PR #2503, opened 2014-11-26
bug 948644 - add filename and version column to missing_symbols

rhelmer:
r? @selenamarie - technically we don't need these to replace modulelist, but anticipate other reports based on this data (see bug for details)

### DIFF
--- a/socorro/external/postgresql/models.py
+++ b/socorro/external/postgresql/models.py
@@ -1545,10 +1545,13 @@ class MissingSymbols(DeclarativeBase):
     __tablename__ = 'missing_symbols'
 
     date_processed = Column(u'date_processed', DATE(), nullable=False)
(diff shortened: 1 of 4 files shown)


socorro/external/postgresql/models.py
```diff
@@ -1545,10 +1545,13 @@ class MissingSymbols(DeclarativeBase):
     __tablename__ = 'missing_symbols'
 
     date_processed = Column(u'date_processed', DATE(), nullable=False)
-    debug_file = Column(u'debug_file', TEXT(), nullable=True, primary_key=True)
+    filename = Column(u'filename', TEXT(), nullable=True)
+    debug_file = Column(u'debug_file', TEXT(), nullable=False)
```
selenamarie: This is going to need a rebase as I'm removing the PRIMARY KEY on this column in another PR.

peterbe:
Needs rebasing.

PR #2506, opened 2014-12-05
Fixes Bug1097988 dataservices in api

selenamarie:
This patch implements a dataservice app that loads configuration and objects with configman and points two crashstats.models classes (Bugs and SignaturesByBugs) at those objects.

I created a decorator for caching: `@memoize` in `webapp-django/crashstats/crashstats/models.py`.  It's purpose is to support `CACHE_MIDDLEWARE` and `CACHE_MIDDLEWARE_FILES` config options for the django app.  I removed the 'dont_cache' option as it is no longer necessary. The decorator will be used on every method that needs it and left off of the methods that don't. The previous implementation funnelled all RESTful functions through the same `fetch` function that implemented the caching. 

Config will be done in `webapp-django/crashstats/settings/local.py` in a dict whose default will be in the corresponding `-dist` file. The local.py on stage/production will need to be updated before pushing code. The change should not cause a problem with *existing* code at all.

@peterbe r?


webapp-django/crashstats/crashstats/models.py
```diff
@@ -161,7 +161,96 @@ def inner(*args, **kwargs):
     return inner
 
 
+def memoize(function):
+    """Decorator for model methods to cache in memory or the filesystem
+    using CACHE_MIDDLEWARE and/or CACHE_MIDDLEWARE_FILES Django config"""
+
+    @functools.wraps(function)
+    def memoizer(*args, **kwargs):
```
selenamarie: Ah yes, that makes it more clear.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -161,7 +161,96 @@ def inner(*args, **kwargs):
...
+
+        def get_cached_result(key):
```
selenamarie: Yes! That was an error in refactoring on my part, rather than intentional. Thanks for catching it.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -161,7 +161,96 @@ def inner(*args, **kwargs):
...
+
+        def refresh_caches(result):
+            cache.set(key, result, instance.cache_seconds)
+            if cache_file and settings.CACHE_MIDDLEWARE_FILES:
+                if not os.path.isdir(os.path.dirname(cache_file)):
+                    os.makedirs(os.path.dirname(cache_file))
+                if instance.expect_json:
+                    json.dump(result, open(cache_file, 'w'), indent=2)
+                else:
+                    open(cache_file, 'w').write(result)
```
selenamarie: Oh good idea. Thanks.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -161,7 +161,96 @@ def inner(*args, **kwargs):
...
+            key = hashlib.md5(stringified_args).hexdigest()
+            cache_file = get_cache_filename(key)
+
+            result = cache.get(key)
+            if result is None:
+                # Didn't find it in the cache, so run our function
+                result = function(*args, **kwargs)
+                refresh_caches(result)
+            return result
+
```
selenamarie: Ahhh yeah, good catch. That is much nicer. Thanks.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -161,7 +161,96 @@ def inner(*args, **kwargs):
...
+            if instance.expect_json:
+                cache_file = os.path.join(
+                    cache_file,
+                    '%s.json' % key
+                )
+            else:
+                cache_file = os.path.join(
+                    cache_file,
+                    '%s.dump' % key
+                )
```
selenamarie: It was repeated twice. I thought I was copying existing functionality, but probably missed something.

I like your suggested version much better.

socorro/external/postgresql/service_base.py
```diff
@@ -31,6 +31,30 @@ class PostgreSQLWebServiceBase(DataserviceWebServiceBase):
...
+            'hits': (
+                'id',
+                'signature',
+            )
+        },
+    )
+    required_config.add_option(
+        'required_params',
+        default=('signatures',),
+    )
```
peterbe: These two options above don't belong in the service_base.py

selenamarie: oh yeah.. sigh.. /me moves them back

webapp-django/crashstats/settings/local.py-dist
```diff
@@ -168,3 +168,19 @@ SECRET_KEY = 'you must change this'
...
+build_path = os.environ.get(
+    'TRAVIS_BUILD_DIR',
+    os.environ.get(
+        'WORKSPACE',  # Jenkins
+        '..'  # development where webapp is in a sub-directory of socorro
+    )
+)
+
+DATASERVICE_INI = build_path + '/config/dataservice.ini'
+assert os.path.isfile(DATASERVICE_INI), DATASERVICE_INI
```
peterbe: Let's move all of this to `settings/base.py`.

webapp-django/crashstats/settings/local.py-dist
```diff
@@ -168,3 +168,19 @@ SECRET_KEY = 'you must change this'
...
 #    'label': 'File a bug in bugzilla'
 #}
+#
+# We're adding configman to django, and so we need to specify a path 
+# to the .ini files needed. The setting below is appropriate for local
+# development where you invoke manage.py in the webapp-django/ directory
+# this configuration is for our travis environment
+import os
+build_path = os.environ.get(
+    'TRAVIS_BUILD_DIR',
```
peterbe: We should perhaps move this to one of the tests. Don't we have something specific in the ./scripts/ files that is specific to travis?

peterbe: Maybe not.

webapp-django/crashstats/settings/local.py-dist
peterbe: Basically, `settings/local.py` is just for exceptional overrides and things, like passwords, that can never be in `settings/base.py`.

scripts/bootstrap.sh
```diff
@@ -36,6 +36,15 @@ make
...
 
+# setup any unset test configs and databases without overwriting existing files
+pushd config
+for file in *.ini-dist; do
+    if [ ! -f `basename $file -dist` ]; then
+        cp $file `basename $file -dist`
+    fi
+done
+popd
+
```
peterbe: If we no longer are going to depend on a `config/dataservice.ini` file, can we move this back into `scripts/test.sh`

socorro/external/postgresql/service_base.py
peterbe: sorry :)

peterbe:
I'm r+ on the code as it appears all my many nits have been addressed (apart from some Bug specific options being service_base.py)

I haven't tested it locally on my laptop yet. I'm going to do that tomorrow.

selenamarie:
@peterbe OK I have this working on my dev instance :) 

The instructions I made a while back for testing dataservice are here: https://gist.github.com/selenamarie/fec696bf042164ea123e

selenamarie:
@peterbe alright! I have updated base.py and local.py-dist and I updated our stage config with the changes. I had to add a try/except on stage because the code isn't pushed yet.  

Are we good to merge?

webapp-django/crashstats/settings/local.py-dist
```diff
@@ -168,3 +168,43 @@ SECRET_KEY = 'you must change this'
...
+# At a minimum, you'll probably want to change db username/password. All
+# dataservice objects inherit resource configuration and so can all
+# have their database resource configuration set once in 'secrets.postgresql'
+# and 'resource.postgresql' keys.
+# Here's an example:
+#from . import base
+#try:
+#    DATASERVICE_CONFIG_BASE = base.DATASERVICE_CONFIG_BASE
+#except AttributeError:
+#    pass
```
peterbe: This doesn't make any sense. 
First of all, we would always have a `base.DATASERVICE_CONFIG_BASE` and if we did get an `AttributeError` it wouldn't later be possible to do `DATASERVICE_CONFIG_BASE.update(`

webapp-django/crashstats/settings/local.py-dist
```diff
@@ -168,3 +168,43 @@ SECRET_KEY = 'you must change this'
...
+#DATASERVICE_CONFIG_BASE.update({
+#    'secrets': {
+#       'postgresql': {
+#           'database_password': 'newpassword',
+#            'database_username': 'newusername',
```
peterbe: Can you add the others too. Like `database_name` and `database_hostname`

webapp-django/crashstats/settings/local.py-dist
selenamarie: If someone was deploying this change and upgrading from a previous version of socorro, and they pushed their config changes before updating their app -- without this try/except block our config would break their app.

peterbe:
Two last comments. But if you fix that, r+

webapp-django/crashstats/settings/local.py-dist
peterbe: Why would why they have a `local.py-dist` that is more up to date than
`settings/base.py`?!


On Mon, Jan 12, 2015 at 2:30 PM, Selena Deckelmann <notifications@github.com
> wrote:

> In webapp-django/crashstats/settings/local.py-dist
> <https://github.com/mozilla/socorro/pull/2506#discussion-diff-22828498>:
>
> > @@ -168,3 +168,43 @@ SECRET_KEY = 'you must change this'
> >  #           'component=General&groups=client-services-security',
> >  #    'label': 'File a bug in bugzilla'
> >  #}
> > +
> > +
> > +# Dataservice API configuration
> > +# Extend dataservices settings from settings/base.py here
> > +# At a minimum, you'll probably want to change db username/password. All
> > +# dataservice objects inherit resource configuration and so can all
> > +# have their database resource configuration set once in 'secrets.postgresql'
> > +# and 'resource.postgresql' keys.
> > +# Here's an example:
> > +#from . import base
> > +#try:
> > +#    DATASERVICE_CONFIG_BASE = base.DATASERVICE_CONFIG_BASE
> > +#except AttributeError:
> > +#    pass
>
> If someone was deploying this change and upgrading from a previous version
> of socorro, and they pushed their config changes before updating their app
> -- without this try/except block our config would break their app.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mozilla/socorro/pull/2506/files#r22828498>.
>



-- 
Peter Bengtsson
Mozilla Web Engineering

webapp-django/crashstats/settings/local.py-dist
selenamarie: Yeah, no problem.

webapp-django/crashstats/settings/local.py-dist
selenamarie: Haha - well in our case, we needed to push `local.py` out before we pushed code. Is it that terrible? I don't mind removing it... I only put it in there because it caught me by surprise when I tested the change before I committed it to our puppet repo.

peterbe: That's something we should do in **our** local.py then. Also, it needs indentation to work at all. I mean the `DATASERVICE_CONFIG_BASE.update({` line should be between the `try:` and `except`

PR #2527, opened 2014-12-23
Fixes bug 1114552 - Added Query service to es module.

AdrianGaudebert:
@phrawzty r?

### DIFF
--- a/socorro/external/es/query.py
+++ b/socorro/external/es/query.py
@@ -1,0 +1,86 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import datetime
+import elasticsearch
+import json
+import re
+
+from socorro.external import (
+    BadArgumentError,
+    DatabaseError,
+    MissingArgumentError,
+    ResourceNotFound,
+)
+from socorro.external.es.base import ElasticsearchBase
+from socorro.external.es.supersearch import BAD_INDEX_REGEX
+from socorro.lib import datetimeutil, external_common
+
+
+class Query(ElasticsearchBase):
+    '''Implement the /query service with ElasticSearch. '''
+
+    filters = [
+        ('query', None, 'str'),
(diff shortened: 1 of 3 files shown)


socorro/external/es/query.py
```diff
@@ -0,0 +1,88 @@
...
+        except ValueError:
+            raise BadArgumentError(
+                'query',
+                msg="Invalid JSON value for parameter 'query'"
+            )
+
+        # Set indices.
+        indices = []
+        if not params.indices:
+            # By default, use the last two indices.
```
AdrianGaudebert: The UI actually even fills the input with the last two indices by default.

socorro/external/es/query.py
```diff
@@ -0,0 +1,88 @@
...
+            today = datetimeutil.utc_now()
+            last_week = today - datetime.timedelta(days=7)
+
+            indices = self.generate_list_of_indexes(last_week, today)
+        elif len(params.indices) == 1 and params.indices[0] == 'ALL':
+            # If we want all indices, just do nothing.
+            pass
+        else:
```
AdrianGaudebert: It will fail because, most probably, ``ALL`` is not a valid index.

socorro/external/es/query.py
```diff
@@ -0,0 +1,88 @@
...
+            results = connection.search(
+                body=query,
+                **search_args
+            )
+        except elasticsearch.exceptions.NotFoundError, e:
+            missing_index = re.findall(BAD_INDEX_REGEX, e.error)[0]
+            raise ResourceNotFound(
+                "elasticsearch index '%s' does not exist" % missing_index
+            )
+        except elasticsearch.exceptions.TransportError, e:
```
AdrianGaudebert: I'm not too keen on changing the features during this rewrite, as you know. Plus I am not sure what the value of a retry would be here, like, if we add it here why not add it to all other services? I'm sure users don't mind hitting F5 that much. :)

AdrianGaudebert:
@phrawzty comment addressed!

phrawzty:
`r+` pending Jenkins. :fish_cake:

PR #2537, opened 2015-01-02
fixes Bug 1117168 extended quit check to transform rules

twobraids:
Processor2015 is not as responsive to SIGTERM as the HybridProcessor was.  That's because the TransformRule system doesn't honor the quit_check callback function.  

This PR extends the quit_check call back function to the TransformRule system


rhelmer:
r+

PR #2540, opened 2015-01-04
fixes Bug 1117590 - removed bogus module on exception

twobraids:
the code that moved the delete of the temporary dump files somehow acquired a module qualifier.  and that prevents the processor from logging file deletion problems.

this is the fix.


lonnen:
+1

PR #2542, opened 2015-01-05
added logging of saved to transform functions

twobraids:
the old HBase modules logged the "saved" message.  I've instead moved that to the transform functions rather than than crashstorage classes.


rhelmer:
r+

PR #2545, opened 2015-01-05
fixes Bug 1117961 - ensure that all passwords are tagged as secret

twobraids:
configman can offer sensitive status to options that should not be echoed to stdout, stderr or logs.  Make sure that all Socorro apps use that tagging for sensitive options


phrawzty:
dance

PR #2546, opened 2015-01-05
Fixes bug 1117907 remove unneeded indexes and fks

selenamarie:
This removes a bunch of things we tried out in prod but aren't being used. A search of index scans and index tuple fetches reveals that at most, one of the following indexes was used at most 4 times over its life going back to early 2014.

RIP unused indexes!

The foreign keys are also unused and unnecessary and cluttering up the database.


selenamarie:
@rhelmer r?

rhelmer:
r+

PR #2548, opened 2015-01-06
Bug 1111536 - Improved get_mapping unit tests.

AdrianGaudebert:
The bug was invalid, but better unit tests are better. 

@phrawzty or @peterbe  r?


phrawzty:
Mo' tests mo' better.  `r+`

tests

PR #2550, opened 2015-01-06
bug 1118446 -  Implement HTTP symbol fetching in stackwalker

luser:
There's better documentation in https://github.com/luser/socorro/commit/cc2849ed2eb35c56df567748c6f9f095958eca69 for what I actually implemented here.

This allows using the stackwalker in the current mode unless we specify the new options, and we should be able to transition to the S3 storage incrementally.

### DIFF
--- a/minidump-stackwalk/http_symbol_supplier.cc
+++ b/minidump-stackwalk/http_symbol_supplier.cc
@@ -1,0 +1,275 @@
+// Copyright (c) 2011 The Mozilla Foundation
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//     * Neither the name of The Mozilla Foundation nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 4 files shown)


rhelmer:
This lgtm, @twobraids want to take a look over the C++ ?

minidump-stackwalk/http_symbol_supplier.cc
```diff
@@ -0,0 +1,319 @@
...
+    dirs.push_back(dir);
+    string new_dir = dirname(dir);
+    if (new_dir == dir || dir.empty()) {
+      break;
+    }
+    dir = new_dir;
+  }
+  for (auto d = dirs.rbegin(); d != dirs.rend(); d++) {
+    if (mkdir(d->c_str(), 0755) != 0) {
+      BPLOG(ERROR) << "Error creating " << dir << ": " << errno;
```
twobraids: feeling a need to at least say something about this code: it might be nice in logging the problem here to say what subdir actually failed by indicating where in the iterator we were when mkdir failed.  

and a question prompted by my rusty C++ skills, the iterator in the for loop, is it just a ```size_t``` or is an object of some sort?  If it is an object, is its copy constructor being invoked by use of the post increment rather than pre increment in the loop?  or is that not an issue in modern C++?

luser: That's a good point, I should log the exact dir that couldn't be created.

rbegin() returns a reverse iterator, so it's an object. The postincrement operator on that just mutates its internal state, iterator traversal should be plenty fast in modern C++.

twobraids: actually speed isn't my concern, it's mainly a refresher in my own understanding of C++.  

Twenty years ago doing code review a Merrill Lynch, demonstrated a "gotcha" to a skeptical audience about the use preincrement vs postincrement for iterators that encapsulated cursors in relational databases.  Using post increment actually invoked the copy constructor, spawning a new cursor and throwing it away on each iteration. Is that sort of unintended side effect something that we still have to watch out for, or does the compiler optimize that away these days?

luser: This is still true, although I suspect (but haven't verified) that the compiler can optimize it away. I'll fix it anyway as I'm using preincrement everywhere else (probably from the Google style guide for that very reason).

PR #2551, opened 2015-01-07
fix bug 1118224 - check for existing buildID before trying to insert

rhelmer:
r? @selenamarie

### DIFF
--- a/socorro/cron/jobs/ftpscraper.py
+++ b/socorro/cron/jobs/ftpscraper.py
@@ -432,6 +432,7 @@ class FTPScraperCronApp(BaseCronApp, ScrapersMixin):
                         self._is_final_beta(version)
                         and build_type == 'release'
                         and version > '26.0'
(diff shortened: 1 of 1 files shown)


selenamarie:
Is the issue is that we can have a build that doesn't have a buildID but is a release? I couldn't get sentry to reset my password, so I haven't seen the exact error that triggered this issue. 

r+ for the code.. just trying to also understand what was wrong

rhelmer:
@selenamarie it's actually that we're picking up other files like```.mozinfo.json```. A warning is printed at https://github.com/mozilla/socorro/blob/master/socorro/cron/jobs/ftpscraper.py#L109 when this happens, but it doesn't actually skip the file.

This patch makes the caller skip if it happens, just like in https://github.com/mozilla/socorro/blob/master/socorro/cron/jobs/ftpscraper.py#L415

PR #2552, opened 2015-01-08
bug 1119223 - run sphinx in travis

peterbe:
Making a PR to see if this works as expected in travis.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -14,6 +14,8 @@
 
 ./scripts/integration-test.sh
 
+./scripts/docs.sh
+
 ./scripts/analysis.sh
 
 ./scripts/install.sh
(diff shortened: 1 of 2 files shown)


peterbe:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1119223#c2

I decided this was a bad idea. I don't want to install spinx (and all its dependencies) into our build process. However, if it was possible to set up a temporary virtualenv and run `pip install -r docs/requirements.txt` then I think we could make the spinx build part of the travis build.

peterbe:
@rhelmer Suppose I didn't want to use `source "$VIRTUAL_ENV/bin/activate"` in this patch and instead something like `virtualenv docs-env; source docs-env/bin/activate; pip install -r ...` would that be a good idea?

peterbe:
It's a bad idea. Let's do it with a webhook instead. 
screenshot 2015-01-09 15 30 46

Let's travis overhead and noise this way.

PR #2553, opened 2015-01-08
No Bug - added new crash source for fs

twobraids:
as per request by @rhelmer, a restoration of the ability for the processor to use the FS classes as a new crash source

### DIFF
--- a/socorro/unittest/external/fs/test_fs_new_crash_source.py
+++ b/socorro/unittest/external/fs/test_fs_new_crash_source.py
@@ -1,0 +1,54 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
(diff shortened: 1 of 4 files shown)


socorro/unittest/external/fs/test_fs_new_crash_source.py
```diff
@@ -0,0 +1,54 @@
...
+        for a_crash_id in range(10):
+            yield str(a_crash_id)
+
+    def ack_crash(self, crash_id):
+        return crash_id
+
+
+#==============================================================================
+class TestConnection(TestCase):
+    """Test PostgreSQLBase class. """
```
rhelmer: Should that be ```s/PostgreSQLBase/FSNewCrashSource/``` ?

twobraids:
@rhelmer, you are, of course, correct. As this test file comes directly from the RabbitMQ tests, I looked there and found the same misidentified class.  I've corrected both locations.

rhelmer:
r+ this wfm

PR #2556, opened 2015-01-09
Fixes bug 1119594: Deprecate old lib and scripts that depended on them

selenamarie:
@twobraids r?


twobraids:
oh, yes, please.  r+

PR #2557, opened 2015-01-09
Fixes nobug issue with SQL for FK drops

selenamarie:
The SQL in this migration was incorrect!  Now corrected with testing via IRC from a user that was affected.


selenamarie:
@rhelmer r?

rhelmer:
r+

PR #2558, opened 2015-01-09
Fixes bug 1119840 - remove PK from missing_symbols

selenamarie:
No migration included as we never created a primary key in production.


selenamarie:
@rhelmer r?

rhelmer:
r+ thanks for fixing my mistake :)

PR #2559, opened 2015-01-12
bug 1120551 - bump forced versions correlation thingy for ff 36

lonnen:
should go out today or tomorrow. @KaiRo-at can you double check the versions override when you get a chance?


KaiRo-at:
This looks good, r=me - and thanks for thinking of it!

PR #2560, opened 2015-01-13
Fixes bug 1120953 - Fixed errors with Super Search in new es module.

AdrianGaudebert:
@phrawzty r?

### DIFF
--- a/socorro/unittest/external/es/test_supersearch.py
+++ b/socorro/unittest/external/es/test_supersearch.py
@@ -561,6 +561,16 @@ class IntegrationTestSuperSearch(ElasticsearchTestCase):
             _facets=['unkownfield']
         )
 
+    def test_get_missing_index(self):
+        config = self.get_mware_config(es_index='socorro_test_reports_%W')
+        api = SuperSearch(config=config)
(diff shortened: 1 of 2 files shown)


socorro/unittest/external/es/test_supersearch.py
```diff
@@ -561,6 +561,16 @@ def test_get_with_facets(self):
             _facets=['unkownfield']
         )
 
+    def test_get_missing_index(self):
```
phrawzty: Minor nit, but "missing" is ambiguous here.  To be clear, this means "test what happens if a request for a non-existent index is made", correct?  If so, prefer something like `test_get_nonexistent_index`.  Again, minor, but to strive for clarity is a noble and virtuous pursuit. :haircut:

phrawzty:
`r+` pending Jenkins.

socorro/unittest/external/es/test_supersearch.py
AdrianGaudebert: How about ``test_get_against_missing_index``? I don't like the words "non-existent". :-)

PR #2561, opened 2015-01-13
Fixes bug 1120984 create 'make dev' instead of bootstrap

selenamarie:
@rhelmer not exactly what we talked about, but one way of handling this. r?


rhelmer:
r+

PR #2564, opened 2015-01-14
fixes bug 1119347 - Admin creation of API tokens

peterbe:
r? @rhelmer 

@AdrianGaudebert wanna take a quick look too?

### DIFF
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -30,7 +30,6 @@ from crashstats.api.cleaner import Cleaner
 from crashstats.dataservice import models
 models = models  # silence pyflakes
 
-
 logger = logging.getLogger('crashstats_models')
 
 
--- a/webapp-django/crashstats/manage/forms.py
+++ b/webapp-django/crashstats/manage/forms.py
@@ -2,8 +2,10 @@ import datetime
 
 from django.contrib.auth.models import User, Group, Permission
 from django import forms
+from django.utils import timezone
 
 from crashstats.crashstats.forms import BaseForm, BaseModelForm
+from crashstats.tokens.models import Token
 
 
 class SkipListForm(BaseForm):
@@ -38,6 +40,82 @@ class FilterEventsForm(BaseForm):
 
     user = forms.CharField(required=False)
     action = forms.CharField(required=False)
+
+
+class FilterAPITokensForm(BaseForm):
+
+    user = forms.CharField(required=False)
+    key = forms.CharField(required=False)
+    expired = forms.ChoiceField(required=False, choices=(
+        ('', 'All'),
+        ('yes', 'Expired'),
+        ('no', 'Not expired'),
+    ))
+
+
+class APITokenForm(BaseModelForm):
+
+    user = forms.CharField(required=True)
+    expires = forms.ChoiceField(required=True)
+
+    class Meta:
+        model = Token
+        fields = ('user', 'expires', 'notes', 'permissions')
+
+    def __init__(self, *args, **kwargs):
+        self.possible_permissions = kwargs.pop('possible_permissions', [])
+        expires_choices = kwargs.pop('expires_choices', [])
+        super(APITokenForm, self).__init__(*args, **kwargs)
+        self.fields['permissions'].choices = (
+            (x.pk, x.name) for x in self.possible_permissions
+        )
+        self.fields['user'].widget = forms.widgets.TextInput()
+        self.fields['user'].label = 'User (by email)'
+        self.fields['expires'].choices = expires_choices
+
+    def clean_user(self):
+        value = self.cleaned_data['user']
+        try:
+            user = User.objects.get(email__istartswith=value.strip())
+            if not user.is_active:
+                raise forms.ValidationError(
+                    '%s is not an active user' % user.email
+                )
+            return user
+        except User.DoesNotExist:
+            raise forms.ValidationError('No user found by that email address')
+        except User.MultipleObjectsReturned:
+            raise forms.ValidationError(
+                'More than one user found by that email address'
+            )
+
+    def clean_expires(self):
+        value = self.cleaned_data['expires']
+        return timezone.now() + datetime.timedelta(days=int(value))
+
+    def clean(self):
+        cleaned_data = super(APITokenForm, self).clean()
+        if 'user' in cleaned_data and 'permissions' in cleaned_data:
+            user = cleaned_data['user']
+            for permission in cleaned_data['permissions']:
+                if not user.has_perm('crashstats.' + permission.codename):
+                    only = [
+                        p.name for p in self.possible_permissions
+                        if user.has_perm('crashstats.' + p.codename)
+                    ]
+                    msg = (
+                        '%s does not have the permission "%s". ' % (
+                            user.email,
+                            permission.name
+                        )
+                    )
+                    if only:
+                        msg += ' Only permissions possible are: '
+                        msg += ', '.join(only)
+                    else:
+                        msg += ' %s has no permissions!' % user.email
+                    raise forms.ValidationError(msg)
+        return cleaned_data
 
 
 class GroupForm(BaseModelForm):
--- a/webapp-django/crashstats/manage/static/manage/js/api_tokens.js
+++ b/webapp-django/crashstats/manage/static/manage/js/api_tokens.js
@@ -1,0 +1,194 @@
+/*global alert:true location:true PaginationUtils:true */
+(function ($, document) {
+    'use strict';
+
+    function start_loading() {
+        $('.pleasewait').show();
+    }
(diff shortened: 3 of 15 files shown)


peterbe:
The reason I added the `webapp-django/__init__.py` file is because `coverage` won't work unless you have this.

AdrianGaudebert:
Your branch needs to be rebased

webapp-django/crashstats/manage/forms.py
```diff
@@ -40,6 +42,84 @@ class FilterEventsForm(BaseForm):
...
 
+class FilterAPITokensForm(BaseForm):
+
+    user = forms.CharField(required=False)
+    key = forms.CharField(required=False)
+    expired = forms.CharField(required=False)
+
+    def clean_expired(self):
+        value = self.cleaned_data['expired']
+        if value not in ('yes', 'no', ''):
```
AdrianGaudebert: Why not make ``expired`` a ``ChoiceField`` instead?

webapp-django/crashstats/manage/forms.py
```diff
@@ -40,6 +42,84 @@ class FilterEventsForm(BaseForm):
...
+            (x.pk, x.name) for x in self.possible_permissions
+        )
+        self.fields['user'].widget = forms.widgets.TextInput()
+        self.fields['user'].label = 'User (by email)'
+        self.fields['expires'].choices = expires_choices
+
+    def clean_user(self):
+        value = self.cleaned_data['user']
+        try:
+            user = User.objects.get(email__istartswith=value.strip())
```
AdrianGaudebert: If we have two users, one with email ``me@mail.co`` and the other with email ``me@mail.com``, the first one won't ever be able to use this form, because that query will always return 2 results. I think it would be better to make sure that this query cannot raise ``User.MultipleObjectsReturned``.

webapp-django/crashstats/manage/static/manage/js/api_tokens.js
```diff
@@ -0,0 +1,228 @@
...
+        $('.pleasewait').show();
+    }
+
+    function stop_loading() {
+        $('.pleasewait').hide();
+    }
+
+    function fetch() {
+
+        function show_pagination_links(count, batch_size, page) {
```
AdrianGaudebert: This looks like logic that already exists in other scripts and could be shared. Would it be possible to be more DRY here?

webapp-django/crashstats/manage/views.py
```diff
@@ -789,3 +792,140 @@ def _get_edit_url(action, extra):
...
+                expires=data['expires'],
+            )
+            for permission in data['permissions']:
+                token.permissions.add(permission)
+
+            messages.success(
+                request,
+                'API Token for %s created. ' % token.user.email
+            )
+            return redirect('manage:api_tokens')
```
AdrianGaudebert: Don't you want to return something if the form is not valid?

webapp-django/crashstats/manage/views.py
```diff
@@ -789,3 +792,140 @@ def _get_edit_url(action, extra):
...
+    except (ValueError, AssertionError):
+        return http.HttpResponseBadRequest('invalid page')
+
+    items = []
+    batch_size = settings.API_TOKENS_ADMIN_BATCH_SIZE
+    batch = Paginator(tokens.select_related('user'), batch_size)
+    batch_page = batch.page(page)
+
+    # build up a dict of permission id -> permission name
+    _permissions_names = {}
```
AdrianGaudebert: Why did you put an underscore in those variable names?

AdrianGaudebert:
A few concerns, but overall the code looks good. I haven't tested it locally though, I leave that to @rhelmer. :-)

webapp-django/crashstats/manage/forms.py
peterbe: You are right. Done.

webapp-django/crashstats/manage/static/manage/js/api_tokens.js
peterbe: You're creating more work for me :(
You unfortunately you're right. I'll do some refactoring.

webapp-django/crashstats/manage/views.py
peterbe: It already does that. If NOT `if form.is_valid():`

webapp-django/crashstats/manage/views.py
peterbe: I guess it's a peterism. I use underscore to signify that the variable is only relevant in a temporary manner. It's not a variable that I want to keep and return as a context variable. 
Do you think it's ugly?

peterbe:
@AdrianGaudebert I'm assigning the review over to you unless @rhelmer beats you to it once the tests pass.

lonnen:
I want to hold on this for a minute until peter and I can work through token stuff this afternoon. It will be hard to remove this once we land it and people start relying on it.

lonnen:
k. I'm converted. This is a good idea.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -24,7 +24,6 @@
 from crashstats import scrubber
 from crashstats.api.cleaner import Cleaner
 
-
```
peterbe: sorry. this was me mucking around with some hacks that I later undid.

webapp-django/crashstats/manage/views.py
AdrianGaudebert: No, it's fine. I was just wondering what the idea was.

webapp-django/crashstats/manage/views.py
AdrianGaudebert: I don't understand your answer, here's the structure I see here: 

```python
if request.method == 'POST':
    if form.is_valid():
        # do something if the form is valid
else:
    # do something if not using POST
```

If ``form.is_valid()`` returns a falsy value, that's not handled. Unless I am not looking at it correctly, but then I'd love an explanation because it is not obvious. :)

peterbe: That's actually a super common pattern in Django. If you look around https://github.com/mozilla/airmozilla/tree/master/airmozilla/manage/views you'll see lots of these. E.g. https://github.com/mozilla/airmozilla/blob/master/airmozilla/manage/views/locations.py#L122-L131

Let's beef up your example code:
```python
if request.method == 'POST':
    form = MyForm(request.POST)
    if form.is_valid():
        # do something if the form is valid
       return redirect('...')  # this is important
else:
    # do something if not using POST
    form = MyForm(initial={'some': 'value'})
return render(request, 'mytemplate.html', {'form': form})
```
And the template looks like this:
```html
<html>
<form action="." method="post">
{{ form }} <!-- Note! that we let django render the whole form for itself -->
<input type="submit">
</form>
</html>
```
So, initially the view is a GET and then it will fall in that `else:` block. 
The django form instance will turn into HTML automatically. 
E.g. 
```html
<label id="id_some">Some:</label>
<input name="some" id="id_some" value="value">
```
Now suppose you fill something into that form and submit the form. Now the view function is called with a POST. 
Ok. Suppose the `MyForm` class for some reason doesn't like one of the submitted values, then when it gets to the line: `return render(request, 'mytemplate.html', {'form': form})` that `form` instance variable will be a "bound form". It means it has data and it has run the validation. 
And again, it will render the template. 
Now, this time, the django form instance is smart about how it *renders itself*. Instead of the above snippet it'll render itself like this:
```html
<label id="id_some">Some:</label>
<ul class="errorlist">
    <li>Some form validation error message</li>
</ul>
<input name="some" id="id_some" value="entered value">  <!-- assuming you typed in "entered value" -->
```
If you have time (Ha! who does?!) I would recommend you whip up something similar and see it in action.

rhelmer:
r+ code looks good, and WFM locally.

webapp-django/crashstats/manage/views.py
AdrianGaudebert: OK I get it now. Thanks for this awesome explanation Peter!

webapp-django/crashstats/manage/forms.py
AdrianGaudebert: @peterbe You didn't address this nor explained why you didn't. Should I be worried? :-)

peterbe: You should be worried. :) 
Sort of. 
My excuse, if you can call it that, was I that I forgot to address it. I don't think I want(ed) change the code but I should have responded. 

The example you mention is extremely rare. That you have users with `@mail.co` and `@mail.com` on the same project is super rare. 

Let's address this if and when it becomes a problem. 

Also, it wouldn't have been a problem if select2 didn't have a bug that prevented me from writing a select2 widget that uses AJAX and can accept 1 value into a type input field. It's a [known open bug](https://github.com/select2/select2/issues/2800).

PR #2566, opened 2015-01-14
fixes Bug 1121462 - add generic external process rule

twobraids:
this PR provides two processor rules (though does not actually set them up for use - separate bug for that).  The first rule is a generic base class for rules that need to invoke an external process.  The second rule is an implementation that uses the base class to invoke ted's dump-lookup tool.


twobraids:
@rhelmer r?

rhelmer:
r+

PR #2568, opened 2015-01-14
fixes bug 1121749 - Unable to nose test specific files due to models.py loading

peterbe:
@rhelmer or @selenamarie r?


selenamarie:
Ahhh r+ thank you for fixing this.

PR #2569, opened 2015-01-15
fixes bug 1119378 - Upgrade to Django 1.6

peterbe:
@rhelmer r?

### DIFF
--- a/webapp-django/crashstats/api/views.py
+++ b/webapp-django/crashstats/api/views.py
@@ -99,10 +99,6 @@ class APIWhitelistError(Exception):
 class MultipleStringField(forms.TypedMultipleChoiceField):
     """Field that do not validate if the field values are in self.choices"""
 
-    # def to_python(self, value):
-    #     """Override checking method"""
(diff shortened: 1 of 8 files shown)


peterbe:
I've tested it a bunch locally and it works fine. One big change was around the list type inputs for test driving the API and that works.

webapp-django/crashstats/api/views.py
```diff
@@ -99,9 +99,9 @@ class APIWhitelistError(Exception):
 class MultipleStringField(forms.TypedMultipleChoiceField):
     """Field that do not validate if the field values are in self.choices"""
 
-    def to_python(self, value):
-        """Override checking method"""
-        return map(self.coerce, value)
+    # def to_python(self, value):
+    #     """Override checking method"""
+    #     return map(self.coerce, value)
```
rhelmer: why is this commented out?

peterbe: Because I'm sloppy. Will fix.

rhelmer:
r+

PR #2571, opened 2015-01-15
Fixes bug 1117911 crontabber to truncate raw/processed crash partitions >2 weeks

selenamarie:
We may not want to make this a default for our open source users.

Maybe what I should do is add configuration that sets the default number of weeks to something much larger, and then our local config is 2 weeks.  Thoughts?


selenamarie:
@rhelmer r?

rhelmer:
@selenamarie code lgtm - I agree on making the default longer than 2 weeks (maybe 6 months?) and overriding it in config.

Would you mind mentioning this in the docs too? A lot of people ask how they can automatically purge old data so I think it'll be useful, but surprising to some (especially after ```reports``` goes away)

selenamarie:
This change should be accompanied by a crontabber.ini update in stage and prod.

PR #2572, opened 2015-01-15
Nobug refactor duplicates sql

selenamarie:
Refactoring so this is a bit easier to read for future dbas.


selenamarie:
@twobraids r? this is just to make this procedure easier to read.

selenamarie:
Needs an edit for alembic before merging

twobraids:
r+ I see no problems

PR #2576, opened 2015-01-16
bug 1119525 - fix get-minidump-instructions build in vagrant

luser:
Some simple workarounds to support GCC 4.4.

PR #2577, opened 2015-01-16
fix bug 1122587; add libcurl-devel & stackwalker patch

phrawzty:
stackwalker needs `libcurl-devel` to compile.

`r?` @rhelmer
cc @luser


phrawzty:
Travis will fail - there needs to be a patch for stackwalker as well.

phrawzty:
Dropping this PR in favour of #2576 since @luser and I are now walking on each other's toes.

PR #2578, opened 2015-01-16
Fixes bug 1122311 remove simplejson from requirements

selenamarie:
Editor removed some whitespace issues as well. I didn't do a full PEP8 on these.. it would be a pretty major on the scripts and I don't have a test environment for the hbase/s3 scripts.


selenamarie:
@peterbe we pyelasticsearch depends on simplejson :/ So, I can't remove it until we get rid of pyelasticsearch -- which I understand is in the process of being deprecated.

r? anyway?

peterbe:
r+ but...

Perhaps you want to take this opportunity to upgrade simplejson in requirements.txt. We're currently relying on an outdated version.

PR #2581, opened 2015-01-16
bug 1122319 - Start using south migrations

peterbe:
@rhelmer r?

So, one caveat is that when you go from not-using-south to start using it you basically need to run a one-off command for an installed app (namely waffle). See https://bugzilla.mozilla.org/show_bug.cgi?id=1122319#c1

Another caveat is that once this is deployed and in master, every developer will need to run: `./manage.py migrate tokens 0001 --fake` and `./manage.py migrate symbols 0001 --fake` on his/her computer. That's less scary compared to having to run the `./manage.py migrate waffle --fake` command once on the webheads.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -144,8 +144,6 @@ funfactory==2.3.0
 https://github.com/jbalogh/test-utils/archive/4ae9327.zip#egg=test-utils
 # sha256: o71QUd0X6dUOip8tNiUI6C1wYwEP3AbV4HzjYI54D-Y
 jingo==0.7
-# sha256: nc6AOA28EXa-GdHOg2C5QC4d5SpHNG_3WeV-3zgxHRw
-https://github.com/django-nose/django-nose/archive/10475f1.zip#egg=django-nose
 # sha256: 0p07MkSEjnM0DFU_lgrFzRSlNmEa7tbcC7rj0bZJvZg
 https://github.com/fwenzel/django-sha2/archive/3ba2b47.zip#egg=django-sha2
 # sha256: A1gwhyCNozcxug18_9RjJTmJQa1rctOt-AnP7_yR0PM
@@ -174,3 +172,7 @@ ndg-httpsclient==0.3.2
 # sha256: bTs1EQn0GQzZdA12M4RpkIGl8uuNc9qn5IUAn7ZGGMA
 # sha256: 5PgdU8Uz9r2VJrBH8Ef3sQHCSrFzOcGnrY-YslwQHqs
 pyasn1==0.1.7
+# sha256: 02C9MYmPnfWfb6p4ZVEGW7pFs15-48ObOBtPv-9zkvQ
+South==1.0.2
+# sha256: NmfSakH-wwNkoO9yWAgyylMogC1VP21ucq9awhyzY2U
+django-nose==1.3
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -246,6 +246,12 @@ echo "Running Django syncdb"
     &> /var/log/socorro/django-syncdb.log
 error $? "django syncdb failed `cat /var/log/socorro/django-syncdb.log`"
 
+echo "Running Django migrate tokens"
+/data/socorro/socorro-virtualenv/bin/python \
+    /data/socorro/webapp-django/manage.py migrate --noinput \
+    &> /var/log/socorro/django-migrate.log
+error $? "OperationalError `cat /var/log/socorro/django-migrate.log`"
+
 # move new socorro.tar.gz over old now that the installation was
 # succesful.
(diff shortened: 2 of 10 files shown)


peterbe:
Hmmm... on master `time ./manage.py test`:
```
real	0m6.900s
user	0m6.575s
sys	0m0.314s
````

Using this branch
```
real	0m50.416s
user	0m40.959s
sys	0m9.448s
```
Something bad going on.

**Comment by mythmon** [comment]
Try `FORCE_DB=0 REUSE_DB=1 ./manage.py test`. One of those is for South, the other is for Django nose. I don't know which is which.

**Comment by willkg** [comment]
If you don't need data migrations to run when you're running tests, you can set `SOUTH_TESTS_MIGRATE = False`.

http://south.readthedocs.org/en/latest/unittests.html

Also, I want to say that both south and django-nose override syncdb and so you might have to have them in the right order in `INSTALLED_APPS`. Something like that.

**Comment by peterbe** [comment]
> Also, I want to say that both south and django-nose override syncdb and so you might have to have them in the right order in INSTALLED_APPS. Something like that.

Yeah, if you don't have `django_nose` last, you don't get nose basically.

**Comment by peterbe** [comment]
Sigh. Nothing works. 

I think `SOUTH_TESTS_MIGRATE` is only something you can use if you let South "hijack" the `test` command.

I have the latest `django-nose` and `south` installed. 

Doing `FORCE_DB=0 REUSE_DB=1 ./manage.py test` still takes 50+ seconds (instead 6). 

It's late Friday and I'm getting tired. Perhaps I missed something.

**Comment by peterbe** [comment]
Re-opening so travis restarts.

File: scripts/package/after-install.sh [hunk 1]
```diff
@@ -48,6 +48,14 @@ if [ $? != 0 ]; then
...
 fi
 
+/data/socorro/socorro-virtualenv/bin/python \
+    /data/socorro/webapp-django/manage.py migrate --noinput \
+    &> /var/log/socorro/django-migrate.log
+if [ $? != 0 ]; then
+    echo "WARN could not run django migration"
+    echo "See /var/log/socorro/django-migrate.log for more info"
+fi
+
```
rhelmer: thanks for fixing this up :+1:

scripts/deploy.sh
```diff
@@ -246,6 +246,12 @@ echo "Running Django syncdb"
...
 
+echo "Running Django migrate tokens"
+/data/socorro/socorro-virtualenv/bin/python \
+    /data/socorro/webapp-django/manage.py migrate --noinput \
+    &> /var/log/socorro/django-migrate.log
+error $? "OperationalError `cat /var/log/socorro/django-migrate.log`"
+
 # move new socorro.tar.gz over old now that the installation was
 # succesful.
 mv socorro-new.tar.gz socorro.tar.gz
```
rhelmer: We could probably just delete ```deploy.sh``` now, I'll file a bug

requirements.txt
```diff
@@ -173,3 +173,7 @@ ndg-httpsclient==0.3.2
 # sha256: bTs1EQn0GQzZdA12M4RpkIGl8uuNc9qn5IUAn7ZGGMA
 # sha256: 5PgdU8Uz9r2VJrBH8Ef3sQHCSrFzOcGnrY-YslwQHqs
 pyasn1==0.1.7
+# sha256: 02C9MYmPnfWfb6p4ZVEGW7pFs15-48ObOBtPv-9zkvQ
+South==1.0.2
+# sha256: NmfSakH-wwNkoO9yWAgyylMogC1VP21ucq9awhyzY2U
+django-nose==1.3
```
rhelmer: we already have ```django-nose``` in ```requirements.txt```

rhelmer: This is making ```pip``` complain and not install packages. The old one is a github URL, we should remove that in favor of using the pypi version anyhow.

rhelmer:
Trying this locally - after doing:
```
./manage.py syncdb --noinput
./manage.py migrate tokens 0001 --fake
./manage.py migrate symbols 0001 --fake
./manage.py migrate
```
I get stuck on:
```
django.db.utils.OperationalError: table "waffle_sample" already exists
```

peterbe:
Oh dear, you sent me on a journey @rhelmer :)
I started over and started using postgres for my local instance. I created a bunch of things. Symbols, waffle flags, tokens. Then switched to this branch and ran the necessary **four** commands:

```
./manage.py syncdb --noinput
./manage.py migrate tokens 0001 --fake
./manage.py migrate symbols 0001 --fake
/manage.py migrate waffle --fake
```
The last one you missed in your comment above. 
Now, when you have run those, now you can do `./manage.py migrate` any time.

rhelmer:
r+ @peterbe yep the order in that last comment WFM, thanks!

PR #2583, opened 2015-01-17
(no bug) docs only - do not assume people have a socorro checkout when t...

rhelmer:
...esting install

### DIFF
--- a/docs/systemtest.rst
+++ b/docs/systemtest.rst
@@ -18,13 +18,20 @@ Note: crashme is `currently not working on Mac OS X <https://bugzilla.mozilla.or
 
 Also see: https://developer.mozilla.org/en/Environment_variables_affecting_crash_reporting
 
-If you already have a crash available and wish to submit it, you can
-use the standalone submitter tool (there is an example JSON and dump
-file for your crash are in the "./testcrash" directory you can use)
-::
-  python socorro/collector/submitter_app.py -u http://crash-reports/submit -s ./testcrash/raw/
+You can use the "curl" command to submit a test crash. If you don't
+have a minidump available, you can download a test dump from the
+Socorro repo:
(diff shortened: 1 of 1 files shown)


rhelmer:
Monsoon in IRC pointed out that the system test doc uses things from the repo, but the install instructions don't have you clone the repo, install virtualenv etc.

I think using ```curl``` here would demonstrate more, since it clearly shows that POST is used, and that there are three necessary form fields (```ProductName```, ```Version```, ```upload_file_minidump```) to submit a crash.

docs/systemtest.rst
```diff
@@ -18,13 +18,17 @@ Note: crashme is `currently not working on Mac OS X <https://bugzilla.mozilla.or
...
-If you already have a crash available and wish to submit it, you can
-use the standalone submitter tool (there is an example JSON and dump
-file for your crash are in the "./testcrash" directory you can use)
-::
-  python socorro/collector/submitter_app.py -u http://crash-reports/submit -s ./testcrash/raw/
+You can use the "curl" command to submit a test crash. If you don't
+have a minidump available, you can download a test dump from the
+Socorro repo:
+.. code-block:: bash
+  curl https://raw.githubusercontent.com/mozilla/socorro/master/testcrash/raw/7d381dc5-51e2-4887-956b-1ae9c2130109.dump > test.dump
```
phrawzty: Redirecting STDOUT in this case can have unintended side-effects.  I admit that we're talking weird shell and environment edge cases here, but it's safer to use one of cURL's `-o` or `-O` arguments for this task.

rhelmer:
@phrawzty oh good point on redirecting stdin, fixed that and also reformatted the curl POST a bit.

phrawzty:
`r+` pending Travis.

PR #2584, opened 2015-01-19
style tidying

phrawzty:
Minor style fixes; highly non-critical. :hamster: 

@rhelmer `r?`

### DIFF
--- a/puppet/modules/socorro/manifests/init.pp
+++ b/puppet/modules/socorro/manifests/init.pp
@@ -43,16 +43,16 @@ class socorro::vagrant {
 
   yumrepo {
     'dchen':
-      baseurl  => 'https://repos.fedorapeople.org/repos/dchen/apache-maven/epel-$releasever/$basearch/';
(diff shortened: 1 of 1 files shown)


puppet/modules/socorro/manifests/init.pp
```diff
@@ -62,11 +62,9 @@
...
 
   package {
-    [
-      'apache-maven',
-    ]:
-    ensure  => latest,
-    require => Yumrepo['dchen'];
+    'apache-maven':
+      ensure  => latest,
+      require => Yumrepo['dchen'];
```
rhelmer: Could do this as a separate bug, but we don't need this anymore (since we don't use hadoop/hbase)

rhelmer:
r+

puppet/modules/socorro/manifests/init.pp
phrawzty: I didn't want to alter functionality in any way with this bug - this is purely stylistic.

PR #2585, opened 2015-01-20
[requires.io] dependency update on master branch

peterbe:


requirements.txt
```diff
@@ -1,94 +1,94 @@
...
 # sha256: 2upIE6eJSfnxlBvixW1aviyZqFgjUMrwWYBPtleIrKU
-pyelasticsearch==0.6.1
+pyelasticsearch==0.7.1
 # sha256: QQvqlt8kefozmGxAw-76nB57qVp_cHEcMVhgu-LWbH8
 # sha256: 2Fg3nvWYjUU0u4kJQy1pdCIQCq_ycimdZhM5g2ttrps
-urllib3==1.9.1
+urllib3==1.10
 # sha256: cHay6ATFKumO3svU3B-8qBMYb-f1_dYlR4OgClWntEI
-elasticutils==0.7
+elasticutils==0.10.2
```
AdrianGaudebert: Please revert this, ElasticUtils > 0.7 breaks compatibility with our. And it is deprecated and will be removed this quarter, so no need to upgrade it. Also, it is possible that updating pyelasticsearch will cause problems as well, and as it is also deprecated, you might want to revert it too.

peterbe:
Gah! I didn't make that pull request. I gave requires.io my GitHub access so I could let it find my repos. I didn't know it would make a PR. Annoying!
I'm revoking their app right away.

PR #2587, opened 2015-01-20
Fixes bug 1122145 add fennec release repos to special_product_platforms

selenamarie:
Added documentation to the function and our shipped documentation. Also opening another bug to just rip this junk out and replace with something considerably simpler.

@rhelmer r?


rhelmer:
r+, +1 on doc/comment updates too!

PR #2588, opened 2015-01-20
Fixes Bug 1104317 - adds SignatureRunWatchDog rule to processor

twobraids:
as outlined in Bug 1104317, a signature that includes "::RunWatchdog" is not useful.  Check for that signature and, if found, reprocess the signtaure for Thread-0 instead of the crashing thread, and prepend the string "shutdownhang" to the signature.  Make sure that the new signature doesn't ignore the "ProcessNextEvent" but adds it to the prefix regular expression.

### DIFF
--- a/socorro/processor/signature_utilities.py
+++ b/socorro/processor/signature_utilities.py
@@ -366,6 +366,7 @@ class CSignatureTool(CSignatureToolBase):
           'PORT_.*',
           '_PR_.*',
           'PR_.*',
+          '.*ProcessNextEvent.*',
           'pthread_mutex_lock',
           '_purecall',
           'raise',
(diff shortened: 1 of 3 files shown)


twobraids:
@peterbe r?

socorro/processor/signature_utilities.py
```diff
@@ -683,6 +685,10 @@ def _create_frame_list(
         return frame_signatures_list
 
     #--------------------------------------------------------------------------
+    def _get_crashing_thread(self, processed_crash):
+        return processed_crash.json_dump['crash_info']['crashing_thread']
```
peterbe: We had to do a lot of magic on the webapp because some json dumps were spuriously missing things like this `crashing_thread`. Should we be more careful here?

peterbe: E.g. https://github.com/mozilla/socorro/commit/8b321c1fbf068e897e484aa1c5a020854f39fa56#diff-a7d1c53ffed84fe4a5fee1abc9bd61eaR212
It's very possible I don't know what I'm talking about.

twobraids: such care is not necessary here.  not having the key is a known condition for the signature generation code.  It is expecting a KeyError to trigger the fallback behavior signature generation behavior.

peterbe:
It's pretty hard to review this. My context is very low. But individual lines all look good :)
r+

PR #2589, opened 2015-01-20
Nobug upgrading alembic and sqlalchemy

selenamarie:
This should get us a version of alembic that is ok with multiple branches.


selenamarie:
@peterbe r? - just updating sqlalchemy and alembic, and fixing a long-standing useless warning when a user runs alembic

peterbe:
I don't know what it means to have a blank down-revision but the requirements look good.
r+

selenamarie:
@peterbe it just means there's no parent. i think what happened here was that i had some initial revisions that I deleted, but early versions of alembic didn't catch the error.

PR #2591, opened 2015-01-22
Fixes bug 1124750 - remove edit_product_info()

selenamarie:
Starting down the path of deprecating product_info view


selenamarie:
Probably needs an edit for alembic before merging.

selenamarie:
@peterbe r?

socorro/lib/migrations.py
```diff
@@ -11,17 +11,36 @@
...
     """
-    Takes a list of files and the alembic op object as arguments
-    Load and run CREATE OR REPLACE function commands from files
+    Helper for finding our raw SQL files locally.
+
     Expects files to be in:
         $CWD/socorro/external/postgresql/raw_sql/procs/
     """
     app_path=os.getcwd()
+    return (app_path + '/socorro/external/postgresql/raw_sql/procs/' + filename)
```
peterbe:
```python
return os.path.join(
    os.getcwd(), 
    'socorro/external/postgresql/raw_sql/procs/' + filename
)
```

peterbe: By the way, what's the difference between `os.getcwd()` and simply doing `.`?

socorro/lib/migrations.py
```diff
@@ -11,17 +11,36 @@
...
+    Load and run CREATE OR REPLACE function commands from files
+    """
     for filename in filelist:
-        sqlfile = app_path + '/socorro/external/postgresql/raw_sql/procs/' + filename
+        sqlfile = get_local_filepath(filename)
+        # Capturing "file not exists" here rather than allowing
+        # an exception to be thrown. Some of the rollback scripts
+        # would otherwise throw unhelpful exceptions when a SQL
+        # file is removed from the repo.
+        if local_file_exists(sqlfile) == False:
```
peterbe: What that does is:
```
v = get_local_filepath(v)
v = get_local_filepath(v)
os.path.isfile(v) ??
```
Why not just change it to
```python
if not os.path.isfile(sqlfile):
```

peterbe: Generally, doing `something == True` or `something == False` is considered an anti-pattern and the preferred pattern is to just do `if something:` or `if not something:`. That way it works with `0`, `None`, `False` or `[]` etc.

socorro/lib/migrations.py
```diff
@@ -11,17 +11,36 @@
...
+            print "Did not find %s. Continuing." % sqlfile
```
peterbe: Let's not use `print` for something so important. Also, if you see that switch past in the logs you might not know where it came from. Much better to use. 
```python
import warnings
...
warnings.warn("Did not find %r. Continuing migration." % sqlfile)
```

socorro/lib/migrations.py
peterbe: Using `os.path.join` is better because it's then extremely unlikely to get a path string that contains a double slash like `//`.

selenamarie: Probably nothing in this case... I tend to specify full paths just in case a caller changed directories. I don't think that would happen though here. I'll give it a spin.

peterbe: Depending on `os.getcwd()` makes me nervous. It means that it totally depends on where you are when you kick off the migration. And that shouldn't be necessary because we can figure out what the path is in relation to this file. E.g.
```python
procs_dir = os.path.normpath(os.path.join(
    __file__,
    '../../', 
    'external/postgresql/raw_sql/procs'
))
```

peterbe:
I'm r+ if it works. Some nits here and there. Something might be irrelevant in the big picture.

socorro/lib/migrations.py
selenamarie: Yeah that earlier series of filepath calls was a thinko - cleaning this up.

peterbe: "a thinko" hehe. New word to me but I got it.

socorro/lib/migrations.py
```diff
@@ -10,18 +10,36 @@
...
-    app_path=os.getcwd()
     for filename in filelist:
-        sqlfile = app_path + '/socorro/external/postgresql/raw_sql/procs/' + filename
+        sqlfile = get_local_filepath(filename)
+        # Capturing "file not exists" here rather than allowing
+        # an exception to be thrown. Some of the rollback scripts
+        # would otherwise throw unhelpful exceptions when a SQL
+        # file is removed from the repo.
+        if not os.path.isfile(sqlfile):
+            warnings.warn("Did not find %r. Continuing migration." % sqlfile, UserWarning, 2)
```
peterbe: your pep8 pre-commit checker clearly didn't catch you this time :)

PR #2592, opened 2015-01-22
Fixes bug 1123363 add setup.py develop to make dev

selenamarie:
Longer term fix might be to change bootstrap.sh


rhelmer:
@selenamarie can you remove the ```python setup.py develop``` from the "build from source" doc too pls? :)

rhelmer:
r+

PR #2595, opened 2015-01-26
Fixes bug 1125862 - Removed paireduuid service and associated webapp bits.

AdrianGaudebert:
r? @peterbe or @selenamarie

### DIFF
--- a/socorro/external/postgresql/crashes.py
+++ b/socorro/external/postgresql/crashes.py
@@ -393,7 +393,7 @@ class Crashes(PostgreSQLBase):
                 COUNT(r.id) AS total
         """]
 
-        ## Adding count for each OS
+        # Adding count for each OS
         if hasattr(self.context, 'webapi'):
             context = self.context.webapi
         else:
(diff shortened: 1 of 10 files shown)


rhelmer:
Note that ```reports_clean``` has an hour delay versus ```reports``` which gets crashes inserted real-time.

If this matters, we could use the ```processed_crashes``` table instead, or fetch the crash from S3.

peterbe:
I guess it looks fine. I'm just not clued up enough on this at the moment. I think @selenamarie should review this instead.

AdrianGaudebert:
I don't think the 1 hour delay matters much, this is used by the report/index page to show just a small information about crashes. And it's much simpler to change the code to use reports_clean than processed_crashes. :)

selenamarie:
@AdrianGaudebert individual reports will fail to display if we implement this change because we filter out duplicate reports, which have different UUIDs. We either need to fetch them from `processed_crashes` or find the filtered out crashes in `reports_duplicates` and display only the rows that the crash has in common. I'd guess the proper way of handling this is: 
```
   SELECT duplicate_of as uuid from reports_duplicates where uuid = 'xxx'
```
And if we do find the value in this table, we then use the 'duplicate_of' value instead of `uuid` for filling in the rest of the values.

Briefly, something like: 

```
with duplicate as (
   SELECT 'xxx' as uuid, duplicate_of as dupe from reports_duplicates where uuid = 'xxx'
), 
with original as (
   SELECT uuid from reports_clean where uuid = 'xxx'
)
SELECT coalesce(duplicate.dupe, original.uuid)
FROM duplicate JOIN original USING uuid
...
```

selenamarie:
Another option would be to fetch metadata for a crash from a different crashstorage backend (s3?).

AdrianGaudebert:
Good news, everyone! This service is not used anymore! This is now a PR to remove it entirely. :-)

phrawzty:
good news everyone

peterbe:
rebase?

PR #2597, opened 2015-01-27
Fixes nobug - update pyhs2 to 0.6.0

selenamarie:


selenamarie:
@rhelmer the version of pyhs2 in pypi has now passed the github commit we were pegged on previously so we can commit to 0.6.0 safely for our prod deploy. r?

selenamarie:
and here is the comparison of where we are pegged in the requirements vs the 0.6 release: https://github.com/BradRuderman/pyhs2/compare/48d22...0.6

rhelmer:
r+

PR #2598, opened 2015-01-27
Used for bug 1124872 - script for querying hive ADI

selenamarie:
Stashing this in case someone needs to verify ADI directly from hive again.


selenamarie:
The test failure is due to some pypi issues. Leaving this for now. Feel free to restart the travis job!

PR #2600, opened 2015-01-28
Fixes bug 1053956 deprecate raw_adu table in favor of raw_adi table

selenamarie:
Some cleanup work to deprecate this table. 

I backed up raw_adu from prod yesterday so this migration should be safe to run, provided tests pass.


socorro/external/postgresql/models.py
```diff
@@ -1232,14 +1210,14 @@ class SignatureSummaryArchitecture(DeclarativeBase):
     product_version_id = Column(u'product_version_id', INTEGER(), primary_key=True, nullable=False)
     product_name = Column(u'product_name', TEXT(), nullable=False)
     version_string = Column(u'version_string', TEXT(), nullable=False)
-    report_date = Column(u'report_date', DATE(), primary_key=True, nullable=False, index=True)
+    report_date = Column(u'report_date', DATE(), primary_key=True, nullable=False)
```
rhelmer: Hm why are there a bunch of changes to ```report_date``` column indexes, is that related?

scripts/staging/extractMiniDB.py
```diff
@@ -71,8 +71,8 @@ def rundump(dump_command):
 # and tables with data that needs to be cleaned
 # dump those with no data
 
-matviews = {'raw_adu'
```
rhelmer: hah does this old "mini DB" stuff even work anymore? I doubt it :P probably just remove it in a different bug, updating it now is fine I think

rhelmer:
r+

PR #2601, opened 2015-01-28
fixes bug 1122319 - south migration problem in jenkins

peterbe:
@rhelmer or @AdrianGaudebert r?

I tested this locally. I created a brand new database (e.g `socorro_webapp_new` with `createdb`) and put this in my `settings/local.py`. Then without running either `syncdb` or `migrate` I just went straight to `FORCE_DB=true ./manage.py test` and it works. 

Let's just check this is how it's expected to work in travis.


rhelmer:
r+ if it passes travis too, this makes sense.

PR #2602, opened 2015-01-28
fix bug 1126917; verify gpg keys of repos where possible

phrawzty:
* EPEL is now managed as a packaged install. :smile: 
* Elasticsearch and PGDG have keys associated with them. :relaxed: 
* devtools has no GPG key available. :disappointed: 

I also set up a dep chain where `ca-certificates` -> `yum-plugin-fastestmirror` -> `everything_else`.  This should keep things fairly future-proofed (famous last words).

@rhelmer `r?`


rhelmer:
r+

PR #2603, opened 2015-01-28
Fixes bug 1126920 add 'alembic upgrade heads' to test and deploy script

selenamarie:


scripts/deploy.sh
```diff
@@ -197,7 +197,7 @@ else
     error $? "Could not pushd /data/socorro"
     PYTHONPATH=. ../socorro-virtualenv/bin/python \
         ../socorro-virtualenv/bin/alembic \
-        -c config/alembic.ini upgrade head &> /var/log/socorro/alembic.log
+        -c config/alembic.ini upgrade heads &> /var/log/socorro/alembic.log
```
phrawzty: Would it be useful to see the log during the procedure?  If so, I would prefer `tee` to redirecting STDOUT, especially since this won't send STDERR to said log in the case of problems.  Possibly something like this instead?
```bash
-c config/alembic.ini upgrade heads 2>&1 | tee /var/log/socorro/alembic.log
```

selenamarie: In the past, the details have been shown in the log if there is a failure.. now i am unsure why tho? :D

selenamarie: Is that existing redirect perhaps only for stdout?

phrawzty: Yes, it's STDOUT only, since `&` assumes `&1` in Bash (hence `2>` to capture STDERR).

As for failures being captured in the log already, well, I guess that command doesn't output to STDERR at all, then. :wink:

It's.. not critical.  Just my grizzled old sysadmin brain clicking in. :grin:

phrawzty:
`r+` with nits.

PR #2607, opened 2015-01-29
fixes bug 1085530 - store uploaded symbols to S3

peterbe:
@rhelmer r? cc @twobraids 

So basically, we stop storing the uploaded file. Instead, we immediately unpack it and send it straight into S3. All within one big request.

### DIFF
--- a/docs/configuring-socorro.rst
+++ b/docs/configuring-socorro.rst
@@ -152,3 +152,58 @@ All of these jobs can be enabled or disabled in crontabber configuration or by m
 DEFAULT_JOBS in:
 
 https://github.com/mozilla/socorro/blob/master/socorro/cron/crontabber_app.py
+
+
+Symbols S3 uploads
+------------------
+
+The webapp has support for uploading symbols. This can be done by the user
+either using an upload form or you can HTTP POST directly in. E.g. with curl.
+
+For this to work you need to configure the S3 bucket details. The file
+``webapp-django/crashstats/settings/base.py`` specifies the defaults which
+are all pretty much empty.
+
+First of all, you need to configure the AWS credentials. This is done by
+overriding the following keys::
+
+    AWS_ACCESS_KEY
+    AWS_SECRET_ACCESS_KEY
+
+These settings can not be empty.
+
+Next you have to set up the bucket name. When doing so, if you haven't already
+created the bucket over on the AWS console or other management tools you
+also have to define the location. The bucket name is set by setting the
+following key::
+
+    SYMBOLS_BUCKET_DEFAULT_NAME
+
+And the location is set by setting the following key::
+
+    SYMBOLS_BUCKET_DEFAULT_LOCATION
+
+If you're wondering what the format of the location should be,
+you can see `a list of the constants here <http://boto.readthedocs.org/en/latest/ref/s3.html#boto.s3.connection.Location>`_.
+For example ``us-west-2``.
+
+If you want to have a different bucket name for different user you can
+populate the following setting as per this example:
+
+.. code-block:: python
+
+    SYMBOLS_BUCKET_EXCEPTIONS = {
+        'joe.bloggs@example.com': 'private-crashes.my-bucket',
+    }
+
+That means that when ``joe.bloggs@example.com`` uploads symbols they are
+stored in a different bucket called ``private-crashes.my-bucket``.
+
+If you additionally want to use a different location for this user you
+can enter it as a tuple like this:
+
+.. code-block:: python
+
+    SYMBOLS_BUCKET_EXCEPTIONS = {
+        'joe.bloggs@example.com': ('private-crashes.my-bucket', 'us-east-1'),
+    }
--- a/webapp-django/crashstats/symbols/utils.py
+++ b/webapp-django/crashstats/symbols/utils.py
@@ -5,35 +5,69 @@ import tarfile
 from cStringIO import StringIO
 
 
-def preview_archive_content(file_object, content_type):
-    """return file listing of the contents of an archive file"""
-    out = StringIO()
-    print >>out, "FILENAME".ljust(70),
-    print >>out, "SIZE".ljust(9)
-    print >>out, "-" * 80
+class _ZipMember(object):
+
+    def __init__(self, member, container):
+        self.name = member.filename
+        self.size = member.file_size
+        self.container = container
+
+    def extractor(self):
+        return self.container.open(self.name)
+
+
+class _TarMember(object):
+
+    def __init__(self, member, container):
+        self.member = member
+        self.name = member.name
+        self.size = member.size
+        self.container = container
+
+    def extractor(self):
+        return self.container.extractfile(self.member)
+
+
+def get_archive_members(file_object, content_type):
     if content_type == 'application/zip':
         zf = zipfile.ZipFile(file_object)
         for member in zf.infolist():
-            print >>out, member.filename.ljust(70),
-            print >>out, str(member.file_size).rjust(9)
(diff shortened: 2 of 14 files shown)


rhelmer:
This looks/works great, we have been chatting in IRC trying to work out what happens if an in-progress upload fails.

Right now any upload failure to S3 causes everything to fail, which means the client gets an error and should retry (good). However it means we also don't record anything in the database - I think we should write a row in the database for each file that's uploaded successfully to S3.

peterbe:
For the record, we discussed the risk of lack of atomicity. 
The django side is fully atomic. Meaning if the upload (half-way through) fails because the S3 network connection blows up nothing is recorded in our database and the client gets a 500 error. 
That would mean we'd potentially get some symbols uploaded to S3 that we don't have a record of. 

Some solutions:
1) We currently make a listing of the zip file as a blog of text and later proceed to upload each file. 
We could instead do something like this:
```python
# pseudo code
success = []
try:
    for file in zipfile:
        s3.upload(file)
        success.append(file)
    failed = False
except S3NetworkError:
    failed = True
SymbolsUpload.objects.create(
    user=request.user,
    files='\n'.join(success),
    failed=failed,
    size=zipfile.size,
    name=zipfile.name
)
```

2) We could delete what we've uploaded if it fails. But only delete what wasn't already there. 
This will almost always be tricky if we override a file in S3. 

Also, a risk I can see is that our local postgres connection might fail DURING the request. That means we could be uploading X% symbols but when it's upload 100% our DB connection to postgres might have broken and we can't write down what was upload. 
A solution to that would be to do something like this:
```python
upload = SymbolsUpload.objects.create(
    user=request.user,
    size=zipfile.size,
    name=zipfile.name,
    content=''
)
for file in zipfile:
    s3.bucket.upload(file)
    upload.content += '%s\n' % file
    upload.save()
```

webapp-django/crashstats/symbols/views.py
```diff
@@ -32,6 +36,46 @@ def check_symbols_archive_content(content, header_length=2):
...
+        bucket = conn.get_bucket(settings.SYMBOLS_BUCKET_NAME)
+    except boto.exception.S3ResponseError:
+        bucket = conn.create_bucket(
+            settings.SYMBOLS_BUCKET_NAME,
+            settings.SYMBOLS_BUCKET_LOCATION
+        )
+
+    total_uploaded = 0
+    for member in iterator:
+        key = bucket.new_key(os.path.join(
```
rhelmer: We expect to see a lot of attempts to upload symbols that are already present in S3. We should do a HEAD here and compare the size of the remote and local files, and not bother to upload if it's already present (```key.get_key()``` can do this, see http://boto.readthedocs.org/en/latest/ref/s3.html#boto.s3.bucket.Bucket.get_key)

rhelmer:
@peterbe I think the 3rd solution is the way to go - I don't think the large number of updates will be a big deal since they are all on the same connection and in the same transaction.

peterbe:
We discussed, here's what we will do:
1. For each file we upload/or not we write a link (full S3 URL that contains the bucket name, key, etc) in our big text field 
2. Just before uploading, we check if the key already exists and if it has the same size. E.g. (`boto.get_key()`) If the file does EXISTS we don't bother uploading again. But if the size is different, we upload again.
3. Each file in the zip file will go into the text field but we'll prefix each line with a `=` if the file wasn't re-uploading and `+` if it was uploaded.

peterbe:
@rhelmer time for another review. 
So, I implemented what was mentioned in the comment above **except** instead of storing the absolute URL in the text field I decided on my own format:
```
# when the file didn't exist or the size was different
+BUCKET_NAME,PREFIX/FILENAME
# when the upload was cancelled because it wasn't sufficiently different
=BUCKET_NAME,PREFIX/FILENAME
```

For example:
screenshot 2015-02-05 10 48 09

The reason I didn't use a URL was because then you have to decide on an expiration time and the URL will contain a key. 
Also, suppose the filename contains something that needs to be URL encoded, if you ever want to track an upload you'd have to reverse the URL into a key name (aka S3 filename/filepath). 

r?

rhelmer:
One thing we missed - some partners upload symbols that are private and these should go to a separate bucket. For simplicity, we can have this tied to the account (e.g. "partnerA always uploads to private symbol bucket"), or it could be something configurable by the uploader as an optional param in the URL perhaps.

peterbe:
So, how do we decide this? 
Will a certain partner sometimes upload public and sometimes upload private symbols? If so she'll need an option to override in her uploads. 
Otherwise it might be easier to attach a property to the user that uploads always go into a certain bucket.

rhelmer:
@peterbe currently our only use case is that one partner is always private and the rest are always public, so let's just have the ability to tie an account to a bucket if that's simpler.

peterbe:
Should we hardcode that user email to a specific bucket name? in `settings/base.py`.
Or do we need this to be "bigger". I.e. a new UserProfile module with an admin functionality to edit these etc.

rhelmer:
@peterbe I am OK with hardcoding in config for now, maybe a dict to map users to buckets, with a default? I think in the grand scheme of things we don't want to be hosting private symbols for folks unless we really really need to.

peterbe:
@rhelmer e.g. https://gist.github.com/peterbe/dd685444587068081f42
This means this email will never be able to upload public symbols. 
cc @luser

luser:
That looks fine. I think that's exactly what we want--the ability to set a user as "all uploads are private".

peterbe:
Rebased up so hopefully mergeable again.

webapp-django/crashstats/symbols/utils.py
```diff
@@ -5,36 +5,70 @@
...
 
     elif content_type == 'application/x-tar':
         zf = tarfile.TarFile(fileobj=file_object)
         for member in zf.getmembers():
-            if member.isfile():
-                print >>out, member.name.ljust(70),
-                print >>out, str(member.size).rjust(9)
+            # Sometimes when you make a tar file you get a
+            # smaller index file copy that start with "./._".
+            if member.isfile() and not member.name.startswith('./._'):
```
rhelmer: Hm this is new to me.. looking it up I guess it's an OS X thing, "AppleDouble"?

docs/configuring-socorro.rst
```diff
@@ -152,3 +152,58 @@ All of these jobs can be enabled or disabled in crontabber configuration or by m
 DEFAULT_JOBS in:
 
 https://github.com/mozilla/socorro/blob/master/socorro/cron/crontabber_app.py
+
+
+Symbols S3 uploads
+------------------
```
rhelmer: Nice, thanks for documenting this :+1: "how do I upload symbols" is a super common question

webapp-django/crashstats/symbols/views.py
```diff
@@ -48,6 +49,80 @@ def check_symbols_archive_content(content, header_length=2):
...
+            settings.SYMBOLS_FILE_PREFIX, member.name
+        )
+        key = bucket.get_key(key_name)
+
+        # let's assume first that we need to add a new key
+        prefix = "+"
+        if key:
+            # key already exists, but is it the same size?
+            if key.size != member.size:
+                # we're changing our minds, so it get uploaded again
```
rhelmer: this "changing our minds" comment is confusing.. maybe just ```# file size in S3 is different, upload the new one```?
If you disagree at least ```s/get/gets/``` :)

rhelmer:
r+ with a silly nit about a comment

peterbe:
rebased and squashed. awaiting travis.

PR #2608, opened 2015-01-30
No bug - Added instructions for Elasticsearch to the configure page.

AdrianGaudebert:
@rhelmer r?

### DIFF
--- a/docs/configuring-socorro.rst
+++ b/docs/configuring-socorro.rst
@@ -8,6 +8,19 @@ Socorro produces graphs and reports, most are updated once per day.
 You must enter information about your releases into Socorro in order
 for this to work, and this information must match the incoming crashes.
 
+Configure Elasticsearch
+-----------------------
+
+Some key features of Socorro (like Super Search) depend on Elasticsearch.
+This service needs to be initialized before it can be used:
+
+.. code-block:: bash
+
+    cd scripts && python ./setup_supersearch_app.py
+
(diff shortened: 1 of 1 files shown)


docs/configuring-socorro.rst
```diff
@@ -8,6 +8,19 @@ Socorro produces graphs and reports, most are updated once per day.
 You must enter information about your releases into Socorro in order
 for this to work, and this information must match the incoming crashes.
 
+Configure Elasticsearch
+-----------------------
+
+Many features of Socorro depend on Elasticsearch, and to work you will first
```
phrawzty: This is true but vague - what data, and why?  I would recommend adding some detail as to what this script actually does.

AdrianGaudebert: Mmm this is all explained if you follow the link. I don't really want to repeat things too much. Do you have a suggestion of a good catch phrase to sum it up?

phrawzty: That's a valid point.  I guess it was the name of the script combined with the wording of both blocks that wasn't clear to me.  Perhaps something like:
> Supersearch, which is one of the most useful features of Socorro, depends on Elasticsearch.  The service needs to be initialised before it can be used:
[code-block]
For more comprehensive [etc...]

AdrianGaudebert: I've updated it. Is it better now?

phrawzty:
`r+`

PR #2609, opened 2015-01-30
Fixes bug 1127783 - By default, load Elasticsearch mapping data from dat...

AdrianGaudebert:
...abase instead of a file.

@phrawzty r?

### DIFF
--- a/socorro/external/elasticsearch/crashstorage.py
+++ b/socorro/external/elasticsearch/crashstorage.py
@@ -68,7 +68,7 @@ class ElasticSearchCrashStorage(CrashStorageBase):
     )
     required_config.add_option(
         'use_mapping_file',
-        default=True,
+        default=False,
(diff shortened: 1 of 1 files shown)


AdrianGaudebert:
Hah, only automatic-emails tests are failing. Let's wait until #2610 is merged and the problem will be solved... :-)

phrawzty:
Why do this :question:

AdrianGaudebert:
This configuration option is used to determine if the Elasticsearch mapping for new indices is loaded from a file (in our repository) or from the database (using the Super Search Fields list). It was set to ``True`` by default because that's how it has been working historically. Then we turned it to ``False`` in staging and production. Now that we know loading that data from our database works well, we should make it ``False`` by default, so our external users benefit it.

phrawzty:
Thanks for the clarification!

`r+`

PR #2612, opened 2015-02-02
Fixes Bug 907499 - added checksum to collector

twobraids:
I've benchmarked for two days and can detect no impact on the collectors whats so ever.  Perhaps we're being overly cautious in not wanting to put the md5sum code into the collector.  This is clearly the most logica place to put it.

### DIFF
--- a/socorro/unittest/collector/test_wsgi_breakpad_collector.py
+++ b/socorro/unittest/collector/test_wsgi_breakpad_collector.py
@@ -2,6 +2,8 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import hashlib
+
 import mock
 from nose.tools import eq_, ok_
 from datetime import datetime
@@ -33,6 +35,7 @@ class TestCollectorApp(TestCase):
         config.collector.dump_field = 'dump'
         config.collector.accept_submitted_crash_id = False
         config.collector.accept_submitted_legacy_processing = False
+        config.collector.checksum_method = hashlib.md5
 
         config.crash_storage = mock.MagicMock()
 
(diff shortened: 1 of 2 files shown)


rhelmer:
I don't think there's any harm in making the collector do a checksum of the already in-memory minidump and put that into S3. The part I was more worried about is whether collector would be checking for duplicates and use that to decide whether or not to queue the crash for processing.

I guess we could use S3 for this - just check for the presence of a key with the md5sum for the name - it wouldn't be perfect due to eventual consistency, but it'd probably catch a fair number of them.

If we deferred the duplicate checking to processors they could do this in Postgres instead, which would be totally consistent, but I don't think we'd want to hook collectors up to Postgres at all.

twobraids:
that is the same conclusion that I have drawn.  calculate in the collector - do something with the data elsewhere.

rhelmer:
Another option for consistency is DynamoDB which has (non-default) option for strongly consistent reads.

selenamarie:
@twobraids @rhelmer i concur!

The last bit of IRC conversation led me to: 
```
collector (md5) --> KV store (memcache? redis? aerospikeDB?) <-- processors
     |                                                                  |
     |                                                                  |
     --------------> raw_crash incl md5  <------------------------------
```

selenamarie:
Can we just do this? :checkered_flag: 

I'm super excited to end that reports_duplicates job :)

socorro/unittest/collector/test_wsgi_breakpad_collector.py
```diff
@@ -64,7 +67,11 @@ def test_make_raw_crash(self):
         eq_(rc.some_field, '23')
         eq_(rc.some_other_field, 'XYZ')
 
-    def test_POST(self):
+    @mock.patch('socorro.collector.wsgi_breakpad_collector.time')
```
selenamarie: I like this change to use decorators!  Much more readable.

selenamarie:
r+ i dig the example of using a standard lib with `class_converter` in this patch!  And thanks again for refactoring those tests.

PR #2614, opened 2015-02-03
fix bug 981079 and 1127924 - support stackwalker HTTP symbol fetch and caching

rhelmer:
feedback? @peterbe @luser - this is what we have so far to handle cleanup in the processor.

I have the following qualms about it:
1. pyinotify module is only available for Linux, other platforms won't be able to ```pip/peep install -r requirements.txt``` anymore if we do this as-is
2. pyinotify has no tests of its own (it's basically just a thin wrapper around Linux inotify, maybe doesn't matter?)
3. lars and I aren't sure how to adequately test it (maybe doesn't matter per point 2)

Since the class itself can be optional, if we can figure out a workaround for point 1 then I think we can take this as-is. Any thoughts?


rhelmer:
@twobraids and I discussed possibly abandoning pyinotify and writing symbols to a radix-style directory structure (which we could cleanup in a manner identical to how we do filesystem crash storage) but I am not sure we have time to pull this off (stackwalker isn't expecting this for one thing) , we could write a "companion" class for processor to swap in later that does this if we want to.

luser:
We can certainly modify the stackwalker's symbol lookup code any way we desire, but note that things are the way they are currently because the stackwalker only has the debug file and debug ID to use for symbol lookup, so it does a simple path lookup:
https://code.google.com/p/google-breakpad/source/browse/trunk/src/processor/simple_symbol_supplier.cc#153

peterbe:
I'm quite stumped. That's a massive PR to consider. Especially for someone who is so cold on processor code. 

From experience I have doubts about inotify (and pyinotify) in that it's kinda unreliable. 
I started using it many years ago and maybe things have progressed since those kernels. The problem was that it doesn't work 0.1% of the time. The trouble is usually not when things are left alone but when you do things like killing the process and starting another (new release) or if you have stale events that stack up. 

Colleagues at my last gig had similar problems when inotify was used to spread the disk writes onto a raid of disks as new files came in from a mobile photo app. It just fails strangely some rare few times but when it does you really really scratch your head. 

I think, ideally, one should look at inotify as a luxury. The foundation could be a periodic job (e.g. cron or a daemon while loop with sleep) that gets sped up by a great-if-it-works-no-biggie-if-it-doesn't-inotify-job.

luser:
The key feature inotify gives us here is the ability to maintain the cache in LRU order, so that when we run out of space we can remove the least-recently-used symbol files. Linux filesystems don't maintain access time by default (since it slows down I/O) so otherwise you have no way to know which files have been used most recently without maintaining some extra state somewhere. The best a cron job could do would be to remove some arbitrary set of symbol files to free up space.

rhelmer:
I forked pyinotify in order to move the platform check to import-time instead of install-time. Also opened a PR against upstream: https://github.com/seb-m/pyinotify/pull/92

socorro/processor/breakpad_transform_rules.py
```diff
@@ -459,3 +490,153 @@ def _predicate(
...
+    required_config.add_option(
+        'symbol_cache_path',
+        doc='the path where the symbol cache is found (quote path with '
+        'embedded spaces)',
+        default='/mnt/socorro/symbols',
+    )
+    required_config.add_option(
+        'temporary_file_system_storage_path',
+        doc='a path where temporary files may be written',
+        default='/tmp',
```
rhelmer: @peterbe can I use ```tempfile.gettempdir()``` here? How do I override it at runtime with configman?

twobraids: @rhelmer: running tempfile.gettempdir() will just set the temporary
directory location to a constant at import time.  That constant can be
overridden in the same manner as any other constant that might be put
there.

```gettempdir``` doesn't create a new temporary directory, it just search a
set of known typical temp directory locations.  It'll return the first one
that it finds for which the current user has write permission.

Comment (peterbe) [inline socorro/processor/breakpad_transform_rules.py]: It's no biggie. If someone's system's temp dir is /var/cache/tmp instead, for example, when they run some `--admin.print_conf` command it'll become that instead of `'/tmp'` or the python `tempfile.gettempdir()`

**Comment by rhelmer** [comment]
@peterbe the reason the diff is so big is that it renames all the old "monitor" stuff we no longer use, and makes a "companion" process that can run in a thread alongside the processor. Monitor has not been in use for a long time, and this feature will be off by default (since it breaks non-Linux and not everyone will necessarily use a symbol cache)

**Comment by rhelmer** [comment]
r? @luser @peterbe - I've rebased and made suggested changes, as we discussed in the meeting yesterday I think we should move forward with using pyinotify (using my fork, I've opened a PR upstream we can switch to later)

If we find inotify problematic we can swap out the ```SymbolLRUCacheManager``` for a different ```companion_class``` later, but we'd likely need to adjust stackwalker too if we use a different directory structure (e.g. radix) for cleanup purposes.

File: requirements.txt [hunk 2]
```diff
@@ -124,6 +124,8 @@ pyflakes==0.8.1
 pyhs2==0.6.0
 # sha256: TzJCJuHuhaxBUgV5up3nAylAEmAuU2IX_Q50VuFH2IY
 sasl==0.1.3
+# sha256: 6nj05rvk7z_4OHM6mUIsl7GjE2plb4N3PqnJWaSRRlw
+https://github.com/rhelmer/pyinotify/archive/9ff352f.zip#egg=pyinotify
```
peterbe: Please add a comment above the sha that points to https://github.com/seb-m/pyinotify/pull/92 and that says that we can start using `pyinotify>0.9.5` once that PR has been merged and a new release made.

socorro/processor/breakpad_transform_rules.py
```diff
@@ -353,6 +354,11 @@ def version(self):
 
     #--------------------------------------------------------------------------
     def _execute_external_process(self, command_line, processor_meta):
+        if self.config.get('chatty', False):
```
peterbe: "chatty" :)
Normally the nomenclature is "verbose" or "verbosity".

socorro/processor/breakpad_transform_rules.py
```diff
@@ -374,6 +380,22 @@ def _execute_external_process(self, command_line, processor_meta):
         return external_command_output, return_code
 
     #--------------------------------------------------------------------------
+    @staticmethod
+    def dot_save(a_mapping, key, value):
```
peterbe: I feel like I've seen this solution before. Like somewhere in configman around dotdict. I could be wrong but it seems strange to need a whole new staticmethod for something that is only used twice.

socorro/processor/breakpad_transform_rules.py
```diff
@@ -459,3 +491,153 @@ def _predicate(
...
 
 
+#==============================================================================
+class BreakpadStackwalkerRule2015(ExternalProcessRule):
+
+    required_config = Namespace()
+    required_config.add_option(
+        name='symbols_url',
+        doc='url of the symbol server',
+        default="https://localhost"
```
peterbe: Doesn't this need one of those [likely_to_be_changed=True](https://github.com/mozilla/configman/blob/master/configman/option.py#L33) things.

socorro/processor/breakpad_transform_rules.py
```diff
@@ -459,3 +491,153 @@ def _predicate(
...
+    required_config.command_pathname = change_default(
+        ExternalProcessRule,
+        'command_pathname',
+        '/data/socorro/stackwalk/bin/stackwalker',
+    )
+    required_config.add_option(
+        'symbol_cache_path',
+        doc='the path where the symbol cache is found, this location must be '
+        'readable and writeable (quote path with embedded spaces)',
+        default='%s/symbols' % tempfile.gettempdir(),
```
peterbe: change to 
```python
default=os.path.join(tempfile.gettempdir(), 'symbols'),
```
to be safe.

socorro/processor/breakpad_transform_rules.py
```diff
@@ -459,3 +491,153 @@ def _predicate(
...
+    required_config.add_option(
+        'temporary_file_system_storage_path',
+        doc='a path where temporary files may be written',
+        default=tempfile.gettempdir(),
+    )
+
+
+    #--------------------------------------------------------------------------
+    #def __init__(self, config):
+        #super(BreakpadStackwalkerRule, self).__init__(config)
```
peterbe: rm?

socorro/processor/breakpad_transform_rules.py
```diff
@@ -459,3 +491,153 @@ def _predicate(
...
+    def _temp_raw_crash_json_file(self, raw_crash, crash_id):
+        file_pathname = os.path.join(
+            self.config.temporary_file_system_storage_path,
+            "%s.%s.TEMPORARY.json" % (
+                crash_id,
+                threading.currentThread().getName()
+            )
+        )
+        with open(file_pathname, "w") as f:
+            ujson.dump(dict(raw_crash), f)
```
peterbe: There's no need to create a whole new dict from the `raw_crash`. E.g.:
```python
>>> from configman.dotdict import DotDict
>>> thing = DotDict()
>>> thing.foo = "bar"
>>> import ujson
>>> ujson.dumps(thing)
'{"foo":"bar"}'
```

peterbe: the `dict(some_mapping)` will create a duplicate in memory and considering that these raw crashes are sometimes very large that can mean a lot of memory.

socorro/processor/breakpad_transform_rules.py
```diff
@@ -459,3 +491,153 @@ def _predicate(
...
+
+    #--------------------------------------------------------------------------
+    def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
+        if 'additional_minidumps' not in processed_crash:
+            processed_crash.additional_minidumps = []
+        with self._temp_raw_crash_json_file(
+            raw_crash,
+            raw_crash.uuid
+        ) as raw_crash_pathname:
+            for dump_name, dump_pathname in raw_dumps.iteritems():
```
peterbe: We might not need the `dump_pathname` so no need to use `iteritems()`. Instead we should use `iter(raw_dumps)` or `raw_dumps.iterkeys()`.

socorro/processor/breakpad_transform_rules.py
```diff
@@ -459,3 +491,153 @@ def _predicate(
...
+                # this rule is only interested in dumps targeted for the
+                # minidump stackwalker external program.  As of the writing
+                # of this code, there is one other dump type.  The only way
+                # to differentiate these dump types is by the name of the
+                # dump.  All minidumps targeted for the stackwalker will have
+                # a name with a prefix specified in configuration:
+                if not dump_name.startswith(self.config.dump_field):
+                    # dumps not intended for the stackwalker are ignored
+                    continue
+
```
peterbe: At this point we will need the dump_pathname. So if we start looping over keys instead, here we can do:
```pyhon
dump_pathname = raw_dumps[dump_name]
```

socorro/processor/breakpad_transform_rules.py
```diff
@@ -459,3 +491,153 @@ def _predicate(
...
+                if self.config.chatty:
+                    self.config.logger.debug(
+                        "BreakpadStackwalkerRule: %s, %s",
+                        dump_name,
+                        dump_pathname
+                    )
+
+
+                command_parameters = dict(self.config)
```
peterbe: Is it deliberate to make a copy here? A much neater syntax to achieve this is to make it on the fly. 
```python
command_line = self.config.command_line.format(
    **dict(
        self.config, 
        dump_file_pathname=dump_pathname,
        raw_crash_pathname=raw_crash_pathname
    )
)
```

socorro/processor/symbol_cache_manager.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+#==============================================================================
+class SymbolLRUCacheManager(RequiredConfig):
+    """for cleaning up the symbols cache"""
+    required_config = Namespace()
+    required_config.add_option(
+        'symbol_cache_path',
+        doc="the cache directory to automatically remove files from",
+        default='%s/symbols' % tempfile.gettempdir(),
```
peterbe: This too should ideally be
```python
default=os.path.join(tempfile.gettempdir(), 'symbols'),
```

socorro/unittest/processor/test_breakpad_transform_rules.py
```diff
@@ -580,3 +606,129 @@ def test_default_parameters(self):
...
+        )
+        mocked_subprocess_handle.stdout.read.return_value = (
+            cannonical_stackwalker_output_str
+        )
+        mocked_subprocess_handle.wait.return_value = 0
+
+        rule = BreakpadStackwalkerRule2015(config)
+
+        # the call to be tested
+        print "about to act"
```
peterbe: debugging

peterbe:
I found a bunch of low-hanging fruit stuff. The most alarming is the excessive memory footprint of doing `dict(raw_crash)` when it's not necessary. (and if it is necessary, there should be a bold and clear comment explaining that).

rhelmer:
OK addressed review comments, thanks @peterbe 

Please keep in mind when reading this comment that the default configs for the processor are a bit out of date/unfriendly for local development, which is why some of these config options are a little scarily long :) However these should give us an idea of what we'll need to do to configure this on stage/prod in both PHX and AWS.

I am testing this locally, processor runs fine on Mac unless you enable this feature (as expected):
```
$ socorro processor \
  --companion_process.companion_class='socorro.processor.symbol_cache_manager.SymbolLRUCacheManager'
inotify is not available on darwin
```

Works as expected on Linux:
```
socorro processor \
  --processor.processor_class=socorro.processor.mozilla_processor_2015.MozillaProcessorAlgorithm201 \
  --companion_process.companion_class='socorro.processor.symbol_cache_manager.SymbolLRUCacheManager' \
  --new_crash_source.new_crash_source_class='socorro.external.fs.fs_new_crash_source.FSNewCrashSource' \
  --source.crashstorage_class='socorro.external.fs.crashstorage.FSLegacyDatedRadixTreeStorage' \
  --destination.crashstorage_class='socorro.external.fs.crashstorage.FSLegacyDatedRadixTreeStorage'  \
  --processor.raw_to_processed_transform.BreakpadStackwalkerRule2015.command_pathname=./minidump-stackwalk/stackwalker \
  --processor.raw_to_processed_transform.BreakpadStackwalkerRule2015.symbols_url='https://s3-us-west-2.amazonaws.com/org.mozilla.crash-stats.symbols-public/v1/' \
  --companion_process.symbol_cache_size=1G
```

I can send in crashes, symbols get downloaded and cached, and it seems to clean up if we hit the limit (we'll be able to stress-test this better on stage)

However I just noticed that we're missing a feature - we need to be able to specify two ```symbols_url```s to stackwalker, one for public symbols and one for private. I'll fix that up.

socorro/processor/breakpad_transform_rules.py
rhelmer: @peterbe I agree however there's already precedent for it in other files in this directory so I am inclined to leave it.

socorro/processor/breakpad_transform_rules.py
rhelmer: @peterbe I looked around and didn't see an already-existing utility method... I am inclined to leave this alone unless you have a more specific suggestion :)

socorro/processor/breakpad_transform_rules.py
rhelmer: @peterbe I did address this, but I guess it's just below the diff above so github doesn't consider this comment "outdated" :(

rhelmer:
@luser seriously thought about having the two different ```symbol_cache_path``` options - I am ambivalent, since this feature removes files from the directory specified I think it's safer to have to be explicit.

luser:
@rhelmer I wouldn't block you on that, it just seems odd that you'd have to specify it in two places since you always want it to be the same thing. Feels like one more place for users to screw up their configuration.

rhelmer:
@luser now that I have looked into it I can't see a straightforward way to get at the ```symbol_cache_dir``` as it's not set by processor but by transform rules, which are loaded at runtime (not sure how to find it without hardcoding the name of the transform rule class). I'll file a followup and we can figure it out.

I spoke with Orphis and he would like to see his code integrated into processor, it's at https://gist.github.com/rhelmer/ae1bea9b9f06bfec3dc7 - I'm not sure we need the time-limited or uploading features, just doing an HTTP proxy with boto is pretty simple.

I'll open followup bugs for these two issues.

rhelmer:
Filed bug 1130426 and bug 1130428.

rhelmer:
@peterbe @luser - I've rebased and filed followup bugs for the outstanding issues ted pointed out, are we good to merge this? :)

luser:
LGTM

socorro/processor/breakpad_transform_rules.py
peterbe: Let's leave it as is then. @twobraids can maybe refactor it some day in the future.

socorro/processor/breakpad_transform_rules.py
peterbe: I did put a smiley face on it after all.

peterbe:
r+ as it seems you've addressed my superficial nits :)

I don't know what to think about that sample commandline line. It makes my eyes hurt but I can't really offer anything better :)

twobraids:
that command line could be expressed like this:  

```bash
export processor__processor_class=socorro.processor.mozilla_processor_2015.MozillaProcessorAlgorithm2015 
export companion_process__companion_class=socorro.processor.symbol_cache_manager.SymbolLRUCacheManager
export new_crash_source__new_crash_source_class=socorro.external.fs.fs_new_crash_source.FSNewCrashSource
export source__crashstorage_class=socorro.external.fs.crashstorage.FSLegacyDatedRadixTreeStorage
export destination__crashstorage_class=socorro.external.fs.crashstorage.FSLegacyDatedRadixTreeStorage
export processor__raw_to_processed_transform__BreakpadStackwalkerRule2015__command_pathname=./minidump-stackwalk/stackwalker
export processor__raw_to_processed_transform__BreakpadStackwalkerRule2015__symbols_url="https://s3-us-west-2.amazonaws.com/org.mozilla.crash-stats.symbols-public/v1/"
export companion_process__symbol_cache_size=1G

socorro processor
```

rhelmer:
@twobraids yes good point maybe more readable that way than as command line args :) I think really what makes it overwhelming is 1) need to override so many default options 2) values are so long, I'll file a followup bug on the former, I don't think there's an easy answer for the latter

rhelmer:
Thanks everybody!

rhelmer:
@twobraids filed that followup about more friendly defaults in bug 1130598

PR #2615, opened 2015-02-04
Fixes bug 1129476 add SQL files to setup.py install for alembic

selenamarie:
Adds `data_file` paths for our SQL files so that when the package is installed, `alembic` is able to find the raw SQL files in the `virtualenv`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,5 @@
 import codecs
+import glob
 import os
 from setuptools import setup, find_packages
 
@@ -46,4 +47,12 @@ setup(
         },
     test_suite='nose.collector',
     zip_safe=False,
+    data_files=[
+        ('socorro/external/postgresql/raw_sql/procs',
+            glob.glob('socorro/external/postgresql/raw_sql/procs/*.sql')),
+        ('socorro/external/postgresql/raw_sql/views',
+            glob.glob('socorro/external/postgresql/raw_sql/views/*.sql')),
+        ('socorro/external/postgresql/raw_sql/types',
+            glob.glob('socorro/external/postgresql/raw_sql/types/*.sql')),
+    ],
 ),
--- a/socorro/external/postgresql/postgresqlalchemymanager.py
+++ b/socorro/external/postgresql/postgresqlalchemymanager.py
@@ -62,7 +62,9 @@ class PostgreSQLAlchemyManager(object):
     def create_types(self):
         types_dir = os.path.normpath(os.path.join(
             __file__,
-            'raw_sql/types'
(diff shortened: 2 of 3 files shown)


selenamarie:
@rhelmer r?

rhelmer:
@selenamarie hmm if I ```python setup.py install``` into a virtualenv and then ```cd /tmp```, it doesn't seem to work:
```
sqlalchemy.exc.ProgrammingError: (ProgrammingError) type "major_version" does not exist
```
socorro/external/postgresql/postgresqlalchemymanager.py seems to expect the files to be in ```socorro/external/postgresql/raw_sql/views/*.sql```, does it need be looking on the PYTHONPATH for these instead (is it able to do that?) Or do we need to build a path for it?

selenamarie:
@rhelmer oh yeah i need to fix the paths.. i did it already for alembic, so will just patch it up now.

setup.py
```diff
@@ -46,4 +47,12 @@ def read(fname):
...
     test_suite='nose.collector',
     zip_safe=False,
+    data_files=[
+        ('socorro/external/postgresql/raw_sql/procs',
+            glob.glob('socorro/external/postgresql/raw_sql/procs/*.sql')),
+        ('socorro/external/postgresql/raw_sql/views',
+            glob.glob('socorro/external/postgresql/raw_sql/views/*.sql')),
+        ('socorro/external/postgresql/raw_sql/types',
+            glob.glob('socorro/external/postgresql/raw_sql/types/*.sql')),
+    ],
```
peterbe: For the record, all of this could be replaced with adding 
```
include socorro/external/postgresl/raw_sql/*/*.sql
```
in the `MANIFEST.in` file. 

The question is, what's the difference? Is one better than the other?

selenamarie: This means we have to run `python setup.py sdist` in addition to `python setup.py install`. 

I brought it up in channel earlier this morning and we figured that adding the extra step for this one thing was not helpful.

rhelmer:
r+

I tested this locally, and the ```config/alembic.ini``` had to be present (expected), as well as the ```./alembic``` directory (which is where the .ini file was pointing to - I wasn't expecting this but I  guess it makes sense? Is this supposed to be in the package too? I don't know if alembic would know how to find it)

```socorro setupdb``` seems to find the ```.sql``` files just fine.

selenamarie:
I'll start thinking about how we should handle alembic in a more automated way. We have it documented to use the config file with `-c config/alembic.ini`, but it could be done better.  Thanks for the review.

PR #2616, opened 2015-02-04
Fixes bug 1129508 add info for new product config

selenamarie:
Does some cleanup on the database development docs, remove some deprecated documentation for things that no longer exist and makes a rough stab at helping people who are trying to setup a new install for production in a non-Mozilla environment.

Thanks to never in IRC for filing the bug and providing the source inspiration and steps for getting going.


selenamarie:
r+ from never on irc

PR #2618, opened 2015-02-05
Fixes bug 1129649 - add elasticsearch fixture to package

selenamarie:
Help resolve an error when installing socorro from scratch.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,16 +35,16 @@ setup(
         'Programming Language :: Python :: 2.6',
         'Programming Language :: Python :: 2.7',
         'Topic :: Internet :: WWW/HTTP :: WSGI :: Application',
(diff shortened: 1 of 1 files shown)


selenamarie:
@AdrianGaudebert r?

AdrianGaudebert:
Ah, I get it! This should not be needed anymore as of https://github.com/mozilla/socorro/commit/afc2681d1f1424013bfebdef0300451fe6c11a44. We now, by default, don't need that file at all. Since this json file is going to be removed entirely (actually, we could even do that right now), I'm in favor of not adding it to the package. 

Could you please confirm that you don't get the error with an updated master?

selenamarie:
great! thanks.

PR #2619, opened 2015-02-05
fixes bug 1130048 - Tests should pass in python 2.6 and 2.7

peterbe:
To do:
- [x] Make all tests work in 2.6 and 2.7
- [x] Remove 2.7 from the `.travis.yml` file


peterbe:
@selenamarie @rhelmer (cc @twobraids but he's not available I know)

If I can get an r+ on this, then I'll edit it one last time so the `.travis.yml` doesn't have to do both 2.6 and 2.7.

selenamarie:
r+  thanks @peterbe!

peterbe:
I did a rebase. Let's merge when it goes green.

PR #2621, opened 2015-02-09
fixes bug 1102244 - Hover details on reports/list graphs off screen

peterbe:
@rhelmer r?


peterbe:
The reason it didn't work before was because `d3.event.pageX` (which was already an int) was sent through `parseInt()` with a base of 16 (instead of 10). 
https://github.com/mozilla/socorro/blob/6a0f900b7bcd361020a204c55c2b347c5dee5476/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/report_list_graph.js#L245

lonnen:
r+ :goberserk:

PR #2623, opened 2015-02-10
No bug - Cleanup breakpad.tar.gz before downloading it.

AdrianGaudebert:
@rhelmer this is constantly causing errors for me, forcing me to ``rm -f breakpad.tar.gz`` manually every time I want to run ``make test``. Adding it to the script would make me less frustrated. :-)

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -27,6 +27,7 @@ fi
 
 if [ "`uname -sm`" == "Linux x86_64" ]; then
   # pull pre-built, known version of breakpad
+  rm -f breakpad.tar.gz
   wget -N --quiet 'https://ci.mozilla.org/job/breakpad/lastSuccessfulBuild/artifact/breakpad.tar.gz'
   tar -zxf breakpad.tar.gz
   rm -rf stackwalk


phrawzty:
What is the error that you're trying to solve here?  Without any context it looks like you're having problems with `wget`?  Please add more details.

phrawzty:
```
03:16:57 < adrian> phrawzty: some obscure error caused by breakpad.tar.gz, I'm
                   re-running to get the precise log
03:17:22 < adrian> crap it passed this time
03:17:41 < adrian> I think it happens when there's a new version of breakpad
03:18:00 < adrian> removing that file is the only way around it I found
03:20:13 < phrawzty> adrian: the -N tells wget to enable timestamping, so if the
                     build artefact has the same stamp, it won't re-download it.  If
                     the artefact has a newer stamp, it will downloadn and over-write
                     that which exists locally.
03:20:42 < phrawzty> adrian: rm'ing the file might be a solution, but it also might
                     be covering up a larger problem ?
03:21:44 < adrian> phrawzty: I can close the PR for now and file a bug with a more
                   precise log next time it happens
03:23:29 < phrawzty> adrian: it's your call, but it's worth looking into, imho.
```

AdrianGaudebert:
Will reopen or file a bug with more info.

PR #2624, opened 2015-02-10
Fixes bug 850610 - Filled the uuid field in extensions table.

AdrianGaudebert:
@selenamarie r?


AdrianGaudebert:
Wait, this service doesn't seem to be used by the UI anymore. There is no model calling this endpoint. Can we remove extensions?

selenamarie:
@AdrianGaudebert good question! I'd pose it in the bug.. it doesn't seem to be used in any reports inside the database either.

AdrianGaudebert:
RESOLVED - WILL REMOVE
https://bugzilla.mozilla.org/show_bug.cgi?id=1132068

PR #2626, opened 2015-02-10
fixes bug 1131319 - "Show/hide other threads" setting remembered

peterbe:
@AdrianGaudebert r?

### DIFF
--- a/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/report.js
+++ b/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/report.js
@@ -29,10 +29,23 @@ $(document).ready(function () {
         });
     });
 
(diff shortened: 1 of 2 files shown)


AdrianGaudebert:
This feels like a hack. Relying on the hash to keep the state of a page seems wrong to me, though I understand why you went this way. I just wonder if it wouldn't be better to use a more persistent way, like local storage? 

The code looks good though, so if you strongly disagree with me, feel free to merge. You have my r+. :)

peterbe:
Valid point. But remember that `#allthreads` is an anchor link which means that simply going to it will scroll the chrome into view. 
Also, I managed to achieve the functionality without persist-ency. That must mean localStorage would be overkill.

PR #2628, opened 2015-02-13
DO NOT MERGE Bug1132937 add heroku url

selenamarie:
Testing some changes for working in heroku.


socorro/external/postgresql/postgresqlalchemymanager.py
```diff
@@ -52,9 +63,10 @@ def setup_admin(self):
         if not self.min_ver_check(90300):
             self.session.execute(
                 'CREATE EXTENSION IF NOT EXISTS json_enhancements')
-        self.session.execute('CREATE SCHEMA bixie')
-        self.session.execute(
-            'GRANT ALL ON SCHEMA bixie, public TO breakpad_rw')
+        self.session.execute('CREATE SCHEMA IF NOT EXISTS bixie')
```
rhelmer: didn't we remove the bixie stuff elsewhere? Surprised it's still in here could probably drop it at this point.

lonnen: Plz. `Bixie => :fire: `

socorro/unittest/config/commonconfig.py.dist
```diff
@@ -25,7 +25,7 @@ oldDatabaseName.default = 'socorro_test'
 
 databaseUserName = cm.Option()
 databaseUserName.doc = 'the user name for the database servers'
-databaseUserName.default = 'breakpad_rw'
+databaseUserName.default = 'test'
```
selenamarie: This is legacy stuff, but until it is fully deprecated, I think we should just use 'test' as the user, since this is what we are using for the rest of our unit test configs.

socorro/unittest/config/jenkins.py.dist
```diff
@@ -25,7 +25,7 @@ oldDatabaseName.default = 'socorro_test'
 
 databaseUserName = cm.Option()
 databaseUserName.doc = 'the user name for the database servers'
-databaseUserName.default = 'breakpad_rw'
+databaseUserName.default = 'test'
```
selenamarie: Same as for commonconfig -- this should get deprecated eventually in favor of configman style config. Until then, I recommend using the 'test' user as that's what we use in our unit test configs.

selenamarie:
This patch is unfortunately epic. 

So -- I could carve out four separate patches if that's wanted: 

* Bixie deprecation which got pulled in because of an incidental comment from @rhelmer about some cruft around schema management left in a function
* Cleanup of functions that have weird ownership and aren't part of our core distribution of Socorro (like monitoring tables and some stuff with `postgres` ownership
* Updating our `databaseUserName.default` throughout the tests to use `test` instead of `breakpad_rw`
* Updating setupdb_app.py and related libraries to support the heroku-style URLs and heroku-style databases where we have only one `ROLE` available and no ability to change ownership or update ownership on database objects.

The bixie deprecation and the ownership cleanup require migrations.

But this works!

To deploy socorro database on heroku you can do the following: 
```
heroku addons:add heroku-postgresql:hobby-dev
socorro setupdb --database_url=$(heroku config:get DATABASE_URL -a guarded-crag-2629) --on_heroku --logging.stderr_error_logging_level=0
```

That will give you a verbose deployment :)

The hobby-dev only supports 10k rows, so not useful for a real test, but does validate the approach. 

There are some kinks to work out in the setupdb_app.py -- I disconnect and reconnect quite a few times to set up the database. It would be fair to ask "why are we doing that" and to embark on updating our existing database to reflect the role structure we are actually using, rather than a hypothetical and quite complex system that we don't implement or use.

socorro/external/postgresql/raw_sql/procs/nonzero_string.sql
```diff
@@ -5,7 +5,7 @@ SELECT btrim($1) <> '' AND $1 IS NOT NULL;
 $_$;
 
 
-ALTER FUNCTION public.nonzero_string(citext) OWNER TO postgres;
+-- ALTER FUNCTION public.nonzero_string(citext) OWNER TO postgres;
```
rhelmer: why is this commented out? can just remove the line if it's not useful

socorro/external/postgresql/raw_sql/procs/reports_clean_weekly_partition.sql
```diff
@@ -63,7 +63,7 @@ WHILE rc_indexes[dex] IS NOT NULL LOOP
     dex := dex + 1;
 END LOOP;
 
-EXECUTE 'ALTER TABLE ' || this_part || ' OWNER TO breakpad_rw';
+-- EXECUTE 'ALTER TABLE ' || this_part || ' OWNER TO breakpad_rw';
```
rhelmer: same

selenamarie:
This is being split apart

PR #2630, opened 2015-02-17
fixes bug 1133814 - django-appconf sometimes fails to install with peep

peterbe:
@rhelmer r?

So, the strange thing is that we used to depend on version [0.6](https://pypi.python.org/pypi/django-appconf/0.6) and that only had 1 file to download. So I don't see why we had two shas for that when there was only the tarball available. 

We could investigate this deeper and try to understand why it used to have multiple files on pypi but doesn't any more. 

Instead, I thought I'd just put the latest version in and it works. The [latest version](https://pypi.python.org/pypi/django-appconf/1.0.1) does have a tarball and a wheel.


selenamarie:
@peterbe r+ from me. happy to just upgrade it.

PR #2631, opened 2015-02-18
Bug 1133765 - Update forced versions for correlations for Firefox cycle starting 2015-02-24

KaiRo-at:
The next version of our usual correlation version bump. Should go into production in the week of 2014-02-23.


rhelmer:
I will take care of this - there's a copy running (that uses JSON from postgres not hbase like this old script) that needs to be modified as well.

rhelmer:
I've modified the copy referred to above, so this change should be active starting with tonight's run.

PR #2632, opened 2015-02-19
[DO NOT MERGE] no bug - see if we can run with sudo:no

rhelmer:


rhelmer:
We can indeed do this! I'll file a bug and reopen a proper PR.

PR #2634, opened 2015-02-19
bug 1134686 - Use setUpClass for more efficient test cases

peterbe:
cc @rhelmer @AdrianGaudebert @selenamarie @twobraids 

These changes works locally for me. Instead of creating 93 connections it only creates 23 connections for running the tests in `socorro/unittest/external/postgresql/`.

There are probably many more places where we can reuse fixtures across a whole test case. If you can find more things to tune, please feel free to open a PR against this PR.

### DIFF
--- a/socorro/unittest/external/postgresql/test_backfill.py
+++ b/socorro/unittest/external/postgresql/test_backfill.py
@@ -23,9 +23,9 @@ class TestBackfill(PostgreSQLTestCase):
         """ Populate tables with fake data """
         super(TestBackfill, self).setUp()
 
-        cursor = self.connection.cursor()
+        self.truncate_tables()
 
-        self.tables = []
+        cursor = self.connection.cursor()
 
         for table in staticdata.tables + fakedata.tables:
             # staticdata has no concept of duration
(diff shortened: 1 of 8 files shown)


peterbe:
With the https://github.com/mozilla/socorro/commit/7c493d5f51b178ffb0576bad9b1547368a70b951 commit, the test case went from ~0.22s to ~0.14s.
It's not a big deal but it's an indication that this pattern works and at scale it can make a much bigger difference.

peterbe:
The https://github.com/mozilla/socorro/commit/3d5f62cdd99d8b2b78897ec1f2d027de4fbeb951 change made those tests go from ~0.8s to ~0.4s

peterbe:
The https://github.com/mozilla/socorro/commit/5155e277b837a3a4d0b7ff2eedc09e5eb2b4c656 change made those tests go from ~2.6s to ~0.5s.

peterbe:
I think I'm done optimizing the `socorro/unittest/external/postgresql/` folder now. I ran it repeatedly and only attacked the top 10% (using `--with-timer`) and the rest is already fast enough. 

Who wants to take a look at the ES related tests? (hint hint @AdrianGaudebert or @phrawzty)

AdrianGaudebert:
r+

Regarding the ``elasticsearch`` tests, that module is supposed to be frozen until we are done moving to 1.4 (which is blocked on moving to AWS), and most of the long test will be removed when that's done, so I'm not keen on doing that now.

peterbe:
So, let's stop here for now then. I'm re-running the tests since this getting dated.

PR #2637, opened 2015-02-24
Fixes bug 1136145 - Control SuperSearch whitelist from Super Search Fiel...

AdrianGaudebert:
...ds list.

@peterbe r?

Notes: I looked at the current unit tests and thought they were good enough to test this change (we verify that fields with permissions are not exposed in public mode, and are available in unredacted mode, and we verify that some results are available, which proves that the list is correctly loaded). The function ``get_api_whitelist`` was sufficiently different from the one you wrote in ``crashstats.models`` that I chose to copy it instead of making the original one generic. I think it's easier this way, and less confusing.

### DIFF
--- a/webapp-django/crashstats/supersearch/models.py
+++ b/webapp-django/crashstats/supersearch/models.py
@@ -1,101 +1,38 @@
+import functools
 import json
+
+from django.core.cache import cache
 
 from crashstats import scrubber
 from crashstats.crashstats import models
 
 
-SUPERSEARCH_HITS_WHITELIST = (
-    'additional_minidumps',
-    'addons',
-    'addons_checked',
-    'address',
-    'app_notes',
-    'build_id',
-    'client_crash_date',
-    'completeddatetime',
-    'cpu_info',
-    'cpu_name',
-    'crashedThread',
-    'crash_time',
-    'date',
-    'distributor',
-    'distributor_version',
-    'flash_version',
-    'hangid',
-    'hang_type',
-    'id',
-    'install_age',
-    'java_stack_trace',
-    'last_crash',
-    'platform',
-    'platform_version',
-    'plugin_filename',
-    'plugin_name',
-    'plugin_version',
-    'processor_notes',
-    'process_type',
-    'product',
-    'productid',
-    'reason',
-    'release_channel',
-    'ReleaseChannel',
-    'signature',
-    'startedDateTime',
-    'success',
-    'topmost_filenames',
-    'truncated',
-    'uptime',
-    'user_comments',
-    'uuid',
-    'version',
-    'winsock_lsp',
-    'accessibility',
-    'adapter_device_id',
-    'adapter_vendor_id',
-    'android_board',
-    'android_brand',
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/supersearch/models.py
```diff
@@ -1,101 +1,34 @@
...
-    # deliberately not including:
-    #    email
-    #    url
-    #    exploitability
-)
+def get_api_whitelist(*args, **kwargs):
+
+    def get_from_es(unredacted=False):
+        fields = cache.get('api_supersearch_fields')
+        if fields is None:
```
peterbe: This will always be true :)

webapp-django/crashstats/supersearch/models.py
```diff
@@ -1,101 +1,34 @@
...
+            all = SuperSearchFields().get()
+            fields = []
+            for meta in all.itervalues():
+                if (
+                    meta['is_returned'] and (
+                        unredacted or
+                        not meta['permissions_needed']
+                    )
+                ):
+                    if meta['name'] not in fields:
```
peterbe: just add this to the parent if statement. It's weird that it's not part of it. Is there a reason for this?

peterbe:
Mayhaps I'm not caffinated or drunk enough to understand this but; `unredacted=True` what does that mean? It feels like a double-negative. 
So, setting `unredacted=True` does that mean we include the redacted fields or not? Or is it the other way around. 

I'm still thinking about this and unless I set aside some serious time to understand it I don't by reading that code. If it was something like `include_all_protected_fields=bool` or something my limited brain would stand a better chance :)

Also, it bothers me that the function is inside a closure that is somewhere used in another other form than in...
```python
return models.Lazy(
    functools.partial(get_from_es, *args, **kwargs)
)
```

So all that extra argument is noise. I'm guessing your intent is for this to be used from somewhere else too with a different value sent to `unredacted`. But do we have that? I think not. Will we have that in coming weeks/months?

Perhaps I'm just disappointed in my sleepy IQ but I think this code could do with some simplifications.

webapp-django/crashstats/supersearch/models.py
AdrianGaudebert: Oops... :°)

AdrianGaudebert: Hah, actually, I copy/pasted this from https://github.com/mozilla/socorro/blob/master/webapp-django/crashstats/crashstats/models.py#L61 and guess what? It's the same over there! :P

I'll fix both while I'm at it.

AdrianGaudebert:
@peterbe I think I have addressed your comments. I agree with you on the ``unredacted`` name that was too confusing. I actually don't think I understood myself either, but I went too fast and didn't notice how bad it was.

peterbe:
Ahhh... `include_all_fields` feels soo much better. Thank you!

peterbe:
r+

I don't know if you want to change the argument to a keyword argument but I trust you from hereon to land this.

PR #2638, opened 2015-02-24
Initial changes to support Debian package building

cloakedcode:
I've slaved away on this to get Debian packages building and finally have a usable product. I will add to this PR, or create another one, to update the docs with Debian build examples.

Suffice it to say, setting `BUILD_TYPE` to `deb` when running `make install && make package` does the trick -- assuming you have the build dependencies installed.


rhelmer:
Hey this looks great! I'll take a look, thanks!

scripts/package/deb-after-install.sh
```diff
@@ -0,0 +1,109 @@
...
+chmod 644 /etc/cron.d/socorro
+
+# Link the local Django settings to the distributed settings.
+ln -fs /etc/socorro/local.py $INSTALL_PREFIX/socorro/webapp-django/crashstats/settings/local.py
+
+# change mod_wsgi to python 2.6
+ln -fs /usr/lib/apache2/modules/mod_wsgi.so-2.6 /usr/lib/apache2/modules/mod_wsgi.so
+
+# add postgres user
+#su -c "psql template1 -c \"create user debian_admin with encrypted password '0f6J7sTUspsYzfgnzoTo' superuser\"" postgres
```
rhelmer: probably want to remove this comment

requirements.txt
```diff
@@ -30,6 +30,8 @@ mercurial==2.7.1
 honcho==0.5.0
 # sha256: th84eGr2uNZEtTBVIBLWacdBX89YUt4yDykIANuXmD0
 gunicorn==18.0
+# sha256: 02C9MYmPnfWfb6p4ZVEGW7pFs15-48ObOBtPv-9zkvQ
+South==1.0.2
```
rhelmer: Looks like we already have South specified further down

scripts/package/deb-after-install.sh
```diff
@@ -0,0 +1,109 @@
+#! /bin/bash
```
rhelmer: I wonder if we could safely refactor some of this out, it's really similar to what we have for the RPM packaging. @phrawzty do you have any thoughts on this?

Seeing how long this is I wonder if we're doing too much in ```after-install```, does it seem abnormal for a package install to be doing this much setup? Maybe we can break this out to a cross-distro post-config script?

scripts/package.sh
```diff
@@ -29,13 +29,14 @@ if [ "$BUILD_TYPE" != "tar" ]; then
...
-        --before-remove scripts/package/before-remove.sh \
-        --after-remove scripts/package/after-remove.sh \
+        --before-install scripts/package/$BUILD_TYPE-before-install.sh \
+        --after-install scripts/package/$BUILD_TYPE-after-install.sh \
+        --before-remove scripts/package/$BUILD_TYPE-before-remove.sh \
+        --after-remove scripts/package/$BUILD_TYPE-after-remove.sh \
         --config-files /etc/socorro \
         --exclude *.pyc \
         --exclude *.swp \
+	--deb-field 'Depends: libc6 (>= 2.4), libgcc1 (>= 1:4.1.1), libpq5 (>= 8.3~), libsasl2-2 (>= 2.1.24), libssl1.0.0 (>= 1.0.0), libstdc++6 (>= 4.5), libxml2 (>= 2.7.4), libxslt1.1 (>= 1.1.26), zlib1g (>= 1:1.2.0), openjdk-7-jre-headless, python-virtualenv, postgresql-9.3, postgresql-plperl-9.3, postgresql-contrib-9.3, rsync, rabbitmq-server, elasticsearch, memcached, apache2, libapache2-mod-wsgi, nodejs, node-less' \
```
rhelmer: Thanks for tracking down all these dependencies! (we should really do the same for the RPM)

config/package/socorro/alembic.ini
```diff
@@ -11,7 +11,7 @@ file_template = %%(rev)s_%%(slug)s
 # the 'revision' command, regardless of autogenerate
 # revision_environment = false
 
-sqlalchemy.url = postgresql://postgres@localhost/breakpad
+sqlalchemy.url = postgresql://breakpad_rw:aPassword@localhost/breakpad
```
rhelmer: I'd like to make sure this doesn't break the RPM (travis doesn't test this but we have a jenkins job that builds it)

requirements.txt
rhelmer: This is the cause of the Travis failure on this PR btw:
https://travis-ci.org/mozilla/socorro/builds/52057223#L163

scripts/package.sh
```diff
@@ -29,13 +29,14 @@ if [ "$BUILD_TYPE" != "tar" ]; then
         --vendor "Mozilla" \
         --url "https://wiki.mozilla.org/Socorro" \
         --description "$DESC" \
-        --before-install scripts/package/before-install.sh \
-        --after-install scripts/package/after-install.sh \
-        --before-remove scripts/package/before-remove.sh \
-        --after-remove scripts/package/after-remove.sh \
+        --before-install scripts/package/$BUILD_TYPE-before-install.sh \
```
rhelmer: Good idea

scripts/package/deb-after-install.sh
```diff
@@ -0,0 +1,109 @@
...
+
+# crond doesn't like files with executable bits, and doesn't load
+# them.
+chmod 644 /etc/cron.d/socorro
+
+# Link the local Django settings to the distributed settings.
+ln -fs /etc/socorro/local.py $INSTALL_PREFIX/socorro/webapp-django/crashstats/settings/local.py
+
+# change mod_wsgi to python 2.6
+ln -fs /usr/lib/apache2/modules/mod_wsgi.so-2.6 /usr/lib/apache2/modules/mod_wsgi.so
```
rhelmer: We're planning on dropping the 2.6 requirement soon... you should be able to use 2.7 now and not have to deal with 2.6 if you want.

rhelmer:
@cloakedcode overall this looks really great, thanks!

What distro did you test on (ubuntu/debian, version?)

scripts/package/deb-after-install.sh
phrawzty: Really what's happening here is that we're using FPM instead of the fundamental packaging tools.  This sort of stuff can (and does) occur in packaging all the time but we're not used to seeing it because we don't often examine source packages.  Given that, a cross-distro post-config script would - from a package perspective - be on the janky side; of course, we're already using FPM, which is pretty strong declaration that we're not dedicated to package purity in the first place. :smirk:

Ultimately, either we put this stuff into an omnibus post-install script, or we leave it in the package-specific scripts that FPM expects.  Either way we need to maintain a bunch of shell code, except that in the former case, it's shell code that everybody *expects*, which seems better to me.

cloakedcode:
@rhelmer I tested on Debian 7.1 (stable, wheezy). I haven't tested on Ubuntu, our infra is all Debian.

config/package/socorro/local.py
```diff
@@ -2,8 +2,8 @@
 # These settings overrides what's in settings/base.py
 
 # To extend any settings from settings/base.py here's an example:
-#from . import base
-#INSTALLED_APPS = base.INSTALLED_APPS + ['debug_toolbar']
+from . import base
+INSTALLED_APPS = base.INSTALLED_APPS + ('south',)
```
peterbe: This line shouldn't be needed. That commented out comment is out-of-date. `south` is [already in settings/base.py](https://github.com/mozilla/socorro/blob/master/webapp-django/crashstats/settings/base.py#L29)

scripts/package.sh
```diff
@@ -29,13 +29,14 @@ if [ "$BUILD_TYPE" != "tar" ]; then
...
+        --after-install scripts/package/$BUILD_TYPE-after-install.sh \
+        --before-remove scripts/package/$BUILD_TYPE-before-remove.sh \
+        --after-remove scripts/package/$BUILD_TYPE-after-remove.sh \
         --config-files /etc/socorro \
         --exclude *.pyc \
         --exclude *.swp \
+        --deb-field 'Depends: libc6 (>= 2.4), libgcc1 (>= 1:4.1.1), libpq5 (>= 8.3~), libsasl2-2 (>= 2.1.24), libssl1.0.0 (>= 1.0.0), libstdc++6 (>= 4.5), libxml2 (>= 2.7.4), libxslt1.1 (>= 1.1.26), zlib1g (>= 1:1.2.0), openjdk-7-jre-headless, python-virtualenv, postgresql-9.3, postgresql-plperl-9.3, postgresql-contrib-9.3, rsync, rabbitmq-server, elasticsearch, memcached, apache2, libapache2-mod-wsgi, nodejs, node-less' \
```
Orphis: How about you don't force to have everything installed with the Debian package?
It's probably not really sane to force people to install postgres, elasticsearch, rabbitmq, memcached, node... In my version of this feature, I just had to depend on node-less, the rest seems optional or standard to me.
Also, you couldn't you just use option --depends from fpm?

scripts/package/deb-after-install.sh
Orphis: I'm personally not a big fan of those scripts as they assume that the global configuration is already in place for the post-install scripts to run. If you want a simple package that can be redistributed, you don't want to hardcode the postgresql host and credential in there for alembic to do the upgrades.
The package is also creating all the configuration files and overwriting any existing files, so it's not safe to assume the standard config files are enough for a proper install. A lot of configuration needs to be reapplied after a package install or upgrade.
I'd be more happy with a set of clean puppet files on the side to do everything as needed!

rhelmer:
@Orphis these are all good points. I think the right way to do this (which we're going to pursue next quarter) is to break Socorro up into separate projects, and make the defaults such that things like rabbit, ES and so on are truly optional.

For instance, collector and processor don't really have hard dependencies on postgres/ES/S3 (although processor can optionally store to all of these), but the Django app does have a hard dependency on Postgres (and the search features depend on ES).

However for today we ship everything all together in a single package, with default configs that expect all of these things, so I think it's reasonable to have a .deb equivalent for our current .rpm right now.

Orphis:
@rhelmer The problem is that when you want to setup a new processor by using this package, you'll end up with everything installed. Nobody wants every service installed everywhere.
The setup of Socorro is quite specific already, so I don't think it's even necessary to try to enforce having all the other services installed automatically.

rhelmer:
@Orphis oh I see what you mean, we shouldn't specify hard dependencies from the package... yeah I think you're right.

scripts/package.sh
rhelmer: @cloakedcode would you mind removing this line? I think @Orphis is right that we shouldn't set ```Depends``` here, maybe ```Suggests``` would be ok... then you could still e.g. install this on a dedicated collector or processor node without necessarily having to drag in things like postgres and ES and so forth.

cloakedcode:
@Orphis @rhelmer I've done as suggested and made most of the dependencies suggestions. Keep in mind, some of the previous dependencies in the deb pkg are used in both the deb and rpm post-install scripts (e.g. apache2/httpd). This should be addressed in the rpm scripts. I've tried to address this where able in the deb script.

cloakedcode:
@rhelmer et al. - I don't want steam to be lost here. Is this ready/acceptable to be merged in?

rhelmer:
@cloakedcode understood - yeah I think we can leave the more speculative comments to a separate issue, since this is doing what the RPM does it's Good Enough for now.

It looks like all comments are addressed. Could you rebase this down to a single commit?

Thanks!

cloakedcode:
@rhelmer  As requested, rebased into one commit! Thanks for all your time on this.

rhelmer:
@cloakedcode great, thanks! 
r+, this lgtm (haven't tested but shouldn't break anything as long as the RPM build works)

PR #2641, opened 2015-02-25
Fixes bug 1136917 - update default user to be 'test' for use in tests

selenamarie:
We're using breakpad_rw sometimes and that messes things up once we no longer assume that our core users have superuser privileges.


rhelmer:
r+

PR #2642, opened 2015-02-25
Fixes bug 1136854 removes unnecessary function defs and permission changes

selenamarie:


selenamarie:
@rhelmer ok i removed the commented-out things.. i need this merged before we can proceed with the heroku-specific changes.

selenamarie:
This is just SQL cleanup. Need to merge so that I can get review for the Heroku PR.

PR #2643, opened 2015-02-25
Fixes bug 1132937 add support for heroku URL and heroku pg

selenamarie:
Depends on the changes to test users in an earlier PR.

Also needs the function updates to be rebased in before merging.

### DIFF
--- a/scripts/rabbitmq-integration-test.sh
+++ b/scripts/rabbitmq-integration-test.sh
@@ -53,6 +53,7 @@ function fatal() {
   message=$2
 
   echo "ERROR: $message"
+  cat setupdb.log
 
   cleanup
 
@@ -69,7 +70,7 @@ export PYTHONPATH=.
 echo " Done."
 
 echo -n "INFO: setting up database..."
-python socorro/external/postgresql/setupdb_app.py --database_username=$database_username --database_password=$database_password --database_name=breakpad --database_hostname=$database_hostname --dropdb --force > setupdb.log 2>&1
+python socorro/external/postgresql/setupdb_app.py --database_username=$database_username --database_password=$database_password --database_name=breakpad --database_hostname=$database_hostname --dropdb --force --createdb > setupdb.log 2>&1
 if [ $? != 0 ]
 then
   fatal 1 "setupdb_app.py failed, check setupdb.log"
@@ -103,7 +104,7 @@ echo " Done."
 
 echo -n "INFO: starting up collector, processor and middleware..."
 python socorro/collector/collector_app.py --admin.conf=./config/collector.ini --resource.rabbitmq.host=$rmq_host --secrets.rabbitmq.rabbitmq_user=$rmq_user --secrets.rabbitmq.rabbitmq_password=$rmq_password --resource.rabbitmq.virtual_host=$rmq_virtual_host --resource.rabbitmq.transaction_executor_class=socorro.database.transaction_executor.TransactionExecutor --web_server.wsgi_server_class=socorro.webapi.servers.CherryPy > collector.log 2>&1 &
-python socorro/processor/processor_app.py --admin.conf=./config/processor.ini --resource.rabbitmq.host=$rmq_host --secrets.rabbitmq.rabbitmq_user=$rmq_user --secrets.rabbitmq.rabbitmq_password=$rmq_password --resource.rabbitmq.virtual_host=$rmq_virtual_host --resource.postgresql.database_hostname=$database_hostname > processor.log 2>&1 &
+python socorro/processor/processor_app.py --admin.conf=./config/processor.ini --resource.rabbitmq.host=$rmq_host --secrets.rabbitmq.rabbitmq_user=$rmq_user --secrets.rabbitmq.rabbitmq_password=$rmq_password --resource.rabbitmq.virtual_host=$rmq_virtual_host --resource.postgresql.database_hostname=$database_hostname --secrets.postgresql.database_username=$database_username --secrets.postgresql.database_password=$database_password > processor.log 2>&1 &
 sleep 1
 python socorro/middleware/middleware_app.py --admin.conf=./config/middleware.ini --database.database_hostname=$database_hostname --database.database_username=$database_username --database.database_password=$database_password --rabbitmq.host=$rmq_host --rabbitmq.rabbitmq_user=$rmq_user --rabbitmq.rabbitmq_password=$rmq_password --rabbitmq.virtual_host=$rmq_virtual_host --web_server.wsgi_server_class=socorro.webapi.servers.CherryPy > middleware.log 2>&1 &
 echo " Done."
@@ -128,6 +129,7 @@ function retry() {
       grep 'ERROR' ${name}.log
       if [ $? != 1 ]
       then
+        cat ${name}.log
         fatal 1 "errors found in $name.log"
       fi
       echo "INFO: $name test passed"
--- a/socorro/external/postgresql/postgresqlalchemymanager.py
+++ b/socorro/external/postgresql/postgresqlalchemymanager.py
@@ -8,24 +8,14 @@ Create, prepare and load schema for Socorro PostgreSQL database.
 """
 from __future__ import unicode_literals
 
-import cStringIO
 import logging
 import os
 import re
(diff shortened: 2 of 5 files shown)


socorro/external/postgresql/postgresqlalchemymanager.py
```diff
@@ -43,91 +33,87 @@ def __init__(self, sa_url, logger, autocommit=False):
...
-        self.session.execute('SET check_function_bodies = false')
+        self.on_heroku = on_heroku
+
+    # decorator for heroku stuff
+    def ignore_if_on_heroku(func):
+        def ignored(f_arg, *args, **kwargs):
+            if f_arg.on_heroku:
+                return
+            else:
+                func(f_arg, *args, **kwargs)
```
peterbe: You could just
```python
if not f_arg.on_heroku:
    return func(f_arg, *args, **kwargs)
```

Note also the important `return` that was missing. Otherwise the decorator doesn't work in a "proxy manner".

peterbe: The first parameter you could call `self` instead of `f_arg`.

socorro/external/postgresql/postgresqlalchemymanager.py
```diff
@@ -43,91 +33,87 @@ def __init__(self, sa_url, logger, autocommit=False):
...
         self.session.execute('CREATE EXTENSION IF NOT EXISTS citext')
         self.session.execute('CREATE EXTENSION IF NOT EXISTS hstore')
         # we only need to create the json extension for pg9.2.*
         if not self.min_ver_check(90300):
             self.session.execute(
                 'CREATE EXTENSION IF NOT EXISTS json_enhancements')
+
+    @ignore_if_on_heroku
+    def grant_public_schema_ownership(self, username):
+        logging.debug('granting ownership of public schema')
```
peterbe: This should be `self.logger.debug(...)`

socorro/external/postgresql/postgresqlalchemymanager.py
```diff
@@ -43,91 +33,87 @@ def __init__(self, sa_url, logger, autocommit=False):
...
         self.session.execute(
-            'GRANT ALL ON SCHEMA public TO breakpad_rw')
+            'GRANT ALL ON SCHEMA public TO %s' % username)
 
-    def setup(self):
+    def turn_function_body_checks_off(self):
+        logging.debug('setting body checks off')
```
peterbe: ditto

peterbe:
There are lots of `logging.debug(` which needs to be replaced with `self.logger.debug(`.

scripts/rabbitmq-integration-test.sh
```diff
@@ -53,6 +53,7 @@ function fatal() {
   message=$2
 
   echo "ERROR: $message"
+  cat setupdb.log
```
peterbe: Was this a temporary hacking thing?

rhelmer: It gets the failing log output into the CI output so I think it's a good thing to keep

selenamarie: Yeah - not a temporary hacking thing. We need this for travis.

socorro/external/postgresql/postgresqlalchemymanager.py
```diff
@@ -43,91 +33,87 @@ def __init__(self, sa_url, logger, autocommit=False):
...
+    @ignore_if_on_heroku
+    def set_default_owner(self, database_name, username):
+        logging.debug('setting database %s owner to %s' % (
+            database_name, username
+        ))
         self.session.execute("""
-                ALTER DATABASE %s OWNER TO breakpad_rw
-            """ % database_name)
+                ALTER DATABASE %s OWNER TO %s
+            """ % (database_name, username))
```
peterbe: Shouldn't that be `self.session.execute("ALTER DATABASE %s ...", (database_name, username))`?
You know, in case someone tried to set their username to `'foo'; DROP TABLE some_table`

socorro/external/postgresql/setupdb_app.py
```diff
@@ -219,37 +234,42 @@ def generate_fakedata(self, db, fakedata_days):
...
+        self.database_url = self.config.database_url
+        self.database_name = self.config.database_name
 
-        url_template = connection_url()
-        sa_url = url_template + '/%s' % 'postgres'
+        if self.database_url:
+            connection_url = self.database_url
+        else:
+            if not self.database_name:
+                print "Syntax error: --database_name required"
```
peterbe: Seems pretty late to complain about this.

socorro/external/postgresql/postgresqlalchemymanager.py
selenamarie: I will fix this in a future PR. This is a private admin interface, and while its possible someone might use it maliciously, they'd need admin access to a system that has production credentials first. 

This is not the only place where this might be an issue, and I'd like to fix them all rather than in this one spot.

socorro/external/postgresql/setupdb_app.py
selenamarie: Ok! moving things around a bit.

peterbe:
what about the logging?

socorro/external/postgresql/postgresqlalchemymanager.py
```diff
@@ -43,91 +33,85 @@ def __init__(self, sa_url, logger, autocommit=False):
         self.metadata.bind = self.engine
         self.session = sessionmaker(bind=self.engine)()
         self.logger = logger
+        self.on_heroku = on_heroku
 
-    def setup_admin(self):
-        self.session.execute('SET check_function_bodies = false')
+    # decorator for heroku stuff
+    def ignore_if_on_heroku(func):
```
rhelmer: nice use of a decorator!

rhelmer:
> There are lots of logging.debug( which needs to be replaced with self.logger.debug(.

Looks like this is fixed (closing the loop since github only hides these for line-level comments)

rhelmer:
r+, have not tested but code lgtm and what I'd expect

rhelmer:
I will test this on heroku as soon as I am able, I think it's fine to land now. Thanks @selenamarie !

peterbe:
rebase?

rhelmer:
@selenamarie do you mind rebasing this against master, and squashing so we can merge?

socorro/external/postgresql/setupdb_app.py
```diff
@@ -268,75 +290,88 @@ def create_table(element, compiler, **kw):
...
-            db.set_grants(self.config)  # config has user lists
-            if not self.config['no_staticdata']:
+            if not self.config.get('no_staticdata'):
                 self.import_staticdata(db)
-            if self.config['fakedata']:
-                self.generate_fakedata(db, self.config['fakedata_days'])
+            if self.config.get('fakedata'):
+                self.generate_fakedata(db, self.config.get('fakedata_days'))
             db.commit()
             command.stamp(alembic_cfg, "heads")
```
twobraids: at this point in the code, the ```alembic_cfg``` object has reference to a different ```connection_url``` than the ```PostgreSQLAlchemyManager``` context.  Is this correct?

socorro/external/postgresql/setupdb_app.py
```diff
@@ -268,75 +290,88 @@ def create_table(element, compiler, **kw):
...
                     if not confirm == "y":
                         logging.warn('NOT dropping table')
                         return 2
+                db.drop_database(self.database_name)
+
+        if self.config.createdb:
+            with PostgreSQLAlchemyManager(connection_url, self.config.logger,
+                                          autocommit=False) as db:
+                db.create_database(self.database_name)
+                db.create_roles(self.config)
```
twobraids: with autocommit set to False, why don't we commit at the end?  when the connection is closed at the end of the context-manager section, does it automatically commit or rollback?

socorro/external/postgresql/setupdb_app.py
selenamarie: It is not. This is definitely a bug.

twobraids:
I have superseded this PR with PR 2748

PR #2645, opened 2015-02-27
Fixes bug 1137480 - Removed duplicates from versions in Super Search.

AdrianGaudebert:
@peterbe r?


webapp-django/crashstats/supersearch/forms.py
```diff
@@ -15,6 +15,12 @@
 }
 
 
+# Credits: http://www.peterbe.com/plog/uniqifiers-benchmark
+def uniqify_keep_order(seq):
```
peterbe: cute!

PR #2646, opened 2015-02-27
Fixes bug 1137752 adds reprocessing submission script

selenamarie:
Creates a script for handling bulk reprocessing loads from a production machine.

### DIFF
--- a/config/reprocess_crashlist.ini-dist
+++ b/config/reprocess_crashlist.ini-dist
@@ -1,0 +1,37 @@
+# the fully qualified classname of the app to run
+application=reprocess_crashlist.ReprocessCrashlistApp
+
+[resource]
+
+    [[rabbitmq]]
+
+        +include ./common_rabbitmq.ini
+
+[secrets]
+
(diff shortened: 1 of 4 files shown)


rhelmer:
This looks fine to me, tested on stage and seems to work fine too.

The only thing I see as a potential problem is that this is that it only works when run from the same directory, like ```python ./reprocess_crashlist.py``` - otherwise you get:

```configman.config_exceptions.CannotConvertError: No module named scripts.reprocess_crashlist```

This might make it hard to script, it'd be nice it if didn't care how it was invoked.

selenamarie:
@rhelmer good point! maybe i can add it so that it can be run like the other `socorro` commands.

rhelmer:
@selenamarie oh good idea! I think it just needs to be added here and ```socorro reprocess_crashlist``` would work: https://github.com/mozilla/socorro/blob/master/socorro/app/for_application_defaults.py#L46-L58

rhelmer:
r+

PR #2647, opened 2015-02-27
Fixes bug 1137879 add support for postgres:// or postgresql:// URLs

selenamarie:


socorro/external/postgresql/connection_context.py
```diff
@@ -6,10 +6,14 @@
...
 import psycopg2
 import psycopg2.extensions
+import urlparse
 
 from configman.config_manager import RequiredConfig
 from configman import Namespace
 
+urlparse.uses_netloc.append('postgres')
+urlparse.uses_netloc.append('postgresql')
+
```
peterbe: These might require a comment. Why is it needed? Also, the first part of a URI is usually called the "scheme" E.g. "ftp" but this talks about "netloc" which is usually the alias for "hostname".

socorro/external/postgresql/connection_context.py
```diff
@@ -49,6 +53,13 @@ class ConnectionContext(RequiredConfig):
         reference_value_from='secrets.postgresql',
         secret=True,
     )
+    required_config.add_option(
+        name='database_url',
+        default=None,
+        doc='URI for connecting to the database, '
+            'format: postgres://username:pass@hostname:port/database',
```
peterbe: Hurray!

socorro/external/postgresql/connection_context.py
```diff
@@ -68,9 +79,20 @@ def __init__(self, config, local_config=None):
...
             local_config = config
+
+        if local_config.get('database_url'):
+            url = urlparse.urlparse(local_config['database_url'])
+            local_config['database_username'] = url.username
+            local_config['database_password'] = url.password
+            local_config['database_hostname'] = url.hostname
+            local_config['database_port'] = url.port
+            local_config['database_name'] = url.path[1:]
+
```
peterbe: Hmm... what bothers me here is that we're taking these pieces and then create a string with it later (e.g `self.dsn = ...`)
So, if you leave the port out of the URL, which I think is a sensible thing to do. then `local_config['database_port'] is None` and when this is inserted into the `dsn` string you'll end up with `... port=None ...` which postgres won't like. 

Like we discussed in IRC, I don't think we should do `local_config['database_port'] = url.port or local_config['database_port']` because that's implicit and just odd. 
We could instead do: `local_config['database_port'] = url.port or 5432` but then we have to know that the default postgres port is 5432. 

How about...

We *can't* do `psycopg2.connect('postgres://user@password:localhost/mydb')` but you **can** do `psycopg2.connect(user="user", password="password", host="localhost", dbname="mydb")`
I.e. the equivalent for our use case would be `psycopg2.connect(**kwargs)`. 

A neat thing about this is that we can leave this as none and the psycopg2 will automatically know to set the default. This is also applicable not only to the port but to things like the username. 

[Full documentation here](http://initd.org/psycopg/docs/module.html)

cc @twobraids

socorro/unittest/external/postgresql/unittestbase.py
```diff
@@ -119,6 +127,14 @@ def setUpClass(cls):
...
         cls.config = cls.get_standard_config()
+        if cls.config.get('database_url'):
+            url = urlparse.urlparse(cls.config.database_url)
+            cls.config.database_username = url.username
+            cls.config.database_password = url.password
+            cls.config.database_hostname = url.hostname
+            cls.config.database_port = url.port
+            cls.config.database_name = url.path[1:]
+
         cls.database = db.Database(cls.config)
```
peterbe: Shouldn't this be solved inside that `db.Database` class instead? I.e. *it* could look out for a `database_url` key.

socorro/unittest/external/postgresql/unittestbase.py
```diff
@@ -2,11 +2,15 @@
...
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import urlparse
+
 import socorro.database.database as db
 from configman import ConfigurationManager, Namespace
 from configman.converters import list_converter
 from socorro.unittest.testbase import TestCase
 
+urlparse.uses_netloc.append('postgres')
```
peterbe: Why only one here?

socorro/external/postgresql/connection_context.py
```diff
@@ -6,10 +6,14 @@
 import contextlib
 import psycopg2
 import psycopg2.extensions
+import urlparse
```
peterbe: Ultrapicky, a good convention is to separate python standard lib from third-party libs. Can you move this to appear just after `import contextlib` and then add an extra space. The `import psycopg2` stuff belongs next to the `from configman.config_manager...` lines.

peterbe:
Two things:

1) The problem with `port==None` being inserted into the `self.dsn` string would still become `"... port=None ..."` as a string. 

2) I don't like the DSN. I think we should use keywords arguments instead. Is that out of scope here?

twobraids:
I see a potential problem.  the PG connection context will likely be embedded in namespaces, perhaps several deep.  that would mean that in order for configman to grab the value of the URL from the environment, the environment var would have to reflect that same nesting.   

Do you want to have this PG URL behavior for _all_ the Socorro apps, or do you have it in mind just for one in particular?

selenamarie:
@twobraids we need support for this in all apps. I had originally made a similar change in `setupdb_app.py`, and then realized this was needed everywhere, so I figured that I could set it in connection_context and rely on the `common_database.ini` to cover most of our use cases. 

In my AWS instance, I'm using the integration test command-line invocations that use `resource.postgresql` and that seems to not require additional tweaking.

selenamarie:
Note from IRC conversation:  `resources__postgresql__database_url` is the environment variable needed to override configuration of the URL in production.

socorro/unittest/external/postgresql/unittestbase.py
```diff
@@ -19,11 +21,15 @@ class PostgreSQLTestCase(TestCase):
     required_config = Namespace()
     required_config.namespace('database')
     required_config.add_option(
+        name='database_url',
+        default='postgres://test:aPassword@localhost:5432/socorro_integration_test',
+        doc='Name of database to manage',
+    )
+    required_config.add_option(
```
peterbe: Why keep these old options then?

socorro/external/postgresql/connection_context.py
peterbe:
```python
else:
    print "Deprecationism! Please use database_url instead. Everywhere!!!"
```

socorro/unittest/external/postgresql/unittestbase.py
selenamarie: I was trying not to make the heroku-specific stuff be mandatory. Having the password embedded in the URL might be an issue for some users as well.

peterbe: Using a URL isn't Heroku-specific. I use it whenever I can when I do non-django. The reason I like it is because it files into one environment variable instead of 5. 
Also, wherever you type in `database_password` you'd also type in `database_url`. 

Either way, I'm not convinced. But I am convinced that you know better. :)
My comment was just a question after all :)

socorro/unittest/external/postgresql/unittestbase.py
```diff
@@ -19,11 +21,15 @@ class PostgreSQLTestCase(TestCase):
     required_config = Namespace()
     required_config.namespace('database')
     required_config.add_option(
+        name='database_url',
+        default='postgres://test:aPassword@localhost:5432/socorro_integration_test',
```
peterbe: If the default was `'postgres://localhost:5432/socorro_integration_test'` instead this'd work for most people out-of-the-box because then internally it will connect with the user running nose and with a blank password which is a common setup for local dev.

socorro/unittest/external/postgresql/unittestbase.py
```diff
@@ -2,6 +2,8 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import urlparse
```
peterbe: pyflakes?

socorro/external/postgresql/connection_context.py
peterbe: What about other things like `?sslmode=verify` or something like that? Or `&client_encoding=utf8`. I'm sort of making these up but my point is that there might be additional query string parameters there that someone might want to send in to the `psycopg2.connect(**options)` thing. 
Perhaps this is not the time and place but I think if we refactor ConnectionContext further I'd like to make it stop using a DSN string and instead unpack everything in this URI into the keyword arguments form for `psycopg2.connect()`.

socorro/external/postgresql/connection_context.py
```diff
@@ -49,6 +51,13 @@ class ConnectionContext(RequiredConfig):
         reference_value_from='secrets.postgresql',
         secret=True,
     )
+    required_config.add_option(
+        name='database_url',
+        default=None,
```
twobraids: rather than set the default to none, this is where you should fetch the value from the environment. 
```python
    import os
    required_config.add_option(
        name='database_url',
        default=os.environ('database_url', None),
        # ...
```

selenamarie: Ah perfect. Thank you.

socorro/unittest/external/postgresql/unittestbase.py
selenamarie: good point! Thanks.

socorro/unittest/external/postgresql/unittestbase.py
selenamarie: I think it's worth exploring the rest of the changes in another PR. It's just a lot of code change that touches a bunch of things like integration tests, so I'd like to take this step now, and rip out the rest of the stuff later.

selenamarie: Exploration revealed depedencies on `socorro.database.database` in our tests and middleware. Going back to just supporting the minimal change of adding support for `database_url` without deprecating DSN, pending deprecation of `socorro.database.database`.

scripts/test.sh
```diff
@@ -8,23 +8,27 @@ JENKINS_CONF=jenkins.py.dist
 
 ENV=env
 
-PYTHONPATH="."
+PYTHONPATH='.'
```
peterbe: Why this change? Why quotation at all?

scripts/rabbitmq-integration-test.sh
```diff
@@ -69,7 +70,8 @@ export PYTHONPATH=.
 echo " Done."
 
 echo -n "INFO: setting up database..."
-python socorro/external/postgresql/setupdb_app.py --database_username=$database_username --database_password=$database_password --database_name=breakpad --database_hostname=$database_hostname --dropdb --force > setupdb.log 2>&1
+# Getting config info from database_url
+python socorro/external/postgresql/setupdb_app.py --database_url=$database_url --dropdb --force > setupdb.log 2>&1
```
peterbe: Ahhh.. niiiice. Why didn't we do this earlier?

scripts/test.sh
```diff
@@ -8,23 +8,27 @@ JENKINS_CONF=jenkins.py.dist
...
-fi
-if [ -n "$database_port" ]; then
-    PG_RESOURCES="$PG_RESOURCES resource.postgresql.database_port=$database_port"
-fi
-if [ -n "$database_name" ]; then
-    PG_RESOURCES="$PG_RESOURCES resource.postgresql.database_name=$database_name"
+if [ -n "$database_url" ]; then
+    PG_RESOURCES="resource.postgresql.database_url=$database_url"
+else
+    if [ -n "$database_hostname" ]; then
```
peterbe: A comment here that says that this is all legacy stuff that can be removed some day?

selenamarie: Yeah good idea.

scripts/test.sh
selenamarie: That's debugging cruft. There's no need for the quotes, so I'll just remove them when I add comments about deprecation.

socorro/external/postgresql/connection_context.py
selenamarie: I don't want to put a print statement here, but I'll add a comment. This is used by so many classes, there will be a ton of noise in the logs.

socorro/external/postgresql/connection_context.py
```diff
@@ -68,21 +81,41 @@ def __init__(self, config, local_config=None):
             local_config - this is the namespace within the complete config
                            where the actual database parameters are found"""
         super(ConnectionContext, self).__init__()
+
         self.config = config
+
         if local_config is None:
             local_config = config
+
+        if local_config['database_url']:
```
peterbe: Why is this safe? Why not `if local_config.get('database_url'):`?

socorro/external/postgresql/connection_context.py
peterbe: Thank you!

socorro/external/postgresql/connection_context.py
selenamarie: Oops - I will fix this. Sorry about that.

socorro/external/postgresql/connection_context.py
```diff
@@ -68,21 +81,41 @@ def __init__(self, config, local_config=None):
...
+            url = urlparse.urlparse(local_config['database_url'])
+            local_config['database_username'] = url.username
+            local_config['database_password'] = url.password
+            local_config['database_hostname'] = url.hostname
+            local_config['database_port'] = url.port
+            local_config['database_name'] = url.path[1:]
+
+            self.db_kwargs = {
+                'host': url.hostname or 'localhost',
+                'user': url.username or os.environ.get('USER', 'test'),
```
peterbe: Shouldn't all of these defaults already be built into the default `database_url`? In particular that user field looks like something that can cause debugging confusion some day. 

Why not do something like this:
```
self.db_kwargs = {}
if url.hostname:
    self.db_kwargs['host'] = url.hostname
...
```
Then we're leaving the defaultism up to `psycopg2.connect()` instead.

socorro/external/postgresql/connection_context.py
```diff
@@ -68,21 +81,41 @@ def __init__(self, config, local_config=None):
...
+                'password': url.password or 'aPassword',
+                'port': url.port or 5432,
+                'database': url.path[1:] or 'postgres'
+            }
+
```
peterbe: You might want to add an `else:` here because `self.dsn` does not need to be created if you have `self.db_kwargs`

socorro/external/postgresql/setupdb_app.py
```diff
@@ -219,9 +232,48 @@ def generate_fakedata(self, db, fakedata_days):
...
+            database_url = self.database_url
+            url = urlparse.urlparse(database_url)
+            database_username = url.username
+            database_password = url.password
+            database_hostname = url.hostname
+            database_port = url.port
+            # Set the self.database_name because it wasn't set
+            # intentionally by --database_name if database_url
+            # is defined
+            self.database_name = url.path[1:]
```
peterbe: Hurts my soul that this is the third time I see `urlparse()` being used. Can we simply just add a `def parse_database_url(url)` function to `socorro/lib/database_utils.py` or something?

socorro/external/postgresql/setupdb_app.py
```diff
@@ -219,9 +232,48 @@ def generate_fakedata(self, db, fakedata_days):
...
+            database_username = self.config.get('database_superusername')
+            database_password = self.config.get('database_superuserpassword')
+
+        # construct a URL
+        url = 'postgresql://'
+        if database_username:
+            url += '%s' % database_username
+        if database_password:
+            url += ':%s' % database_password
+        url += '@'
```
peterbe: What if `not database_username and not database_password`?
Then you get `postrgesql://@localhost:5432/foo`? Is that a problem?

socorro/external/postgresql/connection_context.py
```diff
@@ -21,34 +24,44 @@ class ConnectionContext(RequiredConfig):
     required_config.add_option(
         name='database_hostname',
         default='localhost',
-        doc='the hostname of the database',
+        doc='the hostname of the database (deprecated)',
```
peterbe: Wondeful!

peterbe:
r+ with some optional nits

socorro/external/postgresql/connection_context.py
selenamarie: That makes sense to change around the variable setting.

As to your other extremely valid point, I don't have enough time left to debug the default settings as you suggest. 

Our options are: 

* abandon this patch based on your concern and I can come back to it around April 25th. The problem with that is that then the core socorro stuff doesn't support the Heroku URLs, and someone will need to take the Heroku URL from consul, parse it into our configs and update our existing configs as-is.
* Accept the patch even with the annoying defaults in it and revisit the problems with our unit tests, integration tests and jenkins/travis runs later

twobraids:
I am replacing this PR with one that does not suffer override ambiguities.

PR #2648, opened 2015-03-02
fixes Bug 1138488 - added supplemental processing to ProcessorApp

twobraids:
originally slated to create a new processor type, instead this PR modifies the existing processor_app.  It adds the ability for the processor to fetch an existing processed crash along with the raw crash and dumps.  The existing processed crash is then passed on to the configured processor algorithm.  For processor2015, the rules may work with the processed crash at will or ignore it entirely.  

This method of expanding the abilities of the existing processor_app class cuts down significantly on the amount of redundant code in comparison with my original solution.

### DIFF
--- a/socorro/processor/hybrid_processor.py
+++ b/socorro/processor/hybrid_processor.py
@@ -310,7 +310,7 @@ class HybridCrashProcessor(RequiredConfig):
             os.unlink(file_pathname)
 
     #--------------------------------------------------------------------------
-    def convert_raw_crash_to_processed_crash(self, raw_crash, raw_dumps):
+    def process_crash(self, raw_crash, raw_dumps, ignored_processed_crash):
         """ This function is run only by a worker thread.
             Given a job, fetch a thread local database connection and the json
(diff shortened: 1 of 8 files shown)


socorro/processor/hybrid_processor.py
```diff
@@ -310,7 +310,7 @@ def _temp_raw_crash_json_file(self, raw_crash, crash_id):
             os.unlink(file_pathname)
 
     #--------------------------------------------------------------------------
-    def convert_raw_crash_to_processed_crash(self, raw_crash, raw_dumps):
+    def process_crash(self, raw_crash, raw_dumps, ignored_processed_crash):
```
twobraids: this change represents the idea that the ```convert_raw_crash_to_processed_crash``` method name was no longer accurate.  The name ```process_crash``` covers both cases where raw is converted to processed, and processed is augmented by additional processing.

twobraids:
@rhelmer r?

rhelmer:
r+

PR #2649, opened 2015-03-02
Nobug add starting elasticsearch to production install docs

selenamarie:
ES is now a requirement so... maybe we should have this in our production install docs.

### DIFF
--- a/docs/production-install.rst
+++ b/docs/production-install.rst
@@ -69,6 +69,11 @@ Enable RabbitMQ on startup:
   sudo service rabbitmq-server start
   sudo chkconfig rabbitmq-server on
 
+Enable RabbitMQ on startup:
+::
+  sudo service elasticsearch start
+  sudo chkconfig elasticsearch on 
+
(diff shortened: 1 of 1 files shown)


rhelmer:
Do we still have a waffle flag around ```/search``` in the django app?

rhelmer:
r+

selenamarie:
@rhelmer i see some decorators for it and app_api_all in supersearch, api and the signature apps.

rhelmer:
@selenamarie oh looks like that was fixed in bug 1052355, it's not directly related to this doc change anyway (we need it running or processor will complain)

docs/production-install.rst
```diff
@@ -69,6 +69,11 @@ Enable RabbitMQ on startup:
   sudo service rabbitmq-server start
   sudo chkconfig rabbitmq-server on
 
+Enable RabbitMQ on startup:
```
AdrianGaudebert: I suppose you meant ``Enable Elasticsearch`` here? :)

PR #2650, opened 2015-03-03
Fixes Bug 1121469 - a service for tagging a processed crash

twobraids:
here's a dataservice that can add tags to processed crashes one at a time.

These tags can be used to hang flags on a processed crash that will trigger special post processing rules when the crash is run through a secondary processor.  The idea is to put a second set of processors up to handle classifications and other post-processing  transforms on demand.  

For example, the dump analyzer "get-minidump-instructions" is to be run only on demand.  Using this service, we can tag a crash with 'get-minidump-instructions' and use this service to queue the crash for post processing.  The secondary or post-processor_app will grab the crash and run it through its rule set.  Presumably, there will be a rule that will detect the 'get-minidump-instructions' in the predicate and then invoke the appropriate binary on the crash's dumps.


twobraids:
@selenamarie r?

twobraids:
withdrawn pending dataservice refactoring

PR #2651, opened 2015-03-03
the current Mozilla production install has symlinks to these files, make...

rhelmer:
... copies of the RPM version for the moment


selenamarie:
r+ (looks like there are some tabbing issues in there.. but i imagine it doesn't matter)

rhelmer:
@selenamarie thanks - yeah this is just a copy of the "rpm" version of these scripts, we can remove them once we're out of PHX.

PR #2652, opened 2015-03-09
Output code_id in stackwalker's module list

luser:
We'll need this to fix bug 1140358. r? @twobraids


twobraids:
r+   seems perfectly reasonable

PR #2653, opened 2015-03-11
fix bug 1142149 - make it possible to override python2.6 for virtualenv

rhelmer:
r? @phrawzty


rhelmer:
@phrawzty this still defaults to 2.6 (for the benefit of our jenkins/PHX infra), but you can override it e.g.:
```
make test PYTHON=python2.7
```

phrawzty:
Travis failed but it doesn't have anything to do with this PR as far as I can tell.

`r+`

PR #2654, opened 2015-03-11
fix bug 1142182 - make localhost the default in alembic.ini-dist, make j...

rhelmer:
...enkins override

r? @phrawzty @selenamarie - we already have a place for jenkins-specific stuff, it uses the fact that ```$WORKSPACE``` is set to detect jenkins, might be better to use ```$USER``` as @erikrose suggests.


selenamarie:
r+ yay thanks @rhelmer

PR #2657, opened 2015-03-12
Revert "fixes bug 1085530 - store uploaded symbols to S3, r=@rhelmer"

rhelmer:
This reverts commit 91db6f053cfc1ba8524417c2c58345c6338e9939.


rhelmer:
We're not quite ready to ship this yet (still using NFS symbol store), will reland when we are.

PR #2660, opened 2015-03-12
fix bug 1142617 - do not force python version, but allow setting PYTHON ...

rhelmer:
...env var to override

r? @phrawzty


phrawzty:
`r+` pending Jenkins.

PR #2663, opened 2015-03-13
fixes bug 1022097 - switch from check.py to flake8

peterbe:
@rhelmer r?


peterbe:
I might do another temp config to see that travis trips it.

rhelmer:
r+

PR #2665, opened 2015-03-13
fixes bug 1119200, 1119203 - GA analytics to tabs

peterbe:
@AdrianGaudebert r?
cc @rhelmer @lonnen 

I combined two bugs in one git commit message. Let's see if that works. 

To read more about event tracking see https://developers.google.com/analytics/devguides/collection/analyticsjs/events

### DIFF
--- a/webapp-django/crashstats/base/templates/crashstats_base.html
+++ b/webapp-django/crashstats/base/templates/crashstats_base.html
@@ -22,18 +22,6 @@
           ga('send', 'pageview');
         </script>
   {% endif %}
-
-    <script>
-        // Track a page view in our analytics tracker. To be called by pages
-        // that use AJAX requests to load content.
-        {% if GOOGLE_ANALYTICS_ID %}
-            function socorroTrackUse(url) {
-                ga('send', 'pageview', url);
-            }
-        {% else %}
-            function socorroTrackUse(url) {}
-        {% endif %}
-    </script>
     </head>
     <body>
     {{ browserid_info() }}
@@ -174,6 +162,7 @@
     <script src="{{ static('crashstats/js/jquery/jquery-2.0.3.min.js') }}"></script>
     <script src="{{ static('crashstats/js/jquery/plugins/jquery.cookies.2.2.0.js') }}"></script>
     <script src="{{ static('crashstats/js/socorro/nav.js') }}"></script>
+    <script src="{{ static('crashstats/js/socorro/analytics.js') }}"></script>
     {% endcompress %}
     {{ browserid_js() }}
     {% endblock %}
--- a/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/analytics.js
+++ b/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/analytics.js
@@ -1,0 +1,20 @@
+/*global ga */
+
+var Analytics = (function() {
+    'use strict';
+
+    function wrap() {
+        if (typeof ga !== 'undefined') {
+            ga.apply(undefined, arguments);
+        }
+    }
+
+    return {
+        trackTab: function(page, id) {
+            wrap('send', 'event', 'tab', page, id);
+        },
+        trackPageview: function(url) {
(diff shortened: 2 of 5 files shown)


peterbe:
Now, we just need to remember to learn to how to use this inside GA. But at least the data will be there.

AdrianGaudebert:
For Super Search I have been using this function: https://github.com/mozilla/socorro/blob/master/webapp-django/crashstats/base/templates/crashstats_base.html#L30

I now see that the way I did it is much worse than yours (checking ``ga !== undefined`` is much better, and that would allow putting that function in a JS file instead of in that template), but I still like the fact that GA is encapsulated in another function. Wouldn't it be nice to have some sort of object with methods that encapsulate our GA calls? What do you think?

peterbe:
I think the supersearch should use the `event` API and not the `pageview` API. I didn't want to solve that here and now. Care to own that?

> Wouldn't it be nice to have some sort of object with methods that encapsulate our GA calls? What do you think? 

Good thought. The call is very specific to Google Analytics and if we one day were to switch to a GA competitor it's very likely that the API'd be very very different and you'd need to understand that difference where you call things. If we build an awesome and robust wrapper that is used centrally we might be solving a massive problem we don't have. I don't think these small and few `ga(...)` calls are that hard to read and understand as it is today.

schalkneethling:
Hey All ;) You might to have a look at this, that is used extensively on mozilla.org:
https://github.com/mozilla/bedrock/blob/master/media/js/base/global.js#L153

A sample of use may be https://github.com/mozilla/bedrock/blob/master/media/js/base/global.js#L67

AdrianGaudebert:
> If we build an awesome and robust wrapper that is used centrally we might be solving a massive problem we don't have. I don't think these small and few ga(...) calls are that hard to read and understand as it is today. 

I agree it's bike-shedding, but I'm concerned about the quality of our front-end JS code. I feel like we are just piling stuff on top of other things, and that's become a big mess. In this instance it's an example of a bad separation of contexts, the code handling tabs in our pages should not know about Google Analytics imo. Maybe I'm overthinking it and it's not that true, maybe it's true and we should not care because it's already messy and we are likely to rewrite the front-end eventually, or maybe we should start putting more thoughts in the organisation of our JS code. 

However, I am not going to get in the way of such a simple patch. You have my r+. I just wanted to express what I'm feeling and start some sort of discussion about it. 

@schalkneethling What you shared is nice, but not exactly what I was hopping for. I think it's still an improvement from what we currently have, though.

peterbe:
So what do you propose? Basically, we'd need to rely on a globally available function. We could do something like this:
```javascript
/*global TabSwitch */
...
    $('#report-index').tabs({
        selected: 0,
        activate: function(event, ui) {
            TabSwitch.notice('report_index', ui.newPanel.attr('id'));
        }
    });
```
And the definition for that would look something like this:
```javascript
var TabSwitch = (function() {
    return {
        notice: function(page_id, tab_id) {
            if (typeof ga !== 'undefined') 
              ...
        }
    };
})();
```

What else did you have in mind to do with it? We could potentially use it to either update the location hash or to do a pushState thing.

AdrianGaudebert:
I think I would go for something even more focused on analytics. 

```javascript
var Analytics = {
    trackUrl: function () {},
    trackTab: function () {},
}

if (typeof ga !== 'undefined') {
    Analytics.trackUrl = function (url) {
        ga('send', 'pageview', url);
    };

    Analytics.trackTab = function (pageName, tabName) {
        ga('send', 'event', 'tab', pageName, tabName, true);
    };
}
```

The idea here is to not repeat the ``ga is undefined`` test. I could see us making a better system for tabs and history, but that seems out of scope, and probably something to keep for an eventual rewrite.

peterbe:
I like that but I don't like the idea of redefining the functions. I think I'd prefer a module pattern and put the `typeof ga` negotiation inside instead.

AdrianGaudebert:
Awesome! I love it! r+ when travis is happy. :)

PR #2666, opened 2015-03-16
fixes bug 1131398 - move setup of product id map to constructor

twobraids:
this fixes the bug where the logger doesn't work properly when the PG password is incorrect: "Nonetype doesn't have..."

the ProductIdMap is set up by a configman aggregation just like logging is setup by an aggregation.  they are order dependent, but configman does not guarantee order.  the solution is to move the ProductIdMap to the constructor of the rule rather than as an aggregation.


twobraids:
@rhelmer r?

rhelmer:
r+ have not tested but lgtm

PR #2667, opened 2015-03-17
Fixes bug 1144094 - Prevented empty parameters to break Super Search.

AdrianGaudebert:
@peterbe r?

### DIFF
--- a/webapp-django/crashstats/supersearch/tests/test_views.py
+++ b/webapp-django/crashstats/supersearch/tests/test_views.py
@@ -541,6 +541,13 @@ class TestViews(BaseTestViews):
         ok_('1.0' not in response.content)
         ok_('2017' not in response.content)
 
+        # Test missing parameters don't raise an exception.
+        response = self.client.get(
+            url,
+            {'product': 'WaterWolf', 'date': '', 'build_id': ''}
+        )
+        eq_(response.status_code, 200)
+
     @mock.patch('requests.post')
     @mock.patch('requests.get')
     def test_search_results_admin_mode(self, rget, rpost):
--- a/webapp-django/crashstats/supersearch/form_fields.py
+++ b/webapp-django/crashstats/supersearch/form_fields.py
@@ -97,8 +97,9 @@ class MultiplePrefixedValueField(PrefixedField):
                 *args,
                 **kwargs
             )
-            cleaned_values.append(cleaned_value)
-            prefixed_values.append(self.prefixed_value)
+            if cleaned_value is not None:
+                cleaned_values.append(cleaned_value)
+                prefixed_values.append(self.prefixed_value)
 
         self.prefixed_value = prefixed_values
         return cleaned_values


webapp-django/crashstats/supersearch/views.py
```diff
@@ -240,7 +240,14 @@ def get_report_list_parameters(source):
     params = {}
 
     for key, value in source.items():
-        if not value:
+        if (
+            not value or (
+                # if value == [None]
+                isinstance(value, list) and
+                len(value) == 1 and
+                value[0] is None
```
peterbe: What's the difference between this and `if value == [None]`?

.travis.yml
```diff
@@ -16,7 +16,7 @@ services:
 before_install:
   - npm install -g less
   - gem install puppet puppet-lint
-  - wget 'https://download.elasticsearch.org/elasticsearch/elasticsearch/elasticsearch-0.90.10.tar.gz'
+  - wget 'https://download.elasticsearch.org/elasticsearch/elasticsearch/elasticsearch-0.90.10.tar.gz' --no-check-certificate
```
peterbe: Isn't this making things less secure?

webapp-django/crashstats/supersearch/tests/test_views.py
```diff
@@ -851,6 +851,16 @@ def test_get_report_list_parameters(self):
...
 
+        # Test missing parameters don't break.
+        source = {
+            'date': [None],
+            'product': [None],
+            'version': [None],
+            'release_channel': [None],
+            'build_id': [None],
+        }
+        res = get_report_list_parameters(source)
```
peterbe: If the purpose is to check that it just doesn't raise an error, then don't bother with the `res =`

webapp-django/crashstats/supersearch/views.py
```diff
@@ -240,7 +240,14 @@ def get_report_list_parameters(source):
...
+            )
+        ):
```
peterbe: Basically, why is it being sent `[None]` at all? Can we attack that problem first? 
I started looking at [these lines](https://github.com/AdrianGaudebert/socorro/blob/1144094-super-search-form-validation/webapp-django/crashstats/supersearch/views.py#L84-L86) to understand where it might come from but I'm no wiser.

webapp-django/crashstats/supersearch/tests/test_views.py
```diff
@@ -851,6 +851,16 @@ def test_get_report_list_parameters(self):
 
         eq_(res['build_id'], ['12345'])
 
+        # Test missing parameters don't break.
```
peterbe: "break" is ambiguous. Can you replace with with `Test missing parameters don't raise an exception` because that's really what you're after, isn't it?

.travis.yml
AdrianGaudebert: It does, but see https://github.com/mozilla/socorro/pull/2669 and https://bugzilla.mozilla.org/show_bug.cgi?id=1144158 for context. This line breaks our build and it's the only workaround we have for the moment.

webapp-django/crashstats/supersearch/views.py
AdrianGaudebert: Is it OK to do ``if value == [None]``? Iirc doing ``smth == None`` is a bad thing to do, so I thought it would also be bad to do ``== [None]``.

AdrianGaudebert:
@peterbe Brand new patch, as you suggested I went looking for the real culprit upper in the chain, and I found it. This is much simpler now. r?

peterbe:
I'm kinda proud of myself with that first code review :)
Glad you went "up the chain" to the root problem.

PR #2671, opened 2015-03-17
fixes bug 1144293 - added MigrationCrashStore

twobraids:
create a crashstore that wraps two crashstores and decides between them by comparing the date embedded in the crash_id with a threshold date given to the crash store.  If the embedded date is before the threshold, save/get from the fallback storage system.  Otherwise save/fetch from the primary crashstore

### DIFF
--- a/socorro/external/crashstorage_base.py
+++ b/socorro/external/crashstorage_base.py
@@ -652,6 +652,120 @@ class FallbackCrashStorage(CrashStorageBase):
 
 
 #==============================================================================
+class MigrationCrashStorage(FallbackCrashStorage):
+    required_config = Namespace()
+    required_config.add_option(
+        'date_threshold',
+        default="150401",
+        doc="a date before which seconday storage is to "
+            "be used (YYYMMDD)",
+    )
+
+    #--------------------------------------------------------------------------
+    def save_raw_crash(self, raw_crash, dumps, crash_id):
+        """save raw crash data to the primary storage iff the date embedded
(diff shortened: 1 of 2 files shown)


twobraids:
@selenamarie please review to see if this does what you believe to be the right thing.

socorro/external/crashstorage_base.py
```diff
@@ -652,6 +652,120 @@ def new_crashes(self):
 
 
 #==============================================================================
+class MigrationCrashStorage(FallbackCrashStorage):
+    required_config = Namespace()
+    required_config.add_option(
+        'date_threshold',
+        default="150401",
```
selenamarie: This is great!  Would it be prudent to check that this string conforms to our expectations about date format? I'd hate for someone to fat finger the config.

selenamarie:
@twobraids yes! I believe this does exactly what we'd like. I especially like being able to set this to a date in the future -allowing us to have a precise cutover date that falls on a reporting date boundary.

peterbe:
Who should review this?

socorro/external/crashstorage_base.py
twobraids: there are many places in socorro config where fat fingering would cause problems.  There has not been a history of writing code that verifies values coming in from config as being of the correct type or domain.  I don't see a particularly compelling reason to start do such verifications in this location.

selenamarie:
r+

PR #2680, opened 2015-03-19
bug 1130598 - simpler processor default

rhelmer:
r? @twobraids - this makes processor work with collector without any additional configuration, and use a default ruleset that does not use any external resources (like postgres) or anything Mozilla-specific.

### DIFF
--- a/socorro/processor/processor_app.py
+++ b/socorro/processor/processor_app.py
@@ -11,12 +11,15 @@ from configman import Namespace
 from configman.converters import class_converter
 
 from socorro.app.fetch_transform_save_app import FetchTransformSaveApp, main
-from socorro.external.filesystem.crashstorage import FileSystemRawCrashStorage
 from socorro.external.crashstorage_base import (
   PolyCrashStorage,
   CrashIDNotFound,
 )
 from socorro.lib.util import DotDict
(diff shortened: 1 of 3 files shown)


socorro/processor/processor_app.py
```diff
@@ -11,12 +11,15 @@
...
-from socorro.external.filesystem.crashstorage import FileSystemRawCrashStorage
 from socorro.external.crashstorage_base import (
   PolyCrashStorage,
   CrashIDNotFound,
 )
 from socorro.lib.util import DotDict
+from socorro.external.fs.fs_new_crash_source import (
+  FSNewCrashSource
+)
+from socorro.external.fs.crashstorage import FSLegacyDatedRadixTreeStorage
```
twobraids: would you consider using some of the more friendly named FS Crash Stores?  Are you storing both raw and processed in the default ```./crashes```.   If so, I suggest using the ```FSDatedPermanentStorage``` class instead.  it's the same class, it just has a friendlier name.

twobraids:
aside from the suggestion on FS crash storage class, r+

rhelmer:
@twobraids changed the name of FS crash storage class and rebased, thanks!

PR #2681, opened 2015-03-19
fixes bug 1145374 - Ratelimit the search interface

peterbe:
@rhelmer @AdrianGaudebert r?

### DIFF
--- a/webapp-django/crashstats/crashstats/templates/crashstats/ratelimit_blocked.html
+++ b/webapp-django/crashstats/crashstats/templates/crashstats/ratelimit_blocked.html
@@ -1,0 +1,32 @@
+{% extends "crashstats_base.html" %}
+
+{% block page_title %}
+Ratelimit blocked
+{% endblock %}
+
+{% block product_nav_filter %}&nbsp;{% endblock %}
+
+{% block content %}
+<div id="mainbody">
+<div class="page-heading">
+    <h2>Hold your horses!</h2>
+    <h3>Ratelimit blocked</h3>
+</div>
+
+<div class="panel">
+    <div class="body">
+      <p>
+        You have been temporarily blocked due to excessive
+        use that we can't handle at the moment.
+      </p>
+      {% if not request.user.is_authenticated() %}
+      <p>
+        If you can, consider signing in.
+      </p>
+      {% endif %}
+    </div>
+</div>
+
+
+</div>
+{% endblock %}
--- a/webapp-django/crashstats/crashstats/views.py
+++ b/webapp-django/crashstats/crashstats/views.py
@@ -30,6 +30,30 @@ from crashstats.supersearch.models import SuperSearchUnredacted
 datetime.datetime.strptime('2013-07-15 10:00:00', '%Y-%m-%d %H:%M:%S')
 
 
+def ratelimit_blocked(request, exception):
+    # http://tools.ietf.org/html/rfc6585#page-3
+    status = 429
+
+    # If the request is an AJAX on, we return a plain short string.
+    # Also, if the request is coming from something like curl, it will
+    # send the header `Accept: */*`. But if you take the same URL and open
+    # it in the browser it'll look something like:
+    # `Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8`
+    if (
+        request.is_ajax() or
+        'text/html' not in request.META.get('HTTP_ACCEPT', '')
+    ):
+        # Return a super spartan message.
+        # We could also do something like `{"error": "Too Many Requests"}`
+        return http.HttpResponse(
+            'Too Many Requests',
+            status=status,
+            content_type='text/plain'
+        )
+
+    return render(request, 'crashstats/ratelimit_blocked.html', status=status)
+
+
 def robots_txt(request):
     return http.HttpResponse(
         'User-agent: *\n'
--- a/webapp-django/crashstats/settings/base.py
+++ b/webapp-django/crashstats/settings/base.py
@@ -82,6 +82,7 @@ MIDDLEWARE_CLASSES = (
     'commonware.middleware.FrameOptionsHeader',
 
     'waffle.middleware.WaffleMiddleware',
+    'ratelimit.middleware.RatelimitMiddleware',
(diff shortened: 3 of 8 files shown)


AdrianGaudebert:
This is failing: 
```
FAIL: test_search_ratelimited (crashstats.supersearch.tests.test_views.TestViews)
----------------------------------------------------------------------
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/mock.py", line 1224, in patched
    return func(*args, **keywargs)
  File "/home/travis/build/mozilla/socorro/webapp-django/crashstats/supersearch/tests/test_views.py", line 304, in test_search_ratelimited
    eq_(response.status_code, 429)
AssertionError: 200 != 429
```

webapp-django/crashstats/crashstats/views.py
```diff
@@ -31,6 +31,17 @@
 datetime.datetime.strptime('2013-07-15 10:00:00', '%Y-%m-%d %H:%M:%S')
 
 
+def ratelimit_blocked(request, exception):
+    # http://tools.ietf.org/html/rfc6585#page-3
+    status = 429
+    if request.path.startswith('/api/'):
+        # return a super spartan message
+        # we could also do something like `{"error": "Too Many Requests"}`
```
AdrianGaudebert: I've been wondering about this when building APIs recently and I think this is unnecessary. It's true that a string is not a valid JSON output, but do API consumers bother to parse the content when the status of the request is not a 200? I personally don't, so I would not bother making the output a JSON content.

webapp-django/crashstats/settings/base.py
```diff
@@ -346,6 +347,12 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
 # Rate limit for when using the Web API for anonymous hits
 API_RATE_LIMIT = '10/m'
 
+# Rate limit when using the supersearch web interface
+RATELIMIT_SUPERSEARCH = '5/m'
```
AdrianGaudebert: Why this super low number? I could see myself making more than 5 searches in a minute, if I'm having to refine a search with facets or columns for example. I think 10 would be a safer default here.

phrawzty: Perhaps this is out of scope, but would it be possible to set different limits based on whether the user is logged in or not?

webapp-django/crashstats/supersearch/views.py
```diff
@@ -97,6 +98,12 @@ def get_params(request):
...
 
 
+@ratelimit(
+    key='ip',
+    rate=utils.ratelimit_rate,
+    method=ratelimit.ALL,
+    block=True
+)
 @pass_default_context
 def search(request, default_context=None):
```
AdrianGaudebert: This is the wrong view to limit, it is only loaded the initial page but not the results. You want to limit ``search_results`` instead, as that's the view that's calling Elasticsearch.

webapp-django/crashstats/crashstats/templates/crashstats/ratelimit_blocked.html
```diff
@@ -0,0 +1,32 @@
+{% extends "crashstats_base.html" %}
```
AdrianGaudebert: If you're rate-limiting the ``search_results`` view, you will likely want to remove this. Otherwise, it will load the entire crash-stats UI inside a div in Super Search, leading to a very weird result. 

That seems a bit weird though, as that makes this view a lot less generic, so maybe you'll want to take a different approach to this?

webapp-django/crashstats/supersearch/views.py
AdrianGaudebert: To make this statement a bit clearer: 

loading ``/search/`` gets the page layout, all the JS code and stuff, and then it does an AJAX call to ``/search/results/``. That one actually calls the middleware and Elasticsearch. So, as is this patch, if someone where to use a crawler to load ``/search/`` many times, we would be protected. However, if that evil person calls ``/search/results/`` instead, it would have the same effect and we would not be protected. The same goes for someone that loads the UI in a browser and just spams the Search button.

webapp-django/crashstats/crashstats/views.py
peterbe: Yeah, it's confusing. I wonder what GitHub does. Whatever they do we ought to do :)

webapp-django/crashstats/settings/base.py
peterbe: Let's change to `10/m` then. 

Also, being signed in it raises the limit to infinite which is something rather weird because anybody can bother to sign in. It doesn't give them any permissions.

peterbe:
The test failure is annoying because it's not reproducible. Tests pass locally. There must be something with my `settings/local.py` that affects things. Hmm...

peterbe:
@AdrianGaudebert All addressed now. 
I now put the ratelimiter on the `search()` view **and** the `search_results()` view. 
The result is that if you're using the `search` view in the browser (which in itself fires off the `search_results` as AJAX) and you do this repeatedly you get a nice friendly full page HTML error that says "Hold your horses!"
However, if you take that AJAX URL, from the Web Console for example, and run it in curl 10+1 times you get a single `text/html` error that just says "Too Many Requests`. 

Also, suppose you take that AJAX URL and instead of running it in cron you run it in the browser, after 10+1 refreshes you get the full HTML error. 

This is clever because we're no longer dependent on that `if request.path.startswith('/api/')` stuff and the logic applies correctly for the API URLs too because them you can either run with curl or in the browser. 

In whichever case we always set the response HTTP error code to 429.

peterbe:
@AdrianGaudebert I'm changing the review? to you now.

webapp-django/crashstats/supersearch/tests/test_views.py
```diff
@@ -548,6 +571,62 @@ def mocked_get(url, params, **options):
...
+                'signature' in params and
+                '~nsASDOMWindowEnumerator' in params['signature']
+            ):
+                return Response({
+                    "hits": [],
+                    "facets": {},
+                    "total": 0
+                })
+            else:
+                return Response({"hits": [], "facets": [], "total": 0})
```
AdrianGaudebert: You can keep this line only, there's no need to have different blocks for products and stuff.

AdrianGaudebert:
r+

PR #2683, opened 2015-03-20
Fixes bug 1143724 - Used Super Search to power the Reports tab of report...

AdrianGaudebert:
.../list.

@peterbe r?

I realize this is lacking some comments, I'll add them next week (but feel free to complain about it anyway, as to make sure I do not forget ;) ).


AdrianGaudebert:
Now with a nice comment to explain what the new function does.

peterbe:
Great! I saw the followup bug you filed that is going to depend on this one. 

* The `old_names_map` stuff bothers me. Why is it needed? Either we ought to have a loud comment in the code saying "This is deprecated because bla bla bla" or something. Or why not just change the templates as part of this PR. After all, that's what the point is, isn't it, that the template says `{{ some_old_key }}`. 

* By the way, if you insist that we need the `old_names_map` and `new_names_map` business, you might want to just write 1 dict that has `{old_name: new_name}` instead of creating 2 dicts with the values in it.

AdrianGaudebert:
> The old_names_map stuff bothers me. Why is it needed? Either we ought to have a loud comment in the code saying "This is deprecated because bla bla bla" or something. Or why not just change the templates as part of this PR. After all, that's what the point is, isn't it, that the template says {{ some_old_key }}.

I am scared about missing something if I changed the templates. These views are quite complicated. By using this technique, I am pretty sure that nothing is being forgotten (because the Super Search Fields has all those names, we made sure of that when migrating the RawCrash white list). To me this is safer and less complicated to implement. (And also, since I made it a personal goal to remove ``/report/list``, you'll understand that I am reluctant to spend too much time working on it, hehe. )

> By the way, if you insist that we need the old_names_map and new_names_map business, you might want to just write 1 dict that has {old_name: new_name} instead of creating 2 dicts with the values in it.

How do I do the new_name -> old_name conversion using a dict like ``{old_name: new_name}``? (This is an honest question, to me the best way to convert in both directions is to have both associations and just use ``my_dict.get(key, key)``. )

peterbe:
What changed in the last commit? It appears you rebased it so I can't tell. 

Sorry. I don't mean to be a pain in the butt but it seems if we don't address this now (for the sake of finishing your Q goal) we'll never get around to it. Even if we file a bug. 

The thing with the ReportList is that it's quite complicated how it works. These are [the two crucial lines](https://github.com/mozilla/socorro/blob/master/webapp-django/crashstats/crashstats/templates/crashstats/partials/reports.html#L18-L19) and below are some exceptions like `{% elif column.key == 'os_and_version' %}`. And as far as the RawCrash fields, that's not changing with SuperSearch is it? In the old `ReportList().get()` it just joins the rawcrash into the list and then uses the `read_crash_column()` function to extract the column value out.

peterbe:
Are there other places where `ReportList().get()` gets used? I see [this for example](https://github.com/mozilla/socorro/blob/master/webapp-django/crashstats/crashstats/views.py#L1432).

I would rather we try to solve this here and now instead of a hack. If we get something wrong (e.g. data stops appearing) we can just solve that later which feels like something more likely to be done if we filed bugs.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -136,6 +139,79 @@ def build_data_object_for_crash_reports(response_items):
...
+    if 'version' in params:
+        versions = [
+            x.split(':')[-1] for x in params['version'] if x != 'ALL:ALL'
+        ]
+        params['version'] = versions
+
+    if 'process_type' in params and params['process_type'] == 'any':
+        del params['process_type']
+
+    if 'hang_type' in params and params['hang_type'] == 'any':
```
peterbe: Same as saying `if params.get('hang_type') == 'any':`

webapp-django/crashstats/crashstats/views.py
```diff
@@ -136,6 +139,79 @@ def build_data_object_for_crash_reports(response_items):
...
+        del params['hang_type']
```
AdrianGaudebert: In my last rebase, I added those 4 lines. It's removing the default ``any`` value for those 2 fields, because it's not understood by Super Search.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -136,6 +139,79 @@ def build_data_object_for_crash_reports(response_items):
...
+        ):
+            old_names_map[field['in_database_name']] = field['name']
+            new_names_map[field['name']] = field['in_database_name']
+
+    # Convert parameters to what Super Search accepts.
+    params = kwargs.copy()
+    if 'signature' in params:
+        params['signature'] = '=' + params['signature']
+
+    if 'end_date' in params or 'start_date' in params:
```
peterbe: here too. `params.get('end_date')`

webapp-django/crashstats/crashstats/views.py
peterbe: or
```python
for key in 'hang_type', 'process_type':
    if params.get(key) == 'any':
        del params[key]
```

AdrianGaudebert:
@peterbe I think I have addressed all of your concerns!

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/report_list_reports.js
```diff
@@ -111,7 +111,7 @@ Panels.register('reports', function() {
 
 var Columns = (function() {
     var COLUMNS = null;
-    var KEY = 'report-list-report-columns';
+    var KEY = 'report-list-reports-columns';
```
peterbe: clever!

webapp-django/crashstats/crashstats/views.py
```diff
@@ -136,6 +140,68 @@ def build_data_object_for_crash_reports(response_items):
...
 
+def get_report_list_from_super_search(**kwargs):
+    '''Return results as expected by the report_list views, but loaded
+    from Super Search.
+
+    This function converts given parameters to match what Super Search expects.
+    It changes some values (like version and _sort parameters). It then
+    queries Super Search, and maps the field names in the results to what
+    the report/list/ views and templates expect.
+    '''
```
peterbe: What's with these `'''` comments?! :)
you're the only one doing them. I guess it's your personal touch to say "adrian was here" :)

webapp-django/crashstats/crashstats/views.py
```diff
@@ -136,6 +140,68 @@ def build_data_object_for_crash_reports(response_items):
...
+
+    for key in ('process_type', 'hang_type'):
+        # For those keys, 'any' is not supported and should be an empty value.
+        if params.get(key) == 'any':
+            del params[key]
+
+    for param in params:
+        if param in old_names_map:
+            params[old_names_map[param]] = params[param]
+            params[param] = None
```
peterbe: why doesn't that do `del params[param]`? Perhaps I'm not understanding the code but basically this means that the key is not a known, SuperSearch field, right?
Perhaps what I'm saying is that this code deserves a human-friendly comment.

peterbe:
r+ with nit. Feel free to address that and merge freely.

webapp-django/crashstats/crashstats/views.py
AdrianGaudebert: Hah yeah sorry I like those better than ``"""``, especially since we use single quotes for everything else. But consistency first! I'll change that.

webapp-django/crashstats/crashstats/views.py
AdrianGaudebert: Adding a comment. The reason is that we're looping on ``params``, so it's not trivial to do that deletion. And making it ``None`` has the same effect.

PR #2686, opened 2015-03-23
Fixes nobug issue with migration drop ordering

selenamarie:
Fixes something that broke the stage refresh. Sorry! Production database has some things in it that our repo does not. :/


rhelmer:
r+

PR #2689, opened 2015-03-25
fixes Bug 1147519 - remove "Apache" from names

twobraids:
the Socorro code often links the words Apache and ModWSGI. Since modwsgi works under more than just Apache, the names that label modwsgi things ought not be saddled with the name Apache, too.


rhelmer:
@twobraids ```mod_wsgi``` is apache-specific too - there're other WSGI servers with different names (```uwsgi```, ```gunicorn```, etc) - I'd just go with ```WSGI```

twobraids:
@rhelmer  this one is for you!   r?

rhelmer:
r+ thanks this is great :)

PR #2692, opened 2015-03-27
bug 1148310 - always try to restart socorro apps

rhelmer:
r? @phrawzty - in the process of bootstrapping a consul cluster and connecting clients to it, I found that envconsul will generally retry until consul is working but will quit in some circumstances, which means the service will go into the failed state.


phrawzty:
We'll have to keep an eye on how this behaves generally speaking - for now, `r+` all the way.

PR #2694, opened 2015-03-27
No Bug - upgrade to latest configman

twobraids:
upgrade to the latest configman enabling several features useful to Socorro:

1) support for "--admin.print_conf=env"  will output a series of shell compatible environment variables for all the configman options
2) support round tripping without editing when creating "ini" files.  The output "ini" file will have things left as defaults commented out, thing that changed will not be commented out.
3) several bug fixes


twobraids:
@rhelmer   r?

rhelmer:
r+ wfm locally

PR #2696, opened 2015-03-30
Fixes bug 1149064 - Used Super Search to power the Comments tab of repor...

AdrianGaudebert:
...t/list/.

@peterbe r?


peterbe:
So, if we do this, it'll disable [the API endpoint](https://crash-stats.mozilla.com/api/#CommentsBySignature). I guess it makes sense because you can do the same with SuperSearch (if you speak black supersearchmagic fluently). Should there be some sort of deprecation. 
After all, the endpoint has been used [437 times](https://crash-stats.mozilla.com/admin/analyze-model-fetches/) since we switched on that analytics early last week.

What do you think @rhelmer and @lonnen ?

peterbe:
A more important problem is that I think we need to solve the problem of capping which fields we want to get back when doing SuperSearch queries. To get this:
screenshot 2015-03-30 15 40 25
It had to download a **1.0Mb** blob of JSON. All it needs is the comment, the date submitted and the email address. It should be a couple of bytes only. Not a million :)

By the way, I think we'd be 99% there if we just exlude the json_dump in the result. 

Because we can't reuse the cache can we when the parameter is different (e.g. `user_comments='!__null__'`)

peterbe:
This is basically depending on https://github.com/mozilla/socorro/pull/2710 right?

AdrianGaudebert:
Not technically, but yes, we might as well wait for it to be in as to avoid putting too much pressure on stage. :)

About the API endpoint, we have been removing a bunch of them recently. Even ``ReportList`` got removed yesterday, and it has seen about the same amount of usage as ``CommentsBySignature``. I guess this will be an excellent opportunity for us to show that ``SuperSearch`` can do it. :)

AdrianGaudebert:
@peterbe ready for review!

peterbe:
Clearly works!
screenshot 2015-04-02 07 54 27

PR #2695, opened 2015-03-27
fix bug 1148628 - run crontabber under envconsul if no /etc/socorro/cron...

rhelmer:
...tabber.ini

r? @phrawzty

PR #2698, opened 2015-03-30
Bug1148628 run crontabber under envconsul

rhelmer:
r? @phrawzty - same as PR #2695 but on a sensible branch name this time.


rhelmer:
Carrying over @phrawzty's r+ from PR #2695

PR #2699, opened 2015-03-30
fixes bug 1149196 - report_list  ValueError on install_time sometimes

peterbe:
@rhelmer r?

(cc @AdrianGaudebert but I think you're on PTO)


rhelmer:
r+, yeah looks like ES is storing this as a string the way the client sends it in.

peterbe:
Is my change reliable though. I'm also highly skeptical whenever I see any `mydatetime.strftime(...)` in python code. Converting dates to strings belongs in the view or the `json_view` decorator. Only.

rhelmer:
@peterbe honestly I think we should be parsing this on the way into ES, if that isn't possible for some reason we could push it to the model perhaps?

PR #2700, opened 2015-03-30
fix bug 1149236 - pull processed JSON from postgres, hbase is gone from ...

rhelmer:
...mozilla infra

r? anybody - long overdue to get this checked in properly.


rhelmer:
We need to land this and enable it in crontab, and disable the temporary one in socorro user's crontab on the prod admin box.

lonnen:
r+. once this lands we need to disable the socorro user's crontab or prod will get weird on us

PR #2701, opened 2015-03-30
fixes bug 1149239 - Unreliable JSON disk IO for models fetch cache

peterbe:
@rhelmer or @AdrianGaudebert r?

I've seen cases where the stored JSON file is corrupt. If it is we shouldn't block but just remove the file and not rely on that as a cache.

Also, I found that there was no test coverage for retrieving from the middleware file cache.

### DIFF
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -195,10 +195,23 @@ def memoize(function):
                     os.remove(cache_file)
                 else:
                     logger.debug("CACHE FILE HIT %s" % stringified_args)
-                    if instance.expect_json:
-                        return json.load(open(cache_file))
-                    else:
-                        return open(cache_file).read()
+                    delete_cache_file = False
+                    with open(cache_file) as f:
+                        if instance.expect_json:
+                            try:
+                                return json.load(f)
+                            except ValueError:
+                                logger.warn(
+                                    "%s is not a valid JSON file and will "
+                                    "be deleted" % (
+                                        cache_file,
+                                    )
+                                )
+                                delete_cache_file = True
+                        else:
+                            return f.read()
+                    if delete_cache_file:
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/crashstats/models.py
```diff
@@ -195,10 +195,23 @@ def get_cached_result(key, instance, stringified_args):
...
+                                    "%s is not a valid JSON file and will "
+                                    "be deleted" % (
+                                        cache_file,
+                                    )
+                                )
+                                delete_cache_file = True
+                        else:
+                            return f.read()
+                    if delete_cache_file:
+                        os.remove(cache_file)
```
AdrianGaudebert: I suppose you had to delay the removal of this file because it can't be done while it's open, right?

webapp-django/crashstats/crashstats/models.py
```diff
@@ -357,10 +370,24 @@ def fetch(
...
+                                            "will be deleted" % (
+                                                cache_file,
+                                            ),
+                                            exc_info=True
+                                        )
+                                        delete_cache_file = True
+                                else:
+                                    return f.read(), True
+                            if delete_cache_file:
+                                os.remove(cache_file)
```
AdrianGaudebert: I see a pattern here, how about putting that in a function? :-)

AdrianGaudebert:
r+ with a DRY nit. Also, that's an awesome test you added there!

webapp-django/crashstats/crashstats/models.py
peterbe: yup.

peterbe:
Thanks for the nit but I don't care enough because the second memoizer is going away in Q2 when @twobraids and I have re-written @selenamarie 's dataservice to an all new thing. E.g. dataservice version 2.0

PR #2702, opened 2015-03-30
fixes bug 1148587 - 0x221 is the wrong adapter hex

peterbe:
@selenamarie r? or if you're too busy getting started in your new team, @rhelmer r?

So, if you look at [Adapter Graphics Report](https://crash-stats.mozilla.com/report/list?signature=mozilla%3A%3Alayers%3A%3ADataTextureSourceD3D9%3A%3AUpdateFromTexture%28IDirect3DTexture9*%2C+nsIntRegion+const*%29&range_value=28&range_unit=days) in the Signature Summary you'll see that all the hexes for the adapters we don't recognized are padded as 4 characters. E.g. `0x0221`. And `'0x0221' != '0x221'` even though `hex(int('0x0221', 16)) == hex(int('0x221', 16))` is true. 

So that's why we miss so many adapters in the signature summary that starts with a leading 0. This PR fixes that.


selenamarie:
@peterbe r+!  Thank you for fixing this.

PR #2703, opened 2015-03-31
Fixes Bug 1149587 - enable crontabber to use socorro_app framework

twobraids:
when the socorro_app system was created, crontabber_app did not work with that system.  It derived from an earlier version of the app class system.  Since it split off from the Socorro, it took a copy of that earlier app system with it.  

this PR ought to bring crontabber into the modern app system fold...


peterbe:
Reviewer?

twobraids:
@peterbe, I will be asking you to review, but it isn't ready yet.  The error that Travis encounters is different than the error in my local install.

twobraids:
withdrawn until  there's more time to work on it

PR #2704, opened 2015-03-31
fixes bug 1149578 - install_time not always present in correlation signatures

peterbe:
@AdrianGaudebert r?


peterbe:
@rhelmer it's trivial and Adrian's asleep. If we get this in soon we might get less errors paging on stage.

PR #2706, opened 2015-03-31
Bug1149716 partitioning problems

selenamarie:
While reviewing our tables and the sizes of tables and partitions, I discovered we had many partitioned tables that were empty.

These tables haven't had any data put into them because we lacked `INSERT` triggers to handle the partitioning. 

The original partitioning strategy relied on our processors to use a date-specific table name on insert to get the data into the right place. This is fine, but it meant that we didn't have triggers set up as part of our typical partition setup. 

Because of this oversight, we have a few large tables that need to have data moved around to support our true partitioning. 

This patch addresses new incoming data. A separate bug will be filed to update our current data so that affected data is moved into appropriate partitioned tables.

This issue affects `missing_symbols` which has already grown to be a 50 GB table.


peterbe:
Selena says @rhelmer should be ideal reviewer :)

selenamarie:
@lars also said he'd have a look

twobraids:
I see that you've modified the insert logic so that the correct table partition is used for inserts rather than the parent table.  The mention of triggers in the initial problem statement leads me to believe that triggers were what you expected. I'm curious as to why you didn't implement a solution using triggers.

selenamarie:
@twobraids Because we bulk-insert and know which day we're dealing with ahead of time, it makes sense from an efficiency standpoint to insert into the table that we know the data is going to. 

In addition, my fears were:

* implementing something with a Postgres feature few people on the team were familiar with
* having the solution fail in a difficult to debug way, being too slow or causing locking problems

twobraids:
@selenamarie, that makes perfect sense, thank you for the answer

r+

PR #2707, opened 2015-04-01
do not accept crashes on Rahukaalam

rhelmer:
r? @lars @peterbe @AdrianGaudebert 

We should not be accepting crashes during times of Rahukaalam. This will require breakpad client support (/cc @luser), which should show a helpful error message to the user and retry submission later.

No tests, for proof denies faith. Also, once this feature is enabled, existing unit tests will fail during Rahukaalam - this is as it should be.

### DIFF
--- a/socorro/collector/wsgi_breakpad_collector.py
+++ b/socorro/collector/wsgi_breakpad_collector.py
@@ -2,6 +2,8 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import astral
+import datetime
 import web
 import time
 
@@ -46,6 +48,11 @@ class BreakpadCollector(RequiredConfig):
             'hash value',
         default='hashlib.md5',
         from_string_converter=class_converter
+    )
+    required_config.add_option(
+        'reject_crash_on_rahukaalam',
+        doc='reject crashes during Rahukaalam',
+        default=False
     )
 
     #--------------------------------------------------------------------------
@@ -116,6 +123,18 @@ class BreakpadCollector(RequiredConfig):
         if raw_crash.legacy_processing == IGNORE:
             self.logger.info('%s ignored', crash_id)
             return "Unsupported=1\n"
+        if self.config.collector.reject_crash_on_rahukaalam:
+            # only accept crash if not currently Rahukaalam
+            # http://en.wikipedia.org/wiki/Rahukaalam
+            a = astral.Astral()
+            latitude_mountain_view = '37.3894'
+            longitude_mountain_view = '122.0819'
+            now = datetime.datetime.utcnow()
+            try:
+                a.rahukaalam_utc(now, latitude_mountain_view, longitude_mountain_view)
+            except astral.AstralError:
+                # throw crash away, for now is an inauspicious time
+                return "Rahukaalam=1\n"
 
         self.config.crash_storage.save_raw_crash(
           raw_crash,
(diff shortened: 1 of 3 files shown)


phrawzty:
`s/rahukaalam/රාහු_කාලය/ig`

socorro/collector/wsgi_breakpad_collector.py
```diff
@@ -116,6 +123,18 @@ def POST(self, *args):
         if raw_crash.legacy_processing == IGNORE:
             self.logger.info('%s ignored', crash_id)
             return "Unsupported=1\n"
+        if self.config.collector.reject_crash_on_rahukaalam:
+            # only accept crash if not currently Rahukaalam
+            # http://en.wikipedia.org/wiki/Rahukaalam
+            a = astral.Astral()
+            latitude_mountain_view = '37.3894'
```
phrawzty: We'll need to determine the exact latitude and longitude of `us-west-2` in order to properly implement Rahukaalam checking.

Eventually we'll need to do the same for `us-east-1`, which should be easy enough.

twobraids:
While I commend the effort, I must say that this implementation is plain wrong.  There's way too much hard coding and not nearly enough abstraction. As you well know, once we've deployed Rahukaalam, we'll soon need Yamagandakaala and then Vishaghati.  Additionally, implementing this in the WSGIBreakpadCollector is not nearly as favorable as just creating a rule within the LegacyThrottler. 

I've already got a branch (likely written a couple years ago, so it may have suffered a some bit rot) that implements the ```InauspiciousThrottlerRule``` (implemented originally as a Ragnarök detector).  

```python
class InauspiciousThrottlerRule(tuple):
    required_config = Namespace()
    required_config.add_option(
        'crash_field',
        doc='the name of a field within a crash',
        default='*'
    )
    required_config.add_option(
        'inauspicious_function',
        doc='full python path to a method calculating determining an '
            'inauspicious moment',
        default='socorro.collector.ragnaroek.RagnaroekDetector',
    )
    required_config.add_option(
        'appliciation_percentage',
        doc="what percentage of crashes should be this rule apply to",
        default=100,
    )
    
    def __new__(cls, config):
        self.config = config
        return super(InauspiciousThrottlerRule, cls).__new__(
            cls, 
            (
                config.crash_field, 
                config.inauspicious_function,
                config.appliciation_percentage
            )
        )
```

With this class we can then specify in _configuration_ a throttler rule for any inauspicious moment:

```
[throttler]
    throttler_rules = [Rahukaalam, Yamagandakaala, Vishaghati, Ragnaroek]

    [[Rahukaalam]]
        crash_field=*
        inauspicious_function=socorro.collector.rahukaalam.is_.rahukaalam
        appliciation_percentage=100

    [[Yamagandakaala]]
        # ....
```

So, in all, I must say __r-__ to your implementation

socorro/unittest/collector/test_wsgi_breakpad_collector.py
```diff
@@ -36,6 +36,7 @@ def get_standard_config(self):
         config.collector.accept_submitted_crash_id = False
         config.collector.accept_submitted_legacy_processing = False
         config.collector.checksum_method = hashlib.md5
+        config.collector.reject_crash_on_rahukaalam = False
```
peterbe: This should be 
```python
config.collector.reject_crash_on_රාහු_කාලය = False
```
None of that medieval ascii bull-fecalmatter here please.

rhelmer:
Oops, turns out this is ancient astrology and not real life. My mistake!

phrawzty:
meow

PR #2708, opened 2015-04-01
Bug 427585 Change default table sort on topcrasher to descending

Tchanders:


AdrianGaudebert:
Thank you for contributing! Could you please create an account on bugzilla so we can assign you to the associated bug? 

r+

Tchanders:
I've made an account - my Bugzilla user name is also Tchanders. Thanks!

PR #2710, opened 2015-04-01
Fixes bug 1144569 - Return only a subset of columns in Super Search

AdrianGaudebert:
@peterbe r?


webapp-django/crashstats/crashstats/views.py
```diff
@@ -200,6 +200,12 @@ def get_report_list_from_super_search(**kwargs):
         params['_sort'] = '-' + params['_sort']
     del params['_reverse']
 
+    # Correct the 'os_and_version' column name.
+    if 'os_and_version' in params['_columns']:
+        cols = params['_columns']
+        cols[cols.index('os_and_version')] = 'platform'
```
peterbe: Only necessary if the order matters. You could just 
```python
cols.remove('os_and_version')
cols.extend(['platform', 'version'])
```

webapp-django/crashstats/crashstats/views.py
```diff
@@ -1440,6 +1446,14 @@ def report_list(request, partial=None, default_context=None):
                 include_raw_crash = True
                 break
 
+        api_columns = [x['key'] for x in context['columns']]
+        # The correlations tab requires some keys to always be there. Since
+        # they are low cost, we add them here, to keep the advantages of the
+        # shared cache between the two tabs.
```
peterbe: excellent comment!

peterbe:
Is the missing Date column due to this change? In my model-cache files I see the JSON files have a key called "date". 
screenshot 2015-04-01 11 31 50

peterbe:
r+ if you can fix the Date column :)

lonnen:
I decided to push to prod without landing this. We can do another push tomorrow morning after this lands. If the push is problematic I'll roll back prod.

PR #2712, opened 2015-04-02
Bug 1056984 Remove external link icons from "Signature report" links

Tchanders:
Also add titles to links and italicise the link in the "Signature facet" tab

### DIFF
--- a/webapp-django/crashstats/supersearch/templates/supersearch/search_results.html
+++ b/webapp-django/crashstats/supersearch/templates/supersearch/search_results.html
@@ -52,7 +52,7 @@
                         {% else %}
                             href="{{ url('crashstats:report_list') }}?signature={{ crash.signature | urlencode }}&amp;{{ report_list_query_string }}"
                         {% endif %}
-                            class="external-link"
+                            title="Click to leave this search and go to this signature report"
                         >
                             Signature report
                         </a>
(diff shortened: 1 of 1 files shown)


webapp-django/crashstats/supersearch/templates/supersearch/search_results.html
```diff
@@ -90,22 +90,24 @@
...
+                                    _no_base=True,
+                                    signature=hit.term,
+                                    page=None,
+                                    _columns=None,
+                                    _facets=None
+                                ) }}"
+                            {% else %}
+                                href="{{ url('crashstats:report_list') }}?signature={{ hit.term | urlencode }}&amp;{{ report_list_query_string }}"
+                            {% endif %}
+                                title="Click to leave this search and go to this individual report"
```
AdrianGaudebert: Please change ``individual`` to ``signature`` here.

AdrianGaudebert:
Could we use a different icon here in replacement of the "external" one? Something very light, just to show that those links will take you outside of the Search page. I'm open to suggestion, or to disagreement. :)

webapp-django/crashstats/supersearch/templates/supersearch/search_results.html
Tchanders: OK done

Tchanders:
Perhaps it would make more sense to add icons (e.g. a plus) to the left of the search terms, to indicate that they are not links, but filters?

AdrianGaudebert:
That could be a nice thing indeed! I have filed a bug for that, since it's a bit out of scope here: https://bugzilla.mozilla.org/show_bug.cgi?id=1152243 (feel free to work on it :) )

r+

PR #2715, opened 2015-04-03
fixes Bug 1150970 - added case where type is datetime rather than str

twobraids:
the MissingSymbolsRule in the processor can fail if the date_processed is an actual datetime rather than a string.  Because the crashstorage classes may stringify datetimes and that change propagates to the processed crash given to other crashstorage modules, the crashstorage modules must be defensive and should not assume types.

This PR fixes the utility method that creates a table partition suffix to accept either datetime instances or strings.

@selenamarie   r?


selenamarie:
@twobraids thank you! r+

PR #2716, opened 2015-04-03
also fixes Bug 1150970 - stopped MissingSymbolRule from spoiling its own SQL

twobraids:
the MissingSymbolsRule despoiled its own SQL by save the result after it formatted it.  That meant thet the second time around, there were no formatting variables in the string and formatting failed.  

fixed that problem and added tests to make sure that the rule works more than once.


twobraids:
@selenamarie r?

selenamarie:
:beetle: :zap: r+

PR #2719, opened 2015-04-06
fix bug 1151435 - provide way to override content-type based on file ext...

rhelmer:
...ension

r? @peterbe - I couldn't figure out a great way in the test to look at the boto ```key``` object, so I passed the content type through the symbol model (might be useful to record this info I guess?)

What do you think?

### DIFF
--- a/webapp-django/crashstats/symbols/models.py
+++ b/webapp-django/crashstats/symbols/models.py
@@ -9,6 +9,7 @@ class SymbolsUpload(models.Model):
     filename = models.CharField(max_length=100)
     size = models.IntegerField()
     created = models.DateTimeField(default=timezone.now)
+    content_type = models.TextField(null=True)
 
     def __repr__(self):
         return '<%s: %s (%d)...>' % (
--- a/webapp-django/crashstats/symbols/tests/test_views.py
+++ b/webapp-django/crashstats/symbols/tests/test_views.py
@@ -43,6 +43,7 @@ class TestViews(BaseTestViews):
         self.uploaded_keys = {}
         self.known_bucket_keys = {}
         self.created_buckets = []
+        self.created_keys = []
         mocked_connect_s3 = self.patcher.start()
 
         def mocked_get_bucket(*a, **k):
@@ -60,6 +61,7 @@ class TestViews(BaseTestViews):
                 mocked_key.set_contents_from_string.side_effect = mocked_set
                 mocked_key.key = key_name
                 mocked_key.bucket = mocked_bucket
+                self.created_keys.append(mocked_key)
                 return mocked_key
 
             def mocked_get_key(key_name):
@@ -67,6 +69,7 @@ class TestViews(BaseTestViews):
                 if key_name in self.known_bucket_keys:
                     mocked_key = mock.Mock()
                     mocked_key.key = key_name
+                    mocked_key.content_type = 'application/binary-octet-stream'
                     mocked_key.size = self.known_bucket_keys[key_name]
                     mocked_key.bucket = mocked_bucket
                     return mocked_key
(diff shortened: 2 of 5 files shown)


webapp-django/crashstats/symbols/models.py
```diff
@@ -9,6 +9,7 @@ class SymbolsUpload(models.Model):
     filename = models.CharField(max_length=100)
     size = models.IntegerField()
     created = models.DateTimeField(default=timezone.now)
+    content_type = models.TextField()
```
peterbe: Migration missing. You need to generate one with:
```
./manage.py schemamigration symbols --auto
```

If you do that you'll run into a problem that this is a new field but it doesn't have a default or allows nulls. 

I would recommend setting `content_type = models.TextField(null=True)`.

webapp-django/crashstats/symbols/tests/test_views.py
```diff
@@ -39,6 +39,8 @@ def setUp(self):
         super(TestViews, self).setUp()
         self.tmp_dir = tempfile.mkdtemp()
 
+        settings.SYMBOLS_MIME_OVERRIDES['jpeg'] = 'text/plain'
```
peterbe: This is not a good idea because you're potentially affecting the settingss as a global object across other tests. Instead use the context manager which resets the settings object automatically. E.g.
```python
with self.settings(SYMBOLS_MIME_OVERRIDES={'jpeg': 'text/plain'}):
    ...
```

peterbe: NB: this is something you'll have to do in the tests themselves. Not in the `setUp`.

webapp-django/crashstats/symbols/tests/test_views.py
```diff
@@ -67,6 +69,7 @@ def mocked_get_key(key_name):
                 if key_name in self.known_bucket_keys:
                     mocked_key = mock.Mock()
                     mocked_key.key = key_name
+                    mocked_key.content_type = 'application/binary-octet-stream'
```
peterbe: Now I don't understand. This is not testing the `key.content_type = settings.SYMBOLS_MIME_OVERRIDES[ext]` line, is it?

rhelmer: > Now I don't understand. This is not testing the `key.content_type = settings.SYMBOLS_MIME_OVERRIDES[ext]` line, is it?

Well, it is setting a default which we are overriding with ```text/plain``` in the settings and testing that the override worked later... can you think of a better way to do this?

webapp-django/crashstats/symbols/views.py
```diff
@@ -95,6 +95,12 @@ def unpack_and_upload(iterator, symbols_upload, bucket_name, bucket_location):
 
             file = StringIO()
             file.write(member.extractor().read())
+
+            for ext in settings.SYMBOLS_MIME_OVERRIDES:
+                if key_name.endswith(ext):
```
peterbe: Should we use `os.path.splitext()` instead? e.g. if one of the keys in `SYMBOLS_MIME_OVERRIDES` is 'ar' this'd be true if the file is `foo.bar` and `foo.tar`.

rhelmer: > Should we use os.path.splitext() instead? e.g. if one of the keys in SYMBOLS_MIME_OVERRIDES is 'ar' this'd be true if the file is foo.bar and foo.tar.

Well we do ```endswith('.tar.gz')``` elsewhere in this file:
https://github.com/mozilla/socorro/blob/master/webapp-django/crashstats/symbols/views.py#L154

So if we change it in one place we should probably do all :)

webapp-django/crashstats/symbols/tests/test_views.py
rhelmer: I could make the default ```None``` instead, maybe would be less confusing?

peterbe: I don't know anymore. I tried to make sense of it during lunch but I can't find any place where it even tests what gets set in the mocked objects. I saw, just leave it as is. Don't let this comment block.

webapp-django/crashstats/symbols/views.py
peterbe: So, we can simply change to `if key_name.endswith('.%s' % ext):` ?

webapp-django/crashstats/symbols/tests/test_views.py
rhelmer: (for the benefit of anyone reading along, @peterbe figured out a good way to do this and I merged a PR in, github is not smart enough to collapse this though)

peterbe:
r+ if travis is OK.

PR #2720, opened 2015-04-06
fixes bug 1151455 - signature summary constraints on report_date

peterbe:
@rhelmer r?

I'm confident this fixes it. See https://gist.github.com/peterbe/d09cd6fd7765642453ea

### DIFF
--- a/socorro/external/postgresql/signature_summary.py
+++ b/socorro/external/postgresql/signature_summary.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import datetime
 import logging
 
 from socorro.external.postgresql.base import PostgreSQLBase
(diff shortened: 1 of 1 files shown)


rhelmer:
r+ - I am pretty sure this was changed by commit 921139ca566dabc31b419589474746aa9681f341

socorro/external/postgresql/signature_summary.py
```diff
@@ -109,6 +110,9 @@ def get(self, **kwargs):
         params['version'] = tuple(versions)
 
         all_results = {}
+        assert isinstance(params['start_date'], datetime.date)
+        assert isinstance(params['end_date'], datetime.date)
```
peterbe: The reason for this added verbosity is to future foot-gun-proof us. Later this month @twobraids and I are planning to rewrite large parts of the implementation classes to use a new annotation system. In particular so that the webapp can scan and auto-generate the API documentation. Then we're going to get rid of the `filters = [` lists found in places like this. And since a `datetime.datetime` is "convertable" to `datetime.date` we must let this happen on the psycopg level because it goes horribly wrong. 

So by putting these asserts in we can assure that we don't accidentally allow any `datetime.datetime` instances in as parameters.

PR #2721, opened 2015-04-06
 Fixes bug 1137879 add support for postgres URLs - alternate

twobraids:
NO MERGE - this is an experiment

this is an alternate for selena's PG URL implementation.  I think it is simpler.


twobraids:
withdrawn pending further experimentation and tests

PR #2722, opened 2015-04-07
BugFix 1151754

dehachoudhry:
While building Socorro from source (github master); 
`git clone https://github.com/mozilla/socorro`
`cd socorro`
`make dev`

error in files dumplookup.cc, get-minidump-instructions.cc & stackwalker.cc in /minidump-stackwalk

```stackwalker.cc: In function ‘std::string {anonymous}::ToHex(uint64_t)’:
stackwalker.cc:117:33: error: format ‘%lx’ expects argument of type ‘long unsigned int’, but argument 3 has type ‘uint64_t {aka long long unsigned int}’ [-Werror=format=]
   sprintf(buffer, "0x%lx", value);```

`all %lx / %ld should be %llx / %lld`
https://bugzilla.mozilla.org/show_bug.cgi?id=1151754

### DIFF
--- a/minidump-stackwalk/dumplookup.cc
+++ b/minidump-stackwalk/dumplookup.cc
@@ -70,19 +70,19 @@ void print_frame(const StackFrame* frame) {
       printf("!%s", frame->function_name.c_str());
       if (!frame->source_file_name.empty()) {
         string source_file = PathnameStripper::File(frame->source_file_name);
-        printf(" [%s : %d + 0x%lx]",
+        printf(" [%s : %d + 0x%llx]",
                source_file.c_str(),
                frame->source_line,
                frame->instruction - frame->source_line_base);
       } else {
-        printf(" + 0x%lx", frame->instruction - frame->function_base);
+        printf(" + 0x%llx", frame->instruction - frame->function_base);
       }
     } else {
-      printf(" + 0x%lx",
+      printf(" + 0x%llx",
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by luser** [comment]
Unfortunately this isn't quite right. Right now the code builds without warnings when doing a 64-bit build, but prints warnings (that get turned into errors by -Werror) when doing a 32-bit build. Your patch would fix the warnings on a 32-bit build but they would warn on a 64-bit build. If I build this test program on a 64-bit system:
```c
#include <stdio.h>
#include <stdint.h>

int main()
{
  uint64_t i = 1;
  printf("%llx\n", i);
  return 0;
}
```

I get:
```
/tmp/test.c:7:3: warning: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘uint64_t’ [-Wformat=]
   printf("%llx\n", i);
   ^
```

The right fix here is to use the format macros from inttypes.h, like so:
```c
#include <stdio.h>
#define __STDC_FORMAT_MACROS
#include <inttypes.h>
#include <stdint.h>

int main()
{
  uint64_t i = 1;
  printf("%" PRIx64 "\n", i);
  return 0;
}
```

It's ugly but it's the right fix. If you can update your patch to use that we can land it. Thanks!

AdrianGaudebert:
Closing this for now. Please reopen when problems are addressed.

Thanks for contributing! :)

PR #2723, opened 2015-04-07
Fixes bug 1137879 add support for postgres URLs

twobraids:
This PR enables the PG ConnectionContext classes to use a PG URL existing as an environment variable ```database_url``` to prime the configuration defaults.  For example if the environment variable is: ```postgres://test:aPassword@localhost:5432/socorro_migration_test```, then the class ```socorro.external.postgresql.connection_context.ConnectionContext``` will have the following defaults automatically:
``` python
database_hostname=localhost
database_username=test
database_password=aPassword
database_port=5432
database_name=socorro_migration_test
```
 It also facilitates removing the old ```socorro.database.Database``` class from the middleware.  The middleware integration tests are now also free of the dreaded ```.../unittest/confg/commonconfig.py```.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -8,23 +8,28 @@ JENKINS_CONF=jenkins.py.dist
 
 ENV=env
 
-PYTHONPATH="."
+PYTHONPATH=.
 
 PG_RESOURCES=""
-if [ -n "$database_hostname" ]; then
-    PG_RESOURCES="$PG_RESOURCES resource.postgresql.database_hostname=$database_hostname"
-fi
-if [ -n "$database_username" ]; then
-    PG_RESOURCES="$PG_RESOURCES secrets.postgresql.database_username=$database_username"
-fi
-if [ -n "$database_password" ]; then
-    PG_RESOURCES="$PG_RESOURCES secrets.postgresql.database_password=$database_password"
-fi
-if [ -n "$database_port" ]; then
-    PG_RESOURCES="$PG_RESOURCES resource.postgresql.database_port=$database_port"
-fi
-if [ -n "$database_name" ]; then
-    PG_RESOURCES="$PG_RESOURCES resource.postgresql.database_name=$database_name"
+if [ -n "$database_url" ]; then
+    echo database_url is present, specifying parameters on the command line is not necessary
+else
+    # This clause is all legacy and can be removed once we switch to use database_url in config
+    if [ -n "$database_hostname" ]; then
+        PG_RESOURCES="$PG_RESOURCES resource.postgresql.database_hostname=$database_hostname"
+    fi
+    if [ -n "$database_username" ]; then
+        PG_RESOURCES="$PG_RESOURCES secrets.postgresql.database_username=$database_username"
+    fi
+    if [ -n "$database_password" ]; then
+        PG_RESOURCES="$PG_RESOURCES secrets.postgresql.database_password=$database_password"
+    fi
+    if [ -n "$database_port" ]; then
+        PG_RESOURCES="$PG_RESOURCES resource.postgresql.database_port=$database_port"
+    fi
+    if [ -n "$database_name" ]; then
+        PG_RESOURCES="$PG_RESOURCES resource.postgresql.database_name=$database_name"
+    fi
 fi
 
 RMQ_RESOURCES=""
@@ -84,11 +89,11 @@ for file in *.py.dist; do
 done
 popd
 
-PYTHONPATH=$PYTHONPATH $SETUPDB --database_name=socorro_integration_test --database_username=$database_username --database_hostname=$database_hostname --database_password=$database_password --database_port=$database_port --database_superusername=$database_superusername --database_superuserpassword=$database_superuserpassword --dropdb --logging.stderr_error_logging_level=40 --unlogged --no_staticdata
+PYTHONPATH=$PYTHONPATH $SETUPDB --database_name=socorro_integration_test --dropdb --logging.stderr_error_logging_level=40 --unlogged --no_staticdata
 
-PYTHONPATH=$PYTHONPATH $SETUPDB --database_name=socorro_test --database_username=$database_username --database_hostname=$database_hostname --database_password=$database_password --database_port=$database_port --database_superusername=$database_superusername --database_superuserpassword=$database_superuserpassword --dropdb --no_schema --logging.stderr_error_logging_level=40 --unlogged --no_staticdata
+PYTHONPATH=$PYTHONPATH $SETUPDB --database_name=socorro_test --dropdb --no_schema --logging.stderr_error_logging_level=40 --unlogged --no_staticdata
 
-PYTHONPATH=$PYTHONPATH $SETUPDB --database_name=socorro_migration_test --database_username=$database_username --database_hostname=$database_hostname --database_password=$database_password --database_port=$database_port --database_superusername=$database_superusername --database_superuserpassword=$database_superuserpassword --dropdb --logging.stderr_error_logging_level=40 --unlogged
+PYTHONPATH=$PYTHONPATH $SETUPDB --database_name=socorro_migration_test --dropdb --logging.stderr_error_logging_level=40 --unlogged
 
 PYTHONPATH=$PYTHONPATH ${VIRTUAL_ENV}/bin/alembic -c config/alembic.ini downgrade -1
 PYTHONPATH=$PYTHONPATH ${VIRTUAL_ENV}/bin/alembic -c config/alembic.ini upgrade heads
--- a/socorro/database/database.py
+++ b/socorro/database/database.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import urlparse
 
 import psycopg2
 import psycopg2.extensions
--- a/socorro/external/postgresql/base.py
+++ b/socorro/external/postgresql/base.py
@@ -9,11 +9,12 @@ import logging
 
 import psycopg2
 
-import socorro.database.database as db
 from socorro.external import DatabaseError
 
-from .dbapi2_util import execute_query_fetchall, single_value_sql
-
+from socorro.external.postgresql.dbapi2_util import (
+    execute_query_fetchall,
+    single_value_sql
+)
 logger = logging.getLogger("webapi")
 
 
@@ -41,22 +42,17 @@ class PostgreSQLBase(object):
 
         """
         self.context = kwargs.get("config")
-        if hasattr(self.context, 'database'):
-            # XXX this should be replaced with connection_context instead
-            self.context.database['database_host'] = \
-                self.context.database.database_hostname
-            self.context.database['database_port'] = \
-                self.context.database.database_port
-            self.context.database['database_name'] = \
-                self.context.database.database_name
-            self.context.database['database_username'] = \
-                self.context.database.database_username
-            self.context.database['database_password'] = \
-                self.context.database.database_password
-            self.database = db.Database(self.context.database)
-        else:
-            # the old middleware
-            self.database = db.Database(self.context)
+        try:
+            self.database =  self.context.database_class(
+                self.context
+            )
+        except KeyError:
+            # some tests seem to put the database config parameters
+            # into a namespace called 'database', others do not
+            self.database =  self.context.database.database_class(
+                self.context.database
+            )
+
 
     @contextlib.contextmanager
     def get_connection(self):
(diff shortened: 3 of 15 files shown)


socorro/unittest/external/postgresql/unittestbase.py
```diff
@@ -119,7 +98,8 @@ def setUpClass(cls):
         case (aka. test class).
         """
         cls.config = cls.get_standard_config()
-        cls.database = db.Database(cls.config)
+        # cls.database = db.Database(cls.config)
```
twobraids: remove

twobraids:
@peterbe r?

scripts/test.sh
```diff
@@ -8,23 +8,30 @@ JENKINS_CONF=jenkins.py.dist
...
-if [ -n "$database_password" ]; then
-    PG_RESOURCES="$PG_RESOURCES secrets.postgresql.database_password=$database_password"
-fi
-if [ -n "$database_port" ]; then
-    PG_RESOURCES="$PG_RESOURCES resource.postgresql.database_port=$database_port"
-fi
-if [ -n "$database_name" ]; then
-    PG_RESOURCES="$PG_RESOURCES resource.postgresql.database_name=$database_name"
+if [ -n "$database_url" ]; then
+    echo database_url supercedes database environment variables for nosetests
```
twobraids: these two comments are no longer accurate - remove

socorro/database/database.py
```diff
@@ -154,6 +155,14 @@ class Database(object):
   #-----------------------------------------------------------------------------------------------------------------
   def __init__(self, config, logger=None):
     super(Database, self).__init__()
+    if config.get('database_url'):
```
twobraids: superfluous - remove

socorro/external/postgresql/setupdb_app.py
```diff
@@ -50,47 +52,52 @@ class SocorroDBApp(App):
 
     required_config = Namespace()
 
+    # Specifying database_url overrides all previously set
```
twobraids: comment no longer accurate - remove

twobraids:
woot!  new feature with a net reduction in the number of lines of code!

scripts/test.sh
peterbe: +1

socorro/external/postgresql/base.py
```diff
@@ -41,22 +42,17 @@ def __init__(self, *args, **kwargs):
...
+            self.database =  self.context.database_class(
+                self.context
+            )
+        except KeyError:
+            # some tests seem to put the database config parameters
+            # into a namespace called 'database', others do not
+            self.database =  self.context.database.database_class(
+                self.context.database
+            )
+
```
peterbe: Good riddance!

socorro/external/postgresql/base.py
```diff
@@ -41,22 +42,17 @@ def __init__(self, *args, **kwargs):
...
-            # the old middleware
-            self.database = db.Database(self.context)
+        try:
+            self.database =  self.context.database_class(
+                self.context
+            )
+        except KeyError:
+            # some tests seem to put the database config parameters
+            # into a namespace called 'database', others do not
+            self.database =  self.context.database.database_class(
```
peterbe: Deprecation warning? 
Or perhaps just fix the tests that goes into this except block once and for all?

socorro/external/postgresql/base.py
```diff
@@ -87,11 +83,13 @@ def query(self, sql, params=None, error_message=None, connection=None):
                 fresh_connection = True
             # self.context.logger.debug(connection.cursor.mogrify(sql, params))
             results = execute_query_fetchall(connection, sql, params)
+            connection.commit()
```
peterbe: Isn't it the job of the transactionexecutor to worry about transactions? I forget. It's a mess in my head.

socorro/external/postgresql/connection_context.py
```diff
@@ -2,15 +2,32 @@
...
-from configman.config_manager import RequiredConfig
-from configman import Namespace
+from configman import RequiredConfig, Namespace
+from configman.memoize import memoize
 
+#------------------------------------------------------------------------------
+# we're going to end up parsing the same url several times, there's no need to
+# actually do it more than once, so we wrap the method with a memoizer to speed
+# things along.
+@memoize()
```
peterbe: That'll save hundreds of nanoseconds. :)

socorro/external/postgresql/base.py
twobraids: they are tests for the middleware, all of this code goes away when the middleware does, so I'm not very interested in spending time correcting its tests.

socorro/external/postgresql/connection_context.py
```diff
@@ -2,15 +2,32 @@
...
+def parse_url(*args):
+    return urlparse(*args)
 
+#------------------------------------------------------------------------------
+def default_from_url(field, default):
+    if 'database_url' in os.environ:
```
peterbe: This should be `if os.environ.get('database_url'):` otherwise the next line might fail. 
A user might think he's undoing a setting by executing `export database_url=`

socorro/external/postgresql/setupdb_app.py
```diff
@@ -233,9 +212,35 @@ def generate_fakedata(self, db, fakedata_days):
...
+            database_username = self.config.get('database_superusername')
+            database_password = self.config.get('database_superuserpassword')
+
+        # construct a URL
+        url = 'postgresql://'
+        if database_username:
+            url += '%s' % database_username
+        if database_password:
+            url += ':%s' % database_password
+        url += '@'
```
peterbe: shouldn't this be:
```python
if database_username or database_password:
    url += '@'
```

socorro/external/postgresql/setupdb_app.py
```diff
@@ -233,9 +212,35 @@ def generate_fakedata(self, db, fakedata_days):
...
+        if superuser:
+            database_username = self.config.get('database_superusername')
+            database_password = self.config.get('database_superuserpassword')
+
+        # construct a URL
+        url = 'postgresql://'
+        if database_username:
+            url += '%s' % database_username
+        if database_password:
+            url += ':%s' % database_password
```
peterbe: Do these two lines need to be indented under the `if database_username:`?

peterbe:
A couple of nits and questionmarks but if it works, it works. r+

socorro/external/postgresql/base.py
twobraids: while I converted the middleware to use the current PG ConnectionContext class to make configuration easier, converting the middleware to use the TransactionExecutor was going to be a big job.  Rather than spend the time to do the conversion on code that is already doomed, I just ensured that transactions are at least properly closed.  This is something that the original code should have been doing anyway.

socorro/external/postgresql/connection_context.py
twobraids: I did that because you complained about rerunning the parser several times in selena's original PR.  I figured you know something about the parsers expense that I did not, I made sure it only ran once.

socorro/external/postgresql/setupdb_app.py
twobraids: this is a code block lifted directly from selena's original PR, I'd didn't even examine it.  I'll take a closer look.

socorro/external/postgresql/connection_context.py
peterbe: That was not the point I made. The point I made was about DRY violating code. E.g. the 
```
parts = urlparse(url)
database_host = parts.hostname
database_name = parts.path[1:]
...
```

peterbe: Rewriting the standard lib `urlparse` just to put a memoize on it gains very little in my (stringent) opinion. It's a I/O-less operation that takes virtually no time at all and swapping `parse_url` for `urlparse` isn't winning us much. It just adds moving parts and future-developer-questionmarks. :)

Strangely, now that I look at your rewrite from Selena I don't see multiple places where it converts the URL into a dict (or DotDict like thing). What I wanted was to only see the `urlparse.urlparse` used in one single place in a neat little function in some `util.py` somewhere.

socorro/external/postgresql/connection_context.py
peterbe: Why is only ever coming from the env? What if the database_url comes from an .ini file?
I like env variables and I'm sure it'll work but what if you put the database_url in the .ini or as a argument (e.g. `--database_url=postgres://...`)

socorro/external/postgresql/connection_context.py
peterbe: I sort of see what's going on now. There are two instances of it being "used" repeatedly. 

My original complaint (I most likely mishandled my comments) was about the DRY violation of using `urlparse.urlparse()` logic spread in multiple places with miniscule difference. 

Now I see what you're doing with the `required_config.add_option()` as class attributeS repeatedly. That I don't mind being repeated. The "re-parsing" overhead is negligible. 

I think what I don't like about all of these `default_from_url()` is that when my eyes are on them I don't know what this "url" is. The `default_from_url` isn't a "tight" function. It depends on something globally outside of its internal scope. 

You know me, I like bikeshedding and high quality code that is dumb and easy to understand. 

Solution 1)

```python
from utils import parse_database_url

db_defaults = {
    'hostname': 'localhost', 
    'name': 'breakpad', 
    ...
}
if os.environ.get('database_url'):
    # oh! much better defaults are available
    db_defaults.update(parse_database_url(os.environ['database_url')))

class ConnectionContext(...):
    required_config = Namespace()
    required_config.add_option(
        name='database_hostname',
        default=db_defaults['hostname'],
        doc='the hostname of the database',
        reference_value_from='resource.postgresql'
    )
    ...etc....
```

Solution 2)

Get rid of ALL explosion of database credentials everywhere. ONLY allow for database URLs to configure things ever. E.g. disallow the use of relying on all other defaults but deliberately overriding the password with `setupdb_app.py --database_password=imspecial`.

I'm guessing this is harder to do due to legacy but you will know much better than I. 

Solution 3)

Ignore everything I say because things will be replaced in a month or two anyway. :)

twobraids: Solution 1 would be functionally equivalent to what I have now.  So it is a matter of style.  I prefer my method because the action of determining the default to use is encapsulated in a function that could be easily replaced.  I'm amused by your objection to the method not being _tight_ - if _tight_ means that it cannot use global state, then a _tight_ function could never use the environment.  An alternative method that would allow environment to be passed, as in dependency injection, in would be needless complication to the function call.  The only reason that your code proposal doesn't suffer the same criticism is the fact that it isn't encapsulated into a function.  

Solution 2 - I have several objections to that method:

I implemented such a thing on Monday as a separate PG ConnectionContext class with the idea that through config you could choose to use either class depending on how you wanted to set the PG credentials.  That method however, did nothing to address the problem that the environment variable ```database_url``` does not map in configman to our database credentials (```resource.postgresql.database_url```)  So additional measures similar to what I did in my code here would still have to be done to get the environment variable's value into the right configman option.  

Then, in order to allow for overriding certain fields within the URL, I'd still have to have the configman options for each of the component parts.  At that point, I have to establish a overriding heirarchy between the __url__ and the component parts.  If the user were to override on the command line, the same value that is the default for ```database_name```, do I override it in the __url__?  How do I detect that the default from the user specified value.  This was flaw in the original PR as it did not establish a predictable overriding scheme.  In ```test.sh``` I found a place where the __url__ overrode the component parts default, but on the next line the opposite was true, the component parts overrode the __url__.  It was too confusing, so I opted for a method that was perfectly predictable.  

My final argument against that is the legacy configuration and code.  I wanted to minimize the size of the job both in the code in the PR and later during deployments.

Solution 3 - is what I believe I have implemented...

socorro/external/postgresql/connection_context.py
twobraids: see my objections to Solution 2 above.

socorro/external/postgresql/connection_context.py
peterbe: I read your comment on Solution 2. Twice. I still don't understand it but I understand that you know this domain much better than me. Legacy is definitely at play. And transition from legacy to modern is the crux. 

My dream is to have it like you do in sugardough (the mozilla preferred way of doing modern django set ups) in that the one and only way to set credentials is with a URL. 

What would it take to get rid of ALL "database_something" (where "something" is **not** "url")?

twobraids: I wrote this on Monday and rejected it:  https://gist.github.com/twobraids/83be93e6eef41e8fda7c
because it would be a huge chore to migrate.

socorro/external/postgresql/setupdb_app.py
twobraids: I hadn't noticed that ```setupdb_app.py``` effectively has no tests.  In addition, the file that does much of the work, ```postgresqlalchemymanager.py```, doesn't even have a test file.

PR #2725, opened 2015-04-09
Fixes bug 1000160 - Set the correct version number in processed crashes.

AdrianGaudebert:
@twobraids Does this look good to you? 

Note that this is not ready to be merged yet. I need to figure out a transition plan, and to modify a bunch of code related to those version numbers.

### DIFF
--- a/socorro/external/postgresql/base.py
+++ b/socorro/external/postgresql/base.py
@@ -113,7 +113,7 @@ class PostgreSQLBase(object):
             if not connection:
                 connection = self.database.connection()
                 fresh_connection = True
-            # self.context.logger.debug(connection.cursor.mogrify(sql, params))
+            # logger.debug(connection.cursor().mogrify(sql, params))
             result = actor_function(connection, sql, params)
             connection.commit()
         except psycopg2.Error, e:
@@ -282,6 +282,23 @@ class PostgreSQLBase(object):
                         )
                     )
                     version_index += 2
+
+                # Bug 1000160. The logic to convert version numbers to what
+                # they should actually be has been moved to the processors.
+                # However, there will be a period of time when we will have
+                # both the "correct" version number and the "wrong" one in our
+                # database. We thus need to query for both here.
+                versions_where.append(
+                    PostgreSQLBase.build_version_where(
+                        product,
+                        version,
+                        version_index,
+                        sql_params,
+                        None,
+                        config
+                    )
+                )
+                version_index += 2
 
             if versions_where:
                 sql_where.append("(%s)" % " OR ".join(versions_where))
--- a/socorro/external/postgresql/raw_sql/procs/001_update_reports_clean.sql
+++ b/socorro/external/postgresql/raw_sql/procs/001_update_reports_clean.sql
@@ -241,7 +241,7 @@ SET product_version_id = product_versions.product_version_id
 FROM product_versions, new_reports
 WHERE reports_clean_buffer.uuid = new_reports.uuid
     AND new_reports.product = product_versions.product_name
-    AND new_reports.version = product_versions.release_version
+    AND new_reports.version = product_versions.version_string
     AND reports_clean_buffer.release_channel = product_versions.build_type
     AND reports_clean_buffer.release_channel <> 'beta';
 
@@ -253,7 +253,7 @@ SET product_version_id = product_versions.product_version_id
 FROM product_versions JOIN product_version_builds USING (product_version_id), new_reports
 WHERE reports_clean_buffer.uuid = new_reports.uuid
     AND new_reports.product = product_versions.product_name
-    AND new_reports.version = product_versions.release_version
+    AND new_reports.version = product_versions.version_string
     AND reports_clean_buffer.release_channel = product_versions.build_type
     AND reports_clean_buffer.build = product_version_builds.build_id
     AND reports_clean_buffer.release_channel = 'beta';
--- a/socorro/processor/mozilla_transform_rules.py
+++ b/socorro/processor/mozilla_transform_rules.py
@@ -816,3 +816,109 @@ class MissingSymbolsRule(Rule):
         except KeyError:
             return False
         return True
+
+
+#==============================================================================
+class BetaVersionRule(Rule):
+    required_config = Namespace()
+    required_config.add_option(
+        'database_class',
+        doc="the class of the database",
+        default=
+        'socorro.external.postgresql.connection_context.ConnectionContext',
+        from_string_converter=str_to_python_object,
+        reference_value_from='resource.postgresql',
+    )
+    required_config.add_option(
+        'transaction_executor_class',
+        default="socorro.database.transaction_executor."
+                "TransactionExecutorWithInfiniteBackoff",
+        doc='a class that will manage transactions',
+        from_string_converter=str_to_python_object,
+        reference_value_from='resource.postgresql',
+    )
+
+    #--------------------------------------------------------------------------
+    def __init__(self, config):
+        super(BetaVersionRule, self).__init__(config)
+        database = config.database_class(config)
+        self.transaction = config.transaction_executor_class(
+            config,
+            database,
+        )
+        self._versions_data_cache = {}
+
+    #--------------------------------------------------------------------------
+    def version(self):
+        return '1.0'
+
+    #--------------------------------------------------------------------------
+    def _get_version_data(self, product, version, release_channel, build_id):
+        key = '%s:%s:%s:%s' % (product, version, release_channel, build_id)
+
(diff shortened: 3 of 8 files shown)


socorro/processor/mozilla_transform_rules.py
```diff
@@ -816,3 +816,79 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
+                processed_crash['version'],
+                processed_crash['release_channel'],
+                processed_crash['build'],
+            )
+            real_version = self.versions_data.get(key)
+
+            if real_version:
+                processed_crash['version'] = real_version
+            elif processed_crash['release_channel'].lower() == 'beta':
+                # This is a beta version but we do not have data about it. It
```
twobraids: you might want to add a note to the ```processor_meta.processor_notes``` (list of strings) to say express what happened here.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -816,3 +816,79 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
+                # could be because we don't have it yet (if the cron jobs are
+                # running late for example), so we mark this crash. This way,
+                # we can reprocess it later to give it the correct version.
+                processed_crash['version'] += 'b0'
```
twobraids: if this crash is reprocessed for some reason other than this _unknown beta_ situation and it is still an unknown beta, then this line will tag an additional 'b0' to the end of the version.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -816,3 +816,79 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
+                pvb.build_id,
+                pv.version_string
+            FROM product_versions pv
+                LEFT JOIN product_version_builds pvb ON
+                    (pv.product_version_id = pvb.product_version_id)
+        """
+        results = transaction(
+            execute_query_fetchall,
+            sql
+        )
```
twobraids: how frequently does this table change?  This class is instantiated once during processor startup so the content of this lookup table will persist unchanged until the SIGHUP or restart of the processor.

rhelmer: This table is refreshed daily by crontabber ```ProductVersionsCronApp```. It may also be updated more frequently by the admin UI.

twobraids: Is there another way to detect if a crash is from a beta version?  If we could detect that it is a beta version in the rule predicate then we could  do the lookup dynamically in the body of the action.  We could do a time based cache on the results to limit the number of hits on PG, but still allow it to be dynamic enough to catch changes to that table.

rhelmer: We can tell if it's from *a* beta version based on the channel, we just don't have any way tell tell exactly *which* beta version this is unless we look in ```product_versions```.

AdrianGaudebert: I'm not even sure this logic is needed for beta only. At the moment, we are doing this transform logic for all non-release channels in our middleware. But maybe "no release" is a good enough rule, since it will already remove a lot of crashes?

AdrianGaudebert:
@twobraids OK I have moved the SQL query to action time, added something to the ``processor_notes``, and added some logic change to the middleware so that nothing breaks during the transition.

AdrianGaudebert:
Should I add another rule, in the classification rules system, to mark all crashes with a ``b0`` suffix so they are easily re-processable? Or is it too early? Do we have the reprocessing system yet?

socorro/processor/mozilla_transform_rules.py
```diff
@@ -816,3 +816,93 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
+            key = '%s:%s:%s:%s' % (product, version, channel, build)
+            versions_data[key] = real_version
+
+        return versions_data
+
+    #--------------------------------------------------------------------------
+    def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
+        try:
+            if processed_crash['release_channel'].lower() == 'release':
+                return True
```
rhelmer: I think you could do this for ```!= 'beta'```, why not?

rhelmer: We don't have this problem with Nightly or Aurora channels, it's only Beta that send crash reports without the specific beta number. The former two are always ```a1``` and ```a2``` respectively no matter what.

socorro/external/postgresql/base.py
```diff
@@ -85,7 +85,7 @@ def query(self, sql, params=None, error_message=None, connection=None):
             if not connection:
                 connection = self.database.connection()
                 fresh_connection = True
-            # self.context.logger.debug(connection.cursor.mogrify(sql, params))
+            # logger.debug(connection.cursor().mogrify(sql, params))
```
rhelmer: This is funny and I have used it before too :) Why don't we allow it to be on for ```debug```, we can run production in ```error``` and below if we don't like it there? Seems like this is what log levels are intended for.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -816,3 +816,93 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
+
+            versions_data = self._get_versions_data()
```
twobraids: not exactly what I had in mind, sorry for not being clearer.  

Rather than refresh the whole cached version of the table every time, change the query so that it looks up the result for this particular product/version/channel/build.  Keep a cache of the results of this query.  

When entering the ```_action```, lookup the product/version/channel/build in the cache, if it is not there, then do the query and add the results to the cache.

socorro/processor/mozilla_transform_rules.py
twobraids: this test to see if the ```_action``` should apply at all ought to be put in the ```_predicate``` method.   See the implementation https://github.com/mozilla/socorro/blob/master/socorro/processor/mozilla_transform_rules.py#L462 for an example.

socorro/external/postgresql/base.py
AdrianGaudebert: Historically we have always showed ``debug`` in our production logs, and that has been a pain. But we could easily change that I suppose. 

The other reason I can think about is that this is an option that you only want to turn on when you are debugging those SQL queries, which is not always the case. It's quite painful to have to go through a lot of log when debug something unrelated, so I think less log is better. 

But maybe it's just convenient to me because I know exactly where to go to turn this debugging on...

socorro/processor/mozilla_transform_rules.py
AdrianGaudebert: @rhelmer that's what we do in the middleware, and as far as I remember I was told to do so by Josh. Maybe he told me to do that to be future-ready, in case we started doing rapid aurora and such? 

For context in the middleware, see: https://github.com/mozilla/socorro/blob/master/socorro/external/postgresql/base.py#L400 and https://github.com/mozilla/socorro/blob/master/config/middleware.ini-dist#L472

socorro/external/postgresql/base.py
rhelmer: Yeah me too (and don't let me derail this PR with this discussion!), it honestly wouldn't hurt to have this in prod either and would be helpful sometimes. If we're running with debug on in prod we kind of deserve/expect to have debug output, and this is not so much really.

socorro/processor/mozilla_transform_rules.py
rhelmer: Hm well right now it's just a waste since we only take action on beta, I think it's not worth it.

AdrianGaudebert: Good to know! I'll change that.

AdrianGaudebert:
OK I think this only needs a migration now. @twobraids, @rhelmer, did I forget anything?

rhelmer:
r+

twobraids:
r+

socorro/external/postgresql/raw_sql/procs/001_update_reports_clean.sql
```diff
@@ -253,7 +253,7 @@ SET product_version_id = product_versions.product_version_id
 FROM product_versions JOIN product_version_builds USING (product_version_id), new_reports
 WHERE reports_clean_buffer.uuid = new_reports.uuid
     AND new_reports.product = product_versions.product_name
-    AND new_reports.version = product_versions.release_version
+    AND new_reports.version = product_versions.version_string
```
AdrianGaudebert: @selenamarie Can you review this little change please? I don't know enough about our schema to know if that will have any consequences.

PR #2726, opened 2015-04-09
Bug 1152243 Add an icon to links that add terms to the current search

Tchanders:
Also remove external-link class from uuids and italicise


webapp-django/crashstats/supersearch/templates/supersearch/search_results.html
```diff
@@ -24,7 +24,7 @@
         <tbody>
             {% for crash in query.hits %}
             <tr>
-                <td><a href="{{ url('crashstats:report_index', crash.uuid) }}" class="external-link">{{ crash.uuid }}</a></td>
+                <td><em><a href="{{ url('crashstats:report_index', crash.uuid) }}">{{ crash.uuid }}</a></em></td>
```
AdrianGaudebert: I'm not fan of this ``<em>`` here, I'd rather have those crash ids have the same CSS they have in the signature report page: https://github.com/mozilla/socorro/blob/master/webapp-django/crashstats/signature/static/signature/css/signature_report.less#L51-L56

AdrianGaudebert:
I like this! The icon looks good and links are a bit more explicit now. 

KaiRo has been asking in the bug that we change this view a bit, and would like the link on the actual signature to send to the signature report page, instead of the ``Signature report`` link. I'm not sure what the best solution would be here. Replace the ``Signature report`` link with an ``Add term`` link? What do you think? 

Also, where does the icon come from? Does it have a free license?

Tchanders:
Hi Adrian,

I made the icon myself so it's public domain.

Have also made those changes to the crash ids, as you'll see...

In terms of the Signature report link, I agree with you that it might look
strange to have all the filters in the table look the same, except for the
signatures. I guess it's a case of consistency of the signature links
throughout the site vs consistency of filters within the table... To me it
does look clearer that the signature is a filter rather than a link to the
page now there is an icon, but maybe I'm biased! I'm happy to try the Add
term link if you think it would look clearer for regular users.

Thalia

AdrianGaudebert:
r+

PR #2729, opened 2015-04-09
fix bug 1152641 - nginx needs to accept larger uploads for crash-{stats,...

rhelmer:
...reports}

r? @jdotpz


jdotpz:
That would be an R+.
http://stream1.gifsoup.com/view7/2598739/r-keine-lust-o.gif

Server context is right for that directive, and we believe the directive understand's Gs.

rhelmer:
Thanks! Worst case scenario we bust stage, that's One Way To Learn I guess (pretty sure this works fine, I see references to it but nginx docs aren't clear.)

PR #2730, opened 2015-04-09
Revert "fix bug 1123908 - use env vars for django config"

rhelmer:
This reverts commit e9bf32ef4ee6cc1f1b5c20d0a02a8cb175f107c1.


rhelmer:
This isn't working on stage/prod (silently failing on stage), complaining about ```SECRET_KEY``` not being set so my ```local.py``` fallback is not working :(

rhelmer:
Actually figured out a fix for this, PR #2731

PR #2731, opened 2015-04-09
bug 1123908 - this errors out in PHX since there is no default for SECRE...

rhelmer:
...T_KEY, put a default back in for now

r? @peterbe


rhelmer:
The problem is decouple is throwing an exception because ```SECRET_KEY``` has no default *before* local.py is imported, and PHX doesn't set it in the env

So I think we should just set a default for now so PHX works, and remove it again once we're out of there, since I don't want to fix all the myriad PHX scripts.

peterbe:
r+

PR #2732, opened 2015-04-10
(no bug) some places still link to our old install docs, provide a landi...

rhelmer:
...ng page here

r? anybooody


AdrianGaudebert:
:+1:

PR #2735, opened 2015-04-13
Fix bug 1151436. Handle signatures over 255 chars correctly.

jkozera:
Runs SigTrunc before SignatureRunWatchDog to truncate signatures with "shutdownhang" correctly. Also adds a simple test for mozilla_processor_2015.py.

### DIFF
--- a/socorro/unittest/processor/test_mozilla_processor_2015.py
+++ b/socorro/unittest/processor/test_mozilla_processor_2015.py
@@ -1,0 +1,107 @@
+import ujson
+
+from configman import ConfigurationManager
+from configman.dotdict import DotDict
+
+from mock import Mock, patch
+from nose.tools import eq_, ok_
+
+from socorro.processor.mozilla_processor_2015 import (
+    MozillaProcessorAlgorithm2015
+)
+from socorro.unittest.testbase import TestCase
+
+
+thread_over_255_chars = {
+    "frames": [
+        {
+            "file": "nsTerminator.cpp:604367e1fa5e",
+            "frame": 0,
+            "function": ("mozilla::`anonymous namespace::"
+                         "RunWatchdog(void *)" + "a"*255),
+            "function_offset": "0x0",
+            "line": 151,
+            "module": "xul.dll",
+            "module_offset": "0x783f2b",
+            "offset": "0x67903f2b",
+            "registers": {
+                "eax": "0x0000003f",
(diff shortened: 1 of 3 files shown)


rhelmer:
Thanks, this lgtm!

However does it really fix bug 1151436 ? It seems like it puts in a test that shows that processor is not causing the problem (which is great!)

I think the problem in bug 1151436 is that our ```reports``` table in postgres has a limit on the ```signature``` of 256 characters (```varchar(256)```)

One fix for that would be to make it ```text``` instead, but I don't know the implications of altering the reports table (it's large and has many partitions and is pretty central to our current reporting system so a bit hard to test.)

I am going to ask @AdrianGaudebert in the bug if we're planning to re-land the Elasticsearch backend for the ```/report/list``` page, which should fix this as well.

socorro/unittest/processor/test_mozilla_processor_2015.py
```diff
@@ -0,0 +1,107 @@
...
+            raw_crash,
+            raw_dumps,
+            DotDict()
+        )
+
+        ok_(processed_crash.success)
+        eq_(processed_crash.processor_notes,
+            'dwight; MozillaProcessorAlgorithm2015; '
+            'SignatureTool: signature truncated due to length; '
+            'SignatureTool: signature truncated due to length')
```
twobraids: this is interesting - I'm not sure, off hand, why the processor note about truncation is repeated.   Looking in production database, I can only see that condition when a crash is reprocessed, and in that situation, the two messages are not contiguous.    I'm going to spend some time investigating why it repeats here.

twobraids: ah, the signature generator, a class that predates the rule system has a truncation routine in it. So a long signature is truncated on generation.  Then other rules lengthen the signature, and then the SigTrunc rule truncates it a second time.  Not ideal, but out of the scope of this bug to repair.

twobraids:
r+  
thank you very much for this patch.  You did a great job.  I'll be merging this into the code base later today.

twobraids:
I've created http://bugzil.la/1155165 for the duplication issue.  @jkozera if you have an interest, you are invited to look at that bug, too.  Don't feel obligated, it'll get resolved by someone eventually.  

Again, thank you for the resolution of this problem.

PR #2736, opened 2015-04-13
Fixes Bug 713973, Bug 1042420, Bug 1044183, Bug 913581, Bug 999923, deprecate & remove all the things

twobraids:
the time has come for the code removal party.  All this stuff is old, unused, unwanted and unloved.  It's time to remove them all.  Farewell, ```schema.py```


peterbe:
I can't see any file that I have a relationship with that I would understand. r+

AdrianGaudebert:
I would be in favor of simply testing this on stage. It's difficult to catch the subtle errors that this would cause in a local environment, so we might as well just merge and watch it make wonders, or open the gates of Hell. 

r+

selenamarie:
:100: Well, 102, but close enough :dancer:

twobraids:
blocked by the daily_url cron.

twobraids:
@rhelmer r?

we'll have to run this in stage for a bit to ensure that I've not inadvertently killed something that we actually still use.

rhelmer:
r+ daily_url cron is dead

rhelmer:
We need to do a push, going to merge this after that so it has some time to bake on stage.

lonnen:
This will break some tooling the GFX team has and really relies on.

rhelmer:
aaand we're back to CSV being a blocker :)

lonnen:
@twobraids had a workaround for this. It has either already landed or it was not opened, but either way he will track down the branch

lonnen:
deleted

PR #2737, opened 2015-04-15
fix bug 1154935 - jenkins never actually modifies the alembic.ini

rhelmer:


twobraids:
that looks better
r+

rhelmer:
Not waiting for travis, this is in a block that doesn't run there and we already verified it fixes jenkins.

PR #2738, opened 2015-04-17
bug 1155504 - support apache-style userdirs for crash-analysis

rhelmer:
r? @jdotpz - nginx config similar to existing Apache on https://crash-analysis.mozilla.com/

### DIFF
--- a/config/package/nginx/conf.d/socorro-analysis.conf
+++ b/config/package/nginx/conf.d/socorro-analysis.conf
@@ -1,0 +1,12 @@
+server {
+    listen 80;
+    server_name crash-analysis;
+
+    # Support Apache-style userdirs
+    # see: http://wiki.nginx.org/UserDir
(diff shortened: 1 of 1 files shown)


phrawzty:
The configuration looks fine at first glance (yay), but I'm curious as to *why* we want to do this?

rhelmer:
> The configuration looks fine at first glance (yay), but I'm curious: why do we want to do this?

So we can move https://crash-analysis.mozilla.com/ to AWS. As I mentioned in the bug, I'd like to integrate the stuff hosted here into Socorro proper, but people still depend on some of the reports hosted there.

rhelmer:
Oh, and I am wrong about this server even using ```UserDir``` (there are user directories stuck directly into Apache's ```DocumentRoot```

NEVERMIND :neutral_face:

PR #2739, opened 2015-04-17
bug 1155504 - enable autoindex for crash-analysis

rhelmer:
@phrawzty I think this is what we really need for https://crash-analysis.mozilla.com

### DIFF
--- a/config/package/nginx/conf.d/socorro-analysis.conf
+++ b/config/package/nginx/conf.d/socorro-analysis.conf
@@ -1,0 +1,9 @@
+server {
+    listen 80;
+    server_name crash-analysis;
+
+    location / {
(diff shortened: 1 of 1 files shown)


config/package/nginx/conf.d/socorro-analysis.conf
```diff
@@ -0,0 +1,9 @@
+server {
+    listen 80;
+    server_name crash-analysis;
+
+    location / {
+        index  index.html index.htm;
```
phrawzty:
```
    index  index.html index.htm;
    #    ^^  whitespace?
```

phrawzty:
`r+` with nit.

PR #2740, opened 2015-04-17
(no bug) fix syntax error in example nginx config files

rhelmer:
r? @jdotpz just updated stage and discovered I got this wrong


phrawzty:
`r+`

PR #2742, opened 2015-04-18
no bug - fix config key, should be username not password here

rhelmer:
r? @peterbe

### DIFF
--- a/webapp-django/crashstats/settings/base.py
+++ b/webapp-django/crashstats/settings/base.py
@@ -581,7 +581,7 @@ DATASERVICE_CONFIG_BASE.update({
                 'aPassword'
             ),
             'database_username': config(
-                'DATASERVICE_DATABASE_PASSWORD',
+                'DATASERVICE_DATABASE_USERNAME',
                 'breakpad_rw'
             ),
             'database_hostname': config(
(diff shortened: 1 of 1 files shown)


webapp-django/crashstats/settings/base.py
```diff
@@ -555,7 +555,7 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
                 'aPassword'
             ),
             'database_username': config(
-                'DATASERVICE_DATABASE_PASSWORD',
+                'DATASERVICE_USERNAME',
```
peterbe: Shouldn't that be `DATASERVICE_DATABASE_USERNAME` to follow suit?

peterbe:
Would you mind perhaps tying this to a bug and include one more fix? [This](https://github.com/rhelmer/socorro/blob/nobug-fix-config-key/webapp-django/crashstats/settings/base.py#L569) should be `database_port`.

peterbe:
The reason for tying it to a bug is because I think yanking the `local.py-dist` is quite a landmark move.

peterbe:
If you merge and rebase in https://github.com/rhelmer/socorro/pull/23 this should be good to go.

rhelmer:
@peterbe hah! I didn't actually mean to `rm` the `local.py-dist` in this PR, just fix that one typo! I've reverted that bit.

rhelmer:
@peterbe ok how's that

peterbe:
r+

PR #2743, opened 2015-04-20
Bug 1156321 attributeerror on connectionrollback in postgres base

peterbe:
r? @twobraids 

I actually think we can refactor even more. I don't like that the big fat `try: ... except psycopg2.Error` block encompasses more than 1 thing. However, the DRY was a bigger violation for now. 

This is actually a fairly high priority fix because it without it we'll never have constructive error messages when the connection can't be made.

### DIFF
--- a/socorro/external/postgresql/base.py
+++ b/socorro/external/postgresql/base.py
@@ -75,27 +75,13 @@ class PostgreSQLBase(object):
                       will be opened.
 
         """
-        fresh_connection = False
-        try:
-            if not connection:
-                connection = self.database.connection()
-                fresh_connection = True
-            # self.context.logger.debug(connection.cursor.mogrify(sql, params))
-            results = execute_query_fetchall(connection, sql, params)
-            connection.commit()
-        except psycopg2.Error, e:
-            if error_message is None:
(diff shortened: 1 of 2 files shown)


socorro/external/postgresql/base.py
```diff
@@ -111,25 +97,34 @@ def count(self, sql, params=None, error_message=None, connection=None):
...
+        self, transformer, sql, error_message, params=None, connection=None
+    ):
         fresh_connection = False
         try:
             if not connection:
                 connection = self.database.connection()
                 fresh_connection = True
             # self.context.logger.debug(connection.cursor.mogrify(sql, params))
-            result = single_value_sql(connection, sql, params)
+            result = transformer(connection, sql, params)
```
twobraids: I do not understand the name of this method, ```transformer```.  Perhaps, something like ```transaction_function``` would be clearer.

twobraids:
minor nit, otherwise, r+
note this code will likely go away entirely with our upcoming dataservices initiative

peterbe:
I changed the name from `transformer` to `actor_function`. 
Will merge as soon as travis has had a second go at it.

PR #2745, opened 2015-04-20
fixes bug 1156394 - Use dj-database-url to unpackage settings.DATABASES

peterbe:
r? @rhelmer

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -184,3 +184,6 @@ flake8==2.4.0
 mccabe==0.3
 # sha256: ufv3R_qOcR0zDy9Da5suz17tnrZ6Y3yoGrzRHCq-3sg
 python-decouple==2.3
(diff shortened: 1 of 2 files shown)


peterbe:
Oh. I think I missed the ci part.

peterbe:
Actually, this is tricky. 
You actually get a warning in the [travis tests](https://api.travis-ci.org/jobs/59293735/log.txt?deansi=true) because the travis tests rely on the defaults. (e.g. `django.db.backends.sqlite3` and `sqlite.crashstats.db`)

So, we actually only want to bother with the deprecation warning IF the `DATABASE_URL` is NOT set and some of the `DATABASE_*` settings are.  I'll need to modify the code.

peterbe:
Right. Addressed in the last commit.

peterbe:
@rhelmer ping

rhelmer:
So nice. Now we just need to switch everything over to it :)

rhelmer:
r+

PR #2749, opened 2015-04-27
fixes bug 1158929 - Avoid pyinotify like the plague on OSX

peterbe:
@rhelmer r?

### DIFF
--- a/socorro/processor/symbol_cache_manager.py
+++ b/socorro/processor/symbol_cache_manager.py
@@ -10,12 +10,18 @@ from ordereddict import OrderedDict
 from configman import Namespace, RequiredConfig
 
 
-if os.uname()[0] == 'Darwin':
-    # You're on OSX! Avoid pyinotify like the plague
+if os.uname()[0] != 'Linux':
+    # You're not on Linux! Avoid pyinotify like the plague
     class ProcessEvent(object):
         # Defining a class means we can't define the EventHandler class
         # without indenting the whole thing in an if-block.
         pass
+
+    import warnings
+    warnings.warn(
+        'ProcessEvent is disabled on operating systems that does not '
+        'have inotify in its kernel.'
+    )
 else:
     import pyinotify
     from pyinotify import ProcessEvent
--- a/socorro/processor/symbol_cache_manager.py
+++ b/socorro/processor/symbol_cache_manager.py
@@ -5,15 +5,11 @@
 import os
 import sys
 import tempfile
-import threading
-from datetime import datetime
-from collections import defaultdict
 from ordereddict import OrderedDict
 
 from configman import Namespace, RequiredConfig
-from configman.converters import class_converter, timedelta_converter
 
-from socorro.lib.datetimeutil import utc_now, UTC
+
 if os.uname()[0] == 'Darwin':
     # You're on OSX! Avoid pyinotify like the plague
     class ProcessEvent(object):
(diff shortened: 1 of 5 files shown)


socorro/processor/symbol_cache_manager.py
```diff
@@ -14,11 +14,19 @@
 from configman.converters import class_converter, timedelta_converter
 
 from socorro.lib.datetimeutil import utc_now, UTC
-import pyinotify
+if os.uname()[0] == 'Darwin':
+    # You're on OSX! Avoid pyinotify like the plague
+    class ProcessEvent(object):
+        # Defining a class means we can't define the EventHandler class
+        # without indenting the whole thing in an if-block.
+        pass
```
rhelmer: Can we log a warning message too? This is just going to silently not work otherwise right?

socorro/processor/symbol_cache_manager.py
```diff
@@ -14,11 +14,19 @@
 from configman.converters import class_converter, timedelta_converter
 
 from socorro.lib.datetimeutil import utc_now, UTC
-import pyinotify
+if os.uname()[0] == 'Darwin':
```
rhelmer: You should make it `!= 'Linux'`, since this won't work on anything else (Windows, BSD, etc)

socorro/processor/symbol_cache_manager.py
```diff
@@ -5,27 +5,31 @@
...
-from collections import defaultdict
 from ordereddict import OrderedDict
 
 from configman import Namespace, RequiredConfig
-from configman.converters import class_converter, timedelta_converter
 
-from socorro.lib.datetimeutil import utc_now, UTC
-import pyinotify
 
+if os.uname()[0] == 'Darwin':
```
rhelmer: You should make it != 'Linux', since this won't work on anything else (Windows, BSD, etc)

peterbe:
Wow! That became a serious tangent on python decorators. 

So it turns out that writing a class decorator that triggers a `raise SkipTest` was easy but one that doesn't, really really hard. 

Researching this caused me to did *deeeep* and there's almost no existing examples that gets this right. Long story short, I needed a kick ass decorator that I can set on classes, class methods and functions that **conditionally** raises a `SkipTest`.

peterbe:
cc @twobraids for the part of `skip_if`

socorro/processor/symbol_cache_manager.py
```diff
@@ -5,27 +5,37 @@
...
+        # Defining a class means we can't define the EventHandler class
+        # without indenting the whole thing in an if-block.
+        pass
 
-#==============================================================================
-class EventHandler(pyinotify.ProcessEvent):
-    #--------------------------------------------------------------------------
+    import warnings
+    warnings.warn(
+        'ProcessEvent is disabled on operating systems that does not '
```
rhelmer: "that do not"

peterbe: Who deploys on non-Linux anyway :)

I actually don't know what to make of this comment. My intention was just around development, not about use or deployment. Now you're making me un-confident.

rhelmer:
r+

socorro/processor/symbol_cache_manager.py
rhelmer: We discussed in the meeting, but to close the loop - I am thinking of someone on Windows/Mac/BSD/etc. that has enabled ```SymbolLRUCache``` feature and doesn't have ```inotify``` available, they will get some warning about `ProcessEvent` and most likely not know what that means.

If the warning instead (or additionally) mentioned `SymbolLRUCache`, they'd have some idea of what went wrong.

rhelmer:
r+ if `SymbolLRUCache` is mentioned in the warning ;)

PR #2750, opened 2015-04-27
bug 1158877 - move the daily URL cron to crontabber

peterbe:
for @twobraids's eyes only for now


twobraids:
withdrawn as it will be just be rendered obsolete with the move to AWS.

PR #2751, opened 2015-04-27
fixes Bug 1159776 - stop existing 'dump_checksums' in raw crashes from causing trouble

twobraids:
raw crashes that already have a "dump_checksums" when resubmitted will cause trouble in collection.  The value will overwrite the new checksums being created by collector, causing a TypeError.


rhelmer:
r+

PR #2752, opened 2015-04-28
(no bug) - more explicit instructions on how to get consul going

rhelmer:
r? anybody

### DIFF
--- a/docs/configuring-socorro.rst
+++ b/docs/configuring-socorro.rst
@@ -16,9 +16,18 @@ Configuring Socorro
 Socorro uses a distributed configuration service called Consul to hold
 configuration - https://consul.io/intro/getting-started/install.html
 
-This must be running in order for Socorro apps to start up and access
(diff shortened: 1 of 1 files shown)


phrawzty:
A relative path for `-data-dir` - that seems odd, no?

rhelmer:
@phrawzty true... the docs use `/tmp` but that seems rude. What do you think would be better?

phrawzty:
We use `/var/lib/consul` for reasonable reasons, so that's a possibility.  Honestly, imho anything would be better than a relative directory - we could even just put `-data-dir /dir/of/your/choosing/` fwiw. :cake:

rhelmer:
@phrawzty let's use `/var/lib/consul` and throw a `sudo` in there, lack of write permission was the thing I was worried about.

phrawzty:
This is high-quality bikeshed material, so let's just `r+` and move along. :wink:

PR #2753, opened 2015-04-28
fix bug 1157346 - do not activate nginx configs on package install

rhelmer:
r? @phrawzty as discussed last week


phrawzty:
`r+`

PR #2754, opened 2015-04-29
Bug1159450 more explicit config optional webapp pg

rhelmer:
r? @twobraids @phrawzty @peterbe - this patch makes it clear in the docs that PostgreSQL and the Django webapp are both optional components. I think we should warn external installers away from them, and suggest using ES+Kibana instead.

I find doc changes like this hard to read as diffs, I suggest doing `make html` in the `socorro/docs/` dir if you want to check out what recommended install looks like now, or looking at my copy on http://people.mozilla.org/~rhelmer/Bug1159450/configuring-socorro.html


docs/configuring-socorro.rst
```diff
@@ -3,271 +3,138 @@
...
 configuration - https://consul.io/intro/getting-started/install.html
 
 Consul must be running in order for Socorro apps to start up and access
 their configuration. Normally you want to run a cluster (see the docs above)
-but to start in a single node configuration::
+but to start in a single node configuration add this to
+/etc/consul/server.json::
 
-  sudo consul agent -server -bootstrap-expect 1 -data-dir /var/lib/consul
+  {
```
phrawzty: This block assumes that the Consul package was installed from our repo (and, thus, includes `/etc/consul/common.json`).  Given that this is only true if the user is using our repo - the suggestion for which occurs in `production-install.rst` *only* - I don't know if the assumption in question is a safe one.

In other words, either we mandate that our repo be used from the get-go, or we must assume that it isn't being used at all and write our documentation accordingly.  In the latter case, the previous revision (command-line invocation) is the only safe possibility.

docs/configuring-socorro.rst
```diff
@@ -3,271 +3,138 @@
...
-That means that when ``joe.bloggs@example.com`` uploads symbols they are
-stored in a different bucket called ``private-crashes.my-bucket``.
-
-If you additionally want to use a different location for this user you
-can enter it as a tuple like this:
+Mozilla runs a service at https://crash-stats.mozilla.org that produces
+graphs and reports for developers.
 
-.. code-block:: python
+Both the crash-stats app and the PostgreSQL schema it depends on are very
```
phrawzty: Is there some way to make this section be in an 180 point flashing font? :laughing:

docs/configuring-crash-stats.rst
```diff
@@ -0,0 +1,310 @@
+.. index:: configuring-crashstats
+
+.. _configuring-crashstats-chapter:
+
+Configuring Crash-Stats and PostgreSQL
+======================================
+
```
phrawzty: Right now it's unclear what parts in this section are useful (if any) outside of our use-case. Let's be sure to be absolutely clear about the elements of general vs. Moz interest.

docs/configuring-crash-stats.rst
```diff
@@ -0,0 +1,310 @@
...
+Both the Django socorro-webapp and the socorro-middleware REST service
+must be running::
+
+    sudo systemctl enable socorro-middleware socorro-webapp
+
+Configure Nginx
+---------------
+
+Both socorro-webapp and socorro-middleware should be fronted by a
+webserver like Nginx.
```
phrawzty: *Why* should they be fronted?

docs/configuring-crash-stats.rst
```diff
@@ -0,0 +1,310 @@
...
+You can leave the default "socorro-middleware" in socorro-middleware.conf
+
+Copy these .sample files to .conf and restart Nginx to activate::
+
+  sudo systemctl restart nginx
+
+Cron jobs
+---------
+
+Socorro uses a crontab manager called Crontabber. This needs
```
phrawzty: Suggest adding a link to Crontabber proper.

"This needs to be in [...]" is ambiguous; prefer something like "crontabber needs to run regularly from the system cron, for example [...]".

docs/configuring-crash-stats.rst
```diff
@@ -0,0 +1,310 @@
...
+
+Socorro produces graphs and reports, most are updated once per day.
+
+You must enter information about your releases into Socorro in order
+for this to work, and this information must match the incoming crashes.
+
+Becoming a superuser
+--------------------
+
+If you're starting a fresh new Socorro instance without any users at
```
phrawzty: :+1: !

docs/configuring-crash-stats.rst
```diff
@@ -0,0 +1,310 @@
...
+was named "KillerApp" then it will appear at:
+
+http://crash-stats/home/products/KillerApp
+
+You should also change the DEFAULT_PRODUCT in local.py (/etc/socorro/local.py
+in a packaged install, ./webapp-django/crashstats/settings/local.py otherwise).
+
+Active Daily Install (ADI)
+--------------------------
+
```
phrawzty: This, for example, is highly Moz-specific.

docs/configuring-socorro.rst
rhelmer: The production installation instructions use our RPMs, no ifs/ands/buts.

docs/configuring-socorro.rst
rhelmer: In a previous revision, instead of "These components are totally optional" below I had "You probably don't want to use it". Could go back to something stronger like that?

docs/configuring-crash-stats.rst
rhelmer: I think the easy answers are because Nginx listens on port 80 and can handle static resources more efficiently. 

Nginx also buffers slow clients which I know can easily lead to a DoS with gunicorn at least (described http://gunicorn-docs.readthedocs.org/en/latest/deploy.html), uwsgi's direct HTTP serving feature may be able to handle this idk.

docs/configuring-crash-stats.rst
rhelmer: What should we link to, more documentation on crontabber or to the github project?

Also I don't mind changing this but note this whole document was copied verbatim out of the original "Configuring Socorro" top-level doc.

docs/configuring-crash-stats.rst
rhelmer: Yes this whole document is, that's why we're warning people away from PG+Crash-Stats right? :)

rhelmer:
Just to clarify in case it's not worded strongly enough - the intent of this PR is to separate the "Configuring Crash-Stats and PostgreSQL" out of the main installation flow (the "main" flow is what you get if you mash the "Next" button) and to warn people away from using these optional components, since they are very Mozilla-specific.

Over time we may choose to more strongly deprecate crash-stats, but I don't want to go quite that far today. For now I think if external Socorro users avoid Postgres and Crash-Stats then they'll be much, much more supportable and easier for everyone involved.

rhelmer:
I worry that the "curl" commands in here are pretty unreadable. In a future version I am hoping we can show how to set things using environment variables and then move on to Consul, but we hardcode envconsul in our systemd service files right now so I don't think it's very realistic today.

docs/configuring-crash-stats.rst
rhelmer: No sections in ```configuring-crashstats``` doc are useful outside of Mozilla. I'll state this at the top of the doc too.

docs/configuring-crash-stats.rst
rhelmer: Just to clarify further - ADI is highly-Mozilla specific, and it is absolutely required for Postgres reporting and crash-stats to function.

docs/configuring-crash-stats.rst
phrawzty: Maybe add some of that sweet sweet explanation in the doc, then. :smile:

docs/configuring-crash-stats.rst
phrawzty: Socorro uses a scheduler called [Crontabber](https://github.com/mozilla/crontabber) that needs to be triggered via system cron:
```
# /etc/cron.d/socorro-crontabber
*/5 * * * * socorro /data/socorro/application/scripts/crons/crontabber.sh
```
More documentation about Crontabber is [available here](https://crontabber.readthedocs.org/en/latest/).

docs/configuring-crash-stats.rst
```diff
@@ -0,0 +1,374 @@
...
+that Socorro and crash-stats expect.
+
+Configure Socorro Processor
+---------------------------
+
+Socorro Processor must be configured to store crashes in Elasticsearch
+as well as PostgreSQL.
+
+NOTE - you must replace @@@DATABASE_HOSTNAME@@@, @@@DATABASE_USERNAME@@@ and
+@@@DATABASE_PASSWORD@@@ below with valid entries for your PostgreSQL install
```
peterbe: I don't know what these `@@@` are but if you want to make the strings monospace you can just use a double back-tick.

rhelmer:
I'm maintaining a built copy of this on http://people.mozilla.org/~rhelmer/Bug1159450/configuring-socorro.html for easier review.

peterbe:
r+ on the few things I get.

docs/configuring-socorro.rst
rhelmer: I went strong again.

rhelmer:
OK thanks for the review comments - I am going to go ahead and land this now. Still very open to changes ! But I literally had someone come today and start setting up crash-stats who really doesn't need to, using the current instructions which are unclear about how to use Consul and what it's for ("instructions unclear, infra stapled to forehead")

PR #2755, opened 2015-04-29
typo in chown command, needs to specify username

rhelmer:
r? @phrawzty - not sure how this is being installed successfully anywhere with this bug, but apparently it is!


phrawzty:
Haha *woops*. `r+`

PR #2758, opened 2015-04-30
Remerge all the things

AdrianGaudebert:
@peterbe @lonnen 

This is the counterpart of https://github.com/mozilla/socorro/pull/2714, it re-merges all the things that I had to remove because of [bug 1000160](https://bugzilla.mozilla.org/show_bug.cgi?id=1000160). Now that this bug is solved, we should be able to re-land all those things. All those commits where in master once, there should be no need to review them again. 

It might be better to wait a bit before merging though, as to make sure that the work for bug 1000160 works as intended, and to make things easier to deal with in case it does not. But at least the PR is ready.


peterbe:
+1 on letting the fix in 1000160 really land on prod and be tested there.

PR #2759, opened 2015-04-30
tests for cleaner

peterbe:


peterbe:
It's ok that the tests fail. Lars and I are actively working on this.

PR #2761, opened 2015-04-30
fixes bug 1159993 - rewrite release channel for certain Fennec

twobraids:
add new rule to rewrite the release channel for fennec to 'beta' when (product name == Fennec and build_id == 20150427090529 and channel == 'release')


rhelmer:
r+ looks like what I'd expect..

PR #2763, opened 2015-05-01
docs only - show how to get a mware+crash-stats that really starts up

rhelmer:
r? @peterbe @phrawzty @twobraids - I think we could make configuration easier here, but we could do worse than showing explicit working `curl` commands. At least this will get someone from 0 to a working system.

Also, I am starting to think we should adopt the config setup we're using over in https://github.com/mozilla/socorro-infra/tree/master/socorro-config - that is, keep these in "KEY=VALUE" format in a file and have a bulk loader, maybe something like `setup-socorro.sh consul```? All these `curls` make my eyes kind of glaze over.

### DIFF
--- a/docs/configuring-crash-stats.rst
+++ b/docs/configuring-crash-stats.rst
@@ -5,16 +5,30 @@
 Configuring Crash-Stats and PostgreSQL
 ======================================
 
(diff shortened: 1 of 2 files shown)


rhelmer:
Closed in favor of PR# 2766

PR #2764, opened 2015-05-01
Revert "Bug 1152243 Add an icon to links that add terms to the current s...

rhelmer:
...earch"

This reverts commit 2b502dc9f3e85e6c57bc58b9039c7a1405dc3d8c.

r? @peterbe - reverting this on prod at the request of @KaiRo-at and dmajor.


peterbe:
Perhaps I read the feedback from @KaiRo-at too quickly but wouldn't it be better to make the link do two things instead? E.g. Kairo says he sometimes wants to middleclick on the signature to go to the report list for that signature. But if he single-clicks it should add it to the search. 

The way to have both pieces of functionality would be to...:
```html
<a href="/report/list?signature=foo" class="term" data-field="signature">foo</a>
```
...with...:
```javascript
$('a.term').click(function() {
    currentMode.addTerm(
            $(this).data('field'),
            $(this).text().trim()
        );
    submitButton.click();
    return false;
})
```

Thus, single-click amends the current super search, and middleclick goes to the signature report.

peterbe:
Or, perhaps instead of using a hyperlink make it something like this instead:
```html
<td>
    <a class="term" data-field="signature" data-value="foobar"
         title="Add this signature to the current search"
    ></a> foobar
</td>
```
...with...:
```css
a.term {
    cursor: pointer;
    display:block;
    width: 16px; height: 16px;
    background: url('plus.png') no-repeat;
}
```
This'd apply equally not just to signatures but also to all other possible fields that are displayed.

KaiRo-at:
IMHO, we should revert the patch that landed for now, and then discuss on the bug how to fix it in a better way.

peterbe:
for the record; r+ but I don't think we should give up entirely.

rhelmer:
@peterbe thanks! I agree, I reopened the bug and we can work on it there and re-land.

PR #2766, opened 2015-05-01
Bug1160603 load config files into consul

rhelmer:
r? @peterbe @twobraids - more readable config, that happens to be be more amenable to loading directly into the environment. This contains the fixes from and replaces PR #2763

### DIFF
--- a/docs/configuring-crash-stats.rst
+++ b/docs/configuring-crash-stats.rst
@@ -5,16 +5,30 @@
 Configuring Crash-Stats and PostgreSQL
 ======================================
 
-https://crash-stats.mozilla.org is Mozilla's front-end to expose crash-stats
-to developers and interested users. It is written with Mozilla's unique 
-requirements in mind, and will probably not be useful to other Socorro users.
+You probably do not want to use this, unless you have *very* similar requirements
+to Mozilla. Try using Kibana instead.
 
-You probably do not want to use this, unless you are Mozilla.
+https://crash-stats.mozilla.org is Mozilla's front-end to expose crash data
+to developers and interested users.
+
+Some of the unique features it provides are:
+
+* public view of crash reports, with private data redacted
+* ability to assign users fine-grained permissions to private data
+* graphs and reports showing crashes per Active Daily Install (ADI)
+* ability to ingest Mozilla release engineering metadata
+* Bugzilla integration
+* extensive support for the Firefox rapid release cycle
+
+All of these features are baked in and are not trivial to disable or adjust.
+
+Again, if you are not Mozilla you almost certainly would be happier without
+using the Crash-Stats frontend.
 
 Install Memcached
 -----------------
 
-Crash-stats makes heavy use of memcached, you definitely want to use it::
+Crash-Stats makes heavy use of memcached::
 
   sudo yum install memcached
   sudo systemctl enable memcached
@@ -33,7 +47,7 @@ Now you can actually install the packages:
 
 Initialize and enable PostgreSQL on startup:
 ::
-  sudo service postgresql-9.3 initdb
+  sudo /usr/pgsql-9.3/bin/postgresql93-setup initdb
   sudo systemctl enable postgresql-9.3
 
 Modify postgresql config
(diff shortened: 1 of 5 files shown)


rhelmer:
Now we can use syntax highlighting for config!

I've pushed a built copy up to http://people.mozilla.org/~rhelmer/Bug1160603/configuring-socorro.html - I think it's downright readable now.

docs/configuring-socorro.rst
```diff
@@ -27,13 +27,37 @@ The Socorro systemd service scripts use envconsul
...
+    web_server__wsgi_server_class='socorro.webapi.servers.WSGIServer'
+
+Put this into a file named "collector.conf" in your socorro-config folder. 
+
+For processing to work, the path to your debug symbols must be provided:
+
+.. code-block:: bash
+
+    # This should be replaced with the path to your debug symbols.
+    processor__raw_to_processed_transform__BreakpadStackwalkerRule__processor_symbols_pathname_list='@@@PATH_TO_YOUR_SYMBOLS@@@'
```
rhelmer: @twobraids - the config key for the path to symbols seems crazy long, can we do anything to make this shorter? 

I know it's down in one one of the processor transform rulesets which makes sense from our perspective, but it probably seems like a really fundamental setting from the point of view of someone installing Socorro.

If you think we should leave as-is, then I have a second proposal in mind - we should change the default to something friendlier, I'd feel better telling people "put your symbols in `~socorro/symbols/` or else do this more complicated config.

docs/configuring-crash-stats.rst
```diff
@@ -68,77 +82,161 @@ Create database, set up schema
...
-  curl -s -X PUT -d "socorro.external.fs.crashstorage.FSTemporaryStorage" localhost:8500/v1/kv/socorro/processor/storage.crashstorage0_class=socorro.external.fs.crashstorage.FSTemporaryStorage
-  curl -s -X PUT -d "socorro.external.es.crashstorage.ESCrashStorage" localhost:8500/v1/kv/socorro/processor/destination.storage1.crashstorage_class
-  curl -s -X PUT -d "socorro.external.es.crashstorage.PostgreSQLCrashStorage" localhost:8500/v1/kv/socorro/processor/destination.storage2.crashstorage_class
-  curl -s -X PUT -d "@@@DATABASE_HOSTNAME@@@" localhost:8500/v1/kv/socorro/processor/resource.postgresql.database_hostname
-  curl -s -X PUT -d "@@@DATABASE_USERNAME@@@" localhost:8500/v1/kv/socorro/processor/secrets.postgresql.database_username
-  curl -s -X PUT -d "@@@DATABASE_PASSWORD@@@" localhost:8500/v1/kv/socorro/processor/secrets.postgresql.database_password
+Now load the contents of your socorro-config directory::
 
+  cd ./socorro-config
+  sudo setup-socorro.sh consul
```
rhelmer: Although I think this is an improvement, there are a few things I don't like about this:
* this is repeated in several places
* if you follow the instructions you need to reload everything, which is slow (but not harmful)

Maybe it should take an argument like `sudo setup-socorro.sh consul <config_file>`?

I am also not sure how clear it is that you need to create a `socorro-config` directory and create these as new files in there, but this is the best way I've found for dealing with Consul and it's what we do in AWS.

I'll see if the Consul folks are interested in implementing their own bulk-load command, since this seems like a reasonable workflow. Otherwise it's easy to lose track of things in the Consul Web UI or get lost in a sea of `curl` commands.

rhelmer: The only similar open issue I see right now is https://github.com/hashicorp/consul/issues/886 which would allow bulk loading in the sense of atomic change of several keys at once (which is something we'd like too), but nothing specifically about a command for loading multiple config files into Consul as a simple operation (e.g. `consul load <config>`)

I'll file a bug on the latter.

rhelmer: Filed https://github.com/hashicorp/consul/issues/899

peterbe:
Any specific bits you think I should look at? I see a lot of changes that are out of my scope/domain.

docs/configuring-socorro.rst
rhelmer: Note on the above, we did end up changing the default to `~/socorro/symbols` so I will make this change in the doc.

twobraids: the ```reference_value_from``` can be used to create a shorter form of the configuration key.  However, I do not think that it would make a new value that was short enough.  ```reference_value_from``` specifies an alternative namespace (or nested namespaces) where a default for the option can be found.  For example:
``` python
n0.n1.n2.add_option('o1', default=99, reference_value_from='r')
```
produces: 
```python
#n0.n1.n2.o1=99
r.o=99
```
on the command line, the value can be changed by referencing it using either path:
```bash
$ ./my_app.py --n0.n1.n2.o1=65
$ ./my_app.py --r.o1=65
# the original form overrides the reference_value_from. the value will be 65 in the next invocation
$ ./my_app.py --n0.n1.n2.o1=65 --r.o1=222  
```
reference values can be shared between several options.  This is what we do with database and other crash storage credentials.  That allows multiple sections of code that need access to PG, to declare their need within their own namespace context, but all pick up their config from a common location.  

The reference value system doesn't allow for changing the name of a option.  ```reference_value_from``` always is a list of namespaces into which the original name of the option will be inserted.  

It would not bee too difficult to create a new ```alias=``` parameter for the Option object to create a true shortcut to a long deeply nested value.

twobraids:
I'm not noticing anything that stands out as incorrect.

rhelmer:
> Any specific bits you think I should look at? I see a lot of changes that are out of my scope/domain.

Well from that perspective it'd be helpful to know if it seems to make sense :) This PR fixes a few errors along the way so I am going to merge now, but if you see something that doesn't make sense or just seems weird let me know.

PR #2765, opened 2015-05-01
contributes to fixing bug 1159993 - fix fennec ADI channel

twobraids:
integrated the update sql as a final step in the fetch ADI from hive job

PR #2767, opened 2015-05-02
Fennec adi correction

rhelmer:
This replaces PR #2765


rhelmer:
As this code is from lars originally, even though I opened the PR, I am going to r+ :)

PR #2768, opened 2015-05-03
bug 1160603 - use a simpler default for symbols, do not hardcode mozilla...

rhelmer:
... products

r? @twobraids - I mentioned this over in https://github.com/mozilla/socorro/pull/2766/files#r29558442 - we shouldn't harcode Mozilla paths in here in any case, and we decided for S3 to go with a flat namespace for all products and we don't use any of these paths ourselves anymore.

Anyone setting up a Socorro install from scratch today I'd advise to just put all symbols into a flat namespace somewhere easy like `~socorro/symbols/`.


twobraids:
I totally agree with this effort
r+

PR #2769, opened 2015-05-04
Include Linux lsb-release info in stackwalker JSON output

luser:
Linux dumps include this info, I'd like to get it into the JSON. The data is simple key=value pairs:
```
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.2 LTS"
```
so most of this patch is "implement stock Python string methods in C++" (split, startswith, lower).

The only goofy thing here is that description field that is quoted, I didn't strip the quotes. I could do that without much trouble I just wasn't sure if it was worth it. The output looks like this:
```json
{
   "lsb_release" : {
      "codename" : "trusty",
      "description" : "\"Ubuntu 14.04.2 LTS\"",
      "id" : "Ubuntu",
      "release" : "14.04"
   }
}
```
r? @twobraids

### DIFF
--- a/minidump-stackwalk/stackwalker.cc
+++ b/minidump-stackwalk/stackwalker.cc
@@ -748,6 +748,54 @@ static void ConvertMemoryInfoToJSON(Minidump& dump,
 
 }
 
+static vector<string> split(const string& s, char delimiter) {
+  vector<string> bits;
+  size_t prev = 0;
+  size_t pos = s.find(delimiter);
+  while (true) {
+    size_t count = pos == string::npos ? string::npos : pos - prev;
+    bits.push_back(s.substr(prev, count));
+    if (pos == string::npos) {
+      break;
+    }
+    prev = pos + 1;
+    pos = s.find(delimiter, prev);
+  }
+  return bits;
+}
+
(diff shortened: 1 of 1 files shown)


twobraids:
I really think the quotes in the description value field ought to be stripped.  It would not surprise me if we encountered something within socorro that failed to quote them during a serialization process. 

Then again, if you're feeling daring, we could let it stand.

luser:
Yeah, that's fair, although hopefully everything is using a JSON serializer to handle JSON. It looks a little weird, so I added a function to strip quotes, so the output looks like this now:
```json
{
   "lsb_release" : {
      "codename" : "trusty",
      "description" : "Ubuntu 14.04.2 LTS",
      "id" : "Ubuntu",
      "release" : "14.04"
   }
}
```

twobraids:
I approve:  r+

luser:
Rebased and landed as https://github.com/mozilla/socorro/commit/7d2713ea2e40463fd06edde0e9a27117acc7745c

PR #2770, opened 2015-05-04
Fixes bug 1161030 - Added support for generic beta versions to Super Search.

AdrianGaudebert:
@peterbe r?

This needs to be merged before we can take https://github.com/mozilla/socorro/pull/2758 in.

### DIFF
--- a/socorro/lib/search_common.py
+++ b/socorro/lib/search_common.py
@@ -126,7 +126,16 @@ class SearchBase(object):
                 values = [values]
 
             if values is not None:
-                no_operator_param = None
+                # There should only be one parameter with no operator, and
+                # we want to stack all values into it. That's why we want
+                # to keep track of it.
+                # Actually, we want _two_ parameters with no operator: one
+                # for each possible value of "operator_not".
+                no_operator_param = {
+                    True: None,
+                    False: None
+                }
+
                 for value in values:
                     operator = None
                     operator_not = False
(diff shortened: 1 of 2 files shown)


socorro/lib/search_common.py
```diff
@@ -162,27 +171,29 @@ def get_parameters(self, **kwargs):
...
                         parameters[param.name].append(SearchParam(
                             param.name, value, operator, param.data_type,
                             operator_not
                         ))
 
-                if no_operator_param:
-                    parameters[no_operator_param.name].append(
-                        no_operator_param
-                    )
+                for key in (True, False):
```
peterbe: `for key in no_operator_param:`

peterbe: actually, you'll need the value so... 
```python
for key, value in no_operator_param.items():
    if value:
        parameters[value.name].append(value)
```

socorro/lib/search_common.py
```diff
@@ -310,6 +321,42 @@ def fix_hang_type_parameter(parameters):
...
 
+    @staticmethod
+    def fix_version_parameter(parameters):
+        """Correct the version parameter.
+
+        If version is something finishing with a 'b' and operator is
+        'has terms', replace the operator with 'starts with' to query all
+        beta versions.
+        """
+        if 'version' in parameters:
```
peterbe: shouldn't that be `if parameters.get('version'):` because none of the stuff below matters if it's there but empty.

socorro/lib/search_common.py
```diff
@@ -310,6 +321,42 @@ def fix_hang_type_parameter(parameters):
...
                 hang_type.data_type = 'int'
 
+    @staticmethod
+    def fix_version_parameter(parameters):
+        """Correct the version parameter.
+
+        If version is something finishing with a 'b' and operator is
+        'has terms', replace the operator with 'starts with' to query all
+        beta versions.
+        """
```
peterbe: Add a comment that explains that it's ok to use this for ALL products because we're using the rapid-beta process for all products we bother to support. (as discussed)

socorro/lib/search_common.py
peterbe: Or, 
```pyhon
if not parameters.get('version'):
    return
```
...and save yourself a bunch of indentation later :)

socorro/lib/search_common.py
```diff
@@ -310,6 +321,42 @@ def fix_hang_type_parameter(parameters):
...
+                        new_values.append(val)
+
+                if new_values:
+                    version.value = new_values
+                else:
+                    marked_for_deletion.append(version)
+
+            for param in marked_for_deletion:
+                parameters['version'].remove(param)
+
```
peterbe: What you're trying to achieve with the `marked_for_deletion` is easier done this way:
```python
for version in parameters['version'][:]:
      if some_reason(version):
          parameters['version'].remove(version)
```
So basically, by looping over a copy you can afford to do the `.remove()` thing. 

I had to double-check my argument. See [this snippet](https://gist.github.com/peterbe/04186d8ebc7c773c41ad) compared to [this one](https://gist.github.com/peterbe/efa61b928e612532755e) which is the wrong way to do it.

peterbe: Also, if you loop over a copy you don't get the weird thing of adding a value to the list during the iteration.

peterbe:
r+ with stylistic nits

PR #2772, opened 2015-05-06
fixes Bug 1161067, Bug 1161026 - added 2 new prefix signature patterns

twobraids:
adds 'msvcr120\.dll@0x.*' and "CleanupPerAppKey"  to the prefix skiplist


rhelmer:
r+

PR #2773, opened 2015-05-06
fixes Bug 1161648 - new raw & processed copier app for ES migration

twobraids:
This PR introduces a new app that can copy the raw and processed json version of a crash from a source to a destination.  it takes this form:
```bash
$ socorro raw_and_processed_copier 
```
This PR also adds a special ES crashstorage class that employs a redactor to edit keys from the processed crash.  The idea is to use it to remove any key from the processed crash in the form '...json_dump'.

### DIFF
--- a/socorro/collector/crashmover_app.py
+++ b/socorro/collector/crashmover_app.py
@@ -7,6 +7,11 @@
 from configman import Namespace, class_converter
 
 from socorro.app.fetch_transform_save_app import FetchTransformSaveApp, main
+from socorro.external.crashstorage_base import (
+    PolyCrashStorage,
+    CrashIDNotFound,
+)
+from socorro.external.es.crashstorage import ESCrashStorageNoStackwalkerOutput
 
 
 #==============================================================================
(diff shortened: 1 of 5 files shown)


socorro/collector/crashmover_app.py
```diff
@@ -103,5 +108,101 @@ def transform(self, crash_id):
...
+    #--------------------------------------------------------------------------
+    # new_crash_source namespace
+    #     this namespace is for config parameter having to do with the source
+    #     of new crash_ids.
+    #--------------------------------------------------------------------------
+    required_config.namespace('new_crash_source')
+    required_config.new_crash_source.add_option(
+        'new_crash_source_class',
+        doc='an iterable that will stream crash_ids needing processing',
+        default='socorro.processor.timemachine.PGQueryNewCrashSource',
```
rhelmer: We want to run this in parallel across multiple machines (and probably multiple threads per machine as well) - maybe we should make RabbitMQ the default (with a special queue name)

rhelmer:
`PGQueryNewCrashSource` is fine for a default, I'll configure it as needed.

rhelmer:
r+

PR #2774, opened 2015-05-06
fix bug 1162085 - persist uwsgi permissions on reboot

rhelmer:


config/package/tmpfiles.d/uwsgi.conf
```diff
@@ -0,0 +1 @@
+d /run/uwsgi/socorro 0755 socorro nginx -
```
phrawzty: When the RPM is built via [`scripts/install.sh`](https://github.com/mozilla/socorro/blob/master/scripts/install.sh#L21-L22) this conf file will get dropped into `/etc/tmpfiles.d/uwsgi.conf`. Unfortunately, according to the [packaging guidelines](https://fedoraproject.org/wiki/Packaging:Guidelines#binfmt.d.2C_sysctl.d_and_tmpfiles.d), this is an inappropriate behaviour - instead, said conf should be dropped into `/usr/lib/tmpfiles.d/`.

Interestingly, this would be easier to deal with if we were using a real spec.  In any case, we'll need to modify the behaviour of `scripts/install.sh` if we want to adhere to the guidelines.

rhelmer: > 

When the RPM is built via scripts/install.sh this conf file will get dropped into /etc/tmpfiles.d/uwsgi.conf. Unfortunately, according to the packaging guidelines, this is an inappropriate behaviour - instead, said conf should be dropped into /usr/lib/tmpfiles.d/.
>
> Interestingly, this would be easier to deal with if we were using a real spec. In any case, we'll need > to modify the behaviour of scripts/install.sh if we want to adhere to the guidelines.
>
> Thoughts?

I think we're already fairly impure, just by virtue of using `fpm` and shipping a virtualenv (the "right" way would be to have RPM dependencies packaged properly for every one of our ~70 Python dependencies)

That said, we tried to do this right for systemd service files (which have a similar "RPM-provided files go to /usr/lib, human overrides go to /etc", so let's try to do it right here too.

rhelmer: The only problem with doing it "right", or at least right-ish, is that we assume everything in `./package/conf` is going into `/etc` in the staged install area, maybe we should make everything in there go into `/` in the staged install area instead.

rhelmer: The advantage to this is that we can move the systemd service files into `./config/package/` too, they are in an odd place now.

phrawzty: > maybe we should make everything in there go into / in the staged install area instead.

dangerously

rhelmer:
So turns out this is *now* blocking https://github.com/mozilla/socorro-infra/pull/86 because we install the Socorro RPM then build an AMI and shut down :)

I made the requested changes, going to land now. r=me

PR #2779, opened 2015-05-11
fixes Bug 1163421 moved PG OperationalError to Conditional list

twobraids:
Moved the PG OperationalError to the Conditional list so that the OOM can be treated as fatal.  

For the record, I do not believe that this will resolve the larger issue of problems in the processor using too much memory.


rhelmer:
r+, travis is not setting github status for some reason :/ so I'll check on that before merging

rhelmer:
Oh, turned out travis is down for maintenance. Let jenkins do it!

PR #2782, opened 2015-05-13
no bug - stopped PG json saves

twobraids:
this is an experiment to see if stopping the json saves, forestalls a psycopg2 memory leak.


rhelmer:
r+

PR #2783, opened 2015-05-13
no  bug - fix rabbit integration test to not force use of PG for processed crashes

twobraids:
furthering the no json in PG test  to forestall the debilitating memory leak.


rhelmer:
R+

PR #2784, opened 2015-05-13
fixes bug 1164465 - psycopg2 2.5.5

peterbe:
@twobraids r?
cc @rhelmer 
So, the major difference between psycopg2 2.4 and 2.5 (as far as I can see it affecting us) is...

* When you query a table that has a `json` column, what you get is a dict (or whatever is appropriate). Not a string that is JSON text that you need to convert with `json.loads()`. crontabber supports both old and new versions of psycopg2 because it does [this trick](https://github.com/mozilla/crontabber/blob/11bcf6e2e73b1f660ad736142ef52c32affc2170/crontabber/app.py#L245-L246) which makes it more agnostic.
* The actual error from within postgres gets put into exceptions the normal way. No more of this `some_exception.pgcode` or `some_exception.pgerror` See

```python
>>> import psycopg2
>>> psycopg2.__version__
'2.5.5 (dt dec pq3 ext)'
>>> try:
...   psycopg2.connect("dbname='foo'")
... except psycopg2.OperationalError as x:
...   exception = x
...
>>> exception.args[0]
'FATAL:  database "foo" does not exist\n'
>>> dir(exception)
['__class__', '__delattr__', '__dict__', '__doc__', '__format__', '__getattribute__', '__getitem__', '__getslice__', '__hash__', '__init__', '__module__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__setstate__', '__sizeof__', '__str__', '__subclasshook__', '__unicode__', '__weakref__', 'args', 'cursor', 'diag', 'message', 'pgcode', 'pgerror']
>>> exception.pgcode
>>> exception.pgerror
>>>
```

However, I'm not 100% certain that this is what it will do if you get one of those strange `SSL SYSCALL error: EOF detected` errors wrapped in a `ProgrammingError` exception.

### DIFF
--- a/socorro/external/postgresql/connection_context.py
+++ b/socorro/external/postgresql/connection_context.py
@@ -153,7 +153,7 @@ class ConnectionContext(RequiredConfig):
         pass
 
     #--------------------------------------------------------------------------
-    def is_operational_exception(self, msg):
+    def is_operational_exception(self, exp):
         """return True if a conditional exception is actually an operational
         error. Return False if it's a genuine error that should probably be
         raised and propagate up.
@@ -162,16 +162,17 @@ class ConnectionContext(RequiredConfig):
         operational exception "labelled" wrong by the psycopg2 code error
         handler.
         """
-        if msg.pgerror in ('SSL SYSCALL error: EOF detected', ):
-            # Ideally we'd like to check against msg.pgcode values
+        message = exp.args[0]
+        if message in ('SSL SYSCALL error: EOF detected', ):
+            # Ideally we'd like to check against exp.pgcode values
             # but certain odd ProgrammingError exceptions don't have
             # pgcodes so we have to rely on reading the pgerror :(
             return True
 
         if (
-            isinstance(msg, psycopg2.OperationalError) and msg.pgerror not in (
-            'out of memory',
-        )):
+            isinstance(exp, psycopg2.OperationalError) and
+            message not in ('out of memory',)
+        ):
             return True
 
         # at the of writing, the list of exceptions is short but this would be
--- a/requirements.txt
+++ b/requirements.txt
@@ -47,8 +47,8 @@ lxml==3.4.1
 pgxnclient==1.2.1
 # sha256: Zm_VIvShBZ1bi0mW5Xb6eGL1BVKOEc8e9VRjEcYMPk0
 pika==0.9.8
-# sha256: SLcCp8pHnhvCwae4GHWgfUdCmBMmBZk-LLl59eCCd9c
-psycopg2==2.4.5
+# sha256: xsMwyYqwjFY58MzX4hG4Kzo58AGBPc6mu4JKmdKzT-I
+psycopg2==2.5.5
 # sha256: U6gisBsElN-WWMBOhyFhgOXsvoWxeGc92Eix-AnVwQA
 # sha256: u5PYXKflDWjkGXhSu226YJi7dyUA-C9dsv9KgHEVBck
 elasticsearch==1.2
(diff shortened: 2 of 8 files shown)


socorro/external/postgresql/connection_context.py
```diff
@@ -162,14 +162,15 @@ def is_operational_exception(self, msg):
...
+        message = exp.args[0]
+        if message in ('SSL SYSCALL error: EOF detected', ):
+            # Ideally we'd like to check against exp.pgcode values
             # but certain odd ProgrammingError exceptions don't have
             # pgcodes so we have to rely on reading the pgerror :(
             return True
 
         if (
-            isinstance(msg, psycopg2.OperationalError) and msg.pgerror not in (
+            isinstance(exp, psycopg2.OperationalError) and message not in (
```
twobraids: Now I'm guilty of superficial comments about style in a code review...   This if statement would be easier to read at a glance, if it were written as:
```python
        if (
            isinstance(exp, psycopg2.OperationalError)
            and message not in ('out of memory', )
        ):
```

twobraids:
please examine the code in ```.../socorro/external/postgresql/crashstorage.py:422-423 to see if it ought to be updated, too.

twobraids:
this PR needs to be rebased with master.  After that ```.../socorro/external/postgresql/crashstorage.py:192-231``` needs to be examined for the use of json.dumps.  If the psycopg2 reads json columns directly into dict objects, does it also need dict objects for writes?

peterbe:
Regarding line 422-423:
```python
>>> cur=c.cursor()
>>> try:
...   cur.execute('select roses from hell')
... except psycopg2.ProgrammingError as x:
...   err = 'relation "%s" does not exist' % ('hell',)
...   assert err in str(x)
...
>>>
```
In other words, works as expected.

twobraids:
the question is not the error reporting, the question is in regards the necessity of using ```json.loads```.  Is ```crash``` a string or a dict?

peterbe:
Regarding inserting dicts, you have to convert them it seems:
```python
>>> import psycopg2
>>> psycopg2.__version__
'2.5.5 (dt dec pq3 ext)'
>>> conn=psycopg2.connect('dbname=crontabber_exampleapp')
>>> c=conn.cursor()
>>> c.execute("insert into crontabber (app_name, last_error) values('delete_me', %s)", ({'pi':3.14, 'jerks': ['jackson']},))
psycopg2.ProgrammingError: can't adapt type 'dict'
>>> import json
>>> c.execute("insert into crontabber (app_name, last_error) values('delete_me', %s)", (json.dumps({'pi':3.14, 'jerks': ['jackson']}),))
```
But you get it **back** as a dict when you query:
```python
>>> c.execute("select last_error from crontabber where app_name='delete_me'")
>>> first, =c.fetchone()
>>> first
{u'jerks': [u'jackson'], u'pi': 3.1400000000000001}
```

twobraids:
your most recent comment then implies that .../postgresql/crashstorage.py:422-423 needs to be changed.  ```crash``` will be coming back as a dict not a string that needs to be deserialized into a dict.

peterbe:
Oh pardon me. I was looking at the exception handling. You're right, no need to do a `json.loads()`.

I [changed that line](https://github.com/mozilla/socorro/commit/6c056fc107dc78808c814cd4d9e66c66321b5fe2) now and I did check that tests travelled across that path. What's bewildering though is that how the tests could have passed **before**. Really strange.

peterbe:
Oh, it did [fail the tests](https://travis-ci.org/mozilla/socorro/builds/63038759). I see, this is because of the rebase.

peterbe:
@twobraids passing now

twobraids:
then r+

peterbe:
one last travis after a rebase

PR #2787, opened 2015-05-14
fixes bug 1164922 - laglog sometimes chokes on null

peterbe:
@twobraids r?

Note: The old tests never actually tested anything. It never iterated anything back when looping over that `each_server` result. So none of the crucial stuff was ever tested. 

This bug was about protecting against sporadic and bad NULLs coming back from `pg_stats_replication` (most likely related and due to high load or something) but I sorta kill two birds with one stone with this PR.


twobraids:
looks good to me.
r+

PR #2788, opened 2015-05-15
fixes bug 1165012 - strip "report this bug" summary

peterbe:
@AdrianGaudebert r?


AdrianGaudebert:
I like this PR very much! Nice generic solution and nice improvement to readability. 

r+

PR #2789, opened 2015-05-19
fixes bug 1166063 - Support uploads from https://pci-ids.ucw.cz/

peterbe:
@rhelmer r?

Like I said in IRC, I played with this and all the unknown on [this page](https://crash-stats.mozilla.com/report/list?product=Firefox&signature=nsIPresShell%3A%3AAddRefreshObserverInternal%28nsARefreshObserver*%2C+mozFlushType%29) would be resolved. 

It's hard to gauge the **quality** of the database other than the merits of quantity which is partly the reason I've decided to NOT wipe away the old stuff for dealing with CSV files from PCIDatabase.com


rhelmer:
r+ thanks for the thorough test code on that!

PR #2793, opened 2015-05-20
bug 1161812 - copy files to place PHX deploy expects, since that is n…

rhelmer:
…ot running out of the package

r? @peterbe


peterbe:
r+

PR #2794, opened 2015-05-20
fix bug 1166782 - use consistent versioning scheme for RPM

rhelmer:
r? @phrawzty


phrawzty:
`r+` based on the thread in [bug 1166782](https://bugzilla.mozilla.org/show_bug.cgi?id=1166782).

PR #2797, opened 2015-05-20
fixes bug 1166838 - new peep hashes for github.com .zip files

peterbe:
@rhelmer r?

PS. I deleted all [cache repositories](https://travis-ci.org/mozilla/socorro/caches) on Travis just prior to starting this PR.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -124,7 +124,7 @@ pyhs2==0.6.0
 sasl==0.1.3
 # this is a temporary fork until https://github.com/seb-m/pyinotify/pull/92
 # is merged and a new pyinotify release is available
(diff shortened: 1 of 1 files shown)


rhelmer:
r+ if travis is happy. I'll take care of Jenkins caches (if it needs it)

peterbe:
What the heck! Are we [not using peep in travis](https://github.com/mozilla/socorro/blob/master/.travis.yml#L24)?
That means that travis won't check any of the peep hashes.

rhelmer:
> That means that travis won't check any of the peep hashes. 

It looks like we're running `pip` and then later `peep`.

rhelmer:
Not waiting for Travis, it's crawling right now. Jenkins will test this.

PR #2798, opened 2015-05-20
fixes bug 1166866 - No pip install in travis

peterbe:
@rhelmer r?


rhelmer:
I suspect this is there as some kind of optimization - e.g. pip actually installs packages (and caches them), peep checks hashes but doesn't need to install anything. I'll see if there's anything in the PR or bug about it when it landed.

rhelmer:
Let's look at the before/after travis times to see if my comment about this being an optimization is correct.

PR #2802, opened 2015-05-21
fix bug 1167017 - persist /var/lock/socorro on reboot

rhelmer:
@phrawzty `/var/lock/` is a symlink to `/run/lock/` now, so we need to make a `tmpfiles.d` entry to get it to persist on reboot.

I think we should ultimately switch away from custom lock/unlock function to `flock`, but we'll probably need this dir anyway.


phrawzty:
`r+`

PR #2804, opened 2015-05-22
(no bug) - remove useless PG storage class from docs

rhelmer:
r? anybody


phrawzty:
`r+` on the syntax of the doc change - I'll trust you on the accuracy of the information being presented. :grinning:

PR #2805, opened 2015-05-22
bug 1167256 - ensure that betas get correlation reports

rhelmer:
r? @KaiRo-at


rhelmer:
This will work for the weekend, but we don't want to have to do this twice per week. I'll work on a more permanent solution in the meantime.

KaiRo-at:
r=me as a stopgap measure until next week.

PR #2806, opened 2015-05-22
fixes bug 1167787 - Install browser header for Pingdom in socorroweb

peterbe:
@jdotpz or @rhelmer r?

So this is the just the code. It hinges on the environment variable `PINGDOM_RUM_ID=555f618cabe53d05797ee534` being set.


jdotpz:
yay

rhelmer:
r+, thanks for making it work with `python-decouple` :+1:

PR #2808, opened 2015-05-27
Bug 1152243 Make signature link in supersearch results go to signatur…

Tchanders:
…e report

Also make 'Add term' link for signatures and bring back '+' icons for search terms, but without line breaks this time


webapp-django/crashstats/supersearch/templates/supersearch/search_results.html
```diff
@@ -54,9 +45,20 @@
...
-                    </em>
+                        {% endif %}
+                    {% endif %}
+
+                    {% if column == 'signature' %}
+                    &nbsp;
+                    <a href="{{ current_url }}&amp;{{ column }}=%3D{{ crash[column] }}" class="term" data-field="{{ column }}" data-content="{{ crash[column] }}" title="Add this term to the search form">
+                        Add term
+                    </a>
                     {% endif %}
```
AdrianGaudebert: I would put this block one level up (inside the ``if crash[column]`` block) so that it doesn't show this link for empty signatures. Signatures are never supposed to be empty for us, but it might be for other users.

AdrianGaudebert:
The 'add' icon is missing.

Tchanders:
@AdrianGaudebert Have addressed both of those in the new commit

AdrianGaudebert:
r+

PR #2813, opened 2015-05-28
bug 1162085 - sample config path is wrong

rhelmer:
The path to the uwsgi sockets in the systemd configs was changed in 1f5a4f2f946a662470ef3be077c445e082b0b22e so the path in the nginx sample configs should be too.


phrawzty:
`r+`

PR #2814, opened 2015-05-28
Bug 1169375 Make facet links add terms on supersearch

Tchanders:


Tchanders:
@AdrianGaudebert I think this should fix this bug. I missed the behaviour of these links because I wasn't aware of the multiple facet tabs, so there may be more use cases I'm not aware of...

phrawzty:
Thanks for the contribution, @Tchanders !

PR #2815, opened 2015-05-29
Fixes bug 1152925 - don't get search results on initial load

Tchanders:


webapp-django/crashstats/supersearch/static/supersearch/js/socorro/search.js
```diff
@@ -297,18 +297,22 @@ $(function () {
...
-                // When the form has finished loading, we get sanitized parameters
-                // from it and show the results. This will avoid strange behaviors
-                // that can be caused by manually set parameters, for example.
-                var params = self.getParams();
-                updatePublicApiUrl(params);
-
-                params.page = page;
-                var url = prepareResultsQueryString(params);
-                showResults(resultsURL + url);
+                if (!dontRun) {
```
AdrianGaudebert:
```javascript
if (dontRun) {
    return;
}
// rest of the function
```

Writing it this way avoids a level of indentation, which is always a fine thing to do, while keeping the code as readable, if not more.

Tchanders:
@AdrianGaudebert Should have drawn your attention to this earlier - that change has been made.

PR #2816, opened 2015-05-29
Fixes Bug 1167767 - add statsd storage class to count calls to crashstore

twobraids:
this is a stab at a crashstorage class that will log any call to its method to statsd.


rhelmer:
r+, this will work for getting a count. We should think about making a wrapper class too, so we could easily log how long it takes us to write to the data stores.

rhelmer:
We'll need the datadog package in `requirements.txt`, I'll do that in a separate PR.

PR #2818, opened 2015-05-30
(no bug, docs) - local.py is deprecated, do not suggest modifying it …

rhelmer:
…here. suggest base.py until this is covered by django-decouple


jdotpz:
This is good documentation pruning.  r+

PR #2819, opened 2015-06-01
fixes Bug 1170168 - added statsd benchmarking wrapper

twobraids:
here's an additional class for statsd crashstorage - a benchmarking class  that calls the ```timing``` method.  Any call to a wrapped crashstorage class is timed and the information sent to statsd.


rhelmer:
r+ lgtm - going to land now and test on AWS stage

PR #2820, opened 2015-06-01
(no bug) - crontabber should use envconsul -once

rhelmer:
r? @jdotpz / @peterbe - we don't want envconsul to run as a daemon, by default it listens for changes and will restart if there are config changes to the socorro/common or socorro/crontabber prefixes.

We don't want envconsul to kill crontabber while it is running just to push out config changes, crontabber should finish the current run and pick them up on the next.


peterbe:
Not sure what review here, but you're right, it's important that it gets to finish its work before being killed.

rhelmer:
Good enough @peterbe :+1:

PR #2821, opened 2015-06-01
fixes bug 1164465 - psycopg2 2.6

peterbe:
Going to r someone as soon as I see this working once in travis.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -47,8 +47,15 @@ lxml==3.4.1
 pgxnclient==1.2.1
 # sha256: Zm_VIvShBZ1bi0mW5Xb6eGL1BVKOEc8e9VRjEcYMPk0
 pika==0.9.8
-# sha256: xsMwyYqwjFY58MzX4hG4Kzo58AGBPc6mu4JKmdKzT-I
-psycopg2==2.5.5
+
+# psycopg2 2.6.1 isn't out on pypi at the time of writing and we can't
+# wait so we'll point to the sha for now.
+# Once 2.6.1 on pypi (see
+# https://github.com/psycopg/psycopg2/issues/324)
+# we'll change this line back to `psycopg2==2.6.1`
+# sha256: GY81FHaFAe2lKH7s1bunn6TjV2x0n-v76ndjBDmjFz0
+https://github.com/psycopg/psycopg2/archive/d661652.zip#egg=psycopg2
+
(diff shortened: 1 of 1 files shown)


rhelmer:
This should work, *but* wouldn't fix the deadlock problem we're seeing in AWS load-testing since psycopg2 2.6.1 isn't out yet :(

@peterbe - should we use a git sha until 2.6.1 is released?

peterbe:
Oh. I see! That's problematic. I hate when this happens. But we have little choice don't we. This bug you mention that 2.6.1 fixes, is hurting us actively this week already.

peterbe:
So, according to this, it is a release http://initd.org/psycopg/docs/news.html

peterbe:
The two tickets under 2.6.1 on http://initd.org/psycopg/docs/news.html don't mention https://github.com/psycopg/psycopg2/pull/290

rhelmer:
@peterbe yeah... 2.6.1 isn't on pypi it looks like, but it doesn't contain the fix it won't help us anyway.

rhelmer:
OK so it looks like 2.6.1 *does* contain this fix, but it's not posted on pypi yet.

requirements.txt
```diff
@@ -47,8 +47,15 @@ lxml==3.4.1
 pgxnclient==1.2.1
 # sha256: Zm_VIvShBZ1bi0mW5Xb6eGL1BVKOEc8e9VRjEcYMPk0
 pika==0.9.8
-# sha256: xsMwyYqwjFY58MzX4hG4Kzo58AGBPc6mu4JKmdKzT-I
-psycopg2==2.5.5
+
+# psycopg2 2.6.1 isn't out on pypi at the time of writing and we can't
+# wait so we'll point to the sha for now.
+# Once 2.6.1 on pypi (see
+# https://github.com/psycopg/psycopg2/issues/324)
```
rhelmer: Good idea to comment this

rhelmer:
r+ from me, if it passes travis - I am testing on AWS as we speak as well.

peterbe:
I'm going to rebase this once it finishes (if it finishes) if you don't mind waiting a bit. I don't want that bug to close until we can get off the github sha.

dvarrazzo:
FYI, 2.6.1 is now released on PyPI, thank you for your patience.

PR #2825, opened 2015-06-02
Add a --symbols-tmp argument to stackwalker to set temp directory for symbol download

luser:
r? @rhelmer who ran into this issue when trying to move the symbol cache to local disk on EC2 instances, where /tmp is EBS. We use `rename` to move the temp file to its final location but that doesn't work across filesystems.


rhelmer:
r+

PR #2826, opened 2015-06-02
Fixes Bug 1170168 even more - adjustments to datadog statsd

twobraids:
pave over the differences between datadog statsd and regular statsd


rhelmer:
r+

PR #2827, opened 2015-06-02
bug 1164465 - provide upgraded fork with standard version until new o…

rhelmer:
…ne is released on pypi

r? @peterbe


peterbe:
If travis rss, I'll arse+

PR #2828, opened 2015-06-03
Fixes bug 1066881 - redirect if product is missing and user is logged in

Tchanders:


Tchanders:
@AdrianGaudebert Looking at the bug title, "missing product should redirect to /admin if user is logged-in and superuser" - are the current checks sufficient to determine whether the user is a superuser, or just whether they are logged in?

AdrianGaudebert:
Good catch, it should indeed check if the user is a superuser. Otherwise it'll redirect to the "login" page, which we probably don't want.

webapp-django/crashstats/crashstats/decorators.py
```diff
@@ -43,7 +46,17 @@ def pass_default_context(view):
...
+        try:
+            kwargs['default_context'] = utils.build_default_context(
+                product,
+                versions,
+                user_authenticated
+            )
+        except utils.MissingProductError:
+            return redirect(
+                reverse('manage:products') + '?' +
+                urllib.urlencode({'product': product})
```
AdrianGaudebert: You don't need to build that URL yourself. Take a look at the examples here: https://docs.djangoproject.com/en/1.6/topics/http/shortcuts/#examples You can do something like ``redirect('url', product=product)``. This way you don't need to use urllib, and it's easier to maintain.

webapp-django/crashstats/crashstats/decorators.py
```diff
@@ -43,7 +46,17 @@ def pass_default_context(view):
...
-        kwargs['default_context'] = utils.build_default_context(product,
-                                                                versions)
+        user_authenticated = request.user.is_authenticated()
+        try:
+            kwargs['default_context'] = utils.build_default_context(
+                product,
+                versions,
+                user_authenticated
+            )
+        except utils.MissingProductError:
```
peterbe: Can you not do `except http.Http404:` here?

peterbe: that way `build_default_context` becomes independent of users auth status.

Tchanders: Thanks, I've updated this

Tchanders:
Is this better?

AdrianGaudebert:
The build is failing! But once it's green, you have an r+. :)

Tchanders:
This seems to have worked. Is it good to go?

AdrianGaudebert:
Yes! Do you mind squashing your commits into just one? And then I'll merge.

PR #2829, opened 2015-06-03
No Bug  - corrected getting name of statsd wrapped crashstore

twobraids:
tried to get the name from the wrong object...


rhelmer:
r+

PR #2830, opened 2015-06-04
Fixes bug 1143724 - Used Super Search to power the Reports tab of rep…

AdrianGaudebert:
…ort/list.

@peterbe r?

That's another one of those commits that once were in master but got removed. It's the same content but since I had to do intense conflict resolution, I guess it would be good to give a quick look.

### DIFF
--- a/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/report_list_reports.js
+++ b/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/report_list_reports.js
@@ -111,7 +111,7 @@ Panels.register('reports', function() {
 
 var Columns = (function() {
     var COLUMNS = null;
-    var KEY = 'report-list-report-columns';
+    var KEY = 'report-list-reports-columns';
 
     // New API, new object || Where the polyfill lives
     var storage = navigator.storage || navigator.alsPolyfillStorage;
--- a/webapp-django/crashstats/crashstats/views.py
+++ b/webapp-django/crashstats/crashstats/views.py
@@ -20,7 +20,11 @@ from session_csrf import anonymous_csrf
 
 from . import forms, models, utils
 from .decorators import check_days_parameter, pass_default_context
-from crashstats.supersearch.models import SuperSearchUnredacted
+from crashstats.supersearch.models import (
+    SuperSearch,
+    SuperSearchFields,
+    SuperSearchUnredacted,
+)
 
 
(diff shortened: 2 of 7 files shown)


peterbe:
A "quick look" he says to a 300 line diff :)

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/report_list_reports.js
```diff
@@ -111,7 +111,7 @@ Panels.register('reports', function() {
 
 var Columns = (function() {
     var COLUMNS = null;
-    var KEY = 'report-list-report-columns';
+    var KEY = 'report-list-reports-columns';
```
peterbe: I remember :)

webapp-django/crashstats/crashstats/views.py
```diff
@@ -1382,36 +1445,40 @@ def report_list(request, partial=None, default_context=None):
...
                 break
 
+        api_columns = [x['key'] for x in context['columns']]
+        # The correlations tab requires some keys to always be there. Since
+        # they are low cost, we add them here, to keep the advantages of the
+        # shared cache between the two tabs.
+        api_columns += [
+            'platform', 'version', 'release_channel', 'install_time', 'date'
+        ]
+
```
peterbe: rm

peterbe:
Well, it works :)

What I don't like is that it says "Url" instead of "URL". And "Uuid", "Cpu arch", "Json dump" and "Productid" to mention a few. Also, why does it show "Productid" twice? Is that because `"productid".title() == `"ProductID".title()`?

webapp-django/crashstats/crashstats/views.py
AdrianGaudebert: rm? Why should I remove this?

AdrianGaudebert:
Yeah the fact that it appears twice is due to the fact that we expose the field from the processed crash and from the raw crash. I suppose we could change the Super Search Fields list to not exposed the one in the processed crash?

webapp-django/crashstats/crashstats/views.py
peterbe: oops. my bad. I didn't see that it was actually used.

peterbe:
>  I suppose we could change the Super Search Fields list to not exposed the one in the processed crash? 

+1

AdrianGaudebert:
I will come back to this later!

PR #2831, opened 2015-06-04
fixes Bug 1171475 - refactor statsd for use as rule benchmarking, too

twobraids:
the statsd benchmarking wrapper for crashstorage is very useful.  We need something like it for the TransformRule system, too.  In fact, the crashstore system is so generic, the same wrappers ought to be used for both situations.

Refactor the statsd class for use in both crashstorage and transform rules.

Take Note:  this PR changes the names of some of the configuration requiredments for the statsd classes.  These changes made the code more generic so that it could be used in two different systems without having names of crashstorage system leak into the config of the transform rule system:

```
active_counter_list  --> active_list
wrapped_crashstore --> wrapped_object_class
prefix --> statsd_prefix
```
Existing configuration will have to be updated to reflect these new names. 

Also note that the benchmarking crashstorage class now honors the ```active_list``` in the same manner that the counte class.


rhelmer:
r+

PR #2832, opened 2015-06-04
fixes Bug 1163831 - collapse templates in signatures

twobraids:
normalize all signature that contain templates to this form: 
```
Alpha<Bravo<Charlie>, Delta>::Echo<Foxtrot> --> Alpha<T>::Echo<T>
```
remove old normalization rule from Bug 481445 that normalized templates with integer agruments
```
f<3>(x, y, z) -->  f<T>(x, y, z)
```

### DIFF
--- a/socorro/processor/signature_utilities.py
+++ b/socorro/processor/signature_utilities.py
@@ -84,7 +84,6 @@ class CSignatureToolBase(SignatureTool):
 
         self.fixup_space = re.compile(r' (?=[\*&,])')
         self.fixup_comma = re.compile(r',(?! )')
-        self.fixup_integer = re.compile(r'(<|, )(\d+)([uUlL]?)([^\w])')
 
     #--------------------------------------------------------------------------
     def normalize_signature(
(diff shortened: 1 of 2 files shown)


socorro/processor/signature_utilities.py
```diff
@@ -109,15 +108,30 @@ def normalize_signature(
...
+
             if self.signatures_with_line_numbers_re.match(function):
                 function = "%s:%s" % (function, line)
             # Remove spaces before all stars, ampersands, and commas
             function = self.fixup_space.sub('', function)
             # Ensure a space after commas
             function = self.fixup_comma.sub(', ', function)
-            # normalize template signatures with manifest const integers to
-            #'int': Bug 481445
-            function = self.fixup_integer.sub(r'\1int\4', function)
```
rhelmer: Why is this being removed?

twobraids: this is the removal of the template-with-integer-arguments originally for Bug 481445.  This new method of collapsing templates in signature normalization supersedes that system.

rhelmer:
r+

PR #2834, opened 2015-06-05
Fixes Bug 1124241 - added jit category external program

twobraids:
I've written this in the form of a classifier rule.  This is all it takes to put the stdio results of an external program into a single key in a processed crash.  This is a fine example of a classifier that anyone could write to include in Socorro.  

Placing the output of this external program into the "classifications.jit" namespace within the processed crash, makes it readily available for access by future rules, such as signature manipulation. At the same time this imposes a hierarchical structure on what will likely evolve into many different types of crash classification systems.  Signature itself is just another classifications system that happens to have gotten into Socorro first and so remains entrenched.  I encourage others to write classifiers. A new one could, perhaps, end up being far more useful than the current signature system tracing back to 2007.   

Once this predicate and form of the invocation are accepted, we can enable it at any time by uncommenting the block in ```socorro/processor/mozilla_processor_2015.py``` in this PR.

This PR makes no attempt to ensure the building, loading, provisioning, or verification of the name of the ```jit-crash-categorize``` external program. Collaboration on that is welcome.

### DIFF
--- a/socorro/processor/breakpad_transform_rules.py
+++ b/socorro/processor/breakpad_transform_rules.py
@@ -68,7 +68,6 @@ class BreakpadStackwalkerRule(Rule):
         default=tempfile.gettempdir(),
     )
 
-
     #--------------------------------------------------------------------------
     def __init__(self, config):
         super(BreakpadStackwalkerRule, self).__init__(config)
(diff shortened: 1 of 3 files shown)


luser:
https://github.com/mozilla/socorro/pull/2839 added the jit-crash-categorize code, FYI.

twobraids:
discussion on this topic has been with the consumers of the data in https://bugzilla.mozilla.org/show_bug.cgi?id=1124241  

I realize that there isn't a lot of agreement on how to detect the target crashes for the application of this external tool.  @rhelmer, I suggest putting the PR on hold/withdraw until Q3 and consensus on the predicate algorithm.

rhelmer:
r+

PR #2835, opened 2015-06-05
Fixes Bug 1172511 - examples of CountAnythingRules for the processor

twobraids:
this is a demo of what can be done with the statsd classes and processor rules...  

this change creates a base counter rule class and then uses it to define two processor rules that detect conditions during processing and then increment a statsd counter.  

CountStackWalkerTimeoutKills  will track how many times the stackwalker is killed for taking too long.
CountStackWalkerFailures will track how many times the stackwalker crashes or returns a bad error code.


lonnen:
@AdrianGaudebert could you have a look at these when you return?

rhelmer:
r+ lgtm, these won't be activated until we enable them via config

AdrianGaudebert:
Don't we have unit tests for rules? If not, should we have some? It looks all simple and good, but unit tests are still a nice thing to have imho.

PR #2836, opened 2015-06-05
Fixes Bug 1162703 - added mozilla::ipc::MessageChannel::Send to prefix

twobraids:
add mozilla::ipc::MessageChannel::Send to the prefix list in processor signature generation


twobraids:
r+ selfie for such a minor change.

PR #2837, opened 2015-06-05
fixes Bug 1171437 - added option to remove arguments from signatures

twobraids:
this is an experimental change to signature generation. It removes the arguments from C++ signatures during signature normalization.  We want to try this to see how this affects top signature reports.  It has been made to be turned on/off (default off) via configuration so we can see if it is helpful or harmful.

### DIFF
--- a/socorro/processor/signature_utilities.py
+++ b/socorro/processor/signature_utilities.py
@@ -69,6 +69,13 @@ class CSignatureToolBase(SignatureTool):
     breakpad C/C++ stacks.  It provides a method to normalize signatures
     and then defines its own '_do_generate' method."""
 
+    required_config = Namespace()
+    required_config.add_option(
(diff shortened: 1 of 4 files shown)


twobraids:
@rhelmer this PR can be merged benignly as it is config'd off.  However, there are folks keenly interested in seeing it tried for a day to see how it affects the top crash reports.  Should they determine that the changes are harmful, they'll want it removed, and, I assume, a start a big reprocessing job.

Do we have the bandwidth to do this thing before both Whistler and EOQ ?  Or should be put it off until July?

rhelmer:
@twobraids I think we can squeeze this in - would you mind bringing it up tomorrow? It's not on my near-term radar.

rhelmer:
r+

PR #2839, opened 2015-06-08
bug 1171937 - import jit-crash-categorize code

luser:
r? @rhelmer

I don't think the jit-crash-categorize code itself needs a review, dmajor and I have both been in it and we can probably trade reviews on it in the future.


rhelmer:
r+

PR #2840, opened 2015-06-08
fixes bug 1131987 - symbol upload and wildcard emails for private

peterbe:
@AdrianGaudebert r?

Sorry Adrian if this feels out of context for you. It's just Python. I didn't want to steal Rob's attention.


AdrianGaudebert:
r+

PR #2841, opened 2015-06-08
fixes bug 1172693 - Document how to configure S3 symbol uploads

peterbe:
@rhelmer r?

Note: I make the bug of this PR depend on [1131987](https://bugzilla.mozilla.org/show_bug.cgi?id=1131987) which isn't [resolved yet](https://github.com/mozilla/socorro/pull/2840) but I know once it lands you're going to be able to do the wildcards stuff.

### DIFF
--- a/webapp-django/crashstats/settings/base.py
+++ b/webapp-django/crashstats/settings/base.py
@@ -625,15 +625,29 @@ AWS_SECRET_ACCESS_KEY = config('AWS_SECRET_ACCESS_KEY', '')
 # Information for uploading symbols to S3
 SYMBOLS_BUCKET_DEFAULT_NAME = config('SYMBOLS_BUCKET_DEFAULT_NAME', '')
 
-# To set overriding exceptions by email, use:
-SYMBOLS_BUCKET_EXCEPTIONS = {
-    # e.g.
-    # 'joe.bloggs@example.com': 'private-crashes.my-bucket'
-    # or you can specify it as a tuple of (name, location)
-    # 'joe@example.com': ('my-bucket', 'USWest1')
-    config('SYMBOLS_BUCKET_EXCEPTIONS_USER', ''):
-        config('SYMBOLS_BUCKET_EXCEPTIONS_BUCKET', '')
-}
+# The format for this is `email:bucketname, email2:bucketname, etc`
+SYMBOLS_BUCKET_EXCEPTIONS = config('SYMBOLS_BUCKET_EXCEPTIONS', '', cast=Csv())
+SYMBOLS_BUCKET_EXCEPTIONS = dict(
+    x.strip().split(':', 1) for x in SYMBOLS_BUCKET_EXCEPTIONS
+)
+# We *used* to allow just one single key/value override exceptions
+# we need to continue to support for a little bit.
+if (
+    config('SYMBOLS_BUCKET_EXCEPTIONS_USER', '') and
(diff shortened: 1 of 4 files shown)


rhelmer:
These instructions depend on `local.py` and don't fit with the newer docs. I actually ran into this earlier today - the way we are doing `python-decouple` right now only allows for a single override.

peterbe:
Hmm.... Right you are. The only thing that makes sense is to make it a JSON string. 
E.g. 
```python
import json
SYMBOLS_BUCKET_EXCEPTIONS = json.loads(config('SYMBOLS_BUCKET_EXCEPTIONS', '{}')
```
What do you think about that?

I see, that at the moment we have [this](https://github.com/mozilla/socorro/blob/e0d45869f34db8f05a0e14af761cbd16be140f0a/webapp-django/crashstats/settings/base.py#L616-L623) which will only ever give us the option to override 1 thing.

rhelmer:
@peterbe yeah I have been avoiding sticking JSON into values, although it does seem like a way out :/

One thing that the `python-decouple` folks support and recommend is to use comma-delimited values as lists, so we could perhaps break this down in a way that's more readable as separate keys with lists?

peterbe:
Yeah, my thoughts too actually. 

So, all the code that uses the setting expects it to be a dict but perhaps we can simply do this:
```python
SYMBOLS_BUCKET_EXCEPTIONS = config('SYMBOLS_BUCKET_EXCEPTIONS', Csv())
SYMBOLS_BUCKET_EXCEPTIONS = dict(.... for x in SYMBOLS_BUCKET_EXCEPTIONS)
```
that way, we interface with it as a CSV on the config side. E.g.:

```
export SYMBOLS_BUCKET_EXCEPTIONS=p@mozilla.com:peters-bucket, *@example.com:exampleincs-bucket:us-east-3
```
And the code that depends on it gets a dict. All is well. 

You like? I code. :)

rhelmer:
@peterbe sure that looks fine, note that currently we only have a single exception but we may need to add more someday.

peterbe:
@rhelmer So now it's been fixed. You can now easily specify multiple exceptions with a CSV approach.

peterbe:
@rhelmer ping for a final r?

peterbe:
Rebased and squashed.

webapp-django/crashstats/settings/base.py
```diff
@@ -625,15 +625,29 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
...
-    # e.g.
-    # 'joe.bloggs@example.com': 'private-crashes.my-bucket'
-    # or you can specify it as a tuple of (name, location)
-    # 'joe@example.com': ('my-bucket', 'USWest1')
-    config('SYMBOLS_BUCKET_EXCEPTIONS_USER', ''):
-        config('SYMBOLS_BUCKET_EXCEPTIONS_BUCKET', '')
-}
+# The format for this is `email:bucketname, email2:bucketname, etc`
+SYMBOLS_BUCKET_EXCEPTIONS = config('SYMBOLS_BUCKET_EXCEPTIONS', '', cast=Csv())
+SYMBOLS_BUCKET_EXCEPTIONS = dict(
```
lonnen: Why immediately overwrite this car with a fresh dict?

lonnen: retracted. I see it now.

PR #2842, opened 2015-06-09
Output symbol cache stats in stackwalker output

luser:
This includes disk cache hits/misses as well as elapsed HTTP fetch time for misses. It looks like this:
```json
{
   "modules" : [
      {
         "base_addr" : "0xd50000",
         "code_id" : "54224E6B6000",
         "debug_file" : "plugin-container.pdb",
         "debug_id" : "42331CE9D3A74178A1753DD1D9115F8C2",
         "end_addr" : "0xd56000",
         "filename" : "plugin-container.exe",
         "loaded_symbols" : true,
         "symbol_disk_cache_hit" : false,
         "symbol_fetch_time" : 2344.384033203125,
         "version" : "32.0.3.5379"
      },
      {
         "base_addr" : "0x5e070000",
         "code_id" : "54224D93192e000",
         "debug_file" : "xul.pdb",
         "debug_id" : "553EB6DB78C34A58BE608DB8784798642",
         "end_addr" : "0x5f99e000",
         "filename" : "xul.dll",
         "loaded_symbols" : true,
         "symbol_disk_cache_hit" : true,
         "version" : "32.0.3.5379"
      }
   ]
}
```

We should be able to use this with https://github.com/mozilla/socorro/pull/2835 to get stats on our local disk symbol cache hit rates as well as how much time we're spending fetching symbols via HTTP.

@rhelmer, @twobraids, I don't know if either of you would like to look at this.

### DIFF
--- a/minidump-stackwalk/http_symbol_supplier.cc
+++ b/minidump-stackwalk/http_symbol_supplier.cc
@@ -33,6 +33,7 @@
 
 #include <curl/curl.h>
 #include <sys/stat.h>
+#include <sys/time.h>
 #include <unistd.h>
 
 #include <errno.h>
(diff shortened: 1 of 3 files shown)


twobraids:
r+

PR #2843, opened 2015-06-09
fixes bug 1166068 - Graphics hexes in tooltips

peterbe:
@AdrianGaudebert r?

@KaiRo-at, 
I don't know what you mean by [*"for the first entry in"*](https://bugzilla.mozilla.org/show_bug.cgi?id=1166068#c6) but do you like what you see in the screenshot?

screenshot 2015-06-09 13 35 13


KaiRo-at:
That's exactly what I meant. I was only referring to the first row/adapter in the table given that there are multiple adapters listed. r=me on the result (I can't review the code itself).

PR #2844, opened 2015-06-10
fixes Bug 1158877 - rework the daily_url cron job

twobraids:
making a version much closer to the original, make a version of the daily-url for crontabber
* eliminate the dependency on the deprecated IdCache object
* use the crontabber decorators for external process

### DIFF
--- a/socorro/cron/jobs/daily_url.py
+++ b/socorro/cron/jobs/daily_url.py
@@ -2,9 +2,6 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-###!/usr/bin/python
-
-
 import datetime
 import gzip
 import csv
@@ -97,7 +94,10 @@ def setup_query_parameters(config, day):
             else:
                 prod_phrase = "and r.product = '%s'" % config.product
     except Exception, x:
-        config.logger.error(repr(x), exc_info=True)
+        config.logger.error(
+            'unable to create product phrase for query',
+            exc_info=True
+        )
     ver_phrase = ''
     try:
         if config.version != '':
@@ -107,7 +107,10 @@ def setup_query_parameters(config, day):
             else:
                 ver_phrase = "and r.version = '%s'" % config.version
     except Exception, x:
-        config.logger.error(repr(x), exc_info=True)
+        config.logger.error(
+            'unable to create product phrase for query',
+            exc_info=True
+        )
 
     return DotDict({
         'now_str': now_str,
@@ -135,8 +138,7 @@ def gzipped_csv_files(config, day, gzip=gzip, csv=csv):
         lineterminator='\n'
     )
 
-    pubic_out_filename = ("%s-pub-crashdata.csv.gz"
-                          % day.strftime('%Y%m%d'))
+    public_out_filename = "%s-pub-crashdata.csv.gz" % day.strftime('%Y%m%d')
     public_out_pathname = None
     public_out_directory = config.get('public_output_path')
     public_gzip_file_handle = None
@@ -144,7 +146,7 @@ def gzipped_csv_files(config, day, gzip=gzip, csv=csv):
     if public_out_directory:
         public_out_pathname = os.path.join(
             public_out_directory,
-            pubic_out_filename
+            public_out_filename
         )
         public_gzip_file_handle = gzip.open(public_out_pathname, "w")
         public_csv_file_handle = csv.writer(
@@ -285,7 +287,10 @@ def dailyUrlDump(
                     column_value_list = process_crash(crash_row)
                     write_row(output_context, column_value_list)
     except Exception, x:
-        config.logger.error(repr(x), exc_info=True)
+        config.logger.error(
+            'major problem running dailyUrlDump',
+            exc_info=True
+        )
         raise
 
 
--- a/socorro/cron/jobs/daily_url.py
+++ b/socorro/cron/jobs/daily_url.py
@@ -2,103 +2,329 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import contextlib
-import csv
+###!/usr/bin/python
+
+
 import datetime
 import gzip
+import csv
 import os.path
-import subprocess
+import re
+
+import contextlib
+
 
 from configman import Namespace
-from crontabber.base import BaseCronApp
+
 from crontabber.mixins import (
-    as_backfill_cron_app,
-    with_postgres_transactions,
-    with_single_postgres_transaction
+    using_postgres,
+    with_subprocess,
+    as_backfill_cron_app
 )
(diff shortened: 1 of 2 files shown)


socorro/cron/jobs/daily_url.py
```diff
@@ -2,103 +2,329 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import contextlib
-import csv
+###!/usr/bin/python
```
peterbe: ?

socorro/cron/jobs/daily_url.py
```diff
@@ -2,103 +2,329 @@
...
+        if config.product != '':
+            if ',' in config.product:
+                prod_list = [x.strip() for x in config.product.split(',')]
+                prod_phrase = (
+                    "and r.product in ('%s')" % "','".join(prod_list)
+                )
+            else:
+                prod_phrase = "and r.product = '%s'" % config.product
+    except Exception, x:
+        config.logger.error(repr(x), exc_info=True)
```
peterbe: No need to repeat the exception value. 
`config.logger.error('unable to create product phrase for query', exc_info=True)`

socorro/cron/jobs/daily_url.py
```diff
@@ -2,103 +2,329 @@
...
+    ver_phrase = ''
+    try:
+        if config.version != '':
+            if ',' in config.product:
+                ver_list = [v.strip() for v in config.version.split(',')]
+                ver_phrase = ("and r.version in ('%s')" % "','".join(ver_list))
+            else:
+                ver_phrase = "and r.version = '%s'" % config.version
+    except Exception, x:
+        config.logger.error(repr(x), exc_info=True)
```
peterbe: ditto

socorro/cron/jobs/daily_url.py
```diff
@@ -2,103 +2,329 @@
...
+    config.private_out_pathname = private_out_pathname
+    private_gzip_file_handle = gzip.open(private_out_pathname, "w")
+    private_csv_file_handle = csv.writer(
+        private_gzip_file_handle,
+        delimiter='\t',
+        lineterminator='\n'
+    )
+
+    pubic_out_filename = ("%s-pub-crashdata.csv.gz"
+                          % day.strftime('%Y%m%d'))
```
peterbe: **pubic** :)
(I know it was already like that but now's the chance to fix it)

socorro/cron/jobs/daily_url.py
```diff
@@ -2,103 +2,329 @@
...
+    config.public_out_pathname = public_out_pathname
+    yield (private_csv_file_handle, public_csv_file_handle)
+    private_gzip_file_handle.close()
+    if public_gzip_file_handle:
+        public_gzip_file_handle.close()
+
+
+#------------------------------------------------------------------------------
+linux_line_re = re.compile(
+    r'0\.0\.0 [lL]inux.+[lL]inux$|[0-9.]+.+(i586|i686|sun4u|i86pc|x86_64)?'
```
peterbe: I don't know why but there seems something odd with this. 
```python
>>> bool(linux_line_re.match('0.0.0 linuxxxxxxxxxxxxxx'))
True
```
The `.+` after the first `[lL]inux` means anything can come next. It doesn't have to end with `[lL]inux`. 

Also, the second part (after the `|` sign) doesn't require that it ends with any of those things like i586 or i686 etc. 
E.g.
```python
>>> bool(linux_line_re.match('12'))
True
>>> bool(linux_line_re.match('..'))
True
```

socorro/cron/jobs/daily_url.py
```diff
@@ -2,103 +2,329 @@
...
 
 
+#------------------------------------------------------------------------------
+def get_appropriat_os_version(name, original_version):
+    """
+    If this is a linux os, chop out all the gubbish retaining only the actual
+    version numbers and, if available, the architecture name.
+    """
+    ret = original_version
+    if 'Linux' != name:
```
peterbe: why not `if name.lower() != 'linux':`?

socorro/cron/jobs/daily_url.py
twobraids: yeah, that's weird.

socorro/cron/jobs/daily_url.py
```diff
@@ -2,103 +2,329 @@
...
+            x = ','.join(str(bugid) for bugid in x)
+        if i == 15:  # r.user_comments
+            x = x.replace('\t', ' ')  # per bug 519703
+        if i == 17:  # r.email -- show 'email' if the email is likely useful
+            # per bugs 529431/519703
+            if '@' in x:
+                x = 'yes'
+            else:
+                x = ''
+        if type(x) == str:
```
peterbe: `if isinstance(x, basestring)` in case we want to be compatible with unicode strings.

socorro/cron/jobs/daily_url.py
twobraids: this is code relocated from the IdCache class - as that class is deprecated, I copied this part literally.  if it has a problem, we ought to deal with it in a separate Bug/PR.

socorro/cron/jobs/daily_url.py
```diff
@@ -2,103 +2,329 @@
...
+                    if headers_not_yet_written:
+                        write_row(
+                            output_context,
+                            [x[0] for x in db_cursor.description]
+                        )
+                        headers_not_yet_written = False
+                    column_value_list = process_crash(crash_row)
+                    write_row(output_context, column_value_list)
+    except Exception, x:
+        config.logger.error(repr(x), exc_info=True)
```
peterbe: What's the point of this? Doesn't crontabber handle all error catching?

peterbe: ...and reporting.

socorro/unittest/cron/jobs/test_daily_url.py
```diff
@@ -82,6 +84,7 @@ def test_basic_run_job_no_data(self):
             private = now.strftime('%Y%m%d-crashdata.csv.gz')
             public = now.strftime('%Y%m%d-pub-crashdata.csv.gz')
             ok_(private in os.listdir(self.tempdir))
+            print "DDDDDDD",  os.listdir(self.tempdir)
```
peterbe: debugging

peterbe:
I feel like I could go into a lot more depth here but I also think it's a not very important piece of code that just needs to work (whatever it takes!) till we can convince certain people to use the API. 

r+ with nits

socorro/unittest/cron/jobs/test_daily_url.py
```diff
@@ -107,10 +110,12 @@ def test_run_job_no_data_but_scped(self):
           private_ssh_command='chmod 0640 /var/data/*',
         )
 
-        def comm():
+        def comm(*args, **kwargs):
             # no errors
             return '', ''
 
+        print self.Popen
```
twobraids: Remove debug

twobraids:
withdrawn as they now say they no longer need it.

PR #2846, opened 2015-06-10
Fixes Bug 1130960 - added new RE to prefix list for UTF signatures

twobraids:
added "(NS_)?(Lossy)?(Copy|Append|Convert).*UTF.*" to the prefix skip list


rhelmer:
r+

PR #2848, opened 2015-06-10
No Bug - corrected platform to os_name

twobraids:
stupidly used "platform" instead of "os_name" to detect windows.


rhelmer:
r+

PR #2854, opened 2015-06-13
Fixes bug 1171968 Show individual loaders in Aggregations tab

Tchanders:


Tchanders:
FAO @AdrianGaudebert

AdrianGaudebert:
This looks and works really good! 

I think there should be some logic about the previous way of showing the loader that could be removed, can you make sure of that?

AdrianGaudebert:
I was wrong, there's is nothing to remove. r+!

PR #2856, opened 2015-06-15
fixes bug 1174822 - Be less picky about archive files' mimetype

peterbe:
@rhelmer r? 

cc @luser 

Rob, 
We have an incident where a contact at Adobe is unable to upload a zip file he/she sends. Our webapp constantly rejects it because when our Django first looks up its mimetype it finds that it is "application/octet-stream" which isn't [one of the allowed ones](https://github.com/mozilla/socorro/blob/b576425da157047b184b76d5ffef8c92a53d3d22/webapp-django/crashstats/symbols/forms.py#L9-L11) and thus it never even bothers trying to unzip it. 
When I tried upload the exact same file from my laptop into my localhost webapp (and uploading to my personal S3 account) it always works. 
So I suspect that there's something strange with out Windows packed that file that our python 2.6 on our RedHat production server interrogates it. After all, django relies on the built-in mimetypes lib which I think uses some OS routines and clearly there's a clash going on there. 

In this patch I remove the pre-checking of the mimetype of the uploaded file and just look at the file name extension. And hence just try to unpack it and if it yields a `BadZipfile` exception it turns that into a `400 Bad Request` error back to the client. 

This is tested by me creating a file called `tests/notazipdespiteitsname.zip` which is just a text file but does get sent to `unzip` which of course fails and hence the `400 Bad Request` in the test.


rhelmer:
r+, yeah I guess we can't assume the user agent will get this right

peterbe:
> r+, yeah I guess we can't assume the user agent will get this right

I don't think it's the user agent really. I think it's our Red Hat server. After all, the file Piyush sent me was the same that got the wrong mimetype on prod. But on my laptop the mimetype became `application/zip`.

PR #2858, opened 2015-06-16
bug 1171937 - copy other stackwalker binaries

rhelmer:
r? @twobraids


twobraids:
r+   

I have nothing more to add to this, as it appears that I already did that last week.

PR #2859, opened 2015-06-17
Fixes bug 1171975 - Port graph tab from report/list/

Tchanders:


Tchanders:
Work in progress, for @AdrianGaudebert

webapp-django/crashstats/signature/templates/signature/signature_graph.html
```diff
@@ -0,0 +1,8 @@
+<div class="panel">
+	<header class="title">
+		<h2>ADU by signature for {{ form.channel }} channel</h2>
+	</header>
+	<div class="body">
+    	<div id="adu-graph"></div>
```
AdrianGaudebert: indent

webapp-django/crashstats/signature/static/signature/js/signature_report.js
```diff
@@ -334,6 +334,186 @@ $(function () {
...
+        }
+
+        function showGraph(graphOptions) {
+            // Remove previous results and show loader.
+            contentElt.empty();
+            addLoaderToElt(contentElt);
+
+            // Add product and signature to graphOptions
+            var params = getParamsWithSignature();
+            graphOptions.product_name = params.product[0];
```
AdrianGaudebert: ``product`` is not a mandatory parameter, that can sometimes be empty.

webapp-django/crashstats/signature/templates/signature/signature_report.html
```diff
@@ -135,6 +141,41 @@
...
+        Displays a graph of crashes/ADU against build date
+        TODO: Clean up classes and labels
+        TODO: Sort out datepicker (dates currenty filled with dummy data)
+        TODO: pass product through
+    -->
+    <section class="panel" id="graph-panel"
+        data-source-url="{{ url('signature:signature_graph') }}"
+        data-jsonurl="{{ url('crashstats:adu_by_signature_json') }}"
+        data-signature="{{ signature }}"
+        data-product="WaterWolf"
```
AdrianGaudebert: You can remove ``data-signature`` and ``data-signature``.

webapp-django/crashstats/signature/templates/signature/signature_report.html
```diff
@@ -135,6 +141,41 @@
...
 
+    <!-- Graph panel.
+        Displays a graph of crashes/ADU against build date
+        TODO: Clean up classes and labels
+        TODO: Sort out datepicker (dates currenty filled with dummy data)
+        TODO: pass product through
+    -->
+    <section class="panel" id="graph-panel"
+        data-source-url="{{ url('signature:signature_graph') }}"
+        data-jsonurl="{{ url('crashstats:adu_by_signature_json') }}"
```
AdrianGaudebert: If possible use the public API endpoint instead.

webapp-django/crashstats/signature/static/signature/js/signature_report.js
```diff
@@ -334,6 +334,180 @@ $(function () {
...
+        var contentElt = $('.content', graphPanel);
+        var loaderElt = $('.loader', graphPanel);
+        var selectElt = $('.channels-list', graphPanel);
+
+        // Get urls from graphPanel
+        var dataUrl = graphPanel.data('source-url');
+        var jsonUrl = graphPanel.data('jsonurl');
+
+        // Set up default params
+        var defaultChannel = 'nightly';
```
AdrianGaudebert: That should not be hard-coded. Either use the already existing ``settings.CHANNELS`` (for example using ``settings.CHANNELS[0]``) or add a new setting key to hold that piece of data.

webapp-django/crashstats/signature/static/signature/js/signature_report.js
```diff
@@ -334,6 +334,180 @@ $(function () {
...
+                    value: ''
+                });
+                // Focus the value input of the new fieldset
+                $('fieldset:last-child .value').select2('open');
+                // Prompt the user to choose a product
+                contentElt.empty().text('Please select a product.');
+                return;
+            }
+
+            // Get graph panel -- this might not be necessary
```
AdrianGaudebert: Yeah there's no point getting the graph panel if it's just a simple template that won't change depending on the data. You should instead include it in the global template of that page, and only query for the graph data.

webapp-django/crashstats/signature/templates/signature/signature_report.html
```diff
@@ -135,6 +141,34 @@
...
+    >
+        <header class="title">
+            <h2>Graph</h2>
+        </header>
+
+        <div class="body">
+            <div class="controls">
+                <select class="channels-list">
+                    <option></option>
+                    {# TODO: find out where channels come from #}
```
AdrianGaudebert: I guess you can remove this comment now. :)

webapp-django/crashstats/signature/templates/signature/signature_report.html
```diff
@@ -135,6 +141,34 @@
         </div>
     </section>
 
+    <!-- Graph panel.
+        Displays a graph of crashes/ADU against build date
+    -->
+    <section class="panel" id="graph-panel"
+        data-source-url="{{ url('signature:signature_graph') }}"
+        data-jsonurl="{{ url('signature:signature_report') }}graphdata/"
```
AdrianGaudebert: There should be an actual name to get that URL. Adding ``graphdata/`` at the end is bad practice.

AdrianGaudebert: The name you defined in ``urls.py`` is ``signature_graph_data``.

webapp-django/crashstats/signature/views.py
```diff
@@ -296,3 +301,66 @@ def signature_comments(request):
     data['query'] = search_results
 
     return render(request, 'signature/signature_comments.html', data)
+
+
+@waffle_switch('signature-report')
+def signature_graph(request):
+    '''Return a graph of crashes/ADU against build date. '''
+
+    return render(request, 'signature/signature_graph.html')
```
AdrianGaudebert: Remove this and instead include it in the ``signature_report`` template, that will save an unnecessary HTTP request.

webapp-django/crashstats/signature/templates/signature/signature_report.html
Tchanders: Just remembered why I did this: url('signature:signature_graph_data') didn't work, so I looked at the aggregations view - which is the other view where a second argument is passed through in addition to the request - and copied that pattern. Is there a better way to do it?

AdrianGaudebert: Aha, alright, my bad then! =)

webapp-django/crashstats/signature/templates/signature/signature_report.html
```diff
@@ -145,8 +184,12 @@
 <script>
 var BASE_URL = location.protocol + '//' + location.host;
 var FIELDS = {{ fields | json_dumps }};
+var DEFAULT_CHANNEL = {{ channels[3] | json_dumps }};
```
AdrianGaudebert: This is bad: if a user changes the configuration, you cannot be assured that ``channels[3]`` is ``nightly``. You'll need to create a different settings option to have that data. 

(In a previous comment I said you could use ``channels[0]``. That's fine to do because we often consider that the first element of a list is the default value. However here it seems ``nightly`` is not the first element, and should not be, so it was just a bad suggestion on my part. )

Tchanders: Should it just go back to 'nightly' then?

webapp-django/crashstats/signature/templates/signature/signature_report.html
```diff
@@ -145,8 +184,12 @@
 <script>
 var BASE_URL = location.protocol + '//' + location.host;
 var FIELDS = {{ fields | json_dumps }};
+var DEFAULT_CHANNEL = {{ channel | json_dumps }};
```
AdrianGaudebert: Not a big deal but there's no need to using ``json_dumps`` here, since it's a simple string. You can just use ``... = '{{ channel }}';``.

AdrianGaudebert:
I see 2 points per date in the new version, and only one in the old one. Is that expected? Also, in this example, the graph doesn't show the "June 26" date. 

screenshot - 02072015 - 10 56 22

AdrianGaudebert:
Also, the new graph doesn't say what the build id is. That's a regression from the old one! :)

Tchanders:
Ah yes I wanted to ask you about the x axis - it's to do with the library. By default, the graph looks like this:
selection_011
I extended the x axis to the left using a setting (https://github.com/mozilla/metrics-graphics/wiki/Axes#xax_start_at_min) - but it seems this means you lose the first tick:
selection_012

Do you think the default is better?

webapp-django/crashstats/signature/static/signature/css/signature_report.less
```diff
@@ -115,3 +115,11 @@ a.crash-id {
...
     }
 }
+
+#graph-panel {
+    .controls {
+        .channels-list {
+            width: 30%;
+        }
+    }
+}
```
AdrianGaudebert: You still have a newline missing here.

AdrianGaudebert:
Regarding the x axis of the graph, that looks like a bug. Let's keep as is right now and then ask the metrics-graphics team to help us solve it.

webapp-django/crashstats/signature/templates/signature/signature_report.html
```diff
@@ -5,8 +5,13 @@
     {{ super() }}
     {% compress css %}
 <link href="{{ static('crashstats/js/lib/select2/select2.css') }}" type="text/css" rel="stylesheet">
+<link href="{{ static('crashstats/css/lib/metricsgraphics.css') }}" type="text/css" rel="stylesheet">
     {% endcompress %}
 
+<link rel="stylesheet" type="text/css" href="{{ static('crashstats/css/lib/jquery-ui.css') }}" media="screen" />
+<link rel="stylesheet" type="text/css" href="{{ static('crashstats/css/lib/jquery-ui.structure.css') }}" media="screen" />
+<link rel="stylesheet" type="text/css" href="{{ static('crashstats/css/lib/jquery-ui.theme.css') }}" media="screen" />
```
AdrianGaudebert: Are those still needed? I think you included them for the date input, but it's no longer there, so we can probably avoid included those css files, right?

Tchanders: Oops yes you're right - have fixed that.

PR #2860, opened 2015-06-18
fixes bug 1179456 - a ES bulk load within crashstorage context

twobraids:
we'd like to try the ES bulk load to see how it would work in a crash storage class.

@AdrianGaudebert  - I suspect that you're more familiar with the elasticsearch python API than the rest of us.  could you take a look at this to see if the implementation makes any sense?  The form of the call to ```helpers.bulk``` was very confusing and I have no idea if I have done it correctly.

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -3,7 +3,6 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import elasticsearch
-from esalsticsearch import helpers
 
 from socorro.external.crashstorage_base import CrashStorageBase
 from socorro.external.es.index_creator import IndexCreator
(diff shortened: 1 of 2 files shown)


socorro/external/es/crashstorage.py
```diff
@@ -214,3 +215,83 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
...
+
+    #--------------------------------------------------------------------------
+    def close(self):
+        self.task_queue.put(None)
+        self.consuming_thread.join()
+
+    #--------------------------------------------------------------------------
+    def _consuming_thread_func(self):  # execute the bulk load
+        es = self.es_context()
+        helpers.bulk(
```
twobraids: @AdrianGaudebert  this is the line that I really like you to take a look at

socorro/external/es/crashstorage.py
```diff
@@ -3,6 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import elasticsearch
+from esalsticsearch import helpers
```
AdrianGaudebert: typo

socorro/external/es/crashstorage.py
```diff
@@ -208,9 +211,182 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
...
+            # First create a datetime object from the string in the crash
+            # report.
+            crash_date = datetimeutil.string_to_datetime(
+                crash_document['processed_crash']['date_processed']
+            )
+            # Then convert it back to a string with the expected formatting.
+            crash_date_with_t = datetimeutil.date_to_string(crash_date)
+            # Finally, re-insert that string back into the report for indexing.
+            crash_document['processed_crash']['date_processed'] = \
+                crash_date_with_t
```
AdrianGaudebert: Can you not use ``reconstitute_datetimes`` here instead?

socorro/external/es/crashstorage.py
```diff
@@ -208,9 +211,182 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
...
+        def _consumer_iter(self):
+            try:
+                while True:
+                    crash_document = self.task_queue.get()
+                    if crash_document is None:
+                        self.done = True
+                        break
+                    try:
+                        yield crash_document  # execute the task
+                    except Exception:
```
AdrianGaudebert: Should this not capture more specific exceptions? I suppose you want to catch problems coming from the ``elasticsearch`` library, so that would be ``elasticsearch.exceptions.ElasticsearchException``.

socorro/external/es/crashstorage.py
```diff
@@ -208,9 +211,182 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
...
+                        self.config.logger.error(
+                            "Error in processing a job",
+                            exc_info=True
+                        )
+            except Exception:
```
AdrianGaudebert: Same thing?

AdrianGaudebert:
r+ when my few questions are answered or addressed. :sunny:

PR #2861, opened 2015-06-18
Revert "Fixes bug 1174169 - Fixed magic beta filter for report/list/."

rhelmer:
This reverts commit 498f453772e276be2ddc54a41f3b5060bd1136df.

This is causing /report/index to not show stack traces on prod, and the "raw dumps" tab on that same page to be blank. (/cc @AdrianGaudebert )

r? @peterbe

### DIFF
--- a/socorro/external/postgresql/base.py
+++ b/socorro/external/postgresql/base.py
@@ -412,6 +412,9 @@ class PostgreSQLBase(object):
             if channel.startswith(tuple(config.non_release_channels)):
                 # This is a non-release channel.
 
+                # Use major_version instead of full version.
+                sql_params[version_param] = version_data["major_version"]
+
                 # Restrict by release_channel.
                 version_where.append("r.release_channel ILIKE '%s'" % channel)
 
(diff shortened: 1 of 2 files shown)


peterbe:
r+

peterbe:
I'm hesitant. What if reverting that isn't related? 

I moved the above mentioned crash from prod to stage and it appears correctly when I load it up locally.
That's before applying this revert (e.g. running off master).

peterbe:
The "Raw Dump" tab depends on there being a `json_dump` or a `dump` in the `UnredactedCrash` output:
And clearly there are no such keys [in the unredacted crashes](https://crash-stats.mocotoolsprod.net/api/UnredactedCrash/?crash_id=40e903b3-c251-4f7e-b2df-0de7e2150618&datatype=unredacted)

Much due to sheer ignorance of how this beta mess works, I honestly don't understand the code but this revert seems to be able selecting a recordset of stuff. Not about the **content** of the unredacted crash.

peterbe:
For what it's worth, when I compare https://crash-stats.mocotoolsprod.net/api/UnredactedCrash/?crash_id=40e903b3-c251-4f7e-b2df-0de7e2150618&datatype=unredacted with http://socorro.dev/api/UnredactedCrash/?crash_id=40e903b3-c251-4f7e-b2df-0de7e2150618&datatype=unredacted (my local instance running off the stage PG) the results are totally different. I moved the crash over with that little `./crash_prod_to_stage.py` script which is black magic to me. 

The one moved over to stage [looks like this](https://gist.github.com/peterbe/094e32620244fc9d2c17)

rhelmer:
I tested reverting this on AWS prod and it works :)

PR #2862, opened 2015-06-18
Revert "Revert "Fixes bug 1174169 - Fixed magic beta filter for repor…

rhelmer:
…t/list/.""

This reverts commit 7d183d541cafb4755b7eb3d19c006b1fa8e8c3c1.

@peterbe r? - I was wrong before about this causing the error we were seeing on production - that was a totally unrelated config change.


peterbe:
r+

PR #2863, opened 2015-06-21
fix bug 1176543 - collector should handle gzipped form posts

rhelmer:
r? @twobraids / @peterbe 

I am not sure that nginx will be able to handle gzipped HTTP POST as Apache did for us in the past, and we have at least Fennec clients out there sending these.

Unfortunately I wasn't able to reuse the multipart form parser `web.py` uses since they don't expose it in a useful way AFAICT, it's basically just a wrapper around `cgi.FieldStorage` using the `web.utils.storage` class though.

Besides the unit test, I have tried this on an EC2 instance with nginx+uwsgi, and it seems to work as expected with the client in https://gist.github.com/rhelmer/c9e3c861e3844d9cb142 (which you'll notice is very similar to the unit test)

### DIFF
--- a/socorro/collector/wsgi_breakpad_collector.py
+++ b/socorro/collector/wsgi_breakpad_collector.py
@@ -8,6 +8,8 @@ import zlib
 import cgi
 import cStringIO
 
+from contextlib import closing
+
 from socorro.lib.ooid import createNewOoid
 from socorro.lib.util import DotDict
 from socorro.collector.throttler import DISCARD, IGNORE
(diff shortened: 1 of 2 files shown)


socorro/collector/wsgi_breakpad_collector.py
```diff
@@ -84,8 +87,32 @@ def _make_raw_crash_and_dumps(self, form):
...
+        if web.ctx.env.get('HTTP_CONTENT_ENCODING') == 'gzip':
+            gzip_header = 16 + zlib.MAX_WBITS
+            data = zlib.decompress(web.webapi.data(), gzip_header)
+            fp = cStringIO.StringIO(data)
+            e = web.ctx.env.copy()
+
+            # this is how web.webapi.rawinput() handles
+            # multipart/form-data, as of this writing
+            fs = cgi.FieldStorage(fp=fp, environ=e, keep_blank_values=1)
+            fsdict = dict([(k, fs[k]) for k in fs.keys()])
```
twobraids: this code creates a dict on this line, only to tear it apart and make another dict out of it on line 108 (```web.utils.storage``` is a dict).

socorro/collector/wsgi_breakpad_collector.py
```diff
@@ -84,8 +87,32 @@ def _make_raw_crash_and_dumps(self, form):
...
+            def process_fieldstorage(fs):
```
twobraids: this method is deceptive.  Embedded here it looks like a closure.  However, the variable ```fs``` is a parameter rather than the ```fs``` value from the outer scope.

twobraids:
I'm assuming that the additional code is actually a cut and paste from another source.  It just doesn't look like what I've come to know as your style.  I've posted a PR to this branch in your repo with my verison of this same thing with what I believe is a clearer organization of the GZIP transformation.  

Accept or reject as at your will.  Since it works, I'll give this a begrudging r+, but if you take my changes, i'd give out an r++

rhelmer:
@twobraids right, the bit to parse and wrap the multipart form was selectively copied from https://github.com/webpy/webpy/blob/32c891c64bd838ad972239f81254fdfb0c784bf4/web/webapi.py#L279-L323

I like your version more. I worry a little about divergence, but I think it's probably not an issue there.

To really solve my concern, the right thing would not be to copy/paste but instead to extend web.py so their parser could be called from outside code. That's not trivial to do unfortunately, and no guarantee that they will accept.

All that said, I think your version is short and clear enough that we can keep it up to date. Our unit tests should break if something changes on their side.

PR #2865, opened 2015-06-23
Fixes bug 1165869 - Run the elasticsearch cleanup cron job weekly.

AdrianGaudebert:
anyone r? (one liner, config change)


phrawzty:
`r+`

For reference:
```
15:11:13 < adrian> phrawzty: it was related to PHX, but then I think it actually makes
                   sense to run it more regularly, we still want to store as few data as
                   possible in AWS
```

PR #2870, opened 2015-06-26
fixes Bug 1177715 - added missing tests, and associated corrections

twobraids:
the tests for the JIT classifier were written to a file that was never checked in to git, nor did its name start with the word "test".  The tests were never run, I didn't notice, and interpreted silence as success.  Once tests were put into the proper file, some obvious defects were found and then corrected.


rhelmer:
r+

PR #2872, opened 2015-06-29
fixes bug 1177131 - Store symbols with gzip compression

peterbe:
@rhelmer r?
CC @luser 

Ted, so to test this I used my personal S3 account. But because the only zip file I had available was one I was given from Adobe I can't upload it and have you view it because it's a private one. 
If you can give me a (public) zip file I'll upload it locally and then share the URLs with you.


peterbe:
Note-to-reviewer (Rob): Perhaps we wait till Ted and I can do a manual test from my laptop.

rhelmer:
@peterbe - @luser just pointed out that we only want to compress `.sym` files, `.pdb` files are already compressed.

luser:
Yes, clients will be uploading native debug symbols which are all already compressed. (.pd_, .gz, .tar.bz2). I'll find you some useful test data shortly.

peterbe:
So, I'll whitelist the types that *should* be compressed and all else leave as is.

luser:
@peterbe : grab the -crashreporter-symbols-full.zip packages from the subdirs here:
http://ftp.mozilla.org/pub/mozilla.org/firefox/try-builds/tmielczarek@mozilla.com-9cc4ba78daca/

Those are exactly what our build machines upload.

peterbe:
@luser Ok. I just uploaded the latest mac zip file here: https://s3.amazonaws.com/peterbe.org.allizom.crash-stats.staging.crashes/v1/firefox-41.0a1-Darwin-20150630043537-x86-macosx64-symbols.txt
The whole directory is publicly available.

peterbe:
@rhelmer when reviewed, let me rebase to get rid of the breaking commit as that would potentially destroy git bisect'isms.

luser:
I mentioned this on IRC, but you should really try uploading the * -crashreporter-symbols-full.zip* packages, since those are what our build machines upload, and they contain non-.sym files (the native debug symbols).

luser:
I tested by downloading the Win32 build from that try push, crashing it locally with "crash me now", then running a local build of the socorro stackwalker, pointing its --http-symbols-path at peterbe's s3 bucket, and the symbols look fine:
```json
         {
            "frame" : 1,
            "function" : "ffi_call_win32",
            "function_offset" : "0x34",
            "module" : "xul.dll",
            "module_offset" : "0x19e7fe4",
            "offset" : "0x10a87fe4",
            "trust" : "frame_pointer"
         },
```

The commandline looked like `stackwalker --pretty --symbols-url=https://s3.amazonaws.com/peterbe.org.allizom.crash-stats.staging.crashes/v1/ --symbols-cache=/tmp/symbols ~/12375637-7171-400a-bbef-8ea975efcea4.dmp` FWIW.

peterbe:
In other words, we just need @rhelmer's code blessing?

luser:
@rhelmer : code review ping?

rhelmer:
r+

PR #2873, opened 2015-06-30
re-fixes Bug 1171437 - added exceptions to args collapsing

twobraids:
discovered a case where the simplistic approach of removing stuff between parens was the wrong thing.  There exists a signature fragment of this form:   .... (anonymous namespace)::

added code to prevent that example from being collapsed.

### DIFF
--- a/socorro/processor/signature_utilities.py
+++ b/socorro/processor/signature_utilities.py
@@ -93,6 +93,65 @@ class CSignatureToolBase(SignatureTool):
         self.fixup_comma = re.compile(r',(?! )')
 
     #--------------------------------------------------------------------------
+    @staticmethod
+    def _is_exception(
+        exception_list,
+        remaining_original_line,
+        line_up_to_current_position
+    ):
+        for an_exception in exception_list:
+            if remaining_original_line.startswith(an_exception):
+                return True
+            if line_up_to_current_position.endswith(an_exception):
+                return True
+        return False
+
+    #--------------------------------------------------------------------------
+    def _collapse(
+        self,
+        function_signature_str,
+        open_string,
+        replacement_open_string,
+        close_string,
+        replacement_close_string,
+        exception_substring_list=(),  # list of exceptions that shouldn't collapse
+    ):
+        """this method takes a string representing a C/C++ function signature
+        and replaces anything between to possibly nested delimiters"""
+        target_counter = 0
+        collapsed_list = []
+        exception_mode = False
+
+        def append_if_not_in_collapse_mode(a_character):
+            if not target_counter:
+                collapsed_list.append(a_character)
+
+        for index, a_character in enumerate(function_signature_str):
+            if a_character == open_string:
+                if self._is_exception(
+                    exception_substring_list,
+                    function_signature_str[index + 1:],
+                    function_signature_str[:index]
+                ):
+                    exception_mode = True
+                    append_if_not_in_collapse_mode(a_character)
+                    continue
+                append_if_not_in_collapse_mode(replacement_open_string)
+                target_counter += 1
+            elif a_character == close_string:
+                if exception_mode:
+                    append_if_not_in_collapse_mode(a_character)
+                    exception_mode = False
+                else:
+                    target_counter -= 1
+                    append_if_not_in_collapse_mode(replacement_close_string)
+            else:
+                append_if_not_in_collapse_mode(a_character)
+
+        edited_function = ''.join(collapsed_list)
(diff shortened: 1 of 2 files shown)


peterbe:
reviewer?

socorro/unittest/processor/test_signature_utilities.py
```diff
@@ -50,6 +50,7 @@ def setup_config_C_sig_tool(
         pr='pre1|pre2',
         si='fnNeedNumber',
         ss=('sentinel', ('sentinel2', lambda x: 'ff' in x)),
+        ellipsis_arguments=False
```
peterbe: Why is this an argument? It's never set to true. The `config.collapse_arguments` is hardcoded to True.

socorro/processor/signature_utilities.py
```diff
@@ -93,6 +93,63 @@ def __init__(self, config, quit_check_callback=None):
         self.fixup_comma = re.compile(r',(?! )')
 
     #--------------------------------------------------------------------------
+    def _is_exception(
+        self,
+        exception_list,
+        remaining_original_line,
+        line_up_to_current_position
+    ):
```
peterbe: Why is this a class method?
Seems it's only used from within the loop in `_collapse`

socorro/unittest/processor/test_signature_utilities.py
twobraids: artifact from an earlier version.  all mentions of ```ellipsis_arguments``` are now being removed.

socorro/processor/signature_utilities.py
```diff
@@ -93,6 +93,63 @@ def __init__(self, config, quit_check_callback=None):
...
+                    exception_mode = False
+                else:
+                    target_counter -= 1
+                    append_if_not_in_collapse_mode(replacement_close_string)
+            else:
+                append_if_not_in_collapse_mode(a_character)
+
+        edited_function = ''.join(collapsed_list)
+        return edited_function
+
```
peterbe: I can't put my finger on exactly where it goes wrong but this code doesn't protect against unbalanced `open_string` and `close_string`. 
E.g. `_collapse('less<more', '<', '{', '>', 'T}') == 'less{'`
Is that a problem?

socorro/processor/signature_utilities.py
twobraids: to define that method within ```_collapse``` would imply that it is a closure like the other method ```append_if_not_in_collapse_mode```.  In addition, to define it within ```_collapse``` makes unpleasant depth of indentation in a method that already goes uncomfortably deep.

it could though be defined as a ```staticmethod```, but that doesn't really accomplish much important.

socorro/processor/signature_utilities.py
twobraids: as this input comes from the C++ compiler/linker, anything unbalanced would be indicative of a serious complier bug that would not likely have even produced executable code.

socorro/processor/signature_utilities.py
```diff
@@ -93,6 +93,63 @@ def __init__(self, config, quit_check_callback=None):
...
+    #--------------------------------------------------------------------------
+    def _collapse(
+        self,
+        function,
+        open_string,
+        replacement_open_string,
+        close_string,
+        replacement_close_string,
+        exception_substring_list=(),  # list of exceptions that shouldn't collapse
+    ):
```
peterbe: Even though the code is pretty straight forward, now that I've studied it for a while, it's not clear what it does. The name reveals a lot which is a good start. 
How about this for a doc string. 
```
Return a new string where certain enclosed blocks can be replaced. For example
you can turn `somestring[stuff]` into `somestring[]` or `somestring{replacement}`.
```

socorro/processor/signature_utilities.py
```diff
@@ -93,6 +93,63 @@ def __init__(self, config, quit_check_callback=None):
...
+            if remaining_original_line.startswith(an_exception):
+                return True
+            if line_up_to_current_position.endswith(an_exception):
+                return True
+        return False
+
+    #--------------------------------------------------------------------------
+    def _collapse(
+        self,
+        function,
```
peterbe: This put me off. Once I realized that it has nothing to do with callable functions, it made more sense. It should maybe be called "string".

peterbe:
r+ with nits. 
In particular I would prefer if functions that have nothing to do with the object are either utility functions or kept in the class itself but marked with `@staticmethod`.

PR #2874, opened 2015-07-02
No Bug - change default of the boto 'prefix' to empty string

twobraids:
change default of the boto 'prefix' to empty string


rhelmer:
This works around https://github.com/gmr/consulate/issues/51 but is the right thing to do anyway (we decided not to use prefixes in this way)

rhelmer:
r+

PR #2877, opened 2015-07-08
Bug 1181634 - Update forced versions for correlations for Firefox cycle starting 2015-06-29

KaiRo-at:
This adds current aurora and nightly versions as well as all planned beta versions for this cycle.

### DIFF
--- a/scripts/crons/cron_libraries.sh
+++ b/scripts/crons/cron_libraries.sh
@@ -70,7 +70,7 @@ do
   techo "Phase 1: end"
 done
 
-MANUAL_VERSION_OVERRIDE="38.0.5 39.0b5 39.0b6 39.0b7 39.0b4 40.0a2 41.0a1"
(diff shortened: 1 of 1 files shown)


rhelmer:
@jp btw this cronjob lives on the socorroanalysis server now, which is a long-running instance that isn't part of the normal release process... we haven't worked out a good way to automatically update this yet, for now I'd suggest just logging in and updating the RPM (don't kill the instance if you can help it!)

lonnen:
Cool. Let's do this.

lonnen:
:fire:

rhelmer:
@lonnen @jdotpz IIRC the `analysis` node is excluded from production pushes (it's a long-running service like the `admin` node ES, Consul, symbol-api, etc.) so you'll need to update the RPM on that node to make this change live.

We *can and should* make this node able to handle updates, it was just one of the last things to move over and didn't get as much love as the primary services. The primary challenge is that it has an EBS volume, and it has long-running jobs, so we can't safely scale up more.

PR #2887, opened 2015-07-09
2. Rebort! "Revert "fixes bug 1180850 - Crash Me Now for the webapp""

lonnen:
This was reverted to help investigate a production issue that crept in shortly after AWS before we had our monitoring in order. The patch is r+, but we will be relanding and shipping each of these PRs one at a time on July 9 to attempt to find the problematic patch.


jdotpz:
Things are looking good on prod.

screenshot 2015-07-09 13 40 32

PR #2889, opened 2015-07-09
4. Rebort! "Revert "Fixes bug 1171975 - Port graph tab from report/list/""

lonnen:
This was reverted to help investigate a production issue that crept in shortly after AWS before we had our monitoring in order. The patch is r+, but we will be relanding and shipping each of these PRs one at a time on July 9 to attempt to find the problematic patch.


jdotpz:
This is looking good in prod.
screenshot 2015-07-09 16 57 24

PR #2893, opened 2015-07-09
Fixes Bug 1182069 - Redraw GC Crashes graph using MetricsGraphics

Tchanders:


Tchanders:
@AdrianGaudebert How does this look?

AdrianGaudebert:
Code looks good to me! However I find the graph a bit more difficult to read than the previous one. I would make texts a bit darker if possible, and probably slightly bigger as well. 

Also, why does the y axes stop before the highest value? For example if I have a 70k value, the graph will only go up to 60k. That's a bit weird to me.

Tchanders:
The y axis thing is because the axes automatically cover the the range of auto-generated ticks (as discussed in this issue: https://github.com/mozilla/metrics-graphics/issues/419). You can override it, but since our data keeps changing it's hard to predict values that look right in every case... So it might be a choice between having the axis too short sometimes or having the axis too long sometimes.

Regarding the labels, if we make them darker here will the graphs start to look less consistent again? How about rotating them to make them easier to read? Here's what both cases look like:

screenshot from 2015-07-20 13 48 29
screenshot from 2015-07-20 13 44 05

AdrianGaudebert:
Gotcha for the axes. Regarding colors and consistency: how about doing both things? And apply that to all other graphs to stay consistent? :)

Tchanders:
@AdrianGaudebert How's this?

AdrianGaudebert:
r+

PR #2895, opened 2015-07-10
WIP Redraw crashes per ADI using metrics graphics

Tchanders:


AdrianGaudebert:
Looks good to me! The bug in MG with the x axes not showing the left-most value is annoying, but nothing we can do about. The colors would need to be fixed to stay consistent with the current set. 

Also, is it possible to remove the old plot library now, or is it still used by something else?

Tchanders:
I've changed the colours by overriding the metrics graphics classes - is this the best way?

Have also extended the x axis like we decided with the signature graph, and raised the missing tick as an issue: https://github.com/mozilla/metrics-graphics/issues/458

Yes, the old library doesn't need to be included any more!

Tchanders:
Incidentally, the graph colours sometimes stop matching the table colours, depending on which versions are chosen. This also happens with the old graph, however, so I should probably look into it as a separate issue.

webapp-django/crashstats/crashstats/templates/crashstats/daily.html
```diff
@@ -2,6 +2,7 @@
 {% block site_css %}
   {{ super() }}
   {% compress css %}
+  <link rel="stylesheet" type="text/css" href="{{ static('crashstats/css/lib/metricsgraphics.css') }}">
```
AdrianGaudebert: This was actually better along with jquery and others. It's the ``daily.less`` file that should be included after the more generic files. Can you please move that one down to be loaded last?

webapp-django/crashstats/crashstats/static/crashstats/css/daily.less
```diff
@@ -117,6 +117,63 @@ table.graph {
...
+}
+.mg-area2-color {
+    fill: #e31a1c;
+}
+.mg-area3-color {
+    fill: #008800;
+}
+.mg-area4-color {
+    fill: #1f78b4;
+}
```
AdrianGaudebert: You can greatly simplify this by merging selectors and having less rules. For example: 

```css
.mg-line1-color,
.mg-hover-line1-color {
    stroke: #6a3d9a;
}
```

AdrianGaudebert: Also this seems weird to me. Isn't there a way in MG to set those color sets in the javascript directly?

AdrianGaudebert:
What I meant by removing the old files was to remove them from the code base entirely! So ``git rm webapp-django/crashstats/crashstats/static/crashstats/js/flot-0.7/jquery.flot.pack.js`` for example. But of course you need to make sure that those files you are going to remove are indeed not used anymore anywhere in the app (which means using some ``grep`` in the repo).

Tchanders:
Ah right, flot can't be removed yet because it's being used for the home page, which we haven't replaced yet.

webapp-django/crashstats/crashstats/static/crashstats/css/daily.less
Tchanders: It seems not... These colours should be added to the new css file for customising metricsgraphics classes (so e.g. the home page graph will get the same colours). That means this is now dependent on https://github.com/mozilla/socorro/pull/2893

Tchanders:
@AdrianGaudebert Have addressed the above comments in this latest version

PR #2896, opened 2015-07-10
fixes bug 1188085 - socorro.external.es.supersearch.SuperSearch directly

peterbe:
@AdrianGaudebert r? cc @twobraids 

So this stops using the middleware for talking to ES for supersearch queries. 

Adrian, when I started this work I used to hack it so that the `fetch` method first did the usual `requests.get()` and then it did it the new way. So every page I hit, it would make 2 ES queries. 
I used that to make sure the results are the same. They were. Also, I used it to measure if the direct method was much faster. It was. But only by 5-15% depending on the size. So much of the time spent is on the network and when doing local development it uses CherryPy instead of the much better uWSGI we use in production. 

I know the size of the patch is scary. 99.99% of it is fixing how the mocking is done in tests. Instead of mocking `requests.get` it mocks `socorro.external.es.elasticsearch.SuperSearch`

@twobraids Please look over the changes in `supersearch/models.py` and let me know what you think. 
Also, please take a look at my hack in `SocorroCommon.fetch()` where I make it so that you can pass in either an implementation OR a string URL. This makes it really easy for me to keep ALL the existing fancy caching functionality without much extra work. 

@twobraids I'd also like to get some eyes on the logging. Truth is, there is very little logging in `socorro.external.es.supersearch.SuperSearch`. Perhaps a challenge for some other day. 

The plan:

* File a bug and rebase this PRs commit message to mention that bug :)
* Land this and test it in stage. 
* Attack the `SuperSearchFields` so that it too does not use the middleware
* Delete everything from the `middleware_app.py` that has to do with supersearch
* Attack some postgres related model classes. To use the same pattern. One service at a time. 

cc @lonnen and @rhelmer if you're curious.

### DIFF
--- a/socorro/external/es/supersearch.py
+++ b/socorro/external/es/supersearch.py
@@ -187,6 +187,7 @@ class SuperSearch(SearchBase):
         The list of accepted parameters (with types and default values) is in
         the database and can be accessed with the super_search_fields service.
         """
+
         # Filter parameters and raise potential errors.
         params = self.get_parameters(**kwargs)
 
--- a/webapp-django/crashstats/settings/base.py
+++ b/webapp-django/crashstats/settings/base.py
@@ -639,3 +639,24 @@ SYMBOLS_BUCKET_DEFAULT_LOCATION = config(
     'SYMBOLS_BUCKET_DEFAULT_LOCATION',
     None
 )
+
+# Config for when the models pull directly from socorro.external classes.
+# NOTE: This is overwritten, for tests in settings_test.py
+SOCORRO_IMPLEMENTATIONS_CONFIG = {
+    'elasticsearch': {
+        # All of these settings are repeated with sensible defaults
+        # in the implementation itself.
+        # We repeat them here so it becomes super easy to override
+        # from the way we set settings for the webapp.
+        'elasticsearch_urls': config(
+            'ELASTICSEARCH_URLS',
+            'http://localhost:9200',
+            cast=Csv()
+        ),
+        # e.g. (deliberately commented out)
+        # 'elasticsearch_doctype': config(
+        #     'ELASTICSEARCH_DOCTYPE',
+        #     'crash_reports'
+        # )
+    }
+}
--- a/webapp-django/crashstats/supersearch/models.py
+++ b/webapp-django/crashstats/supersearch/models.py
@@ -2,9 +2,40 @@ import functools
 import json
 
 from django.core.cache import cache
+from django.conf import settings
+
+from socorro.external.es import supersearch
+from socorro.middleware.middleware_app import MiddlewareApp
+
+from configman import Namespace, configuration, class_converter
 
 from crashstats import scrubber
 from crashstats.crashstats import models
+
+
+def_source = Namespace()
+def_source.namespace('elasticsearch')
+def_source.elasticsearch.add_option(
+    'elasticsearch_class',
+    doc='a class that implements the ES connection object',
+    default='socorro.external.es.connection_context.ConnectionContext',
+    from_string_converter=class_converter
+)
+
+
+def config_from_configman():
+    return configuration(
+        definition_source=[
+            def_source,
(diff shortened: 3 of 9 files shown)


socorro/external/es/supersearch.py
```diff
@@ -187,6 +187,9 @@ def get(self, **kwargs):
         The list of accepted parameters (with types and default values) is in
         the database and can be accessed with the super_search_fields service.
         """
+        # print "GET KWARGS"
+        # print kwargs
```
AdrianGaudebert: debug leftover

webapp-django/crashstats/settings/base.py
```diff
@@ -629,3 +629,24 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
     'SYMBOLS_BUCKET_DEFAULT_LOCATION',
     None
 )
+
+# Config for when the models pull directly from socorro.external classes.
+# NOTE: This is overwritten, for tests in settings_test.py
+SOCORRO_IMPLMENTATIONS_CONFIG = {
```
AdrianGaudebert: Typo, there's an ``E`` missing.

webapp-django/crashstats/supersearch/models.py
```diff
@@ -2,11 +2,42 @@
...
+)
+
+
+def config_from_configman():
+    return configuration(
+        definition_source=[
+            def_source,
+            # This is a tie-over until we totally get rid of the MiddlwareApp.
+            # At the moment it defines some useful options that are required
+            # by the
```
AdrianGaudebert: By the... ? :)

webapp-django/settings_test.py
```diff
@@ -52,3 +52,11 @@
 SYMBOLS_BUCKET_DEFAULT_NAME = 'my-lovely-bucket'
 SYMBOLS_FILE_PREFIX = 'v99'
 SYMBOLS_BUCKET_DEFAULT_LOCATION = 'us-west-2'
+
+
+# So it never ever actually uses a real ElasticSearch server
+SOCORRO_IMPLMENTATIONS_CONFIG = {
```
AdrianGaudebert: Same typo here.

webapp-django/crashstats/supersearch/models.py
```diff
@@ -2,11 +2,42 @@
...
+            MiddlewareApp.required_config.webapi,
+        ],
+        values_source_list=[
+            settings.SOCORRO_IMPLMENTATIONS_CONFIG,
```
AdrianGaudebert: And the last occurrence of that typo.

AdrianGaudebert:
This feels like a hack to me, and one with a weird taste, but I take it that's the fastest and easiest way to make this huge move. Using code that was initially made for fetching an HTTP API, to call a method on a object in the code, seems wrong, but at the same time I understand the advantages of doing that. 

I think it's fine, but I wanted to write down my concerns.

webapp-django/crashstats/settings/base.py
AdrianGaudebert: Also, why don't you use ``DATASERVICE_CONFIG_BASE`` that Selena added in that same config file a while back? I don't really understand what in here is using her previous work, if anything at all (but I guess it would have been a lot larger if you were not building on top of her work).

peterbe: The reason is pragmatism. I don't want to go near that stuff. 

Configman does some black magic with changing that config. It's a mutable dict. 
I don't want to go near it. The only thing that uses it is the Bugs service. 
When I'm done with this PR, I'll attack the SuperSearchFields next. And when I'm done that I'm going to rewrite the work Selena did on the Bugs service using my new technique. And when I've done that I'm going to attack each and every other postgres or http service implementation.

peterbe:
Nits addressed .

AdrianGaudebert:
r+!

PR #2898, opened 2015-07-14
Fixes Bug1183608 - added logging to assist in confirming acks

twobraids:
while not including remediation, this PR formalizes the logging of the acks so we can match them.

### DIFF
--- a/socorro/external/rabbitmq/crashstorage.py
+++ b/socorro/external/rabbitmq/crashstorage.py
@@ -187,6 +187,7 @@ class RabbitMQCrashStorage(CrashStorageBase):
                         ]
                     self.transaction(
                         self._transaction_ack_crash,
+                        crash_id_to_be_acknowledged,
                         acknowledgement_token
                     )
                     del self.acknowledgement_token_cache[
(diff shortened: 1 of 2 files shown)


socorro/external/rabbitmq/crashstorage.py
```diff
@@ -210,14 +211,20 @@ def _consume_acknowledgement_queue(self):
...
+    def _transaction_ack_crash(
+        self,
+        connection,
+        crash_id,
+        acknowledgement_token
+    ):
         connection.channel.basic_ack(
             delivery_tag=acknowledgement_token.delivery_tag
         )
+        self.config.logger.debug(
```
peterbe: Should be put a `logger.debug(...` **before** the `connection.channel.basic_ack(...)` so that you can see crashes that are **attempted** to be acked. Then you might be able to see which ones are attempted but don't get the `'RabbitMQCrashStorage acking %s with delivery_tag %s'` line.

peterbe:
r+ with nit/question to consider.

socorro/external/rabbitmq/crashstorage.py
twobraids: I specifically moved the logging to after the call so that the log would happen only if the call did not raise an exception.  That way I count successes, not attempts.

lonnen:
question satisfied. second r+ from me. landing.

PR #2899, opened 2015-07-14
Fixes Bug 1183752 - Fair and balanced submitter

twobraids:
the submitter app that we're using now, submits a query to PG to get a list of crashes to submit to staging.  This creates a bias toward crashes that have succeeded in processing and staging doesn't get a realistic distribution of crashes.  

This new method adds a throttling feature to the RabbitMQ crashstorage class.  Defaulting to 100% pass through, this change makes the queue filter crashes based on the percentage that should pass through.

The goal is to add this to the crashmovers as that will push crashes to a new queue in RabbitMQ.  Then the submitter(s) will read from that queue, pull the crashes from AWS and submit them via a proper HTTP POST to staging.


lonnen:
r+

PR #2900, opened 2015-07-14
fixes bug 1183804 - Admin 'Symbols Uploads' needs pagination

peterbe:
@AdrianGaudebert r?

Ping me if you need help understanding anything about symbols uploads and how to fake some content so you can test it.


webapp-django/crashstats/manage/views.py
```diff
@@ -508,11 +508,60 @@ def graphics_devices_lookup(request):
...
+            'size': upload.size,
+            'created': upload.created,
+        })
+    return {
+        'items': items,
+        'count': count,
+        'batch_size': batch_size,
+        'page': page,
+        'user_edit_url': reverse('manage:user', args=('0',)),
+        'symbols_content_url': reverse('symbols:content', args=('0',)),
```
AdrianGaudebert: Not really important, but in terms of API design, I think it's better to send those URLs along with the content and without the fake ``0``. It makes the data slightly bigger (but with 10 results, does that even matter? ) but it makes the front-end code and the API results simpler to understand. 

And if you do that, you can even remove the identifiers entirely, since I don't think you use them for anything else that building URLs.

webapp-django/crashstats/manage/templates/manage/symbols_uploads.html
```diff
@@ -31,30 +51,48 @@
...
+      </p>
+
+      <form action="" id="filter" data-dataurl="{{ url('manage:symbols_uploads_data') }}">
+        <input type="hidden" name="page">
+        <table class="data-table" >
+          <thead>
+            <tr>
+              <th>User</th>
+              <th>Filename</th>
+              <th>&nbsp;</th>
```
AdrianGaudebert: Why no name here?

webapp-django/crashstats/manage/static/manage/js/symbols-uploads.js
```diff
@@ -0,0 +1,118 @@
...
+    'use strict';
+
+    function humanFileSize( bytes, precision ) {
+        var units = [
+           'bytes',
+           'Kb',
+           'Mb',
+           'Gb',
+           'Tb',
+           'Pb'
```
AdrianGaudebert: nit: you have 3-spaces indentation here.

webapp-django/crashstats/manage/static/manage/js/symbols-uploads.js
```diff
@@ -0,0 +1,118 @@
...
+        ];
+        if (isNaN( parseFloat(bytes)) || !isFinite(bytes)) {
+            return '?';
+        }
+        var unit = 0;
+        while (bytes >= 1024) {
+          bytes /= 1024;
+          unit++;
```
AdrianGaudebert: And 2-spaces here. :)

webapp-django/crashstats/manage/static/manage/js/symbols-uploads.js
```diff
@@ -0,0 +1,118 @@
...
+        var $form = $('#filter');
+        var url = $form.data('dataurl');
+        var data = $form.serialize();
+        $.getJSON(url, data, function(response) {
+            stop_loading();
+            $('.count b').text(response.count);
+            $('.count:hidden').show();
+            $('tbody tr', $form).remove();
+            var $tbody = $('tbody', $form);
+            $('tr', $tbody).remove();
```
AdrianGaudebert: Isn't that redundant with line 43?

webapp-django/crashstats/manage/static/manage/js/symbols-uploads.js
```diff
@@ -0,0 +1,118 @@
...
+            function userURL(id) {
+                return response.user_edit_url.replace('0', id);
+            }
+            function contentURL(id) {
+                return response.symbols_content_url.replace('0', id);
+            }
```
AdrianGaudebert: As explained earlier, I think this is the responsibility of the API and not of the front-end. Imo these functions should be removed.

webapp-django/crashstats/manage/templates/manage/symbols_uploads.html
AdrianGaudebert: OK I get it.

webapp-django/crashstats/manage/static/manage/js/symbols-uploads.js
```diff
@@ -0,0 +1,118 @@
...
+            $page.val(parseInt($page.val()) + 1);
+            $('.pagination').hide();
+            fetch_uploads();
+        });
+
+        $('.pagination .previous').click(function(e) {
+            e.preventDefault();
+            var $page = $form.find('[name="page"]');
+            $page.val(parseInt($page.val()) - 1);
+            $('.pagination').hide();
```
AdrianGaudebert: I suggest making this a function: 

```javascript
function updatePage(incr) {
    var $page = $form.find('[name="page"]');
    $page.val(parseInt($page.val())+ incr);
    $('.pagination').hide();
}

updatePage(1);
updatePage(-1);
```

AdrianGaudebert:
Many nits, but overall it looks good. I would be happy to know how to test it indeed, if you could share some instructions on generating some fake data locally.

webapp-django/crashstats/manage/static/manage/js/symbols-uploads.js
peterbe: I plucked that from another app. 
I have hopefully fixed it now.

webapp-django/crashstats/manage/static/manage/js/symbols-uploads.js
peterbe: Yeah. You're right. 
I took that idea from another application where I make the JSON return EVERYTHING and then I do pagination in the webapp (angular). For that case I wanted to avoid repeating the URLs so much in the JSON and instead just sent a "URL template". 

It was premature optimization for this case. Fixed now.

AdrianGaudebert:
r+!

PR #2901, opened 2015-07-14
fixes bug 1183838 - Upgrade to Django 1.7

peterbe:
@AdrianGaudebert r?

So this upgrade trigger a lot of package updates. 
I created a brand new virtualenv based on **python 2.7** (which is a must) and one of the problems is that django's runserver command tries to import `inotify`. When it does so, because it is in the virtualenv, it causes a `SystemExit` and kills runserver. I'm going to update @rhelmer's fork of inotify. 

One of the reasons the PR looks big and scary is because I had to do `git mv crashstats/auth crashstats/authentication` because you can't have two installed apps that have the same suffix. We have `django.contrib.auth` AND `crashstats.auth`.


peterbe:
One thing that might need to be done is change the sha with which we point to @rhelmer's pyinotify fork once this lands https://github.com/rhelmer/pyinotify/pull/2

peterbe:
I realize that I haven't tested the wsgi endpoint. I've only tested it with runserver.

peterbe:
Note-to-self: Here's how I started the wsgi app locally:
```
uwsgi --pythonpath ./webapp-django/ -w webapp-django.wsgi.socorro-crashstats --http :9090 -H ~/virtualenvs/socorro
```
Then visiting `localhost:9090`

peterbe:
@AdrianGaudebert if it wasn't clear. This guy's ready for review again.

.travis.yml
```diff
@@ -4,7 +4,7 @@ cache:
     - $HOME/.cache/pip
 language: python
 python:
-  - "2.6"
+  - "2.7"
```
AdrianGaudebert: Are we there yet? Is it official? :pray:

webapp-django/crashstats/settings/base.py
```diff
@@ -30,7 +30,7 @@
     'django.contrib.sessions',
     'django.contrib.staticfiles',
     'commonware.response.cookies',
-    'south',  # important that django_nose comes AFTER this one
+    # 'south',  # important that django_nose comes AFTER this one
```
AdrianGaudebert: Why not just remove it since it's in django now?

.travis.yml
rhelmer: @AdrianGaudebert Yep, we've been 2.7 only since the AWS switch.

peterbe: @rhelmer I just noticed that 
```
docs/development/build-pkg-from-source.rst
docs/development/codeanddatabaseupdate.rst
docs/development/dev-services.rst
```
Also speak of Python 2.6. 
Can I just blanket update them too?

webapp-django/crashstats/settings/base.py
peterbe: A perfectly good review comment. 
I started that way but clearly didn't get back to removing it properly. 
Thanks!

.travis.yml
rhelmer: @peterbe please do.

AdrianGaudebert:
@peterbe so what's up with ``pyinotify``? I don't see it being updated in the requirements.

peterbe:
> @peterbe so what's up with pyinotify? I don't see it being updated in the requirements. 

Another great piece of PR review right there! I made a patch to Rob's fork but I was waiting to update `requirements.txt` until my [PR](https://github.com/rhelmer/pyinotify/pull/2) got merged. 
Thanks!

.travis.yml
AdrianGaudebert: :sunny:

AdrianGaudebert:
Tested locally, works well! r+

PR #2902, opened 2015-07-15
Fixes Bug 1183179 - Date parse error in signature graph data

Tchanders:


Tchanders:
@peterbe @AdrianGaudebert This also introduces different behaviour for >= and >, and fixes the comments which were the wrong way round

AdrianGaudebert:
r+

PR #2903, opened 2015-07-17
FIxes Bug 1185031 - make Rabbit refuse to offer crashes already in progress

twobraids:
when there duplicate crash_ids in the queues and the duplicates are picked up by the same processor, both the job acknowledgement cache and the dumps on disk experience a collision.  This results in failed queue ack'ng (and therefore stuck jobs) as well as failures in running the stackwalker.  The QueingThread should detect this situation and decline to accept the duplicate crash jobs.


lonnen:
At first blush I wondered if we couldn't resolve the duplicate ID from the queue, knowing that it was already being processed. After thinking on it a moment I agree with this implementation of declining to accept.

If we moved to a single threaded processor this bit of logic could be removed.

r+

PR #2904, opened 2015-07-20
NO BUG - added missing method to NullStorage

twobraids:
the NullCrashStorage class is used in ad hoc testing.  It serves no purpose in actual prod code.  In its use in testing, however, it ought to implement the full API. We don't it want raising errors.  We want it to silently ignore everything that it is told to do.

the method ```get_raw_dumps_as_files``` was missing, I added it...


lonnen:
r+

PR #2905, opened 2015-07-21
NO BUG - api adjustments for compatiblity rabbit / submitter

twobraids:
the API used by the submitter has a slight and annoying variation from the standard crashstorage API.  The changes here enable the rabbitmq crashstorage and new_crash_source classes to be compatible with the way that submitter wants things to work.


rhelmer:
r+

PR #2906, opened 2015-07-22
Fixes Bug 1186505 - unified signature for transform method in FTS apps

twobraids:
When we adopted RabbitMQ for use in the processor, we discovered that acknowledging a item taken from a Rabbit queue had to be done by the same thread that pulled the item from the queue.  We retro fitted the new_crashes iterator and the transform method of the processor to accomplish this.  Unfortunately, we only did that for the processor and its dependencies.

Now that we're wanting to use RabbitMQ in other places, like the submitter, we need to make the same retrofitting to that component.  Rather than doing it entirely piecemeal, refactor the acknowledgement callback so that _any_ component can successfully use RabbitMQ as a crash_id streaming source.  

Pull the "try...finally" block that runs the ack function out of the processor's transform method into the base class.  This allows all FTS apps to have the same behavior.  Change all the subclass "transform" methods into "_transform" and then have the base class "transform" method call the now private subclass "transform" method.

As for the tests, rework a FTS test to ensure that the correct ack function.  For all of the subclasses, this change then becomes transparent.  The existing tests will call the base class "transform" method which will, in turn, call the subclass private "_transform" method for testing.

### DIFF
--- a/socorro/app/fetch_transform_save_app.py
+++ b/socorro/app/fetch_transform_save_app.py
@@ -123,6 +123,8 @@ class FetchTransformSaveApp(App):
             for x in self.source.new_crashes():
                 if x is None:
                     yield None
+                elif isinstance(x, tuple):
+                    yield x  # already in (args, kwargs) form
                 else:
                     yield ((x,), {})  # (args, kwargs)
             else:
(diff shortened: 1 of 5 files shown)


lonnen:
r+

PR #2907, opened 2015-07-22
fixes bug 1186512 - Expose raw ADI counts

peterbe:
@rhelmer r? (informal)

Rob, I just wanted to share what I've got so far. 
Again, here's an example of its use: https://gist.github.com/peterbe/843ec4764e39dc6a364d

The reason I'm not treading with confidence is because I'm not entirely sure how I'm going to need this to work. My objective is to pull in data from this into the webapp and use SuperSearch aggregates to combine crash counts with ADI. In particular for the sake of the /daily report. I haven't yet started looking at throttle values.

### DIFF
--- a/socorro/external/postgresql/adi.py
+++ b/socorro/external/postgresql/adi.py
@@ -1,0 +1,79 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import datetime
+import logging
+
+from socorro.external.postgresql.base import PostgreSQLBase
+from socorro.external import MissingArgumentError
+from socorro.lib import external_common
+
+logger = logging.getLogger("webapi")
+
+
+class ADI(PostgreSQLBase):
+
+    def get(self, **kwargs):
+        yesterday = datetime.datetime.utcnow() - datetime.timedelta(1)
+        tomorrow = yesterday + datetime.timedelta(2)
+        yesterday = yesterday.date()
+        tomorrow = tomorrow.date()
+        filters = [
+            ('start_date', yesterday, 'date'),
+            ('end_date', tomorrow, 'date'),
+            ('product', '', 'str'),
+            ('version', '', 'str'),
+        ]
+        params = external_common.parse_arguments(filters, kwargs)
+        required = (
+            'start_date',
+            'end_date',
+            'product',
+            'version',
+        )
+        missing = []
+        for each in required:
+            if not params.get(each):
+                missing.append(each)
+        if missing:
+            raise MissingArgumentError(', '.join(missing))
+
+        SQL = """
+        SELECT
+            SUM(adi_count) AS adi_count,
+            date::DATE,
+            product_name AS product,
+            product_version AS version,
+            product_os_platform AS platform,
+            update_channel AS release_channel
+        FROM
+            raw_adi
+        WHERE
+            product_name = %(product)s AND
+            product_version = %(version)s AND
+            date BETWEEN %(start_date)s::DATE AND %(end_date)s::DATE
+        GROUP BY
+            date::DATE,
+            product_name,
+            product_os_platform,
+            product_version,
+            update_channel
+        """
+
+        results = self.query(SQL, params)
+
+        fields = (
+            'adi_count',
+            'date',
+            'product',
+            'version',
+            'platform',
+            'release_channel',
+        )
+        rows = []
+        for record in results:
+            row = dict(zip(fields, record))
+            row['date'] = row['date'].strftime('%Y-%m-%d')
+            rows.append(row)
+        return {'hits': rows}
--- a/socorro/middleware/middleware_app.py
+++ b/socorro/middleware/middleware_app.py
@@ -75,6 +75,7 @@ SERVICES_LIST = (
     (r'/supersearch/(.*)', 'supersearch.SuperSearch'),
     (r'/suspicious/(.*)', 'suspicious.SuspiciousCrashSignatures'),
     (r'/util/(versions_info)/(.*)', 'util.Util'),
+    (r'/adi/(.*)', 'adi.ADI'),
 )
 
 # certain items in a URL path should NOT be split by `+`
@@ -550,7 +551,7 @@ class ImplementationWrapper(JsonWebServiceBase):
         except AttributeError:
             try:
                 if (method_name == 'post' and
-                    getattr(instance, 'create', None)):
+                   getattr(instance, 'create', None)):
                     # use the `create` alias
                     method = instance.create
                 elif (method_name == 'put' and
--- a/socorro/unittest/external/postgresql/test_adi.py
+++ b/socorro/unittest/external/postgresql/test_adi.py
@@ -1,0 +1,142 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import datetime
+
+from nose.plugins.attrib import attr
+from nose.tools import eq_, assert_raises
+
+from socorro.external.postgresql.adi import ADI
+from socorro.external import MissingArgumentError
+
+from unittestbase import PostgreSQLTestCase
+
+
+@attr(integration='postgres')  # for nosetests
+class IntegrationTestADI(PostgreSQLTestCase):
+
+    def setUp(self):
+        """Set up this test class by populating the reports table with fake
+        data. """
+        super(IntegrationTestADI, self).setUp()
+        self._truncate()
+        cursor = self.connection.cursor()
+
+        self.date = datetime.datetime(2015, 7, 1)
+        yesterday = self.date - datetime.timedelta(hours=24)
+        products = ('Firefox', 'Thunderbird')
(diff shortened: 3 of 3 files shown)


peterbe:
NOTE: This doesn't yet have tests.

peterbe:
@rhelmer Now it has tests too.

socorro/external/postgresql/adi.py
```diff
@@ -0,0 +1,84 @@
...
+        SQL = """
+        SELECT
+            SUM(adi_count) AS adi_count,
+            date::DATE,
+            product_name,
+            product_os_platform,
+            product_version,
+            update_channel
+        FROM
+            raw_adi
```
rhelmer: Why not `product_adu`? That's already normalized against `product_versions`

socorro/middleware/middleware_app.py
```diff
@@ -551,7 +552,7 @@ def GET(self, *args, **kwargs):
         except AttributeError:
             try:
                 if (method_name == 'post' and
-                    getattr(instance, 'create', None)):
+                   getattr(instance, 'create', None)):
```
rhelmer: looks like this isn't aligned anymore

peterbe: I'm not sure why I did that change. I'm normally not a stickler for pep8 in the non-webapp files. But this was the last change necessary for it to be compliant.

socorro/external/postgresql/adi.py
peterbe: Because this gets us closer to the "raw deal". My subtle goal is to stay away from too much normalization that is based on stored procedures. 
The `raw_adi` table isn't based on a stored procedure but on an insert (insert from select) statement that happens right after the `raw_adi_logs` table has been added to. In a cron job. 

Am I mad for doing this? Is it overzealous? 

Also, using adi_count directly avoids joins. I'm not sure if it makes a big difference.

socorro/external/postgresql/adi.py
```diff
@@ -0,0 +1,84 @@
...
+            'adi_count',
+            'date',
+            'product_name',
+            'product_os_platform',
+            'product_os_version',
+            'product_version',
+            'build',
+            'product_guid',
+            'update_channel',
+            'received_at',
```
AdrianGaudebert: You are querying 6 fields in your SQL query, and then trying to extract 10 values out of it. Something is probably wrong somewhere! :)

socorro/external/postgresql/adi.py
```diff
@@ -0,0 +1,84 @@
...
+        )
+        rows = []
+        for record in results:
+            row = dict(zip(fields, record))
+            row['date'] = row['date'].strftime('%Y-%m-%d')
+            rows.append(row)
+        return {'hits': rows}
+        # return [dict(zip(fields, row)) for row in results]
```
AdrianGaudebert: debug leftover

socorro/unittest/external/postgresql/test_adi.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def setUp(self):
+        """Set up this test class by populating the reports table with fake
+        data. """
+        super(IntegrationTestADI, self).setUp()
+        self._truncate()
+        cursor = self.connection.cursor()
+
+        self.date = datetime.datetime(2015, 7, 1)
+        yesterday = self.date.replace(hour=0) - datetime.timedelta(hours=24)
```
AdrianGaudebert: Why do you need ``.replace(hour=0)``? It should already be ``0`` since you didn't pass a value for the ``hour`` param when you created it. And also, why not just make it a ``datetime.date`` object?

socorro/unittest/external/postgresql/test_adi.py
```diff
@@ -0,0 +1,143 @@
...
+                            '1.0',
+                            version,
+                            '20140903141017',
+                            '{abc}',
+                            channel,
+                        ))
+                        adi_count *= 2
+        cursor.execute('select count(*) from raw_adi')
+        count, = cursor.fetchone()
+        assert count == 2 * 2 * 2 * 2, count
```
AdrianGaudebert: This feels a bit hacky and not "future-proof", if we ever have to add a value somewhere it's going to be weird to have to change this. But it's nitpicking, I don't really mind.

socorro/unittest/external/postgresql/test_adi.py
```diff
@@ -0,0 +1,143 @@
...
+class IntegrationTestADI(PostgreSQLTestCase):
+
+    def setUp(self):
+        """Set up this test class by populating the reports table with fake
+        data. """
+        super(IntegrationTestADI, self).setUp()
+        self._truncate()
+        cursor = self.connection.cursor()
+
+        self.date = datetime.datetime(2015, 7, 1)
```
AdrianGaudebert: Just a question, but do we gain any value by testing with a moving date in this case? Like ``utc_now``?

socorro/external/postgresql/adi.py
```diff
@@ -0,0 +1,84 @@
...
+            update_channel
+        """
+
+        results = self.query(SQL, params)
+
+        fields = (
+            'adi_count',
+            'date',
+            'product_name',
+            'product_os_platform',
```
AdrianGaudebert: What is this field? Is it the same as ``os_name`` in our database / ``platform`` in Super Search (so something like Linux, Windows NT or Mac OS X)?

socorro/external/postgresql/adi.py
AdrianGaudebert: For the sake of consistency, I would like that we use the same names that are displayed in Super Search for all new services, even if that is different from what the database has. It will make it easier to for us to build new features when we're mixing Super Search with other things (and we are going to mix it with the data this ADI service exposes), and it will make the public API more consistent. 

in database         | in super search
--------------------|------------------------------
product_name        | product
product_os_platform | platform (I suppose)
product_os_version  | platform_version (I suppose)
product_version     | version
build               | build_id
product_guid        | product_id (I suppose)
update_channel      | release_channel (I suppose)

peterbe: > You are querying 6 fields in your SQL query, and then trying to extract 10 values out of it. 

Oops. Fixed now. The others just became `None`.

peterbe: Well, at first I objected to that. Some day we might replace ES with Cassandra or NextCoolThing and then the names would no longer make sense. 

**However**, some of the column name aliases you propose I think are really good. Independent of the fact that that's what supersearch calls them. I'll edit some of them.

socorro/unittest/external/postgresql/test_adi.py
peterbe: Oops. I used to use `datetime.datetime.utcnow()` but that just felt like a bad idea.

socorro/unittest/external/postgresql/test_adi.py
peterbe: I think you're right. I like having the assert statement there but it's strange to type 16 like that.

peterbe: Actually the more I look at it I think it just needs a more human-friendly comment.

socorro/unittest/external/postgresql/test_adi.py
peterbe: No great reason. I just like that it doesn't depend on the current date. It feels more solid this way.

socorro/external/postgresql/adi.py
peterbe: It's stuff like 'Windows_NT`, `Darwin`, `Linux`, `SunOS`, `Windows_95`. And lots and lots of other junk. 

Perhaps we ought to make it an optional where clause to filter by this.

peterbe: What do you think? Should I add a filtering option so that it only counts adi for certain known platform values? Is that going to be important to build "Crashes per user"?

socorro/external/postgresql/adi.py
AdrianGaudebert: Note that it is not Elasticsearch that replaces those names, but Super Search. So even if we were to rewrite Super Search with any other back-end, the Super Search Fields list would stay the same and we would still expose the same names. Also note that those names are not hard-coded and can be changed in our admin zone (even if I believe we should avoid that because it would probably be a breaking change in the API).

socorro/external/postgresql/adi.py
AdrianGaudebert: So if I understood our "path" correctly, in the near future we are going to use this service along with Super Search to build our graphs. What worries me here is that this field seems to contain data that is not available from Super Search: the "platform" can be one of "linux", "mac os x" or "windows nt" and rarely anything else. I don't think we have the more precise OS names like "Darwin" or "SunOS", so I'm concerned that we might be confused when trying to filter by platform. If it's inconsistent we should either make it consistent (either by putting the appropriate data in ES, through the processors like we did with the beta magic numbers or by making it simpler in our Postgres tables) or we do not expose this at all.

peterbe:
@AdrianGaudebert see latest commits

AdrianGaudebert:
Looking good! Do you want to keep it in the middleware, or should it be exposed in the public API as well?

peterbe:
So is that an r+?

I guess it wouldn't hurt to have it in the public web API. Are these numbers supposed to be protected?
What do you think @rhelmer  and @lonnen ?
The question is, should you be possible to extract ADI numbers from the public web API without permissions?

AdrianGaudebert:
Yes, r+ from me.

PR #2908, opened 2015-07-22
Fixes Bug 1186505 - complete a half-assed unification of 'source_iterator'

twobraids:
nope nope nope 

the PR that closed this Bug 1186505 was half-assed and incomplete.  

every instance of the method 'source_iterator' must accept from its internal iterator crashes either in the form of 
    1)  crash_id
    2)  ((crash_id,), {'additional': 'kwargs'})
and should always return the second form.

ALSO:

the rabbitmq method 'new_crashes' must run the '_consume_acknowledgement_queue' BEFORE checking for the end of iteration.  If it does not, then the last crash will not be acknowledged in the queue.

### DIFF
--- a/socorro/collector/crashmover_app.py
+++ b/socorro/collector/crashmover_app.py
@@ -78,8 +78,15 @@ class ProcessedCrashCopierApp(FetchTransformSaveApp):
           self.quit_check
         )
         for x in self.iterator():
-            self.config.logger.debug('yielding %s', x)
-            yield x  # (args, kwargs) or None
+            if x is None:
+                self.config.logger.debug('yielding %s', x)
+                yield None
+            elif isinstance(x, tuple):
+                self.config.logger.debug('yielding %s', x)
+                yield x  # (args, kwargs) or None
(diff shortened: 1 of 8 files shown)


lonnen:
```
lonnen: why not convert everything to ((x, ), {})
lars: I could go back to ensure that all crashstore iterators do that form, 
but _only_ rabbitmq needs it.  I would rather the iterators for the other crash 
store have a more obvious simple form - iterate over a stream of crashids - 
just the ids.   when implementing a new crashstore, I don't want to have to 
remember that an unrelated external resource needs a special form.  let the 
code deal with unifying the special form.

lars: honestly, I think an improvement to this solution would be to push the form
conversion down toward the base class of these three apps.  solve in only one 
place rather than three 

lars: however, the three case are just enough different that would mean 
introducing a new private method - converting the specialzations to use that 
private method.  The same technique that I used in the last PR for the method "transform"
```

lonnen:
r+ with this implementation. lets not dive into the base class yet

PR #2909, opened 2015-07-23
Fixes bug 1186903 - Raise BadArgumentError when on errors due to the …

AdrianGaudebert:
…user's input.

@peterbe r?

### DIFF
--- a/socorro/external/es/super_search_fields.py
+++ b/socorro/external/es/super_search_fields.py
@@ -6,7 +6,7 @@ import datetime
 import elasticsearch
 
 from socorro.external import (
-    InsertionError,
+    BadArgumentError,
     MissingArgumentError,
     ResourceNotFound,
 )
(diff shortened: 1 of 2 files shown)


peterbe:
r+

socorro/external/es/super_search_fields.py
```diff
@@ -92,9 +92,10 @@ def create_field(self, **kwargs):
             )
         except elasticsearch.exceptions.ConflictError:
             # This field exists in the database, it thus cannot be created!
-            raise InsertionError(
-                'The field "%s" already exists in the database, '
-                'impossible to create it. ' % params['name']
+            raise BadArgumentError(
+                'name',
+                msg='The field "%s" already exists in the database, '
+                    'impossible to create it. ' % params['name'],
```
peterbe: I would recommend using `%r` instead if you ever have a need to use `"%s"`. E.g. 
```
msg='The field %r already ...'
```

PR #2910, opened 2015-07-23
bug 1185751 - Get ADI pulled in on stage

peterbe:
@twobraids r?

I did some indentation fixes here and there too. That's why it looks like a bigger diff. 
Those changes makes the code read better I think. 

I'm now going to try to set up a crontabber instance locally that just runs this job.

### DIFF
--- a/socorro/cron/jobs/fetch_adi_from_hive.py
+++ b/socorro/cron/jobs/fetch_adi_from_hive.py
@@ -5,21 +5,16 @@
 import codecs
 import datetime
 import urllib2
-import csv
-import getpass
 import os
 import tempfile
 import unicodedata
 
 import pyhs2
 
-from configman import Namespace
+from configman import Namespace, class_converter
 from crontabber.base import BaseCronApp
-from crontabber.mixins import (
-    as_backfill_cron_app,
-    with_postgres_transactions,
-    with_single_postgres_transaction
-)
+from crontabber.mixins import as_backfill_cron_app
+from socorro.external.postgresql.connection_context import ConnectionContext
 from socorro.external.postgresql.dbapi2_util import execute_no_results
 
 """
@@ -128,9 +123,8 @@ _FENNEC38_ADI_CHANNEL_CORRECTION_SQL = """
               and build = '20150427090529'
               and date > '2015-04-27';"""
 
+
 @as_backfill_cron_app
-@with_postgres_transactions()
-@with_single_postgres_transaction()
 class FetchADIFromHiveCronApp(BaseCronApp):
     """ This cron is our daily blocklist ping web logs query
         that rolls up all the browser checkins and let's us know
@@ -167,7 +161,6 @@ class FetchADIFromHiveCronApp(BaseCronApp):
         doc='Password to connect to Hive with',
         secret=True)
 
-
     required_config.add_option(
         'hive_database',
         default='default',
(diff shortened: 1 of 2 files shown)


socorro/cron/jobs/fetch_adi_from_hive.py
```diff
@@ -183,13 +176,99 @@ class FetchADIFromHiveCronApp(BaseCronApp):
...
+        tx_class = self.config.primary_destination.transaction_executor_class
+        primary_transaction = tx_class(
+            self.config,
+            primary_database,
+        )
+        transactions = [primary_transaction]
+
+        db_class = self.config.secondary_destination.database_class
+        secondary_database = db_class(self.config.secondary_destination)
+        if secondary_database.config != primary_database.config:
```
twobraids: nice touch

peterbe:
How I tested this locally:

1. Make sure you're on VPN
2. Generate a new crontabber.ini file with `python socorro/cron/crontabber_app.py --admin.dump_conf=../crontabber-hive.ini`
3. Edit `../crontabber-hive.ini` and make sure the `jobs=` line looks like this:
    ```
    jobs=socorro.cron.jobs.fetch_adi_from_hive.FetchADIFromHiveCronApp|1d
    ```
4. Create a new database and load in the same stuff you have in your local `breakpad` table. 
    ```
    createdb breakpad_deleteme
    pg_dump breakpad > /tmp/breakpad.sql
    psql breakpad_deleteme < /tmp/breakpad.sql
    ```
5. Remove (or ignore) the config of all other jobs and focus on FetchADIFromHiveCronApp. Here's what mine looked like:
    ```
   [[class-FetchADIFromHiveCronApp]]
        # Hostname to run Hive query on
        hive_host={{REDACTED IN CASE IT MATTERS}}
        [[[secondary_destination]]]

            # the name of the database
            # see "resource.postgresql.database_name" for the default or override it here
            database_name=breakpad_deleteme
    ```
Note that I had to manually delete the `#query=` part. The only overriding changes should be `hive_host` and `secondary_destination.database_name`. 
6. Run it `python socorro/cron/crontabber_app.py --admin.conf=../crontabber-hive.ini`
7. Wait. About 10min
8. `psql breakpad` and `select count(*) from raw_adi_logs`
9. `psql breakpad_deleteme` and `select count(*) from raw_adi_logs`

Should have about 466k records in each.

socorro/cron/jobs/fetch_adi_from_hive.py
```diff
@@ -183,13 +176,99 @@ class FetchADIFromHiveCronApp(BaseCronApp):
...
+        db_class = self.config.primary_destination.database_class
+        primary_database = db_class(self.config.primary_destination)
+        tx_class = self.config.primary_destination.transaction_executor_class
+        primary_transaction = tx_class(
+            self.config,
+            primary_database,
+        )
+        transactions = [primary_transaction]
+
+        db_class = self.config.secondary_destination.database_class
```
twobraids: actually, perhaps these four lines would be better as:
```python
    if  (
        db_class = self.config.secondary_destination.database_class
        and self.config.secondary_destination != self.config.primary_destination
    ):
        db_class = self.config.secondary_destination.database_class
        secondary_database = db_class(self.config.secondary_destination)
```
That way we can turn off the secondary by setting the class to the empty string.

peterbe: You mean to change 
```python
    required_config.secondary_destination.add_option(
        'database_class',
        default=ConnectionContext,
        doc='the class responsible for connecting to Postgres',
    )
```
...to...:
```python
    required_config.secondary_destination.add_option(
        'database_class',
        default='',
        doc='the class responsible for connecting to Postgres',
    )
```
I'm not convinced. Its confusing what the value should be. And it's something else you also need to configure if you want to enable the second destination. Right now you just need to override the database credentials. Having to also configure the `database_class` would be another thing you have to get right.

twobraids:
r+ pending changes from conversation

PR #2911, opened 2015-07-24
Fixes Bug 1187325 - Add tests for signature graph

Tchanders:


Tchanders:
@AdrianGaudebert Is this along the right lines? It's a bit different from the other tests because the graph view only returns data. So it just tests whether the parameters are passed through to the api correctly and also tests the date parsing.

PR #2912, opened 2015-07-24
Fixes bug 1186355 - Added simple date histogram to Super Search middl…

AdrianGaudebert:
…eware.

@peterbe r?

This is the PR that will allow us to build time graphs on top of Super Search. cc @Tchanders

### DIFF
--- a/socorro/external/es/supersearch.py
+++ b/socorro/external/es/supersearch.py
@@ -161,16 +161,24 @@ class SuperSearch(SearchBase):
                 for key in bucket:
                     # Go through all sub aggregations. Those are contained in
                     # all the keys that are not 'key' or 'count'.
-                    if key in ('key', 'doc_count'):
+                    if key in ('key', 'key_as_string', 'doc_count'):
                         continue
 
                     sub_aggs[key] = [
-                        {'term': x['key'], 'count': x['doc_count']}
+                        {
+                            # For date data, Elasticsearch exposes a timestamp
+                            # in 'key' and a human-friendly string in
+                            # 'key_as_string'. We thus check if the later
+                            # exists to expose it, and return the normal
+                            # 'key' if not.
+                            'term': x.get('key_as_string', x['key']),
+                            'count': x['doc_count'],
+                        }
                         for x in bucket[key]['buckets']
                     ]
 
                 aggs[agg]['buckets'][i] = {
-                    'term': bucket['key'],
+                    'term': bucket.get('key_as_string', bucket['key']),
                     'count': bucket['doc_count'],
                 }
 
@@ -209,12 +217,18 @@ class SuperSearch(SearchBase):
             for param in sub_params:
 
                 if param.name.startswith('_'):
+                    # By default, all param values are turned into lists,
+                    # even when they have and can have only one value.
+                    # For those we know there can only be one value,
+                    # so we just extract it from the made-up list.
                     if param.name == '_results_offset':
                         results_from = param.value[0]
                     elif param.name == '_results_number':
                         results_number = param.value[0]
                     elif param.name == '_facets_size':
                         facets_size = param.value[0]
+                    elif param.name == '_histogram_interval.date':
+                        histogram_interval_date = param.value[0]
                     # Don't use meta parameters in the query.
                     continue
 
@@ -402,7 +416,7 @@ class SuperSearch(SearchBase):
                 )
 
         # Create signature aggregations.
-        if params['_aggs.signature']:
+        if params.get('_aggs.signature'):
             sig_bucket = A(
                 'terms',
                 field=self.get_field_name('signature'),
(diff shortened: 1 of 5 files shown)


scripts/setup_supersearch_app.py
```diff
@@ -21,7 +21,7 @@
 class SetupSuperSearchApp(generic_app.App):
     """Index supersearch fields data into elasticsearch. """
 
-    app_name = 'setup-supersearch'
+    app_name = 'middleware'
```
peterbe: Why this name change? What does it even do?

socorro/external/es/supersearch.py
```diff
@@ -422,6 +432,28 @@ def get(self, **kwargs):
 
             search.aggs.bucket('signature', sig_bucket)
 
+        # Create date histograms.
+        if params.get('_histogram.date'):
```
peterbe: Shouldn't this be `elif`?

socorro/external/es/supersearch.py
```diff
@@ -214,6 +222,8 @@ def get(self, **kwargs):
                         results_number = param.value[0]
                     elif param.name == '_facets_size':
                         facets_size = param.value[0]
+                    elif param.name == '_histogram_interval.date':
+                        histogram_interval_date = param.value[0]
```
peterbe: I'm sure it's correct but why the first one? Perhaps explain it to a dummy like me with a code comment.

socorro/external/es/supersearch.py
peterbe: Or can you aggregate on signature **and** on date?

socorro/lib/datetimeutil.py
```diff
@@ -172,3 +173,14 @@ def datestring_to_weekly_partition(date_str):
...
+
+def datetime_to_midnight(date):
+    """Return the midnight datetime of a given datetime.
+
+    Essentially all this does is return the same date with hour, minute,
+    second and microsecond set to 0.
+    """
+    return datetime.datetime(
+        date.year, date.month, date.day, tzinfo=date.tzinfo
+    )
```
peterbe: I'm sceptical to "polluting" this lib with a function that is only used in one place. 
Normally you use `mydatetimeinstance.replace` for this kind of thing. 

In your `test_get_with_date_histogram` all you need to do is add a little function that looks like this:

```python
def dt_to_midnight(d):
    return d.replace(hour=0, minute=0, second=0, microsecond=0)
```

My concern is that we're modifying a core lib just for the sake of a test. 
That's one thing. Another thing is that if we (some other developer) some day refactors `test_get_with_date_histogram` so it no longer needs `datetimeutil.datetime_to_midnight` then that person might forget to spot that `datetimeutil.py` can be cleaned up and slimmed down. And before you know it grows and becomes dead weight. 

That's why I always think you should only add things to your `main.css` if there's CSS you need on every (or nearly every) page.

peterbe:
I'm still struggling to understand why this is called a histogram. First of all, I can't see where it sorts the keys (aka. the dates) which is a prerequisite for a histogram. Perhaps that's just because I couldn't see that in the code. 
Also, why is the aggregate on signature not a "histogram"? Is it just because you can't sort signatures in some naturally useful way. 

Basically, code-complexity alarm bells ring my attention. The code is complex enough as it is. Now I have to juggle aggregates and histograms. Can we make things simpler?

Lastly, as far as I understand, a histogram is not a histogram until you display it. What we've been looking at here is the underlying data used for a future graph. That graph might be a histogram but the data isn't. 

I don't want to stand in your way. Just taking the time to raise some questions that might be worth wondering one extra time.

scripts/setup_supersearch_app.py
AdrianGaudebert: That has nothing to do here, it was a local change that sneaked in.

socorro/external/es/supersearch.py
AdrianGaudebert: You can aggregate on all fields in the same request (which is cool, because that means you can build many things with just one request to ES :) ).

socorro/external/es/supersearch.py
AdrianGaudebert: Yep, that's because all values are turned into lists, even when they have and can have only one value. I'll add a comment.

AdrianGaudebert:
The name "histogram" comes from Elasticsearch, as that is what they call the feature that I am using to generate the data. Date histograms are just a different kind of aggregations, and the data is actually returned as part of the ``facets`` part of the results. The difference in treatment comes from the fact that by default, facets (or aggregations, it's the same thing in our context) are using the ``terms`` aggregation from Elasticsearch - that's the case for the ``signature`` aggregation for example. 

Now we need to exploit some more specific features: nested aggregations (what I added for TCBS) and histogram aggregations (to generate graphs of data over time). The easiest way to add those to the existing SuperSearch is by creating new parameters. For the nested aggregations, I added a simple key called ``_aggs.signature``, on your suggestion, so that in the future we can easily add more possible parameters keeping the same pattern of ``_aggs.<field_name>``. The same goes here, where I'm adding a single param called ``_histogram.date`` (well, and its counterpart, ``_histogram_interval.date``, but they work the same way) with the idea that in the future we can easily extend that pattern to include more fields (Elasticseach support histograms on any number or date fields, and the one we will very likely need in the near future is ``build_id``). 

The sorting of the data is all done in Elasticsearch: it generates the keys per date interval, puts data in buckets accordingly, and then run some other facets in each bucket. 

Now, on to naming, I am not sure what else I should call that. Since that is an aggregation and is returned as one, I thought it good to name it differently, hence the ``date_histogram`` name in the results. I'm not good enough at English to come up with a better name for this type of data, to me it's a "histogram" because the ES documentation told me it was. :) If you have a better suggestion, I am all ears!

peterbe:
If it's a "ElasticSearch thing" calling it histogram then I'm totally OK with. That's using the tools correctly. 
It was just a guttural reaction that it seemed odd to call the data a histogram, but if that's how ES works that's cool. 

r+

PR #2914, opened 2015-07-27
fixes bug 1187497 - Cannot read compressed symbols

peterbe:
@rhelmer r?

### DIFF
--- a/webapp-django/crashstats/symbols/tests/test_views.py
+++ b/webapp-django/crashstats/symbols/tests/test_views.py
@@ -722,9 +722,11 @@ class TestViews(BaseTestViews):
             'Content-Encoding': 'gzip'
         })
         # The sample.zip file contains the file xpcshell.sym and it's
-        # 1156 bytes when un-archived. But the file is compressed so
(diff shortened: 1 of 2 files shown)


rhelmer:
r+, @peterbe how are you going to deal with the existing symbols, they'll need to be re-uploaded right?

peterbe:
> ... how are you going to deal with the existing symbols, they'll need to be re-uploaded right?

I'm not super sure. The processors (Ted's code) is doing fine with the "badly encoded" gzip files. And as discussed in the bug, if you can't use `wget`, use `curl --compressed` and if you can't use python `requests.get` you can use `urllib2`. 
Over time, this will sort of solve itself too. What do you think?
How can we even re-upload those files? Some monster script that downloads them all and re-encodes them and uploads again?

rhelmer:
@peterbe well we wouldn't need to do all of them, just the ones that got uploaded without gzip header. But, I think you're right that it probably does not matter as the snappy server (symbolapi) as well as stackwalker now works ok now.

PR #2915, opened 2015-07-27
fixes Bug 1168511 - filter all strings in raw crash input to remove \x00

twobraids:
we seem to be getting crashes with null bytes embedded in strings.  Ensure that the collector sanitizes these strings so that the null bytes don't come through and upset the delicate sensibilities of PostgreSQL.


lonnen:
r+

PR #2917, opened 2015-07-27
fixes bug 1165742 - throw out async-local-storage-with-Promise

peterbe:
@AdrianGaudebert r?

Now that we no longer rely on `IndexedDB` you can use the report/list in private browsing. 
Keeps things simpler also.


AdrianGaudebert:
Testing locally, I get this error when I refresh the page while being on the Reports tab: ``TypeError: Columns.value(...) is null - report_list_reports.js:79:16`` (it doesn't happen if I reload while being on any other tab).

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/report_list_reports.js
```diff
@@ -66,10 +66,6 @@ var Reports = (function() {
        activate: function(url) {
            url = url || null;
            if (loaded) return;
-           if (Columns.value() === null) {
-               // the storage hasn't returned yet!
-               return Columns.load(Reports.activate);
-           }
```
AdrianGaudebert: I do not understand why you remove this code, and I guess it's what causes the error I'm seeing.

peterbe: That's why we do code reviews :)
Again :)

I didn't mean to delete all of it. It used to be that `Columns.load()` was async so it had to be self-recursive. Now that it's sync, I can just call `Columns.load()` instead of this "Is it ready? No? Ok, then re-run".

peterbe:
I think I didn't think deep enough about it yesterday. 
Because the loading from `localStorage` is synchronous I just embedded the loading inside the `values()` function instead. Makes things even simpler.

AdrianGaudebert:
Sweet! r+!

PR #2921, opened 2015-07-29
Fixes bug 1188786 - Removed duplicate uuid columns in Reports tab of …

AdrianGaudebert:
…signature report page.

@peterbe r?

Same fix is in place in Super Search. It's difficult to test (it would require loading the entire DOM to verify that the list of columns doesn't have 'uuid', and I thought it was trivial enough to not be worth the test).

### DIFF
--- a/webapp-django/crashstats/signature/views.py
+++ b/webapp-django/crashstats/signature/views.py
@@ -135,12 +135,20 @@ def signature_reports(request):
         x for x in data['columns'] if x in allowed_fields
     ]
 
-    params['_columns'] = data['columns']
+    # Copy the list of columns so that they can differ.
+    params['_columns'] = list(data['columns'])
 
     # The uuid is always displayed in the UI so we need to make sure it is
     # always returned by the model.
     if 'uuid' not in params['_columns']:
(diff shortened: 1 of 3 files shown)


webapp-django/crashstats/signature/views.py
```diff
@@ -135,13 +135,18 @@ def signature_reports(request):
...
     # The uuid is always displayed in the UI so we need to make sure it is
     # always returned by the model.
     if 'uuid' not in params['_columns']:
         params['_columns'].append('uuid')
 
+    # The `uuid` field is a special case, it will not be displayed like
+    # normal columns.
+    if 'uuid' in data['columns']:
+        data['columns'].remove('uuid')
+
```
peterbe: This is a mess :)
A couple of lines above it says `if 'uuid' not in params['_columns']`. That changes `_columns` which was created from `columns` and then later it possible changes `columns`. 
This `params` vs. `data` makes my head spin.

AdrianGaudebert:
@peterbe I added some better comments, hope that makes it clearer!

peterbe:
r+

PR #2923, opened 2015-07-29
Fixes Bug 1187642 - wrapped rabbitmq basic_get in transaction object

twobraids:
the queuing thread of the processors (and stage-submitter) do not recover from connection errors on that thread.  The iterator is not protected by a transaction object because for other queue types, that can result a long running transaction and excessive locking.  This historical decision is not necessary as rabbit's ```...basic_get``` method can easily be wrapped in a transaction object.  

This PR does the wrapping and new tests ensure that a broken connection is re-established rather than just allowing the thread to hang.


lonnen:
r+ gogogogo

PR #2925, opened 2015-07-29
fixes Bug 1158877 - rework the daily_url cron job

twobraids:
this is a hand off to someone else (lonnen, peterbe?)  I'm not sure of the status of this PR from some months ago.  It passes its own tests, and has been rebased with master.   I've not run this anywhere, so it needs some manual testing...

### DIFF
--- a/socorro/cron/jobs/daily_url.py
+++ b/socorro/cron/jobs/daily_url.py
@@ -2,103 +2,334 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import contextlib
-import csv
 import datetime
 import gzip
+import csv
 import os.path
-import subprocess
+import re
+
+import contextlib
+
 
 from configman import Namespace
-from crontabber.base import BaseCronApp
+
 from crontabber.mixins import (
-    as_backfill_cron_app,
-    with_postgres_transactions,
-    with_single_postgres_transaction
+    using_postgres,
+    with_subprocess,
+    as_backfill_cron_app
 )
-from socorro.database.cachedIdAccess import IdCache
+from crontabber.base import BaseCronApp
+
+from socorro.external.postgresql.dbapi2_util import (
+    execute_no_results
+)
 from socorro.lib.util import DotDict
 
-SQL = """
-select
-  r.signature,  -- 0
-  r.url,        -- 1
-  'http://crash-stats.mozilla.com/report/index/' || r.uuid as uuid_url, -- 2
-  to_char(r.client_crash_date,'YYYYMMDDHH24MI') as client_crash_date,   -- 3
-  to_char(r.date_processed,'YYYYMMDDHH24MI') as date_processed,         -- 4
-  r.last_crash, -- 5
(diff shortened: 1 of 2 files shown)


lonnen:
I haven't forgotten about this. Will try to get to it tomorrow.

lonnen:
@peterbe can you pick this up? I suspect this branch is incomplete

peterbe:
I certainly can. But it'll take some while. I'd need to boot up my context too. 
I have more urgent things on my todo list but happy to do the work. How urgent is this?

socorro/cron/jobs/daily_url.py
```diff
@@ -2,103 +2,334 @@
...
-  '%(yesterday_str)s' <= r.date_processed and r.date_processed < '%(now_str)s'
-  %(prod_phrase)s %(ver_phrase)s
-order by 5 -- r.date_processed, munged
-"""
 
+sql = """
+      select
+        r.signature,  -- 0
+        r.url,        -- 1
+        'http://crash-stats.mozilla.com/report/index/' || r.uuid as uuid_url, -- 2
```
peterbe: https not http

socorro/cron/jobs/daily_url.py
```diff
@@ -2,103 +2,334 @@
...
+        r.hangid, --22
+        r.reason, --23
+        r.process_type, --24
+        r.app_notes, --25
+        r.install_age, --26
+        rd.duplicate_of, --27
+        r.release_channel, --28
+        r.productid --29
+      from
+        reports r left join reports_duplicates rd on r.uuid = rd.uuid
```
peterbe: Why not reports_clean?

socorro/cron/jobs/daily_url.py
```diff
@@ -2,103 +2,334 @@
...
+                        )
+                        headers_not_yet_written = False
+                    column_value_list = process_crash(crash_row)
+                    write_row(output_context, column_value_list)
+    except Exception, x:
+        config.logger.error(
+            'major problem running dailyUrlDump',
+            exc_info=True
+        )
+        raise
```
peterbe: I don't see the point of the lines above. If there is an exception, crontabber will spit out it to logging and it will also put the full exception into state.

peterbe:
Note to self, here's how you run just one crontabber app with some custom settings:
```
env crontabber.class-DailyURLCronApp.output_path=/tmp/dailyurl \
crontabber.jobs="socorro.cron.jobs.daily_url.DailyURLCronApp|1d" \
python socorro/cron/crontabber_app.py
```

socorro/cron/jobs/daily_url.py
twobraids: ```reports_clean``` did not exist when this was written.  Changing the source table now increases the risk of introducing some some small subtle problem.

socorro/cron/jobs/daily_url.py
twobraids: while I haven't looked it up, it is likely an artefact of the original code. Change as little as possible to avoid introducing new behaviors.

peterbe:
This is going to become a real-time report instead of a cron job.

PR #2927, opened 2015-07-30
Fixes Bug 1188793 - [Signature report] Closing an aggregation panel d…

Tchanders:
…oesn't work


Tchanders:
@AdrianGaudebert It's because it was looking for the delete button inside the wrong element... Have defined the button and bound the event to that just to be safe.

AdrianGaudebert:
Yay! Thanks! r++

PR #2928, opened 2015-07-30
fixes bug 1189394 - Update travis to use Elasticsearch 1.4

peterbe:
Testing to find out what version of ES travis installs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,11 +16,7 @@ services:
 before_install:
   - npm install -g less
   - gem install puppet puppet-lint
-  - wget --no-check-certificate 'https://download.elasticsearch.org/elasticsearch/elasticsearch/elasticsearch-0.90.10.tar.gz'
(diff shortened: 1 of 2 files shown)


peterbe:
@AdrianGaudebert Now I it installs and runs ES 1.4.5 and it seems to get past that problem where groovy is needed. However, at the time of writing it fails on something in `test_supersearch.py`. Care to take a look?

Actually, I just restarted the tests seconds ago, in case it was a matter of "bad luck" but I think that was wishful thinking.

AdrianGaudebert:
@peterbe I think this might be a specific test that's failing with ES1.4 for some weird reason. Maybe we can wait for https://github.com/mozilla/socorro/pull/2929 to be merged before doing this one? The other option could be to mark the failing test as running only on ES 0.90.

peterbe:
I agree on doing that PR dependency. To #2929 !

AdrianGaudebert:
Once #2929 is merged and this turns green, r+.

PR #2930, opened 2015-07-30
bug 1189432 - remove metro firefox

peterbe:
@rhelmer like this?

### DIFF
--- a/alembic/versions/b99155654de_sunset_all_future_metrofirefox.py
+++ b/alembic/versions/b99155654de_sunset_all_future_metrofirefox.py
@@ -1,0 +1,29 @@
+"""sunset all future MetroFirefox
+
+Revision ID: b99155654de
+Revises: 56f5cdf9bcdb
+Create Date: 2015-07-31 15:51:26.282796
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = 'b99155654de'
+down_revision = '56f5cdf9bcdb'
+
+from alembic import op
+from socorro.lib.migrations import load_stored_proc
+
+
+def upgrade():
+    op.execute("""
+        UPDATE product_versions
+        SET sunset_date = NOW()
+        WHERE
+            product_name = 'MetroFirefox' AND
+            sunset_date >= NOW()
+    """)
+    load_stored_proc(op, ['update_product_versions.sql'])
+
+
+def downgrade():
+    load_stored_proc(op, ['update_product_versions.sql'])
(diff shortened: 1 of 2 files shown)


rhelmer:
Pretty much. Need a migration too.

peterbe:
Would it just be `DELETE from releases_recent where product_name='MetroFirefox'`?

rhelmer:
@peterbe hmm well at minimum you would need to re-load the stored procedure you modified.

Here is an example:
https://github.com/mozilla/socorro/blob/master/alembic/versions/4a068f8838c6_bug_959731_update_product_versions_.py

peterbe:
...as well as the delete? Delete first, then the `load_stored_proc(op, ['update_product_versions.sql'])` call?

rhelmer:
I'm not sure what the purpose of the delete is. Do you want to let this age out?

AFAIK we've never tried to totally remove a product from Socorro before, I don't know what all needs to be removed. We let things like Camino age out.

alembic/versions/b99155654de_sunset_all_future_metrofirefox.py
```diff
@@ -0,0 +1,27 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'b99155654de'
+down_revision = '56f5cdf9bcdb'
+
+from alembic import op
+
+
+def upgrade():
+    op.execute("""
```
rhelmer: you also need `load_stored_proc(op, ['update_product_versions.sql'])`, to load the stored proc

alembic/versions/b99155654de_sunset_all_future_metrofirefox.py
```diff
@@ -0,0 +1,27 @@
...
+        UPDATE product_versions
+        SET sunset_date = NOW()
+        WHERE
+            product_name = 'MetroFirefox' AND
+            sunset_date >= NOW()
+    """)
+
+
+def downgrade():
```
rhelmer: I don't think it's worth trying to undo the product_versions update in the case of a downgrade, but you should load the stored proc again (the assumption being that if we downgrade, we want to revert the stored proc change), just `load_stored_proc(op, ['update_product_versions.sql'])` again

peterbe:
`load_stored_proc(op, ['update_product_versions.sql'])` added

rhelmer:
r+

We should try it out on stage, need to figure out exactly how to set up alembic so it works (we always used to run it on the DB as the postgres user, which isn't doable with RDS)

peterbe:
Is alembic not automatically run? So now when I land this it won't execute the migration??

rhelmer:
@peterbe alembic migrations have never been automatic, and we haven't had to run one in AWS yet. It should just be a matter of making an `alembic.ini` and running it somewhere that has access to the database.

PR #2932, opened 2015-08-01
Bug 1190046 - Update forced versions for correlations for Firefox cycle starting 2015-08-10

KaiRo-at:
This adds current aurora and nightly versions as well as all planned beta versions for this cycle, and should land on prod in the week of August 10.

Note the conversation from https://github.com/mozilla/socorro/pull/2877 - we need to update the crashanalysis node with that code to actually make it live.


lonnen:
mergeing so this goes out with the next prod push.

PR #2935, opened 2015-08-05
Fixes bug 1184173 - Implemented Top Crashes in real time using Super …

AdrianGaudebert:
…Search.

@peterbe r?

This is losing some features from the current TCBS page: 

- [ ] startup crash icon (would require a new feature in Super Search)
- [ ] versions count column (would require a third Super Search query, or a new feature in Super Search)
- [ ] first appearance column (would require a new API endpoint to get that data from postgres)
- [x] per-day data (because I've been matching existing features for the moment)
- [ ] graph data (removed on purpose)
- [ ] by build date (removed on purpose)

Since this is a separate page, no linked from anywhere, I would be in favor of merging this asap, then showing it to a few people and building missing features as follow-ups.

### DIFF
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -870,6 +870,11 @@ class Platforms(SocorroMiddleware):
             for x in result['hits']
         ]
 
+    def get_all(self):
+        """Return all platforms without reducing them to a pre-configured list.
+        """
+        return super(Platforms, self).get()
+
 
 class CrashesPerAdu(SocorroMiddleware):
     # Fetch records for active daily installs.
--- a/webapp-django/crashstats/settings/base.py
+++ b/webapp-django/crashstats/settings/base.py
@@ -35,16 +35,19 @@ INSTALLED_APPS = (
 
     # Application base, containing global templates.
     '%s.base' % PROJECT_MODULE,
-    # Example code. Can (and should) be removed for actual projects.
+
+    # Other Socorro apps.
     '%s.dataservice' % PROJECT_MODULE,
     '%s.crashstats' % PROJECT_MODULE,
     '%s.api' % PROJECT_MODULE,
     '%s.manage' % PROJECT_MODULE,
     '%s.supersearch' % PROJECT_MODULE,
     '%s.signature' % PROJECT_MODULE,
+    '%s.topcrashers' % PROJECT_MODULE,
     '%s.authentication' % PROJECT_MODULE,
     '%s.tokens' % PROJECT_MODULE,
     '%s.symbols' % PROJECT_MODULE,
+
     'django.contrib.messages',
     'raven.contrib.django.raven_compat',
     'waffle',
--- a/webapp-django/crashstats/topcrashers/templates/topcrashers/topcrashers.html
+++ b/webapp-django/crashstats/topcrashers/templates/topcrashers/topcrashers.html
@@ -1,0 +1,212 @@
+{% extends "crashstats_base.html" %}
+{% block site_css %}
+  {{ super() }}
+
+  {% compress css %}
+  <link href="{{ static('crashstats/css/topcrashers.less') }}" rel="stylesheet" type="text/less" media="screen" />
+  {% endcompress %}
+{% endblock %}
+
+{% block page_title %}
+Top Crashers for {{ product }} {{ version }}
+{% endblock %}
+
+{% block content %}
+    <div id="mainbody">
+        <div class="page-heading">
+            <h2>
+                Top Crashers for <span id="current-product">{{ product }}</span>
+                <span id="current-version">{{ version }}</span>
+            </h2>
+            <ul class="options">
+                <li>
+                    <a
+                        href="{{ change_query_string(_tcbs_mode=None) }}"
+                        {% if query.mode == 'realtime' %} class="selected"{% endif %}
+                    >Near Real Time</a>
+                </li>
+                <li>
+                    <a
+                        href="{{ change_query_string(_tcbs_mode='byday') }}"
+                        {% if query.mode == 'byday' %} class="selected"{% endif %}
+                    >By Day</a>
+                </li>
+            </ul>
+        </div>
+
+        <div class="panel">
+            <div class="body">
+
+            {% if total_crashing_signatures %}
+            <div>
+                Top {{ total_crashing_signatures }} Crashing Signatures.
+                <time class="start-date">{{ query.start_date }}</time> through
+                <time class="end-date">{{ query.end_date }}</time>.
+                <p>The report covers <span class="percentage">{{ number_of_crashes }} ({{ total_percentage | round(2) }}%)</span> of all {{ total_number_of_crashes }} crashes during this period. </p>
+            </div>
+            {% endif %}
+
+            <ul class="tc-duration-type tc-filter">
+                <li class="tc-selector-heading">Type:</li>
+                {% for c in process_type_values %}
+                <li>
+                    <a href="{{ change_query_string(process_type=c) }}" {% if crash_type == c %} class="selected" {% endif %} >
+                        {{ c|capitalize }}
+                    </a>
+                </li>
+                {% endfor %}
+            </ul>
+            <ul class="tc-duration-days tc-filter">
+                <li class="tc-selector-heading">Days:</li>
+                {% for day in possible_days %}
+                <li><a href="{{ change_query_string(days=day) }}" {% if days == day %} class="selected" {% endif %}>{{ day }}</a></li>
+                {% endfor %}
+            </ul>
+            <ul class="tc-per-platform tc-filter">
+                <li class="tc-selector-heading">OS:</li>
+                <li><a href="{{ change_query_string(platform=None) }}" {% if not os_name %} class="selected" {% endif %}>All</a></li>
+                {% for os in platform_values %}
+                <li><a href="{{ change_query_string(platform=os) }}" {% if os_name == os %} class="selected" {% endif %}>{{ os }}</a></li>
+                {% endfor %}
+            </ul>
+            <ul class="tc-result-count tc-filter">
+                <li class="tc-selector-heading">Result Count:</li>
+                {% for res_count in result_counts %}
+                <li><a href="{{ change_query_string(_facets_size=res_count) }}" {% if result_count == res_count %} class="selected" {% endif %}>{{ res_count }}</a></li>
+                {% endfor %}
+            </ul>
+            {% if total_crashing_signatures %}
+            <table id="{% if not os_name %}signature-list{% else %}peros-tbl{% endif %}" class="data-table tablesorter">
+                <thead>
+                    <tr>
+                        <th scope="col">Rank</th>
+                        <th title="The percentage of crashes against overall crash volume" scope="col">%</th>
+                        <th title="The change in percentage since the {{ query.start_date }} report" scope="col">Diff</th>
+                        <th scope="col">Signature</th>
+                        {% if not os_name %}
+                        <th title="Crashes across platforms" scope="col">Count</th>
+                        <th scope="col">Win</th>
+                        <th scope="col">Mac</th>
+                        <th scope="col">Lin</th>
+                        {% else %}
+                        <th scope="col">{{ os_name }}</th>
+                        {% endif %}
+                        <th scope="col">Is GC</th>
+                        <th class="bugzilla_numbers" scope="col">Bugzilla IDs</th>
+                        <th title="Does not imply Causation" scope="col">Correlation</th>
+                    </tr>
+                </thead>
+                <tbody>
+                {% for crash in tcbs %}
+                    <tr>
+                        <td class="rank">
+                            {{ crash.rank }}
+                            {% if crash.rank_diff != 'new'
+                               and crash.rank_diff >= 5
+                               or crash.rank_diff <= -5 %}
+                            <span {% if crash.rank_diff > 1 %} class="moving-up" {% elif crash.rank_diff < -1 %} class="moving-down" {% endif %} title="Movement in rank since the {{ query.start_date }} report">{{ crash.rank_diff }}</span>
+                            {% endif %}
+                        </td>
+                        <td>{{ crash.percent | round(2) }}%</td>
+                        {% if crash.diff == 'new' %}
+                        <td title="This is a new signature">new</td>
+                        {% else %}
+                        <td title="A change of {{ crash.diff | round(2) }}% from {{ crash.previous_percent | round(2) }}%">{{ crash.diff | round(2) }}%</td>
+                        {% endif %}
+                        <td class="signature-column">
+                            <a class="signature" href="{{ url('signature:signature_report') }}?product={{ product }}&amp;signature={{ crash.signature|urlencode }}&amp;version={{ version }}" title="{{ crash.signature }}">{{ crash.signature|truncate(length=80, killwords=True) }}</a>
+                            <div class="sig-history-container hide">
+                                <input type="hidden" class='ajax-signature' name="ajax-signature-1" value="{{ crash.signature }}" />
+                            </div>
+                            <div class="signature-icons">
+                                {% if crash.hang_count > 0 %}
+                                <img src="{{ static('img/3rdparty/fatcow/stop16x16.png') }}" alt="Hanged Crash" title="Hanged Crash" class="hang" height="16" width="16">
+                                {% endif %}
+
+                                {% if crash.plugin_count > 0 %}
+                                <img src="{{ static('img/3rdparty/fatcow/brick16x16.png') }}" alt="Plugin Crash" title="Plugin Crash" class="plugin" height="16" width="16">
+                                {% else %}
+                                <img src="{{ static('img/3rdparty/fatcow/application16x16.png') }}" width="16" height="16" alt="Browser Crash" title="Browser Crash" class="browser" />
+                                {% endif %}
+                            </div>
+                        </td>
+                        {% if not os_name %}
+                        <td>{{ crash.count }}</td>
+                        <td>{{ crash.win_count }}</td>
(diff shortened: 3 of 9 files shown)


AdrianGaudebert:
http://crashstats/topcrashers/?product=WaterWolf

peterbe:
Remember, we get a "0.0%" thing in the header
screenshot 2015-08-06 08 27 09

peterbe:
As agreed in meeting, no r+ until we get the toggle between real-time and yesterday's. :)

peterbe:
I did some basic benchmarking (using randomization)

```
WITHOUT ANY CACHING (about 15 rounds per URL):

URL                                                     Average    Median
/topcrashers/?product=Firefox&version=41.0a2&days=7     2.2944     2.2452
/topcrasher/products/Firefox/versions/41.0a2?days=7     1.7973     1.8018
```
and with model caching switched on
```
WITH CACHING (about 15 rounds per URL):

URL                                                     Average    Median
/topcrashers/?product=Firefox&version=41.0a2&days=7     1.2246     1.2415
/topcrasher/products/Firefox/versions/41.0a2?days=7     0.3227     0.3341
```
```

**Comment by peterbe** [comment]
Let's not read into any absolute numbers, but it's strange that the SuperSearch one is almost 4 times slower when caching is allowed to help. But only 25% slower when no caching is involved. Perhaps there's something strange going on with crashstats.supersearch.SuperSearch and caching now that it's not going via the middleware.

File: webapp-django/crashstats/crashstats/models.py [hunk 1]
```diff
@@ -826,6 +826,11 @@ def get(self):
             for x in result['hits']
         ]
 
+    def get_all(self):
+        """Return all platforms, not just the hardcoded ones.
```
peterbe: "not just the hardcoded ones" is not a true comment. 
It's all platforms without reducing it down to a list of dicts with the additional information whether it should be displayed or not.

webapp-django/crashstats/settings/base.py
```diff
@@ -35,16 +35,19 @@
...
     '%s.crashstats' % PROJECT_MODULE,
     '%s.api' % PROJECT_MODULE,
     '%s.manage' % PROJECT_MODULE,
     '%s.supersearch' % PROJECT_MODULE,
     '%s.signature' % PROJECT_MODULE,
+    '%s.topcrashers' % PROJECT_MODULE,
     '%s.authentication' % PROJECT_MODULE,
     '%s.tokens' % PROJECT_MODULE,
     '%s.symbols' % PROJECT_MODULE,
+
```
peterbe: Strange line :)

webapp-django/crashstats/topcrashers/templates/topcrashers/topcrashers.html
```diff
@@ -0,0 +1,197 @@
...
+  {% endcompress %}
+{% endblock %}
+
+{% block page_title %}
+Top Crashers for {{ product }} {{ version }}
+{% endblock %}
+
+{% block content %}
+    <div id="mainbody">
+        <input type="hidden" name="csrfmiddlewaretoken" value="{{ csrf_token }}">
```
peterbe: What's this for?

webapp-django/crashstats/topcrashers/templates/topcrashers/topcrashers.html
```diff
@@ -0,0 +1,197 @@
...
+
+        <div class="panel">
+            <div class="body">
+
+            {% if total_crashing_signatures %}
+            <div>
+                Top {{ total_crashing_signatures }} Crashing Signatures.
+                <time class="start-date">{{ query.start_date }}</time> through
+                <time class="end-date">{{ query.end_date }}</time>.
+                <p>The report covers <span class="percentage" title="{{ number_of_crashes }} ({{ (total_percentage * 100) | round(2) }}%)">{{ number_of_crashes }} ({{ (total_percentage * 100) | round(2) }}%)</span> of all {{ total_number_of_crashes }} crashes during this period. </p>
```
peterbe: This `title` attribute doesn't add any value.

peterbe: It irks me that we have a variable called `total_percentage` but then we multiply it with 100. If it was `total_ratio * 100` or something it'd make sense.

webapp-django/crashstats/topcrashers/templates/topcrashers/topcrashers.html
```diff
@@ -0,0 +1,197 @@
...
+                        <td class="rank">
+                            {{ crash.rank }}
+                            {% if crash.rank_diff != 'new'
+                               and crash.rank_diff not in [-1,0,1]
+                               and crash.rank_diff >= 5
+                               or crash.rank_diff <= -5 %}
+                            <span {% if crash.rank_diff > 1 %} class="moving-up" {% elif crash.rank_diff < -1 %} class="moving-down" {% endif %} title="Movement in rank since the {{ tcbs.start_date }} report">{{ crash.rank_diff }}</span>
+                            {% endif %}
+                        </td>
+                        <td>{{ (crash.percent * 100) | round(2) }}%</td>
```
peterbe: ditto

webapp-django/crashstats/topcrashers/tests/test_views.py
```diff
@@ -0,0 +1,223 @@
...
+    @staticmethod
+    def setUpClass():
+        TestViews.switch = Switch.objects.create(
+            name='topcrashers-real-time',
+            active=True,
+        )
+
+    @staticmethod
+    def tearDownClass():
+        TestViews.switch.delete()
```
peterbe: Don't these two need to call the parent setUpClass and tearDownClass ? 
Why are they static methods?!

peterbe: Yeah, it should be
```python
@classmethod
def setUpClass(cls):
    super(TestViews, cls).setUpClass()
    TestViews.switch ...
```

webapp-django/crashstats/topcrashers/tests/test_views.py
```diff
@@ -0,0 +1,223 @@
...
+    SUPERSEARCH_FIELDS_MOCKED_RESULTS,
+)
+
+
+class TestViews(BaseTestViews):
+    base_url = reverse('topcrashers:topcrashers')
+
+    @staticmethod
+    def setUpClass():
+        TestViews.switch = Switch.objects.create(
```
peterbe: Kill kill kill all waffle switches that aren't useful. 
Waffle flags are useful when you're swapping one feature for another. But in this case, it's a new URL. And it's not a secret. Waffle is a pain to deal with locally and it's annoying that you have manually mess around with django commands on the servers. 

I basically think we should delete almost ALL waffle switches everywhere. Perhaps you know better examples, but I can't think of a single one that is useful to us. 

Also, note that with waffle switches the view can't begin until Django has talked to the postgres database so they're often a performance penalty.

webapp-django/crashstats/topcrashers/tests/test_views.py
```diff
@@ -0,0 +1,223 @@
...
+                     "signature": u"FakeSignature1 \u7684 Japanese"}
+                ]
+            }
+        rpost.side_effect = mocked_post
+
+        def mocked_get(url, params, **options):
+            assert 'supersearch' in url
+
+            if 'supersearch/fields' in url:
+                return Response(SUPERSEARCH_FIELDS_MOCKED_RESULTS)
```
peterbe: One of us will have to change this when supersearchfields are served straight outta socorro.external.es :)

webapp-django/crashstats/topcrashers/tests/test_views.py
```diff
@@ -0,0 +1,223 @@
...
+        supersearch().get.side_effect = mocked_supersearch_get
+
+        # first without a version
+        no_version_url = self.base_url + '?product=WaterWolf'
+        url = self.base_url + '?product=WaterWolf&version=19.0'
+        reports_count_default = url
+        reports_count_100 = (
+            self.base_url + '?product=WaterWolf&version=19.0&_facets_size=100'
+        )
+
```
peterbe: Unless there's a really obvious reason for doing it this way, it's it better to let `self.client` take care of turning these variables into query strings? E.g. `self.client.get(self.base_url, {'product': 'WaterWolf'})`

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -0,0 +1,302 @@
...
+    default_date_range = datetime.timedelta(days=7)
+
+    greater_than = None
+    lower_than = None
+
+    if not parameters.get('date'):
+        lower_than = timezone.now()
+        greater_than = lower_than - default_date_range
+    else:
+        for param in parameters['date']:
```
peterbe: Is this used today? I don't see a way to send in a date parameter.

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -0,0 +1,302 @@
...
+    '''Return the results of a search. '''
+    results = []
+
+    params = kwargs
+    params['_aggs.signature'] = [
+        'platform',
+        'is_garbage_collecting',
+        'hang_type',
+        'process_type',
+    ]
```
peterbe: The magic sauce! I'll try to remember this. But probably will forget and come begging for your help later.

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -0,0 +1,302 @@
...
+            for row in sig_process:
+                if row['term'].lower() == 'plugin':
+                    hit['plugin_count'] = row['count']
+
+            # Number of hang crashes.
+            hit['hang_count'] = 0
+
+            sig_hang = hit['facets']['hang_type']
+            for row in sig_hang:
+                if row['term'] in (1, -1):
```
peterbe: This deserves a comment.

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -0,0 +1,302 @@
...
+                    hit['hang_count'] = row['count']
+
+        # Run the same query but for the previous date range, so we can
+        # compare the rankings and show rank changes.
+        dates = get_date_boundaries(params)
+        delta = (dates[1] - dates[0]) * 2
+        params['date'] = [
+            '>=' + (dates[1] - delta).isoformat(),
+            '<=' + dates[0].isoformat()
```
peterbe: Shouldn't that be `<`? That's normally what "between" means when it comes to date ranges.

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -0,0 +1,302 @@
...
+        ]
+
+        previous_range_results = api.get(**params)
+        total = previous_range_results['total']
+
+        compare_signatures = {}
+        if (total > 0 and 'signature' in previous_range_results['facets']):
```
peterbe: brackets

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -0,0 +1,302 @@
...
+            signatures = previous_range_results['facets']['signature']
+            for i, hit in enumerate(signatures):
+                compare_signatures[hit['term']] = {
+                    'count': hit['count'],
+                    'rank': i + 1,
+                    'percent': 1.0 * hit['count'] / total
```
peterbe: Do '100.0 * hit['count'] / total` here instead and it starts to make sense calling it "percent".

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -0,0 +1,302 @@
...
+                hit['previous_percent'] = sig['percent']
+            else:
+                hit['diff'] = 'new'
+                hit['rank_diff'] = 0
+                hit['previous_percent'] = 0
+
+    return search_results
+
+
+@waffle_switch('topcrashers-real-time')
```
peterbe: kill kill kill :)

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -0,0 +1,302 @@
...
+                return redirect(url)
+    else:
+        versions = versions.split(';')
+
+    if len(versions) == 1:
+        context['version'] = versions[0]
+
+    release_versions = [x['version'] for x in context['releases'][product]]
+    if context['version'] not in release_versions:
+        raise http.Http404('Unrecognized version')
```
peterbe: Ugh! I sadly recognize so much of these 20 lines above. It hurts my eyes every time. 
SOME day we'll get this simplified and beautiful again.

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -0,0 +1,302 @@
...
+    # the template. This way we avoid hardcoding it twice and
+    # have it defined in one common location.
+    context['result_counts'] = settings.TCBS_RESULT_COUNTS
+    if result_count not in context['result_counts']:
+        result_count = settings.TCBS_RESULT_COUNTS[0]
+
+    context['result_count'] = result_count
+    context['query'] = {
+        'end_date': end_date,
+        'start_date': end_date - datetime.timedelta(days=days),
```
peterbe: Neat!

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    api_results = get_topcrashers_results(
+        product=product,
+        version=context['version'],
+        platform=os_name,
+        process_type=crash_type,
+        date=[
+            '<=' + end_date.isoformat(),
+            '>' + context['query']['start_date'].isoformat()
```
peterbe: Date betweens should always be `>=` the start date and `<` the end date.

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -0,0 +1,302 @@
...
+
+    count_of_included_crashes = 0
+    signatures = []
+    for crash in tcbs[:int(result_count)]:
+        signatures.append(crash['signature'])
+        count_of_included_crashes += crash['count']
+
+    context['number_of_crashes'] = count_of_included_crashes
+    context['total_percentage'] = api_results['total'] and (
+        count_of_included_crashes / api_results['total']
```
peterbe: Why don't you worry about this being 0?

peterbe:
A few nits but it's looking great!
I'll review again when you have the "real-time toggle".

webapp-django/crashstats/topcrashers/views.py
AdrianGaudebert: I do, that's what the ``api_results['total'] and `` is for. :)

webapp-django/crashstats/topcrashers/views.py
AdrianGaudebert: No, that's just the safest way to process dates from a Super Search form. Even if it isn't used now, I'd like to keep it around so that it becomes very simple to build a more "Super Search-y" interface on top of this.

webapp-django/crashstats/topcrashers/views.py
AdrianGaudebert: I don't understand this comment. :)

AdrianGaudebert:
@peterbe now with "day time" data (and bad names for the different modes, but we can discuss that with users like KaiRo) and all comments addressed.

webapp-django/crashstats/topcrashers/views.py
peterbe: Oh I was just teasing. You don't need to use brackets in python.

AdrianGaudebert: Ahaha that's me doing too much JS those days... :)

webapp-django/crashstats/topcrashers/tests/test_views.py
```diff
@@ -0,0 +1,276 @@
...
+            # Now test the "day time" data.
+            response = self.client.get(self.base_url, {
+                'product': 'WaterWolf',
+                'version': '19.0',
+                '_tcbs_mode': 'daytime',
+            })
+            eq_(response.status_code, 200)
+            ok_(today in response.content)
+            ok_(now not in response.content)
```
peterbe: Neat!

AdrianGaudebert:
@peterbe names changed!

peterbe:
It still bothers me that it is so slow. If we don't care now and file a new bug we might never get around to doing something about it. Perhaps there's some low-hanging fruit fixes that we can easily fix. Or is all the slowness due to ES being so slow. (ES is awesome but we are issuing a complex query).
For example, we're doing two queries right? Just to get the rank difference. For the second query do we select too much maybe? Is there inefficiencies in our python code once the data comes back or is it all spent inside ES?

I talked to Kairo about it and he said the "First appearance" column is really important. I don't know who else is it important so. We tend to take Kairo's words as words being from all of CrashKill. 
Should we look into doing a separate PG to get those dates for all the signatures we intend to display?

Imperfections aside, we just simple HAVE TO HAVE this new Top Crashers. Already today people are impatient and don't want to wait 24 hours. As we add more and more non-C++ crashes we just have to be closer to real-time. So because of that, I've talked myself into r+'ing this. After all, it's not *yet* replacing any existing functionality.

AdrianGaudebert:
So do I have an r+ here, or do you want me to work on the perf issues first? I don't really now. :)

Also, local perfs are sometimes misleading, so I would like to test this on stage at least before drawing any conclusion. Not that I disagree with your numbers, just to see how it behaves in a more realistic environment.

peterbe:
Yes. r+

The performance irks me and I don't want us to lose sight of it. Performance does matter. If pages feel sluggish and annoying people won't like anything we do and that's bad all around. 

We also need to find out if the "First appearance" is important to the whole CrashKill team or if its absence makes the new report useless.
Let's also file a bug about **investigating** why it is so slow.

webapp-django/crashstats/topcrashers/tests/test_views.py
AdrianGaudebert: Heh, I forgot to do that!

PR #2937, opened 2015-08-05
fixes bug 1187986 - revert S3 calling format

peterbe:
@rhelmer r?


rhelmer:
r+

The underlying cause here seems to be that S3 is sending a redirect and Boto doesn't want to follow it, for reference the exception when using `OrdinaryCallingFormat` in AWS prod right now is:

`S3ResponseError: S3ResponseError: 301 Moved Permanently`

PR #2939, opened 2015-08-05
better errors on bucket creation errors

peterbe:
@rhelmer r? 
We're getting desperate. It's past 4pm.


rhelmer:
r+, I think we're fine to go to prod with what we have (per IRC) but we want this for debugging stage and just for future debugging.

PR #2924, opened 2015-07-29
Bug 1188087 supersearchfields directly via ssocorro

peterbe:
@AdrianGaudebert This works locally when clicking around and stuff. The tests also pass locally. 

I had to do a lot of deep debugging to get this to work and I probably have a lot of ugly code lying around. I just wanted you to get involved early and I wanted to see how tests work out on travis.

PR #2940, opened 2015-08-06
Fixes Bug 1191509 - New signature reports tab to show graphs based on date histogram aggregations

Tchanders:


Tchanders:
@AdrianGaudebert  A few things about these graphs:

I wasn't sure what default options were best, so I just went for Product for now.

I had to populate elastic search with some extra stuff here in order to get different lines to show on the graphs, so I added three extra fake signatures with c. 2, 10 and 20 products (this many products may be unrealistic, but there might be this many of some other field). Metrics Graphics only shows the first five lines by default:
screenshot from 2015-08-06 16 57 01
(You can see by the legend that there are actually c. 20 lines of data here.)

Adding classes for the extra lines does make them display, but it looks too crowded:
screenshot from 2015-08-06 16 59 34
(...in any case you still get the same problem - if you add twenty extra classes but there are thirty lines, only the first twenty will show up.)

Tchanders:
We could sort by which lines have the highest counts and just show the top 5 (but we'd have to choose between highest average and highest maximum count). Or we could just show the graph with 5 lines but let the user know that there are more lines, and let them select which they want to see?

webapp-django/crashstats/signature/static/signature/js/signature_tab_graphs.js
```diff
@@ -0,0 +1,147 @@
...
+            'text': field.text
+        }));
+    });
+
+    // Append the control elements.
+    this.$controlsElement.append(this.$selectElement, $('<hr>'));
+
+    // Set the placeholder.
+    this.$selectElement.select2({
+        'placeholder': 'Aggregate on...',
```
AdrianGaudebert: ``s/Aggregate on/Graph of``. Or something better.

webapp-django/crashstats/signature/views.py
```diff
@@ -238,6 +238,58 @@ def signature_aggregation(request, aggregation):
 
 
 @waffle_switch('signature-report')
+@utils.json_view
+def signature_graphs(request, field):
+    '''Return the aggregation of a field. '''
```
AdrianGaudebert: Needs updating.

webapp-django/crashstats/signature/views.py
```diff
@@ -238,6 +238,58 @@ def signature_aggregation(request, aggregation):
...
+
+    params['signature'] = '=' + signature
+    params['_results_number'] = 0
+    params['_results_offset'] = 0
+    params['_histogram.date'] = [field]
+
+    data['current_url'] = '%s?%s' % (
+        reverse('signature:signature_report'),
+        current_query.urlencode()
+    )
```
AdrianGaudebert: What's the point of this? Is it being used?

webapp-django/crashstats/signature/views.py
```diff
@@ -238,6 +238,58 @@ def signature_aggregation(request, aggregation):
...
+    try:
+        search_results = api.get(**params)
+    except models.BadStatusCodeError, e:
+        # We need to return the error message in some HTML form for jQuery to
+        # pick it up.
+        return http.HttpResponseBadRequest('<ul><li>%s</li></ul>' % e)
+
+    data['aggregates'] = search_results['facets']['histogram_date']
+
+    data['total_count'] = search_results['total']
```
AdrianGaudebert: Same here, this isn't being used.

AdrianGaudebert:
Let's show only the top 4 lines, and use those with the maximum total number of results. To find out which are the top results, add a facet on the field, and use the results from that facet: they will be sorted by descending number. 

For example: http://socorro/supersearch/?_histogram.date=version&_facets=version

Returns:
```json
{
  "facets": {
    "histogram_date": ["..."] ,
    "version": [
      { "term": "WaterWolf", "count": 42 },
      { "term": "NightTrain", "count": 41 },
    ]
  }
}
```

On the graph, if there are more than 4 different lines, add a note saying something like "Showing top 4 results out of XX, not showing: a, b, c... ".

webapp-django/crashstats/signature/views.py
Tchanders: Oops, this was naively copied over from the aggregations view. In fact it doesn't seem to be being used there either.

AdrianGaudebert: Feel free to remove it there as well then!

webapp-django/crashstats/signature/views.py
Tchanders: This made me realise I was checking for incorrectly whether there was data to draw the graph... Corrected now hopefully.

Tchanders:
New version with those changes.

AdrianGaudebert:
This is looking very very good! I'm excited to finally see graphs in the signature report page!!! :)

I think the only comment I have is that when you hover a particular point, the MG tip tells you the number of crashes but does not tell you what it is for (for example, if I'm looking at a graph of count of crashes per version, when I hover a point I'd like it to tell me that these numbers are for version X or Y). 

And now for the actual last comment: where are my unit tests? :)

Tchanders:
@AdrianGaudebert Here they are!

AdrianGaudebert:
Looking excellent! However, you are conflicting with @peterbe work in #2924, and he is going to merge that today. Once that PR is merged, can you please rebase this one and change your unit test code so that it matches the new way of mocking SuperSearch introduce in #2924?

Tchanders:
@AdrianGaudebert Good to squash?

AdrianGaudebert:
Yes! Excellent! I'll merge as soon as it's ready to go. :)

Tchanders:
@AdrianGaudebert Done!

AdrianGaudebert:
Heh, I just noticed: when you have several graphs and you hover a point in one of the graphs, it changes the top-right legend for all graphs! Not sure if that's a bug in your code or in MG, but that's not a blocker anyway. I'm merging this anyway, can you file a bug to figure that issue out?

Tchanders:
Oops, yes I should be able to fix that!

PR #2941, opened 2015-08-06
fixes bug 1190899 - Output crontabber status in json with a wrap-up status

peterbe:
@AdrianGaudebert r?

/cc @rhelmer @jdotpz 

So, the output is either just `{"status": "OK"}` if all is green or... something like this:
screenshot 2015-08-06 11 42 50


webapp-django/crashstats/crashstats/views.py
```diff
@@ -1763,6 +1764,44 @@ def crontabber_state(request, default_context=None):
...
 
 
+@utils.json_view
+def crontabber_status(request):
+    api = CrontabberState()
+
+    # start by assuming the status is OK which means no jobs are broken
+    context = {'status': 'OK'}
+
+    all = api.get()['state']
```
AdrianGaudebert: ``all`` is a function in the standard library of Python, I would recommend not using it. (Plus it's a terrible variable name really. :) )

https://docs.python.org/2/library/functions.html#all

webapp-django/crashstats/crashstats/views.py
```diff
@@ -1763,6 +1764,44 @@ def crontabber_state(request, default_context=None):
...
+    broken = [
+        name for name, state in all.items()
+        if state['error_count']
+    ]
+    blocked = [
+        name for name, state in all.items()
+        if set(broken) & set(state['depends_on'])
```
AdrianGaudebert: What's wrong with ``state['depends_on'] in broken``? It seems easier to understand to me, and I don't think perf matter a lot here (though I would suspect it's actually faster to just use ``in`` instead of creating 2 new sets).

webapp-django/crashstats/crashstats/views.py
```diff
@@ -1763,6 +1764,44 @@ def crontabber_state(request, default_context=None):
...
+    ]
+
+    # This infinite loop recurses deeper and deeper into the structure
+    # to find all jobs that are blocked. If we find that job X is blocked
+    # by job Y in iteration 1, we need to do another iteration to see if
+    # there are jobs that are blocked by job X (which was blocked by job Y)
+    while True:
+        also_blocked = [
+            name for name, state in all.items()
+            if name not in blocked and set(blocked) & set(state['depends_on'])
```
AdrianGaudebert: Same here, this a bit tricky to understand and I think would be easier to understand with just an ``in``. But maybe I'm actually not understanding it?

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -2659,6 +2659,59 @@ def test_crontabber_state(self):
...
+                        'depends_on': [],
+                    },
+                    'job2': {
+                        'error_count': 0,
+                        'depends_on': ['job1'],
+                    },
+                    'job3': {
+                        'error_count': 0,
+                        'depends_on': ['job2'],
+                    }
```
AdrianGaudebert: I think this would deserve to test even more things. I would add at least one non-failing test and a few more levels of dependency of failing jobs, to test that ``while`` loop in the code.

AdrianGaudebert:
Some nits but overall it's great! r+

webapp-django/crashstats/crashstats/views.py
peterbe: taken care of.

webapp-django/crashstats/crashstats/views.py
peterbe: It's about comparing one list with another list and looking for an intersection.

PR #2942, opened 2015-08-07
Fixes bug 1192012 - Rewrote setup_supersearch_app.py file to use ES1.…

AdrianGaudebert:
…4 style bulk indexing.

@peterbe r?

### DIFF
--- a/scripts/setup_supersearch_app.py
+++ b/scripts/setup_supersearch_app.py
@@ -11,6 +11,7 @@ mapping for processed and raw crashes.
 
 import json
 import os
+from elasticsearch.helpers import bulk
 
 from configman import Namespace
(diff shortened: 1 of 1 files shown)


peterbe:
How do I test this? I ran it (had to cd into the scripts directory) and it passed. It took no time at all.

AdrianGaudebert:
It tests itself, at the end of the script there's a count to ES and that number of results is compared with the number of fields in the origin file. So if there was no exception at the end of the script's log, it worked fine. Now if you want to test it in the webapp, go to the Super Search page and see that there are fields listed in the form. Or open the admin panel, go the Super Search Fields page and verify that the content there is correct.

peterbe:
r+ then

PR #2944, opened 2015-08-11
Fixes Bug 1193367 - IndexError in signature_graph_data view

Tchanders:


Tchanders:
@AdrianGaudebert Do you mean like this?

AdrianGaudebert:
That's one way of solving the problem, it makes the django view return a 400 error instead of a 500 (and thus we do not receive an alert about it), but I think you also want to prevent the JS script to send a query if the product is missing (which happens at the moment, causing the error). I would keep both things in place (what you currently have, + a check to not send the query from the client if there's no product).

Tchanders:
I see, it was checking but it was making the request regardless. Is this check sufficient, or do you think there should be an error raised?

AdrianGaudebert:
r+

PR #2945, opened 2015-08-11
fixes bug 1193536 - re-encode old .sym files

peterbe:
@rhelmer r?

I've successfully run this on my laptop using my personal S3 account. 
A bunch of URLs that that couldn't be opened in the browser before [can now be opened in the browser](https://s3.amazonaws.com/peterbe.org.allizom.crash-stats.staging.crashes/v1/AccessibleMarshal.pdb/83C8F71548914DEFB5A49C880BCDFD3B1/AccessibleMarshal.sym)

Once this lands we'll need to run this from somewhere on a stage node. You start the script with:

```
./manage.py reencode-bad-sym-files
```

By default it will attempt go through every file uploaded. So it'll that a freakishly long time. An alternative is to add `--max-uploads=3` (for example). That'll pick 3 random uploads and go through each file in those.

### DIFF
--- a/webapp-django/crashstats/symbols/management/commands/reencode-bad-sym-files.py
+++ b/webapp-django/crashstats/symbols/management/commands/reencode-bad-sym-files.py
@@ -1,0 +1,132 @@
+import gzip
+from cStringIO import StringIO
+
+from optparse import make_option
+
+import requests
+import boto.s3.connection
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+from django.template.defaultfilters import filesizeformat
+from django.utils import timezone
+
+from crashstats.symbols.models import SymbolsUpload
+
+
+class Command(BaseCommand):
+    """
+    On July 20 2015 we released an improvement to the symbols uploader so
+    that all .sym files in the zip archive bundles would be gzipped.
+    That makes them smaller.
+    However, we made a mistake in that we didn't encode the file correctly,
+    just encoding the content.
+    To correct this we have this script.
+    """
+
+    help = (
+        'Run this command to re-encode .sym files that were uploaded '
+        'without first being properly encoded with the gzip header signature.'
+    )
+
+    option_list = BaseCommand.option_list + (
+        make_option(
+            '--dry-run',
+            action='store_true',
+            dest='dry_run',
+            default=False,
+            help='Only find the files to re-encode'
+        ),
+        make_option(
+            '--first-date',
+            dest='first_date',
+            default='2015-07-20',  # the day we started compressing (132)
+            help='Upload date range start'
+        ),
+        make_option(
+            '--end-date',
+            dest='end_date',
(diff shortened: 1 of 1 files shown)


webapp-django/crashstats/symbols/management/commands/reencode-bad-sym-files.py
```diff
@@ -0,0 +1,132 @@
...
+        ),
+        make_option(
+            '--max-uploads',
+            dest='max_uploads',
+            default='0',
+            help='Max. number of SymbolsUploads to work on (default all)'
+        ),
+    )
+
+    def _parse_date(self, date):
```
rhelmer: why the custom date parsing code? We have `python-dateutil` in `requirements.txt` so why not:
```
dateutil.parser.parse()
```
?

If you want to stick to stdlib for some reason why not `datetime.datetime.strptime`?

webapp-django/crashstats/symbols/management/commands/reencode-bad-sym-files.py
```diff
@@ -0,0 +1,132 @@
...
+                        if not options['dry_run']:
+                            downloaded = key.get_contents_as_string()
+                            real_content = downloaded.decode('zlib')
+                            out = StringIO()
+                            with gzip.GzipFile(fileobj=out, mode='w') as f:
+                                f.write(real_content)
+                            value = out.getvalue()
+                            key.set_contents_from_string(value, {
+                                'Content-Encoding': 'gzip'
+                            })
```
rhelmer: How many objects are we talking about? An alternative fix that doesn't require downloading/re-encoding/uploading would be to just set `Content-Encoding: deflate`

I think your approach is cleaner overall, it'd be less confusing in the future.

rhelmer:
lgtm, curious why you used custom date parsing code though.

rhelmer:
Looks like we *also* have `iso8601` in `requirements.txt`, so could use that instead as well.

webapp-django/crashstats/symbols/management/commands/reencode-bad-sym-files.py
peterbe: It was to get the timezone information into it. That's what django.utils.timezone.now does.

webapp-django/crashstats/symbols/management/commands/reencode-bad-sym-files.py
peterbe: I don't know they that deflate thing works but it feels like a hack on top of a hack. We should encode the files properly.

peterbe:
Perhaps the right thing to do would have been to use a proper date parser but it's quite immaterial since we'll soon delete this script. And besides, we'll just use the default dates.

I can't tell how many uploads there were on prod between these dates but on stage it's 1,449 uploads!! And 467,868 individual `.sym` files. 

I think we should be careful with running this script and use the `--max-uploads=10` or something and see how long it takes on stage.

PR #2947, opened 2015-08-12
Adding Rank to ProdVersions

yglazko:
Trying to add a 'rank column' to Prod Versions in Signature Summary. Easy fix, good for relman.

### DIFF
--- a/webapp-django/crashstats/crashstats/views.py
+++ b/webapp-django/crashstats/crashstats/views.py
@@ -2019,7 +2019,8 @@ def signature_summary(request):
             'product': r['product_name'],
             'version': r['version_string'],
             'percentage': format_float(r['percentage']),
-            'numberOfCrashes': r['report_count']})
+            'numberOfCrashes': r['report_count'],
+            'rank': r.get('currentRank', '')})
     for r in result['uptimeRange']:
         signature_summary['uptimeRange'].append({
             'range': r['category'],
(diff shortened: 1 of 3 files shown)


rhelmer:
Thanks! The diff here looks good to me.

Can you please squash these commits into one?
Also is there a bugzilla bug #? If so mention it in the commit message.

yglazko:
Does this commit work? https://github.com/yglazko/socorro/commit/3a948fd4d637287c29c45fc21ae94521c4168cba

webapp-django/crashstats/crashstats/views.py
```diff
@@ -2019,7 +2019,8 @@ def format_float(number):
             'product': r['product_name'],
             'version': r['version_string'],
             'percentage': format_float(r['percentage']),
-            'numberOfCrashes': r['report_count']})
+            'numberOfCrashes': r['report_count'],
+            'rank': r.get('currentRank', '')})
```
peterbe: I can't find a single place where this is ever present. I clicked on a bunch of signatures under Firefox 41 and it's never there. I think I need to dig deeper to try to understand what you tried achieve.

peterbe: The query that gets the data to here is this one: https://github.com/mozilla/socorro/blob/master/socorro/external/postgresql/signature_summary.py#L152-L179
No `currentRank`.

peterbe: `currentRank` is a TCBS thing (Top Crashers by Signature). You're trying to use data from the signature summary.

peterbe:
I think this PR needs some more thought. It can't work.

yglazko:
Closing after discussion with Kairo, going to do something with the crash volume /overall volume instead. Thank you!

PR #2949, opened 2015-08-13
fixes bug 1132652 - Windows 10, not Windows Unknown

peterbe:
@twobraids r?

cc @selenamarie @rhelmer 

The three operations in the migration have plenty of comments that explain what's going on. Helpfully :)

Ultimately, what I'm trying to achieve is fixing [a page like this](https://crash-stats.allizom.org/report/list?product=Firefox&range_value=7&range_unit=days&date=2015-08-12&signature=nsACString_internal%3A%3AReplace%28unsigned+int%2C+unsigned+int%2C+char+const*%2C+unsigned+int%29&version=Firefox%3A41.0a2#tab-sigsummary) where it says "Windows Unknown" in the "Operating System" signature summary box. 

One thing that is still a bit of a mystery to me is why we have so few minor versions of different major windows versions. 

```
breakpad=> select * from windows_versions ;
 windows_version_name | major_version | minor_version
----------------------+---------------+---------------
 Windows NT           |             3 |             5
 Windows NT           |             4 |             0
 Windows 98           |             4 |             1
 Windows Me           |             4 |             9
 Windows 2000         |             5 |             0
 Windows XP           |             5 |             1
 Windows Vista        |             6 |             0
 Windows 7            |             6 |             1
 Windows 8            |             6 |             2
 Windows 8.1          |             6 |             3
 Windows 10           |            10 |             0
(11 rows)
```
(this is stage data)

Why only 2 minor versions at most of Windows NT?! Were there not many more minor versions of Windows during the long time that these operating systems have existed??


KaiRo-at:
Why the different minor versions there? It should all be 10.0 AFAIK, so minor version should always be 0 for Win10 right now, right? Or did you see something else in examples?

peterbe:
I don't know what inserts into `os_versions` but there's records in there under three different minor numbers:
```
breakpad=> select * from os_versions where os_name = 'Windows' and major_version = 10;
 os_version_id | os_name | major_version | minor_version | os_version_string
---------------+---------+---------------+---------------+-------------------
            85 | Windows |            10 |             0 | Windows Unknown
           105 | Windows |            10 |             5 | Windows Unknown
           114 | Windows |            10 |             4 | Windows Unknown
(3 rows)
```
(again, this is stage data)
Intestingly now reports in `reports_clean` depend on those other ones
```
breakpad=> select count(*) from reports_clean where os_version_id=85;
 count
--------
 184680
(1 row)

breakpad=> select count(*) from reports_clean where os_version_id=105;
 count
-------
     0
(1 row)

breakpad=> select count(*) from reports_clean where os_version_id=114;
 count
-------
     0
(1 row)
```

So, what are these minor version 4 and 5?!?
Can we just delete them and just proceed to correct the existing ones with:
```
UPDATE os_versions
        SET
        os_version_string = 'Windows 10'
        WHERE
        os_version_string = 'Windows Unknown' AND
        os_name = 'Windows' AND
        major_version = 10
```
and also fix the old `signature_summary_os` records.

KaiRo-at:
I would vote for killing or ignoring those 4 and 5 minor versions, yes. The 0 is all we care about at this time.

peterbe:
I'll re-write the PR. Kairo has me convinced.

peterbe:
@twobraids I've changed this PR now so that it doesn't mess around with those stray minor versions. 
It just corrects the table `os_versions` and `signature_summary_os`.

I think I know what happened. The `create_os_version_string` SQL function makes an entry to the `os_versions` table based on what's in the `windows_versions` table. That latter table, Selena updated about 5 months ago. But it was too late. An entry in `os_versions` had already been created with `os_name='Windows'` and `major_version=10` that was set to `os_version_string='Windows Unknown'`. 
Now we need to correct that past.

alembic/versions/4afd4e13281d_correct_windows_unknown_to_windows_10.py
```diff
@@ -0,0 +1,65 @@
...
+    # auto-heal but there's really no good way of re-running it for all
+    # past computations. Instead of re-running those, let's just fix it
+    # straight away.
+    op.execute("""
+        UPDATE signature_summary_os
+        SET
+        os_version_string = 'Windows 10'
+        WHERE
+        os_version_string = 'Windows Unknown'
+    """)
```
twobraids: this will touch 296032 entries in the signature_summary_os table.  It could be that the table is seriously in need of truncation of old unimportant data.  However, if this suddenly associates nearly 300K signatures with "Windows 10" and this table is used for further aggregations, then we have the potential to adversely skew the results.  My question, is this a safe thing to do?

peterbe: > My question, is this a safe thing to do?

I don't know :)
I'm totally new to this and just decided to dive right in. 
I searched and searched and found no places where the `signature_summary_os` is being referenced (other than the SQL SELECT that is used to display its content)

peterbe: Wow! on stage...
```
breakpad=> select count(*) from signature_summary_os;
  count
----------
 64531792
(1 row)
```
That's ~65M rows! Why?!

peterbe: Where else is the signature summary used? Does anybody know? @rhelmer @AdrianGaudebert 

When you click on a signature in Top Crashers it loads the signature summary by doing a query that filters on
* signature
* product
* product version
* start date
* end date

The start- and end dates are usually something like 7 days ago and today. Or 28 days ago and today. 
That makes me wonder, when do we ever benefit from `signature_summary_os` data whose `report_date < NOW() - interval(28 days)` if you see what I mean?

peterbe: Should we set up a cron job that truncates these enormous tables?
Should we put our weight behind re-writing all of these into run-time ES queries? [Apparently we need to do something to the ES data first](https://bugzilla.mozilla.org/show_bug.cgi?id=1181650#c2)

peterbe: Also, given that we have good caching of the implementations/middleware, how about doing these queries straight on the `reports_clean` table? Would it be horribly slow and expensive?

twobraids: I strongly suspect that without @selenamarie, much of the day to day PG maintenance is just not getting done.  I wonder if we could do this: 

1) the ```signature_summary_os``` table could just be truncated entirely
2) make the "windows 10" change to the ```os_versions``` table
3) backfill the ```signature_summary_os``` table to the max day in the past that we need data

That would ensure that there are no changes to that would associate Windows 10 with some signature/os_version that is legitimately "Windows Unknown"

peterbe: So, a migration that first delete a `DELETE FROM signature_summary_os`?
But how do we do a backfill in the migration? Do we simply call `backfill_weekly_report_partitions`?
I guess we could do something like this:

```python
    now = datetime.datetime.utcnow()
    start = (now - datetime.timedelta(days=28)).strftime('%Y-%m-%d')
    end = now.strftime('%Y-%m-%d')
    op.execute("""
        SELECT backfill_weekly_report_partitions(%s, %s, 'signature_summary_os')
    """, (start, end))
```

twobraids: perhaps a migration is not the best way to do this.  I see migrations as changes to structure of the database schema, not adjustments to the data within.  Is there precedent for adjusting data in using the migration mechanism?  

I suspect this would be a one-off adjustment to the data.

peterbe: I quickly skimmed through the last couple of versions of migrations and it's a mix of actually changing tables & columns and just running various data ops.

peterbe:
For the record, we will do...

* change os_versions
* delete all from signature_summary_os
* re-apply the signature_summary_os backfilling'ly for the past 28 days but do a PG commit between each date.

I'll re-work the PR as soon as time allows.

PR #2950, opened 2015-08-14
Fixes Bug 1168511 (again) - add ability to test keys for null bytes

twobraids:
it turns out that the null string can occurr in both the key and the value.  The previous fix for this problem only focused on the values for the keys, not the keys themselves.  

This fixes that oversight.


rhelmer:
r+

PR #2952, opened 2015-08-16
Fixes Bug 1194306 - Style signature page panel delete buttons on hover

Tchanders:


Tchanders:
@AdrianGaudebert Made this a div instead of a link

PR #2954, opened 2015-08-17
Bug 1184173 - Added missing hash for freezegun package.

AdrianGaudebert:
@jdotpz r? This should fix the build.


jdotpz:
Arrrr+
As soon as Travis says yay, at least.

PR #2955, opened 2015-08-17
fixes bug 1195345 - bad error on supersearch when the date range is too large

peterbe:
@AdrianGaudebert r?


AdrianGaudebert:
r+

PR #2956, opened 2015-08-17
fixes bug 1195406 - BadArgumentErrors need to become 400 Bad Requests

peterbe:
@AdrianGaudebert r?

So when I started I thought I'd apply this to that little `SocorroMiddleware.fetch` method that actually makes the method call to a thing like `socorro.external.es.supersearch.SuperSearch.get` and at that point immediately convert the call to a 400 Bad Request. 

But then I changed my mind. If someone uses the public web API and supply some nasty value (e.g. `date=2015-02-32`) that our first line of defense can't catch and it goes into the implementation classes then it was the user's fault for supplying bad data. 

However, if it happens in a django view (e.g. `report_list()`) or the super search django view, the the user didn't type things in. Most likely they just click things. Granted, on the super search UX you *do* type in values. However, **that** we should catch with our first-line defense. I.e. the form validation. And if the user types in "2015-02-32" we should display a soft form validation error. If that doesn't protect from bad queries down to the implementation we have failed in protecting it and we should deal with it as it happens.

### DIFF
--- a/webapp-django/crashstats/api/tests/test_views.py
+++ b/webapp-django/crashstats/api/tests/test_views.py
@@ -9,6 +9,7 @@ from django.conf import settings
 import mock
 from nose.tools import eq_, ok_
 
+from socorro.external import BadArgumentError, MissingArgumentError
 from crashstats.base.tests.testbase import TestCase
 from crashstats.crashstats.tests.test_views import (
     BaseTestViews,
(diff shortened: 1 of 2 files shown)


peterbe:
Also, sorry for editing something unrelated. I edited that first because I wrote my new test based on copy the one I changed. It was a hangover miss from last weeks monster patch to re-write how mocking is done of supersearch so I decided to throw caution to the wind and let it sneak in.

AdrianGaudebert:
r+

PR #2957, opened 2015-08-18
more Fixes Bug 1168511 - separate unicode & str cases in cleaning of null bytes.

twobraids:
the collector bad bytes cleanup is failing on unicode/str conversions.  separate the cases so that can be more controlled in the cleanup of bad bytes.


peterbe:
r+

PR #2958, opened 2015-08-18
fixes bug 1195920 - django 1.7.10

peterbe:
@AdrianGaudebert @lonnen r?

I installed it on my local laptop, was able to sign out and sign in again.


AdrianGaudebert:
r+

PR #2961, opened 2015-08-20
Fixes Bug 1119431 - Create a profile page for a signed in user

Tchanders:


Tchanders:
@AdrianGaudebert Here you go! I wasn't quite sure what to call it - have switched between "your details" and "landing page". Also it's still including the css and js files from the previous apps - I wasn't quite sure how separate to make it yet.

AdrianGaudebert:
So this makes me wonder about a few things. The page name doesn't sound right to me. I would go for something like "profile" (it's the best name I can think of). "Landing page" doesn't mean much, and is just a term that we use to describe it because we want it to be the page that users will see by default. 

I also wonder if the API Tokens part should actually be there. It seems out-of-place to me on such a page. Notably because it doesn't just list tokens but allows one to create or delete them. Imo that should stay in a separate page. 

@lonnen what do you think? Here's a screenshot: 

landing-page

Tchanders:
How about keeping "Your generated tokens" and having a link to "Manage tokens"?

AdrianGaudebert:
* rename the app to ``profile``
* rename the url to ``profile/``
* change the heading to ``Your Profile``
* for tokens, keep only the list and add a link to the tokens page: ``Manage your tokens``
* add a link to the admin panel if user is a superuser
* move the email address to the header (out of "Your permissions")
* all data about tokens should fit in one single panel (titled ``API Tokens``)

AdrianGaudebert:
* remove link from the footer

Tchanders:
@AdrianGaudebert Here's the latest version with those changes, and have also neatened up the design a bit. (The view can be simplified now, but thought I'd ask for your feedback on the new look first...)

Tchanders:
Hi @peterbe, Adrian asked me to cc you in on this in case you have any thoughts. I'll be working with him tomorrow and I think he's hoping to merge then if possible.

webapp-django/crashstats/profile/views.py
```diff
@@ -0,0 +1,100 @@
...
+            # avoid the rate limiter.
+            form = None
+
+    context['form'] = form
+    context['your_tokens'] = (
+        models.Token.objects
+        .filter(user=request.user)
+        .order_by('-created')
+    )
+    context['absolute_base_url'] = (
```
peterbe: What is this for?

webapp-django/crashstats/profile/templates/profile/profile.html
```diff
@@ -0,0 +1,198 @@
...
+                        </tr>
+                        {% else %}
+                        <tr>
+                            <th>Expires:</th>
+                            <td>
+                                {{ token.expires.strftime('%Y-%m-%d %H:%M:%S%Z') }}
+                                (<time class="in" data-date="{{ token.expires.isoformat() }}">{{ token.expires.isoformat() }}</time> from now)
+                            </td>
+                        </tr>
+                        {% endif %}
```
peterbe: This block is a bit repeated. You could do something like this:
```
<tr>
  <th>{% if token.is_expired %}Expired:{% else %} Expires:{% endif %}</th>
   ...
```
if you see what I mean. Line163 and 171 is repeated.

webapp-django/crashstats/profile/templates/profile/profile.html
```diff
@@ -0,0 +1,198 @@
...
+                                    {{ token.notes | nl2br }}
+                                {% else %}
+                                    <i>no notes</i>
+                                {% endif %}
+                            </td>
+                        </tr>
+                    </table>
+                </div>
+                {% else %}
+                <p>You currently have no tokens generated (<a href="{{ url('tokens:home') }}">manage</a>).</p>
```
peterbe: "manage" isn't super useful. It would be better with something like "Manage your API tokens"

webapp-django/crashstats/profile/templates/profile/profile.html
```diff
@@ -0,0 +1,198 @@
...
+            </tbody>
+        </table>
+        {% endif %}
+        </div>
+    </div>
+
+    <div class="page-section">
+        <div class="panel">
+            <div class="title">
+                <h2>Your crash reports</h2>
```
peterbe: The title of the page is "Your Profile" (with a capital P). Then this should be "Your Crash Reports". Or, the title should maybe be "Your profile". Either way, it should be consistent.

webapp-django/crashstats/profile/templates/profile/profile.html
```diff
@@ -0,0 +1,198 @@
...
+
+{% endblock %}
+
+{% block site_js %}
+    {{ super() }}
+
+<script src="{{ static('crashstats/js/moment.min.js') }}"></script>
+
+    {% compress js %}
+<script type="text/javascript" src="{{ static('crashstats/js/socorro/your_crashes.js') }}"></script>
```
peterbe: Was this file added some other time?

webapp-django/crashstats/profile/templates/profile/profile.html
```diff
@@ -0,0 +1,198 @@
...
+    </div>
+
+    <div class="panel">
+        <div class="title">
+            <h2>Your permissions</h2>
+        </div>
+        <div class="body">
+        {% if request.user.is_superuser %}
+            <p>
+                You are a <b>superuser</b>. You have <b>unrestricted access to everything</b> on the <a href="{{ url('manage:home') }}">admin</a> page.
```
peterbe: Can we make "admin page" the label for the link. Not just "admin".

webapp-django/crashstats/profile/templates/profile/profile.html
```diff
@@ -0,0 +1,198 @@
...
+
+ --><div class="page-section">
+        <div class="panel">
+            <div class="title">
+                <h2>API Tokens</h2>
+            </div>
+
+            <div class="body">
+                {% if your_tokens %}
+                <p>Here is a list of your currently generated tokens (<a href="{{ url('tokens:home') }}">manage</a>):</p>
```
peterbe: Same here.

webapp-django/crashstats/profile/views.py
```diff
@@ -0,0 +1,100 @@
...
+
+from crashstats.crashstats.decorators import pass_default_context
+from crashstats.supersearch.models import SuperSearchUnredacted
+from crashstats.tokens import models
+from crashstats.tokens import forms
+
+
+@pass_default_context
+@login_required
+@transaction.atomic
```
peterbe: No need for this since this view isn't changing any data.

webapp-django/crashstats/profile/views.py
```diff
@@ -0,0 +1,100 @@
...
+def profile(request, default_context=None):
+    context = default_context or {}
+    context['permissions'] = (
+        Permission.objects.filter(content_type__model='')
+        .order_by('name')
+    )
+
+    one_month_ago = (
```
peterbe: Ideally you don't want to be tied to a value in a variable name. Can you call it something else? Also, 4 weeks is not a month. The concept of "One month ago" can't exist without context.

webapp-django/crashstats/profile/views.py
peterbe: Oh. I was wrong. This view DOES create data. 
However, I think it's a mistake to repeat the code for creating tokens here. It should be done over in the API Tokens view.

webapp-django/crashstats/urls.py
```diff
@@ -40,6 +40,10 @@
     # if we ever use the Django admin we might want to change this URL
     (r'^admin/', include('crashstats.manage.urls', namespace='manage')),
     (r'', include('django_browserid.urls')),
+    (r'^profile/', include(
+        'crashstats.profile.urls',
+        namespace='profile'
```
peterbe: Excellent! Thank you for making it another app.

webapp-django/crashstats/profile/views.py
```diff
@@ -0,0 +1,100 @@
...
+            # If you *have* permissions, you can actually create a
+            # token without selecting *any* permissions. The point of
+            # that is to avoid the rate limiter.
+            # If you don't have any permissions attached to your user
+            # account means you haven't been hand curated by any
+            # administrator and if that's the case you shouldn't be able
+            # avoid the rate limiter.
+            form = None
+
+    context['form'] = form
```
peterbe: Is this used?

peterbe:
Hi @Tchanders !
I just did a quick fly-by on some low-hanging fruit. I haven't tested the branch on my laptop. (no good excuse other than being "busy")

It seems that the `your_crashes.js` is busy. 

There are quite a few nits to address. 

Also, For this page to be successful (and this might be further down the line) I think we should start doing some client-side-only tracking of what reports, products and versions the user is interested in. 
E.g. If I click "Top Crashers" for Firefox 43, I think there should be a section like "Your favorite products" that lists Top Crashers, Top Changers, etc. 
We can do this by putting in a little `localStorage` that pushes to an array which indicates which products and versions you're interested in. Or which signatures you're interested in. 
All of that could wait of course.

webapp-django/crashstats/profile/templates/profile/profile.html
AdrianGaudebert: Yeah it's being used in the ``your crashes`` page, though as that one is doomed to disappear, I wouldn't bother optimizing there and compress this file with others.

Tchanders:
Screenshots:

screenshot from 2015-08-25 10 12 51
screenshot from 2015-08-25 10 13 41

AdrianGaudebert:
Please make the email address bold.

webapp-django/crashstats/profile/templates/profile/profile.html
Tchanders: This is the only change I didn't make, since lines 164 and 172 would have to be if-else as well (the different classes make the text around the dates different).

Tchanders:
Thanks for your really helpful feedback @peterbe. I've made those changes (and more from @AdrianGaudebert).

Yes Adrian also mentioned local storage too, I agree it would enhance user experience a lot.

peterbe:
@Tchanders What's the status of things? I know Adrian is on vacation this week. When does your internship end? Are you pinging me so that you can complete this PR before the end of the week?

I like the idea of a javascript solution to "track" what the user is interested so we can provide useful links. But all of that can wait I guess. 

What I'm deeply pessimistic about is the repetition of the "Your API Tokens" here on this page. If we're going to have it there, then we should remove it from the other old page.

Tchanders:
@peterbe My internship ends today, but I'll still keep an eye on this. As far as I know this is in Adrian's hands - I was mainly pinging you to say thanks.

Everything on this page is repetitions at the moment. The idea is to introduce it gradually (there is no link to it from anywhere else), so the old "permissions" and "your crashes" pages, are still there - as is the summary of tokens on the "tokens" page. These will eventually be replaced by the new page.

peterbe:
@Tchanders In that case, THANK YOU!!! for your contributions these last couple of months. 
Please do keep an eye on us and remember that we will always be open source and a non-profit :)

I'll chat more to Adrian next week when he comes back.

webapp-django/crashstats/profile/tests/test_views.py
```diff
@@ -0,0 +1,125 @@
...
+from crashstats.supersearch.models import SuperSearchFields, SuperSearch
+from crashstats.crashstats.tests.test_views import BaseTestViews
+
+
+class TestViews(BaseTestViews):
+
+    def _login(self):
+        user = User.objects.create_user('test', 'test@mozilla.com', 'secret')
+        assert self.client.login(username='test', password='secret')
+        return user
```
AdrianGaudebert: This is already in the base class, there is no need to repeat it here.

webapp-django/crashstats/profile/tests/test_views.py
```diff
@@ -0,0 +1,125 @@
...
+                    'query_type': 'string',
+                    'namespace': 'processed_crash',
+                    'form_field_choices': None,
+                    'permissions_needed': ['crashstats.view_pii'],
+                    'default_value': None,
+                    'is_exposed': True,
+                    'is_returned': True,
+                    'is_mandatory': False,
+                }
+            }
```
AdrianGaudebert: You don't need to change this I think, the mocking done by default in the base class should be enough for you. Though maybe you are doing something specific here, but if not I would try to remove this to make the test smaller.

webapp-django/crashstats/profile/tests/test_views.py
```diff
@@ -0,0 +1,125 @@
...
+        # Now log in for the remaining tests.
+        user = self._login()
+
+        # Test with results and check email is there.
+        response = self.client.get(url)
+        eq_(response.status_code, 200)
+        ok_(user.email in response.content)
+        ok_('1234abcd-ef56-7890-ab12-abcdef130801' in response.content)
+        ok_('1234abcd-ef56-7890-ab12-abcdef130802' in response.content)
+        ok_('test@mozilla.com' in response.content)
```
AdrianGaudebert: This is the same as testing ``user.email``.

AdrianGaudebert:
A few things to do on the tests, and then it's good to go! If you have some time to finish it that would be super cool, otherwise I'll take care of it next week. Thanks Thalia!

Tchanders:
@AdrianGaudebert Thanks, have made those changes.

AdrianGaudebert:
And there we merge! Good job @Tchanders ! :)

PR #2963, opened 2015-08-20
fixes bug 1196751 - Implementation instantiation gets overly cached

peterbe:
@AdrianGaudebert r?

Note the left-over comment in the code with the print statement :)
I'll rebase that away when I get the r+.

### DIFF
--- a/webapp-django/crashstats/crashstats/tests/test_views.py
+++ b/webapp-django/crashstats/crashstats/tests/test_views.py
@@ -453,7 +453,7 @@ class BaseTestViews(DjangoTestCase):
         # the same request. This is great for low-level performance but
         # it makes it impossible to test classes that are imported only
         # once like they are in unit test running.
-        SocorroCommon._implementations = {}
+        SocorroCommon.clear_implementations_cache()
 
     def _login(self):
         user = User.objects.create_user('test', 'test@mozilla.com', 'secret')
--- a/webapp-django/crashstats/manage/views.py
+++ b/webapp-django/crashstats/manage/views.py
@@ -30,6 +30,7 @@ from crashstats.crashstats.models import (
     Platforms
 )
 from crashstats.supersearch.models import (
+    SuperSearch,
     SuperSearchField,
     SuperSearchFields,
     SuperSearchMissingFields,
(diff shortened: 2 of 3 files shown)


webapp-django/crashstats/manage/views.py
```diff
@@ -638,6 +638,7 @@ def supersearch_field_create(request):
 
     # Refresh the cache for the fields service.
     SuperSearchFields().get(refresh_cache=True)
+    SuperSearchField.clear_implementations_cache()
```
AdrianGaudebert: ``SuperSearchFields``, not ``SuperSearchField``. :)

AdrianGaudebert:
So this doesn't work because of that missing ``s`` in the class name. Here's how I test it: 

1. go to Super Search and run any search
2. go to admin panel and create a new field (easy way: go to missing fields and add any of those, make sure it has "is returned" and "is exposed" checked)
3. go back to Super Search and run a search with that new field in either the "facets" or "columns" lists
4. verify it does show that new field's data

webapp-django/crashstats/manage/views.py
peterbe: Shouldn't matter. But I'll change it.

peterbe:
@AdrianGaudebert Sorry. It took me a while but now this should be ready for one last review.

AdrianGaudebert:
r+!

socorro/external/es/supersearch.py
```diff
@@ -27,6 +27,12 @@ def __init__(self, *args, **kwargs):
         )
 
         self.all_fields = SuperSearchFields(config=self.config).get_fields()
+        # XXX I'm deliberately leaving this comment in for you Adrian
+        # so when you try to do a manual testing you should see this
+        # print statement happen before AND AFTER you do a super search search
+        # after you've made any super search field change.
+        print "IN socorro.external.es.SUPERSEARCH get all_fields",
+        print len(self.all_fields)
```
AdrianGaudebert: don't forget to remove this :)

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -453,7 +453,8 @@ def tearDown(self):
         # the same request. This is great for low-level performance but
         # it makes it impossible to test classes that are imported only
         # once like they are in unit test running.
-        SocorroCommon._implementations = {}
+        SocorroCommon.clear_implementations_cache()
+        # _implementations = {}
```
AdrianGaudebert: ... and while you're at it I guess you can remove this as well.

PR #2966, opened 2015-08-25
fixes bug 1198528 - status page tries to display python data as json

peterbe:
@AdrianGaudebert r?

It's not a security thing because you can't have access to what gets printed there.


AdrianGaudebert:
Exceptional r+ from the pool. :)

peterbe:
Ha! Go back to the pool!

PR #2967, opened 2015-08-26
fixes bug 1198533 - More info about failed crontabber jobs

peterbe:
@lonnen r?

It looks like this:
screenshot 2015-08-26 09 54 45

Stage wasn't reporting any errors at the moment so I had to fake the json. That's why I added some `<` and `>` to make sure those got escaped properly.

### DIFF
--- a/webapp-django/crashstats/crashstats/static/crashstats/css/crontabber_state.css
+++ b/webapp-django/crashstats/crashstats/static/crashstats/css/crontabber_state.css
@@ -31,6 +31,9 @@ div.panel.ongoing {
     margin: 0 auto;
     display: none;
 }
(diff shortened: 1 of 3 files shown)


lonnen:
Do the errors have a consistent schema? I'm sure we can have slightly nicer formatting than a json blob in the page

peterbe:
> Do the errors have a consistent schema? I'm sure we can have slightly nicer formatting than a json blob in the page

As of today, the `last_error` is always a dict that has the keys `traceback`, `exc_type`, `exc_value` but I just didn't feel it worth it to do something fancier. Keeping it simple as a JSON should do for now.

PR #2968, opened 2015-08-26
fixes bug 1132652 - Windows 10, not Windows Unknown

peterbe:
@twobraids r?

cc @selenamarie 

Because I had closed https://github.com/mozilla/socorro/pull/2949 and then rebased this branch, it couldn't re-open the pull request.

### DIFF
--- a/alembic/versions/4afd4e13281d_correct_windows_unknown_to_windows_10.py
+++ b/alembic/versions/4afd4e13281d_correct_windows_unknown_to_windows_10.py
@@ -1,0 +1,73 @@
+"""correct 'Windows Unknown' records
+
+Revision ID: 4afd4e13281d
+Revises: b99155654de
+Create Date: 2015-08-13 13:42:53.101703
+
+Aims to solve: https://bugzilla.mozilla.org/show_bug.cgi?id=1132652
+which was re-opened.
+"""
+import datetime
+
+# revision identifiers, used by Alembic.
+revision = '4afd4e13281d'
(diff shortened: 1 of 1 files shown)


peterbe:
One thing I admit not knowing very well, is partitions work. I'm assuming it all pretty automated.

alembic/versions/4afd4e13281d_correct_windows_unknown_to_windows_10.py
```diff
@@ -0,0 +1,68 @@
...
+    """)
+
+    backfill_signature_summary_os()
+
+
+def backfill_signature_summary_os():
+    # We only ever, at most, show 28 days in the report list that
+    # shows this signature summary.
+    op.execute("""
+        DELETE FROM signature_summary_os
```
twobraids: Instead of DELETE, use TRUNCATE.  The latter doesn't index scan the table so it is faster and it reclaims memory without a need for VACUUM.

peterbe:
Now using `TRUNCATE` instead.

peterbe:
I've removed the `TRUNCATE` now.

twobraids:
well, lets see if I can prompt you to put something else in and then take it out 24 hours later...
eh, too much trouble.  That leaves me with no choice but to "r+" - run with it

peterbe:
Now I just need to figure out who to sleep with to get these migrations executed :)
Or better, find out how I can do it myself.

PR #2971, opened 2015-09-01
Fixes bug 1194048 - Added a rank to product/versions signature summary.

AdrianGaudebert:
@peterbe r?


AdrianGaudebert:
Preview:
sigsum-prodver-rank

peterbe:
Reluctantly r+

My first reaction was; why aren't we doing this in the template instead of pretending it's part of the data. 
Then I learned that we're using Mustache which doesn't support `{{@index}}` or any helpers or anything like that. 

Secondly, I wonder, why the heck do we do this in Mustache in the first place!? If we're going to have Crash-stats be a fat-client web app with JSON + javascript then we should do it properly with something from 2014 at least. 

There's little reason not just just render this stuff as HTML (ie. django and jinja templates) and insert the HTML with jQuery.

PR #2972, opened 2015-09-01
Fixes Bug 1200628 - altered RunWatchDog sig rule to account for arg collapsing

twobraids:
two problems:
1. the RunWatchDog rule will not trigger properly if arguments are collapsed. solution: remove the arguments from the predicate
2. the RunWatchDog rule on triggering, regenerates the signature forcing it to use thread-0. The collapse arguments boolean does not extend to this regeneration.  Solution: add a "reference_value_from" to make configman signature arguements act globally instead of only locally.  Then set "resource.signature.collapse_arguments=True" in consulate.


lonnen:
R+

PR #2973, opened 2015-09-01
Bug 1196867 new crashes per user using supersearch

peterbe:
@AdrianGaudebert here's a sneak peek. It gives pretty nice results on the "By Crash Date" but we have some work to do on the "By Build Date" still since PG seems to yield such totally different numbers. 

Todo:

- [x] Hang type
- [x] respect the OS parameter choice
- [x] flake8
- [x] all tests passing
- [ ] ~~investigate why the "By Build date" numbers are so different in the PG based on (can maybe wait)~~
- [x] a link to the old daily report. 
- [x] hide the "By Build date" option

### DIFF
--- a/socorro/external/es/supersearch.py
+++ b/socorro/external/es/supersearch.py
@@ -195,7 +195,6 @@ class SuperSearch(SearchBase):
         The list of accepted parameters (with types and default values) is in
         the database and can be accessed with the super_search_fields service.
         """
-
         # Filter parameters and raise potential errors.
         params = self.get_parameters(**kwargs)
 
--- a/socorro/external/postgresql/adi.py
+++ b/socorro/external/postgresql/adi.py
@@ -23,14 +23,16 @@ class ADI(PostgreSQLBase):
             ('start_date', yesterday, 'date'),
             ('end_date', tomorrow, 'date'),
             ('product', '', 'str'),
-            ('version', '', 'str'),
+            ('versions', [], 'list'),
+            ('platforms', [], 'list'),
         ]
         params = external_common.parse_arguments(filters, kwargs)
         required = (
             'start_date',
             'end_date',
             'product',
-            'version',
+            'versions',
+            'platforms',
         )
         missing = []
         for each in required:
@@ -40,40 +42,40 @@ class ADI(PostgreSQLBase):
             raise MissingArgumentError(', '.join(missing))
 
         SQL = """
-        SELECT
-            SUM(adi_count) AS adi_count,
-            date::DATE,
-            product_name AS product,
-            product_version AS version,
-            product_os_platform AS platform,
-            update_channel AS release_channel
-        FROM
-            raw_adi
-        WHERE
-            product_name = %(product)s AND
-            product_version = %(version)s AND
-            date BETWEEN %(start_date)s::DATE AND %(end_date)s::DATE
-        GROUP BY
-            date::DATE,
-            product_name,
-            product_os_platform,
-            product_version,
-            update_channel
+            SELECT
+                SUM(adu_count) AS adi_count,
+                adu_date AS date,
+                pv.build_type,
+                pv.version_string AS version
+            FROM
+                product_adu
+            LEFT OUTER JOIN product_versions pv USING (product_version_id)
+            WHERE
+                pv.product_name = %(product)s
+                AND pv.version_string IN %(versions)s
+                AND os_name IN %(platforms)s
+                AND adu_date BETWEEN %(start_date)s AND %(end_date)s
+            GROUP BY
+                adu_date,
+                build_type,
+                version_string
         """
 
+        params['versions'] = tuple(params['versions'])
+        params['platforms'] = tuple(params['platforms'])
         results = self.query(SQL, params)
 
         fields = (
             'adi_count',
             'date',
-            'product',
+            'build_type',
             'version',
-            'platform',
-            'release_channel',
         )
         rows = []
         for record in results:
             row = dict(zip(fields, record))
             row['date'] = row['date'].strftime('%Y-%m-%d')
+            # BIGINTs become Decimal which becomes floating point in JSON
+            row['adi_count'] = long(row['adi_count'])
             rows.append(row)
         return {'hits': rows, 'total': len(rows)}
--- a/socorro/external/postgresql/product_build_types.py
+++ b/socorro/external/postgresql/product_build_types.py
@@ -1,0 +1,44 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import logging
+
+from socorro.external import MissingArgumentError
+from socorro.external.postgresql.base import PostgreSQLBase
+from socorro.lib import external_common
+
+
+logger = logging.getLogger("webapi")
+
+
+class ProductBuildTypes(PostgreSQLBase):
+
+    def get(self, **kwargs):
+        """Return a dict that holds the throttling value per build type
+        for a specific product."""
+        filters = [
+            ('product', None, 'str'),
+        ]
+        params = external_common.parse_arguments(filters, kwargs)
+        required = ('product',)
+        for key in required:
+            if not params.get(key):
+                raise MissingArgumentError(key)
+
+        sql = """
+            SELECT
+                build_type,
+                throttle
+            FROM product_build_types
+            WHERE product_name = %(product)s
+        """
+        results = self.query(sql, params)
+
+        build_types = {}
+        for row in results:
+            build_types[row[0]] = row[1]
+
+        return {
+            'hits': build_types,
+        }
--- a/socorro/unittest/external/postgresql/test_adi.py
+++ b/socorro/unittest/external/postgresql/test_adi.py
@@ -26,11 +26,50 @@ class IntegrationTestADI(PostgreSQLTestCase):
         self.date = datetime.datetime(2015, 7, 1)
         yesterday = self.date - datetime.timedelta(hours=24)
         products = ('Firefox', 'Thunderbird')
-        versions = ('39', '40')
-        platforms = ('Linux', 'Darwin')
+
+        versions = ('39.0', '40.0')
+        platforms = ('Linux', 'Windows')
         channels = ('release', 'beta')
+
+        build_date = yesterday - datetime.timedelta(days=30)
+        sunset_date = yesterday + datetime.timedelta(days=30)
+
+        for platform in platforms:
+            cursor.execute("""
+                INSERT INTO os_names
+                (os_short_name, os_name)
+                VALUES
+                (%s, %s)
+            """, (platform.lower()[:3], platform))
+            cursor.execute("""
+                INSERT INTO os_name_matches
+                (match_string, os_name)
+                VALUES
+                (%s, %s)
+            """, (platform, platform))
+
         adi_count = 1
+        product_version_id = 0
         for product in products:
+            cursor.execute("""
+                INSERT INTO products
+                (product_name, sort, release_name)
+                VALUES
+                (%s, 1, %s)
+            """, (
+                product, product.lower()
+            ))
+            cursor.execute("""
+                INSERT into product_productid_map (
+                    product_name,
+                    productid
+                ) VALUES (
+                    %s, %s
+                )
+            """, (
+                product,
+                product.lower() + '-guid',
+            ))
             for version in versions:
                 for platform in platforms:
                     for channel in channels:
@@ -48,7 +87,7 @@ class IntegrationTestADI(PostgreSQLTestCase):
                                 received_at
                             )
                             VALUES (
-                                %s, %s, %s, %s, %s, %s, %s, %s, %s, now()
+                                %s, %s, %s, %s, %s, %s, %s, %s, %s, NOW()
                             )
                         """, (
                             adi_count,
@@ -62,12 +101,36 @@ class IntegrationTestADI(PostgreSQLTestCase):
                             channel,
                         ))
                         adi_count *= 2
-        cursor.execute('select count(*) from raw_adi')
-        count, = cursor.fetchone()
-        # We expect there to be 2 channels per every 2 platforms,
-        # per every 2 versions per every 2 products.
-        assert count == 2 * 2 * 2 * 2, count
+
+                product_version_id += 1
+                cursor.execute("""
+                    INSERT INTO product_versions
+                    (product_version_id, product_name, major_version,
+                     release_version, version_string, version_sort,
+                     build_date, sunset_date, featured_version,
+                     build_type, build_type_enum)
+                    VALUES (%s, %s, %s, %s, %s, %s, %s, %s, 't', %s, %s)
+                """, (
+                    product_version_id,
+                    product,
+                    version,
+                    version,
+                    version,
+                    '0000' + version.replace('.', ''),
+                    build_date,
+                    sunset_date,
+                    'release',
+                    'release',
+                ))
+
+        cursor.callproc('update_adu', [yesterday.date()])
         self.connection.commit()
+
+        cursor = self.connection.cursor()
+        cursor.execute('select count(*) from product_adu')
+        count, = cursor.fetchone()
+        # expect there to be 2 * 2 * 2 rows of product_adu
+        assert count == 8, count
 
     def tearDown(self):
         self._truncate()
@@ -76,7 +139,10 @@ class IntegrationTestADI(PostgreSQLTestCase):
     def _truncate(self):
         cursor = self.connection.cursor()
         cursor.execute("""
-            TRUNCATE raw_adi CASCADE
+            TRUNCATE
+                raw_adi, products, product_versions, product_adu,
+                product_productid_map, os_names, os_name_matches
+            CASCADE
         """)
         self.connection.commit()
 
@@ -92,7 +158,8 @@ class IntegrationTestADI(PostgreSQLTestCase):
             start_date=start,
             end_date=end,
             product='Firefox',
-            version='42'
+            versions=['42'],
+            platforms=['Linux', 'Windows'],
         )
         eq_(stats['hits'], [])
         eq_(stats['total'], 0)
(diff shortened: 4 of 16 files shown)


peterbe:
Yes, I know there's tons of pep8 errors that prevents this from actually running on travis. Working on that.

socorro/external/postgresql/adi.py
```diff
@@ -23,14 +23,16 @@ def get(self, **kwargs):
             ('start_date', yesterday, 'date'),
             ('end_date', tomorrow, 'date'),
             ('product', '', 'str'),
-            ('version', '', 'str'),
+            ('versions', [], 'list'),
+            ('platforms', [], 'list'),
```
AdrianGaudebert: I do not really like the ``s`` at the end of parameters anymore. But maybe it makes more sense here. In the SuperSearch world everything can be a list so it's a bother to make everything plural. 

I guess the actual question is: do we want to be consistent with the SuperSearch API? In which case we should not add ``s`` to parameters.

AdrianGaudebert:
Wouldn't this new crashes-per-user page be better off in a separate django app? I really hate our ``crashstats/crashstats/views.py``, it's way too long and difficult to read, and adding more things into it pains me.

peterbe:
Adrian and I discussed the "By Build date" option. If we can't figure out why the numbers are so different in PG, we'll move ahead with that option commented out.

peterbe:
@AdrianGaudebert All the little checkboxes above are all checked. :)

socorro/external/es/supersearch.py
```diff
@@ -195,7 +195,11 @@ def get(self, **kwargs):
         The list of accepted parameters (with types and default values) is in
         the database and can be accessed with the super_search_fields service.
         """
-
+        with open('/tmp/supersearch.log', 'a') as f:
+            print >>f, "KWARGS"
+            from pprint import pprint
+            pprint(kwargs, f)
+            print >>f, ""
```
AdrianGaudebert: debug leftover

socorro/external/postgresql/adi.py
```diff
@@ -40,40 +42,41 @@ def get(self, **kwargs):
...
+            WHERE
+                pv.product_name = %(product)s
+                AND pv.version_string IN %(versions)s
+                AND os_name IN %(platforms)s
+                AND adu_date BETWEEN %(start_date)s AND %(end_date)s
+            GROUP BY
+                adu_date,
+                build_type,
+                version_string
+            --ORDER BY adu_date
```
AdrianGaudebert: debug leftover, I guess?

socorro/external/postgresql/product_build_types.py
```diff
@@ -0,0 +1,44 @@
...
+            WHERE product_name = %(product)s
+        """
+        results = self.query(sql, params)
+
+        build_types = {}
+        for row in results:
+            build_types[row[0]] = row[1]
+
+        return {
+            'build_types': build_types,
```
AdrianGaudebert: I would call that key ``hits`` for consistency.

socorro/unittest/external/postgresql/test_adi.py
```diff
@@ -62,13 +101,37 @@ def setUp(self):
...
+                    version,
+                    version,
+                    '0000' + version.replace('.', ''),
+                    build_date,
+                    sunset_date,
+                    'release',
+                    'release',
+                ))
+
+        cursor.callproc('update_adu', [yesterday.date()])
```
AdrianGaudebert: Nice, we're starting to test stored procedures in our integration tests, that's great!

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/crashes_per_user.js
```diff
@@ -0,0 +1,78 @@
...
+        } else {
+            $('#crashes-per-adi-graph')
+                .text('No results were found.');
+        }
+    }
+
+    $('#daily_search_version_form_products').change(function() {
+        var urlForm = $('#daily_search_version_form').attr('action'),
+            product = $(this).find(':selected').val(),
+            url = urlForm + '?p=' + product + window.location.hash;
```
AdrianGaudebert: Ugh, I really don't like this style of writing ``var`` only once. I like when it's at the beginning of each line, it's much much clearer imo. (I suspect this is not your doing and you just copied this file, but now is probably a good time to clean this up. :) )

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/crashes_per_user.js
```diff
@@ -0,0 +1,78 @@
...
+        window.location = url;
+    });
+
+    $('#daily_search_os_form_products').change(function() {
+        var urlForm = $('#daily_search_os_form').attr('action'),
+            product = $(this).find(':selected').val(),
+            url = urlForm + '?p=' + product + window.location.hash;
```
AdrianGaudebert: Same thing here.

webapp-django/crashstats/crashstats/templates/crashstats/crashes_per_user.html
```diff
@@ -0,0 +1,275 @@
+{% extends "crashstats_base.html" %}
+{% block site_css %}
+  {{ super() }}
```
AdrianGaudebert: indentation is weird in this file, it's mostly 4-spaces but sometimes it's 2-spaces.

webapp-django/crashstats/crashstats/templates/crashstats/crashes_per_user.html
```diff
@@ -0,0 +1,275 @@
...
+
+{% block page_title %}
+Crashes per User for {{ product }}
+{% endblock %}
+
+{% block content %}
+<div id="mainbody">
+
+<div class="page-heading">
+    <p style="float: right">
```
AdrianGaudebert: I'm pretty sure that would feel a lot better in the ``less`` file. :) 

(Unless you are concerned about us forgetting to remove the associated rule when we remove this link. Which sounds valid to me... This shouldn't stick around for too long so I suppose it's fine. )

webapp-django/crashstats/crashstats/templates/crashstats/crashes_per_user.html
```diff
@@ -0,0 +1,275 @@
...
+</div>
+
+
+<div class="panel daily_search">
+    <div class="title daily_search_title">
+        <h2>Options</h2>
+    </div>
+
+    <div class="body daily_search_body">
+        <div id="daily_search">{# is this div tag necessary? #}
```
AdrianGaudebert: Did you figure it out?

webapp-django/crashstats/crashstats/templates/crashstats/crashes_per_user.html
```diff
@@ -0,0 +1,275 @@
...
+
+                <table class="by_version">
+                    <tr>
+                        <th>Product</th>
+                        <td>
+                          <label class="visually-hidden" for="daily_search_version_form_products">Select Product</label>
+                        <select id="daily_search_version_form_products" name="p">
+                          {% for product_name in releases %}
+                            <option value="{{ product_name }}" {% if product == product_name %}selected{% endif %}>{{ product_name }}</option>
+                            {% endfor %}
```
AdrianGaudebert: This is a good example of the weird indentation of this file. :)

webapp-django/crashstats/crashstats/templates/crashstats/crashes_per_user.html
```diff
@@ -0,0 +1,275 @@
...
+                    {% for i in range(4)|reverse %}
+                    <tr>
+                        {% if loop.first %}
+                        <th rowspan="4">Versions</th>
+                        {% endif %}
+
+                        <td>
+                          <label class="visually-hidden" for="version{{ i }}">Select Version</label>
+                            <select id="version{{ i }}" name="v">
+                                <option value="">-- versions --</option>
```
AdrianGaudebert: Should this value be disabled? Or could it be replaced with a placeholder, even? (Maybe that's a bit out of scope... )

webapp-django/crashstats/crashstats/templates/crashstats/crashes_per_user.html
```diff
@@ -0,0 +1,275 @@
...
+        <div class="title">
+            <h2>Crashes per ADI</h2>
+            <div class="choices">
+                <ul>
+                    {% if request.META.get('QUERY_STRING') %}
+                    <li><a href="?{{ request.META.get('QUERY_STRING') }}&amp;format=csv">csv</a></li>
+                    {% else %}
+                    <li><a href="?format=csv">csv</a></li>
+                    {% endif %}
+                </ul>
```
AdrianGaudebert: Wow, does that still work? Are people using that option? Do we want to keep it around? 

I was under the impression that we were trying to remove CSV's as much as possible to push people towards the API. Maybe now is a good time to do that here?

webapp-django/crashstats/crashstats/templates/crashstats/crashes_per_user.html
```diff
@@ -0,0 +1,275 @@
...
+  {% endcompress %}
+
+  {% compress js %}
+  <script src="{{ static('crashstats/js/lib/d3.min.js') }}"></script>
+  <script src="{{ static('crashstats/js/lib/metricsgraphics.min.js') }}"></script>
+  <script src="{{ static('crashstats/js/socorro/crashes_per_user.js') }}"></script>
+  {% endcompress %}
+  <script>
+    var DATA = {{ graph_data | json_dumps }};
+    window.socGraphByReportType = false;
```
AdrianGaudebert: If this is always false, I'm quite sure it can be removed and the JS file can be simplified.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -817,24 +893,293 @@ def append_row_blob(blob, labels):
...
+
+@pass_default_context
+def crashes_per_user(request, default_context=None):
+    context = default_context or {}
+
+    context['products'] = context['currentproducts']['products']
+
+    platforms_api = models.Platforms()
+    platforms = platforms_api.get()
+    form_class = forms.DailyFormByVersion
```
AdrianGaudebert: What is the point of this variable? It seems to be used only once afterwards.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -817,24 +893,293 @@ def append_row_blob(blob, labels):
...
         try:
-            blob = data['totals'][product_version]
+            params['platforms'] = params.pop('os')
         except KeyError:
-            continue
-        append_row_blob(blob, totals_labels)
-    writer.writerow(row)
-    return response
+            params['platforms'] = None
+    else:
```
AdrianGaudebert: Why not reverse this ``if`` and gain one level of indentation, since you have a ``return`` here?

webapp-django/crashstats/crashstats/views.py
```diff
@@ -817,24 +893,293 @@ def append_row_blob(blob, labels):
...
+    if isinstance(start_date, datetime.datetime):
+        start_date = start_date.date()
+
+    context['start_date'] = start_date.strftime('%Y-%m-%d')
+    context['end_date'] = end_date.strftime('%Y-%m-%d')
+
+    context['duration'] = abs((start_date - end_date).days)
+    context['dates'] = utils.daterange(start_date, end_date)
+
+    context['hang_type'] = params.get('hang_type') or 'any'
```
AdrianGaudebert: Is that not equivalent to ``params.get('hang_type', 'any')``?

webapp-django/crashstats/crashstats/views.py
```diff
@@ -817,24 +893,293 @@ def append_row_blob(blob, labels):
...
+
+    default = request.session.get('date_range_type', 'report')
+    context['date_range_type'] = params.get('date_range_type') or default
+
+    # if params.get('hang_type') == 'any':
+    #     hang_type = None
+    # else:
+    #     hang_type = params.get('hang_type')
```
AdrianGaudebert: debug leftover?

webapp-django/crashstats/crashstats/views.py
```diff
@@ -817,24 +893,293 @@ def append_row_blob(blob, labels):
...
+
+    api = SuperSearchUnredacted()
+    # XXX HACK! Because of a bug in SuperSearch we can't search for
+    # '41.0b' so we have to change that to '41.0b1' :(
+    # See https://bugzilla.mozilla.org/show_bug.cgi?id=1197251
+    if '41.0b' in params['versions']:
```
AdrianGaudebert: So that's actually a problem with all versions ending in ``b``, not 41 specifically. You'll want to make this slightly more generic.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -817,24 +893,293 @@ def append_row_blob(blob, labels):
...
+        params['versions'].insert(
```
AdrianGaudebert: The position of that parameter does not matter, you can simply ``append``.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -817,24 +893,293 @@ def append_row_blob(blob, labels):
...
+    else:
+        params['date'] = [
+            '>=' + start_date.strftime('%Y-%m-%d'),
+            '<' + end_date.strftime('%Y-%m-%d'),
+        ]
+
+    # XXX SuperSearch always returns `hits` even if you're only interested
+    # in facets.
+    # Setting this to 0 means it returns the default number of hits.
+    # Adrian, do we have a bug for this?
```
AdrianGaudebert: I just filed one: https://bugzilla.mozilla.org/show_bug.cgi?id=1207085

webapp-django/crashstats/crashstats/views.py
```diff
@@ -817,24 +893,293 @@ def append_row_blob(blob, labels):
...
+        version = group['version']
+        build_type = group['build_type']
+        throttle = product_build_types[build_type]
+        if version not in adi_by_version:
+            adi_by_version[version] = {}
+        date = group['date']
+        count = group['adi_count']
+        adi_by_version[version][date] = [count, throttle]
+
+    cadu = {}
```
AdrianGaudebert: I don't understand this variable name, I thus believe it is not a very good one. :->

socorro/external/postgresql/adi.py
peterbe: I think we discussed this elsewhere. The fact that SuperSearch allows a value to be a string or a list is an inherited thing from out of ES. If you aks me, it's kinda strange to not call it "versions", "products", "platforms" everywhere. We often have code that looks something like this:
```python
def find_things(platforms):
    if isinstance(platforms, basestring):
        # allows people to be lazy and type `find_things('osx')`
         platforms = [platforms]
```

I think the `s` adds value to the code-reader and implementor. You can immediately see that the implementation allows for it to be a multi-value. Like `FirstDateSignatures.get(signatures=[....])`. That tells, just from reading that that you can use a list. 

If we don't have the `s`, people might think that you can send either a value or a list of values and it might break horribly if it doesn't work. 

In the world of PG, which we're never going to get rid of, I think it makes sense to have humane sounding arguments.

socorro/external/postgresql/product_build_types.py
peterbe: agreed.

webapp-django/crashstats/crashstats/templates/crashstats/crashes_per_user.html
peterbe: Sigh. We need to figure this out. 
I think I have my editor to default to 2-spaces and that's why I sometimes ruin an existing file. 
But want to know what's really sad? [THIS!](https://gist.github.com/peterbe/16f8692b31eeb88d34a1)
You can read more about it in this blog post: http://www.peterbe.com/plog/find-what-indentation-your-files-use

I think you and I need to sit down and make a decision of what we prefer. 
Then we set aside a bit of time to make some quick PRs that correct existing files and when we've done that we add a `.editorconfig` file to the root of the project. 

[In airmozilla](https://github.com/mozilla/airmozilla/blob/master/.editorconfig#L13) I use 2-spaces but it's not for any particular strong reason. Just anything being decided is nice. 

What's certainly clear is that my additions to the file shouldn't mix. If the file **was** 4-spaces, my additions should be 4-spaces too. /me corrects.

webapp-django/crashstats/crashstats/templates/crashstats/crashes_per_user.html
peterbe: I've moved it to the `.less` file now. 
It's a strange issue actually. I've studied the subject some and listened to a long podcast about it. 
There's a school of thought that the only thing you do with global CSS is things like `reset.css` and some odd exceptions. Everything else you do inline styles. Actually, in an era of tools like React you don't do `<div id="body" style="font-family:...">` because you don't first create the HTML at all. Instead you just write the styles directly into th DOM elements. And instead of setting `backgroundColor="red"` you use `backgroundColor=generalBackgroundColorTheme1` so it' variables and not values. 
The advantage is that styling becomes tied to the code that actually needs it. 
However, I'm not sold on it. It has advantages and disadvantages.

webapp-django/crashstats/crashstats/templates/crashstats/crashes_per_user.html
peterbe: Removed now. It made sense when you could chose "Crashes per OS"

webapp-django/crashstats/crashstats/templates/crashstats/crashes_per_user.html
peterbe: Still works. And people are using it. I don't know how much. 

It wouldn't be easy to replace that with instructions how how to use the API because the output involves doing gets on the ADI, the ProductBuildTypes AND the SuperSearch.

webapp-django/crashstats/crashstats/views.py
peterbe: No, that's different. I want it to be `'any'` if the `hang_type` is not set at all or is falsy.

webapp-django/crashstats/crashstats/views.py
peterbe: Order matters for the 4 version select boxes. I want it to be be 43, 42, 41.0b1, 40. Not 43, 42, 40, 41.0b1

webapp-django/crashstats/crashstats/views.py
peterbe: Corrected. Thanks.

webapp-django/crashstats/crashstats/views.py
peterbe: Changed to `graph_data` now.

AdrianGaudebert:
Tested locally, r+. Let's merge and test this on stage / prod with real data.

PR #2974, opened 2015-09-02
fixes bug 1201270 - Migration to backfill last 28 days of signature_summary_os

peterbe:
@twobraids r?

### DIFF
--- a/alembic/versions/16b2bee7db72_backfill_signature_summary_os_28_days.py
+++ b/alembic/versions/16b2bee7db72_backfill_signature_summary_os_28_days.py
@@ -1,0 +1,47 @@
+"""backfill signature_summary_os 28 days
+
+Revision ID: 16b2bee7db72
+Revises: 4afd4e13281d
+Create Date: 2015-09-02 13:53:13.505082
+
+"""
+import datetime
+
+# revision identifiers, used by Alembic.
+revision = '16b2bee7db72'
+down_revision = '4afd4e13281d'
+
+from alembic import op
+
+
+def upgrade():
+    # We only ever, at most, show 28 days in the report list that
+    # shows this signature summary.
+    today = datetime.datetime.utcnow().date()
+    for i in range(28, 0, -1):
+        then = today - datetime.timedelta(days=i)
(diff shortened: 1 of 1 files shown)


twobraids:
looks like you may have to run both stored procs?  one to ensure the partition exists, the other to fill it with data?

peterbe:
Ok. Pushed another attempt to fix that.

peterbe:
Seems happier now.

alembic/versions/16b2bee7db72_backfill_signature_summary_os_28_days.py
```diff
@@ -0,0 +1,47 @@
...
+    # We only ever, at most, show 28 days in the report list that
+    # shows this signature summary.
+    today = datetime.datetime.utcnow().date()
+    for i in range(28, 0, -1):
+        then = today - datetime.timedelta(days=i)
+        till = then + datetime.timedelta(days=1)
+        op.execute(
+            """
+            SELECT
+                backfill_weekly_report_partitions(
```
twobraids: No need to act: but we end up running the ```backfill_weekly...``` stored procedure 7 times more than necessary.  However, this is only run once and that stored procedure can cope with the unnecessary runs without raising an exception.

twobraids:
r+ if you're satisfied with the data that gets inserted.

peterbe:
What did you mean by the "7 times more"?

alembic/versions/16b2bee7db72_backfill_signature_summary_os_28_days.py
twobraids: the partitioning stored procedure, when given a date, will figure find the date of the previous Monday and then create a partition for it.  You run that procedure for each day of the week - effectively telling the procedure to make the partition seven times.  It does make the partition on the first call.  On each of the subsequent six calls, it figures out the date of the Monday and then sees that there already is a partition for that week.  It silently quits at that point, declining to make a partition that already exists.

peterbe:
A ha! So it's benign. Saves me from having to figure out which of all those is a Monday etc.

PR #2975, opened 2015-09-03
The debian postrm cleanup should happen on purge

pcabido:
The socorro files (configs specifically) and user should only be removed on
purge.
https://wiki.debian.org/AccountHandlingInMaintainerScripts has some
good guidelines regarding account deletion.


peterbe:
@jdotpz can you review this?

jdotpz:
r+

PR #2976, opened 2015-09-03
Revert "Fixes bug 1194048 - Added a rank to product/versions signatur…

AdrianGaudebert:
…e summary."

This reverts commit 776df0d877ecb0137d4df831878bed3f690d1898.

cc @peterbe, @KaiRo-at has vetoed that we do not do this, so backing it out.


peterbe:
Have I?
What I objected to is fact that we're doing all of that stuff in Mustache and not on the server-side.

AdrianGaudebert:
No no, KaiRo did, not you. :)

PR #2977, opened 2015-09-03
alembic to be able to read from env

peterbe:
@twobraids r?
CC @selenamarie @rhelmer 

So, to run the migrations (or just see what the current head is) I have to edit `/etc/socorro/alembic.ini` and change the line `sqlalchemy.url = ...` every time the admin node is restarted. That cumbersome. 
I.e. the following:
```
vi /etc/socorro/alembic.ini
cd /data/socorro/application/
. /data/socorro/socorro-virtualenv/bin/activate
alembic -c /etc/socorro/alembic.ini current
```

The problem is that the admin node is ephemeral and my changes to `/etc/socorro/alembic.ini` are wiped. 

As of this patch, I can override `sqlalchemy.url` by running it like this:

```
env sqlalchemy.url=postgresql://USER:PASSW@HOSTNAME/breakpad alembic -c config/alembic.ini current
```
I get the values from:
```
envconsul -prefix socorro/common -sanitize -upcase env | grep POSTGRESQL
```
Which would make it easy to put together that on-the-fly environment variable.

### DIFF
--- a/alembic/env.py
+++ b/alembic/env.py
@@ -1,4 +1,5 @@
 from __future__ import with_statement
+import os
 from alembic import context
 from sqlalchemy import engine_from_config, pool
 from logging.config import fileConfig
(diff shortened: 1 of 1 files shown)


selenamarie:
+1 cool!

peterbe:
Ping @twobraids

lonnen:
r+

PR #2978, opened 2015-09-03
Reencode to work without setting calling format

peterbe:
@rhelmer r?

So this was thorny to say the least. 
My local python can't connect to S3 buckets with setting calling_format. And centos can't connect to S3 buckets if you do. 

See extended comment. 

Secondly, you can't, on my Python 2.7 OSX do something like `requests.get('https://s3.amazonaws.com/peterbe.org.allizom.crash-stats.staging.crashes/v1/symbolPush/C356CD0633314E568FDCFF475732FE741/NPSWF64.sym')` 
So I changed how the URL is made.

### DIFF
--- a/webapp-django/crashstats/symbols/management/commands/reencode-bad-sym-files.py
+++ b/webapp-django/crashstats/symbols/management/commands/reencode-bad-sym-files.py
@@ -1,10 +1,11 @@
 import gzip
+import ssl
 from cStringIO import StringIO
 
 from optparse import make_option
 
 import requests
-import boto.s3.connection
+import boto
 
 from django.conf import settings
 from django.core.management.base import BaseCommand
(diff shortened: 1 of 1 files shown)


webapp-django/crashstats/symbols/management/commands/reencode-bad-sym-files.py
```diff
@@ -104,7 +126,17 @@ def handle(self, *args, **options):
...
+                    # which will produce a URL like
+                    # https://my.bucket.name.s3.amazonaws.com/key/name.ext
+                    # which can be troublesome for Python 2.7 on OSX to
+                    # GET because of the dots, we instead make our own
+                    # URL.
+                    url = 'https://s3.amazonaws.com/{}/{}'.format(
+                        key.bucket.name,
+                        key.name
+                    )
                     print url
```
rhelmer: I know this isn't part of the patch but is this `print` on purpose?

rhelmer:
r+ good job on the comments, after reading the links in there I agree. Unfortunate we need to work around like this to work on both Mac and Linux, but so it goes.

webapp-django/crashstats/symbols/management/commands/reencode-bad-sym-files.py
peterbe: yep that was on purpose. On line 128 it prints the size and then after 10 spaces it prints the url.

PR #2979, opened 2015-09-08
fixes bug 1202739 - Monitor crash-analysis.m.c for valid files

peterbe:
@AdrianGaudebert r?

You can test this simply by going to `http://localhost:8000/monitoring/crash-analysis-health/`


webapp-django/crashstats/crashstats/templates/crashstats/report_index.html
```diff
@@ -517,7 +517,9 @@
           {% if total_correlations > 0 %}
           <div id="correlation" class="ui-tabs-hide"><h2>Correlations for {{ product }} {{ version }} {{ report.os_name }} </h2>
             <p>Below are tables of correlation data generated by <a href="http://dbaron.org/mozilla/topcrash-modules">dbaron's scripts</a>. If this crash signature is
-      a top crasher, then we should be able to load it from <a href="http://crash-analysis.mozilla.com/crash_analysis/">the original text files</a>. Only the first 2MB of each Platform's output is loaded. If this screen has data from multiple product/version/platforms, the crashiest was picked ({{ product }} {{ version }} {{ report.os_name }}).</p>
+              a top crasher, then we should be able to load it from
+              <a href="{{ CRASH_ANALYSIS_URL }}">the original text files</a>.
+              Only the first 2MB of each Platform's output is loaded. If this screen has data from multiple product/version/platforms, the crashiest was picked ({{ product }} {{ version }} {{ report.os_name }}).</p>
```
peterbe: Not only was it hardcoded before, it was wrong too (missing https).

webapp-django/crashstats/monitoring/views.py
```diff
@@ -0,0 +1,81 @@
...
+                    url,
+                )
+            )
+            continue  # no point checking its content
+
+        # The output is expected to be something like this:
+        # <a href="file.txt">file.txt</a> DD-Mon-YYYY HH:MM    X\r\n
+        # where 'X' is an integer that represents the file's size
+        good_lines = [
+            x for x in response.text.splitlines()
```
peterbe: `response.text` is better than `response.content` on a `requests.get()` returned object. The `content` is the raw stuff. The `text` is the content with the encoding applied.

peterbe:
@lonnen It's not urgent. Perhaps keep @AdrianGaudebert as the r? He also has his webapp up and running so he can easily hit that URL. 

@rhelmer No need from a review from you, but let me shine some light on my business logic in this code. So, it looks at "today", "today - 1 day", "today - 2 days" and then looks for the relevant sub-directories on [crash-analysis](https://crash-analysis.mozilla.com/crash_analysis/) if today's directory doesn't exist it doesn't raise that as an error. It just yields a warning. That's because this montoring might execute a couple of hours before the crash-analysis cron jobs have had a chance to do their thing of the day even if it runs only a couple of hours after midnight.

webapp-django/crashstats/monitoring/views.py
lonnen: multi-line list comprehensions always throw me in python. consider revising

webapp-django/settings_test.py
```diff
@@ -60,3 +60,9 @@
         'elasticsearch_urls': ['http://example:9123'],
     },
 }
+
+
+# Make sure we never actually hit a real URL when testing the
+# Crash-analysis monitoring.
+CRASH_ANALYSIS_URL = 'https://crashanalysis.m.c/something/'
+CRASH_ANALYSIS_HEALTH_DAYS = 2
```
lonnen: HEALTH_DAYS is a strange name. Consider a comment explaining this somewhere. I had to read the code to infer what it was doing.

webapp-django/crashstats/monitoring/views.py
peterbe: They're actually very common. It's more efficient that having to set up a for loop a new variable to append to. 
It's a common pattern throughout the socorro code.

webapp-django/settings_test.py
peterbe: Fair point. I'll think up something better.

lonnen:
r+

PR #2980, opened 2015-09-09
bug 1193536 - only reencode S3 uploads of matching bucket name

peterbe:
@rhelmer r?


peterbe:
This'll avoid trying to fix prod S3 files when running this on stage.

peterbe:
@AdrianGaudebert or @twobraids I know you might not know what's going on here but @rhelmer might not have time to look at this. And remember, it's a one-off script that we'll hopefully use once and burn after. 

Fixing this script blocks us from being able to clean up bad symbols uploads and that affects the pending code changes @luser has in mind in the near future.

rhelmer:
r+

PR #2981, opened 2015-09-10
fixes bug 1203593 - signature summaries html in django instead of mustache

peterbe:
@AdrianGaudebert r?

This might seem like a big and unnecessary change. It's not really big because I just moved things and did a search-and-replace. Instead of making the view return JSON it returns a regular rendered HTML template. I did clean up `report_list_signature_summary.js` a bit with some jquery love. 

It only took me about 1-2 in total to do this but perhaps it was silly because hopefully in a couple of months we'll have the whole of report list replaced. But till then, we can at least have something much faster and simplified.


webapp-django/crashstats/crashstats/templates/crashstats/signature_summary_tables.html
```diff
@@ -0,0 +1,256 @@
...
+    <table class="sig-dashboard-tbl data-table initially-hidden">
+        <caption>
+            <span class="sigtbl-title">Mobile Devices:</span>
+            Mobile devices represented in the signature's crash statistics.
+        </caption>
+        <thead>
+            <tr>
+                <th scope="col">Manufacturer</th>
+                <th scope="col">Model</th>
+                <th scope="col"><a href="http://en.wikipedia.org/wiki/Android_%28operating_system%29#Platform_usage">API Level</a></th>
```
AdrianGaudebert: Just a nit, but should we link to the https version instead?

AdrianGaudebert:
Looking good to me, but travis is red. :cry:

PR #2982, opened 2015-09-11
fixes bug 1204108 - Don't truncate signature URLs

peterbe:
@AdrianGaudebert r?


AdrianGaudebert:
This is looking good. The job failed but it seems to be unrelated, so I restarted it. My only question is: what about KaiRo's concern in the bug?

PR #2983, opened 2015-09-11
fixes bug 1186474 - deprecate crash trends report

peterbe:
@AdrianGaudebert r?

I'm going to investigate if there are matviews or stored procedures that can be deprecated too.


peterbe:
@twobraids @rhelmer @selenamarie 
Anybody know any deep reason why the nightly_builds table can't be remove now that the "Crash Trends" report is going away?

I've never actually understood or used the damn report so I'm not super clued up on its context. 

cc @lonnen 

By the way, in the last month 0.10% of our GA traffic goes to Crash Trends. Within that 0.10%, about 2% have bothered to look at any other product than Firefox.

AdrianGaudebert:
Red is my favorite color. :)

AdrianGaudebert:
This looks pretty straightforward. I don't know why that test is failing, nor do I know anything about the tables involved, so I'll you fight this. I also guess we can indeed deprecate some more stored procs, but once again I know too little about our database to be sure.

peterbe:
@selenamarie as of [this commit](https://github.com/peterbe/socorro/commit/20ecaffe7a6b1b45b747b170f6cda1d5385abb17) this PR now deletes the model and its alembic migration will pass tests since the downgrade sort of works. Is this correct?

peterbe:
So, @AdrianGaudebert are you ok with me landing this? I basically means that the Crash Trends report disappears and that the underlying data is wiped too. 

@twobraids @rhelmer This feels like quite a big change, can you think of ANY reason NOT to delete the nightly_builds table?

AdrianGaudebert:
If we delete the data, there is no coming back. Are we sure it's OK to do that? @lonnen?

AdrianGaudebert:
@peterbe let's bring this up at the meeting today.

peterbe:
Consensus was that it's ok for it to be deleted. 

So, can I merge this now?

AdrianGaudebert:
Yes!

PR #2985, opened 2015-09-11
fixes bug 1192014 - Purge pyelasticsearch

peterbe:
@AdrianGaudebert r?

It's hard to see if there are "left-over dependencies" that we can live without. 
Imagine a script you run over night that tests all sorts of combinations of removing items from `requirements.txt` and sees if the whole test suite passes.


AdrianGaudebert:
The problem with that approach, @peterbe, is that we have some scripts that are not called during testing. But maybe it has become marginal now, since we have deprecated a whole lot of untested files. 

Anyway, I can't find no ``pyelasticsearch`` in the repo, so r+.

PR #2986, opened 2015-09-11
fixes bug 1190443 - Error message on SuperSearch AJAX search inserts whole HTML

peterbe:
@AdrianGaudebert r?

This is now what it looks like when you get a 400 AJAX query:
screenshot 2015-09-11 15 50 00

And this is what it looks like when you get a 500 error:
screenshot 2015-09-11 15 50 44


AdrianGaudebert:
That is great! Thanks Peter!

PR #2987, opened 2015-09-11
bug 1198531 - Get rid of LagLog

peterbe:
@twobraids or @AdrianGaudebert r?

Note! This does NOT remove the `lag_log` table. 
The reason is that, to do that, I have to write an alembic migration and I don't want to do that right now. Reason being that https://github.com/mozilla/socorro/pull/2983 contains an alembic migration off the current revision sha. 
So if I make an alembic migration today (prior to #2983 landing), we'll have an annoying alembic conflict. 
Note that this commit's message doesn't close the bug.


AdrianGaudebert:
Looks good to me! Will we have to make some changes to our configuration on stage and prod? 

r+

peterbe:
> Looks good to me! Will we have to make some changes to our configuration on stage and prod? 

Not really. The admin node needs to be upgraded so that the list of crontabber jobs is refreshed.

PR #2988, opened 2015-09-14
Fixes Bug 1204563 - make the submitter more friendly to other crashstores

twobraids:
the variants of the crash storage classes that are defined along with the submitter, to not follow the crash storage API correctly. For example, the file system crash source for the submitter doesn't iterate over crash_ids, but crash paths instead.  

pave over these differences in the submitter so that the submitter is capable of using the standard crash storage API.  +tests

### DIFF
--- a/socorro/collector/submitter_app.py
+++ b/socorro/collector/submitter_app.py
@@ -6,8 +6,12 @@
 
 
 import time
-import os.path
 import json
+
+from os import (
+    path,
+    listdir
+)
 
 from configman import Namespace, RequiredConfig
 from configman.converters import class_converter
@@ -17,7 +21,6 @@ from socorro.external.crashstorage_base import CrashStorageBase
 from socorro.external.filesystem.filesystem import findFileGenerator
 from socorro.lib.util import DotDict
 from socorro.external.postgresql.dbapi2_util import execute_query_iter
-
 
 
 #==============================================================================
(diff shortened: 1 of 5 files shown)


socorro/collector/submitter_app.py
```diff
@@ -66,6 +74,8 @@ def get_raw_dumps_as_files(self, dump_pathnames):
         parameters:
         dump_pathnames - a tuple of paths. the second element and beyond are
                          the dump pathnames"""
+        dumps = {}
```
lonnen: This is unused as far as I can tell

socorro/collector/submitter_app.py
```diff
@@ -237,12 +258,36 @@ def __init__(self, config):
...
+            # unpacks the tuple so that they can be used separately in the
+            # right context
+            crash_id, paths = crash_id
         if self.config.submitter.dry_run:
             print crash_id
         else:
-            raw_crash = self.source.get_raw_crash(crash_id)
-            dumps = self.source.get_raw_dumps_as_files(crash_id)
-            self.destination.save_raw_crash(raw_crash, dumps, crash_id)
+            # in the case where the path has a non-None value, that means
```
lonnen: Well commented.

socorro/external/crashstorage_base.py
```diff
@@ -140,6 +140,25 @@ def save_raw_crash(self, raw_crash, dumps, crash_id):
         pass
 
     #--------------------------------------------------------------------------
+    def save_raw_crash_with_file_dumps(self, raw_crash, dumps, crash_id):
+        """this method that saves  both the raw_crash and the dump, must be
+        overridden in any implementation.  Unlike the previous method, the
+        dumps are in the form
```
twobraids: this is rather incomplete

twobraids:
@lonnen is this a thumbs up or down?  This will become a blocker if it doesn't get merged soon.

PR #2990, opened 2015-09-15
Fixes bug 1195306 - Added a new service to get the first appearance d…

AdrianGaudebert:
…ate of signatures.

@peterbe r?

### DIFF
--- a/socorro/external/postgresql/signature_first_date.py
+++ b/socorro/external/postgresql/signature_first_date.py
@@ -1,0 +1,40 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from socorro.external import MissingArgumentError
+from socorro.external.postgresql.base import PostgreSQLBase
+from socorro.lib import datetimeutil, external_common
+
+
+class SignatureFirstDate(PostgreSQLBase):
+    filters = [
+        ('signatures', None, ['list', 'str']),
+    ]
+
+    def post(self, **kwargs):
+        params = external_common.parse_arguments(self.filters, kwargs)
+
+        if not params['signatures']:
+            raise MissingArgumentError('signatures')
+
+        sql_params = [tuple(params['signatures'])]
+        sql = """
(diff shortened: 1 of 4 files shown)


socorro/middleware/middleware_app.py
```diff
@@ -65,6 +65,7 @@
     (r'/releases/(channels|featured|release)/(.*)', 'releases.Releases'),
     (r'/report/(list)/(.*)', 'report.Report'),
     (r'/server_status/(.*)', 'server_status.ServerStatus'),
+    (r'/signature/first_date', 'signature_first_date.SignatureFirstDate'),
```
peterbe: I wonder why all the other ones have `(.*)` at the end? Sigh! Can't wait to rewrite this.

socorro/external/postgresql/signature_first_date.py
```diff
@@ -0,0 +1,40 @@
...
+        ('signatures', None, ['list', 'str']),
+    ]
+
+    def post(self, **kwargs):
+        params = external_common.parse_arguments(self.filters, kwargs)
+
+        if not params['signatures']:
+            raise MissingArgumentError('signatures')
+
+        sql_params = [tuple(params['signatures'])]
```
peterbe: I'm stumped. How does this work? If I call it with `SignatureFirstDate(config).post(signatures='onestring')` where do you make that into a tuple?

peterbe: The tests don't lie. I guess the `parse_arguments()` function makes sure to turn strings into lists.

peterbe:
r+

socorro/external/postgresql/signature_first_date.py
AdrianGaudebert: That's exactly it. Since the filters above declare that ``signatures`` must be a list of strings, ``parse_arguments`` makes sure it is one.

socorro/middleware/middleware_app.py
AdrianGaudebert: Legacy! There was a time when our URLs looked like this: ``/search/product/Firefox/version/20.2a/...``. Though I realize now that I forgot a trailing slash that every other URL has. I'll add it before merging.

PR #2991, opened 2015-09-15
fixes bug 1204949 - Rearrange monitoring URLs

peterbe:
@AdrianGaudebert r?


peterbe:
With this in place, perhaps we can set up a monitoring script, that Pingdom can ping, that does a SuperSearch query. It runs something like every 5 minutes so it should be really minimal or perhaps it's ok that it tests the caching infrastructure. What do you think? CC @jdotpz

AdrianGaudebert:
Yes yes yes! Less code in the ``crashstats`` app, I like that! 

r+

What's the purpose of this monitoring app in your opinion Peter? Is it just for other machines to check the status of socorro, or could it also be used by humans? The underlying question is: should we also move ``crontabber-state`` and ``server-status`` into this app?

peterbe:
> What's the purpose of this monitoring app in your opinion Peter? 

The purpose is for tools like Pingdom. Or possibly some other Nagios-like thing when we move to that in the future. 

The `crontabber-state` and `server-status` are going to merge into one **human** friendly page and not be considered part of this stuff.

PR #2992, opened 2015-09-15
fixes bug 1204960 - crash-analysis health warnings

peterbe:
@rhelmer r?


rhelmer:
r+, this could be a sign of a problem with the script but I think it'd be better to fix the script to not copy 0-byte files to the output dir.

PR #2993, opened 2015-09-16
Fixes bug 1195310 - Showed startup crash icon in new topcrashers.

AdrianGaudebert:
@peterbe r?

### DIFF
--- a/webapp-django/crashstats/topcrashers/views.py
+++ b/webapp-django/crashstats/topcrashers/views.py
@@ -76,7 +76,9 @@ def get_topcrashers_results(**kwargs):
         'is_garbage_collecting',
         'hang_type',
         'process_type',
+        '_histogram.uptime',
     ]
+    params['_histogram_interval.uptime'] = 60
 
     # We don't care about no results, only facets.
     params['_results_number'] = 0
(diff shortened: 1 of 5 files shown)


webapp-django/crashstats/topcrashers/views.py
```diff
@@ -136,6 +138,16 @@ def get_topcrashers_results(**kwargs):
                 if row['term'] in (1, -1):
                     hit['hang_count'] += row['count']
 
+            # Number of startup crashes.
+            hit['startup_percent'] = 0
+
+            sig_startup = hit['facets']['histogram_uptime']
+            for row in sig_startup:
+                if row['term'] == 0:
+                    hit['startup_percent'] = (
```
peterbe: Is there any other purpose of having this variable? 
In the django template, you could ideally move some of the business logic in here in the view code. E.g.
```python
startup_ratio = 1.0 * row['count'] / hit['count']
# here goes some comment that explains why 0.5 (aka 50%) is used and hardcoded in the code. 
hit['startup_crash'] = startup_ratio > 0.5
```

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -136,6 +138,16 @@ def get_topcrashers_results(**kwargs):
...
+                        100.0 * row['count'] / hit['count']
```
peterbe: Why don't you have to worry about `hit['count'] != 0`?

peterbe:
Seems to work. I get a bunch of rockets using stage data. 

Some stylistic things to consider about the 50% thing being put into the template code. 

r+

webapp-django/crashstats/topcrashers/views.py
AdrianGaudebert: Yep, that's right. I just copied the way it was done in the old TCBS, but yours is better.

webapp-django/crashstats/topcrashers/views.py
AdrianGaudebert: Super Search won't return any result where ``count`` is zero.

PR #2994, opened 2015-09-16
Fixes Bug 1205350 - collector2015 - allow multiple collector types

twobraids:
the collector may be called on to collect things other than breakpad type crashes.  Enable the collector to have multiple services at different URLs that can route data-payloads to any other crash storage systems.
features:
1. This change should remain backwards compatible with the existing collector.
2. Separate the URI from the implementation class, so that service implementation classes can be reused
3. Config for each service should be in its own namespace
4. each service should have its own crash storage class so that different types of data-payloads may be routed to different places
5. each service implementation should brand the raw_crash with a ```type_tag``` to further allow sorting or routing
6. the ```type_tag``` should be used as the return prefix when the collector acknowledges receipt from the submitting client
7. the breakpad collector's ```dump_id_prefix``` ("BP-") should be used as the ```type_tag``` for the breakpad crashes
8. create a new ```GenericCollector``` implementation that can accept nearly any type of submission form payload and wrap it into a json "raw_crash" -- it should accept multiple file uploads and calculate hashes for them.
9. add the ```GenericCollector``` to the integration tests to prove that we can accept generic crashes.

### DIFF
--- a/scripts/rabbitmq-integration-test-2015.sh
+++ b/scripts/rabbitmq-integration-test-2015.sh
@@ -21,22 +21,10 @@ fi
 
 function cleanup_rabbitmq() {
   echo -n "INFO: Purging rabbitmq queue 'socorro.normal'..."
-  python scripts/test_rabbitmq.py \
-      --test_rabbitmq.purge='socorro.normal' \
-      --test_rabbitmq.rabbitmq_host=$rmq_host \
-      --test_rabbitmq.rabbitmq_user=$rmq_user \
-      --test_rabbitmq.rabbitmq_password=$rmq_password \
-      --test_rabbitmq.rabbitmq_vhost=$rmq_virtual_host \
-      > /dev/null 2>&1
+  python scripts/test_rabbitmq.py --test_rabbitmq.purge='socorro.normal' --test_rabbitmq.rabbitmq_host=$rmq_host --test_rabbitmq.rabbitmq_user=$rmq_user --test_rabbitmq.rabbitmq_password=$rmq_password --test_rabbitmq.rabbitmq_vhost=$rmq_virtual_host > /dev/null 2>&1
   echo " Done."
   echo -n "INFO: Purging rabbitmq queue 'socorro.priority'..."
-  python scripts/test_rabbitmq.py \
-      --test_rabbitmq.purge='socorro.priority' \
-      --test_rabbitmq.rabbitmq_host=$rmq_host \
-      --test_rabbitmq.rabbitmq_user=$rmq_user \
-      --test_rabbitmq.rabbitmq_password=$rmq_password \
-      --test_rabbitmq.rabbitmq_vhost=$rmq_virtual_host \
-      > /dev/null 2>&1
+  python scripts/test_rabbitmq.py --test_rabbitmq.purge='socorro.priority' --test_rabbitmq.rabbitmq_host=$rmq_host --test_rabbitmq.rabbitmq_user=$rmq_user --test_rabbitmq.rabbitmq_password=$rmq_password --test_rabbitmq.rabbitmq_vhost=$rmq_virtual_host > /dev/null 2>&1
   echo " Done."
 }
 
@@ -46,6 +34,7 @@ function cleanup() {
   echo "INFO: cleaning up crash storage directories"
   rm -rf ./primaryCrashStore/ ./processedCrashStore/
   rm -rf ./crashes/
+  rm -rf ./submissions
 
   echo "INFO: Terminating background jobs"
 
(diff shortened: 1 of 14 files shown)


socorro/collector/collector_app.py
```diff
@@ -100,6 +112,78 @@ def main(self):
...
+        {
+            "name": "generic",
+            "uri": "/some/other/uri",
+            "service_implementation_class":
+                "socorro.collector.generic_collector.GenericCollector"
+        }
+        ]''',
+        doc='json-like list of services to be offered by this collector:'
+            '[{"name": sevice-name, "uri": uri, '
+            '"service_implementation_class": class-name], ...]',
```
AdrianGaudebert: One of those ``]`` wants to be a ``}`` instead.

socorro/lib/converters.py
```diff
@@ -108,6 +110,94 @@ def to_str(cls):
...
+                subordinate_namespace_names.append(service_name)
+                # create the new Namespace
+                required_config.namespace(service_name)
+                a_class_namespace = required_config[service_name]
+                a_class_namespace.add_option(
+                    "service_implementation_class",
+                    doc='fully qualified classname',
+                    default=service_implementation_class,
+                    from_string_converter=class_converter,
+                    likely_to_be_changed=True,
```
AdrianGaudebert: That's an impressive argument's name. What does it do?

socorro/unittest/collector/test_wsgi_breakpad_collector.py
```diff
@@ -495,3 +504,490 @@ def test_no_x00_character(self):
...
+        c = BreakpadCollector2015(config)
+
+        rawform = DotDict()
+        rawform.ProductName = 'FireSquid'
+        rawform.Version = '99\x00'
+        rawform.dump = DotDict({'value': 'fake dump', 'file': 'faked file'})
+        rawform.aux_dump = DotDict({'value': 'aux_dump contents', 'file': 'silliness'})
+        rawform[u'some_field\u0000'] = '23'
+        rawform[u'some_\u0000other_field'] = ObjectWithValue('XYZ')
+        rawform.uuid = '332d798f-3c42-47a5-843f-a0f892140107'
```
AdrianGaudebert: You seem to be repeating those lines over and over in your tests. Could you put them into a method instead?

socorro/unittest/collector/test_wsgi_breakpad_collector.py
```diff
@@ -495,3 +504,490 @@ def test_no_x00_character(self):
...
+        erc.timestamp = 3.0
+        erc.submitted_timestamp = '2012-05-04T15:10:00'
+        erc.throttle_rate = 100
+        erc.uuid = '332d798f-3c42-47a5-843f-a0f892140107'
+        erc.dump_checksums = {
+            'dump': '2036fd064f93a0d086cf236c5f0fd8d4',
+            'aux_dump': 'aa2e5bf71df8a4730446b2551d29cb3a',
+        }
+        erc.type_tag = 'bp'
+        erc = dict(erc)
```
AdrianGaudebert: Same here, it's being about the same lines of code for each test.

scripts/rabbitmq-integration-test-2015.sh
```diff
@@ -0,0 +1,283 @@
...
+  cleanup
+
+  exit $exit_code
+}
+
+echo -n "INFO: setting up environment..."
+source ${VIRTUAL_ENV:-"socorro-virtualenv"}/bin/activate >> setup.log 2>&1
+if [ $? != 0 ]
+then
+  fatal 1 "could activate virtualenv"
```
AdrianGaudebert: I suppose you meant ``could not activate virtualenv``?

scripts/rabbitmq-integration-test-2015.sh
```diff
@@ -0,0 +1,283 @@
...
+fi
+export PYTHONPATH=.
+echo " Done."
+
+echo -n "INFO: setting up database..."
+socorro setupdb --dropdb --force --createdb > setupdb.log 2>&1
```
AdrianGaudebert: Why not use the test database instead of the default one?

scripts/rabbitmq-integration-test-2015.sh
```diff
@@ -0,0 +1,283 @@
...
+if [ $? != 0 ]
+then
+  fatal 1 "crontabber weekly-reports-partitions failed, check setupdb.log"
+fi
+echo " Done."
+
+echo -n "INFO: configuring backend jobs..."
+for p in collector processor middleware
+do
+  cp config/${p}.ini-dist config/${p}.ini
```
AdrianGaudebert: Instead of being destructive of the user's config, why not copy it to a different, temporary folder that you would then use in the services?

scripts/rabbitmq-integration-test-2015.sh
```diff
@@ -0,0 +1,283 @@
...
+    --web_server.wsgi_server_class=socorro.webapi.servers.CherryPy \
+    --web_server.ip_address=0.0.0.0 \
+    > collector.log 2>&1 &
+socorro processor \
+    --admin.conf=./config/processor.ini \
+    --resource.rabbitmq.host=$rmq_host \
+    --secrets.rabbitmq.rabbitmq_user=$rmq_user \
+    --secrets.rabbitmq.rabbitmq_password=$rmq_password \
+    --resource.rabbitmq.virtual_host=$rmq_virtual_host \
+    --resource.postgresql.database_hostname=$database_hostname \
```
AdrianGaudebert: If I'm correct this variable uses the test database by default when running tests. So I think the ``setupdb`` you do earlier do not impact the same database.

AdrianGaudebert:
As far as I understand this PR, it's looking good to me. My concerns lie in the tests, and the main one is that the new integration test you added is destructive. I sometimes  run ``make test``, and I won't be pleased to discover that it ruined my local database and config files. And I'm quite sure it's possible to make it non-destructive.

scripts/rabbitmq-integration-test-2015.sh
twobraids: this file is basically a copy of the original ```rabbitmq-integration-test.sh``` that we've been using for years with a couple additional tests to accommodate new crash types.  

Problems with its structure/functioning from the original ought to be dealt in a separate PR.  In fact, I advocate for replacing this mess entirely.

twobraids:
Do you detect that the new second integration test works differently from the existing one?  They ought to be identical in their side effects.  If the new second integration tests destroys a database/configuration then that's because the existing one does the same thing.  Aside from the new crash types, I didn't want the new one to change the behavior of the original.

scripts/rabbitmq-integration-test.sh
```diff
@@ -64,16 +78,17 @@ echo -n "INFO: setting up environment..."
 source ${VIRTUAL_ENV:-"socorro-virtualenv"}/bin/activate >> setup.log 2>&1
 if [ $? != 0 ]
 then
-  fatal 1 "could activate virtualenv"
+  fatal 1 "couldn't activate virtualenv"
```
twobraids: hmm, I caught that misstatement here, but not in my copy of this file.

AdrianGaudebert:
r+

PR #2995, opened 2015-09-16
Bug 1205014 - Update forced versions for correlations for Firefox cycle starting 2015-09-21

KaiRo-at:
This adds current aurora and nightly versions as well as all planned beta versions for this cycle, and should land on prod early in the week of September 21.

Note that need to update the crashanalysis node with that code to actually make it live.


twobraids:
@KaiRo-at this looks fine to me r+

PR #2999, opened 2015-09-17
Fixes bug 1200596 - Added a link to signature report from report/list/.

AdrianGaudebert:
@peterbe r?

### DIFF
--- a/webapp-django/crashstats/crashstats/templates/crashstats/report_list_base.html
+++ b/webapp-django/crashstats/crashstats/templates/crashstats/report_list_base.html
@@ -9,6 +9,9 @@ Crash Reports in {{ signature }}
 
     <div class="page-heading">
         <h2>Crash Reports for {{ signature }}</h2>
+        <a href="{{ url('signature:signature_report') }}?{{ super_search_query_string }}">
+            See the equivalent Signature report page
+        </a>
         <div>
             <ul class="options">
               {% for day in [3, 7, 14, 28] %}
--- a/webapp-django/crashstats/crashstats/tests/test_views.py
+++ b/webapp-django/crashstats/crashstats/tests/test_views.py
@@ -27,7 +27,7 @@ from django.core.urlresolvers import reverse
 from django.contrib.contenttypes.models import ContentType
 
 from crashstats.base.tests.testbase import DjangoTestCase
-from crashstats.crashstats import models
+from crashstats.crashstats import models, views
 from crashstats.crashstats.management import PERMISSIONS
 from crashstats.supersearch.tests.common import (
     SUPERSEARCH_FIELDS_MOCKED_RESULTS,
@@ -243,6 +243,28 @@ def mocked_post_nohits(**options):
 
 def mocked_post_threeothersigs(**options):
     return BUG_STATUS
+
+
+class TestHelpFunctions(DjangoTestCase):
+    def test_get_super_search_style_params(self):
+        params_in = {
+            'signature': 'foo',
+            'product': ['WaterWolf'],
+            'platform': 'Linux',
+            'version': 'WaterWolf:1.0a2',
+            'end_date': datetime.datetime(2000, 1, 8),
+            'start_date': datetime.datetime(2000, 1, 1),
+        }
+        params_out = views.get_super_search_style_params(**params_in)
+        params_exp = {
+            'signature': 'foo',
+            'product': ['WaterWolf'],
+            'platform': 'Linux',
+            'version': '1.0a2',
+            'date': ['>=2000-01-08T00:00:00', '<2000-01-01T00:00:00'],
+        }
+
+        ok_(params_out, params_exp)
 
 
 class RobotsTestViews(DjangoTestCase):
@@ -3601,7 +3623,7 @@ class TestViews(BaseTestViews):
             if link.text and 'View ALL' in link.text:
                 ok_(urllib.quote_plus(sig) in link.attrib['href'])
 
-    def test_report_list_columns_offered(self):
+    def test_report_list_all_link_columns_offered(self):
         url = reverse('crashstats:report_list')
         response = self.client.get(url, {'signature': 'sig'})
         eq_(response.status_code, 200)
--- a/webapp-django/crashstats/crashstats/views.py
+++ b/webapp-django/crashstats/crashstats/views.py
@@ -4,6 +4,7 @@ import datetime
 import logging
 import math
 import isodate
+import urllib
 from collections import defaultdict
 from operator import itemgetter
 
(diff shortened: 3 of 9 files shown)


webapp-django/crashstats/crashstats/templates/crashstats/report_list_base.html
```diff
@@ -9,6 +9,12 @@
 
     <div class="page-heading">
         <h2>Crash Reports for {{ signature }}</h2>
+        &nbsp;&nbsp;
+        <em>
+            <a href="{{ url('signature:signature_report') }}?{{ super_search_query_string }}">
+                See the equivalent Signature report page
+            </a>
+        </em>
```
peterbe: That's a lot of stuff that's best done with CSS. The `&nbsp;&nbsp;` for example should be `padding-left` and the `<em>` tag to wrap a `<a>` tag is best done with `font-style:italic`

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -245,6 +245,28 @@ def mocked_post_threeothersigs(**options):
...
+        params_exp = {
+            'signature': 'foo',
+            'product': ['WaterWolf'],
+            'platform': 'Linux',
+            'version': '1.0a2',
+            'date': ['>=2000-01-08T00:00:00', '<2000-01-01T00:00:00'],
+        }
+
+        ok_(params_out, params_exp)
+
```
peterbe: Excellent!

webapp-django/crashstats/crashstats/views.py
```diff
@@ -201,6 +201,39 @@ def get_timedelta_from_value_and_unit(value, unit):
     return date_delta
 
 
+def get_super_search_style_params(**kwargs):
+    if 'signature' not in kwargs:
+        raise ValueError('"signature" is a mandatory parameter')
+
+    params = {
+        'signature': kwargs.get('signature'),
```
peterbe: This should be `kwargs['signature']` since the use of `.get()` implies that it can not be there.

webapp-django/crashstats/supersearch/form_fields.py
```diff
@@ -116,6 +118,12 @@ def value_to_string(self, value):
         except AttributeError:  # when value is None
             return value
 
+    def strptime(self, value, format):
```
peterbe: What's this for?

AdrianGaudebert: Ah, this is fixing a bug that I found as a side effect of this change. It changes the behavior of the ``DatetimeField`` class from django to use isodate.parse_datetime as a first option whenever trying to parse a datetime. By default, django doesn't understand isoformat dates (the ones with a ``T`` in the middle) and that's bothering me. I figured the easiest way to deal with that in the current system is to overwrite this method to use ``isodate``. 

An other solution was to extend the list of accepted date formats passed to this field, but in the mess of heritage around here, I failed to do that.

AdrianGaudebert:
@peterbe all things addressed.

peterbe:
r+ with nits.

webapp-django/crashstats/supersearch/form_fields.py
peterbe: Hmm... 
Considering that `settings.USE_TZ` is on by default, the regular Django `DateTimeField` will return a timezone aware datetime object. Whereas `isodate.parse_datetime(value)` won't. 
We might just be lucky that things work today, but it's dangerous territory. 
Like mixing unicode and bytestring. 

In general, the winner always; Use UTC, Use Unicode, Use timezone aware dates everywhere. 

The default Django `DateTimeField` returns a timezone aware datetime object, 
```
>>> from django.forms import DateTimeField
>>> d=DateTimeField()
>>> d.clean('2015-09-23 12:13:14')
datetime.datetime(2015, 9, 23, 12, 13, 14, tzinfo=<django.utils.timezone.LocalTimezone object at 0x10b162850>)
```
Strangely, the same test doesn't work with your class:
```
>>> from crashstats.supersearch.form_fields import DateTimeField
>>> d=DateTimeField()
>>> d.clean('2015-09-23 12:13:14')
  File "<console>", line 1, in <module>
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/supersearch/form_fields.py", line 100, in clean
    **kwargs
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/supersearch/form_fields.py", line 49, in clean
    cleaned_value = super(PrefixedField, self).clean(*args, **kwargs)
  File "/Users/peterbe/virtualenvs/socorro/lib/python2.7/site-packages/django/forms/fields.py", line 150, in clean
    value = self.to_python(value)
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/supersearch/form_fields.py", line 46, in to_python
    return super(PrefixedField, self).to_python(value)
  File "/Users/peterbe/virtualenvs/socorro/lib/python2.7/site-packages/django/forms/fields.py", line 504, in to_python
    result = super(DateTimeField, self).to_python(value)
  File "/Users/peterbe/virtualenvs/socorro/lib/python2.7/site-packages/django/forms/fields.py", line 416, in to_python
    raise ValidationError(self.error_messages['invalid'], code='invalid')
ValidationError: [u'Enter a valid date/time.']
>>> d.clean('2015-09-23T12:13:14')
  File "<console>", line 1, in <module>
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/supersearch/form_fields.py", line 100, in clean
    **kwargs
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/supersearch/form_fields.py", line 49, in clean
    cleaned_value = super(PrefixedField, self).clean(*args, **kwargs)
  File "/Users/peterbe/virtualenvs/socorro/lib/python2.7/site-packages/django/forms/fields.py", line 150, in clean
    value = self.to_python(value)
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/supersearch/form_fields.py", line 46, in to_python
    return super(PrefixedField, self).to_python(value)
  File "/Users/peterbe/virtualenvs/socorro/lib/python2.7/site-packages/django/forms/fields.py", line 504, in to_python
    result = super(DateTimeField, self).to_python(value)
  File "/Users/peterbe/virtualenvs/socorro/lib/python2.7/site-packages/django/forms/fields.py", line 416, in to_python
    raise ValidationError(self.error_messages['invalid'], code='invalid')
ValidationError: [u'Enter a valid date/time.']
```
So basically, that simple `.clean()` method doesn't work on our custom class at all.

AdrianGaudebert: I can reproduce that behavior, and I do not understand what is happening there. The thing is, it works fine in unit and manual tests, so that really confuses me! 

I've added the timezone like you suggested though.

webapp-django/crashstats/supersearch/form_fields.py
```diff
@@ -116,6 +119,13 @@ def value_to_string(self, value):
...
+        try:
+            d = isodate.parse_datetime(value)
+            return d.replace(tzinfo=utc)
```
peterbe: This could have been done on one line. But that's not important.

peterbe:
So, I did some digging into your `DateTimeField` class. The reason why the `DateTimeField().clean('2015-09-23 13:14:15')` OR `DateTimeField().clean('2015-09-23T13:14:15.123456')` doesn't work is because, by the time the `strptime` is called, the value it's given is `2`. I.e. the first character of the string input. 

Basically, I think it's because of the `MultiplePrefixedValueField` thing. I guess it means you simply can't do `.clean('some string')` on it. 

However, this works
```python
>>> DateTimeField().clean(['2015-09-23 12:13:44'])
[datetime.datetime(2015, 9, 23, 12, 13, 44, tzinfo=<django.utils.timezone.LocalTimezone object at 0x10b04b510>)]
>>> d.clean(['2015-09-23T12:13:44'])
[datetime.datetime(2015, 9, 23, 12, 13, 44, tzinfo=<UTC>)]
```

I think I don't understand how that stuff works. The prefixing things. 
One thought, is that maybe we shouldn't mess inside the `strptime` method but some other one that gets called BEFORE that, when the string comes in the first time.

AdrianGaudebert:
@peterbe tests pass, and your patch is in. Are we good here?

peterbe:
r+

PR #3002, opened 2015-09-22
Fixes bug 1207085 - Do not consider `0` as a falsy value in API.

AdrianGaudebert:
@peterbe r?

I'm also upgrading ``elasticsearch-dsl`` to fix a bug: when you were querying for 0 results, it would instead return Elasticsearch's default number of results, which is 10. That's fixed with the latest version.

### DIFF
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -674,10 +674,12 @@ class SocorroMiddleware(SocorroCommon):
 
         for param in self.get_annotated_params():
             name = param['name']
-            if param['required'] and not kwargs.get(name):
-                raise RequiredParameterError(name)
             value = kwargs.get(name)
-            if not value:
+
+            # 0 is a perfectly fine value, it should not be considered "falsy".
+            if not value and value is not 0:
+                if param['required']:
+                    raise RequiredParameterError(name)
                 continue
 
             if isinstance(value, param['type']):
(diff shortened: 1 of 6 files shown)


webapp-django/crashstats/crashstats/models.py
```diff
@@ -674,10 +674,13 @@ def kwargs_to_params(self, kwargs):
...
-                raise RequiredParameterError(name)
             value = kwargs.get(name)
-            if not value:
+
+            # 0 is a perfectly fine value, it should not be considered "falsy".
+            if param['required'] and not value and value != 0:
+                raise RequiredParameterError(name)
+
+            if not value and value != 0:
                 continue
```
peterbe: So, if this `continue` happens, does that not mean the value isn't made available? Can't you just do:
```python
if not value and value != 0:
    if param['required']:
        raise RequiredParameterError(name)
    continue
```

peterbe: I can't put my finger on it but it feels strange that you're doing the ` ... != 0` check twice.

peterbe:
@AdrianGaudebert Note that the tests don't pass.

AdrianGaudebert:
Well, TIL: 

```python
> False != 0
False
> False is not 0
True
```

AdrianGaudebert:
@peterbe comments addressed!

peterbe:
Awesome! r+

PR #3004, opened 2015-09-23
Bug 1207638 all html templates to use 4 spaces for indentation

peterbe:
@AdrianGaudebert r?
I know there's not much to "review" here. I'm just using the Beautify plugin in Atom and then I manually correct the Jinja code (because Beautify doesn't support that). 
I guess if it passes tests, merge away.


peterbe:
Actually @AdrianGaudebert lemme pile on a couple of more files. When I'm doing I'll rebase it down to one or two neat commits.

peterbe:
@AdrianGaudebert I'm going to stop there. I'll leave the bug open and continue to do more PRs against. I just don't want these PRs to linger as they can cause unnecessary merge conflicts. r?

AdrianGaudebert:
This is great! Merge away!

PR #3005, opened 2015-09-24
Fixes Bug1208129 - add proto_signature to processed crash

twobraids:
one of the eternal complaints about Socorro is the inability to search for crashes based on frames of the crashing thread's stack.  While the signature is based on the frames of the crashing thread, it is heavily edited.  The stack frames themselves are removed from the processed crash because of disk space issues.  

During the process of creating a signature, there is a point when all the normalized frame signature are assembled into a list.  This is just before the application of the skip list regular expressions.  Taking this list of normalized frame signatures and concatenating them into one long string introduces a way to search the stack for patterns in calls.  

Normally, after signature generation is complete, the list is thrown away.  Rather than that, we should explore its usefulness as a search tool.  I've dubbed this string the "proto_signature" and saved it under that name in the processed crash.

It isn't very useful for bucketing on its own, but an searchable exploratory tool it may be very useful.

### DIFF
--- a/socorro/processor/signature_utilities.py
+++ b/socorro/processor/signature_utilities.py
@@ -821,6 +821,7 @@ class SignatureGenerationRule(Rule):
             processed_crash.get('hang_type', ''),
             crashed_thread,
         )
+        processed_crash.proto_signature = ' | '.join(signature_list)
         processed_crash.signature = signature
         if signature_notes:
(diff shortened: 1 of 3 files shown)


AdrianGaudebert:
r+

I'm going to add this field to our Super Search Fields list straight away so that it gets indexed correctly from the start.

PR #3006, opened 2015-09-24
Fixes Bug 1208143 - adjustments to Collector2015 - scoping, tests, refactoring

twobraids:
the colletor2015 has some object scoping problems when used with nginx.  It seems that Apache ModWSGI and the other webservers seem to manage to keep the Socorro App object in scope.  With nginx, that object goes out of scope and is garbage collected causing all sorts of trouble for Collector2015.  The solution is to have the wrapped service classes keep a reference to global config object.  This hangs on the config object, so when the Socorro app object goes out of scope, the configman config object does not.y

In addition, with collector2015, the individual services are responsible for their own crashstorage and throttler classes. Caching them means not having to reinstantiate them for each crash received.

Finally, the 'wsgi/collector.py' file hard coded the implementation of the collector.  This made it impossible for configman to override it.  Switch it to the Generic App and then use config to choose the correct collector class.

NOTE: this PR _requires_ a config change to the collectors!
```consulate kv set socorro/collector/application 'socorro.collector.collector_app.CollectorApp'```

### CODE REVIEW

File: socorro/unittest/webapi/test_class_partial.py [hunk 1]
```diff
@@ -0,0 +1,108 @@
...
+        del d['x']
+        # the DotDictWithAcquisition should try to look to its parent
+        # for the key "bad_key", but the weakly referenced parent should be
+        # gone and raise a ReferenceError
+        try:
+            a.config.bad_key
+        except ReferenceError:
+            # this is the correct behavior
+            # 'assertRaises' was not used because we're not testing a callable
+            pass
```
AdrianGaudebert: You can still use ``assert_raises`` from ``nose.tools``, for example: 

```python
assert_raises(
    ReferenceError,
    lambda a: a.config.bad_key,
    a
)
```

AdrianGaudebert:
Looks good to me! r+ with the tiny nit of using ``assert_raises`` in tests.

PR #3008, opened 2015-09-25
fixes bug 1208172 - swap supersearch-custom-query waffle flag to negative

peterbe:
@AdrianGaudebert r?


AdrianGaudebert:
Great!

PR #3011, opened 2015-09-29
Fixes bug 1197251 - Rewrote SuperSearch filters combination. 

AdrianGaudebert:
Replaced the wilcard and match_phrase queries with filter queries, allowing them to be combined with other filters in boolean filters. 
Values of a filter now combine with "OR" instead of "AND", except for ranges.

@peterbe r?

### DIFF
--- a/socorro/external/es/supersearch.py
+++ b/socorro/external/es/supersearch.py
@@ -44,7 +44,7 @@ class SuperSearch(SearchBase):
         with self.es_context() as conn:
             return conn
 
-    def generate_list_of_indices(self, from_date, to_date, es_index=None):
+    def get_list_of_indices(self, from_date, to_date, es_index=None):
         """Return the list of indices to query to access all the crash reports
         that were processed between from_date and to_date.
 
@@ -81,7 +81,7 @@ class SuperSearch(SearchBase):
             if '<' in date.operator:
                 end_date = date.value
 
-        return self.generate_list_of_indices(start_date, end_date)
+        return self.get_list_of_indices(start_date, end_date)
 
     def format_field_names(self, hit):
         """Return a hit with each field's database name replaced by its
@@ -209,7 +209,7 @@ class SuperSearch(SearchBase):
         )
 
         # Create filters.
-        filters = None
+        filters = []
         histogram_intervals = {}
 
         for field, sub_params in params.items():
@@ -248,22 +248,47 @@ class SuperSearch(SearchBase):
                 elif param.data_type == 'str' and not param.operator:
                     param.value = [x.lower() for x in param.value]
 
+                # Operators needing wildcards, and the associated value
+                # transformation with said wildcards.
+                operator_wildcards = {
+                    '~': '*%s*',  # contains
+                    '$': '%s*',  # starts with
+                    '^': '*%s'  # ends with
+                }
+                # Operators needing ranges, and the associated Elasticsearch
+                # comparison operator.
+                operator_range = {
+                    '>': 'gt',
+                    '<': 'lt',
+                    '>=': 'gte',
+                    '<=': 'lte',
+                }
+
                 args = {}
                 filter_type = 'term'
                 filter_value = None
+
                 if not param.operator:
                     # contains one of the terms
                     if len(param.value) == 1:
                         val = param.value[0]
-                        if not isinstance(val, basestring) or (
-                            isinstance(val, basestring) and ' ' not in val
-                        ):
-                            filter_value = val
 
-                        # If the term contains white spaces, we want to perform
-                        # a phrase query. Thus we do nothing here and let this
-                        # value be handled later.
+                        if not isinstance(val, basestring) or ' ' not in val:
+                            # There's only one term and no white space, this
+                            # is a simple term filter.
+                            filter_value = val
+                        else:
+                            # If the term contains white spaces, we want to
+                            # perform a phrase query.
+                            filter_type = 'query'
+                            args = Q(
+                                'simple_query_string',
+                                query=param.value[0],
+                                fields=[name],
+                                default_operator='and',
+                            ).to_dict()
                     else:
+                        # There are several terms, this is a terms filter.
                         filter_type = 'terms'
                         filter_value = param.value
                 elif param.operator == '=':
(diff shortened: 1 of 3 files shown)


socorro/external/es/supersearch.py
```diff
@@ -81,7 +81,7 @@ def get_indices(self, dates):
             if '<' in date.operator:
                 end_date = date.value
 
-        return self.generate_list_of_indices(start_date, end_date)
+        return self.get_list_of_indices(start_date, end_date)
```
peterbe: That's a very "Lars name" :)
Why isn't it just `get_indices()`?

socorro/external/es/supersearch.py
```diff
@@ -248,22 +248,50 @@ def get(self, **kwargs):
...
                 filter_value = None
+
                 if not param.operator:
                     # contains one of the terms
                     if len(param.value) == 1:
                         val = param.value[0]
+
                         if not isinstance(val, basestring) or (
                             isinstance(val, basestring) and ' ' not in val
                         ):
```
peterbe: Wait! Isn't that the same as `if isinstance(val, basestring) and ' ' not in val:`?

socorro/external/es/supersearch.py
```diff
@@ -248,22 +248,50 @@ def get(self, **kwargs):
...
                             filter_value = val
-
-                        # If the term contains white spaces, we want to perform
-                        # a phrase query. Thus we do nothing here and let this
-                        # value be handled later.
+                        else:
+                            # If the term contains white spaces, we want to
+                            # perform a phrase query.
+                            filter_type = 'query'
+                            query = Q(
```
peterbe: Perhaps change to `args = Q(...).to_dict()` if you don't need the `query` variable.

socorro/external/es/supersearch.py
```diff
@@ -271,93 +299,48 @@ def get(self, **kwargs):
...
                     }
                 elif param.operator == '__null__':
-                    # is null
                     filter_type = 'missing'
                     args['field'] = name
+                elif param.operator in operator_wildcards:
+                    filter_type = 'query'
+
+                    if field_data['has_full_version']:
+                        name = '%s.full' % name
```
peterbe: I don't understand these two lines. Can you add a comment?

socorro/external/es/supersearch.py
```diff
@@ -271,93 +299,48 @@ def get(self, **kwargs):
...
+
+                    q_args = {}
+                    q_args[name] = (
+                        operator_wildcards[param.operator] % param.value
```
peterbe: Is `param.value` **always** a string?

socorro/external/es/supersearch.py
```diff
@@ -271,93 +299,48 @@ def get(self, **kwargs):
...
                 if filter_value is not None:
                     args[name] = filter_value
 
                 if args:
+                    new_filter = F(filter_type, **args)
                     if param.operator_not:
-                        new_filter = ~F(filter_type, **args)
-                    else:
-                        new_filter = F(filter_type, **args)
+                        new_filter = ~new_filter
```
peterbe: smart!

socorro/lib/search_common.py
```diff
@@ -375,7 +375,8 @@ def fix_version_parameter(parameters):
         # Iterate on a copy so we can delete elements while looping.
         for version in list(parameters['version']):
```
peterbe: Why does `parameters['version']` need to be converted to a list before you can iterate over it?

socorro/unittest/external/es/test_supersearch.py
```diff
@@ -426,6 +426,65 @@ def test_get_with_bool_operators(self):
...
+            'product': 'EarthRacoon',
+            'date_processed': self.now,
+        })
+        self.refresh_index()
+
+        res = self.api.get(
+            signature=['js', '~::'],
+        )
+
+        eq_(res['total'], 3)
```
peterbe: Why 3? You have 2 signatures that **don't** have `js` **and** `::`. The first and the second one (above) have `::` and `js` in them.

socorro/unittest/external/es/test_supersearch.py
```diff
@@ -426,6 +426,65 @@ def test_get_with_bool_operators(self):
...
+        eq_(len(res['hits']), 3)
```
peterbe: I think you should do something like...:
```python
eq_([x['signature'] for x in res['hits']], [sig1, sig2, sig3])
```
That way the test gets a lot easier to read.

socorro/unittest/external/es/test_supersearch.py
```diff
@@ -426,6 +426,65 @@ def test_get_with_bool_operators(self):
...
+            'app_notes': 'mozilla bar',
+            'product': 'EarthRacoon',
+            'date_processed': self.now,
+        })
+        self.refresh_index()
+
+        res = self.api.get(
+            signature=['js', '~::'],
+        )
+
```
peterbe: remove this newline since the assertions below belongs to this `res`.

socorro/unittest/external/es/test_supersearch.py
```diff
@@ -426,6 +426,65 @@ def test_get_with_bool_operators(self):
...
+            signature=['js', '~::'],
+            product=['WaterWolf', 'EarthRacoon'],
+        )
+
+        eq_(res['total'], 3)
+        eq_(len(res['hits']), 3)
+
+        res = self.api.get(
+            signature=['js', '~::'],
+            app_notes=['foo bar'],
```
peterbe: By the way, how do you actually search for something with a space in it? I.e. to find "this is foo bar" but not "foo" or "bar".

socorro/external/es/supersearch.py
AdrianGaudebert: Because ``get_indices`` already exists... :)

socorro/external/es/supersearch.py
AdrianGaudebert: Nope, but it's the same as ``if not isinstance(val, basestring) or ' ' not in val``. :)

socorro/external/es/supersearch.py
AdrianGaudebert: Yes, in this case it always is a string, the ``get_parameters`` method made sure of that.

socorro/lib/search_common.py
AdrianGaudebert: It is not converted, it is copied, so that we can alter the original list while iterating on it.

socorro/unittest/external/es/test_supersearch.py
AdrianGaudebert: Yes, that's the whole point of this PR: it changes the behavior so that it looks for everything that has ``js`` **or** ``::``. :)

socorro/external/es/supersearch.py
peterbe: Hmmm... I'll leave it to your discretion then.

socorro/lib/search_common.py
peterbe: I see.

peterbe:
r+ from me, but r- from travis.

PR #3014, opened 2015-09-30
bug 1012759 - API to audit uploaded symbols

peterbe:
@AdrianGaudebert r?

Note! This is the first time we use our Public API to expose something stored "in Django". 
I chose to not bother with the caching. First of all, this endpoint requires auth so it's quite unlikely that one of the people with a permission will abuse it. And it's simple and fast anyway. The other reason is that I didn't want to spend time inventing that when there wasn't a real need. One thing at a time. 

This commit doesn't have a "fixes" in the commit message. 
That's because, once deployed on stage and prod we need to execute the `./manage.py syncdb` run so that it picks up the new permission that I defined in the `crashstats/management/__init__.py`

Also, even when that's done, I don't think the bug is ready until the likes of Ted and Benjamin gets his group(s) configured so they can actually use this new endpoint.

### DIFF
--- a/webapp-django/crashstats/symbols/models.py
+++ b/webapp-django/crashstats/symbols/models.py
@@ -1,6 +1,10 @@
+import datetime
+
 from django.db import models
 from django.contrib.auth.models import User
 from django.utils import timezone
+
+from crashstats.crashstats import models as crashstats_models
 
 
 class SymbolsUpload(models.Model):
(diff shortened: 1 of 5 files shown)


webapp-django/crashstats/symbols/views.py
```diff
@@ -103,7 +103,7 @@ def unpack_and_upload(iterator, symbols_upload, bucket_name, bucket_location):
         key = bucket.get_key(key_name)
 
         # let's assume first that we need to add a new key
-        prefix = "+"
+        prefix = '+'
```
peterbe: This change is unrelated but it was burning my retinas when I say it was the only string not in single quotes. Call me anal :)

webapp-django/crashstats/symbols/models.py
```diff
@@ -18,3 +22,81 @@ def __repr__(self):
...
+            query = query.filter(user__email__icontains=kwargs['user_search'])
+        if kwargs.get('filename_search'):
+            query = query.filter(filename__icontains=kwargs['filename_search'])
+        if kwargs.get('content_search'):
+            query = query.filter(content__icontains=kwargs['content_search'])
+
+        query = query.select_related('user')
+        hits = []
+
+        for each in query:
```
AdrianGaudebert: This ``each`` variable looks very JavaScript-y to me.

AdrianGaudebert:
r+

webapp-django/crashstats/symbols/models.py
peterbe: fixed in rebase

PR #3015, opened 2015-10-01
fixes Bug 1210260 - correlation script compatible crashstore

twobraids:
the correlations scripts write to and then read crashes from tar files.  Make  writing and reading CrashStorage classes that produce the same kind of tar'd gziped files.

### DIFF
--- a/socorro/external/fs/crashstorage.py
+++ b/socorro/external/fs/crashstorage.py
@@ -23,6 +23,13 @@ from socorro.lib.ooid import dateFromOoid, depthFromOoid
 from socorro.lib.datetimeutil import utc_now
 from socorro.lib.util import DotDict
 
+
+def dates_to_strings_for_json(obj):
+    if isinstance(obj, datetime.datetime):
+        return obj.isoformat()
+    return json.JSONEncoder.default(self, obj)
+
+
 @contextmanager
 def using_umask(n):
     old_n = os.umask(n)
@@ -181,7 +188,7 @@ class FSRadixTreeStorage(CrashStorageBase):
         processed_crash = processed_crash.copy()
         f = StringIO()
         with closing(gzip.GzipFile(mode='wb', fileobj=f)) as fz:
-            json.dump(processed_crash, fz, default=self.json_default)
+            json.dump(processed_crash, fz, default=dates_to_strings_for_json)
         self._save_files(crash_id, {
             crash_id + self.config.jsonz_file_suffix: f.getvalue()
         })
@@ -250,12 +257,6 @@ class FSRadixTreeStorage(CrashStorageBase):
         if not os.path.exists(parent_dir):
             raise CrashIDNotFound
         shutil.rmtree(parent_dir)
-
-    @staticmethod
-    def json_default(obj):
-        if isinstance(obj, datetime.datetime):
-            return obj.strftime("%Y-%m-%d %H:%M:%S.%f")
-        raise TypeError
 
 
 class FSLegacyRadixTreeStorage(FSRadixTreeStorage):
@@ -639,17 +640,13 @@ FSPermanentStorage = FSLegacyRadixTreeStorage
 FSDatedPermanentStorage = FSLegacyDatedRadixTreeStorage
 
 
-class TarFileCrashStore(CrashStorageBase):
+#==============================================================================
+class TarFileWritingCrashStore(CrashStorageBase):
     required_config = Namespace()
     required_config.add_option(
         name='tarball_name',
         doc='pathname to a the target tarfile',
-        default='fred.tar'
-    )
-    required_config.add_option(
-        name='temp_directory',
-        doc='the pathname of a temporary directory',
-        default='/tmp'
+        default=datetime.datetime.now().strftime("%Y%m%d")
     )
     required_config.add_option(
         name='tarfile_module',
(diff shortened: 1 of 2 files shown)


socorro/unittest/external/fs/test_tar_crashstore.py
```diff
@@ -1,85 +1,143 @@
...
+        result_gzip_fp = gzip.GzipFile(
+            fileobj=result_tar_fp.extractfile(
+                '091204bd-87c0-42ba-8f58-554492141212.jsonz'
+            )
         )
-        tar_store.os_module.unlink.assert_called_once_with(
-            '/tmp/091204bd-87c0-42ba-8f58-554492141212.jsonz'
+        reconstituted_processed_crash_as_str = result_gzip_fp.read().strip()
+
+        self.assertEqual(
```
peterbe: how about using `eq_` here too?

socorro/external/fs/crashstorage.py
```diff
@@ -663,68 +659,115 @@ class TarFileCrashStore(CrashStorageBase):
...
+
+    #------------------------------------------------------------------------------
+    def _create_tarfile(self):
+        """subclasses that have a different way of openning or creating
+        the tar file pointer can override this method.  Useful for creating
+        text buffer tarfiles or using temporary files"""
+        return self.tarfile_module.open(self.config.tarball_name, 'r')
+
+    #------------------------------------------------------------------------------
+    def __init__(self, config, quit_check_callback=None):
```
peterbe: If we're not using the `quit_check_callback` perhaps we shouldn't have it as a keyword parameter. 
If we have to adhere to some "standard" the right thing to do would be to use `**kwargs` or something.

socorro/external/fs/crashstorage.py
```diff
@@ -663,68 +659,115 @@ class TarFileCrashStore(CrashStorageBase):
...
+        name='gzip_module',
+        doc='a module that supplies the gzip interface',
+        default='gzip',
+        from_string_converter=class_converter
+    )
+
+    #------------------------------------------------------------------------------
+    @staticmethod
+    def stringify_datetimes(obj):
+        if isinstance(obj, datetime.datetime):
```
peterbe: If you change this to `datetime.date` the if statement would be true for **both** `datetime.date` **and** `datetime.datetime` objects.

socorro/external/fs/crashstorage.py
```diff
@@ -663,68 +659,115 @@ class TarFileCrashStore(CrashStorageBase):
...
+            return obj.strftime("%Y-%m-%dT%H:%M:%S.%f")
```
peterbe: Can we use `return obj.isoformat()` instead? It's the same format.
```
>>> datetime.datetime.utcnow().isoformat()
'2015-10-02T13:47:27.912984'
```

socorro/external/fs/crashstorage.py
```diff
@@ -663,68 +659,115 @@ class TarFileCrashStore(CrashStorageBase):
...
-            )
+
+
+#==============================================================================
+class TarFileSequentialReadingCrashStore(CrashStorageBase):
+    required_config = Namespace()
+    required_config.add_option(
+        name='tarball_name',
+        doc='pathname to a the target tarfile',
+        default='fred.tar'
```
peterbe: It's always odd with config options where the default is non-nonsensical. I'd hate to have to block deployment on first having to change the config, when it'd be possible to get a good default right first. 

Don't we have some parameter to `add_option()` that encourages config'ors to have to change it?

socorro/external/fs/crashstorage.py
```diff
@@ -663,68 +659,115 @@ class TarFileCrashStore(CrashStorageBase):
...
+    def close(self):
+        self.tar_fp.close()
+
+    #------------------------------------------------------------------------------
+    def get_unredacted_processed(self, crash_id_ignored):
+        """we don't implement random access in this class, the next
+        one is all you get no matter what you ask for"""
+        a_tar_info_object = self.tar_fp.next()
+        if a_tar_info_object is None:
+            raise CrashIDNotFound(crash_id_ignored)
```
peterbe: This isn't very helpful. I think it should be `raise CrashIDNotFound('no next tar object available')`

socorro/external/fs/crashstorage.py
```diff
@@ -663,68 +659,115 @@ class TarFileCrashStore(CrashStorageBase):
...
-        from_string_converter=class_converter
-    )
 
+    #------------------------------------------------------------------------------
     @staticmethod
     def stringify_datetimes(obj):
         if isinstance(obj, datetime.datetime):
-            return obj.strftime("%Y-%m-%d %H:%M:%S.%f")
-        raise TypeError
+            return obj.strftime("%Y-%m-%dT%H:%M:%S.%f")
```
peterbe: ditto.

Why is the code repeated anyway? 
Why isn't this a module-level function instead then?

socorro/external/fs/crashstorage.py
```diff
@@ -663,68 +659,115 @@ class TarFileCrashStore(CrashStorageBase):
...
+        crash_id = processed_crash["crash_id"]
+
+        compressed_crash = StringIO()
+        gzip_file = self.gzip_module.GzipFile(fileobj=compressed_crash, mode='w')
+        gzip_file.write(processed_crash_as_string)
+        gzip_file.close()
+        compressed_crash.seek(0)
+        tarinfo = self.tarfile_module.TarInfo('%s.jsonz' % crash_id)
+        tarinfo.size = len(compressed_crash.getvalue())
+        self.tar_fp.addfile(tarinfo, compressed_crash)
```
peterbe: Uh?! If you've read the StringIO thing to get its length (line above), how come you don't have to `seek(0)` on it again?

socorro/external/fs/crashstorage.py
twobraids: I want all the crashstorage systems to have identically the same constructor signature.  We do use the ```quit_check_callback``` in most cases. It just wasn't needed here.  But I think it important to keep the signature a standard.

socorro/external/fs/crashstorage.py
twobraids: good point

socorro/external/fs/crashstorage.py
twobraids: sure

socorro/external/fs/crashstorage.py
twobraids: ugh, that "fred.tar" was an oversight on my part.  That should have been switched to something better.  and, yes, there is a parameter called ```likely_to_be_changed```.

socorro/external/fs/crashstorage.py
twobraids: another good point

socorro/external/fs/crashstorage.py
twobraids: that is an error.  I believe the second is supposed to raise a ```TypeError``` rather than defer to the base class.  I'll have to reread the docs on tarfiles.

socorro/unittest/external/fs/test_tar_crashstore.py
```diff
@@ -1,85 +1,143 @@
...
 
-class TestTarCrashStorage(TestCase):
 
+#==============================================================================
+class TestTarFileWritingCrashStorage(TestCase):
+
+    #--------------------------------------------------------------------------
     def setUp(self):
-        super(TestTarCrashStorage, self).setUp()
+        super(TestTarFileWritingCrashStorage, self).setUp()
```
peterbe: you can delete these two lines because there's no overriding going on.

socorro/unittest/external/fs/test_tar_crashstore.py
```diff
@@ -1,85 +1,143 @@
-from nose.tools import eq_, ok_, assert_raises
-from mock import Mock, patch
+from nose.tools import ok_, eq_
+from mock import Mock
 from datetime import datetime
 
 import json
+import tarfile
+import gzip
```
peterbe: I know we don't have any strict guidelines on style, but common practice is to separate standard lib from third-party libs.

socorro/external/fs/crashstorage.py
twobraids: one of my goals here is to remain as backwards compatible with the existing tools as possible.  I should have marked this section of code as an exact quote of the original correlation script code for producing the tar file.  I didn't want to make any changes for fear of introducing a different result.

socorro/unittest/external/fs/test_tar_crashstore.py
```diff
@@ -1,85 +1,143 @@
...
+            reconstituted_processed_crash_as_str
         )
 
 
+#==============================================================================
+class TestTarFileSequentialReadingCrashStorage(TestCase):
+
+    #--------------------------------------------------------------------------
+    def setUp(self):
+        super(TestTarFileSequentialReadingCrashStorage, self).setUp()
```
peterbe: not needed

socorro/unittest/external/fs/test_tar_crashstore.py
```diff
@@ -1,85 +1,143 @@
...
+    #--------------------------------------------------------------------------
     def setUp(self):
-        super(TestTarCrashStorage, self).setUp()
+        super(TestTarFileWritingCrashStorage, self).setUp()
 
+    #--------------------------------------------------------------------------
     def _get_config(self):
         config = DotDict()
         config.logger = Mock()
         config.tarball_name = '/tmp/a_tarball_name.tar'
```
peterbe: The `/tmp` directory is bad to hardcode. It's much better to let the OS specify the temporary directory. E.g.

```python
config.tarball_name = os.path.join(tempfile.gettempdir(), 'a_tarball_name.tar')
```

socorro/unittest/external/fs/test_tar_crashstore.py
```diff
@@ -1,85 +1,143 @@
...
+
+    #--------------------------------------------------------------------------
+    def setUp(self):
+        super(TestTarFileSequentialReadingCrashStorage, self).setUp()
+
+    #--------------------------------------------------------------------------
+    def _get_config(self):
+        config = DotDict()
+        config.logger = Mock()
+        config.tarball_name = '/tmp/a_tarball_name.tar'
```
peterbe: ditto

socorro/unittest/external/fs/test_tar_crashstore.py
```diff
@@ -1,85 +1,143 @@
...
+        config.redactor_class = Redactor
+        config.forbidden_keys = ''
+
+        return config
+
+    #--------------------------------------------------------------------------
+    def test_get_processed(self):
+        config = self._get_config()
+        config.tarfile_module = tarfile
+        config.gzip_module = gzip
```
peterbe: why these two lines? Isn't it already that by default (by looking at `_get_config()`)

socorro/unittest/external/fs/test_tar_crashstore.py
```diff
@@ -1,85 +1,143 @@
...
+
+        reconstituted_processed_crash_1 = reading_tar_store.get_processed(
+            'it does not matter what is said here'
+        )
+        eq_(processed_crash_1, reconstituted_processed_crash_1)
+
+        reconstituted_processed_crash_2 = reading_tar_store.get_processed(
+            'it does not matter what is said here'
+        )
+        eq_(processed_crash_2, reconstituted_processed_crash_2)
```
peterbe: Why are you doing two crashes? What does that test that 1 crash doesn't? If there's an explanation, perhaps a comment would be good.

socorro/external/fs/crashstorage.py
```diff
@@ -639,7 +639,8 @@ def _get_base(self, crash_id):
 FSDatedPermanentStorage = FSLegacyDatedRadixTreeStorage
 
 
-class TarFileCrashStore(CrashStorageBase):
+#==============================================================================
+class TarFileWritingCrashStore(CrashStorageBase):
```
peterbe: Either the name is misleading or we need a doc string here. 
After all, it takes the gzips the processed crash to disk.

socorro/external/fs/crashstorage.py
```diff
@@ -663,68 +659,115 @@ class TarFileCrashStore(CrashStorageBase):
...
-    )
 
+    #------------------------------------------------------------------------------
     @staticmethod
     def stringify_datetimes(obj):
         if isinstance(obj, datetime.datetime):
-            return obj.strftime("%Y-%m-%d %H:%M:%S.%f")
-        raise TypeError
+            return obj.strftime("%Y-%m-%dT%H:%M:%S.%f")
+        return json.JSONEncoder.default(self, obj)
```
peterbe: I bet this line is never called. `json.dumps` will only call this method on objects it doesn't know how to deal with.

peterbe:
I don't think it matters but all that config for being able to override `tarfile` and `gzip` modules incurs a bunch of extra code and you can avoid all of it with just using `@mock.patch('tarfil.TarFile')` in the tests. 

r+ on the code with some nits. The biggest nit is the repeated `json.dumps` class encoder which needs some love.

PR #3016, opened 2015-10-06
Fixes bug 1200596 - Added a link to signature report from report/list/

AdrianGaudebert:
@peterbe Same 2 commits that you reviewed, plus a new one that fixes that ``DateTimeField`` class that caused problems on prod.


peterbe:
r+ but I'm not sure we need to guard for `ValueError`

PR #3017, opened 2015-10-06
Fixes bug 1211875  -Added links between old and new TCBS.

AdrianGaudebert:
@peterbe r?


webapp-django/crashstats/crashstats/urls.py
```diff
@@ -59,9 +59,6 @@
     url('^topcrasher' + products + versions + '$',
         views.topcrasher,
         name='topcrasher'),
-    url('^topcrasher' + products + versions + '$',
-        views.topcrasher,
-        name='topcrasher'),
```
AdrianGaudebert: This rule was a duplicate of the one right before.

webapp-django/crashstats/crashstats/templates/crashstats/topcrasher.html
```diff
@@ -33,6 +33,9 @@
                   {% endif %}
                 {% endif %}
             </ul>
+            <a href="{{ url('topcrashers:topcrashers') }}?product={{ product }}&amp;version={{ versions[0] }}&amp;date_range_type={{ date_range_type }}&amp;process_type={{ crash_type.replace('all', 'any') }}&amp;platform={{ os_name }}&amp;_facets_size={{ result_count }}&amp;days={{ days }}">
+                See the new Top Crashers report
```
peterbe: Too verbose IMHO. If it was going to be a full sentence like that you should make it `See the <a>new Top Crashers</a> report`. 
Alternatively you can just make the link a short noun. E.g. `<a>New Top Crashers report</a>`.

webapp-django/crashstats/crashstats/templates/crashstats/topcrasher.html
```diff
@@ -33,6 +33,9 @@
                   {% endif %}
                 {% endif %}
             </ul>
+            <a href="{{ url('topcrashers:topcrashers') }}?product={{ product }}&amp;version={{ versions[0] }}&amp;date_range_type={{ date_range_type }}&amp;process_type={{ crash_type.replace('all', 'any') }}&amp;platform={{ os_name }}&amp;_facets_size={{ result_count }}&amp;days={{ days }}">
```
peterbe: Wait?! Don't we have a template helper for doing things like this?

webapp-django/crashstats/topcrashers/templates/topcrashers/topcrashers.html
```diff
@@ -18,20 +18,17 @@
...
+            <a href="{{ url(
+                'crashstats:topcrasher',
+                product=query['product'],
+                versions=query['versions'],
+                date_range_type='report',
+                crash_type=query['crash_type'].replace('any', 'all'),
+                os_name=query['os_name'],
+                result_count=query['result_count'],
+            ) }}?days={{ days }}">
+                See the old Top Crashers report
```
peterbe: Ditto

peterbe:
r+ with some nitty thoughts. Generally, I trust you got it right.

webapp-django/crashstats/crashstats/templates/crashstats/topcrasher.html
AdrianGaudebert: As far as I know, we have a helper to replace elements of the current query string, but this is building a new URL entirely. I don't see what helper could make that easier.

webapp-django/crashstats/crashstats/templates/crashstats/topcrasher.html
AdrianGaudebert: That's the exact same phrasing I have been using for report/list/ and signature report. Should I also change it there?

peterbe: Up to you. You don't have to agree. I think I spoke mainly from my gut. Since we don't have any hard rules on these kinds of things I can't state a firm opinion. 
What do you think looks best?

PR #3019, opened 2015-10-06
fixes bug 1198531 - drop laglog table

peterbe:
@AdrianGaudebert r?
Or @twobraids

### DIFF
--- a/alembic/versions/30adec758cc9_drop_lag_log_table.py
+++ b/alembic/versions/30adec758cc9_drop_lag_log_table.py
@@ -1,0 +1,32 @@
+"""drop lag_log table
+
+Revision ID: 30adec758cc9
+Revises: 4350f1383a9b
+Create Date: 2015-10-06 11:39:19.353086
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '30adec758cc9'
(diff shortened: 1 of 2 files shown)


alembic/versions/30adec758cc9_drop_lag_log_table.py
```diff
@@ -0,0 +1,32 @@
...
+
+def downgrade():
+    op.execute("""
+        CREATE TABLE lag_log (
+            replica_name text NOT NULL,
+            moment timestamp with time zone NOT NULL,
+            lag integer NOT NULL,
+            master text NOT NULL
+        )
+    """)
```
AdrianGaudebert: That should not be needed (hopefully) but this does not create the primary key of the table.

AdrianGaudebert:
r+

alembic/versions/30adec758cc9_drop_lag_log_table.py
peterbe: Just FYI, this is an affect of how alembic is tested on travis. Every migration is stepped forward and back just to see that that works. Realistically we probably don't ever care about going backwards.

PR #3021, opened 2015-10-07
fixes Bug 1179456 - ES bulk load within crashstorage context

twobraids:
retry this months later to see if it works.


AdrianGaudebert:
r+

PR #3022, opened 2015-10-07
fixes Bug 1212334 - disambiguate crash dumps being in memory or files

twobraids:
_**This PR may look daunting but it isn't that complicated.** It introduces two classes that take the place of what were simple ```dict``` mappings.  It touches a lot of files, but in a very rote manner._

There has been a bifurcation in the crash storage data throughout the history of the classes.  The crash dumps have two different representations:

    1) a mapping of crash names to binary data blobs
    2) a mapping of crash names to file pathnames.

It has been a manner of gentleman's agreement that two types do not mix. However, as the two methods have evolved in parallel the distinction has become more and more inconvenient.

This PR introduces two classes that encapsulate the two representations.  They each include a method to turn itself into the other. Any crash storage class on receiving a mapping of names/dumps can then just call the method to convert what they've got into what they want.  

In addition, each of the crashstorage classes all implemented the same thing when asked to produce the dumps as files on a file system.  This eliminates that duplication - there is only one place where a dumps are written to a file system for temporary use.

### DIFF
--- a/socorro/collector/submitter_app.py
+++ b/socorro/collector/submitter_app.py
@@ -17,7 +17,11 @@ from configman import Namespace, RequiredConfig
 from configman.converters import class_converter
 
 from socorro.app.fetch_transform_save_app import FetchTransformSaveApp, main
-from socorro.external.crashstorage_base import CrashStorageBase
+from socorro.external.crashstorage_base import (
+    CrashStorageBase,
+    FileDumpsMapping,
+    MemoryDumpsMapping,
+)
 from socorro.external.filesystem.filesystem import findFileGenerator
 from socorro.lib.util import DotDict
 from socorro.external.postgresql.dbapi2_util import execute_query_iter
@@ -59,7 +63,7 @@ class SubmitterFileSystemWalkerSource(CrashStorageBase):
         )
 
     #--------------------------------------------------------------------------
-    def get_raw_crash(self, path_tuple):
+    def get_raw_crash(self, (prefix, path_tuple)):
         """the default implemntation of fetching a raw_crash
         parameters:
            path_tuple - a tuple of paths. the first element is the raw_crash
@@ -68,13 +72,23 @@ class SubmitterFileSystemWalkerSource(CrashStorageBase):
             return DotDict(json.load(raw_crash_fp))
 
     #--------------------------------------------------------------------------
-    def get_raw_dumps_as_files(self, dump_pathnames):
+    def get_raw_dumps(self, prefix_path_tuple):
+        file_dumps_mapping = self.get_raw_dumps_as_files(prefix_path_tuple)
+        return file_dumps_mapping.as_memory_dumps_mapping()
+
+    #--------------------------------------------------------------------------
+    def get_raw_dumps_as_files(self, prefix_path_tuple):
         """the default implemntation of fetching a dump.
         parameters:
         dump_pathnames - a tuple of paths. the second element and beyond are
                          the dump pathnames"""
-        return dict(zip(self._dump_names_from_pathnames(dump_pathnames[1:]),
-                        dump_pathnames[1:]))
+        prefix, dump_pathnames = prefix_path_tuple
+        return FileDumpsMapping(
+            zip(
+                self._dump_names_from_pathnames(dump_pathnames[1:]),
+                dump_pathnames[1:]
+            )
+        )
 
     #--------------------------------------------------------------------------
     def _dump_names_from_pathnames(self, pathnames):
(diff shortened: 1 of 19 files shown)


peterbe:
Reviewer?

socorro/collector/submitter_app.py
```diff
@@ -59,22 +63,33 @@ def __init__(self, config, quit_check_callback=None):
         )
 
     #--------------------------------------------------------------------------
-    def get_raw_crash(self, path_tuple):
+    def get_raw_crash(self, prefix_path_tuple):
```
peterbe: You can write `get_raw_crash(self, (prefix, path_tuple)):`

socorro/external/boto/crashstorage.py
```diff
@@ -177,6 +178,8 @@ def build_s3_dirs(prefix, name_of_thing, crash_id):
     #--------------------------------------------------------------------------
     @staticmethod
     def do_save_raw_crash(boto_s3_store, raw_crash, dumps, crash_id):
+        if dumps is None:
+            dumps = MemoryDumpsMapping()
```
peterbe: should it perhaps, after that line, do something like `assert isinstance(dumps, MemoryDumpsMapping)`?

socorro/external/crashstorage_base.py
```diff
@@ -17,6 +18,100 @@
...
+
+    1) a mapping of crash names to binary data blobs
+    2) a mapping of crash names to file pathnames.
+
+    It has been a manner of gentleman's agreement that two types do not mix.
+    However, as the two methods have evolved in parallel the distinction has
+    become more and more inconvenient.
+
+    This class represents case 1 from above.  It is a mapping of crash dump
+    names to binay representation of the dump.  Before using a mapping, a given
```
peterbe: s/binay/binary

socorro/external/hb/crashstorage.py
```diff
@@ -10,7 +10,10 @@
```
peterbe: Why do we bother with this file?

socorro/unittest/external/boto/test_crashstorage.py
```diff
@@ -20,6 +26,8 @@
 from socorro.external.crashstorage_base import CrashIDNotFound
 import socorro.unittest.testbase
 
+TEMPDIR = tempfile.gettempdir()
```
peterbe: This pollutes. 
It leaves 
```
-rw-r--r--  1 peterbe  staff      29 Oct  7 12:29 936ce666-ff3b-4c7a-9674-367fe2120408.upload_file_minidump.TEMPORARY.dump
-rw-r--r--  1 peterbe  staff      36 Oct  7 12:29 936ce666-ff3b-4c7a-9674-367fe2120408.flash_dump.TEMPORARY.dump
-rw-r--r--  1 peterbe  staff      33 Oct  7 12:29 936ce666-ff3b-4c7a-9674-367fe2120408.city_dump.TEMPORARY.dump
``` 
in my temp directory when I run the tests. 
That's not great. 

I think what you should do is something like this:
...
Actually, to demonstrate I just wrote the code. I'll send this as a counter PR.

peterbe: This instead https://github.com/twobraids/socorro/pull/29

socorro/external/happybase/crashstorage.py
```diff
@@ -10,7 +10,9 @@
     HappyBaseConnectionContext
```
peterbe: Again, why do we bother with this class?

peterbe:
I **think** I get it. Instead of `dict` you use your own class that makes a better distinction between crash names `-->` binary blobs and crash names `-->` file paths. 
Code looks good apart from some nits

r+ if you accept my PR that stops polluting outside the tests.

socorro/external/boto/crashstorage.py
twobraids: no, because it is perfectly legitimate for a FileDumpsMapping to have been passed in at that time.  If something other than those two were passed and that thing did not have a ```as_memory_dumps_mapping``` method, then effectively, line 203 would be doing that assertion for us.   Perhaps we could assert that whatever is passed in must have a ```as_memory_dumps_mapping``` attribute.

socorro/external/hb/crashstorage.py
twobraids: because, last I heard, we have external users that are still on HBase and not on S3

twobraids:
> r+ if you accept my PR that stops polluting outside the tests.
I have, indeed, included that change in the code even though I consider it to be unnecessary. POSIX and FHS both define the use of ```/tmp``` as a place where any process may create temporary files.  They both also suggest that there be a method of cleaning the directory of abandoned files because there is no requirement for an app to remove its file once they've completed their useful life.

PR #3024, opened 2015-10-08
Fixes bug 1212851 - Link to Profile page instead of Your Crashes and …

AdrianGaudebert:
…Permissions.

@peterbe r? cc @lonnen 

Here's what it looks like:
footer-profile


webapp-django/crashstats/base/templates/crashstats_base.html
```diff
@@ -144,26 +144,26 @@
...
-                      |
-                      {{ browserid_logout(text='Sign out') }}
-                    {% else %}
-                      {# the redirect_next function comes from the browserid context processor #}
-                      {{ browserid_login(text='Sign in', color='dark', next=redirect_next()) }}
+                {% if request.user.is_authenticated() %}
+                    <a href="{{ url('profile:profile') }}"
+                      title="Your Profile"
+                    >
+                        {{ request.user.email }}
```
peterbe: I don't like that. It makes it look like `<a href="mailto:me@dot.com">me@dot.com</a>`. I know that's not true but that's what my eyes see. 
I think you should reverse the words. So it's `<a href=... title="You are signed in as {{ request.user.email }}">Your Profile</a>`

peterbe:
It's dangerous to do this without taking responsibility of the now never-linked-to Your Reports page. 
Is that page completely useless now? Can we remove its django view, its tests and its CSS (if there is any) and make that URL (permanent) redirect to `/profile/`? Also, that redirect should be commented in `urls.py` as a legacy thing which will remind us in 2016 to just delete it.

peterbe:
In your HTML because you do...
```
<a href=... 
 >
    {{ request.user.email  }}
</a>
```
there's significant whitespace between "...@mozilla.com" and the `<` which is noticable when you hover over the link.
screenshot 2015-10-08 14 36 12

Isn't there a CSS trick to avoid this? ...without having to change the template code to be `}}</a>`

peterbe:
Next, what about the "API Tokens" page. It's still linked in the footer. Not only does that need to go it also needs to be properly deprecated. 

Sorry for being a killjoy.

AdrianGaudebert:
I'm going to convoke @lonnen here, since I acted on his recommendations about the link. 

Regarding the URLs and views, I agree they need to be removed and a redirect must happen. I'll add that. As for the API Tokens page, it still has all the tools to control tokens, the Profile page only lists what you currently have. So it has to stay around.

peterbe:
Thanks for deprecating the Your Permissions page. Great!

I still think it's a bit weird to list the API tokens in two places. The user might be confused as to which is what. (when in truth, it's the same stuff just displayed in two places). Another thought that occurred is; why show a "preview" of the API tokens but not the Symbols uploaded? 

Perhaps we just miss more things to show on the Your Profile page. So we're "borrowing" snippets from other pages to fill it up. 
If that's the case, one thing I would LOVE to see on this page is a summary of the bugs associated with the crashes I have. **That** would be awesome. 

Another cool feature is something we can't do yet; to display links to reports or UUIDs you've recently visited. 

Sorry for not having crystal clear constructive criticism but I'm trying to speak my gut. Don't let things block just because I say things but please pause to consider. Perhaps one solution would be to just say `<a href="/api/tokens/">You have <b>3</b> active and <b>2</b> expired API tokens</a>`.

Lastly, I still think it's a bad idea to make the link `>email@example.com</a>` and not `title="Signed in as email@example.com">Your Profile</a>`.

peterbe:
rebase rebase rebase

AdrianGaudebert:
OK so here's another proposal for the link in the footer: 

profile-link-2

peterbe:
It's disconnected. There's a `|` character in between each when, in a sense, the "headline" for all of these is "myemail@example.com". But it's better. 

Personally I think you can drop the email address and change the `Your Profile` link to:
```
<a href="/profile" title="You're signed in as adrian@mozilla.com">Your Profile</a> | ...
```

peterbe:
r+ 

@lonnen r?

PR #3025, opened 2015-10-08
fixes Bug 1212917 - refactor iteration - push to base classes

twobraids:
unify the iterator models between the processor and submitter families of FTS apps.  Since the correlation reports will be encoded as TransformRules, this means using the processor as the base app class.  However, the processor is long running and its iteratation system doen't ever raise the StopIteration exception.  The submitter app class has the ability to allow an iterator to run to exhaustion and then quit, but doesn't have a good method of using a TransformRule system.

Push all the iteration down to the FetchTransformSaveApp class rather that repeating components through the class hierarchy.  Give the processor the ability to limit iteration to "forever", "all" or a specific number of iterations.

DO NOT BREAK BACKWARDS COMPATIBILITY WITH EXISTING CONFIGURATION.

### DIFF
--- a/socorro/app/fetch_transform_save_app.py
+++ b/socorro/app/fetch_transform_save_app.py
@@ -56,10 +56,10 @@ class FetchTransformSaveApp(App):
     # iterator 'new_crashes' and the accessors 'get_raw_crash' and
     # 'get_raw_dumps'
     required_config.source.add_option(
-      'crashstorage_class',
-      doc='the source storage class',
-      default=None,
-      from_string_converter=class_converter
+        'crashstorage_class',
+        doc='the source storage class',
+        default=None,
+        from_string_converter=class_converter
     )
     required_config.destination = Namespace()
     # For destination, the storage class should implement the 'save_raw_crash'
@@ -67,17 +67,23 @@ class FetchTransformSaveApp(App):
     # or transform methods and therefore completely redefine what api calls
     # are relevant.
     required_config.destination.add_option(
-      'crashstorage_class',
-      doc='the destination storage class',
-      default=None,
-      from_string_converter=class_converter
+        'crashstorage_class',
+        doc='the destination storage class',
+        default=None,
+        from_string_converter=class_converter
     )
     required_config.producer_consumer = Namespace()
     required_config.producer_consumer.add_option(
-      'producer_consumer_class',
-      doc='the class implements a threaded producer consumer queue',
-      default='socorro.lib.threaded_task_manager.ThreadedTaskManager',
-      from_string_converter=class_converter
+        'producer_consumer_class',
+        doc='the class implements a threaded producer consumer queue',
+        default='socorro.lib.threaded_task_manager.ThreadedTaskManager',
+        from_string_converter=class_converter
+    )
+    required_config.add_option(
+        'number_of_submissions',
+        doc="the number of crashes to submit (all, forever, 1...)",
+        short_form='n',
+        default='forever'
     )
 
     ###########################################################################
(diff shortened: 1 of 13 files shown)


socorro/app/fetch_transform_save_app.py
```diff
@@ -104,32 +110,140 @@ def get_application_defaults():
...
+    # the following methods setup a nested hierarchy of iterators.  Each layer
+    # in the hierarchy adds a feature to the enclosed iterator.
+
+    # The  iterator at the core, the innermost, will yield a list of crash_ids
+    # from some source. A subclass may define its own source and makes it
+    # available by overriding the method "_create_iter".
+    #
+    # That innermost iterator is then wrapped by the "_basic_iterator" method.
+    # this iterator changes the form of the yielded values to the (*args,
+    # **kwargs) for required by the TaskManager.  It also gives derived classes
```
twobraids: ... as required....

socorro/app/fetch_transform_save_app.py
```diff
@@ -104,32 +110,140 @@ def get_application_defaults():
...
+    # the oportunity to run callbacks between each iteration or a the point
```
twobraids: ... at the point...

socorro/app/fetch_transform_save_app.py
```diff
@@ -104,32 +110,140 @@ def get_application_defaults():
...
+    # that the innermost iterator is exhausted.
+    #
+    # The outermost lay of the iterator nesting imposes length limitations
```
twobraids: ... layer...

socorro/app/fetch_transform_save_app.py
```diff
@@ -104,32 +110,140 @@ def get_application_defaults():
...
-                            # yield None to give the caller the chance to sleep
+                arg_kwarg_form = ((x,), {})  # (args, kwargs)
+                self.config.logger.debug('yielding %s', arg_kwarg_form)
+                yield arg_kwarg_form
+            self._action_between_each_iteration()
+        else:
+            # when the iterator is exhausted, yield None as this is an
+            # indicator to some of the clients to take an action.
+            # This is a moribund action, but in this current refactoring
+            # we don't want change old behavior
```
twobraids: ... want to change...

socorro/app/fetch_transform_save_app.py
```diff
@@ -104,32 +110,140 @@ def get_application_defaults():
...
+            yield None
+        self._action_after_iteration_completes()
+
+    #--------------------------------------------------------------------------
+    def _infinite_iterator(self):
+        """this iterator wraps the "_basic_iterator" when the configuration
+        specifies that the "number_of_submissions" is set to "forever".
+        Whenever the "_basic_iterator" is exhausted, it is called again to
+        restart the iteration.  It is up to the implementation of the innermost
+        iteration to define what starting over means.  Some iterators may
```
twobraids: ... iterator to define...

socorro/app/fetch_transform_save_app.py
```diff
@@ -269,3 +382,43 @@ def main(self):
...
+        'new_crash_source_class',
+        doc='an iterable that will stream crash_ids needing processing',
+        default='',
+        from_string_converter=class_converter
+    )
+
+    #--------------------------------------------------------------------------
+    def _create_iter(self):
+        # while the base class ties the iterator to the class specified as the
+        # crash data "source", this class introduces the a different stream of
```
twobraids: ... introduces a different...

socorro/app/fetch_transform_save_app.py
```diff
@@ -104,32 +110,140 @@ def get_application_defaults():
...
     def __init__(self, config):
         super(FetchTransformSaveApp, self).__init__(config)
         self.waiting_func = None
+        # select the iterator type based on the "number_of_submissions" config
+        if config.number_of_submissions == 'forever':
+            self.source_iterator = self._infinite_iterator
+        elif config.number_of_submissions == 'all':
+            self.source_iterator = self._all_iterator
+        else:
+            self.source_iterator = self._limited_iterator
```
peterbe: You could have just...:
```python
self.source_iterator = {
    'forever': self._infinite_iterator,
    'all': self._all_iterator,
}.get(config.number_of_submissions, self._limited_iterator)
```
An advantage is that you don't have to repeat `self.source_iterator` 3 times.

socorro/app/fetch_transform_save_app.py
```diff
@@ -104,32 +110,140 @@ def get_application_defaults():
...
-        crashstorage class's 'new_crashes' method."""
-        while(True):  # loop forever and never raise StopIteration
-            for x in self.source.new_crashes():
-                if x is None:
-                    yield None
-                elif isinstance(x, tuple):
-                    yield x  # already in (args, kwargs) form
-                else:
-                    yield ((x,), {})  # (args, kwargs)
+    def _action_after_iteration_completes(self):
```
peterbe: Perhaps, again, this is a matter of style/taste but this looks like it gives an implementor a chance to override or plug in to the API. But the method is named as a private method which tells the implementor to "stay away".

socorro/app/fetch_transform_save_app.py
```diff
@@ -104,32 +110,140 @@ def get_application_defaults():
...
+                yield None
+            elif isinstance(x, tuple):
+                self.config.logger.debug('yielding %s', x)
+                yield x  # already in (args, kwargs) form
             else:
-                yield None  # if the inner iterator yielded nothing at all,
-                            # yield None to give the caller the chance to sleep
+                arg_kwarg_form = ((x,), {})  # (args, kwargs)
+                self.config.logger.debug('yielding %s', arg_kwarg_form)
+                yield arg_kwarg_form
```
peterbe: There's a lot of repetition in this block. All for the sake of items that are not None and not a tuple. 
You could reduce it down to 
```python
for x in self._create_iter():
    if is is not None and not isinstance(x, tuple):
        x = ((x,), {})
    self.config.logger.debug('yielding %r', x)
    yield x
```
(Note also how I suggest you use `%r` instead of `%s` because it's more useful from a debugging purposes. Otherwise, `None` and `'None'` both become "None`)
```

File: socorro/app/fetch_transform_save_app.py [hunk 10]
```diff
@@ -269,3 +382,43 @@ def main(self):
         self._setup_source_and_destination()
         self.task_manager.blocking_start(waiting_func=self.waiting_func)
         self._cleanup()
+
+
+#==============================================================================
+class FetchTransformSaveWithSeparateNewCrashSourceApp(FetchTransformSaveApp):
```
peterbe: a handful of a name

socorro/app/socorro_app.py
```diff
@@ -146,6 +146,7 @@ class SocorroApp(RequiredConfig):
     #--------------------------------------------------------------------------
     def __init__(self, config):
         self.config = config
+        self.app_instance_name = self._app_instance_name()
```
peterbe: what's the point of this?

socorro/collector/submitter_app.py
```diff
@@ -313,40 +263,52 @@ def source_iterator(self):
...
-            for crash_id in self.new_crash_source.new_crashes():
-                yield crash_id
-
-    #--------------------------------------------------------------------------
-    def _all_iterator(self):
-        for crash_id in self.new_crash_source.new_crashes():
-            yield crash_id
+            for crash_id in self._basic_iterator():
+                if crash_id is not None:
+                    yield crash_id
```
peterbe: Or...
```
for crash_id in super(SubmitterApp, self)._infinite_iterator():
    if crash_id is not None:
        yield crash_id
```
That way, this overrider doesn't need to know it's `self._basic_iterator`.

socorro/collector/submitter_app.py
```diff
@@ -313,40 +263,52 @@ def source_iterator(self):
...
 
     #--------------------------------------------------------------------------
     def _limited_iterator(self):
```
peterbe: How is this different from it's parent class? If it is, it should be explained in a doc string or a comment like you did above in `SubmitterApp._infinite_iterator`.

socorro/collector/submitter_app.py
```diff
@@ -313,40 +263,52 @@ def source_iterator(self):
...
+            # have to use it to lookup the crash rather than the crash_id.
+            self.config.logger.debug('paths: %s', paths)
+            raw_crash = self.source.get_raw_crash(crash_id)
+            self.config.logger.debug('raw_crash: %s', raw_crash)
+            dumps = self.source.get_raw_dumps_as_files(crash_id)
+            self.destination.save_raw_crash_with_file_dumps(
+                raw_crash,
+                dumps,
+                crash_id
+            )
```
peterbe: Cool!

socorro/unittest/processor/test_processor_app.py
```diff
@@ -45,16 +45,17 @@ def get_standard_config(self):
...
-        )
+        class FakedNewCrashSource(object):
+            def __init__(self, *args, **kwargs):
+                pass
+            def new_crashes(self):
+                return sequencer(((1,), {}),
+                                 2,  # ensure both forms acceptable
+                                 None,
+                                 ((3,), {}))()
+        config.new_crash_source.new_crash_source_class = FakedNewCrashSource
```
peterbe: nice! But you don't need the `__init__` method if you're not going to set anything, no?

socorro/app/fetch_transform_save_app.py
twobraids: good point, I like that is much more succinct

socorro/app/fetch_transform_save_app.py
twobraids: that depends on if you interpret the single underbar as meaning off limits to subclasses or not.  One of the reasons that I like C++ so much is that it has public, protected and private.  Does "private" in Python mean off limits even for derived classes? If so, then OO inheritance has been neutered to the point of unusable.  In C++, methods in classes are marked protected to specifically invite derived classes to override and offer an alternate implementation without exposing that implementation to the public interface.  In Python, if we only allow derived classes to override public methods, then we've thrown code reuse and encapsulation out the window.  We end up with fat interfaces to classes and derived classes that have to repeat code from superclasses.  At that point inheritance isn't very useful because we can accomplish the same thing with weaker duck typing.

socorro/app/fetch_transform_save_app.py
twobraids: yep, again more succinct.

socorro/collector/submitter_app.py
twobraids: yes! I've gotten you to advocate for adding another layer of indirection...

socorro/app/fetch_transform_save_app.py
peterbe: > In Python, if we only allow derived classes to override public methods then we've thrown code reuse and encapsulation out the window.

Nothing in Python is protected as private and as something you're not allowed (compiler-level) to override. The only time the `_prefix` actually does something "in terms of compiler" is when you do `from mymodule import *` then you don't get the things that start with `_`. (doesn't stop you from doing `from mymodule import _prefixed_thing`)

The `_prefix` is there to tell the code reader that this is a blackbox thing that you don't want to rely on when you override the class because there's no promises that it won't change drastically. ...even though the publicly named methods will stay as per the promised API. 

If you don't use the _prefix to tell the code reader *"Take caution, this is an "internal" thing that you might not want to mess with"* then you might as well name all methods and variables without the prefix. 

What's wrong with calling it `action_after_iteration_completes()` or `basic_iterator`? 

Perhaps you prefer the _prefix on your methods to imply something of value to the code reader but it's probably not the usual thing. Because, you're not a usual guy Lars. :)

peterbe: NOTE! This has nothing to do with me, as a code reviewer, standing in your way. Just a discussion that, if raised, my help us both become better coders.

socorro/collector/submitter_app.py
twobraids: In both case, the value None is being filtered from the stream of values being yielded from the iterator. A better solution to implement a NoOp filter function in this base class and let the SubmitterApp derived class redefine it to detect ```None```.  That way I don't have to reproduce the iterators at all in the derived class.  It's a win for code-reuse.

peterbe: +1

socorro/unittest/processor/test_processor_app.py
twobraids: yeah, it has to exist.  This class will be passed down deep into the processor code. The processor will attempt to instantiate it, passing some configuration values.  Without this constructor, the instantiation will fail because Python will fall all the way back to ```object()``` which accepts no parameters.

socorro/app/fetch_transform_save_app.py
twobraids: It is a fundamental flaw in the Python language to leave such distinctions to the whims of style rather than enforcement in syntax.  

Do you equate programmers using a class and programmers deriving from a class to be the same thing?  I do not, as they are fundamentally different use cases that get to follow different rules.

My interpretation of the single underbar is to speak to the *public* client user of the class saying stay away, this is not for consumption by the user of the class as it is an implementation detail.  However, it is an invitation to the *non-public coder* working on derivations of the class to override to create a different implementation.  If you truly want something to be strictly private and not to be overridden by a subclass, then the double underbar is available to mangle the name and stop inheritance.  I've yet to see what I consider to be a legitimate use of that feature (in the same manner that Java's ```final``` keyword is a way of denying the existence of future uses of a class).
  
We're going to have to agree to disagree about this.  My opinion on this is very deeply held stemming from literally decades of Object Orient Programming experience.

socorro/app/fetch_transform_save_app.py
twobraids: however, the point about the filtering ```None``` out in a different layer of this multilayer iterator onion has made me realize that it will log a yield statement even for items that get filtered out in a different layer.  I've removed the logging and relocated to a place with will only log from the outer most layer.

twobraids:
I've found a untested failing case here.  I'm going to fix it and update this PR

socorro/unittest/processor/test_processor_app.py
peterbe: cool. not important but that would be nice to put in a comment to explain the non-obvious.

peterbe:
r+

peterbe:
One nit :) 
If you want to put the bug number (plus the "fixes bug ") prefix that would be cool.

PR #3026, opened 2015-10-08
crashmover service should run as socorro user. if it does not and it …

rhelmer:
…starts up before collector, the ./crashes dir will be owned by root

r? @jdotpz


jdotpz:
arrrrr+

PR #3027, opened 2015-10-08
fixes bug 1212088 - Expose graphics related CSV report via webapp

peterbe:
@AdrianGaudebert r?
cc @twobraids 

I'd like to get some early eyes on this whilst I wait for the Graphics team to get back to me about which columns we can blanket out. E.g. `adu_count`. 
Once we have the most expensive columns blanket'ed out (e.g. `adu_count` and `duplicate_of`) that query will become many many times faster. 

How to test this? Run the branch and load http://socorro.dev/graphics_report/?date=2015-10-08 for example. Note that you need to be logged in. 

How to test it on the command line...
First you need to generate an API token that has the "Run long queries" permission. 
Then something like this:
```
curl -v -H "Auth-Token: 12312312312312312" http://socorro.dev/graphics_report/?date=2015-10-08
```
Or if you want to test that it gets gzipped properly, download it like this:
```
curl --compressed -v -H "Auth-Token: 12312312312312312" http://socorro.dev/graphics_report/?date=2015-10-08 > /tmp/reports.csv
```
And then you should be able open `/tmp/reports.csv` in Excel or Quip or Libreoffice or whatever you prefer.

### DIFF
--- a/socorro/external/postgresql/graphics_report.py
+++ b/socorro/external/postgresql/graphics_report.py
@@ -62,9 +62,9 @@ This was the original SQL used in the old cron job:
 SQL = """
 SELECT
     r.signature,
-    'URL (removed)' as url,        -- 1
+    NULL as url,        -- 1
     'https://crash-stats.mozilla.com/report/index/' || r.uuid as uuid_url, -- 2
-    to_char(r.client_crash_date,'YYYYMMDDHH24MI') as client_crash_date,   -- 3
+    NULL as client_crash_date,   -- 3
     to_char(r.date_processed,'YYYYMMDDHH24MI') as date_processed,         -- 4
     r.last_crash, -- 5
     r.product,    -- 6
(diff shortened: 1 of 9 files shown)


peterbe:
@AdrianGaudebert Note in the last iteration that I put adu_count sub-select back in because they have stated that they need that. 
I **suspect** that if we can change this query to `reports_clean` AND switch to using `product_adi` I think this query will perform a lot better. 
Having said that, this URL will require a token and I've prepared the graphics team about that. 

So, I suggest we land this (pending review of course!) and evaluate what to do once we see if this is hurting our PG.

socorro/external/postgresql/graphics_report.py
```diff
@@ -0,0 +1,134 @@
...
+    NULL as last_crash, -- 5
+    r.product,    -- 6
+    r.version,    -- 7
+    r.build,      -- 8
+    '' as branch, -- 9
+    r.os_name,    --10
+    r.os_version, --11
+    r.cpu_name || ' | ' || r.cpu_info as cpu_info,   --12
+    r.address,    --13
+    ARRAY(SELECT 1 WHERE FALSE) as bug_list, --14
```
AdrianGaudebert: According to the [doc](http://www.postgresql.org/docs/current/static/arrays.html), you can write ``ARRAY[]`` instead, but I haven't tested it.

socorro/middleware/middleware_app.py
```diff
@@ -76,6 +76,8 @@
     (r'/suspicious/(.*)', 'suspicious.SuspiciousCrashSignatures'),
     (r'/util/(versions_info)/(.*)', 'util.Util'),
     (r'/adi/(.*)', 'adi.ADI'),
+    # the legacy one
```
AdrianGaudebert: This comment doesn't mean much in its context.

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -5117,3 +5118,79 @@ def test_your_permissions_page(self):
...
+            '2015-10-08 23:22:21'
+        )
+        # now fetch it with gzip
+        response = self.client.get(url, data, HTTP_ACCEPT_ENCODING='gzip')
+        eq_(response.status_code, 200)
+        eq_(response['Content-Type'], 'text/csv')
+        eq_(response['Content-Length'], str(len(response.content)))
+        eq_(response['Content-Encoding'], 'gzip')
+        compressed = len(response.content)
+        ok_(compressed < length)
```
AdrianGaudebert: This block could do with some new lines, it's very very compact.

AdrianGaudebert:
r+

socorro/external/postgresql/graphics_report.py
peterbe:
```
breakpad=> select ARRAY[] As things;
ERROR:  cannot determine type of empty array
LINE 1: select ARRAY[] As things;
               ^
HINT:  Explicitly cast to the desired type, for example ARRAY[]::integer[].
breakpad=> select ARRAY(select 1 where false) As things;
 things
--------
 {}
(1 row)
```

peterbe: I tried something like that first but it didn't work. The reason I ended up with `ARRAY(SELECT 1 WHERE FALSE)` was because I found some stackoverflow post that tried to answer the same question I think I had.

socorro/middleware/middleware_app.py
peterbe: lol! Yeah, I better fix that.

peterbe: In fact, there's nothing to say really. Nothing of value. To say that it's a legacy hack, saying that belongs in the doc string of the implementation class.

PR #3028, opened 2015-10-08
fixes bug 1213058 - Link old Crashes per user reports to new one

peterbe:
@AdrianGaudebert r?

### DIFF
--- a/webapp-django/crashstats/crashstats/static/crashstats/css/daily.less
+++ b/webapp-django/crashstats/crashstats/static/crashstats/css/daily.less
@@ -5,6 +5,7 @@ h1 {
     font-size: 1.90em;
     font-weight: bold;
 }
+
 #daily_content {
     color: #000;
     font-family: Arial;
--- a/webapp-django/crashstats/crashstats/templates/crashstats/daily.html
+++ b/webapp-django/crashstats/crashstats/templates/crashstats/daily.html
@@ -21,6 +21,9 @@ Crashes per Active Daily Install for {{ product }}
 <div id="mainbody">
 
 <div class="page-heading">
+    <p class="old-new-report-link">
+        <a href="{{ new_daily_report_url }}">View the <b>new</b> <i>Crashes per user</i> report</a>
+    </p>
     <h2>Crashes per Active Daily Install</h2>
 </div>
 
--- a/webapp-django/crashstats/crashstats/static/crashstats/css/crashes_per_user.less
+++ b/webapp-django/crashstats/crashstats/static/crashstats/css/crashes_per_user.less
@@ -5,9 +5,6 @@ h1 {
     font-size: 1.90em;
     font-weight: bold;
 }
-.page-heading p.daily-report-link {
-    float: right;
-}
 #daily_content {
     color: #000;
     font-family: Arial;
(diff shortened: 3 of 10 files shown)


webapp-django/crashstats/crashstats/static/crashstats/css/daily.less
```diff
@@ -5,6 +5,10 @@ h1 {
     font-size: 1.90em;
     font-weight: bold;
 }
+.page-heading p.new-daily-report-link {
+    float: right;
+}
+
```
AdrianGaudebert: Why not use the same layout that I've used on Signature report / report/list and TCBS? That is, just putting the link right next to the page title. There's even less markup needed. 

Note that I don't mind if you prefer your solution over mine and change how it's done in other places, I just care about consistency. It's good to have patterns that users can reproduce across the website.

peterbe: You're right. I'll do that.

peterbe: Actually, I think I prefer my way of doing it. I.e. to float it to the right. I too value consistency and whatever we decide in this PR, the final result is that it should be the same for all **4** links. 

Reasons I like my link instead:
* Floating it to the right means it doesn't have any alignment issues when it's situated near the header. E.g. see how [badly it's aligned here with the buttons to the immediate left](https://www.dropbox.com/s/ne886x407wxcl95/Screenshot%202015-10-12%2012.52.53.png?dl=0)
* Floating it to the right "indicates" that it's not a part of the main app. E.g. it's not another button to re-display the data differently. 
* By displaying it with mixed bold and mix italics gives it texture to emphasize two different words within the link. E.g. in "View the <b>new</b> <i>Crashes per user</i> report" manages to highlight the name of the report in the link and which kind it is. 
* Making the whole text italic is confusing. Why is it italic (...wonders the user)? Normally italics is used to make sub-parts of a text stand out from the rest. I.e. to emphasize. When all is italics it doesn't serve the purpose it was built for. 

What do you think about that?

AdrianGaudebert: As I said, I do not mind at all going your way. You have strong arguments about it. The only thing is, there are 6 places where we have that kind of links, not 4: TCBS old and new, Daily Crashes old and new, and report/list/ and Signature Report.

peterbe: I'll take the time to correct them in all 6 places. I think it's worth it.

peterbe:
screenshot 2015-10-13 13 48 05
screenshot 2015-10-13 13 47 58
screenshot 2015-10-13 13 47 53
screenshot 2015-10-13 13 47 47
screenshot 2015-10-13 13 47 41
screenshot 2015-10-13 13 47 30

webapp-django/crashstats/crashstats/templates/crashstats/daily.html
```diff
@@ -21,6 +21,9 @@
 <div id="mainbody">
 
 <div class="page-heading">
+    <p class="new-daily-report-link">
```
AdrianGaudebert: You can use the ``old-new-report-link`` class here instead, and then remove the new styling from ``daily.less``.

AdrianGaudebert:
This is looking great! However, as of this change, those styling rules are not used anymore, so it would be nice to remove them: https://github.com/mozilla/socorro/blob/master/webapp-django/crashstats/crashstats/static/crashstats/css/screen.less#L337-L340

AdrianGaudebert:
r+ with nits.

peterbe:
Nits addressed. Just waiting for one final travis round.

PR #3030, opened 2015-10-09
fixes bug 1213334 - sequence for cursor.fetchall()

peterbe:
@twobraids r?

Thanks for the FetchAllSequence class. I really like the much shorter new code we can get out of this.

### DIFF
--- a/socorro/external/postgresql/dbapi2_util.py
+++ b/socorro/external/postgresql/dbapi2_util.py
@@ -4,6 +4,10 @@
 
 """short cuts for ugly Python DBAPI2 syntax"""
 
+from collections import Sequence
+
+from socorro.lib.util import DotDict
+
 
 #==============================================================================
 class SQLDidNotReturnSingleValue (Exception):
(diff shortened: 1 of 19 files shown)


socorro/external/postgresql/dbapi2_util.py
```diff
@@ -15,6 +17,49 @@ class SQLDidNotReturnSingleRow (Exception):
...
+        return value in self.rows
+
+    def __iter__(self):
+        for x in self.rows:
+            yield x
+
+    def __str__(self):
+        return str(self.rows)
+
+    def zipped(self):
```
twobraids: two wish list things here:
* ```zipped``` is not a very descriptive name telling me what it does. ```iterable_with_column_names``` or something like that tells me what I'm getting.
* let me optionally supply the mapping class that I want to use rather than just dict:
```python
   def zipped(self, mapping_class=dict):
       names = [x.name for x in self.description]
       return [mapping_class(zip(names, x)) for x in self.rows]
```
This would allow me to do this:
```python
for x in execute_query_fetch_all(some_connection, 'select a, b from sometable').zipped(DotDict):
    print x.a, x.b
```

twobraids:
r+

socorro/external/postgresql/dbapi2_util.py
peterbe: I actually like `zipped`. It's short and means something. `iterable_with_column_names` also means something but it's much verbose. I guess I get first dibs here because I wrote the patch :)

Regarding the override-ability of setting the `mapping_class`. Perhaps it's just me but I don't like adding code just because it's doable. I'd rather add it the day we need it. After all, our tooling (github, automatic deployment etc) is getting better. 

However, should it not **always** be `DotDict` instead of dict? DotDict isn't any slower because it's based on `dict` so the only "excess" `DotDict` has only comes in when you do the `__getitem__`. 
Actually, I think that looks like quite fancy. To be able to do...:
```python
results = self.query(sql)
for each in results.zipped():
    print each.first_name, each.last_name
```
That's after all what you'd expect from any abstraction.

twobraids: When I first saw ```zipped```, I immediately thought "compressed", as in a zip file.  I had to look at the implementation to see what you really meant.

peterbe:
Incoming update to the patch. Looks like this now:
screenshot 2015-10-09 13 28 30

peterbe:
@twobraids See the last commit. Do you like it?
I know I got your r+ but I'd like to just travis this one last time and if it passes I'll rebase it up.

twobraids:
```python
while not hell.zipped().frozen_over:
    r += 1
```

PR #3031, opened 2015-10-12
fixes bug 1213886 - Remove nose attribute annotations

peterbe:
@AdrianGaudebert r?

Not much reviewing necessary here. If it works, it works.


peterbe:
@twobraids what do you think? Do you ever use the attributes with nosetests?

AdrianGaudebert:
r+, but let's wait for @twobraids's input.

twobraids:
shrug - I never knew what those things did for us anyway.  i won't miss them. r+

PR #3032, opened 2015-10-12
fixes bug 1212969 - display copyable token without whitespace

peterbe:
@AdrianGaudebert r?

Turns out, it's [not trivial](https://css-tricks.com/fighting-the-space-between-inline-block-elements/) to get rid of those spaces between the tags. But I think I got it working.

### DIFF
--- a/webapp-django/crashstats/tokens/templates/tokens/home.html
+++ b/webapp-django/crashstats/tokens/templates/tokens/home.html
@@ -46,7 +46,8 @@
                     <time class="ago" data-date="{{ token.created.isoformat() }}">{{ token.created.isoformat() }}</time>
                 </h4>
                 <p class="code">
-                    <code>{{ token.key[:12] }}
+                    <code>
+                        <span>{{ token.key[:12] }}</span>
                         <span class="rest-hidden">{{ token.key[12:] }}</span>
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/tokens/templates/tokens/home.html
```diff
@@ -46,7 +46,8 @@
                     <time class="ago" data-date="{{ token.created.isoformat() }}">{{ token.created.isoformat() }}</time>
                 </h4>
                 <p class="code">
-                    <code>{{ token.key[:12] }}
```
AdrianGaudebert: Can you not, instead, just add HTML comments so that there is no white spaces, but still a clear indentation? Something like this: 

```html
<code>{{ token.key[:12] }}<!--
    --><span class="rest-hidden">{{ token.key[12:] }}</span><!--
    --><span class="rest-cover">...</span>
</code>
```

It seems much simpler than all these CSS tricks to me.

peterbe: But that's uglier than flexbox. That's a hack. 
I picked flexbox because of https://css-tricks.com/fighting-the-space-between-inline-block-elements/

PR #3033, opened 2015-10-13
fixes bug 1214023 - Monitor if crontabber is failing to start at all

peterbe:
@AdrianGaudebert r?
@twobraids cc

### DIFF
--- a/webapp-django/crashstats/monitoring/tests/test_views.py
+++ b/webapp-django/crashstats/monitoring/tests/test_views.py
@@ -127,11 +127,13 @@ class TestCrontabberStatusViews(BaseTestViews):
 
         def mocked_get(url, **options):
             assert '/crontabber_state/' in url
+            recently = datetime.datetime.utcnow().isoformat()
             return Response({
                 'state': {
                     'job1': {
                         'error_count': 0,
-                        'depends_on': []
+                        'depends_on': [],
+                        'last_run': recently,
                     }
                 }
             })
@@ -147,24 +149,29 @@ class TestCrontabberStatusViews(BaseTestViews):
     def test_crontabber_status_trouble(self, rget):
 
         def mocked_get(url, **options):
+            recently = datetime.datetime.utcnow().isoformat()
             assert '/crontabber_state/' in url
             return Response({
                 'state': {
                     'job1': {
                         'error_count': 1,
                         'depends_on': [],
+                        'last_run': recently,
                     },
                     'job2': {
                         'error_count': 0,
                         'depends_on': ['job1'],
+                        'last_run': recently,
                     },
                     'job3': {
                         'error_count': 0,
                         'depends_on': ['job2'],
+                        'last_run': recently,
                     },
                     'job1b': {
                         'error_count': 0,
                         'depends_on': [],
+                        'last_run': recently,
                     },
                 }
             })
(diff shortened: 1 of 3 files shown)


webapp-django/crashstats/settings/base.py
```diff
@@ -682,3 +682,11 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
 
 
 CRASH_ANALYSIS_URL = 'https://crash-analysis.mozilla.com/crash_analysis/'
+
+# At what point do we consider crontabber to be stale.
+# Ie. if it hasn't run for a certain number of hours we'd consider
+# that a failing situation and it'll trigger monitoring.
+CRONTABBER_STALE_HOURS = config(
+    'CRONTABBER_STALE_HOURS',
+    default=24
```
AdrianGaudebert: That seems very high to me as a default. Shouldn't anyone worry a lot sooner than 24 hours after crontabber started stalling? Or is it because it's possible to configure crontabber to run only once per day in some installs?

webapp-django/crashstats/monitoring/tests/test_views.py
```diff
@@ -127,11 +127,13 @@ def test_crontabber_status_ok(self, rget):
 
         def mocked_get(url, **options):
             assert '/crontabber_state/' in url
+            recently = datetime.datetime.utcnow().isoformat()
```
AdrianGaudebert: Hah, I love that variable name! So much better than ``now``!

webapp-django/crashstats/monitoring/tests/test_views.py
```diff
@@ -178,3 +185,53 @@ def mocked_get(url, **options):
         eq_(data['status'], 'Broken')
         eq_(data['broken'], ['job1'])
         eq_(data['blocked'], ['job2', 'job3'])
+
+    @mock.patch('requests.get')
+    def test_crontabber_status_not_run_for_a_while(self, rget):
+
+        then = (
```
AdrianGaudebert: On the other hand, I don't really understand the meaning of this one.

AdrianGaudebert: I mean, I do understand the meaning of the word, but it's quite unclear because the word ``then`` can have a lot of meanings, especially in programming (I'm thinking about Promises whenever I read it, for example). I'd go for something like ``yesterday`` or ``long_ago`` instead.

AdrianGaudebert:
r+ with nits and questions.

webapp-django/crashstats/settings/base.py
peterbe: Yeah, I'm not so sure myself. We do run a lot of jobs around one area of the day which we consider to be the night. However, we know we have some jobs that are supposed to run every hour. And there's even a job that's supposed to run every 5 min!

However, our goal is to reduce our dependency on cron jobs in general and go towards more run-time reports. I.e. less matviews and more supersearch. 

Heck, to make sense of it all, let's change the default down to 1 hour.

peterbe:
@AdrianGaudebert I addressed the nits and made the new default 2 hours instead. 
[See this code comment](https://github.com/mozilla/socorro/pull/3033/files#diff-10b748dce4f615dd8e2dac2be124d8ccR691)

PR #3034, opened 2015-10-13
fixes bug 1212969 - show API tokens with javascript

peterbe:
@AdrianGaudebert r?

This feels much better. No longer dependent on some hack in HTML or any hack in CSS.

### DIFF
--- a/webapp-django/crashstats/tokens/static/tokens/js/home.js
+++ b/webapp-django/crashstats/tokens/static/tokens/js/home.js
@@ -1,14 +1,28 @@
 $(function() {
-    $('p.code button').click(function(event) {
+
+    'use strict';
+
+    $('.token p.code[data-key]').each(function() {
+        var p = $(this);
+        p.prepend(
+            $('<code>')
+            .addClass('truncated')
+            .text(p.data('key').substr(0, 12) + '…')
+        ).prepend(
+            $('<code>')
+            .addClass('whole')
+            .text(p.data('key'))
+            .hide()
+        );
+    });
+
+    $('.token').on('click', 'p.code button', function(event) {
         event.preventDefault();
         var $button = $(this);
         var prev_text = $button.text();
         $button.text($button.data('toggle'));
         $button.data('toggle', prev_text);
-
-        var $parent = $button.parents('p.code');
-        $('.rest-hidden', $parent).toggle();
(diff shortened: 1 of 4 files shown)


AdrianGaudebert:
I really find this overly complicated. I was going to argue against it (in favor of dead simple HTML comments) but the one thing that sold me is that with this, you can double-click the token to select it (which doesn't work with other ways because the token is considered being two words).

webapp-django/crashstats/tokens/static/tokens/js/home.js
```diff
@@ -1,14 +1,29 @@
...
-    $('p.code button').click(function(event) {
+
+    'use strict';
+
+    $('.token p.code[data-key]').each(function() {
+        var p = $(this);
+        p.prepend(
+            $('<code>')
+            .addClass('truncated')
+            .on('click')
```
AdrianGaudebert: What is this doing?

webapp-django/crashstats/tokens/static/tokens/js/home.js
```diff
@@ -1,14 +1,29 @@
...
+            .text(p.data('key').substr(0, 12) + '…')
+        ).prepend(
+            $('<code>')
```
AdrianGaudebert: I would have written this like that, what do you think? 

```javascript
$('<code>', {'class': 'whole', text: p.data('key')})
```

AdrianGaudebert:
r+ with a question on the ``on('click')`` that doesn't seem to do anything.

webapp-django/crashstats/tokens/static/tokens/js/home.js
peterbe: ha! good catch! I used to have it so that you can click the `<code>` to have the same affect as the `<button>` next to it. But it didn't work because it made it weird when you fumbled and tried to highlight it.

webapp-django/crashstats/tokens/static/tokens/js/home.js
peterbe: I never know which is better. Perhaps it's just personal taste but I like the `.addClass()` and the `.text()` APIs separately. Choices are annoying. If you don't mind, I'll keep this style.

peterbe:
nits addressed. just waiting for travis one last time.

PR #3035, opened 2015-10-14
fixes bug 1214436 - Ability to sync featured versions from prod

peterbe:
@AdrianGaudebert r?

So this is the least hacky thing I can image we start with. I first tried to write it as a script that anybody could own and run with an API token. But then I needed a new permission (and possibly tie that to a group) and our FeaturedReleases model couldn't accept anything other than multipart form.

I say we make this available and it only shows on stage or localhost and it'll make stage look more sane and should solve the problem for QA.

### DIFF
--- a/webapp-django/crashstats/manage/static/manage/js/featured-versions.js
+++ b/webapp-django/crashstats/manage/static/manage/js/featured-versions.js
@@ -2,14 +2,14 @@
 (function ($, document) {
     'use strict';
 
-    function submit_form($form) {
+    function submitForm() {
         // check that every product has at least one featured version
         var no_empty = true;
         $('.product').each(function() {
             if ($('input[type="checkbox"]', this).length && !$('input[type="checkbox"]:checked', this).length) {
                 var product = $('h4', this).text();
                 no_empty = false;
-                alert("No versions selected for " + product);
+                alert('No versions selected for ' + product);
             }
         });
         return no_empty;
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/manage/static/manage/js/featured-versions.js
```diff
@@ -19,8 +19,40 @@
...
+                $('.featured-on-prod button').on('click', function() {
+                    $('input[type="checkbox"]:checked').attr('checked', false);
+                    $('tbody tr').each(function() {
+                        var product = $('td', this).eq(0).text();
+                        var version = $('td', this).eq(1).text();
+                        if (featured[product].indexOf(version) > -1) {
+                            $('input[type="checkbox"]', this).attr('checked', true);
+                        }
+                    });
+                    $('.featured-on-prod p:hidden').show(600);
```
AdrianGaudebert: This doesn't work, it's not ``hidden`` as far as I can tell.

AdrianGaudebert:
It looks good to me. The filtering on the domain name is kinda weird, but I agree it looks like a very simple solution. What worries me is that it still requires a manual step, and I'm afraid we might forget to do it. Or QA will see a failing test and have to go on stage to fix the featured version before re-running the test. It's still a better solution, but clearly not perfect yet.

webapp-django/crashstats/manage/static/manage/js/featured-versions.js
peterbe: Will fix

AdrianGaudebert:
r+

webapp-django/crashstats/manage/static/manage/js/featured-versions.js
peterbe: Actually, the `p` tag looked like this:
```
<p style="display:none">
    Checkboxes updated to match <code>crash-stats.mozilla.com</code>.<br>
    Verify and press <b>Update Featured Versions</b> below.
</p>
```

peterbe:
nits addressed. Going to merge.

PR #3036, opened 2015-10-14
fixes bug 1214718 - SuperSearch should guard against massive _results_number

peterbe:
@AdrianGaudebert What should the number be? Should it be tied to the "Run long queries" permission?

### DIFF
--- a/socorro/external/es/supersearch.py
+++ b/socorro/external/es/supersearch.py
@@ -224,6 +224,8 @@ class SuperSearch(SearchBase):
                         results_from = param.value[0]
                     elif param.name == '_results_number':
                         results_number = param.value[0]
+                        if results_number > 1000:
+                            raise BadArgumentError('_results_number too large')
                     elif param.name == '_facets_size':
                         facets_size = param.value[0]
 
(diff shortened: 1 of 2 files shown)


socorro/external/es/supersearch.py
```diff
@@ -224,6 +224,8 @@ def get(self, **kwargs):
                         results_from = param.value[0]
                     elif param.name == '_results_number':
                         results_number = param.value[0]
+                        if results_number > 1000:
```
AdrianGaudebert: 1000 is good, that's what I thinking of doing.

AdrianGaudebert:
This should not be tied to any permission, because anyone can still get all the results they want by doing some pagination. We just refuse that people get too many results at once.

AdrianGaudebert:
Actually, we could do something completely different. Whenever the number of results is too high, we could split the query and use some river features of Elasticsearch to still get all the results, but as an iteration of smaller chunks. That doesn't solve the fact that we will be returning a huge JSON blob to the client, but at least the strain on our ES will be a lot smaller. 

I wonder if that's worth doing though. Maybe it's still fine to just tell users that they have to do that pagination effort themselves?

PR #3037, opened 2015-10-14
fixes bug 1214023 - crontabber monitoring on date comparison

peterbe:
@lonnen or @AdrianGaudebert r?

I'd like this push this out sooner rather than later. Hence calling for your attention @lonnen. Once this is in I can push to prod.


peterbe:
Basically, once this lands on stage https://crash-stats.allizom.org/monitoring/crontabber/ will start to work again.

lonnen:
:arrow_double_up:

peterbe:
I'll push to prod once this goes into stage.

PR #3038, opened 2015-10-15
Fixes bug 1214558 - Added a first prototype of a Summary tab.

AdrianGaudebert:
@peterbe r?

This is the first step, I'm adding no new features to Super Search, no new data, just using existing features. I could reproduce 4 features perfectly (architecture, process type, flash version and exploitability), and an imperfect operating system summary (it's missing more precise OS names). 

I'm now going to work on adding what's needed to make the remaining 5 features.

### DIFF
--- a/webapp-django/crashstats/signature/static/signature/js/signature_tab_summary.js
+++ b/webapp-django/crashstats/signature/static/signature/js/signature_tab_summary.js
@@ -1,0 +1,20 @@
+/**
+ * Tab for displaying signature summary.
+ * Does not have any panels.
+ * Does not have any controls.
+ *
+ * @extends {SignatureReport.Tab}
+ * @inheritdoc
+ */
+SignatureReport.SummaryTab = function(tabName) {
+
+    var config  = {
+        'panels': false,
+        'dataDisplayType': 'table',
+        'pagination': false
+    };
+
+    SignatureReport.Tab.call(this, tabName, config);
+};
+
+SignatureReport.SummaryTab.prototype = SignatureReport.inherit(SignatureReport.Tab.prototype);
--- a/webapp-django/crashstats/signature/templates/signature/signature_summary.html
+++ b/webapp-django/crashstats/signature/templates/signature/signature_summary.html
@@ -1,0 +1,161 @@
+<div class="panel">
+    <header class="title">
+        <h2>Operating System</h2>
+    </header>
+    <div class="content">
+        <table class="tablesorter data-table">
+            <thead>
+                <tr>
+                    <th class="header">Operating System</th>
+                    <th class="header">Count</th>
+                    <th class="header">Percentage</th>
+                </tr>
+            </thead>
+            <tbody>
+                {% for platform in query.facets.platform %}
+                <tr>
+                    <td>{{ platform.term }}</td>
+                    <td>{{ platform.count }}</td>
+                    <td>{{ (100.0 * platform.count / query.total) | round(1) }}%</td>
+                </tr>
+                {% endfor %}
+            </tbody>
+        </table>
+    </div>
+</div>
+
+<div class="panel">
+    <header class="title">
+        <h2>Uptime Range</h2>
+    </header>
+    <div class="content">
+        <table class="tablesorter data-table">
+            <thead>
+                <tr>
+                    <th class="header">Uptime Range</th>
+                    <th class="header">Count</th>
+                    <th class="header">Percentage</th>
+                </tr>
+            </thead>
+            <tbody>
+                {% for uptime in uptimes %}
+                <tr>
+                    <td>{{ uptime.term }}</td>
+                    <td>{{ uptime.count }}</td>
+                    <td>{{ (100.0 * uptime.count / query.total) | round(1) }}%</td>
+                </tr>
+                {% endfor %}
+            </tbody>
+        </table>
+    </div>
+</div>
+
+<div class="panel">
+    <header class="title">
+        <h2>Architecture</h2>
+    </header>
+    <div class="content">
+        <table class="tablesorter data-table">
+            <thead>
+                <tr>
+                    <th class="header">Architecture</th>
+                    <th class="header">Count</th>
+                    <th class="header">Percentage</th>
+                </tr>
+            </thead>
+            <tbody>
+                {% for cpu_name in query.facets.cpu_name %}
+                <tr>
+                    <td>{{ cpu_name.term }}</td>
+                    <td>{{ cpu_name.count }}</td>
+                    <td>{{ (100.0 * cpu_name.count / query.total) | round(1) }}%</td>
+                </tr>
+                {% endfor %}
+            </tbody>
+        </table>
+    </div>
+</div>
+
+<div class="panel">
+    <header class="title">
+        <h2>Process Type</h2>
+    </header>
+    <div class="content">
+        <table class="tablesorter data-table">
+            <thead>
+                <tr>
+                    <th class="header">Process Type</th>
+                    <th class="header">Count</th>
+                    <th class="header">Percentage</th>
+                </tr>
+            </thead>
+            <tbody>
+                {% for process_type in query.facets.process_type %}
+                <tr>
+                    <td>{{ process_type.term }}</td>
+                    <td>{{ process_type.count }}</td>
+                    <td>{{ (100.0 * process_type.count / query.total) | round(1) }}%</td>
+                </tr>
+                {% endfor %}
+            </tbody>
+        </table>
+    </div>
+</div>
+
+<div class="panel">
+    <header class="title">
+        <h2>Flash&trade; Version</h2>
+    </header>
+    <div class="content">
+        <table class="tablesorter data-table">
+            <thead>
+                <tr>
+                    <th class="header">Flash&trade; Version</th>
+                    <th class="header">Count</th>
+                    <th class="header">Percentage</th>
+                </tr>
+            </thead>
+            <tbody>
+                {% for flash_version in query.facets.flash_version %}
+                <tr>
+                    <td>{{ flash_version.term }}</td>
+                    <td>{{ flash_version.count }}</td>
+                    <td>{{ (100.0 * flash_version.count / query.total) | round(1) }}%</td>
+                </tr>
+                {% endfor %}
+            </tbody>
+        </table>
+    </div>
+</div>
+
+{% if query.facets.histogram_date %}
+<div class="panel">
+    <header class="title">
+        <h2>Exploitability</h2>
+    </header>
+    <div class="content">
+        <table class="tablesorter data-table">
+            <thead>
+                <tr>
+                    <th class="header">Report Date</th>
+                    <th class="header">None</th>
+                    <th class="header">Low</th>
+                    <th class="header">Medium</th>
+                    <th class="header">High</th>
+                </tr>
+            </thead>
+            <tbody>
+                {% for day in query.facets.histogram_date %}
+                <tr>
+                    <td>{{ day.term[:10] }}</td>
+                    <td>{{ day.exploitability.none }}</td>
+                    <td>{{ day.exploitability.low }}</td>
+                    <td>{{ day.exploitability.medium }}</td>
+                    <td>{{ day.exploitability.high }}</td>
+                </tr>
+                {% endfor %}
+            </tbody>
+        </table>
+    </div>
+</div>
+{% endif %}
--- a/webapp-django/crashstats/signature/views.py
+++ b/webapp-django/crashstats/signature/views.py
@@ -1,7 +1,8 @@
+import datetime
+import functools
+import isodate
 import math
 import urllib
-import isodate
-import datetime
 
 from django import http
 from django.core.urlresolvers import reverse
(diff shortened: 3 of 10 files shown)


peterbe:
Great start!
I've tested it and it looks pretty good.

peterbe:
The imperfections in the design of the signature report is bothersome. 
screenshot 2015-10-15 07 29 16
If you count the logo, there are **5 different indentations** going on here. I know it's not as important as the content but it looks like low-quality and it matters in terms of convincing people to change the tools they're used to. Not specific to this PR per se but nevertheless important.

peterbe:
Why is there a rank? Perhaps I'm naîve about how things work but to me, "rank" means there's a better or worse. It doesn't feel natural to say that certain uptime ranges are better or worse than others. (Especially not if you say that `> 1 hour` is ranked higher than `15-60 min`)
screenshot 2015-10-15 07 33 47

peterbe:
I actually liked it better when the `<th>` would say "Percentage" and the numbers in the `<td>` would have the `%` suffix. Makes it look more professional. Non-suffixed (or non-prefixed like $123.45) numbers are reserved for counts or iterations or spreadsheets that are very very restrained in space and richness (e.g. a CSV file or a tiny spreadsheet that has to fit on a tiny device screen)

screenshot 2015-10-15 07 36 44

webapp-django/crashstats/signature/templates/signature/signature_summary.html
```diff
@@ -0,0 +1,171 @@
...
+                    <th class="header">%</th>
+                </tr>
+            </thead>
+            <tbody>
+                {% for platform in query.facets.platform %}
+                <tr>
+                    <td>{{ loop.index }}</td>
+                    <td>{{ platform.term }}</td>
+                    <td>{{ platform.count }}</td>
+                    <td>{{ (platform.count / query.total * 100.0) | round(2) }}</td>
```
peterbe: If count is 3 and total is 100 that should yield 3.0% right?
```
>>> 3 / 10 * 100.0
0.0
```
Perhaps these numbers aren't `int`s but that's not easy to see. 
A simple solution would be to change to 
```
<td>{{ (100.0 * platform.count / query.total) | round(2) }}</td>
```

webapp-django/crashstats/signature/templates/signature/signature_summary.html
```diff
@@ -0,0 +1,171 @@
...
+        <h2>Operating System</h2>
+    </header>
+    <div class="content">
+        <table class="tablesorter data-table">
+            <thead>
+                <tr>
+                    <th class="header">Rank</th>
+                    <th class="header">Operating System</th>
+                    <th class="header">Count</th>
+                    <th class="header">%</th>
```
peterbe: I made it as a general comment before but I think it's better to do it inline. 
I think "Percentage" looks better than "%". The "%" notation is "lazy". 
If it was about money/currency I think the column should be "Amount" (or "Debt" or "Profit" or whatever) and not "$" (or "£").

webapp-django/crashstats/signature/templates/signature/signature_summary.html
peterbe: Unless you have a strong reason to keep it, I would urge that we switch to a single decimal point. The 2 decimal notation is defacto standard for monetary display. E.g. "$123.45". 
If the accuracy is so important then 2 decimal points is far too little, like for scientific measurements. 

I would advocate we switch to 1 decimal point. Especially for percentages.

peterbe:
Any idea why the uptime range numbers are so different?
screenshot 2015-10-15 07 54 55
screenshot 2015-10-15 07 55 12

peterbe:
Man! It's fast! 
I'm just using stage ES which is smaller but it feels really snappy at the moment.

webapp-django/crashstats/signature/static/signature/js/signature_tab_summary.js
```diff
@@ -0,0 +1,21 @@
...
+        'panels': false,
+        'dataDisplayType': 'table',
+        'pagination': false
+    };
+
+    SignatureReport.Tab.call(this, tabName, config);
+
+};
+
+SignatureReport.SummaryTab.prototype = SignatureReport.inherit(SignatureReport.Tab.prototype);
```
peterbe: Is it just me, or does this whole code look quite complex. That's a lot of object orientation, that makes my head spin, just to be able to get out that it's 3 key-value pairs of config data.

webapp-django/crashstats/signature/views.py
```diff
@@ -424,3 +424,112 @@ def signature_graph_data(request, channel):
...
     return data
+
+
+@waffle_switch('signature-report')
+def signature_summary(request):
+    '''Return a list of specific aggregations. '''
+    params = get_validated_params(request)
+    if isinstance(params, http.HttpResponseBadRequest):
+        # There was an error in the form, let's return it.
+        return params
```
peterbe: I know I'm guilty of this pattern too, but a more pythonic thing would be to change `get_validated_params` to raise a `ValidationError(problem)` or something and change these lines to:
```python
try:
    params = get_validated_params(request)
except ValidationError as exc:
    return http.HttpResponseBadRequest(exc)
```

peterbe: Or, if this is going to happen in several other views, make the `get_validated_params` a decorator and change line 430 to `def signature_summary(request, params):`

webapp-django/crashstats/signature/views.py
```diff
@@ -424,3 +424,112 @@ def signature_graph_data(request, channel):
...
+
+    signature = params['signature'][0]
+
+    data = {}
+
+    current_query = request.GET.copy()
+    if 'page' in current_query:
+        del current_query['page']
+
+    data['params'] = current_query.copy()
```
peterbe: Copy a copy?

webapp-django/crashstats/signature/views.py
```diff
@@ -424,3 +424,112 @@ def signature_graph_data(request, channel):
...
+    ):
+        params['_histogram.date'] = ['exploitability']
+
+    api = SuperSearchUnredacted()
+    try:
+        search_results = api.get(**params)
+    except models.BadStatusCodeError, e:
+        # We need to return the error message in some HTML form for jQuery to
+        # pick it up.
+        return http.HttpResponseBadRequest('<ul><li>%s</li></ul>' % e)
```
peterbe: Why is the `BadStatusCodeError` the client's fault?

webapp-django/crashstats/signature/views.py
```diff
@@ -424,3 +424,112 @@ def signature_graph_data(request, channel):
...
+                if uptime['term'] >= uptime_seconds:
+                    uptimes_count[uptime_minutes] += uptime['count']
+                    break
+
+        labels = {
+            0: '< 1 min',
+            1: '1-5 min',
+            5: '5-15 min',
+            15: '15-60 min',
+            60: '> 1 hour'
```
peterbe: Can you move this to above line 479 and do something like 
```
uptimes_count = dict((x, 0) for x in labels)
```
That way you don't have to repeat the "bucket keys".

webapp-django/crashstats/signature/views.py
```diff
@@ -424,3 +424,112 @@ def signature_graph_data(request, channel):
...
+            'medium': 0,
+            'high': 0,
+        }
+        for day in facets['histogram_date']:
+            exploitability = dict(exploitability_base)
+            for expl in day['facets']['exploitability']:
+                if expl['term'] in exploitability:
+                    exploitability[expl['term']] = expl['count']
+            day['exploitability'] = exploitability
+            del day['facets']['exploitability']
```
peterbe: Why delete? That memory will get garbage collected at the end of the wsgi request.

peterbe:
No final r yet. I think there's quite a few things to consider.

webapp-django/crashstats/signature/views.py
AdrianGaudebert: This function is used in a lot of place. I agree that it's not Pythonic, and I don't remember why but I know I had a "good" reason to do it like this. I don't know if doing this refactoring in a PR that's already 450 lines long is a good idea though.

webapp-django/crashstats/signature/static/signature/js/signature_tab_summary.js
AdrianGaudebert: Some day we'll write ES6 code and this will look a lot better. This is the result of Thalia's work, and it's much better than before (all tabs used to be in one big file). I think it can certainly be made better, but that doesn't look like a priority for now.

webapp-django/crashstats/signature/views.py
AdrianGaudebert: Seriously, that's copy/pasting from what we do in all other Super Search-like views. It gets sent back as AJAX, so a user will never see an actual 400 page.

AdrianGaudebert:
Percentage fixed, rank removed, and I just made the tab nav aligned with the rest of the page.

webapp-django/crashstats/signature/views.py
peterbe: We need to fix that. It's warped and just wrong. 
That idea of showing a bullet point list back to the user is commonly something we do to describe validation errors that are caused by the user. In this case, it's our fault that the squirrels in the machines are on strike.

Perhaps I just don't get the full context and this PR might not be the right place to fix the world but we ought to not forget this and address it at some point soon.

webapp-django/crashstats/signature/views.py
peterbe: Hehe. It's the bane of our life! On the one hand we love short and few-line-PRs more than we love oxygen. On the other hand, PRs that repeat old common bad habits is our only chance to tackle technical debt. 

Just forget about it for now. The PR is already long. I often tackle refactoring/technical debt all-of-a-sudden. I.e. I have some downtime between important feature work and just attack it.

AdrianGaudebert: I've fixed that problem.

webapp-django/crashstats/signature/views.py
AdrianGaudebert: There's actually a bug about fixing those issues: https://bugzilla.mozilla.org/show_bug.cgi?id=968392

AdrianGaudebert:
@peterbe r?

webapp-django/crashstats/signature/views.py
```diff
@@ -35,36 +37,36 @@
...
 
-    return params
+        kwargs['params'] = params
+        return view(request, *args, **kwargs)
+    return inner
 
 
 @waffle_switch('signature-report')
+@get_validated_params
 @pass_default_context
```
peterbe: I wonder if this is needed? Isn't that just there so that you can get access to the current products and versions. Try removing it (and the `default_context=None`) and see if it all continues to work.

webapp-django/crashstats/signature/views.py
```diff
@@ -35,36 +37,36 @@
 )
 
 
-def get_validated_params(request):
-    params = get_params(request)
-    if isinstance(params, http.HttpResponseBadRequest):
-        # There was an error in the form, let's return it.
-        return params
+def get_validated_params(view):
```
peterbe: "get_..." is a bad name. It's a decorator now. Consider "check_..." (or "pass_..." for somewhat consistency)

webapp-django/crashstats/signature/views.py
```diff
@@ -35,36 +37,36 @@
...
+            if not params['signature'] or not params['signature'][0]:
+                raise ValidationError(
+                    '"signature" parameter is mandatory'
+                )
+        except ValidationError, e:
+            return http.HttpResponseBadRequest(str(e))
 
-    return params
+        kwargs['params'] = params
+        return view(request, *args, **kwargs)
```
peterbe: It's such a critical thing this, that I wonder if it should be this instead:
```
args += (params,)
```
and the views should be
```
def signature_report(request, params):
```
It always burns and hurts my eyes to see something like this:
```python
def myfunc(thing=None):
    print thing.attribute 
```

peterbe: A note about the use of `+= (params,)` is that suppose you have a view that is defined to take in an ID or something in the `urls.py` then it would end up looking like this:

```python
@get_validated_params
def my_view(request, id, params):
```

webapp-django/crashstats/supersearch/views.py
```diff
@@ -347,11 +352,13 @@ def search_custom(request, default_context=None):
     '''Return the basic search page, without any result. '''
     error = None
     query = None
-    params = get_params(request)
-    if isinstance(params, http.HttpResponseBadRequest):
+
+    try:
+        params = get_params(request)
+    except ValidationError, e:
```
peterbe: `except ValidationError as e:` is the futuristic syntax

peterbe:
r+ with nits

webapp-django/crashstats/signature/views.py
AdrianGaudebert: It is needed to display the global layout of Socorro (``crashstats_base.html``).

webapp-django/crashstats/signature/views.py
AdrianGaudebert: Actually, that last comment you wrote is incorrect. The parameter that you define in urls is passed as a keyword argument, so you have to write: 
```python
@get_validated_params
def my_view(request, params, id)
```

PR #3040, opened 2015-10-15
fixes bug 1123905 - Stop depending on test-utils

peterbe:
@AdrianGaudebert r?

### DIFF
--- a/webapp-django/crashstats/settings/__init__.py
+++ b/webapp-django/crashstats/settings/__init__.py
@@ -1,7 +1,20 @@
-from .base import *
+"""
+By default, crashstats.settings.base has the defaults, but within that file
+those defaults are overwritten by reading from the environment.
+Lastly, all settings in crashstats.settings.test overrides lastly meaning,
+if specified in crashstats.settings.test, that it doesn't matter what's
+in crashstats.settings.base or the environment.
+This makes tests more predictable because we can avoid local settings
+as much as possible.
+"""
+
+import sys
+
+from .base import *  # NOQA
+
 # TODO remove this whole try/except when we can safely stop using local.py
 try:
-    from .local import *
+    from .local import *  # NOQA
     import warnings
     warnings.warn(
         "Use environment variables or a .env file instead of local.py",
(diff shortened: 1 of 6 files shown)


webapp-django/crashstats/settings/test.py
```diff
@@ -15,7 +22,7 @@
 
 # by scrubbing this to something unreal, we can be certain the tests never
 # actually go out on the internet when `request.get` should always be mocked
-MWARE_BASE_URL = 'http://shouldnotactuallybeused'
```
peterbe: oops. I did a tiny change just to assert to myself that it actually worked.

webapp-django/crashstats/settings/__init__.py
```diff
@@ -1,7 +1,20 @@
...
+Lastly, all settings in crashstats.settings.test overrides lastly meaning,
+if specified in crashstats.settings.test, that it doesn't matter what's
+in crashstats.settings.base or the environment.
+This makes tests more predictable because we can avoid local settings
+as much as possible.
+"""
+
+import sys
+
+from .base import *  # NOQA
```
AdrianGaudebert: What's this comment?

webapp-django/crashstats/settings/test.py
```diff
@@ -1,12 +1,19 @@
...
+Ultimately, it helps make sure you never actually use real URLs. For
+example, if a test incorrectly doesn't mock `requests.get()` for
+example, it shouldn't actually try to reach out to a real valid URL.
+"""
 
 CACHE_MIDDLEWARE = True
 CACHE_MIDDLEWARE_FILES = False
 
-import os
-os.environ['FORCE_DB'] = 'true'
```
AdrianGaudebert: Can you explain this change? It seems like it should never have been this way, are you just fixing it while you're at it?

AdrianGaudebert:
r+

webapp-django/crashstats/settings/__init__.py
peterbe: It's a trick to make flake8 shut up.

webapp-django/crashstats/settings/test.py
peterbe: `FORCE_DB` was an invention of test-utils. `django-nose` "copied" that feature but it's called `REUSE_DB` there instead. 

That notion doesn't matter so much for socorro since we do our testing with the sqlite in-ram solution.

PR #3045, opened 2015-10-20
Fixes bug 1212919 - List of fields is now passed to SuperSearch by th…

AdrianGaudebert:
…e django models.

@peterbe r?

### DIFF
--- a/socorro/external/es/supersearch.py
+++ b/socorro/external/es/supersearch.py
@@ -9,6 +9,7 @@ from elasticsearch.exceptions import NotFoundError
 
 from socorro.external import (
     BadArgumentError,
+    MissingArgumentError,
 )
 from socorro.external.es.super_search_fields import SuperSearchFields
 from socorro.lib import datetimeutil
@@ -26,18 +27,14 @@ class SuperSearch(SearchBase):
             self.config.elasticsearch
         )
 
-        self.all_fields = SuperSearchFields(config=self.config).get_fields()
+        super(SuperSearch, self).__init__(*args, **kwargs)
 
+    def build_fields(self):
         # Create a map to associate a field's name in the database to its
         # exposed name (in the results and facets).
         self.database_name_to_field_name_map = dict(
             (x['in_database_name'], x['name'])
             for x in self.all_fields.values()
-        )
-
-        kwargs.update(fields=self.all_fields)
-        super(SuperSearch, self).__init__(
-            *args, **kwargs
         )
 
     def get_connection(self):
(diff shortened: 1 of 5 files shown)


socorro/external/es/supersearch.py
```diff
@@ -26,8 +27,9 @@ def __init__(self, *args, **kwargs):
             self.config.elasticsearch
         )
 
-        self.all_fields = SuperSearchFields(config=self.config).get_fields()
+        super(SuperSearch, self).__init__(*args, **kwargs)
```
peterbe: Awesomenesssauce!

socorro/external/es/supersearch.py
```diff
@@ -195,6 +192,12 @@ def get(self, **kwargs):
         The list of accepted parameters (with types and default values) is in
         the database and can be accessed with the super_search_fields service.
         """
+        # Require that the list of fields be passed.
+        if not kwargs.get('_fields'):
+            raise MissingArgumentError('_fields')
+        self.all_fields = kwargs['_fields']
```
peterbe: great!

socorro/external/es/supersearch.py
```diff
@@ -195,6 +192,12 @@ def get(self, **kwargs):
...
+        self.build_fields()
```
peterbe: You might want to consider calling this `self._build_fields()` because it's an implementation detail of this class that is boring to outside implementers of the class.

socorro/lib/search_common.py
```diff
@@ -140,6 +136,11 @@ def build_filters(self, fields):
     def get_parameters(self, **kwargs):
         parameters = {}
 
+        fields = kwargs.get('_fields')
```
peterbe: `fields = kwargs['_fields']` because it's not an option that this falls back on `None`. 
Keep the `assert` on the next line though.

socorro/unittest/external/es/test_supersearch.py
```diff
@@ -20,6 +21,12 @@
 # logging.getLogger('requests').setLevel(logging.ERROR)
 
 
+class SuperSearchWithFields(SuperSearch):
```
peterbe: Smart!

peterbe:
But where is the django caching of the fields? Now they have to be computed every time for every supersearch request, no?

AdrianGaudebert:
The webapp does caching in the models, and since we query via those models, I *think* we already have caching enabled. See here: https://github.com/mozilla/socorro/blob/master/webapp-django/crashstats/supersearch/models.py#L211

Are my assumptions wrong?

peterbe:
Caching works! r+

AdrianGaudebert:
https://github.com/mozilla/socorro/commit/aea9ad481c5b2d30c7b55ea9f64271af9d2a73e6

PR #3046, opened 2015-10-20
fixes Bug 1212917: cleanup of the iterator refactoring

twobraids:
enhancements to the previous PR under the same name.

### DIFF
--- a/socorro/app/fetch_transform_save_app.py
+++ b/socorro/app/fetch_transform_save_app.py
@@ -86,15 +86,6 @@ class FetchTransformSaveApp(App):
         default='forever'
     )
 
-    ###########################################################################
-    ### TODO: add a feature where clients of this class may register a waiting
-    ### function.  The MainThread will run all the registered waiting
-    ### functions at their configured interval.  A first application of this
-    ### feature will be to allow periodic reloading of config data from a
-    ### database.  Specifically, the skip list rules could be reloaded without
-    ### having to restart the processor.
-    ###########################################################################
-
     #--------------------------------------------------------------------------
     @staticmethod
     def get_application_defaults():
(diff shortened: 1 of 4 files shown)


socorro/app/fetch_transform_save_app.py
```diff
@@ -234,10 +225,18 @@ def _limited_iterator(self):
             for crash_id in self._basic_iterator():
                 if self._filter_disallowed_values(crash_id):
                     continue
+                if crash_id is None:
+                    # it's ok to yield None, however, we don't want it to
+                    # be counted as a yielded value
```
peterbe: excellent comment!

socorro/app/fetch_transform_save_app.py
```diff
@@ -357,9 +356,13 @@ def _setup_task_manager(self):
...
         queue that is the heart of the processor."""
         self.config.logger.info('installing signal handers')
+        # set up the signal handler for dealing with SIGTERM. the target should
+        # be this app instance so the signal handler can reach in and set the
+        # quit flag to be True.  See the 'respond_to_SIGTERM' method for the
+        # more information
         respond_to_SIGTERM_with_logging = partial(
             respond_to_SIGTERM,
-            logger=self.config.logger
+            target=self
```
peterbe: uh? Does the `respond_to_SIGTERM` no longer accept the `logger` option?

peterbe: s/option/argument

peterbe: ignore. I see now.

socorro/lib/task_manager.py
```diff
@@ -39,10 +39,18 @@ def respond_to_SIGTERM(signal_number, frame, logger=None):
...
-    if logger:
-        logger.info('detected SIGTERM')
-    raise KeyboardInterrupt
+    if target:
+        target.config.logger.info('detected SIGTERM')
+        # by setting the quit flag to true, any calls to the 'quit_check'
+        # method that is so liberally passed around in this framework will
+        # result in raising the quit exception.  The current quit exception
+        # is KeyboardInterrupt
+        target.task_manager.quit = True
```
peterbe: nifty

peterbe: Perhaps it doesn't matter but I don't like the idea of sending in an object and having it mutate "over there". 
I'd rather send in the object and have a function on it called. That feels more functional (tm). Perhaps I'm just colored by doing a lot of Javascript.

peterbe: If it was `target.task_manager.quit()` instead, it'd make more sense to put the `self.config.logger.info('detected SIGTERM quit')` in there instead.

peterbe:
r+ with thoughts

PR #3047, opened 2015-10-20
fixes Bug 1216524: added CollectorApp to config instructions

twobraids:
he docs on how to setup Socorro are outdated in the section about setting up the Collector.  The "application" parameter is now required to select which Collector app to use.  We have two of them now:

``` python
# this is the general use case
socorro.collector.collector_app.CollectorApp
```

```python
# this is the Mozilla variant that allows the collector to
# have several services for collecting different types of 
# crashes at different endpoints.
socorro.collector.collector_app.Collector2015App
```

### DIFF
--- a/docs/configuring-socorro.rst
+++ b/docs/configuring-socorro.rst
@@ -7,7 +7,7 @@ Storing configuration in Consul
 -------------------------------
 
 Socorro apps get their configuration from environment variables. We suggest
-using Consul to hold configuration: 
+using Consul to hold configuration:
 https://consul.io/intro/getting-started/install.html
 
 Consul must be running in order for Socorro apps to start up and access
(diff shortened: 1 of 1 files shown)


docs/configuring-socorro.rst
```diff
@@ -7,7 +7,7 @@ Storing configuration in Consul
 -------------------------------
 
 Socorro apps get their configuration from environment variables. We suggest
-using Consul to hold configuration:
```
twobraids: my editor removes trailing whitespace from lines.  sorry about that.

docs/configuring-socorro.rst
```diff
@@ -36,11 +36,13 @@ Below is the minimum viable configuration to get collection working on a
 single node:
 
 .. code-block:: bash
```
twobraids: the following is actually the change of interest...

docs/configuring-socorro.rst
peterbe: Never be sorry for having a correctly configured editor. :)

peterbe:
r+

PR #3048, opened 2015-10-20
fixes bug1216742:  [no merge - work in progress] move FTS transforms into their own hierarchy 	

twobraids:
Each of the FTS has had to define its own transformation method.  This is the method that is run by the workers to transform a raw_crash into a processed crash.  At least, that's what the transform method does in the Processor.  In the crashmover, the transform method just moves raw crashes and dumps from one crashstore to another.  The submitter is very similar to the crashmover, but its transform copies rather than deletes.  The seldom used processed crash copier has a transform method that duplicates just the processed crashes from one storage system to another.  

Rather than have this mash up of each type of app defining its own transform and then duplicating code sections, move all these transforms into their own class hierarchy.  Any app can then just select a goal like "move raw crashes", "copy all crashes", or even copy/move things other than crashes.  

At the same time, disambiguate the word transform, because as it stands right now, it refers to two things: 
    1) the method run by the worker that includes all the steps of FTS.
    2) the T in FTS, how a crash is changed from one form to another.

From now on the former "transform methods" should be referred to as "FTS worker methods".  This leaves the word "transform" as referring only to the T of FTS: the act of 
converting one thing into another.

That gives the apps the ability to define the actual transform, the T, between the fetch and the save of FTS.  Separating these methods makes it simpler to mix and match the components.  It also then sets the stage for creating special purpose processors, copiers, movers, etc.  In a future step, we can get rid of the special apps like processor, submitter, mover as they can be composed from components and the base FTS app can run them.  

The processor consists of:
   1) the iteration system that consumes infinitely
   2) the FTS worker method that both fetches and saves raw_crashes, dumps & processed crashes
   3) a transformation method that applies the rule processing system.

The submitter consists of:
   1) a user selectable iteration system that can work either finitely or infinitely.
   2) an FTS worker method that only reads raw_crashes and dumps
   3) a transformation that doesn't change anything

The crashmover is just like the submitter, except in only allows the infinite iterator

This is opening the door to special purpose FTS apps such as a copier that can read from a PG query and write to the RabbitMQ reprocessing queue.  Another special purpose app came about when a release was made with the wrong channel.  We hand coded a processor that copied and fixed just the affected crashes.  How about a submitter that filters out PII before sending a crash on to a special purpose processor?  Sure thing!  This new system will allow that sort of change without having to add new code.

### DIFF
--- a/scripts/defaults
+++ b/scripts/defaults
@@ -22,9 +22,9 @@ export rmq_host=${rmq_host:-"localhost"}
 export rmq_user=${rmq_user:-"guest"}
 export rmq_password=${rmq_password:-"guest"}
 export rmq_virtual_host=${rmq_virtual_host:-"/"}
-export rmq_normal_queue_name=${rmq_normal_queue_name:-"socoro.integrationtest.normal"}
-export rmq_priority_queue_name=${rmq_priority_queue_name:-"socoro.integrationtest.priority"}
-export rmq_reprocessing_queue_name=${rmq_reprocessing_queue_name:-"socoro.integrationtest.reprocessing"}
+export rmq_normal_queue_name=${rmq_normal_queue_name:-"socorro.integrationtest.normal"}
+export rmq_priority_queue_name=${rmq_priority_queue_name:-"socorro.integrationtest.priority"}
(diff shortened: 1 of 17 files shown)


twobraids:
Dropping this stage of my refactoring effort as there is insufficient time to complete it.

PR #3050, opened 2015-10-21
Fixes bug 1216928 - Exposed SuperSearchFields in public API.

AdrianGaudebert:
@peterbe r?


peterbe:
r+

At the time I've not had a chance to play with it on my laptop but the code looks good. If you're confident it'll work, merge away.

PR #3052, opened 2015-10-21
fixes bug 1216807 - Get rid of funfactory

peterbe:
This might need some care when reviewing. I'm quite confident about the code changes, but what I'm less confident about is the django-compressor stuff. 

I was able to set
```
COMPRESS_ENABLED=False
COMPRESS_OFFLINE=False
```
and run the site fine. The `.less` files were converted to `.css` just fine. 

Also, I was able to change to 
```
COMPRESS_ENABLED=True
COMPRESS_OFFLINE=True
```
and run
```
./manage.py collectstatic --noinput && ./manage.py compress --engine=jinja2 --force
```
like I always do and then runserver and that worked fine too.

### DIFF
--- a/webapp-django/crashstats/settings/base.py
+++ b/webapp-django/crashstats/settings/base.py
@@ -1,29 +1,64 @@
 # This is your project's main settings file that can be committed to your
-# repo. If you need to override a setting locally, use .env or
-# straight up environment variables.
+# repo. If you need to override a setting locally, use .env or environment
+# variables.
 
-from funfactory.settings_base import *
+import os
+import logging
 
 import dj_database_url
 from decouple import config, Csv
 
+
+ROOT = os.path.abspath(
+    os.path.join(
+        os.path.dirname(__file__),
+        '..',
+        '..'
+    ))
+
+
+def path(*dirs):
+    return os.path.join(ROOT, *dirs)
+
+
+SITE_ID = 1
+
+LANGUAGE_CODE = 'en-US'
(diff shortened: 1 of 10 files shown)


peterbe:
Actually, I need to review how the wsgi file works. I see some import of `manage` there which I don't think we should need. I would love to see this resolved first though https://bugzilla.mozilla.org/show_bug.cgi?id=1217184#c1

peterbe:
@AdrianGaudebert merge conflict fixed.

webapp-django/crashstats/settings/base.py
```diff
@@ -543,6 +609,18 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
...
+COMPRESS_CSS_FILTERS = (
+    'compressor.filters.css_default.CssAbsoluteFilter',
+    'compressor.filters.cssmin.CSSMinFilter'
+)
+
+STATICFILES_FINDERS = (
+    'django.contrib.staticfiles.finders.FileSystemFinder',
+    'django.contrib.staticfiles.finders.AppDirectoriesFinder',
+    'compressor.finders.CompressorFinder',
+)
```
AdrianGaudebert: These are set twice in this file (the first occurrence being around line 100).

AdrianGaudebert:
This is looking good. I've been browsing around with these changes and saw no problem. Hopefully stage will accept those changes gladly as well... :)

r+

PR #3053, opened 2015-10-21
fixes Bug 959724 - the set_default problem that overrode underlying config defaults

twobraids:
some derived classes change the underlying configuration defaults in a manner that changes the value for all super classes too.  Make the scope of the change local to the defining class only.

### DIFF
--- a/socorro/processor/mozilla_processor_2015.py
+++ b/socorro/processor/mozilla_processor_2015.py
@@ -1,6 +1,8 @@
-from ujson import dumps
+import ujson
+from configman import Namespace
 
 from socorro.processor.processor_2015 import Processor2015
+from socorro.lib.converters import change_default
 
 #------------------------------------------------------------------------------
 # these are the steps that define processing a crash at Mozilla.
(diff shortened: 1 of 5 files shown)


twobraids:
@peterebe r?

peterbe:
I think that when you set the assignee properly I get a nice email. No need to type the r? comment. 
I'll take heed of this myself for my PRs from now on.

socorro/processor/mozilla_processor_2015.py
```diff
@@ -97,8 +99,9 @@
...
 
-    Processor2015.required_config.rule_sets.set_default(
-        dumps(mozilla_processor_rule_sets),
-        force=True
+    required_config = Namespace()
+    required_config.rule_sets = change_default(
+        Processor2015,
+        'rule_sets',
+        ujson.dumps(mozilla_processor_rule_sets)
     )
```
peterbe: So much better!

socorro/processor/mozilla_processor_2015.py
```diff
@@ -1,6 +1,8 @@
-from ujson import dumps
+import ujson
```
peterbe: I can't put my finger on it but something tells me we should avoid optimization tricks like this. 
The regular standard python json module is good enough 99% of the cases. We use ujson in some parts where shitloads of JSON encoding and decoding is going on. 
So although I don't have a direct concrete example where it breaks; my instincts are at least asking me to raise a concern.

peterbe:
r+ but you might want to at least stop and think about the use of `ujson`.

socorro/processor/mozilla_processor_2015.py
twobraids: when Rob first introduced ujson into the project, we discovered incompatibilities since it doesn't support ```object_hooks```.  I recall the he and I agreed that we would use ujson where it was compatible and only use the native json when we absolutely had to.

PR #3054, opened 2015-10-22
fix missing new line in PR#3047

micxer:
The bash block didn't show up at all in the docs.


AdrianGaudebert:
Yay! Thanks for contributing! :)

PR #3055, opened 2015-10-22
Add missing configuration line that prevents the processor from processing

micxer:
Following the documentation I was missing this in the `processor.conf`. The processor did run but wasn't picking up the crash reports collected by the collector.


twobraids:
r+

PR #3058, opened 2015-10-23
fixes bug 1217917 - Change default FTP scraping URL

peterbe:
cc @KaiRo-at


twobraids:
the change is certainly simple enough.  the test failures are also pretty straight forward.  so once those are resolved, r+

PR #3060, opened 2015-10-24
Fixes Bug 713973, Bug 1042420, Bug 1044183, Bug 913581, Bug 999923, deprecate & remove all the things

twobraids:
This is the replacement for the previous ROCK-THE-SOCORRO-WORLD PR.  The last one failed after a rebase with master inexplicably overwrote the ```.../scripts/build.sh``` with an older version and caused the build to fail.

This version has actually had every file scrutinized.  The only file that seems to have had the effect was the aforementioned ```.../scripts/build.sh```.

I look forward to merging this PR back.


jdotpz:
The deployment bits look good here, arrrr+

PR #3061, opened 2015-10-25
Fixes Bug 1218220 - split the boto connection from the crashstorage class

twobraids:
we're gonna have to store correlation data in S3, it sure would be handy to have a nicely wrapped Boto connection class to use for this.  We've got one, but it is embedded within the BotoS3CrashStorage class. 

Split it out into a TransactionExecutor compatible ConnectionContext class.  That'll give us some nice code reuse.  At the same time, this action should in no way have any repercussions on existing uses the BotoS3CrashStorage class.  In other words, this action should not affect configuration of any existing App.

### DIFF
--- a/socorro/external/boto/connection_context.py
+++ b/socorro/external/boto/connection_context.py
@@ -1,0 +1,245 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import boto
+import boto.s3.connection
+import boto.exception
+
+import json
+import os
+import socket
+import datetime
+import contextlib
+
+from configman import Namespace, RequiredConfig, class_converter
+from configman.converters import class_converter, py_obj_to_str
+from socorro.lib.util import DotDict
+from socorro.lib.converters import change_default
+
+class BotoNotFound(Exception):
+    pass
+
+
+#==============================================================================
+class ConnectionContext(RequiredConfig):
+
+    required_config = Namespace()
+    required_config.add_option(
+        'host',
+        doc="The hostname (leave empty for AWS)",
+        default="",
+        reference_value_from='resource.boto',
+    )
+    required_config.add_option(
+        'port',
+        doc="The network port (leave at 0 for AWS)",
+        default=0,
+        reference_value_from='resource.boto',
+    )
+    required_config.add_option(
+        'access_key',
+        doc="access key",
+        default="",
+        reference_value_from='resource.boto',
+    )
+    required_config.add_option(
+        'secret_access_key',
+        doc="secret access key",
+        default="",
+        secret=True,
+        reference_value_from='secrets.boto',
+    )
+    required_config.add_option(
+        'bucket_name',
+        doc="The name of the bucket.",
+        default='crashstats',
+        reference_value_from='resource.boto',
+        likely_to_be_changed=True,
+    )
+    required_config.add_option(
+        'prefix',
+        doc="a prefix to use inside the bucket",
+        default='',
+        reference_value_from='resource.boto',
+        likely_to_be_changed=True,
+    )
+    required_config.add_option(
+        'calling_format',
+        doc="fully qualified python path to the boto calling format function",
+        default='boto.s3.connection.SubdomainCallingFormat',
+        from_string_converter=class_converter,
+        reference_value_from='resource.boto',
+        likely_to_be_changed=True,
+    )
+
+    operational_exceptions = (
+        socket.timeout,
+        # wild guesses at retriable exceptions
+        boto.exception.PleaseRetryException,
+        boto.exception.ResumableTransferDisposition,
+        boto.exception.ResumableUploadException,
+    )
+
+    conditional_exceptions = (
+        boto.exception.StorageResponseError
+    )
+
+    #--------------------------------------------------------------------------
+    def is_operational_exception(self, x):
+        if "not found, no value returned" in str(x):
+            # the not found error needs to be re-tryable to compensate for
+            # eventual consistency.  However, a method capable of raising this
+            # exception should never be used with a transaction executor that
+            # has infinite back off.
+            return True
+        #elif   # for further cases...
+        return False
+
+
+    #--------------------------------------------------------------------------
+    def __init__(self, config, quit_check_callback=None):
+        self.config =  config
+
+        self._connect_to_endpoint = boto.connect_s3
+        self._calling_format = config.calling_format
+        self._CreateError = boto.exception.S3CreateError
+        self._S3ResponseError = boto.exception.S3ResponseError
+        self._open = open
+
+
+    #--------------------------------------------------------------------------
+    @staticmethod
+    def build_s3_dirs(prefix, name_of_thing, id):
+        """
+        Use S3 pseudo-directories to make it easier to list/expire later
+        {{prefix}}/{{version}}/{{name_of_thing}}/{{id}}
+        """
+        version = 'v1'
+        return '%s/%s/%s/%s' % (prefix, version, name_of_thing, id)
+
+    #--------------------------------------------------------------------------
+    def _get_bucket(self, conn, bucket_name):
+        try:
+            return self._bucket_cache
+        except AttributeError:
+            self._bucket_cache = conn.get_bucket(bucket_name)
+            return self._bucket_cache
+
+    #--------------------------------------------------------------------------
+    def _get_or_create_bucket(self, conn, bucket_name):
+        try:
+            return self._bucket_cache
+        except AttributeError:
+            try:
+                self._bucket_cache = conn.get_bucket(bucket_name)
+            except self._S3ResponseError:
+                self._bucket_cache = conn.create_bucket(bucket_name)
+            return self._bucket_cache
+
+    #--------------------------------------------------------------------------
+    def submit_to_boto_s3(self, id, name_of_thing, thing):
+        """submit something to boto.
+        """
+        if not isinstance(thing, basestring):
+            raise Exception('can only submit strings to boto')
+
+        conn = self._connect()
+        bucket = self._get_or_create_bucket(conn, self.config.bucket_name)
+
+        key = self.build_s3_dirs(self.config.prefix, name_of_thing, id)
+
+        storage_key = bucket.new_key(key)
+        storage_key.set_contents_from_string(thing)
+
+    #--------------------------------------------------------------------------
+    def fetch_from_boto_s3(self, id, name_of_thing):
+        """retrieve something from boto.
+        """
+        conn = self._connect()
+        bucket = self._get_bucket(conn, self.config.bucket_name)
+
+        key = self.build_s3_dirs(self.config.prefix, name_of_thing, id)
+
+        storage_key = bucket.get_key(key)
+        if storage_key is None:
+            raise BotoNotFound('%s not found, no value returned' % id)
+        return storage_key.get_contents_as_string()
+
+    #--------------------------------------------------------------------------
+    def _connect(self):
+        try:
+            return self.connection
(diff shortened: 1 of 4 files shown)


socorro/external/boto/connection_context.py
```diff
@@ -0,0 +1,245 @@
...
+    required_config.add_option(
+        'port',
+        doc="The network port (leave at 0 for AWS)",
+        default=0,
+        reference_value_from='resource.boto',
+    )
+    required_config.add_option(
+        'access_key',
+        doc="access key",
+        default="",
```
peterbe: This should probably also have `secret=True`

socorro/external/boto/connection_context.py
```diff
@@ -0,0 +1,245 @@
...
+        reference_value_from='resource.boto',
+    )
+    required_config.add_option(
+        'secret_access_key',
+        doc="secret access key",
+        default="",
+        secret=True,
+        reference_value_from='secrets.boto',
```
peterbe: Does this and `access_key` not need the `likely_to_be_changed=True`??

socorro/external/boto/connection_context.py
```diff
@@ -0,0 +1,245 @@
...
+        reference_value_from='resource.boto',
+        likely_to_be_changed=True,
+    )
+    required_config.add_option(
+        'calling_format',
+        doc="fully qualified python path to the boto calling format function",
+        default='boto.s3.connection.SubdomainCallingFormat',
+        from_string_converter=class_converter,
+        reference_value_from='resource.boto',
+        likely_to_be_changed=True,
```
peterbe: Why is that?

socorro/external/boto/connection_context.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+    #--------------------------------------------------------------------------
+    @staticmethod
+    def build_s3_dirs(prefix, name_of_thing, id):
+        """
+        Use S3 pseudo-directories to make it easier to list/expire later
+        {{prefix}}/{{version}}/{{name_of_thing}}/{{id}}
+        """
+        version = 'v1'
```
peterbe: you like sustainable and flexible code, so maybe move this to a keyword argument to the function

socorro/external/boto/connection_context.py
twobraids: because we're specifying a class that is specific to S3 that wouldn't be used if we were connecting to Ceph. so this is an option that is likely to change for other implementations.

socorro/external/boto/connection_context.py
```diff
@@ -0,0 +1,245 @@
...
+        self._connect_to_endpoint = boto.connect_s3
+        self._calling_format = config.calling_format
+        self._CreateError = boto.exception.S3CreateError
+        self._S3ResponseError = boto.exception.S3ResponseError
+        self._open = open
+
+
+    #--------------------------------------------------------------------------
+    @staticmethod
+    def build_s3_dirs(prefix, name_of_thing, id):
```
peterbe: In S3 lingo a file name is called a "key". The key, unlike regular file systems can have the "/" character in it. So that this does is it builds a key, not "s3 dirs".

socorro/external/boto/connection_context.py
twobraids: good point

socorro/external/boto/connection_context.py
twobraids: we've not treated this as secret in the past.  To do so now, would require a change to configuration across the system where anything touches S3.

socorro/external/boto/connection_context.py
```diff
@@ -0,0 +1,245 @@
...
+
+    #--------------------------------------------------------------------------
+    def __init__(self, config, quit_check_callback=None):
+        self.config =  config
+
+        self._connect_to_endpoint = boto.connect_s3
+        self._calling_format = config.calling_format
+        self._CreateError = boto.exception.S3CreateError
+        self._S3ResponseError = boto.exception.S3ResponseError
+        self._open = open
```
peterbe: What's the purpose of this?

socorro/external/boto/connection_context.py
```diff
@@ -0,0 +1,245 @@
...
+            return True
+        #elif   # for further cases...
+        return False
+
+
+    #--------------------------------------------------------------------------
+    def __init__(self, config, quit_check_callback=None):
+        self.config =  config
+
+        self._connect_to_endpoint = boto.connect_s3
```
peterbe: The class is all about `boto`. Not about `S3` yet this is hardcoded as the default.

socorro/external/boto/connection_context.py
```diff
@@ -0,0 +1,245 @@
...
+        self._calling_format = config.calling_format
+        self._CreateError = boto.exception.S3CreateError
+        self._S3ResponseError = boto.exception.S3ResponseError
```
peterbe: Another S3 specific thing

peterbe: s/thing/things

socorro/external/boto/connection_context.py
```diff
@@ -0,0 +1,245 @@
...
+        version = 'v1'
+        return '%s/%s/%s/%s' % (prefix, version, name_of_thing, id)
+
+    #--------------------------------------------------------------------------
+    def _get_bucket(self, conn, bucket_name):
+        try:
+            return self._bucket_cache
+        except AttributeError:
+            self._bucket_cache = conn.get_bucket(bucket_name)
+            return self._bucket_cache
```
peterbe:
```
bucket1 = thing._get_bucket(myconnection, 'stuff')  # cache miss
bucket2 = thing._get_bucket(myconnection, 'stuff')  # cache hit
bucketX = thing._get_bucket(myconnection, 'other')  # cache hit!!

```

socorro/external/boto/connection_context.py
```diff
@@ -0,0 +1,245 @@
...
+    #--------------------------------------------------------------------------
+    def _get_or_create_bucket(self, conn, bucket_name):
+        try:
+            return self._bucket_cache
+        except AttributeError:
+            try:
+                self._bucket_cache = conn.get_bucket(bucket_name)
+            except self._S3ResponseError:
+                self._bucket_cache = conn.create_bucket(bucket_name)
+            return self._bucket_cache
```
peterbe: This is DRY. I don't like the idea of the private cache attribute (`self._bucket_cache`) is mentioned in two places. Much better with:
```python
def _get_or_create_bucket(self, conn, bucket_name):
    try:
        return self._get_bucket(conn, bucket_name)
    except self._S3ResponseError:
         conn.create_bucket(bucket_name)
        return self._get_bucket(conn, bucket_name)
```

socorro/external/boto/connection_context.py
```diff
@@ -0,0 +1,245 @@
...
+
+    #--------------------------------------------------------------------------
+    def submit_to_boto_s3(self, id, name_of_thing, thing):
+        """submit something to boto.
+        """
+        if not isinstance(thing, basestring):
+            raise Exception('can only submit strings to boto')
```
peterbe: To avoid the blank exception class and untested code, just use:
```
# can only submit strings to boto
assert isinstance(thing, basestring), type(thing)
```

socorro/external/boto/connection_context.py
```diff
@@ -0,0 +1,245 @@
...
+        conn = self._connect()
+        bucket = self._get_or_create_bucket(conn, self.config.bucket_name)
+
+        key = self.build_s3_dirs(self.config.prefix, name_of_thing, id)
+
+        storage_key = bucket.new_key(key)
+        storage_key.set_contents_from_string(thing)
+
+    #--------------------------------------------------------------------------
+    def fetch_from_boto_s3(self, id, name_of_thing):
```
peterbe: `fetch_from_boto`?

peterbe: Or `get_from_boto` considering that it's always supposed to return the content or raise an exception. To be "fetch" is an activity you can do in the background or something where you don't need the reading.

socorro/external/boto/connection_context.py
```diff
@@ -0,0 +1,245 @@
...
+            if self.config.host:
+                kwargs["host"] = self.config.host
+            if self.config.port:
+                kwargs["port"] = self.config.port
+            self.connection = self._connect_to_endpoint(**kwargs)
+            return self.connection
+
+    #--------------------------------------------------------------------------
+    def _convert_mapping_to_string(self, a_mapping):
+        self._stringify_dates_in_dict(a_mapping)
```
peterbe: Let's do that properly instead with passing an encoder class to `json.dumps`. That way you get it recursive and not just taking care of `datetime.datetime` objects on the "first" level. 

Also, why not isoformat()?

socorro/external/boto/crashstorage.py
```diff
@@ -16,6 +16,10 @@
     CrashIDNotFound,
     MemoryDumpsMapping,
 )
+from socorro.external.boto.connection_context import (
+    ConnectionContext,
+    BotoNotFound
```
peterbe: I kinda wish this was called `KeyNotFound`.

socorro/external/boto/crashstorage.py
```diff
@@ -31,6 +35,13 @@ class BotoS3CrashStorage(CrashStorageBase):
 
     required_config = Namespace()
     required_config.add_option(
+        "resource_class",
+        default=ConnectionContext,
```
peterbe: Why does this get to be written like this when other options have to have their name written out as a string?

socorro/external/boto/crashstorage.py
```diff
@@ -31,6 +35,13 @@ class BotoS3CrashStorage(CrashStorageBase):
...
+        doc="fully qualified dotted Python classname to handle Boto connections",
```
peterbe: actually, this doc string doesn't match the default. Hence my confusion.

socorro/external/boto/crashstorage.py
```diff
@@ -154,45 +106,28 @@ def __init__(self, config, quit_check_callback=None):
 
         self.transaction_for_get = config.transaction_executor_class_for_get(
             config,
-            self,  # we are our own connection
+            self.connection_source,  # we are our own connection
```
peterbe: comment doesn't make sense any more.

socorro/unittest/external/boto/test_connection_context.py
```diff
@@ -0,0 +1,206 @@
...
+from socorro.database.transaction_executor import (
+    TransactionExecutor,
+    TransactionExecutorWithLimitedBackoff,
+)
+from socorro.external.crashstorage_base import CrashIDNotFound
+import socorro.unittest.testbase
+
+
+a_raw_crash = {
+    "submitted_timestamp": "2013-01-09T22:21:18.646733+00:00"
```
peterbe: See, here you're using the isoformat

socorro/unittest/external/boto/test_connection_context.py
```diff
@@ -0,0 +1,206 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import mock
+import json
+import tempfile
+import shutil
+from os.path import join
+
+import boto.exception
+
```
peterbe: If your intention is to "linebreak bucketize" the imports by built-ins and third-party and socorro etc. then the `import mock` should be together with `import boto.exception`

peterbe:
r+ with nits. Especially the use of json.dumps with out a cls encoder.

peterbe:
Counter PR https://github.com/twobraids/socorro/pull/31

twobraids:
this version of this PR has taken it to the logical conclusion of a generic Boto connection and crash storage classes that be used for connection to services other than S3.  Then there are S3 connection & crashstorage classes.  This tightens up the encapsulations and isolates all the S3-isms into the classes.

peterbe:
laaaaand iiit

PR #3062, opened 2015-10-26
Fixes Bug 1218366  fixed missing import in the SocorroLiteProcessor class + a test

twobraids:
We don't use the ScorroLiteProcessor at Mozilla, but our docs lead new users to start with it.  We ought to ensure that it works.  

Fix the missing symbol in the SocorroLiteProcessor and add it to the integration tests to make sure it works.

### DIFF
--- a/socorro/processor/socorrolite_processor_2015.py
+++ b/socorro/processor/socorrolite_processor_2015.py
@@ -1,6 +1,9 @@
 import ujson
+
 from socorro.processor.processor_2015 import Processor2015
 from socorro.lib.converters import change_default
+
+from configman import Namespace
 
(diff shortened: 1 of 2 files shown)


socorro/processor/socorrolite_processor_2015.py
```diff
@@ -1,7 +1,10 @@
 import ujson
+
 from socorro.processor.processor_2015 import Processor2015
 from socorro.lib.converters import change_default
 
+from configman import Namespace
```
peterbe: we need to pyflakes audit our code clearly.

peterbe:
Seems to work fine. r+

PR #3063, opened 2015-10-26
fixes bug 1217828 - make ftpscraper work on https://archive.mozilla.org

peterbe:
Note! This PR does NOT yet have new tests. The reason I'm prematurely making it a PR is to optimize our day. Perhaps you can get started with what we have so far whilst I re-write the unit tests. 

Again, sorry for baking in a bunch of pep8 fixes into the mix. I did that to get momentum on this work last week as a warm-up.

### DIFF
--- a/socorro/cron/jobs/ftpscraper.py
+++ b/socorro/cron/jobs/ftpscraper.py
@@ -1,16 +1,25 @@
+import datetime
+import sys
 import re
 import urllib2
-import lxml.html
+import os
 import json
 import time
-from configman import Namespace
+import urlparse
+
+import mock
+import lxml.html
+
+from configman import Namespace, class_converter
 from crontabber.base import BaseCronApp
 from crontabber.mixins import (
     as_backfill_cron_app,
     with_postgres_transactions
 )
+from socorro.app.socorro_app import App, main
+from socorro.lib.datetimeutil import string_to_datetime
 from socorro.lib import buildutil
-import os
+
 
 """
  Socket timeout to prevent FTP from hanging indefinitely
@@ -20,9 +29,8 @@ import os
 import socket
 socket.setdefaulttimeout(60)
 
+
 #==============================================================================
-
-
 class RetriedError(IOError):
 
     def __init__(self, attempts, url):
@@ -34,17 +42,6 @@ class RetriedError(IOError):
             '<%s: %s attempts at downloading %s>' %
             (self.__class__.__name__, self.attempts, self.url)
         )
-
-
-def urljoin(*parts):
-    url = parts[0]
-    for part in parts[1:]:
-        if not url.endswith('/'):
-            url += '/'
-        if part.startswith('/'):
-            part = part[1:]
-        url += part
-    return url
 
 
 def patient_urlopen(url, max_attempts=4, sleep_time=20):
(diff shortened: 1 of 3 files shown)


socorro/cron/jobs/ftpscraper.py
```diff
@@ -71,25 +80,46 @@ def patient_urlopen(url, max_attempts=4, sleep_time=20):
...
+                    link = link.replace(path, '')
                 if link.startswith(startswith):
-                    results.append(link)
+                    return True
             elif endswith:
                 if link.endswith(endswith):
-                    results.append(link)
+                    return True
+            else:
+                raise NotImplementedError(
```
twobraids: to simplify the logic in this section, this case can be detected at the very beginning of the ```get_links``` method.  Move it up to the top to allow us to bail out early.  It's no use fetching the url if we can't proceed anyway.

twobraids: on doing that suggestion, the code could further simplified from line 105 onward to just:
``` python
                return link.startswith(startswith)
            return link.endswith(endswith)
```
Naming things is something that we often disagree on, but having the variable names match the function names is unnecessary visual overloading.  Please consider changing the variables to something like ```prefix``` & ```suffix``` or ```start_string``` & ```end_string```

peterbe: What about `starts_with` and `ends_with`? Sounds nice to say. 

I'll refactor once I have some tests.

socorro/cron/jobs/ftpscraper.py
```diff
@@ -71,25 +80,46 @@ def patient_urlopen(url, max_attempts=4, sleep_time=20):
...
+        def url_match(link):
+            # The link might be something like "/pub/mobile/nightly/"
+            # but we're looking for a path that starts with "nightly".
+            # So first we need to remove what's part of the base URL
+            # to make a fair comparison.
+            if startswith is not None:
+                # If the current URL is http://example.com/some/dir/
+                # and the link is /some/dir/mypage/ and the thing
+                # we're looking for is "myp" then this should be true
+                path = urlparse.urlparse(url).path
```
twobraids: the ```url_match ``` method is called inside a loop, but the result of this line is invariant through the life of the loop.  Take it out of the method to both avoid unnecessary repetition and make this function look simpler.

socorro/cron/jobs/ftpscraper.py
```diff
@@ -71,25 +80,46 @@ def patient_urlopen(url, max_attempts=4, sleep_time=20):
 
 class ScrapersMixin(object):
 
-    def getLinks(self, url, startswith=None, endswith=None):
+    def get_links(self, url, startswith=None, endswith=None):
 
         html = ''
```
twobraids: drop it.  You just change the value on line 91.

socorro/cron/jobs/ftpscraper.py
```diff
@@ -71,25 +80,46 @@ def patient_urlopen(url, max_attempts=4, sleep_time=20):
...
+            # but we're looking for a path that starts with "nightly".
+            # So first we need to remove what's part of the base URL
+            # to make a fair comparison.
+            if startswith is not None:
+                # If the current URL is http://example.com/some/dir/
+                # and the link is /some/dir/mypage/ and the thing
+                # we're looking for is "myp" then this should be true
+                path = urlparse.urlparse(url).path
+                if link.startswith(path):
+                    link = link.replace(path, '')
```
twobraids: this substitution doesn't persist outside the method and I think that results in an incorrect final result.

Let's say we use your example:
```
url is "http://example.com/some/dir/"
path is then "/some/dir/"
the link is "/some/dir/mypage/"
start_with is "myp"
so we get a match and True is returned.
then down in your loop, the link "/some/dir/mypage" is accepted
and gets married with the url to form:
"http://example.com/some/dir/some/dir/mypage"
```
I don't think that's correct

twobraids: nope, my interpretation is not correct.  more in a minute...

twobraids: disregard, i'm insane

twobraids:
r+ from me

PR #3064, opened 2015-10-26
fixes Bug 1218579  -removed logger reference at the end that might not exist

twobraids:
there was a debug line in the socorro app hierarchy that was inadvertently left in.  in some edge cases it causes an exception during a shut down.  fix it.


peterbe:
r+ if travis approves

PR #3066, opened 2015-10-28
Fixes Bug 1219265 - iterator for products & versions for correlation reports + consolidation

twobraids:
The correlation reports feed on data about specific Product/Versions.  In their original form, they read from a tar file of all the day's processed crashes.  Since in their new form they are in an FTS app, they can use a crash_iter that is more targeted.

Create a new_crash_source module for Postgres that accepts a list of Products/Versions in this form:  "Firefox, 41.0.1, 42.0, 43.0a3; Thunderbird, 38.1.0, 38.2.0; ...".  From that, create a query that will target just those crashes and then offer the stream of crash_ids as an iterator for an FTS app.  

There are at least two other new_crash_source iterators floating around in the Socorro source code.  Consolidate all of these into one module called socorro.external.postgresql.new_crash_source

### DIFF
--- a/socorro/external/postgresql/new_crash_source.py
+++ b/socorro/external/postgresql/new_crash_source.py
@@ -1,0 +1,158 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from datetime import timedelta
+from collections import Sequence
+
+from configman import Namespace, RequiredConfig, class_converter
+
+from socorro.lib.converters import (
+    change_default,
+)
+from socorro.lib.datetimeutil import (
+    utc_now,
+    string_to_datetime,
+)
+from socorro.external.postgresql.dbapi2_util import execute_query_fetchall
+
+
+#==============================================================================
+class PGQueryNewCrashSource(RequiredConfig):
+    """This class is an iterator that will yield a stream of crash_ids based
+    on a query to the PG database."""
+    required_config = Namespace()
+    required_config.add_option(
+        'transaction_executor_class',
+        default="socorro.database.transaction_executor."
+        "TransactionExecutorWithInfiniteBackoff",
+        doc='a class that will manage transactions',
+        from_string_converter=class_converter,
+        reference_value_from='resource.postgresql',
+    )
+    required_config.add_option(
+        'database_class',
+        default='socorro.external.postgresql.connection_context'
+            '.ConnectionContext',
+        doc='the class responsible for connecting to Postgres',
+        from_string_converter=class_converter,
+        reference_value_from='resource.postgresql',
+    )
+    required_config.add_option(
+        'crash_id_query',
+        doc='sql to get a list of crash_ids',
+        default="select 'some_id'",
+        likely_to_be_changed=True,
+    )
+
+    #--------------------------------------------------------------------------
+    def __init__(self, config, name, quit_check_callback=None):
+
+        self.database = config.database_class(
+            config
+        )
+        self.transaction = config.transaction_executor_class(
+            config,
+            self.database,
+            quit_check_callback=quit_check_callback
+        )
+        self.config = config
+        self.name = name
+        self.data = ()
+        self.crash_id_query = config.crash_id_query
+
+    #--------------------------------------------------------------------------
+    def __iter__(self):
+        crash_ids = self.transaction(
+            execute_query_fetchall,
+            self.crash_id_query,
+            self.data
+        )
+
+        for a_crash_id in crash_ids:
+            yield a_crash_id
+
+    #--------------------------------------------------------------------------
+    def close(self):
+        self.database.close()
+
+    #--------------------------------------------------------------------------
+    new_crashes = __iter__
+
+    #--------------------------------------------------------------------------
+    def __call__(self):
+        return self.__iter__()
+
+
+#==============================================================================
+class PGPVNewCrashSource(PGQueryNewCrashSource):
+    required_config = Namespace()
+    required_config.crash_id_query = change_default(
+        PGQueryNewCrashSource,
+        'crash_id_query',
+        "select uuid "
+        "from reports_clean rc join product_versions pv "
+        "    on rc.product_version_id = pv.product_version_id "
+        "where "
+        "%s <= date_processed and date_processed < %s"
+        "and %s between pv.build_date and pv.sunset_date"
+    )
+    required_config.add_option(
+        'date',
+        doc="a date in the form YYYY-MM-DD",
+        default=(utc_now() - timedelta(1)).date(),
+        from_string_converter=string_to_datetime
+    )
+
+    #--------------------------------------------------------------------------
+    def __init__(self, config, name, quit_check_callback=None):
+        super(PGPVNewCrashSource, self).__init__(
+            config,
+            name,
+            quit_check_callback
+        )
+        self.data = (
+            config.date,
+            config.date + timedelta(1),  # add a day
+            config.date
+        )
+
+
+#==============================================================================
+class DBCrashStorageWrapperNewCrashSource(PGQueryNewCrashSource):
+    """This class is both a crashstorage system and a new_crash_source
+    iterator.  The base FTSApp classes ties the iteration of new crashes
+    to the crashstorage system designed as the 'source'.  This class is
+    appropriate for use in that case as a 'source'."""
+
+    required_config = Namespace()
+    required_config.namespace('implementation')
+    required_config.implementation.add_option(
+        'crashstorage_class',
+        default='socorro.external.boto.crashstorage.BotoS3CrashStorage',
+        doc='a class for a source of raw crashes',
+        from_string_converter=class_converter
+    )
+
+    #--------------------------------------------------------------------------
+    def __init__(self, config, name=None, quit_check_callback=None):
+        super(DBCrashStorageWrapperNewCrashSource, self).__init__(
+            config,
+            name=name,
+            quit_check_callback=quit_check_callback
+        )
+        self._implementation = config.implementation.crashstorage_class(
+            config.implementation,
+            quit_check_callback
+        )
+
+    #--------------------------------------------------------------------------
+    def close(self):
+        super(DBCrashStorageWrapperNewCrashSource, self).close()
+        self._implementation.close()
+
+    #--------------------------------------------------------------------------
+    def __getattr__(self, method):
+        def inner(*args, **kwargs):
+            return getattr(self._implementation, method)(*args, **kwargs)
+        return inner
--- a/socorro/app/fetch_transform_save_app.py
+++ b/socorro/app/fetch_transform_save_app.py
@@ -423,7 +423,8 @@ class FetchTransformSaveWithSeparateNewCrashSourceApp(FetchTransformSaveApp):
             self.new_crash_source = \
                 self.config.new_crash_source.new_crash_source_class(
                     self.config.new_crash_source,
-                    self.app_instance_name
+                    name=self.app_instance_name,
+                    quit_check_callback=self.quit_check
                 )
         else:
             # the configuration failed to provide a "new_crash_source", fall
--- a/socorro/collector/submitter_app.py
+++ b/socorro/collector/submitter_app.py
@@ -13,8 +13,7 @@ from os import (
     listdir
 )
 
-from configman import Namespace, RequiredConfig
-from configman.converters import class_converter
+from configman import Namespace
 
 from socorro.app.fetch_transform_save_app import (
     FetchTransformSaveWithSeparateNewCrashSourceApp,
@@ -26,7 +25,6 @@ from socorro.external.crashstorage_base import (
 )
 from socorro.external.fs.filesystem import findFileGenerator
 from socorro.lib.util import DotDict
-from socorro.external.postgresql.dbapi2_util import execute_query_iter
 
 
 #==============================================================================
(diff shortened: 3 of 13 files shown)


socorro/external/postgresql/new_crash_source.py
```diff
@@ -0,0 +1,242 @@
...
+            '.ConnectionContext',
+        doc='the class responsible for connecting to Postgres',
+        from_string_converter=class_converter,
+        reference_value_from='resource.postgresql',
+    )
+    required_config.add_option(
+        'crash_id_query',
+        doc='sql to get a list of crash_ids',
+        default="select 'some_id'",
+        likely_to_be_changed=True,
```
peterbe: SQL in .ini files or in consul is weird. Why does this have to be a option?

It reminds me of the horror of `--crontabber.class-FetchADIFromHiveCronApp.query` and `--crontabber.class-BugzillaCronApp.query`

Don't let me stop you, but for the record I think queries belong in code where you have syntax highlighting and git etc. If they need to be changed, sub-classing is my preferred way.

socorro/external/postgresql/new_crash_source.py
```diff
@@ -0,0 +1,242 @@
...
+
+        for a_crash_id in crash_ids:
+            yield a_crash_id
+
+    #--------------------------------------------------------------------------
+    def close(self):
+        self.database.close()
+
+    #--------------------------------------------------------------------------
+    new_crashes = __iter__
```
peterbe: This looks like a legacy/backwards compat fix where the name has evolved over time. 
Perhaps this is something of FTS that I don't get but can we stick to just one name?

socorro/external/postgresql/new_crash_source.py
```diff
@@ -0,0 +1,242 @@
...
+
+    #--------------------------------------------------------------------------
+    def get_unredacted_processed(self, crash_id):
+        return self._implementation.get_unredacted_processed(crash_id)
+
+    #--------------------------------------------------------------------------
+    def remove(self, crash_id):
+        self._implementation.remove(crash_id)
+
+
```
peterbe: If all this class does it just to wrap the `self._implementation` then there's a much shorter way to do it:
```python
    def __getattr__(self, method):
        def inner(*args, **kwargs):
            return getattr(self._implementation, method)(*args, **kwargs)
        return inner
```
And if you need to NOT do this with any one in particular (e.g. the `close()`) you can simply just leave that one.

socorro/external/postgresql/new_crash_source.py
```diff
@@ -0,0 +1,242 @@
...
+        "where ( "
+        "date_processed >= %s and date_processed < %s "
+        "and ({pv_where_clause}))"
+    )
+    required_config.add_option(
+        "product_version_list",
+        doc='list of the form "p1, v1a, v1b; p2, v2a, v2b; ..."',
+        default='Firefox',
+        from_string_converter=two_level_nested_list_converter(),
+        likely_to_be_changed=True
```
peterbe: Again, it's probably because I don't understand FTS as well as you do but this is a strange thing to change in config. 

Why does it need to be transcoded to a string? Why can't these classes talk to each other and pass a simple `dict` object?

socorro/external/postgresql/new_crash_source.py
```diff
@@ -0,0 +1,242 @@
...
+
+#==============================================================================
+class PGPVNewCrashSource(PGQueryNewCrashSource):
+    required_config = Namespace()
+    required_config.crash_id_query = change_default(
+        PGQueryNewCrashSource,
+        'crash_id_query',
+        "select uuid from reports "
+        "where ( "
+        "date_processed >= %s and date_processed < %s "
```
peterbe: BETWEEN is better.

socorro/external/postgresql/new_crash_source.py
```diff
@@ -0,0 +1,242 @@
...
+        self._implementation.remove(crash_id)
+
+
+#==============================================================================
+class PGPVNewCrashSource(PGQueryNewCrashSource):
+    required_config = Namespace()
+    required_config.crash_id_query = change_default(
+        PGQueryNewCrashSource,
+        'crash_id_query',
+        "select uuid from reports "
```
peterbe: @AdrianGaudebert might cry if he sees a new piece of code that's come to depend on the `reports` table.

socorro/external/postgresql/new_crash_source.py
```diff
@@ -0,0 +1,242 @@
...
+        "product_version_list",
+        doc='list of the form "p1, v1a, v1b; p2, v2a, v2b; ..."',
+        default='Firefox',
+        from_string_converter=two_level_nested_list_converter(),
+        likely_to_be_changed=True
+    )
+    required_config.add_option(
+        'date',
+        doc="a date in the form YYYY-MM-DD",
+        default=(utc_now() - timedelta(1)).date(),
```
peterbe: Again, why is this configurable? It should just be a passing parameter. 
Having a default that freezes when you import the code just feels wrong. 
I've seen this in other places too and it makes it cringe ever time. 

If it really has to be a config value, can the default at least be `default=lambda: =(utc_now() - timedelta(1)).date()`

peterbe: What is `timedelta(1)`? Is that 1 second or one day? I can never remember.

socorro/external/postgresql/new_crash_source.py
```diff
@@ -0,0 +1,242 @@
...
+            ' or '.join(
+                [where_clause_element] * len(config.product_version_list)
+            )
+        )
+        self.crash_id_query = config.crash_id_query.format(
+            pv_where_clause=pv_where_clause,
+        )
+        self.data = [
+            config.date,
+            config.date + timedelta(1)
```
peterbe: if you accept that this should be a lambda instead, above this line it needs to be...:
```python
if callable(config.date):
    config.date = config.date()
```

socorro/external/postgresql/new_crash_source.py
```diff
@@ -0,0 +1,242 @@
...
+        ]
+
+        def flatten(list_of_tuples):
+            for x in iter(list_of_tuples):
+                if isinstance(x, Sequence) and not isinstance(x, basestring):
```
peterbe: If something is a `Sequence` why would it ever be a string?

socorro/external/postgresql/new_crash_source.py
```diff
@@ -0,0 +1,242 @@
...
+        self.crash_id_query = config.crash_id_query.format(
+            pv_where_clause=pv_where_clause,
+        )
+        self.data = [
+            config.date,
+            config.date + timedelta(1)
+        ]
+
+        def flatten(list_of_tuples):
+            for x in iter(list_of_tuples):
```
peterbe: Why wrap a list in `iter()`? 
One advantage of the `iter()` built in is that you can use it to "extract" a generator. I.e. a pump of values (aka. a stream) that doesn't buffer massive arrays in memory. But if it's already list, it doesn't add any value. 
Another advantage of doing `iter(some_list)` is so you can assign it to a variable and call `.next()` but that's already what the for-loop does.

socorro/external/postgresql/new_crash_source.py
```diff
@@ -0,0 +1,242 @@
...
+                if isinstance(x, Sequence) and not isinstance(x, basestring):
+                    for y in x:
+                        yield y
+                else:
+                    yield x
+
+        self.data.extend(
+            flatten(sorted(config.product_version_list.iteritems()))
```
peterbe: If there's a reason why it needs to be sorted, I think this code needs a comment that explains that justification.

socorro/external/postgresql/new_crash_source.py
twobraids: nope, it is not.  between does  this:  lower limit <= item <= upper limit.  That "<=" on the upper limit will cause the query to needlessly hit two partitions on one day each week. Part of my afternoon yesterday was doing ```explain analyze``` on the query.   The use of between doubles the cost of the query.

socorro/external/postgresql/new_crash_source.py
twobraids: again in terms of query optimization, using reports_clean would require joins that takes the cost of the query up by an order of magnitude.  I originally wrote it using both "between" and "reports_clean" and then analyzed the queries.

socorro/external/postgresql/new_crash_source.py
peterbe: As far as I can see, the only reason to sort them is to make the tests predictable. 
If that's the case, I don't think that's fair. The code shouldn't be at the mercy of the tests unless it's for "fluffying things up" so it becomes easier to reach into units.

socorro/external/postgresql/new_crash_source.py
twobraids: as I said yesterday, I'm making PRs smaller by taking the large ones and breaking them up.  This involves putting shims in so that the code continues to work before the next PR takes it to the next step.  In an upcoming PR, this name will migrate completely away from the legacy.  I could have added the name change here to this PR but that made the PR too big.

socorro/external/postgresql/new_crash_source.py
twobraids: I'm iterating over a sequence of sequences [ 'Firefox', ('v1', 'v2', 'v3')] with the aim of flattening it to ['Firefox', 'v1', 'v2', 'v3'].  Since a string is also a sequence, if I did not make the special case, it would flatten to [ 'F', 'i', 'r', 'e', 'f', 'o', 'x', 'v1', 'v2', v3' ]

peterbe:
Two things I struggle to get over and you might need to talk me down from these with reasons beyond my sight. 

* The dependency on the `reports` table. It's a rough and crazy table. We've been trying to stop depending on it. Sure, `reports_clean` is a bit better and the whole `reports -> storedprocedure() -> reports_clean` is just an optimization thing. I know that @AdrianGaudebert and I are trying to make the webapp completely agnostic of the `reports` table **and** of the `reports_clean` table too. In fact, we want to keep PG for some things but reading queries of crashes ES is the better tool. And for the webapp, SuperSearch is the only API for ES. 
So the point of all this, is to get a stream of UUIDs, right? 

* The `PGPVNewCrashSource` class is kinda strange in that it queries and gathers data at "instantiation time". That's kinda unusual. It's not uncommon to see calls to `self._fetch_data()` or something at the bottom of  a class's `__init__` method. That's a bit weird too. 
It might just be that I don't get the whole picture of how this fits together but my job is, I guess, to say that this is odd and it doesn't feel very natural. I would generally prefer if the api was `instance = SomeClass(); print instance.get_data()` rather than `instance = SomeClass(); print instance.data`

socorro/external/postgresql/new_crash_source.py
peterbe: Aha! I see. That's because normally you do things like this:
`BETWEEN date = '2015-10

socorro/external/postgresql/new_crash_source.py
peterbe: Yeah, but that join would be on integer reference instead of doing string matches on the `product_name` and `version_string` and stuff, on every row. If you do a join with `product_versions` you'll only do that string matching once and the join is quick over indexed integer references. 

Another important difference between `reports` and `reports_clean` is that `reports_clean` is always a "day" behind and doesn't have the real-time-ness that `reports` has. However, the correlations script could be made to depend on the `reports-clean` stored procedure.

socorro/external/postgresql/new_crash_source.py
peterbe: Ah! Then you can do this:
```
for x in list_of_tuples:
    if isinstance(x, tuple):
       for y in x:
            yield y
    else:
        yield x
```

socorro/external/postgresql/new_crash_source.py
twobraids: let's say RelEng tagged a release incorrectly with the wrong channel, and we need to just those crashes to a special processor that could correct the problem, how would you do it?  By having the query readily available on the command line, I can reuse this class with a crashmover or submitter route all those crashes into wherever they need to go in minutes.  

If my only option was writing code, then a 30 minute job would become a 3 day ordeal of writing code, making a PR, getting review, sending to staging, making a release.

socorro/external/postgresql/new_crash_source.py
twobraids: yes, I like that.

socorro/external/postgresql/new_crash_source.py
twobraids: I expect that the data may come from a expansion from a json object in the future which doesn't deliver a tuple, but a list instead.  Best to be flexible rather than brittle.

twobraids:
@peterbe on re-examining the use of the "reports_clean" table instead of the "reports" table, I discovered a reason that the use of "reports_clean" trumps the use of "reports".  Even the though the query is more expensive, I can use the "product_versions" table to figure out automatically which products and versions to include in the results.  This eliminates the config parameter of list of products and versions entirely.

twobraids:
@peterbe, even further analysis shows even more fascinating results - 'reports' is a more efficient on prod, but 'reports_clean' is more efficient on staging.  This has to do with memory and use of the tables.  'reports_clean' is not generally in memory because its use is not constant.  'reports' is almost always in memory because the processors are constantly writing to it and updating the index.  That's why it is faster.  If I switch to a reports partition that is not currently active, the query is no longer more efficient than the equivalent query on 'reports_clean'.  On stging, I surmise that there is enough memory for both the current 'reports' partition and 'reports_clean' to be in memory at the same time.

AdrianGaudebert:
I'm going to argue against using the ``reports`` table as well. It is doomed to disappear as soon as we have the resources to make it die, and no new thing should be built on top of it. If reading from the ``reports_clean`` table works, that's great, otherwise Elasticsearch should be the way to go (through Super Search or not).

peterbe:
@AdrianGaudebert isn't `reports_clean` just a cleaned up version of `reports`. I.e. a [stored procedure](https://github.com/mozilla/socorro/blob/master/socorro/cron/jobs/matviews.py#L145) re-writes what it finds for that hour in `reports` and normalizes it and stores in `reports_clean`. In other words, if `reports` is "doomed to disappear" then so is `reports_clean`.

AdrianGaudebert:
@peterbe The plan, as I understood it, was to update that stored procedure you mentioned to use ``processed_crashes`` as a source instead of ``reports``. ``reports_clean`` is better optimized and more tightly controlled than ``reports``, is still and can still be used by our apps. But as you said earlier, Elasticsearch / Super Search should be the standard sources for crashes now, so maybe at some point we will also decommission ``reports_clean``?

socorro/external/postgresql/new_crash_source.py
twobraids: btw, you can see that this class is actually a specialization of a pre-existing class.  It changes the default query of the base class to be the special one for the correlation reports.  There is no need to put the query in an ini file or environment.  However, if the need arises for a special run with a different query to fix a problem, the capability is there.  

While some may deem this to be a flexibility that we don't need, I must point out that the base class is the driver behind the stage submitter, the primary means to reprocess crashes after a signature generation change and it was key to resolving the crash-with-the-wrong-release channel from several months ago.

twobraids:
@peterbe I'm not following your objection to ```PGPVNewCrashSource``` gathering data at instantiation time.  It does not do the query until it is used as an iterator, so it gathers no data at instantiation time.

socorro/external/postgresql/new_crash_source.py
```diff
@@ -0,0 +1,158 @@
...
+    #--------------------------------------------------------------------------
+    def close(self):
+        self.database.close()
+
+    #--------------------------------------------------------------------------
+    new_crashes = __iter__
+
+    #--------------------------------------------------------------------------
+    def __call__(self):
+        return self.__iter__()
```
peterbe: you could just write 
```
__call__ = __iter__
```
I think.

peterbe:
r+

PR #3067, opened 2015-10-28
Bug 1219384 - Update forced versions for correlations for Firefox cycle starting 2015-10-29

KaiRo-at:
This adds current aurora and nightly versions as well as all planned beta versions for this cycle, and should land on prod this week still.

Note that need to update the crashanalysis node with that code (and make sure this version of the script is actually being run) to actually make it live.

### DIFF
--- a/scripts/crons/cron_libraries.sh
+++ b/scripts/crons/cron_libraries.sh
@@ -70,7 +70,7 @@ do
   techo "Phase 1: end"
 done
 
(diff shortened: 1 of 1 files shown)


KaiRo-at:
FWIW, the 42.0b99 version is left in so that we make sure we don't lose this version (RC for the release on Tuesday) for the next week.

PR #3070, opened 2015-10-29
fixes bug 1219776 - Remove beta hack on crashes per user

peterbe:
Wow! I can't believe we let the new `Crashes per User` be landed without tests. 

So there's two big changes here. 

1. Test coverage of the `crashes_per_user`. One when everything is normal. And one when one of the requested versions is one of those "beta" releases. I.e. it ends with a "b". 

2. "Explosion and compression". Basically, when the user requests to see the data for version "XX.b", before doing the ADI query *and* the SuperSearch query, I look into the known list of current releases for all other versions that start with "XX.b" but is not "XX.b". So basically, if the request is `?p=Firefox&version=19.0&version=18.0b` instead of doing the queries on `versions=['19.0', '18.0b']` I convert that to `versions=['19.0', '18.0b1', '18.0b2', '18.0bn', ...]`. And when the results come back, I compress them back all under the version "18.0b". E.g. instead of `[{'count': 20, 'term': '18.0b1'}, {'count': 30, 'term': '18.0b2'}]` it becomes `[{'count': 20 + 30, 'term': '18.0b'}]`. 

I'm sorry that the PR looks big and scary. Be sure to run it locally and compare the numbers for 42.0b on the old and the new daily reports. It adds up on my screen. 
screenshot 2015-10-29 13 36 57

### DIFF
--- a/socorro/external/postgresql/adi.py
+++ b/socorro/external/postgresql/adi.py
@@ -41,7 +41,27 @@ class ADI(PostgreSQLBase):
         if missing:
             raise MissingArgumentError(', '.join(missing))
 
-        SQL = """
+        sql_versions = []
+        for i, version in enumerate(params['versions'], start=1):
+            key = 'version{}'.format(i)
+            # We make a very special exception for versions that end with
+            # the letter 'b'. It means it's a beta version and when some
+            # queries on that version they actually mean all
+            # the "sub-versions". For example version="19.0b" actually
+            # means "all versions starting with '19.0b'".
+            # This is succinct with what we do in SuperSearch.
+            if version.endswith('b'):
+                # exception!
+                sql_versions.append('pv.version_string LIKE %({})s'.format(
+                    key
+                ))
+                version += '%'
+            else:
+                # the norm
+                sql_versions.append('pv.version_string = %({})s'.format(key))
+            params[key] = version
+
+        sql = """
             SELECT
                 SUM(adu_count) AS adi_count,
                 adu_date AS date,
@@ -52,18 +72,20 @@ class ADI(PostgreSQLBase):
             LEFT OUTER JOIN product_versions pv USING (product_version_id)
             WHERE
                 pv.product_name = %(product)s
-                AND pv.version_string IN %(versions)s
+                AND ({})
                 AND os_name IN %(platforms)s
                 AND adu_date BETWEEN %(start_date)s AND %(end_date)s
             GROUP BY
                 adu_date,
                 build_type,
                 version_string
-        """
+        """.format(
+            ' OR '.join(sql_versions)
+        )
 
-        params['versions'] = tuple(params['versions'])
         params['platforms'] = tuple(params['platforms'])
-        results = self.query(SQL, params)
+        assert isinstance(params, dict)
+        results = self.query(sql, params)
 
         rows = []
         for row in results.zipped():
--- a/webapp-django/crashstats/crashstats/tests/test_views.py
+++ b/webapp-django/crashstats/crashstats/tests/test_views.py
@@ -5,6 +5,7 @@ import json
 import mock
 import re
 import urllib
+import random
 
 import pyquery
 
@@ -15,6 +16,7 @@ from django.test.client import RequestFactory
 from django.test.utils import override_settings
 from django.conf import settings
 from django.contrib.auth import REDIRECT_FIELD_NAME
+from django.utils import timezone
 from django.contrib.auth.models import (
     User,
     AnonymousUser,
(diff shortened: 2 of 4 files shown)


webapp-django/crashstats/crashstats/views.py
```diff
@@ -1050,14 +1049,23 @@ def crashes_per_user(request, default_context=None):
     default = request.session.get('date_range_type', 'report')
     context['date_range_type'] = params.get('date_range_type') or default
 
-    api = SuperSearchUnredacted()
-    # XXX HACK! Because of a bug in SuperSearch we can't search for
-    # '4X.0b' so we have to change that to '4X.0b1' :(
-    # See https://bugzilla.mozilla.org/show_bug.cgi?id=1197251
+    params['original_versions'] = params['versions'][:]  # make a copy
```
peterbe: This takes a copy of what the user expected. 
During the course of this function we're transform `params['versions']` and I want an original copy to be displayed in the template rendering (and CSV outputter).

webapp-django/crashstats/crashstats/views.py
```diff
@@ -1050,14 +1049,23 @@ def crashes_per_user(request, default_context=None):
...
             params['versions'].remove(version)
+            # look through all the releases for this product
+            exploded_version_numbers[version] = [
+                x['version']
+                for x in context['releases'][params['product']]
+                if x['version'].startswith(version) and x['version'] != version
+            ]
+    for versions in exploded_version_numbers.values():
+        params['versions'].extend(versions)
+
```
peterbe: So this whole thing simply converts:
```
params['versions'] = ['18.0b', '19.0']
```
into 
```
params['versions'] = ['19.0', '18.0b1', '18.0b2']
exploded_version_numbers = {'18.0b': ['18.0b1', '18.0b2']}
```

webapp-django/crashstats/crashstats/views.py
```diff
@@ -1143,6 +1151,79 @@ def crashes_per_user(request, default_context=None):
...
+                    new['count'] += facet_cluster['count']
+                    date_cluster['facets']['version'].remove(facet_cluster)
+            date_cluster['facets']['version'].append(new)
+        # Lastly for the "summary facet"
+        new_summary_cluster = {'count': 0, 'term': version}
+        for facet_cluster in list(results['facets']['version']):
+            if facet_cluster['term'] in sub_versions:
+                new_summary_cluster['count'] += facet_cluster['count']
+                results['facets']['version'].remove(facet_cluster)
+        results['facets']['version'].append(new_summary_cluster)
```
peterbe: What this thing above means is that [this hack](https://github.com/mozilla/socorro/blob/f1022e9ef9d4df19b1e674ab6b1199d4a561797f/socorro/lib/search_common.py#L385) is "no longer needed". I put it in apostrophes because it's not needed by the Crashes per User report any more.

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -1894,6 +1896,339 @@ def mocked_get(url, params, **options):
...
+            'p': 'WaterWolf',
+            'v': ['20.0', '19.0']
+        })
+        eq_(response.status_code, 200)
+
+        # There's not a whole lot we can easily test in the output
+        # because it's random numbers in the mock functions
+        # and trying to not use random numbers and looking for
+        # exact sums or whatnot is error prone because we might
+        # asset that the number appears anywhere.
```
AdrianGaudebert: Or you can use ``pyquery`` and select just the right elements to check. :) I find it weird that you use ``random`` in unit tests to be honest, it feels wrong because you cannot test that the numbers are correct.

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -1894,6 +1896,339 @@ def mocked_get(url, params, **options):
...
+
+        # Table headers for each version
+        ok_('19.0' in re.findall('>(.*?)</th>', response.content))
+        ok_('20.0' in re.findall('>(.*?)</th>', response.content))
+
+        main_content = response.content.split('<h2>Crashes per ADI</h2>')[1]
+        # There should be 14 days worth of content in the main table.
+        # Plus two for the standard headers
+        eq_(main_content.count('</tr>'), 14 + 2)
```
AdrianGaudebert: I think all of those things would be done better with ``pyquery``.

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -1894,6 +1896,339 @@ def mocked_get(url, params, **options):
...
+        yesterday = timezone.now() - datetime.timedelta(days=1)
+
+        def is_percentage(thing):
+            return thing.endswith('%') and float(thing.replace('%', ''))
+
+        eq_(first_row[0], yesterday.strftime('%Y-%m-%d'))
+        ok_(first_row[1].isdigit())  # crashes
+        ok_(first_row[2].isdigit())  # adi
+        ok_(is_percentage(first_row[3]))  # throttle
+        ok_(is_percentage(first_row[4]))  # ratio
```
AdrianGaudebert: I assume you made sure none of those values can be zero?

webapp-django/crashstats/crashstats/views.py
AdrianGaudebert:
```python
stuff = list(things)
stuff = things[:]
```

I like the first one better, it's easier to understand and doesn't require the ``make a copy`` comment. See the end of http://henry.precheur.org/python/copy_list for opinion.

webapp-django/crashstats/crashstats/views.py
AdrianGaudebert: Most of this should not be needed. Super Search is doing that for you (as proved in your comments in the bug). You only need the part where you merge the versions together, but there is no need to split in the params.

webapp-django/crashstats/crashstats/views.py
AdrianGaudebert: That hack is actually needed, and is where it belongs. It works for Super Search, Signature report and the public API with a little hack in one place. If we were to move that feature to the webapp, we would have to duplicate it in each and every place querying Super Search, including external users' scripts.

webapp-django/crashstats/crashstats/views.py
AdrianGaudebert: I actually think those 12 lines or so should be removed. There is no need to touch the params. Instead you should change the results only, to merge what needs to be merged.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -1143,6 +1151,79 @@ def crashes_per_user(request, default_context=None):
...
+    # But no number of "1.x" if that was what exploded into 1.1, 1.2, 1.3.
+    # So we need to put that back and manually sum the counts. So that
+    # eventually we end up with this instead:
+    #
+    #  version 1.x : 5700,   (=1234+3456+1010 as per the example)
+    #
+    # Note that we assume the throttle to be the same.
+    #
+
+    for version, sub_versions in exploded_version_numbers.items():
```
AdrianGaudebert: I don't like that you rely on the "official" version numbers. Super Search knows better than that, and sometimes even have version numbers that did not make it to our database yet. There's the ``b0`` exception as well (when we did not have data about a build at processing time). 

So I would instead go for a different approach: look at all the versions ending in ``b`` (e.g. ``18.0b``), and then go through all results, looking for version numbers starting with the same number (``18.0b``) and merge those together.

AdrianGaudebert:
r- (sorry :confused:  )

I don't think this is the best way to solve this problem Happy to chat about it if you like!

webapp-django/crashstats/crashstats/views.py
peterbe: I'm not convinced. Neither did I read that long blog post. 

You can do `[:]` on a tuple too and what you get is a copy. It's the equivalent of `import copy; thing = copy.copy(otherthing)`. If you do `list(something)` you always get a list. Even if it was a tuple. Also, if you do `[:]` you **know** it must be a list (or tuple). It can't be a generator because generators don't support slicing. 

I see `list()` often being used in cased where you desperately need to exhaust a generator. I.e. asking it to return all its elements right here right now. 

If you were to use `list()` instead of `[:]` you'd still need to have the... ` # make a copy` comment because it tells the reader that that's why you're doing it. Not because it was a generator that needs to be fully iterated. 

In fact, with `[:]` there is no need to say `# make a copy` because, of course, it can only be a list. 

I guess I'm so used to `[:]` that I was blind to your gut reaction that `list()` "[is] easier to understand". I can change it if you like.

AdrianGaudebert: Well, I just find that ``[:]`` is a lot more cryptic than ``list()``. If you're not well versed into Python, it's difficult to understand the purpose of ``[:]``. With ``list()`` it's clear that you are creating a new object. I understand both, but I tend to use ``list()`` because I think it is simpler to read and understand. 

But I do not care enough to argue more than just that, so it's your call, really. :) (And I do not really mind having both ways in the code base, it's just a styling detail. )

webapp-django/crashstats/crashstats/tests/test_views.py
peterbe: The random fixtures can not be a zero.

webapp-django/crashstats/crashstats/tests/test_views.py
peterbe: done.

webapp-django/crashstats/crashstats/tests/test_views.py
peterbe: you're right.

peterbe:
Ok. I've attacked the low-hanging fruit of feedback. I.e. removing the HTML string hacking and replaced with pyquery. 

The reason why I convert `18.0b` to `18.0b1,18.0b2,18.0bN` is because of [adi.py](https://github.com/mozilla/socorro/blob/master/socorro/external/postgresql/adi.py#L44-L62) It can't do a `LIKE` query but I guess I could extend that to be something like this:
```sql
WHERE pv.product_name = 'WaterWolf'
AND (pv.version_string IN ('19.0') OR pv.version_string LIKE '18.0b%')
```
I.e. Somewhere in `adi.py` we loop over the versions and look for any that ends in a `b`. 

That'd mean you'd be able to API query the ADI endpoint and ask it how many ADI it has for "Firefox 42.0b". [this is not possible](https://crash-stats.mozilla.com/api/ADI/?end_date=2015-11-04&platforms=Linux&product=Firefox&start_date=2015-10-04&versions=42.0b) but if you ask [specifically for 42.0b1](https://crash-stats.mozilla.com/api/ADI/?end_date=2015-11-04&platforms=Linux&product=Firefox&start_date=2015-10-04&versions=42.0b1) it yields results. 

Let's talk on our Thursday meeting.

peterbe:
Ok. Got this URL working locally:
`http://socorro.dev/api/ADI/?end_date=2015-11-04&platforms=Windows&product=Firefox&start_date=2015-10-24&versions=42.0b` and it returns ADI counts of all versions that start with "42.0b"

peterbe:
Yay! Freckin yay @AdrianGaudebert !
It works. Now the ADI can accept queries for versions like `44.0b` and the Crashes per User report matches the old one. r?

socorro/external/postgresql/adi.py
```diff
@@ -51,19 +51,40 @@ def get(self, **kwargs):
...
             LEFT OUTER JOIN product_versions pv USING (product_version_id)
             WHERE
-                pv.product_name = %(product)s
-                AND pv.version_string IN %(versions)s
-                AND os_name IN %(platforms)s
-                AND adu_date BETWEEN %(start_date)s AND %(end_date)s
+                pv.product_name = %%s
+                AND (%s)
+                AND os_name IN %%s
+                AND adu_date BETWEEN %%s AND %%s
```
AdrianGaudebert: Why not keep ``%%(product)s`` and ``%%(start_date)s`` and such? I find it a lot more readable and robust than relying on an ordered list of parameters as soon as there are more than 2 elements.

socorro/external/postgresql/adi.py
```diff
@@ -51,19 +51,40 @@ def get(self, **kwargs):
...
+            # We make a very special exception for versions that end with
+            # the letter 'b'. It means it's a beta version and when some
+            # queries on that version they actually mean all
+            # the "sub-versions". For example version="19.0b" actually
+            # means "all versions starting with '19.0b'".
+            # This is succinct with what we do in SuperSearch.
+            if version.endswith('b'):
+                # exception!
+                sql_versions.append('pv.version_string LIKE %s')
+                version += '%'
```
AdrianGaudebert: This is great!

socorro/external/postgresql/adi.py
peterbe: Is that even possible? Can you do a tuple **and** keyword arguments?

AdrianGaudebert: Ah my bad, I didn't see that you were doing a tuple formatting for the versions. Well it's fine then. It's still possible to make it a keyword formatting, but it will make the logic ugly, so let's forget it.

peterbe: Yeah, I've seen code where it creates a "numbered key". You know, that's maybe not such a bad idea. Long tuples are stupid and more dangerous. I think I want to change to use a dict where I do something like:
```python
for i, version in enumerate(version):
    sql_params['version{}'.format(i)] = version
```

peterbe: The advantage is that it reads some saner and safer with `...WHERE product = %(product_name)s` IMHO.

AdrianGaudebert: I did that in the old advanced search back-end code, and it was quite ugly. But there were a lot of them scattered in different places, so maybe for something small like here it could be OK. Anyway, I agree that dict > tuple.

peterbe:
I'm not entirely sure if I have your r+ but in case I do, I rebased it and included a change to use a dict instead of a long tuple for the params to the SQL query.

AdrianGaudebert:
I knew that image was not a good enough "r+". So: r+ :)

PR #3072, opened 2015-10-30
fixes bug 1219960 - Exploitability report based on SuperSearch

peterbe:
screenshot 2015-10-30 11 02 09
Actually, this screenshot was taken before I had the pagination working. 

I'm really excited about this change! The exploitability is generated by the processors somehow and inserted into the processed_crash which is stored in ES. Right?
That means, if we switch to this new report I've made, we'll be able to tear down sooo much. 
1) all the crontabber jobs related to it
2) the implementation class that queries the `exploitability_reports` table. 
3) the crashstats model for wrapping that implementation class
4) the `exploitability_reports` table which apparently weighs 6.5Gb

I'm not sure how best to accomplish the pagination. Or the filtering. 

#### About the pagination...
We only display X signatures per page on the report. But the `'_aggs.signature': 'exploitability',` query returns ALL signatures of crashes in the last 7 days. 
To be able to do the pagination in ES we first need to sort. The sorting rule to sort by those that have the highest `high+medium` count. Can we do this? Do we need to bother?

Another thing that would be nice is to be able to filter aggregates whose sum of `high`, `medium`, `low` and `none` is greater than 4. 

In all fairness, I don't think it is critical to be able to do all this fancy filtering in the query. The expensiveness lies in doing a "python join" on the bug associations and that's done by first limiting the crashes to one page at a time.


peterbe:
Is it crazy that we're even doing this in python?? Technically we could do it all in Javascript. In fact, I think it doesn't really matter. The django stuff is easier to test with django's test framework and it's kinda neat that the python code ONLY does the M in MVC in a sense. This way, it frees up the javascript code to just do the rendering.

webapp-django/crashstats/crashstats/forms.py
```diff
@@ -519,3 +519,35 @@ class GraphicsReportForm(BaseForm):
...
+        self.available_products = kwargs.pop('available_products')
+        super(ExploitabilityReportForm, self).__init__(*args, **kwargs)
+
+    def clean_product(self):
+        value = self.cleaned_data['product']
+        if value not in self.available_products:
+            raise forms.ValidationError(
+                '{} is not an available product'.format(value)
+            )
+        return value
```
AdrianGaudebert: Would that not be better done with a ``ChoiceField``? You can feed it a list of choices at runtime in the ``__init__``. Some forms already do something similar, like here: https://github.com/mozilla/socorro/blob/master/webapp-django/crashstats/supersearch/forms.py#L47-L61

peterbe: You're right. I think I rushed that.

webapp-django/crashstats/crashstats/templates/crashstats/exploitability_report.html
```diff
@@ -0,0 +1,81 @@
...
+                        <th>Count LOW</th>
+                        <th>Count NONE</th>
+                    </tr>
+                </thead>
+                <tbody>
+                    <tr>
+                        <td colspan="6" class="ajax-loading">
+                            <!-- this gets replaced by jQuery loading from AJAX -->
+                            <img
+                                src="{{ static('img/icons/ajax-loader.gif') }}">
```
AdrianGaudebert: Why JS and not jinja? Jinja is what we use everywhere else, and recently we even removed mustache in favor of jinja, so I don't get why going back to doing templating in JavaScript.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/exploitability_report.js
```diff
@@ -0,0 +1,119 @@
...
+
+        // let's the first piece of data going
+        var page = 1;
+        if (document.location.hash) {
+            page = parseInt(document.location.hash.replace('#', ''), 10);
+        }
+        fetchReports(page);
+    });
+
+}($, document));
```
AdrianGaudebert: This whole file seems out-of-place in the current Socorro. Ideally we will rewrite Socorro as a JS app some day, but at the moment it's a django / jinja app, and it is weird to me that we mix JS and jinja as templating tools. 

Also, understanding an HTML structure built with jQuery is a horrible experience to me. The day we do JS templating, I really hope we'll use something like JSX.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/exploitability_report.js
```diff
@@ -0,0 +1,119 @@
...
+            updatePaginationLinks(
+                page,
+                response.count,
+                response.batch_size
+            );
+            BugLinks.enhance();
+        })
+        .fail(function() {
+            console.error.apply(console, arguments);
+            alert("Loading reports failed.");
```
AdrianGaudebert: Mmm, don't you want to show an error in the HTML content instead? I find alerts to be pretty aggressive.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/exploitability_report.js
```diff
@@ -0,0 +1,119 @@
...
+            } else {
+                page = parseInt($(this).attr('href').replace('#', ''), 10);
+                fetchReports(page);
+            }
+        });
+
+        // let's the first piece of data going
+        var page = 1;
+        if (document.location.hash) {
+            page = parseInt(document.location.hash.replace('#', ''), 10);
```
AdrianGaudebert: It seems weird to me to have the page number in the hash. Wouldn't it be better as a query string parameter? Even if it's only used in JS, it seems cleaner to me.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/exploitability_report.js
```diff
@@ -0,0 +1,119 @@
...
+        if (page * size < count) {
+            $('a.last', parent)
+                .attr('href', '#' + Math.ceil(count / size))
+                .removeClass('impossible');
+        } else {
+            $('a.last', parent)
+                .attr('href', '#' + Math.ceil(count / size))
+                .addClass('impossible');
+        }
+    }
```
AdrianGaudebert: I'm really having trouble with this. We have a pagination macro in jinja that works very well and would allow you to do all this in just one line.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -1354,9 +1357,156 @@ def exploitable_crashes(
     ))
     context['version'] = versions
     context['report'] = 'exploitable'
+    context['new_link_url'] = reverse('crashstats:exploitability_report')
+    query_string = {'product': product}
+    if versions:
+        query_string['version'] = versions
+    context['new_link_url'] += '?' + urllib.urlencode(query_string, True)
```
AdrianGaudebert: Shouldn't that be in the template instead of the view?

webapp-django/crashstats/crashstats/views.py
```diff
@@ -1354,9 +1357,156 @@ def exploitable_crashes(
...
     return render(request, 'crashstats/exploitability.html', context)
 
 
+@utils.json_view
```
AdrianGaudebert: I'm not sure how this decorator works, but that view is not strictly speaking a "JSON" view, it sometimes return HTML too. So that's kind of weird to be mixing both contents imo.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -1354,9 +1357,156 @@ def exploitable_crashes(
...
+    if not form.is_valid():
+        return http.HttpResponseBadRequest(str(form.errors))
+
+    if request.GET.get('format') == 'json':
+        product = form.cleaned_data['product']
+        version = form.cleaned_data['version']
+
+        api = SuperSearchUnredacted()
+        # XXX ADRIAN, how do I use settings.EXPLOITABILITY_BATCH_SIZE
+        # to get a limited set?
```
AdrianGaudebert: ``_facets_size`` defines the number of results for each facet. By default it's 50. However, there is currently now way to paginate over facets, the only thing you can do is to increase ``_facets_size`` and then show only the last X results that you want. 

I'm actually wondering if it's useful to have pagination at all on that page, to be honest?

webapp-django/crashstats/crashstats/views.py
```diff
@@ -1354,9 +1357,156 @@ def exploitable_crashes(
...
+        params = {
+            'product': product,
+            'version': version,
+            '_results_number': 0,
+            '_aggs.signature': 'exploitability',
```
AdrianGaudebert: So what's going to happen here is that the signatures will be sorted by their number of occurrences in crashes. It's basically going to be topcrashers, but you'll have the exploitability of each signature with it. I don't know if that's what we want, or if we want to see the most "exploitable" crashes first, whatever that is.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -1354,9 +1357,156 @@ def exploitable_crashes(
...
+            #      'term': 'My | Signature',
+            #      'facets': {
+            #          'exploitability': [
+            #              {'count': 1, 'term': 'high'},
+            #              {'count': 23, 'term': 'medium'},
+            #              {'count': 11, 'term': 'other'},
+            #
+            # And we only want to include those where:
+            #
+            #   null + none + low + medium + high > 4
```
AdrianGaudebert: I do not understand this logic at all. What are you trying to achieve? What does that ``4`` value represent?

AdrianGaudebert:
So it's difficult to sort results by exploitability in Elasticsearch, and impossible with Super Search's current features. But what you are doing now (sorting in Python) should be fine, if we increase the number of results we get (using ``_facets_size``) and if we filter out some results we don't care about, like everything that has exploitability as ``error``. What do you think about adding ``'exploitability': '!error'`` to the SuperSearch arguments?

AdrianGaudebert:
Oh, we don't display the results where exploitability is "none" or "interesting" as well, so might as well filter those ones out too. 

```python
params = {
    'product': product,
    'version': version,
    'exploitability': ['!none', '!error', '!interesting'],
    '_results_number': 0,
    '_aggs.signature': 'exploitability',
}
```

peterbe:
Correction, as per the discussion, let's filter on 
```
'exploitability': ['!error', '!interesting'],
```

webapp-django/crashstats/crashstats/views.py
```diff
@@ -1354,9 +1357,156 @@ def exploitable_crashes(
...
+                crash = {
+                    'bugs': [],
+                    'signature': signature_facet['term'],
+                    'high_count': 0,
+                    'medium_count': 0,
+                    'low_count': 0,
+                    'none_count': 0,
+                    'null_count': 0,
+                    'url': (
+                        reverse('crashstats:report_list') + '?' +
```
AdrianGaudebert: Can we link to Signature report instead of report/list/?

peterbe:
@AdrianGaudebert much much much simpler now.

webapp-django/crashstats/crashstats/views.py
AdrianGaudebert: This is not needed anymore!

AdrianGaudebert:
I just realized while testing that you're missing a ``context['report'] = 'exploitable'`` somewhere in your view. If you don't add it, the top navigation doesn't work. 

r+ with nits :)

webapp-django/crashstats/crashstats/views.py
peterbe: Maybe you're right but whatever :)
Because the URL is very different (to the old report) if the version is set or not I'd have to mess around a lot with jinja if statements.

PR #3074, opened 2015-11-03
Fixes bug 1215153 - Added links to Profile page, removed tokens.

AdrianGaudebert:
@peterbe r?

### DIFF
--- a/webapp-django/crashstats/profile/static/profile/css/profile.css
+++ b/webapp-django/crashstats/profile/static/profile/css/profile.css
@@ -7,12 +7,16 @@ div.page-section {
     float: right;
 }
 
-p.profile-email {
-    float: right;
-}
-
 .panel .links li {
     display: inline-block;
     font-size: 1.1em;
     margin: 10px;
+}
+
+ul.links.left {
+    float: left;
+}
+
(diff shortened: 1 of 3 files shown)


peterbe:
screenshot 2015-11-03 07 53 29
Alignment is wrong. The top bar (with the links) is larger than the two boxed below.

peterbe:
The way it says my email address in the upper right-hand corner (sort of) is odd. It's sitting there. It's not clickable or anything.

peterbe:
I think you should NOT put the email address alone up there and instead change the link from "Sign out" to "Sign out pbengtsson@mozilla.com"

AdrianGaudebert:
@peterbe comments addressed!

peterbe:
How about this?
screenshot 2015-11-06 08 35 48

It's bad enough that the `<h2>Your Profile</h2>` is floating in the empty sea and not inside a box. 

Either way, if you prefer having the sign out like you have it, then r+
Otherwise, consider my counter PR.

peterbe:
Either accept https://github.com/AdrianGaudebert/socorro/pull/4 or ignore it and I won't be offended at all.

PR #3078, opened 2015-11-05
fixes bug 1221624 - FennecAndroid repository digit trimming

peterbe:
@rhelmer Thank you for helping!
I tested this first by manually running `ftpscraper.py`. And now it inserts builds for `mozilla-aurora-android-api-11`. Before this change, the `repository` value became `mozilla-aurora-android-api` and that's not a valid repository value in `special_product_platforms`.


rhelmer:
r+

PR #3079, opened 2015-11-05
unify close behavior for FTS source/destination & TransformRules

twobraids:
three inheritance trees, FTS Apps, Crashstorge and Transform classes all have their own ways to shut themselves down.  Unify them all to use a simple 'close' method.  This makes them more compatible with each other, predictable and helps them move toward being contextmanager compatible.

### DIFF
--- a/socorro/app/fetch_transform_save_app.py
+++ b/socorro/app/fetch_transform_save_app.py
@@ -374,8 +374,17 @@ class FetchTransformSaveApp(App):
         self.config.executor_identity = self.task_manager.executor_identity
 
     #--------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


socorro/app/fetch_transform_save_app.py
```diff
@@ -387,7 +396,7 @@ def main(self):
         self._setup_task_manager()
         self._setup_source_and_destination()
         self.task_manager.blocking_start(waiting_func=self.waiting_func)
-        self._cleanup()
+        self.close()
```
peterbe: Talked to Lars, these connections are all lazy.

PR #3081, opened 2015-11-05
Fixes bug 1218751 - Use Windows NT instead of just Windows.

AdrianGaudebert:
@peterbe r?

### DIFF
--- a/socorro/processor/mozilla_transform_rules.py
+++ b/socorro/processor/mozilla_transform_rules.py
@@ -1009,6 +1009,10 @@ class OSPrettyVersionRule(Rule):
             # There is nothing we can do if the `os_name` is missing.
             return None
 
+        if not isinstance(processed_crash.os_name, basestring):
+            # This data is bogus, there's nothing we can do.
+            return None
+
         if not processed_crash.get('os_version'):
             # The version number is missing, there's nothing to do.
             return pretty_name
@@ -1022,7 +1026,7 @@ class OSPrettyVersionRule(Rule):
         major_version = int(version_split[0])
         minor_version = int(version_split[1])
 
-        if processed_crash.os_name == 'Windows':
+        if processed_crash.os_name.lower().startswith('windows'):
             # Get corresponding Windows version.
             key = '%s.%s' % (major_version, minor_version)
             if key in self._windows_versions:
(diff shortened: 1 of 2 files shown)


socorro/processor/mozilla_transform_rules.py
```diff
@@ -1022,7 +1022,7 @@ def _get_pretty_os_version(self, processed_crash):
         major_version = int(version_split[0])
         minor_version = int(version_split[1])
 
-        if processed_crash.os_name == 'Windows':
+        if processed_crash.os_name == 'Windows NT':
```
peterbe: let's make this `.startswith(...)` (as discussed)

peterbe:
r+ with nit on the startswith

AdrianGaudebert:
@peterbe rebased

socorro/processor/mozilla_transform_rules.py
```diff
@@ -1022,7 +1022,7 @@ def _get_pretty_os_version(self, processed_crash):
         major_version = int(version_split[0])
         minor_version = int(version_split[1])
 
-        if processed_crash.os_name == 'Windows':
+        if processed_crash.os_name.lower().startswith('windows'):
```
peterbe: Hmm... Sorry but this makes me nervous all of a sudden. 
Suddenly it HAS to ALWAYS be a string. Otherwise it will fail on that `.lower()`. 
Before, if the crash some crazy and screwed up (which isn't unrealistic) there was a chance that the value of `processed_crash.os_name` might be `None` (or perhaps some other crap like a number). Then this processor rule would fail here. 
The test above (on line 1007) only checks if the key is present. It doesn't check that its non-falsy. 

Sorry, but I don't feel confident around this. It feels like it's a lot more important that processor rules don't raise exceptions. 

Please have a think about it and if you don't know with confidence either perhaps it should be:
```python
if (
    processed_crash.os_name and 
    processed_crash.os_name.lower().startswith('windows')
):
```

peterbe: Shouldn't this then be
```python
if pretty_name.lower().startswith('windows'):
```

AdrianGaudebert: It's the same thing (though as soon as I had pushed I realized I should have used ``processed_crash.os_name`` instead in the verification above. )

AdrianGaudebert:
@peterbe updated (with a new commit this time :) )

PR #3083, opened 2015-11-05
Fixes bug 1221952 - Added _aggs parameter for all fields.

AdrianGaudebert:
@peterbe r?


peterbe:
r+ well done!
I like the fact that we started gently and after a couple of weeks upgraded it to be generic.

PR #3088, opened 2015-11-10
Fixes bug 1221957 - Added a Product summary.

AdrianGaudebert:
@peterbe r?

### DIFF
--- a/webapp-django/crashstats/signature/views.py
+++ b/webapp-django/crashstats/signature/views.py
@@ -439,6 +439,8 @@ def signature_summary(request, params):
         params['_histogram.date'] = ['exploitability']
 
     api = SuperSearchUnredacted()
+
+    # Now make the actual request with all expected parameters.
     try:
         search_results = api.get(**params)
     except models.BadStatusCodeError as e:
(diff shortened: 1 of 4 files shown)


peterbe:
The numbers are VERY different
screenshot 2015-11-11 16 53 43

screenshot 2015-11-11 16 54 26

peterbe:
Also, what's up with it being 99.3% only on Firefox 38.4.0esr?!

webapp-django/crashstats/signature/views.py
```diff
@@ -439,6 +439,32 @@ def signature_summary(request, params):
...
+
+    product_version_facet = None
+
+    # If there is a version or a product filter, we need to make a separate
+    # query so that we can show all versions and not just the one asked for.
+    if (
+        ('version' in params and params['version'])
+        or ('product' in params and params['product'])
+    ):
+        params_copy = dict(params)
```
peterbe: that makes a shallow copy. Is that OK? I.e. are there no deeper levels that 1 level?

webapp-django/crashstats/signature/views.py
```diff
@@ -439,6 +439,32 @@ def signature_summary(request, params):
...
+        del params_copy['product']
+        del params_copy['version']
+        params_copy['_aggs.product'] = ['version']
+
+        try:
+            search_results = api.get(**params_copy)
+        except models.BadStatusCodeError as e:
```
peterbe: yeah! "as e"

peterbe:
Now's maybe not the time but that page could really do with some different grid system. When one table gets really large it does get quite messed up. 
screenshot 2015-11-11 16 57 47

AdrianGaudebert:
Stage postgres and ES are very inconsistent. I wouldn't give any credit to those numbers (though if the same discrepancy happens on prod, we'll have a problem). 

Regarding the boxes, it's definitely out of scope, but I agree with you. Ideally we should use some sort of flex system to make all of that better.

webapp-django/crashstats/signature/views.py
AdrianGaudebert: No, I'm not changing the content of the values, just removing some keys, so that's fine.

AdrianGaudebert:
Actually, after talking to KaiRo on IRC, I'm going to revisit this a bit to remove all filters to the query instead of just product and version.

peterbe:
By the way, I can't find anything wrong with the code. So if you're confident the data queries are right and that the results will be more sane on prod, then r+

AdrianGaudebert:
@peterbe I just added a new commit, r?

peterbe:
Otherwise, r+

PR #3089, opened 2015-11-11
fixes bug 1217184 - We have two webapp wsgi scripts

peterbe:


peterbe:
@lonnen wanna merge this one too?

PR #3091, opened 2015-11-12
fixes bug 1224011 - KeyError on histogram_date in Crashes per User

peterbe:
What do you think @AdrianGaudebert ?

### DIFF
--- a/webapp-django/crashstats/crashstats/tests/test_views.py
+++ b/webapp-django/crashstats/crashstats/tests/test_views.py
@@ -6,6 +6,7 @@ import mock
 import re
 import urllib
 import random
+import urlparse
 
 import pyquery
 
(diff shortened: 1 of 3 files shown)


AdrianGaudebert:
If the error is caused by a parameter in the URL, don't you want to do a redirect instead? This way the user at least has a chance to see that something changed in his request. 

If you disagree, this is fine, r+.

peterbe:
Actually, you have me convinced. I think I like that better. It's a little bit more transparent.

peterbe:
This feels a lot better @AdrianGaudebert r?

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -2166,6 +2167,25 @@ def is_percentage(thing):
...
+        # If we use self.assertRedirects() it will actually go to the
+        # redirected URL which we haven't set up mocking for yet.
+        eq_(response.status_code, 302)
+        redirect_url = response['Location']
+        parsed = urlparse.urlparse(redirect_url)
+        eq_(parsed.path, url)
+        ok_('product=Whatever'in parsed.query)
+        ok_('foo=bar'in parsed.query)
+        ok_('date_range_type=build' not in parsed.query)
+        ok_('date_range_type=' not in parsed.query)
```
AdrianGaudebert: These two lines are redundant, the first one is not needed.

AdrianGaudebert:
r+

webapp-django/crashstats/crashstats/tests/test_views.py
peterbe: They are not redundant. They are the making sure that the new redirect URL doesn't have those things in there any more.

AdrianGaudebert: Just for the sake of arguing: if ``date_range_type=`` is not in ``parsed.query``, ``date_range_type=build`` definitely is not in it either. :) So the second line covers both cases, and the first one is unneeded.

AdrianGaudebert: (Though please do not bother changing that, it's really not worth our time. )

peterbe: NOW I see what you mean. Thank you for not giving up on me. We can just keep the ` ok_('date_range_type=' not in parsed.query)` line.

peterbe: oh. I merged it already :)

PR #3092, opened 2015-11-12
Fixes bug 1224151 - Allowed multiple values in graphics_devices service.

AdrianGaudebert:
@peterbe r?

### DIFF
--- a/socorro/external/postgresql/graphics_devices.py
+++ b/socorro/external/postgresql/graphics_devices.py
@@ -15,28 +15,32 @@ class GraphicsDevices(PostgreSQLBase):
 
     def get(self, **kwargs):
         filters = [
-            ("vendor_hex", None, "str"),
-            ("adapter_hex", None, "str"),
+            ("vendor_hex", None, ["list", "str"]),
+            ("adapter_hex", None, ["list", "str"]),
         ]
         params = external_common.parse_arguments(filters, kwargs)
         for key in ('vendor_hex', 'adapter_hex'):
-            if not params[key]:
+            param = params[key]
+            if not param:
                 raise MissingArgumentError(key)
 
-        sql_where = """
-            WHERE
-                vendor_hex = %(vendor_hex)s
-                AND
-                adapter_hex = %(adapter_hex)s
-        """
+            params[key] = tuple(params[key])
+
         sql_query = """
             SELECT
                 vendor_hex, adapter_hex, vendor_name, adapter_name
             FROM graphics_device
+            WHERE vendor_hex IN %(vendor_hex)s
+            AND adapter_hex IN %(adapter_hex)s
         """
-        results = self.query(sql_query + sql_where, params)
+
+        results = self.query(sql_query, params)
         hits = results.zipped()
-        return {'hits': hits, 'total': len(hits)}
+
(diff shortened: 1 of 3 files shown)


socorro/unittest/external/postgresql/test_graphics_devices.py
```diff
@@ -12,7 +12,7 @@
 from .unittestbase import PostgreSQLTestCase
 
 
-#==============================================================================
+# =============================================================================
```
peterbe: This is a larsism that is here to stay. Sure, we don't allow it in the webapp-django directory but for now I suggest you just allow it be by allowing flake8 to check and it should respect our [setup.cfg](https://github.com/mozilla/socorro/blob/master/setup.cfg).

socorro/external/postgresql/graphics_devices.py
```diff
@@ -15,28 +15,32 @@ class GraphicsDevices(PostgreSQLBase):
...
-            ("vendor_hex", None, "str"),
-            ("adapter_hex", None, "str"),
+            ("vendor_hex", None, ["list", "str"]),
+            ("adapter_hex", None, ["list", "str"]),
         ]
         params = external_common.parse_arguments(filters, kwargs)
         for key in ('vendor_hex', 'adapter_hex'):
-            if not params[key]:
+            param = params[key]
+            if not param or (len(param) == 1 and not param[0]):
```
peterbe: What is this?! Something like `[0]` or `['']`? Why is that ever an input to this? Also why is it a problem? Why not just let the query go ahead and if nothing comes back, nothing comes back. 
And why specifically 1? What if there are others. Like `['hex1', 'hex2', '']`

socorro/external/postgresql/graphics_devices.py
```diff
@@ -15,28 +15,32 @@ class GraphicsDevices(PostgreSQLBase):
 
     def get(self, **kwargs):
         filters = [
-            ("vendor_hex", None, "str"),
-            ("adapter_hex", None, "str"),
+            ("vendor_hex", None, ["list", "str"]),
+            ("adapter_hex", None, ["list", "str"]),
```
peterbe: Does that mean it can be either a string or a list? Or a list of strings?

peterbe:
Some questionmarks to be discussed.

socorro/external/postgresql/graphics_devices.py
AdrianGaudebert: It means it has to be a list of strings.

socorro/external/postgresql/graphics_devices.py
AdrianGaudebert: This is a side effect of ``parse_arguments``, which will generate ``['']`` when the argument was empty. It should probably actually be solved in the ``parse_arguments`` function instead.

AdrianGaudebert:
@peterbe addressed

peterbe:
r+

peterbe:
Actually...

I wonder if this will break the admin page where you can look up individual hexes. Care to check that first?

AdrianGaudebert:
@peterbe That was one of the first things I verified. :)

PR #3093, opened 2015-11-13
support loading correlations of multiple versions

peterbe:
I am so sorry that this is so huge. I'll try to explain why that is so with comments within. 

**NOTE!** Please read my general comments first before you dig in. It'll help.


peterbe:
First of all, the general gist here is that instead of loading correlations for each type, it can take a list of types. It used to be:
```
/correlation?product=...&version=...&platform=...&signature=...&correlation_report_type=core-counts
/correlation?product=...&version=...&platform=...&signature=...&correlation_report_type=interesting-addons
/correlation?product=...&version=...&platform=...&signature=...&correlation_report_type=interesting-modules
```
Now it's just one like this:
```
/correlation?product=...&version=...&platform=...&signature=...&correlation_report_types=core-counts&correlation_report_types=interesting-addons&correlation_report_types=interesting-modules
```
Why did I do that? The reason is that I now want to load these repeatedly. One for each version & OS combo. So if I load 3 different versions and 3 different AJAX queries per version it'd be 9 queries. 
Instead, if all types are lumped as one AJAX query, it'll just be 3 queries for 3 versions. 

That was the warm-up.

peterbe:
Note, that the middleware wasn't changed. The underlying `socorro.external.http.correlations.Correlations` class is unchanged and expects one type at a time. 
The reason I didn't attack that too was because I wanted to keep the PR small. Also it'd break the API for people and I'd need to think about a way of solving for that. 

I might try to solve that as a followup commit as part of this PR once I get the r+ on the basic webapp stuff.

peterbe:
Next, instead of taking the os and version that has the most crashes of the `ReportList` query, I now sort (by number of crashes) them as combinations and take the 3 top-most (see `MAX_CORRELATION_COMBOS_PER_SIGNATURE` setting).

peterbe:
Next, notice that I added a whole new file called `accordions.js` to this PR. The reason I did that was that the existing `accordion.js` was a module, meaning you can only have now. 
So I rewrote it as a function that can be loaded with multiple and independent contexts. Otherwise these lines wouldn't be possible:
```javascript
$('.accordion').each(function() {
    (new Accordion(this)).init();
});
```
So, please don't tease me for the quality of that file. It's just a quick re-write of something I have no idea where it came from :)

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/topcrash.js
```diff
@@ -15,23 +15,25 @@ var Correlations = (function() {
...
-        }).join(',');
-
-        $.getJSON(SocReport.sig_base, data, function (json) {
-            $.each(json.hits, function(i, sig) {
-                all_signatures[sig] = 1;
-            });
+        for (var platform in osnames) {
+            data.platforms.push(platform);
+        }
+        $.getJSON(SocReport.sig_base + '?' + $.param(data, true))
```
peterbe: Apparently you can do things like this with `$.getJSON(url, {foo: 'bar'})` but not this: `$.getJSON(url, {foo: ['bar', 'baz']})`. That's why I use `$.param(data, true)`. It then turns it into `'foo=bar&foo=baz'`

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/topcrash.js
```diff
@@ -45,25 +47,25 @@ var Correlations = (function() {
...
+
+        $.getJSON(SocReport.base + '?' + $.param(data, true))
+        .done(function(json) {
+            for (var i in correlations) {
+                var type = correlations[i];
+                var report = '<h3>' + json[type].reason + '</h3>';
+                report += json[type].load.split("\n").join("<br>");
                 row.find('.' + type).html(report);
-                types_done += 1;
+            }
```
peterbe: This ugly construction of a big fat blob of HTML as a string should die. But it is what it is. Hopefully we'll do it differently when we do Correlations on the Signature report.

webapp-django/crashstats/crashstats/templates/crashstats/partials/correlations.html
```diff
@@ -1,82 +1,83 @@
...
+                    </a>
+                </h3>
+                <div id="core-counts-correlation{{ loop.index }}" class="content-pane" role="tabpanel" aria-labelledby="cpucores-heading" tabindex="-1">
+                    <p class="spinner">Loading
+                        <img src="{{ static('img/loading.png') }}" width="16" height="17" alt="loading spinner" />
+                    </p>
+                </div>
+            </li>
+        </ul>
+    </div>
```
peterbe: The gist of this big block of change is that it used to be things like:
```
<div id="something">...</div>
```
Now that I wrap it in a loop you can't repeat the same ID so I basically changed it to:
```
{% for thing in things %}
<div id="something{{ loop.index }}">...</div>
{% endfor %}
```
so that the IDs became unique.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -2063,35 +2061,34 @@ def report_list(request, partial=None, default_context=None):
...
+            {
+                'os': k[0],
+                'version': k[1],
+            }
+            for k, v in sorted(
+                counts.items(),
+                key=lambda x: x[1],
+                reverse=True
+            )
+        ][:settings.MAX_CORRELATION_COMBOS_PER_SIGNATURE]
```
peterbe: Sorry for the messy list comprehension. It might make @lonnen throw up in his mouth a little. 

The more important point here is the crazy fact that we're picking an arbitrary number. Why 3? Why not 2 or 7 or 34? I don't know. I just didn't want to cause too much AJAX queries if there's a signature that has many different kinds of versions and OSes. 

Having said that, this whole logic is based on the **first page** of the report list. (You made this point yourself Adrian). So if a signature has 300 44.0a2 on the first page and 43.0b1 only on the second page, you might miss out on valuable correlations.

What to do? Add a new middleware endpoint thing where you can get aggregates of crashes for that signature and pick the 3 most "crashiest". 
Or, do that with a SuperSearch query? Or, just accept this as much-better-than-it-was-and-it-ll-soon-be-replaced-by-signature-report-anyway.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -2734,8 +2731,21 @@ def correlations_json(request):
...
+        # type at a time. At some point we ought to change that so it
+        # takes in a list instead.
+        # One big change at a time!
+        context[report_type] = api.get(
+            report_type=report_type,
+            product=product,
+            version=version,
+            platform=platform,
+            signature=signature
+        )
```
peterbe: Places like this I want to change too. Maybe. That'd reduced the number of queries to the middleware. 

However, making many small queries is sometimes better than one big monster query. After all, the way correlations work (under the Correlations tab) is that it loads the 3 most interesting kinds and then you can manually extra load other ones one at a time.

peterbe:
@AdrianGaudebert I've incorporated @KaiRo-at 's [points from our discussion](https://bugzilla.mozilla.org/show_bug.cgi?id=1218585#c10) and rebased it into one commit. r?

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/topcrash.js
AdrianGaudebert: Yes, I know this trick quite well, having used it in Super Search a few times.

AdrianGaudebert:
So the main issue that I see with this is that it keeps relying on the reports sent by the ReportList service, and that data is *not* representative of the products/versions/os in the overall data set. 

It is however maybe outside the scope of this work. And I can't think of an easy solution to this within report/list/ (on the other hand, it would be quite easy to solve in signature report using an aggregation).

peterbe:
It bothers me too. But...

1) It doesn't bother Kairo because the crashes we're finding the versions based on is sorted chronologically so the most recent ones are on the first page. 
2) When we move this over to signature report, we can find out the "hottest versions" with a simple super search query.

AdrianGaudebert:
Tested locally, it seems to work fine. r+

PR #3094, opened 2015-11-16
this needs an absolute path

mmohrhard:


rhelmer:
You're right, thank you very much for your contribution @mmohrhard !

PR #3099, opened 2015-11-17
fixes bug 1225600 - Correlation http downloads gets gzip wrong

peterbe:
So this has apparently been a bug for a very long time. 
Any file that ends with a .gz [for example](https://crash-analysis.mozilla.com/crash_analysis/20151116/20151116_Firefox_43.0b2-interesting-modules-with-versions.txt.gz) gets downloaded in raw gzip format and saved in the temp directory as `something.txt` and then when it parses it it always fails because the content is not a regular string.

### DIFF
--- a/socorro/unittest/external/http/test_correlations.py
+++ b/socorro/unittest/external/http/test_correlations.py
@@ -3,9 +3,11 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import datetime
+import gzip
 import shutil
 import os
 import tempfile
+import io
 
 import mock
 from nose.tools import eq_, ok_, assert_raises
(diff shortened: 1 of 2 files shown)


peterbe:
cc @AdrianGaudebert

peterbe:
@twobraids if you don't have time, would you mind @AdrianGaudebert reviews this one instead.

socorro/unittest/external/http/test_correlations.py
```diff
@@ -25,9 +27,20 @@
 
 
 class Response(object):
-    def __init__(self, content=None, status_code=200):
-        self.content = content
+    def __init__(self, content=None, status_code=200, gzip=False):
```
AdrianGaudebert: ``gzip`` is a terrible name for a boolean. Especially since it "conflicts" with the gzip lib that you import. Could you consider something like ``do_gzip`` instead?

AdrianGaudebert:
r+

socorro/unittest/external/http/test_correlations.py
peterbe: fixed. I now call it `compress`. Why didn't I think of that in the first place?!

PR #3100, opened 2015-11-18
Fixes bug 1225567 - Added a note that the product summary does not us…

AdrianGaudebert:
…e filters.

@peterbe r?

### DIFF
--- a/webapp-django/crashstats/signature/templates/signature/signature_summary.html
+++ b/webapp-django/crashstats/signature/templates/signature/signature_summary.html
@@ -52,7 +52,7 @@
 
 <div class="panel">
     <header class="title">
-        <h2>Product</h2>
+        <h2>Product <sup>*</sup></h2>
     </header>
     <div class="content">
         <table class="tablesorter data-table">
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/signature/templates/signature/signature_summary.html
```diff
@@ -52,7 +52,7 @@
 
 <div class="panel">
     <header class="title">
-        <h2>Product</h2>
+        <h2>Product *</h2>
```
peterbe: You should use `<h2>Product<sup>*</sup></h2>` so you get a raised character. 
And use that too at the beginning of the text below.

webapp-django/crashstats/signature/templates/signature/signature_summary.html
```diff
@@ -77,6 +77,9 @@
                 {% endfor %}
             </tbody>
         </table>
+        <em>
```
peterbe: `em` is not a blocking tag. By the way, `em` (and `strong`) is a relic from the XHTML days. The more modern one to use is `i` (and `b`).

peterbe: I.e. I think it should be: `<p class="footnote">` and something like `p.footnote { font-style: italic; }`

AdrianGaudebert:
@peterbe updated!

peterbe:
r+ but before you merge, consider changing the word "week" to 7 days. 
Because the summary isn't from Monday to today. It's from (today - 7days)  to today.

PR #3102, opened 2015-11-19
Fixes bug 1225447 - Split query to graphics_devices into smaller chun…

AdrianGaudebert:
…ks to avoid hitting URL size limit.

@peterbe r?

### DIFF
--- a/webapp-django/crashstats/signature/tests/test_views.py
+++ b/webapp-django/crashstats/signature/tests/test_views.py
@@ -664,7 +664,7 @@ class TestViews(BaseTestViews):
                     'total': 2
                 })
 
-            raise NotImplementedError()
+            raise NotImplementedError(url)
 
         rget.side_effect = mocked_get
 
@@ -730,16 +730,16 @@ class TestViews(BaseTestViews):
                     ],
                     "adapter_vendor_id": [
                         {
-                            "term": "Intel (0x0086)",
+                            "term": "0x0086",
                             "count": 4,
                             "facets": {
                                 "adapter_device_id": [
                                     {
-                                        "term": "Device (0x1234)",
+                                        "term": "0x1234",
                                         "count": 2,
                                     },
                                     {
-                                        "term": "Other (0x1239)",
+                                        "term": "0x1239",
                                         "count": 2,
                                     }
                                 ]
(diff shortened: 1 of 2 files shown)


AdrianGaudebert:
So I had to do as you proposed and split the query, because the ``graphics_devices`` service already has a POST endpoint, so I couldn't just write ``post = get``.

webapp-django/crashstats/signature/tests/test_views.py
```diff
@@ -843,3 +843,69 @@ def mocked_supersearch_get(**params):
...
+
+        rget.side_effect = mocked_get
+
+        def mocked_supersearch_get(**params):
+            ok_('signature' in params)
+            eq_(params['signature'], ['=' + DUMB_SIGNATURE])
+
+            adapters = [
+                {
+                    'term': '0x{0:0>4}'.format(i),
```
peterbe: what does this fancy thing do? :)

webapp-django/crashstats/signature/views.py
```diff
@@ -502,22 +502,30 @@ def signature_summary(request, params):
...
-            devices_results = devices_api.get(
-                vendor_hex=vendor_hexes,
-                adapter_hex=adapter_hexes,
+
+        devices_results = {
+            'hits': [],
+            'total': 0,
+        }
+        # In order to avoid hitting the maximum URL size limit, we split the
+        # query in smaller chunks, and then we reconstruct the results.
```
peterbe: Thank you! for adding a comment.

webapp-django/crashstats/signature/views.py
```diff
@@ -502,22 +502,30 @@ def signature_summary(request, params):
...
+        for i in range(0, len(vendor_hexes), 50):
```
peterbe: I did a some checking. 50 of these amounts to a query string that is 1843 bytes, given a random set, as a single query string. That's quite tiny. 

So let's say that's 37 bytes per vendor/hex combo. 

Nginx by default, can read 4 buffers of 4000 bytes. I think. That means, if you allow for some other headers too, we should be able to send about 3x4000 bytes which would mean roughly we should be able to send 324 vendor/adaptor hex combos per request from our webhead to our middleware server.

webapp-django/crashstats/signature/views.py
```diff
@@ -502,22 +502,30 @@ def signature_summary(request, params):
...
+            vendors = set(vendor_hexes[i:i + 50])
+            adapters = set(adapter_hexes[i:i + 50])
```
peterbe: Can you replace these 3 occurrences of the number 50 with a variable.

webapp-django/crashstats/signature/tests/test_views.py
```diff
@@ -843,3 +843,69 @@ def mocked_supersearch_get(**params):
...
+            if '/graphics_devices' in url:
+                ok_(len(params['vendor_hex']) <= 50)
+                ok_(len(params['adapter_hex']) <= 50)
+
+                return Response({
+                    'hits': [],
+                    'total': 0
+                })
+
+            raise NotImplementedError()
```
peterbe: I usually make that `raise NotImplementedError(url)`

webapp-django/crashstats/signature/tests/test_views.py
```diff
@@ -843,3 +843,69 @@ def mocked_supersearch_get(**params):
...
+                    }
+                }
+                for i in range(3)
+            ]
+
+            res = {
+                "hits": [],
+                "total": 4,
+                "facets": {
+                    "adapter_vendor_id": vendors,
```
peterbe: mixed quotation styles between here and 10 lines up.

webapp-django/crashstats/signature/views.py
```diff
@@ -502,22 +502,30 @@ def signature_summary(request, params):
...
+            res = devices_api.get(
+                vendor_hex=vendors,
+                adapter_hex=adapters,
             )
-        except models.BadStatusCodeError as e:
-            # We need to return the error message in some HTML form for jQuery
-            # to pick it up.
-            return http.HttpResponseBadRequest('<ul><li>%s</li></ul>' % e)
+            devices_results['hits'].extend(res['hits'])
+            devices_results['total'] += res['total']
```
peterbe: smart!

webapp-django/crashstats/signature/tests/test_views.py
AdrianGaudebert: It makes sure there are always 4 chars, and fills the string with zeros to the left if ``i`` is smaller than 4 chars. Do I really have to put a comment to explain that? It's part of the standard python library, not some deep logic that we built.

webapp-django/crashstats/signature/views.py
AdrianGaudebert: I think getting the right number here is going to be the most important part. I got to ``50`` by just guessing from what works on stage and what doesn't, but there's no science there. 

Also as we discovered it's not nginx that blocks but uwsgi, so we are capped at 4096 bytes according to the logs you shared earlier. Can you find out much the "empty" query weights?

webapp-django/crashstats/signature/tests/test_views.py
peterbe: No, I knew it was a format trick. I just didn't recognize it "by heart". 
We only comment on business logic.

PR #3103, opened 2015-11-19
Fixes bug 1225445 - Added _aggs.product.version parameter to Super Search.

AdrianGaudebert:
@peterbe r?

### DIFF
--- a/socorro/external/es/supersearch.py
+++ b/socorro/external/es/supersearch.py
@@ -152,44 +152,36 @@ class SuperSearch(SearchBase):
         needed for backwards compatibility.
         """
         aggs = aggregations.to_dict()
-        for agg in aggs:
-            if 'buckets' not in aggs[agg]:
-                # This is a cardinality aggregation, there are no terms.
-                continue
 
-            for i, bucket in enumerate(aggs[agg]['buckets']):
-                sub_aggs = {}
+        def _format(aggregation):
+            if 'buckets' not in aggregation:
+                # This is a cardinality aggregation, there are no terms.
+                return aggregation
+
+            for i, bucket in enumerate(aggregation['buckets']):
+                new_bucket = {
+                    'term': bucket.get('key_as_string', bucket['key']),
+                    'count': bucket['doc_count'],
+                }
+                facets = {}
+
                 for key in bucket:
                     # Go through all sub aggregations. Those are contained in
                     # all the keys that are not 'key' or 'count'.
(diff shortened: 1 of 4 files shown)


socorro/external/es/supersearch.py
```diff
@@ -422,6 +414,29 @@ def get(self, **kwargs):
...
+
+            for param in params['_aggs.product.version']:
+                self._add_second_level_aggs(
+                    param,
+                    lvl2_bucket,
+                    facets_size,
+                    histogram_intervals,
+                )
+
+            search.aggs.bucket(lvl1_field, lvl1_bucket)
```
peterbe: I don't understand why you hardcoded this. I can see why you hardcoded the `_aggs.product.version` in the `SearchBase` class but these lines above can all be done with a looping over the keys. 

```python
for key in params:
    if key.startswith('_aggs.'):
        lvl1, lvl2 = key.split('.')[1:]
        ...
```

socorro/unittest/external/es/test_supersearch.py
```diff
@@ -1057,6 +1061,20 @@ def test_get_with_sub_aggregations(self):
         version_sub_facet = res['facets']['signature'][0]['facets']['version']
         eq_(len(version_sub_facet), number_of_crashes)
 
+        # Test with a third level aggregation.
+        kwargs = {
+            '_aggs.product.version': ['_cardinality.signature'],
+        }
+        res = self.api.get(**kwargs)
+
```
peterbe: What should happen if you do...
```python
kwargs={'_aggs.xxxxx.yyy': ['_cardinatliy.signature']}
self.api.get(**kwargs)
```

peterbe:
r+ but I think you can refactor that inner part where you break it up. So it's only hardcoded where the key is added to the list of available fields.

socorro/unittest/external/es/test_supersearch.py
AdrianGaudebert: That would simply be ignored. I was actually wondering, should SuperSearch be more aggressive in validating arguments? At the moment it simply just ignores stuff that it cannot use (like incorrectly spelled field names, or incorrect parameters), and returns some results that might not be what you were expecting. Should it instead raise exceptions when something seems off?

peterbe: > Should it instead raise exceptions when something seems off? 

Yes please :)

PR #3106, opened 2015-11-23
bug 1227131 - crontabber app to clean up raw_adi_logs

peterbe:


socorro/unittest/cron/jobs/test_clean_raw_adi_logs.py
```diff
@@ -0,0 +1,72 @@
...
+        assert not information['clean-raw-adi-logs']['last_error']
+        assert information['clean-raw-adi-logs']['last_success']
+
+        # Ensure test row was removed
+        cur.execute("""
+            SELECT report_date FROM raw_adi_logs
+        """)
+        result, = cur.fetchall()
+        report_date = result[0]
+        eq_(report_date, second)
```
adngdb: Should you not assert that ``len(result) == 1`` or something?

adngdb:
I see this job won't be in the default list of jobs, which means we will have to manually enable it. I think that's good. Let's be very careful when we test this on stage. 

r+

socorro/unittest/cron/jobs/test_clean_raw_adi_logs.py
peterbe: No point. First of all, I already check that only one record gets returned from that SQL select. If it was anything other than exactly 1 record (aka "row") the `result, = cur.fetchall()` would fail. 
Then, result is a tuple of length 1 and it's length 1 because I only select 1 column (i.e. `report_date`). If I were to get more or less than 1 we have a serious bug in `psycopg2`.

PR #3107, opened 2015-11-23
bug 1227142 - crontabber app to clean up raw_adi

peterbe:


peterbe:
Actually @twobraids I don't think I want a "single transaction". I don't want to use transactions at all. Do we have a decorator for that?

socorro/cron/jobs/clean_raw_adi.py
```diff
@@ -0,0 +1,34 @@
+from configman import Namespace
+from crontabber.base import BaseCronApp
+from crontabber.mixins import (
+    with_postgres_transactions,
+    with_single_postgres_transaction,
+)
+
+
+@with_postgres_transactions()
+@with_single_postgres_transaction()
```
adngdb: Did you figure out that thing about having no transaction at all?

adngdb:
This is essentially the same as your other PR, so r+.

socorro/cron/jobs/clean_raw_adi.py
peterbe: Nope and I'm not interested. I manually deleted all the old cruft. 
This new job is going to run every day so the amount of stuff it deletes is only going to be a days worth.

PR #3109, opened 2015-11-24
Fixes bug 1227514 - Sorted graphics adapters by count.

AdrianGaudebert:
@peterbe r?

### DIFF
--- a/webapp-django/crashstats/signature/views.py
+++ b/webapp-django/crashstats/signature/views.py
@@ -556,7 +556,14 @@ def signature_summary(request, params):
 
                 graphics.append(entry)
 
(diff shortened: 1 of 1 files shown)


webapp-django/crashstats/signature/views.py
```diff
@@ -556,7 +556,11 @@ def signature_summary(request, params):
 
                 graphics.append(entry)
 
-        facets['adapter_vendor_id'] = graphics
+        facets['adapter_vendor_id'] = sorted(
```
peterbe: Add a comment that explains *why* this couldn't be done in the query and why it has to be done in django.

peterbe:
r+ with nit about lacking comment.

PR #3112, opened 2015-11-25
fixes bug 1227685 - CurrentProducts.get is called too much

peterbe:
It's still bad, but it's a LOT better now. 

Before:
before
**53 `memcache.get` calls**

After:
after
**30 `memcache.get` calls**

What this is, is the metric for rendering 1 Report list page of a randomly picked signature. 

So why does it still need 13 calls just render 1 page? It loads one AJAX call for every tab (which is a good thing) and each time it needs to verify that the product and version supplied in the AJAX URL's query string is sane. 

This is going to make for much less memcache IO and it's going to make the Report list page faster.
There's still more "deeper" work that can be done but considering that the Report list a slowly dying feature it's not worth going further at the moment.

### DIFF
--- a/webapp-django/crashstats/crashstats/forms.py
+++ b/webapp-django/crashstats/crashstats/forms.py
@@ -87,7 +87,6 @@ class ReportListForm(BaseForm):
     signature = form_fields.SignatureField(required=True)
     product = forms.MultipleChoiceField(required=False)
     version = forms.MultipleChoiceField(required=False)
-    platform = forms.MultipleChoiceField(required=False)
     date = forms.DateTimeField(required=False)
     range_value = forms.IntegerField(required=False)
     reason = forms.CharField(required=False)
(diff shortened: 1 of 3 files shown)


webapp-django/crashstats/crashstats/forms.py
```diff
@@ -87,7 +87,6 @@ class ReportListForm(BaseForm):
     signature = form_fields.SignatureField(required=True)
     product = forms.MultipleChoiceField(required=False)
     version = forms.MultipleChoiceField(required=False)
-    platform = forms.MultipleChoiceField(required=False)
```
AdrianGaudebert: Why are you entirely removing the ``platform`` field? I don't get it.

peterbe: Can you find any place where it's used?

AdrianGaudebert: Holy moly, I was totally expecting the old TCBS to add it to links to report/list/ when there's an OS selected, but it's not the case, see: https://crash-stats.mozilla.com/topcrasher/products/Firefox/versions/44.0a2/date_range_type/report/crash_type/browser/os_name/Linux/result_count/50?days=7

Well, then I guess that means...

AdrianGaudebert:
... you just got yourself an r+! :)

peterbe:
haha

PR #3113, opened 2015-11-25
Fixes bug 1227621 - Added a Mobile summary to signature report.

AdrianGaudebert:
Added infinite level aggregations to Super Search.

@peterbe r?

### DIFF
--- a/webapp-django/crashstats/signature/tests/test_views.py
+++ b/webapp-django/crashstats/signature/tests/test_views.py
@@ -746,6 +746,36 @@ class TestViews(BaseTestViews):
                             }
                         }
                     ],
+                    "android_cpu_abi": [
+                        {
+                            "term": "armeabi-v7a",
+                            "count": 4,
+                            "facets": {
+                                "android_manufacturer": [
+                                    {
+                                        "term": "ZTE",
+                                        "count": 4,
+                                        "facets": {
+                                            "android_model": [
+                                                {
+                                                    "term": "roamer2",
+                                                    "count": 4,
+                                                    "facets": {
+                                                        "android_version": [
+                                                            {
+                                                                "term": "15",
+                                                                "count": 4,
+                                                            }
+                                                        ]
+                                                    }
+                                                }
+                                            ]
+                                        }
+                                    }
+                                ]
+                            }
+                        }
+                    ],
                     "histogram_uptime": [
                         {
                             "count": 2,
@@ -809,6 +839,7 @@ class TestViews(BaseTestViews):
         ok_('Product' in response.content)
         ok_('Architecture' in response.content)
         ok_('Process Type' in response.content)
+        ok_('Mobile Devices' in response.content)
         ok_('Graphics Adapter' in response.content)
         ok_('Flash&trade; Version' in response.content)
 
@@ -824,6 +855,7 @@ class TestViews(BaseTestViews):
         ok_('1.1.1.14' in response.content)
         ok_('&lt; 1 min' in response.content)
         ok_('1-5 min' in response.content)
+        ok_('ZTE' in response.content)
         ok_('Intel (0x0086)' in response.content)
 
         user = self._login()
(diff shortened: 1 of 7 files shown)


webapp-django/crashstats/signature/tests/test_views.py
```diff
@@ -746,6 +746,36 @@ def mocked_supersearch_get(**params):
...
+                                                        ]
+                                                    }
+                                                }
+                                            ]
+                                        }
+                                    }
+                                ]
+                            }
+                        }
+                    ],
```
peterbe: this hurts but welcome to the world of aggregate sub-levels I guess.

webapp-django/crashstats/signature/views.py
```diff
@@ -565,6 +568,23 @@ def signature_summary(request, params):
...
+                        mobile_devices.append({
+                            'cpu_abi': cpu_abi['term'],
+                            'manufacturer': manufacturer['term'],
+                            'model': model['term'],
+                            'version': version['term'],
+                            'count': version['count'],
+                        })
+
+        facets['android_cpu_abi'] = mobile_devices
+
```
peterbe: I can't say anything wrong with it except that it looks tremendously complex. 4 nested for-loops is not something you see very often. 

I guess it's just the nature of ES and aggregations. Or possibly that I'm too used to SQL. 
Would there be any way we can abstract this so that view code isn't so deeply nested. 
In particular, considering that you might want to expose this to users; they'd appreciate a simpler output.

peterbe:
We really need to do something about the now scary big size of that view function. I have some ideas. But first; measurement. 

I clicked on 33 different signature reports. Most of them Firefox but also a bunch of FennecAndroid. 
I put markers in the `signature_summary` view function that denotes how long certain sections take. 

They are:
* t2 - from set up till (and including) first SuperSearchUnredacted search
* t3 - also SuperSearchUnredacted `product_results = api.get(**params_copy)`
* t4 - organizing the uptime histogram
* t5 - organizing the adapter_vendor_id *and* doing the `res = devices_api.get(vendor_hex=vendors, adapter_hex=adapters)`
* t6 - organizing the mobile devices facets
* t7 - organizing the exploitability facets

The results are as follows:
```
    AVG     MEDIAN
t2 	401.61 	444.96
t3 	611.43 	638.58
t4 	0.14 	0.09
t5 	1112.0 	989.67
t6 	0.01 	0.0
t7 	0.02 	0.02
```

The first obvious thing to note is that the getting of vendor hexes is SLOOOW. It ruins the whole view. 

Another thing is that that monster 4-level-for-loop only takes virtually no time at all!

peterbe:
r+ by the way

peterbe:
Now, it bothers me that this view is getting so big. We're going to get really good mileage if we can first fix that damn GraphicsDevices thing. Perhaps we can start caching individual combos. E.g.
```python
names = {}
missing = {}
for i, adapter_hex in enumerate(adapters):
   key = 'graphics_adapters' + adapter_hex + ':' + vendors[i]
   name_pair = cache.get(key)
   if name_pair is not None:
       names[(adapter_hex, vendors[i])] = name_pair
  else:
      missing[(adapter_hex, vendors[i])] = key

vendors = []
adapters = []
for adapter_hex, vendor_hex in missing:
    adapters.append(adapter_hex)
    vendors.append(vendor_hex)
# assume that bunch of 50 per query here...
res = devices_api.get(
                vendor_hex=vendors,
                adapter_hex=adapters,
            )
for group in res['hits']:
    name_pair = (group['adapter_name'], group['vendor_name'])
    key = missing[(group['adapter_hex'], group['vendor_hex'])]
    cache.set(key, name_pair, 60 * 60 * 24)
```

Does that make sense?

peterbe:
Another thought is to use a websocket to publish results when they come in. I've used an awesome tool called fanout.io for that. 

A user makes a query. You start multiple background queries on the server. Like this:
```python
for id in request.GET.getlist('ids'):
    tasks.some_query.delay(id)
```
Then, as each job finishes (in the message queue) you let it publish to a websocket channel. 

Then our signature summary page could look something like this:
```javascript
var client = new Faye.Client('https://dfe3c5ed.fanoutcdn.com/bayeux');
client.subscribe(CURRENT_SIGNATURE, function (data) {
    container = $('.' + data.key, container).html(data.html)
})
```

PR #3118, opened 2015-12-02
Fixes bug 1221048 - Fixed link to Plugin Topcrashers in home page.

AdrianGaudebert:
@peterbe somehow I missed that during my local tests and found only on stage just now. r?

### DIFF
--- a/webapp-django/crashstats/crashstats/templates/crashstats/home.html
+++ b/webapp-django/crashstats/crashstats/templates/crashstats/home.html
@@ -61,7 +61,7 @@ Crash Data for {{ product }} {% if version %}{{ version }}{% endif %}
                             </a>
                         </li>
                         <li>
-                            <a href="{{ url('topcrashers:topcrashers') }}?product={{ product }}&amp;versions={{ version }}&amp;crash_type=plugin&amp;days={{days}}">
+                            <a href="{{ url('topcrashers:topcrashers') }}?product={{ product }}&amp;versions={{ version }}&amp;process_type=plugin&amp;days={{days}}">
                                 Top Plugin Crashers
                             </a>
                         </li>


webapp-django/crashstats/crashstats/templates/crashstats/home.html
```diff
@@ -61,7 +61,7 @@ <h2 id="homepage-heading">{{ product }} {% if version %}{{ version }}{% endif %}
                             </a>
                         </li>
                         <li>
-                            <a href="{{ url('topcrashers:topcrashers') }}?product={{ product }}&amp;versions={{ version }}&amp;crash_type=plugin&amp;days={{days}}">
+                            <a href="{{ url('topcrashers:topcrashers') }}?product={{ product }}&amp;versions={{ version }}&amp;process_type=plugin&amp;days={{days}}">
```
peterbe: Every time we have to type `&amp;` kittens die. Don't we have a helper function for making query strings?

AdrianGaudebert: I can't find any helper that does that in our code.

peterbe: hmm... we have that `change_query_string` but it works based off the current query string and we don't want that, we want to start afresh.

peterbe: Perhaps it's time to fix this. Perhaps not as part of this PR. You can decide that. 

We currently have a helper function called `urlencode` which sucks because that's not really what it does. What it does is "url quote" or "url escape". 

Ideally I think the template code should look like this:
```
 <a href="{{ url('topcrashers:topcrashers') }}?{{ urlencode(product=product, versions=version, process_type='plugin', days=days) }}">
```
don't you think?

AdrianGaudebert: I totally agree with you, but I don't have time to build this before the work week, so I filed a bug about it: https://bugzilla.mozilla.org/show_bug.cgi?id=1229832

peterbe: With that filed, r+

peterbe:
...but! I think we should wait for travis before you merge this.

PR #3120, opened 2015-12-03
fixes bug 867392 - Engage bots such as Googlebot

peterbe:


peterbe:
I've change prod consulate to take this on when it becomes available:
```
[centos@prod-admin-i-ecb76837 ~]$ consulate kv set socorro/webapp-django/ENGAGE_ROBOTS True
[centos@prod-admin-i-ecb76837 ~]$ consulate kv get socorro/webapp-django/ENGAGE_ROBOTS
True
```

lonnen:
I'm pretty sure this is not desireable.

peterbe:
So that google can find pages within Crash-stats. It's like the web thing to do. 

However, now that I think about it; I'm no longer convinced of its value. There's basically 0 brand opportunity for Google to discover this site.

peterbe:
```
[centos@prod-admin-i-ecb76837 ~]$ consulate kv rm socorro/webapp-django/ENGAGE_ROBOTS
```

lonnen:
It's also a lot of load and cache churn when they scrape. Baidu has crippled super search a few times ignoring robots.txt

peterbe:
> Baidu has crippled super search a few times ignoring robots.txt

That was **masked** as baidu. It was happening to lots of sites and supposed part of a global hack thing. We weren't the only ones suffering that day. That's how I remember it. 

Anyway, the fewer bots the better.  ...if they can respect our robots.txt file.

PR #3121, opened 2015-12-03
bug 1226345 - fail fetch-adi-from-hive if 0 rows

peterbe:
https://bugzilla.mozilla.org/show_bug.cgi?id=1226345 is a beast. 

Let me share what I know...

First of all, what's happen thrice in the last couple of days (or weeks?) is that Hive has returned 0 rows and thus the whole ADI stored procedure mess has failed. The `fetch-adi-from-hive` job says all is well because nothing broke. 

The concern is that we might be READING **whilst** Hive is being filled up and thus we might accidentally read too few ADI numbers. That's what that bug has spun into a discussion about comparing the results today with yesterday's. Or, we might be able to scrape some VPN protected complicated page and make a judgement call based on that. That's all SUPER scary and doable but means a lot of code. 

This PR takes care of the most basic situation where we get exactly 0 rows. That has actually happened at least 3 times. This PR would have been useful then because it would have disallowed the `fetch-adi-from-hive` job from succeeding and not re-trying 5 minutes later. 

I think this is going to be a much better than nothing solution.

### DIFF
--- a/socorro/cron/jobs/fetch_adi_from_hive.py
+++ b/socorro/cron/jobs/fetch_adi_from_hive.py
@@ -17,6 +17,11 @@ from crontabber.mixins import as_backfill_cron_app
 from socorro.external.postgresql.connection_context import ConnectionContext
 from socorro.external.postgresql.dbapi2_util import execute_no_results
 
+
+class NoRowsWritten(Exception):
+    pass
+
+
 """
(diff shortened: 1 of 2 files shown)


peterbe:
r=@AdrianGaudebert but he's too lazy to open up his own computer.

PR #3127, opened 2015-12-08
No bug - Better shards and replicas numbers with testing with Elasticsearch.

AdrianGaudebert:


peterbe:
I'm restarting the test to check that it works at least one more time.

peterbe:
r+

PR #3130, opened 2015-12-10
fixes bug 1186482 - deprecate /explosive/

peterbe:


AdrianGaudebert:
There are many errors in the tests, did you perhaps forget to remove a bunch of things?

peterbe:
@AdrianGaudebert it passed!

peterbe:
@AdrianGaudebert This blocks my work on switching to pipeline. Just saying :)

scripts/database/create_heroku_stage_database.sh
```diff
@@ -1,13 +1,13 @@
 #!/usr/bin/env bash
```
AdrianGaudebert: Out of scope, but I bet we can actually remove this file.

AdrianGaudebert:
:sunny: :heart_eyes: r+ would review again

peterbe:
I've just rebased it but I think I'm going to merge it after the release. That way we can let it explode (haha!) on stage for a bit first.

PR #3134, opened 2015-12-11
fixes bug 1231934 - Remove akela and socorro-toolbox git submodules

peterbe:


AdrianGaudebert:
I would love to know what those submodules were used for back when they were useful. Nothing deep, just to make sure that we won't hit some obscure deps issue after merging. Who should I ask? Do you know? 

That said, this looks good and travis is happy, so r+. :)

peterbe:
@luser @rhelmer Can either of you give us a one-liner education about what akela and socorro-toolbox was used for?

luser:
No idea, sorry.

PR #3136, opened 2015-12-14
fixes bug 1228122 - Optimize and abstract graphics devices lookup

peterbe:
I think this is an important patch. Now that signature summary is the "default" (well, I mean the new TCBS is default and that new one prefers signature report) we need to prioritize this functionality. The whole page is currently always held up by a ~1 second server rendering delay as it pulls down the graphics devices. 

We could potentially make the whole experience even better by NOT rendering this in django at all. Perhaps instead we display the table with JUST the hexes and then a piece of jQuery sucks up all the hexes and makes an AJAX query and on-the-fly replaces the hexes with real names. However, I'm not convinced that'd make the page a whole lot faster. Response time would be faster but shuffling a ton of hexes back and forth in jQuery would add to extra load strain when you thought the page was loaded.

### DIFF
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -1709,6 +1709,62 @@ class GraphicsDevices(SocorroMiddleware):
     def post(self, payload):
         return super(GraphicsDevices, self).post(self.URL_PREFIX, payload)
 
+    def get_pairs(self, adapter_hexes, vendor_hexes):
+        """return a dict where each tuple of (adapter_hex, vendor_hex)
+        corresponds to a (adapter_name, vendor_name) pair."""
+        assert len(adapter_hexes) == len(vendor_hexes)
+        names = {}
+        missing = {}
+        for i, adapter_hex in enumerate(adapter_hexes):
+            cache_key = (
+                'graphics_adapters' + adapter_hex + ':' + vendor_hexes[i]
+            )
+            name_pair = cache.get(cache_key)
+            key = (adapter_hex, vendor_hexes[i])
+            if name_pair is not None:
+                names[key] = name_pair
+            else:
+                missing[key] = cache_key
+
+        missing_vendor_hexes = []
+        missing_adapter_hexes = []
+        for adapter_hex, vendor_hex in missing:
+            missing_adapter_hexes.append(adapter_hex)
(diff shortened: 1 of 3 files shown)


webapp-django/crashstats/signature/views.py
```diff
@@ -518,17 +497,17 @@ def _transform_graphics_summary(facets):
...
                         names[0],
-                        vendor['term'],
+                        adapter['term'],
                     )
                 if names and names[1]:
-                    entry['adapter'] = '{} ({})'.format(
+                    entry['vendor'] = '{} ({})'.format(
                         names[1],
-                        adapter['term'],
+                        vendor['term'],
```
AdrianGaudebert: Why did you revert ``adapter`` and ``vendor``? If you can avoid it, the diff would be quite smaller.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -1709,6 +1709,62 @@ class GraphicsDevices(SocorroMiddleware):
...
+            if name_pair is not None:
+                names[key] = name_pair
+            else:
+                missing[key] = cache_key
+
+        missing_vendor_hexes = []
+        missing_adapter_hexes = []
+        for adapter_hex, vendor_hex in missing:
+            missing_adapter_hexes.append(adapter_hex)
+            missing_vendor_hexes.append(vendor_hex)
```
AdrianGaudebert:
```python
missing_vendor_hexes = [x[1] for x in missing]
missing_adapter_hexes = [x[0] for x in missing]
```

Isn't that the same thing, more or less? I wonder if there's any impact in doing the loop twice instead of just once.

AdrianGaudebert: A very basic test showed that it's faster to use two loops. But it was very basic and I didn't push it too far.

AdrianGaudebert:
r+

webapp-django/crashstats/signature/views.py
peterbe: Honestly can't remember. It's just that I wanted it to be adapter after vendor always. Perhaps I wrote some other code first and that had adapter followed by vendor. Sorry.

webapp-django/crashstats/crashstats/models.py
peterbe: For small lists like this it doesn't matter.

PR #3137, opened 2015-12-14
bug 1189493 - copy from FennecAndroid to WebappRuntimeMobile

peterbe:
CC @KaiRo-at @twobraids 

There's no such thing as a release called "Fennec" or "FennecAndroid". In the releases, it's called "mobile". (note that the product_name in that stored procedure is a CITEXT so it's case insensitive)
See https://bugzilla.mozilla.org/show_bug.cgi?id=1189493#c6

This PR doesn't solve the bug with getting rid of Fennec as a product version. Just that the `WebappRuntimeMobile` should be created from what really actually is "fennec". 

After this is done, I'll take care of making sure Fennec is no longer a choice as a product version.


adngdb:
@peterbe looking to the bug, I'm not sure this is ready yet. Let me know when you're sure you're doing the right thing. :)

KaiRo-at:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1189493#c13 - I just got the word in the last days that we can change this and make our lives easier: the webapp runtime should go away as well.

peterbe:
This is not the solution to get rid of Fennec. That one needs more thinking. 

However, I'm going to use this code for https://bugzilla.mozilla.org/show_bug.cgi?id=1238129

PR #3138, opened 2015-12-14
fixes bug 1227237 - Upgrade raven and track releases

peterbe:
The reason I'm confident this'll work is because I managed to send [this in](https://sentry.prod.mozaws.net/operations/socorro-stage/group/155635/) 
That's from my laptop using the socorro stage project. Scroll down to the bottom and you can see that I clearly used raven 5.9.0. 

As of this version of raven we'd now be able to append a version for all crashes. Look at [the second code example here](https://docs.getsentry.com/hosted/clients/python/advanced/). [Here](https://docs.getsentry.com/hosted/clients/python/integrations/django/)'s how you do it in Django. The problem is that can only get that version out of the middleware at runtime. Not when the uwsgi workers are starting up. I'll ask @jdotpz if he has some good ideas.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -60,8 +60,9 @@ elasticsearch-dsl==0.0.8
 urllib3==1.9.1
 # sha256: cHay6ATFKumO3svU3B-8qBMYb-f1_dYlR4OgClWntEI
 elasticutils==0.7
-# sha256: wn5AqzzPN_MKn3estJFzcNk0HiWr2o6Uub1IxxJ_fUg
-raven==3.4.1
+# sha256: NizKbyluEUlgRjuihg4eXXDEGmc3OJ6zE73mFBcTcao
+# sha256: xUXIu2A70Kc3PpdgxvsWuwZX1vNW7RIQy7lM5Bjzc74
+raven==5.9.0
 # sha256: KjGJ950ce4ohSaDng8C0IX-tmzCm59YEUPJVPcLA5X4
 simplejson==3.6.5
 # sha256: HwRtz17HEu076GhLnzPJW3bijNHIJdsPXhVXv9h7N0U
@@ -186,3 +187,5 @@ datadog==0.5.0
 freezegun==0.3.5
 # sha256: tYAUqlZA40Vup9vzPxUaHEud9HKJdocwwKAwYaRxOGU
 newrelic==2.56.0.42
+# sha256: VaXcePenQqDnVmRRNP-zm74R2g_qK8D3Bw1A2sIItzI
+contextlib2==0.4.0
--- a/webapp-django/crashstats/settings/base.py
+++ b/webapp-django/crashstats/settings/base.py
@@ -4,6 +4,7 @@
 
 import os
 import logging
+from pkg_resources import resource_string
 
 import dj_database_url
 from decouple import config, Csv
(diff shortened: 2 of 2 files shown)


peterbe:
Actually, I've figured out a neat way to get the version into Django without having to query the middleware at run-time. Since `socorro` is installed into the virtualenv, we can pick it up with setuptools. 
If that doesn't work, it'll just simply not be able to benefit from setting a release on each crash.

requirements.txt
```diff
@@ -186,3 +187,5 @@ datadog==0.5.0
 freezegun==0.3.5
 # sha256: tYAUqlZA40Vup9vzPxUaHEud9HKJdocwwKAwYaRxOGU
 newrelic==2.56.0.42
+# sha256: VaXcePenQqDnVmRRNP-zm74R2g_qK8D3Bw1A2sIItzI
+contextlib2==0.4.0
```
peterbe: Because of https://github.com/getsentry/raven-python/blob/f9060ccf67d8b19a7d06946895b1039f65e67b53/setup.py#L28

webapp-django/crashstats/settings/base.py
```diff
@@ -4,6 +4,7 @@
 
 import os
 import logging
+from pkg_resources import resource_string
```
peterbe: Note, this is a setuptools thing. Every virtualenv has it.

webapp-django/crashstats/settings/base.py
```diff
@@ -643,11 +644,22 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
...
-    'dsn': config('RAVEN_DSN', '')  # see https://errormill.mozilla.org/
-}
-
+# When socorro is installed (python setup.py install), it will create
+# a file in site-packages for socorro called "socorro/socorro_revision.txt".
+# If this socorro was installed like that, let's pick it up and use it.
+try:
+    SOCORRO_REVISION = resource_string('socorro', 'socorro_revision.txt')
+except IOError:
+    SOCORRO_REVISION = None
```
peterbe: A cool thing about this is that we no longer need to depend on the middleware doing this. 
And things like whatsdeployed.io should be able to get to this much quicker than having to execute [this beast](https://github.com/mozilla/socorro/blob/085f61878a4d6ad2384e1314e13a875ed666447e/socorro/external/postgresql/server_status.py#L19-L86) every time. 

The `except IOError` is for local devs.

webapp-django/crashstats/settings/base.py
```diff
@@ -643,11 +644,22 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
...
+
+# Raven sends errors to Sentry.
+# The release is optional.
+raven_dsn = config('RAVEN_DSN', '')
+if raven_dsn:
```
peterbe: This if statement is very important. I'm glad I tested it. What it means is that if you have...
```
RAVEN_CONFIG = {
    'dsn': '',
}
```
...you'd get this error:
```
  File "/Users/peterbe/virtualenvs/socorro/lib/python2.7/site-packages/raven/contrib/django/models.py", line 175, in sentry_exception_handler
    client.captureException(exc_info=sys.exc_info(), request=request)
  File "/Users/peterbe/virtualenvs/socorro/lib/python2.7/site-packages/raven/contrib/django/models.py", line 54, in <lambda>
    __getattr__ = lambda x, o: getattr(get_client(), o)
  File "/Users/peterbe/virtualenvs/socorro/lib/python2.7/site-packages/raven/contrib/django/models.py", line 155, in get_client
    instance = Client(**options)
  File "/Users/peterbe/virtualenvs/socorro/lib/python2.7/site-packages/raven/base.py", line 144, in __init__
    self.set_dsn(dsn, transport)
  File "/Users/peterbe/virtualenvs/socorro/lib/python2.7/site-packages/raven/base.py", line 199, in set_dsn
    transport_registry=self._registry,
  File "/Users/peterbe/virtualenvs/socorro/lib/python2.7/site-packages/raven/conf/remote.py", line 75, in from_string
    raise InvalidDsn(ERR_UNKNOWN_SCHEME.format(url.scheme, value))
InvalidDsn: Unsupported Sentry DSN scheme:  ()
```

webapp-django/crashstats/settings/base.py
```diff
@@ -643,11 +644,22 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
...
+    RAVEN_CONFIG = {
+        'dsn': raven_dsn,
+        'release': SOCORRO_REVISION,
+    }
```
peterbe: Note, if you don't define the `RAVEN_CONFIG` variable here, the `raven.contrib.django.raven_compat` app is smart enough to back off and not set up raven.

AdrianGaudebert:
Excellent pull request, very well explained! Thanks @peterbe! 

r+

PR #3141, opened 2015-12-16
Fixes bug 1228969 - Removed the version check from new topcrashers.

adngdb:
@peterbe r?


peterbe:
This is the future. r+

PR #3142, opened 2015-12-16
fixes bug 1233194 - 6 uwsgi workers for webheads

peterbe:
Perhaps now's not the time but I'm going to see about [uwsgitop](https://github.com/xrmx/uwsgitop)
You need to start uwsgi with `--stats`.

I noticed that our memory usage on all three webheads hovers steadily around 1Gb at the time of writing. Our m3.xlarge instances have [15Gb of ram](https://aws.amazon.com/ec2/instance-types/) and I noticed that at the time of writing free tells me we have 14Gb total and we use 1Gb. 
So let's see what those numbers are when we go to a higher number of processes (aka. workers).


jdotpz:
Didn't know this was stalled waiting for an r+.

So here is an r+

PR #3143, opened 2015-12-17
fixes bug 1230925 - Switch from django-compressor to django-pipeline

peterbe:
It works on my laptop at least. 

If you set `DEBUG=false` you need to have run `./manage.py collectstatic`. When you do, you'll notice that it successfully bundles and compresses assets and give them cool names like `/static/css/crashstats-base.min.949336cd20ab.css`.

### DIFF
--- a/docs/development/dev-services.rst
+++ b/docs/development/dev-services.rst
@@ -203,10 +203,6 @@ Install dependencies
   sudo pip-2.7 install docutils
   brew install mercurial
 
-Install lessc
-::
-  sudo npm install -g less
-
 Set your PATH
 ::
   export PATH=/usr/local/bin:/usr/local/sbin:$PATH
--- a/webapp-django/crashstats/crashstats/static/crashstats/css/screen.less
+++ b/webapp-django/crashstats/crashstats/static/crashstats/css/screen.less
@@ -430,17 +430,9 @@ html.production {
             }
             th.sortable {
                 padding-right: 17px;
-                background: url(../../img/2.0/sort-none.png) no-repeat right center;
             }
             th.sortable:hover {
                 cursor: pointer;
-                background-image: url(../../img/2.0/sort-none-hover.png);
-            }
-            th.sortable.sort-desc {
-                background-image: url(../../img/2.0/sort-desc.png);
-            }
-            th.sortable.sort-asc {
-                background-image: url(../../img/2.0/sort-asc.png);
             }
         }
         tbody {
@@ -717,7 +709,7 @@ div.admin {
 }
 #close_simplebox {
     position:absolute;
-    background:transparent url(/img/icons/close_round.png) 0 0 no-repeat;
+    background:transparent url(../../img/icons/close_round.png) 0 0 no-repeat;
     margin-top:-20px;
     border:0;
     width:30px;
@@ -977,7 +969,7 @@ a.correlation-toggler {
     width: 40px
 }
 a.open-external {
-    background: url(/img/icons/external.png) no-repeat scroll left top transparent;
+    background: url(../../img/icons/external.png) no-repeat scroll left top transparent;
     color: #0A2533;
     margin-left: 20px;
     padding: 2px 0 2px 20px;
--- a/webapp-django/crashstats/settings/base.py
+++ b/webapp-django/crashstats/settings/base.py
@@ -9,6 +9,8 @@ from pkg_resources import resource_string
 import dj_database_url
 from decouple import config, Csv
 
+from bundles import PIPELINE_CSS, PIPELINE_JS  # NOQA
+
 
 ROOT = os.path.abspath(
     os.path.join(
@@ -60,10 +62,10 @@ PROJECT_MODULE = 'crashstats'
 ROOT_URLCONF = '%s.urls' % PROJECT_MODULE
 
 INSTALLED_APPS = (
-    'compressor',
+    'pipeline',
     'django_browserid',
-    'django.contrib.auth',
     'django.contrib.contenttypes',
+    'django.contrib.auth',
     'django.contrib.sessions',
     'django.contrib.staticfiles',
     'commonware.response.cookies',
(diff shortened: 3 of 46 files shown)


docs/development/dev-services.rst
```diff
@@ -203,9 +203,9 @@ Install dependencies
   sudo pip-2.7 install docutils
   brew install mercurial
 
-Install lessc
+Install node dependencies
 ::
-  sudo npm install -g less
+  npm install less
```
adngdb: Since we now have a ``package.json`` and the install is part of ``bootstrap.sh``, I suspect this can actually be entirely removed. As it stands, it will install it the root directory and not it ``./webapp-django``, so I don't think it will be any help.

webapp-django/crashstats/settings/base.py
```diff
@@ -604,24 +587,32 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
...
-COMPRESS_CSS_FILTERS = (
-    'compressor.filters.css_default.CssAbsoluteFilter',
-    'compressor.filters.cssmin.CSSMinFilter'
-)
+#
+# COMPRESS_ROOT = STATIC_ROOT
+# COMPRESS_CSS_FILTERS = (
+#     'compressor.filters.css_default.CssAbsoluteFilter',
+#     'compressor.filters.cssmin.CSSMinFilter'
+# )
```
adngdb: Why are those lines commented? Should they be removed instead?

webapp-django/crashstats/settings/bundles.py
```diff
@@ -0,0 +1,468 @@
...
+    'tokens': {
+        'source_filenames': (
+            'tokens/js/home.js',
+        ),
+        'output_filename': 'js/tokens.min.js',
+    },
+}
+
+# This is sanity checks, primarily for developers. It checks that
+# you haven't haven't accidentally make a string a tuple with an
```
adngdb: "you haven't <strike>haven't</strike> accidentally <strike>make</strike>made a string a tuple with an"

webapp-django/crashstats/settings/bundles.py
```diff
@@ -0,0 +1,468 @@
...
+# excess comma, no underscores in the bundle name and that the
+# bundle file extension is either .js or .css.
+# We also check, but only warn, if a file is re-used in a different bundle.
+# That's because you might want to consider not including that file in the
+# bundle and instead break it out so it can be re-used on its own.
```
adngdb: Oh wow, that is super cool!

webapp-django/crashstats/settings/bundles.py
```diff
@@ -0,0 +1,468 @@
...
+    for k, v in config.items():
+        assert isinstance(k, basestring), k
+        out = v['output_filename']
+        assert isinstance(v['source_filenames'], tuple), v
+        assert isinstance(out, basestring), v
+        assert '_' not in out
+        assert out.endswith('.min.css') or out.endswith('.min.js')
+        for asset_file in v['source_filenames']:
+            if asset_file in _used:
+                # Consider using warnings.warn here instead
```
adngdb: Yeah, I was going to make the same comment, using ``print`` seems wrong, ``logger`` seems better.

webapp-django/crashstats/settings/bundles.py
```diff
@@ -0,0 +1,468 @@
...
+    for asset_file in _trouble:
+        print "REPEATED", asset_file
+        found_in = []
+        sets = []
+        for k, v in config.items():
+            if asset_file in v['source_filenames']:
+                found_in.append(k)
+                sets.append(set(list(v['source_filenames'])))
+        print "FOUND IN", found_in
+        print "ALWAYS TOGETHER WITH", set.intersection(*sets)
```
adngdb: And same for those prints.

webapp-django/crashstats/crashstats/static/crashstats/css/screen.less
```diff
@@ -430,17 +430,9 @@ html.production {
...
             }
             th.sortable:hover {
                 cursor: pointer;
-                background-image: url(../../img/2.0/sort-none-hover.png);
-            }
-            th.sortable.sort-desc {
-                background-image: url(../../img/2.0/sort-desc.png);
-            }
-            th.sortable.sort-asc {
-                background-image: url(../../img/2.0/sort-asc.png);
```
adngdb: Can you explain this change? Why do you have to change the css? Plus it seems to me that you are removing "features". Unless those rules are never used?

docs/development/dev-services.rst
peterbe: There are two bootstrap.sh

The one in `webapp-django/bin/bootstrap.sh` will run the `npm install` in the `webapp-django` directory.

peterbe: But you're right. We don't need to explictly say these instructions because `webapp-django/bin/bootstrap.sh` *does* indeed take care of that.

webapp-django/crashstats/crashstats/static/crashstats/css/screen.less
peterbe: pipeline is much stricter. These files don't exist. The paths don't work but there simply is no `sort-none.png` file anywhere so it's not a matter of the path. 

If you reference images that can't be found `./manage.py collectstatic` fails hard. 

If the files didn't exist before, I'm sure no feature would be affected.

webapp-django/crashstats/settings/bundles.py
peterbe: The whole point of this stuff is to help developers hacking on the code. E.g. you and me. Thus isn't never ever going to be run in production or in run-time. So the ideal place for it is `stdout`. No doubt. 

However, normally in these kinds of cases you use `warnings.warn()` which, by default, yields to `stdout`. 
But, the benefit of doing...:
```python
import warnings
warnings.warn('message')
```
is that that message will only **once** be printed on `stdout`. However, that's only true once the python process has started and kept running. But when you do things like `./manage.py runserver` and change a file it will **restart** the python process each time. The same is true if you run the test repeatedly. It restarts the python process. So we're not getting the benefit of warnings not being repeated. 

Lastly, using `warnings.warn` (or `logger.info` for that matter) is much more verbose. With `warnings.warn('message')` you get...:
```
/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/settings/bundles.py:459: UserWarning: message
  warnings.warn('message')
```
That's a lot noisier than just doing `print 'message'`. 

One thing we should consider is perhaps not bother with ANY of these if you're NOT in debug mode.

webapp-django/crashstats/crashstats/static/crashstats/css/screen.less
adngdb: Ah, great then! It's good that we remove broken rules!

webapp-django/crashstats/settings/bundles.py
adngdb: You're saying this won't be run in prod, and I kinda get it, because I suppose this runs only when you compress / bundle the files, so for prod that happens only during the build step. But in that case, wouldn't it be a bit better to have all that code wrapped in a ``if settings.DEBUG:`` block? 

Otherwise, I agree on ``print`` over others.

adngdb:
r+

PR #3144, opened 2016-01-06
Fixes bug 1237290 - Avoid using Groovy with Elasticsearch when updati…

adngdb:
…ng a field.

### DIFF
--- a/socorro/external/es/super_search_fields.py
+++ b/socorro/external/es/super_search_fields.py
@@ -153,20 +153,15 @@ class SuperSearchFields(ElasticsearchBase):
             if key not in kwargs:
                 del params[key]
 
-        # Before making the change, make sure it does not break indexing.
-        new_mapping = self.get_mapping(overwrite_mapping=params)
-
-        # Try the mapping. If there is an error, an exception will be raised.
-        # If an exception is raised, the new mapping will be rejected.
-        self.test_mapping(new_mapping)
-
         es_connection = self.get_connection()
+        es_index = self.config.elasticsearch.elasticsearch_default_index
+        es_doc_type = 'supersearch_fields'
 
         # First verify that the field does exist.
         try:
             old_value = es_connection.get(
-                index=self.config.elasticsearch.elasticsearch_default_index,
-                doc_type='supersearch_fields',
+                index=es_index,
+                doc_type=es_doc_type,
                 id=params['name'],
             )['_source']  # Only the actual document is of interest.
         except elasticsearch.exceptions.NotFoundError:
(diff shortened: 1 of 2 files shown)


socorro/external/es/super_search_fields.py
```diff
@@ -176,39 +171,71 @@ def update_field(self, **kwargs):
...
-            # before we do the merge, so that it is entirely overwritten.
-            es_connection.update(
-                index=self.config.elasticsearch.elasticsearch_default_index,
-                doc_type='supersearch_fields',
-                body={'script': 'ctx._source.remove("storage_mapping")'},
-                id=params['name'],
+            # objects in the back-end). We therefore want to perform the merge
+            # ourselves, and remove the field from the database before
+            # re-indexing it.
+            new_doc = old_value.copy()
```
peterbe: Why is this necessary?

peterbe: The only other time `old_value` is used is when you later reference `old_value['storage_mapping']` but surely that's equal to what it was before.

socorro/external/es/super_search_fields.py
```diff
@@ -153,20 +153,15 @@ def update_field(self, **kwargs):
...
-        # Before making the change, make sure it does not break indexing.
-        new_mapping = self.get_mapping(overwrite_mapping=params)
-
-        # Try the mapping. If there is an error, an exception will be raised.
-        # If an exception is raised, the new mapping will be rejected.
-        self.test_mapping(new_mapping)
-
         es_connection = self.get_connection()
+        es_index = self.config.elasticsearch.elasticsearch_default_index
+        es_doc_type = 'supersearch_fields'
```
peterbe: thank you!

socorro/external/es/super_search_fields.py
```diff
@@ -176,39 +171,71 @@ def update_field(self, **kwargs):
                 'be created before it can be updated. ' % params['name']
```
peterbe: It says `'The field "%s"...` but isn't that the ID of the document?

socorro/external/es/super_search_fields.py
```diff
@@ -176,39 +171,71 @@ def update_field(self, **kwargs):
...
             self.config.logger.info(
-                'elasticsearch mapping changed for field "%s", '
+                'Elasticsearch mapping changed for field "%s", '
                 'was "%s", now "%s"',
                 params['name'],
                 old_value['storage_mapping'],
-                params['storage_mapping'],
+                new_doc['storage_mapping'],
+            )
+        else:
```
peterbe: Why is this not 
```python
(
    'in_database_name' in params
    and params['in_database_name'] != old_value['in_database_name']
)
```

socorro/external/es/super_search_fields.py
```diff
@@ -176,39 +171,71 @@ def update_field(self, **kwargs):
...
+                id=new_doc['name'],
+            )
+            es_connection.index(
+                index=es_index,
+                doc_type=es_doc_type,
+                body=new_doc,
+                id=new_doc['name'],
+                op_type='create',
+                refresh=True,
             )
```
peterbe: Does it not have some sort of reindex method you can use?

peterbe:
I think my brain is still a bit in vacation-mode. So the point of this is that if you change the `storage_mapping` it completely deletes and re-inserts? And if you change the `in_database_name` it verifies the mapping. Right?

The code then makes sense. r+

Sadly, I'm not fully understanding what the `storage_mapping` is. Is this simply why type of the field is? And `in_database_name` is just the name of a key?

socorro/external/es/super_search_fields.py
adngdb: ``params['name']`` is the name of the field (so something like "signature" or "product") and is also used as the ID of the field, since it is unique.

socorro/external/es/super_search_fields.py
adngdb: I don't follow you. This block gets executed only if ``params['storage_mapping']`` is different than ``old_value['storage_mapping']``, so if I do not make a copy here, the later use of old_value will show the new storage mapping.

socorro/external/es/super_search_fields.py
adngdb: No, that's the hole point of this, the "reindex" method would be what is called "update", and it performs a merge that doesn't work as we need. 

Note that since the list of fields is heavily cached in the front-end, this should not lead to any problem (for example, someone querying SuperSearch right in between the ``delete`` and the ``index``).

socorro/external/es/super_search_fields.py
adngdb: Because we want to run that ``update`` every time the storage mapping was not changed,  not just when ``in_database_name`` was changed. 

Basically, what this does is, make a special case when the storage mapping is changed where we do the "merge" of the old and new documents manually, because we have a very specific use case. In every other case, the ``update`` method just does the job perfectly.

adngdb:
Yeah, there are two things here. The first one is an optimization: we verify the mapping only if there is a change that can impact it, that is when either ``in_database_name`` or ``storage_mapping`` change. 

The second one is what you said, when ``storage_mapping`` is changed we need to delete the existing document and reindex it, so that we have control over the merge of the 2 documents. The reason is that ``storage_mapping`` is a JSON document, and Elasticsearch interprets it as a branch of the document, and thus has a weird update mechanism. 

For example, say we have this document: 
```json
{
  "name": "foo",
  "storage_mapping": {
    "type": "string",
    "analyzer": "keyword"
  }
}
```

Now suppose that we want to change that mapping to this: 
```json
{
  "name": "foo",
  "storage_mapping": {
    "type": "number"
  }
}
```

If we rely on the ``update`` method of Elasticsearch, here is what we get in the end: 
```json
{
  "name": "foo",
  "storage_mapping": {
    "type": "number",
    "analyzer": "keyword"
  }
}
```

The ``analyzer`` key is still there, even though we wanted to remove it. That is why, up to know, we had a groovy script that went in first and removed the content of the ``storage_mapping`` key before we updated the doc. Since we cannot rely on groovy, the solution I found is to delete the document, do the merging ourselves, and reindex the new document.

peterbe:
r+ but you might want to wait with merging till Lars is done with his staging debugging.

PR #3145, opened 2016-01-06
fixes bug 1237424 - Upgrade pika libraries to 0.10.0

peterbe:


jdotpz:
r+ when Travis is happy and we re-clear deploys

PR #3146, opened 2016-01-07
fixes bug 1237666 - Break out pyinotify install for non-linux

peterbe:
This solves two things. 

1. Now you can do `peep install -r requirements.txt` on non-Linux without having to rely on rhelmer's fork on github (which is not on pypi). 
2. rhelmer's fork made it installable on OSX but once installed it would cause problems with Django because Django tries to use it (based on it being import'able) but will of course fail. That's why we can do add [this hack](https://github.com/rhelmer/pyinotify/pull/2) to rhelmer's fork.


adngdb:
Successfully installed ``pyinotify`` on my Linux, r+.

PR #3147, opened 2016-01-07
fixes bug 1237671 - Refactor out funfactory python dependencies we longer needed

peterbe:
Yes, I did upgrade uWSGI and gunicorn. That's because if I didn't they wouldn't compile any more on my machine. 

What `commonware` did for us is...:

* Made sure all cookies are httponly and secure by default. We manage that with `SESSION_SECURE_COOKIE = True` and there's now a new `SESSION_COOKIE_HTTPONLY` setting which Django respects.
* It sets a `X-Frame-Options: DENY` header. (see `curl -v https://crash-stats.mozilla.com >/dev/null`) We can do that with stock Django now.

### DIFF
--- a/webapp-django/crashstats/settings/base.py
+++ b/webapp-django/crashstats/settings/base.py
@@ -68,7 +68,6 @@ INSTALLED_APPS = (
     'django.contrib.auth',
     'django.contrib.sessions',
     'django.contrib.staticfiles',
-    'commonware.response.cookies',
     'django_nose',
     'session_csrf',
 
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/settings/base.py
```diff
@@ -626,6 +625,15 @@ def JINJA_CONFIG():
 # then set this to False
 SESSION_COOKIE_SECURE = config('SESSION_COOKIE_SECURE', True, cast=bool)
```
adngdb: Isn't this useless now that you have removed ``commonware``?

adngdb:
r+

PR #3150, opened 2016-01-11
Bug 1238595 syntaxerror in compressed jquery ui

peterbe:
@adngdb So yuglify has [a bug that can't handle certain jquery files](https://github.com/yui/yuglify/issues/19).  And sadly yuglify uses the old UglifyJS and not UglifyJS2. Switched fixed the problem. 

What `yuglify` was was a wrapper over `cssmin` and (the old) `uglifyjs`. So instead of keeping it, I decided it's better we just use explicitly `cssmin` and (the new) `uglifyjs`. 

Yeah, it's confusing. Compare: [uglify-js](https://www.npmjs.com/package/uglify-js) and [uglify](https://www.npmjs.com/package/uglifyjs). 

Also, [this was confusing](https://twitter.com/peterbe/status/686607276254298113)

### DIFF
--- a/webapp-django/crashstats/settings/bundles.py
+++ b/webapp-django/crashstats/settings/bundles.py
@@ -260,7 +260,7 @@ PIPELINE_JS = {
         'source_filenames': (
             'crashstats/js/socorro/exploitability.js',
         ),
-        'output_filename': 'js/.min.js',
+        'output_filename': 'js/exploitability.min.js',
     },
     'gccrashes': {
         'source_filenames': (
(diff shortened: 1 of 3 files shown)


webapp-django/crashstats/settings/bundles.py
```diff
@@ -260,7 +260,7 @@
         'source_filenames': (
             'crashstats/js/socorro/exploitability.js',
         ),
-        'output_filename': 'js/.min.js',
+        'output_filename': 'js/exploitability.min.js',
```
adngdb: Hah, good catch!

adngdb:
I'm confused, I've been unable to reproduce the problem we saw on stage locally. But this works, and given the link you gave, I believe it will fix the issue. So, let's go with it!

PR #3153, opened 2016-01-13
fixes bug 1238129 - Retire WebappRuntime and WebappRuntimeMobile prod…

peterbe:
…ucts from Socorro

This PR will make it so that we no longer pretend that there are releases called `WebappRuntime` or `WebappRuntimeMobile`.


adngdb:
I can find an occurrence of "WebAppRuntime" here: https://github.com/mozilla/socorro/blob/master/socorro/cron/jobs/fetch_adi_from_hive.py#L88 — should that be removed as well? 

Otherwise, looks good to me. r+

peterbe:
> I can find an occurrence of "WebAppRuntime" here: https://github.com/mozilla/socorro/blob/master/socorro/cron/jobs/fetch_adi_from_hive.py#L88 — should that be removed as well? 

That's fine to keep. It just corrects the name. Plus it's unlikely that we'll get anything in on that any more anyway.

PR #3158, opened 2016-01-15
fixes bug 1239315 - rename "crashes-per-user" to "crashes-per-day"

peterbe:


webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -2322,6 +2322,19 @@ def mocked_supersearch_get(**params):
         # put it back some something > 0
         models.CurrentProducts.cache_seconds = 60
 
+    def test_crashes_per_user_redirect(self):
+        """At some point in late 2016 we can remove this test."""
```
adngdb: I'm pretty sure we'll be looking at that sometime in 2018 and be like heh.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -908,8 +908,23 @@ def append_row_blob(blob, labels):
...
+    is because then we can't carry the query string.
+    """
+    url = reverse('crashstats:crashes_per_day')
+    if request.META.get('QUERY_STRING'):
+        url += '?{}'.format(request.META['QUERY_STRING'])
+    # At some point later in 2016, we can confidently change this to:
+    # `permanent=True` when we know the redirect is working correctly.
+    # In the transition time, it's safer to use a temporary redirect
+    # since permanent redirects tend to get very stuck in the brower.
+    return redirect(url, permanent=not settings.DEBUG)
```
adngdb: I find it weird to have this redirect as a view instead of directly in the ``urls.py`` file. I'm not sure the flexibility in dev is worth the added complexity.

adngdb:
r+ with a question on the redirect.

peterbe:
Thank you for pushing me to look harder. You can actually get the query string to be part of the redirect directly from `urls.py`. See updated commit. 
I'm going to rebase and land.

m8ttyB:
Issue filed to update the affected tests -- https://github.com/mozilla/Socorro-Tests/issues/346

PR #3160, opened 2016-01-15
fixes bug 1240107 - Expose a /healthcheck endpoint that ELB can hit

peterbe:


peterbe:
@adngdb @jdotpz 
So, note the "XXX" comment in there. I don't know how to make a really basic ES query from the webapp. It would be good to have that, so that when we point Pingdom on this we include testing that the ES server is at least responding. We actually have `settings.SOCORRO_IMPLEMENTATIONS_CONFIG['elasticsearch']['elasticsearch_urls']` available to us and we could use that. 

@adngdb Can you help out with that?

@jdotpz Do you mind waiting till we've had some help from Adrian?

jdotpz:
NP, take all the time you need.  I'll just have to get busy PTO'ing :)

peterbe:
@adngdb I've removed the fs disk check after talking to @jdotpz on irc. 
Also I changed the ES check to be mocked. 
r?

adngdb:
r+

PR #3162, opened 2016-01-20
fixes bug 1230314 - Delete suspicious_crash_signatures table

peterbe:


alembic/versions/9371b45451b_remove_explosiveness_and_suspicious.py
```diff
@@ -0,0 +1,61 @@
...
+    """)
+    op.execute("""
+        CREATE OR REPLACE FUNCTION update_explosiveness(updateday date)
+        RETURNS boolean
+        LANGUAGE plpgsql
+            AS $$
+        BEGIN
+        RETURN TRUE;
+        END; $$;
+    """)
```
adngdb: Why don't you just run the files containing the actual functions? I don't think the idea of this is simply to test, but to keep some sort of way to go back in time. It might not be useful to us, but maybe someone outside Mozilla will be relying on this?

adngdb:
I'm in favor of keeping our alembic migrations correct in both downgrade and upgrade. Otherwise, r+.

alembic/versions/9371b45451b_remove_explosiveness_and_suspicious.py
peterbe: But those files are deleted now.

peterbe:
> I'm in favor of keeping our alembic migrations correct in both downgrade and upgrade. Otherwise, r+. 

Deletion of tables doesn't really have a "downgrade". If you delete a table, it implies that you're also deleting all its content. So it's silly to change your mind. 
We never use the downgrade.

peterbe:
If we decide to want that function back (`update_explosiveness` and `backfill_explosiveness`) we should write a new migration that inserts it again. And we'd resurrect the `.sql` files from git history. *Then* we'd be able to do `load_stored_proc(op, ['update_explosiveness.sql'])`.

peterbe:
Am I being stubborn? Please don't hesitate to say if you feel I'm wrong or you need to chat about it some more. Don't let me steamroll this.

adngdb:
For some reason in my head alembic downgrades where tied to git commits, but that obviously isn't the case. Of course we can't include files that do not exist. Forget my silly comments and go ahead! :)

peterbe:
Did this never pass travis? 
Because now travis is failing with
```
sqlalchemy.exc.ProgrammingError: (ProgrammingError) syntax error at end of input
LINE 3:     
            ^
 '\n        DROP FUNCTION backfill_explosiveness\n    ' {}
```

PR #3164, opened 2016-01-21
Bug 1241343 - Update forced versions for correlations for Firefox cycle starting 2016-01-25

KaiRo-at:
This adds current aurora and nightly versions as well as all planned beta versions for this cycle, and should land on prod early next week (2016-01-25 or closely after).

Note that once again we need to update the crash-analysis node with that code (and make sure this version of the script is actually being run) to actually make it live.

### DIFF
--- a/scripts/crons/cron_libraries.sh
+++ b/scripts/crons/cron_libraries.sh
@@ -69,7 +69,7 @@ do
   techo "Phase 1: end"
 done
 
-MANUAL_VERSION_OVERRIDE="44.0b1 44.0b2 44.0b3 44.0b4 44.0b5 44.0b6 44.0b7 44.0b8 44.0b9 44.0b99 45.0a2 46.0a1"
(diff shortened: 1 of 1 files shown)


peterbe:
If I merge this now, nothing happens. Right?
The real work is in "updating" that `/data/bin/cron_libraries.sh` file on prod crash-analysis. 
The really important thing is that we need to apply this ON the 25th of Jan, so it takes effect on the wee hours of the 26th.

peterbe:
How about we merge this now and book a calendar appointment between you, @jdotpz and myself to go in and do the upgrade however that's best done?

jdotpz:
If we go ahead and move where the cron job is pointing to instead, we can have one person simply run a 'sudo yum install socorro' to update to the current socorro code with the correct versions.

KaiRo-at:
Yes, we can merge this right away only only deploy ideally on Monday.

And also yes, I think we should point the cron job to the socorro version of this script and do this simple deploy then.

peterbe:
r+

peterbe:
@jdotpz, @KaiRo-at and I will ping you on Monday to do that rearrange and yum thing.

PR #3167, opened 2016-01-21
Fixes bug 1237589 - Removed remnants of automatic-emails code from es…

adngdb:
… module.


socorro/unittest/external/es/test_index_creator.py
```diff
@@ -22,19 +20,6 @@ def __init__(self, *args, **kwargs):
 
         self.config = self.get_tuned_config(IndexCreator)
 
-    def tearDown(self):
```
peterbe: Don't you need to also delete the `socorro_test_temp` directory?

socorro/unittest/external/es/test_index_cleaner.py
```diff
@@ -102,17 +102,16 @@ def test_delete_old_indices(self):
...
 
-        assert self.index_client.exists('socorro_emails')
+        assert self.index_client.exists('socorro_test_temp')
 
         api = IndexCleaner(self.config)
         api.delete_old_indices()
 
         # Verify the email index is still there.
-        ok_(self.index_client.exists('socorro_emails'))
+        ok_(self.index_client.exists('socorro_test_temp'))
```
peterbe: I see a `create_index()` but no `delete_index()`. Thus, isn't this unit test leaking?

peterbe:
r+ if you address the problem with leaving a temp index around. It's best practice to not have tests create things that it doesn't delete.

peterbe:
I suggest you rebase against master and let the tests run again.

socorro/unittest/external/es/test_index_creator.py
adngdb: That's done in the base class' ``tearDown`` method.

adngdb: Actually, wait, ``socorro_test_temp`` is not created in this file, so why should it be removed here?

socorro/unittest/external/es/test_index_cleaner.py
adngdb: It is done line 43, using ``self.indices`` as the list of indices to delete.

PR #3172, opened 2016-01-22
Bug 1242024 - Use a regional S3 connection context

peterbe:
CC @twobraids 

Lars, you are the ideal reviewer for this PR but I don't want to assign you out of principle. 
However, the change is small but quite significant. 

The most important change is on line 270 in `socorro/external/boto/crashstorage.py`.
It changes the default to use this new regional connection context class. I don't know if we've specified an override in consul for this but even if we have it's not a huge concern since the old class is still there. 

Another important "trick" is that the default region is `us-west-2` which is what I know we Mozilla use. By setting it as a default we can pursue this new code into prod without having to config change anything. 

By the way, I was able to now get my report index page (using `org.allizom.crash-stats.rhelmer-test.crashes`) to work in my local middleware server. All I had to do was edit the `hbase.calling_format` to `boto.s3.connection.OrdinaryCallingFormat` and it just work. (Why oh why do we still refer to this as hbase?! I hope that can die when we fully deprecate the middleware)

This little PR is approximately a weeks worth of work. With actual real coding just in the last 2 hours of that week.

### DIFF
--- a/socorro/external/boto/connection_context.py
+++ b/socorro/external/boto/connection_context.py
@@ -13,6 +13,8 @@ import boto.exception
 
 from configman import Namespace, RequiredConfig, class_converter
 
+from socorro.lib.converters import change_default
+
 
 class KeyNotFound(Exception):
     pass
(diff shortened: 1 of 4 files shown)


twobraids:
this looks good to me
r+

PR #3173, opened 2016-01-23
nobug -- fixup breakpad bucket name

lonnen:
switch breakpad to using a non-dot bucket name.

I made a hyphenated bucket with the same contents to satisfy the wildcard cert. This will prevent the build from failing on systems that actually validate the cert.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -27,7 +27,7 @@ fi
 
 if [ "`uname -sm`" == "Linux x86_64" ]; then
   # pull pre-built, known version of breakpad
-  wget -N --quiet 'https://s3-us-west-2.amazonaws.com/org.mozilla.breakpad/breakpad.tar.gz'
+  wget -N --quiet 'https://org-mozilla-breakpad.s3-us-west-2.amazonaws.com/breakpad.tar.gz'
   tar -zxf breakpad.tar.gz
   rm -rf stackwalk
   mv breakpad stackwalk


twobraids:
I don't think this is necessary - we've resolved the dotted problem by changing the way we connect to S3,  See PR 3172

lonnen:
They have a common root issue, but this is a different problem affecting our build process. During the build, we wget a precompiled breakpad from a bucket serving static files. Because of the dots in the bucket name the cert looks invalid to wget, which only affects recent versions that validate the cert (vagrant box impacted, build box not yet impacted)

peterbe:
Both versions work. I.e. `wget -N --quiet 'https://s3-us-west-2.amazonaws.com/org.mozilla.breakpad/breakpad.tar.gz'` **and** `wget -N --quiet 'https://org-mozilla-breakpad.s3-us-west-2.amazonaws.com/breakpad.tar.gz'`
In the former you have to specify the region name. In the latter you don't. 

Either way. I don't think this is related to how we're going to do crash storage. It's a baby-step in the right direction. Right now I don't see where there was a problem in the build process because the old way should work (and be secure). 

r+ but would like to know what the problem was.

peterbe:
Upon chatting about this on IRC. The problem is that certain `wget`s have a problem even with `s3-us-west-2.amazonaws.com`. E.g.
```
RROR: certificate common name “*.s3-us-west-2.amazonaws.com” doesn’t match requested host name “s3-us-west-2.amazonaws.com”.
```
By moving to the subdomain calling forward, certain `wget`s will work.

PR #3176, opened 2016-01-27
bug 1223169 - ignore skiplist additions for gfx error logging

KaiRo-at:
This adds CrashStatsLogForwarder::CrashAction and mozilla::gfx::Log to the ignore skiplist, which should give us the behavior Milan requested in the bug.


adngdb:
r+

PR #3178, opened 2016-01-28
Fixes bug 1242983 - Added a mapping for correlations.

adngdb:
This mapping will probably need to be tuned a bit, but at least now you have a foundation on which you can start building your correlations app.

### DIFF
--- a/socorro/external/es/correlations.py
+++ b/socorro/external/es/correlations.py
@@ -1,0 +1,56 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import json
+import os
+
+from configman import Namespace, RequiredConfig
+from configman.converters import class_converter
+
+
+HERE = os.path.dirname(os.path.abspath(__file__))
+
+
+class Correlations(RequiredConfig):
+
+    required_config = Namespace()
+    required_config.add_option(
+        'index_creator',
+        default='socorro.external.es.index_creator.IndexCreator',
+        from_string_converter=class_converter,
+        doc='a class that can create Elasticsearch indices',
+    )
+    required_config.add_option(
+        'elasticsearch_correlations_index_settings',
+        default='%s/mappings/correlations_index_settings.json' % HERE,
+        doc='the file containing the mapping of the indexes receiving '
+            'correlations data',
+    )
+    required_config.add_option(
+        'elasticsearch_correlations_index',
+        default='socorro_correlations_%Y%m',
+        doc='the index that handles data about correlations',
+    )
+
+    def __init__(self, config):
+        super(Correlations, self).__init__()
+        self.config = config
+
+        self.indices_cache = set()
+
+    def get_index_for_date(self, date):
+        return date.strftime(self.config.elasticsearch_correlations_index)
+
+    def create_correlations_index(self, es_index):
+        """Create an index to store correlations. """
+        if es_index not in self.indices_cache:
+            settings_json = open(
+                self.config.elasticsearch_correlations_index_settings
+            ).read()
+            es_settings = json.loads(settings_json)
+
+            index_creator = self.config.index_creator(self.config)
+            index_creator.create_index(es_index, es_settings)
+
+            self.indices_cache.add(es_index)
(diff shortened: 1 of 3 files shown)


socorro/external/es/correlations.py
```diff
@@ -0,0 +1,58 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import json
+import os
+
+from configman import Namespace, RequiredConfig
+from configman.converters import class_converter
+
+
+DIRECTORY = os.path.dirname(os.path.abspath(__file__))
```
peterbe: This is a pretty bad name :)
Perhaps call it `HERE` or `THIS_DIRECTORY`.

socorro/external/es/mappings/correlations_index_settings.json
```diff
@@ -0,0 +1,65 @@
...
+        "correlations": {
+            "_all": {
+                "enabled": false
+            },
+            "_source": {
+                "compress": true
+            },
+            "properties": {
+                "product": {
+                    "type": "string",
```
peterbe: Why isn't this `analyzer: keyword` too? Is it out realistic fear that we might have corrections for "Firefox" AND "firefox"?

socorro/external/es/mappings/correlations_index_settings.json
```diff
@@ -0,0 +1,65 @@
...
+                },
+                "version": {
+                    "type": "string",
+                    "analyzer": "keyword",
+                    "norms" : {
+                        "enabled" : false
+                    }
+                },
+                "platform": {
+                    "type": "string",
```
peterbe: Again, why not `keyword`?

socorro/external/es/mappings/correlations_index_settings.json
```diff
@@ -0,0 +1,65 @@
...
+                    "norms" : {
+                        "enabled" : false
+                    }
+                },
+                "signature": {
+                    "type": "string",
+                    "index": "not_analyzed",
```
peterbe: Pardon my ignorance but what does this do?

socorro/external/es/mappings/correlations_index_settings.json
```diff
@@ -0,0 +1,65 @@
...
+                },
+                "notes": {
+                    "type": "string",
+                    "norms" : {
+                        "enabled" : false
+                    }
+                },
+                "date": {
+                    "type": "date",
+                    "format": "date_time_no_millis"
```
peterbe: I don't even think we'll have a time.

peterbe:
This is excellent! But I think I need a bit more handholding on how to "embed" and actually use this.

peterbe:
Perhaps we can chat about this in more detail in our 1:1

socorro/external/es/mappings/correlations_index_settings.json
```diff
@@ -0,0 +1,65 @@
+{
+    "mappings": {
+        "correlations": {
```
peterbe: note-to-self, this is the "doctype".

socorro/unittest/external/es/test_correlations.py
```diff
@@ -0,0 +1,52 @@
...
+
+        correlations = Correlations(config=self.config)
+        correlations.create_correlations_index(
+            datetime.datetime.utcnow().date()
+        )
+
+        expected_index = today.strftime(ES_CORRELATIONS_INDEX)
+        self.indices.append(expected_index)
+
+        ok_(self.index_client.exists(expected_index))
```
peterbe: Perhaps add another couple of lines like this:
```python
# We should be able to create the index again without errors
correlations.create_correlations_index(
     datetime.datetime.utcnow().date()
)
```

socorro/external/es/mappings/correlations_index_settings.json
adngdb: Yeah, right, this should just be a ``"date"``.

socorro/external/es/correlations.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self, config):
+        super(Correlations, self).__init__()
+        self.config = config
+
+        self.indices_cache = set()
+
+    def get_index_for_date(self, date):
+        return date.strftime(self.config.elasticsearch_correlations_index)
+
+    def create_correlations_index(self, date):
```
peterbe: Can we change this to `(self, ex_index)` and remove like 47 instead. 
That way you can do this:
```python
idx = self.get_index_for_date(somedate)
self.create_correlation_index(idx)
self.connection.index(
    index=idx,
    doctype="correlations",
    ...
```

adngdb:
@peterbe updated!

peterbe:
r+

PR #3179, opened 2016-01-28
fixes bug 1241528 - Drop peep and use pip 8

peterbe:


setup.py
```diff
@@ -39,7 +39,7 @@ def read(fname):
     keywords=['socorro', 'breakpad', 'crash', 'reporting', 'minidump',
               'stacktrace'],
     packages=find_packages(),
-    install_requires=[], # use peep -r requirements.txt instead
+    install_requires=[], # use pip -r requirements.txt instead
```
erikrose: Weird!

scripts/bootstrap.sh
```diff
@@ -12,12 +12,12 @@ fi
...
 # install dev + prod dependencies
-${VIRTUAL_ENV}/bin/pip install tools/peep-2.*tar.gz
-${VIRTUAL_ENV}/bin/peep install --download-cache=./pip-cache -r requirements.txt
+${VIRTUAL_ENV}/bin/python tools/pipstrap.py
+${VIRTUAL_ENV}/bin/pip install --require-hashes -r requirements.txt
 
 # only install `linux-requirements.txt` if you're on linux
 if [[ `uname` == 'Linux' ]]; then
-  ${VIRTUAL_ENV}/bin/peep install --download-cache=./pip-cache -r linux-requirements.txt
+  ${VIRTUAL_ENV}/bin/pip install --require-hashes -r linux-requirements.txt
```
erikrose: [Environment markers](https://www.python.org/dev/peps/pep-0496/) might let you avoid the branch here. They're supported in reqs files from pip 6 onward.

docs/development/python-dependencies.rst
```diff
@@ -40,36 +30,57 @@ Pinning exact versions is important because it makes deployment
...
 Whilst it's a given that you pin the exact version of the package you
 now depend on, that package might have its own dependencies and
 sometimes they're not pinned to specific version. For example,
-`web.py` depends on `somepackage` but doesn't state what exact
+``mypackage`` depends on ``somepackage`` but doesn't state what exact
 version. Therefore, it's your job to predict this before it's
-installed as a nested dependency. So, do this::
+installed as a nested dependency.
+
+The best approach is to inspect the ``setup.py`` of the new package(s)
```
peterbe: If I `somepackage==1.2.3` and its `setup.py` has a `install_requires=['depends==1.0.0']` will pip notice that too?

setup.py
peterbe: Well, how else should you do it? 

In projects past I've made this a list by doing something like `install_requires=open('requirements.txt').read().splitlines()` but that's kinda silly because oftentimes `requirements.txt` has things you don't actually need in runtime.

scripts/bootstrap.sh
peterbe: Cool! I'll play with that. If that works, we'd be able to have just the one `requirements.txt` and let pip decide if certain packages should be avoided.

docs/development/python-dependencies.rst
erikrose: pip will go all the way down the dependency tree. Is that what you're asking?

setup.py
erikrose: For my packages, requirements.txt represents a known-good set of dependencies: the set I've tested with and the one I deploy with. It uses only the == comparator. setup.py's install_requires is a more permissive set of versions, generally using > and <. It's targeted more at people importing my project as a part of their own, for which they might need different exact versions of deps.

erikrose: The really weird thing about [] is that you can't pip install socorro and have a reasonable thing happen.

adngdb: Well, I don't think we want people to be able to ``pip install socorro``, to be honest.

adngdb:
r+

docs/development/python-dependencies.rst
peterbe: I've updated the documentation to say this. I.e. let `pip install` tell you which nested packages you missed to also spec in.

PR #3182, opened 2016-02-03
fixes bug 1237703 - Django 1.8

peterbe:


webapp-django/crashstats/authentication/management/commands/makesuperuser.py
```diff
@@ -10,7 +10,8 @@ def get_input(text):
 
 class Command(BaseCommand):
 
-    help = 'emailaddress [, otheremailaddress, ...]'
+    def add_arguments(self, parser):
+        parser.add_argument('emailaddress', nargs='+', type=str)
```
adngdb: So you require the argument to be passed with the command, but then it still asks you to write it a second time.

webapp-django/crashstats/settings/base.py
```diff
@@ -42,15 +53,9 @@ def path(*dirs):
...
-    'django.template.loaders.filesystem.Loader',
-    'django.template.loaders.app_directories.Loader',
-)
-
-TEMPLATE_DIRS = (
-    path('templates'),
-)
+# TEMPLATE_DIRS = (
+#     path('templates'),
+# )
```
adngdb: Why keep this commented out?

webapp-django/crashstats/settings/base.py
```diff
@@ -160,15 +209,16 @@ def JINJA_CONFIG():
...
-}
+# # This disables all mail_admins on all django.request errors.
+# # We can do this because we use Sentry now instead
+# LOGGING = {
+#     'loggers': {
+#         'django.request': {
+#             'handlers': []
+#         }
+#     }
+# }
```
adngdb: And same here.

webapp-django/crashstats/settings/base.py
```diff
@@ -114,9 +120,9 @@ def JINJA_CONFIG():
 
 # Because Jinja2 is the default template loader, add any non-Jinja templated
 # apps here:
-JINGO_EXCLUDE_APPS = (
-    'browserid',
-)
+# JINGO_EXCLUDE_APPS = (
+#     'browserid',
+# )
```
adngdb: Same here.

adngdb:
A few nits but nothing seemed to be broken during my local tests. r+

webapp-django/crashstats/authentication/management/commands/makesuperuser.py
peterbe: What do you mean?

webapp-django/crashstats/authentication/management/commands/makesuperuser.py
```diff
@@ -10,7 +10,8 @@ def get_input(text):
...
 
     def handle(self, *args, **options):
         if not args:
```
willkg: I think this check is wrong. If you're using the `add_arguments` thing, then it uses argparse, so then the value is in `options['emailaddress']`.

This is the equivalent command in Fjord:

https://github.com/mozilla/fjord/blob/master/fjord/base/management/commands/ihavepower.py

peterbe: Oops. I think I hasted. I'll rectify.

peterbe: https://github.com/mozilla/socorro/pull/3185

PR #3183, opened 2016-02-03
Fixes bug 1245558 - Added failures handling to save_raw_and_processed…

adngdb:
… method of PolyCrashStorage class.

### DIFF
--- a/socorro/external/crashstorage_base.py
+++ b/socorro/external/crashstorage_base.py
@@ -143,15 +143,16 @@ class Redactor(RequiredConfig):
     def redact(self, a_mapping):
         """this is the function that does the redaction."""
         for a_key in self.forbidden_keys:
-            sub_mapping =  a_mapping
-            sub_keys =  a_key.split('.')
+            sub_mapping = a_mapping
+            sub_keys = a_key.split('.')
             try:
                 for a_sub_key in sub_keys[:-1]:  # step through the subkeys
                     sub_mapping = sub_mapping[a_sub_key.strip()]
                 del sub_mapping[sub_keys[-1]]
             except KeyError:
-                pass  # this is okay, our key was already deleted by
-                      # another pattern that matched at a higher level
+                # this is okay, our key was already deleted by
+                # another pattern that matched at a higher level
+                pass
 
     #--------------------------------------------------------------------------
     def __call__(self, a_mapping):
@@ -589,13 +590,30 @@ class PolyCrashStorage(CrashStorageBase):
     #--------------------------------------------------------------------------
     def save_raw_and_processed(self, raw_crash, dump, processed_crash,
                                crash_id):
+        storage_exception = PolyStorageError()
         for a_store in self.stores.itervalues():
-            a_store.save_raw_and_processed(
-              raw_crash,
-              dump,
-              processed_crash,
-              crash_id
-            )
+            self.quit_check()
+            try:
+                a_store.save_raw_and_processed(
+                    raw_crash,
+                    dump,
+                    processed_crash,
+                    crash_id
+                )
+            except Exception:
+                store_class = getattr(
+                    a_store, 'wrapped_object', a_store.__class__
+                )
+
+                self.logger.error(
+                    '%r failed (crash id: %s)',
+                    store_class,
+                    crash_id,
+                    exc_info=True
+                )
+                storage_exception.gather_current_exception()
+        if storage_exception.has_exceptions():
+            raise storage_exception
 
 
 #==============================================================================
@@ -931,7 +949,7 @@ class PrimaryDeferredStorage(CrashStorageBase):
         self.deferred_store = config.deferred.storage_class(
             config.deferred,
             quit_check_callback
-       )
+        )
         self.logger = self.config.logger
 
     #--------------------------------------------------------------------------
--- a/socorro/unittest/external/test_crashstorage_base.py
+++ b/socorro/unittest/external/test_crashstorage_base.py
@@ -110,8 +110,10 @@ class TestBase(TestCase):
                     )
 
             values = ['eins', 'zwei', 'drei']
+
             def open_function(*args, **kwargs):
                 return values.pop(0)
+
             crashstorage = MyCrashStorageTest(
                 config,
                 quit_check_callback=fake_quit_check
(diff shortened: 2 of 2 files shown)


adngdb:
cc @twobraids

socorro/external/crashstorage_base.py
```diff
@@ -589,13 +590,27 @@ def save_processed(self, processed_crash):
...
+        storage_exception = PolyStorageError()
         for a_store in self.stores.itervalues():
-            a_store.save_raw_and_processed(
-              raw_crash,
-              dump,
-              processed_crash,
-              crash_id
-            )
+            self.logger.debug('writing to %s' % (
+                hasattr(a_store, 'wrapped_object') and
```
peterbe: Don't use `hasattr`. https://hynek.me/articles/hasattr/

You can do this line instead with `getattr(a_store, 'wrapped_object', a_store.__class__)`.

By the way, shouldn't it be `a_store.__class__.__name__`? If `a_store` has a `__str__` method, you'd execute that, when in fact, I think you want to log the name of the class.

socorro/external/crashstorage_base.py
```diff
@@ -589,13 +590,27 @@ def save_processed(self, processed_crash):
...
+
+            self.quit_check()
+            try:
+                a_store.save_raw_and_processed(
+                    raw_crash,
+                    dump,
+                    processed_crash,
+                    crash_id
+                )
+            except Exception, x:
```
peterbe: The modern notation is `except Exception as x:` and I'd rather see `exception` or `exp` instead of just `x` as a variable name (...when it's not just inside a list comprehension loop).

socorro/external/crashstorage_base.py
```diff
@@ -589,13 +590,27 @@ def save_processed(self, processed_crash):
...
+                self.logger.error('%s failure: %s', a_store.__class__,
+                                  str(x), exc_info=True)
```
peterbe: Why include the exception object twice? The `exc_info=True` already takes care of turning the exception into a readable string.

peterbe: Also, same point as above about `a_store.__class__` vs `a_store.__class__.__name__`.

peterbe: One thing to consider including in the error message is the crash ID. That's be something interesting to actually use to investigate where it *did* get stored.

socorro/external/crashstorage_base.py
peterbe: Socorro as logging debug on by default in all its syslog. That means that this debug logging will always be sent to syslog in production. Is it worth keeping?

Also, should we switch of debug logging throughout socorro's production systems?

socorro/external/crashstorage_base.py
```diff
@@ -589,13 +590,27 @@ def save_processed(self, processed_crash):
     #--------------------------------------------------------------------------
     def save_raw_and_processed(self, raw_crash, dump, processed_crash,
                                crash_id):
+        storage_exception = PolyStorageError()
```
peterbe: It always bothers me to see a class instance get created without not necessarily being used. It adds noise and makes it read like "We create it, therefore we will need it" when that's not the case. 
I'd venture to do `storage_exception = None` (for the benefit of the last two lines) and then do something like
```python
except Exception:
    if storage_exception is None:
```

socorro/external/crashstorage_base.py
adngdb: This log has been very helpful while debugging yesterday, but yeah, maybe we can do with less noise and just add it whenever we need it. I'll actually keep it in there, but commented out, so that we can easily enable it live.

peterbe: If it's been helpful, you can elevate it to `info` instead. 

I think we're abusing the logging system. A better pattern, in general, is to have lots of `logging.debug(...)` and stuff in the code and some config switch you can flip on to enable it all and when you're done, you flip it back to just `info`. 

So, what I'm saying is, perhaps leave that line as is (it's not a performance penalty if logging is configured to just drop it) and some day better tackle the ability to, easily, switch from regular to debug mode across our servers.

socorro/external/crashstorage_base.py
adngdb: I'm not too sure about this, it would make the code overly complicated to avoid create a single object that doesn't take much memory nor processing time.

peterbe: Sure. Your call. 
I'm just try to share my guttural feedback. More verbose code is sometimes easier to read that code that raises little questions in your head. Like 
```python
print sum(int(num) for xs in l if xs > 50 for num in xs if int(num) < 100)
```
versus
```python
total = 0
for group in groups:
    for number in group:
        number = int(number)
        if number < 100:
            total += number
print total
```

I'm not trying to convince you of anything. Just some loud thinking.

adngdb: Actually, the ``PolyStorageError`` object is always used, with this line: ``if storage_exception.has_exceptions():``. So it does make sense to create it every time, and it would make the code really a lot more complex for almost nothing to make it ``None`` at first. 

So, I'm not going to change this. :)

adngdb:
@peterbe What do you think now?

socorro/external/crashstorage_base.py
```diff
@@ -589,13 +590,28 @@ def save_processed(self, processed_crash):
...
         for a_store in self.stores.itervalues():
-            a_store.save_raw_and_processed(
-              raw_crash,
-              dump,
-              processed_crash,
-              crash_id
+            store_name = getattr(
+                a_store, 'wrapped_object', a_store.__class__
             )
+            # self.logger.debug('writing %s to %s' % (crash_id, store_name))
```
peterbe: debugging left.

socorro/external/crashstorage_base.py
```diff
@@ -589,13 +590,28 @@ def save_processed(self, processed_crash):
...
+            try:
+                a_store.save_raw_and_processed(
+                    raw_crash,
+                    dump,
+                    processed_crash,
+                    crash_id
+                )
+            except Exception:
+                self.logger.error(
+                    '%s failure: %s', store_name, exc_info=True
```
peterbe: excessive `%s`

socorro/external/crashstorage_base.py
peterbe: So, now that you've removed the debug line, the `store_name` variable is not necessarily used. 
That means you can do this line *inside* the `except Exception:` block.

peterbe:
I'm still not convinced about the exception handling. Perhaps I just don't have enough contextual experience with this stuff. If it does happen, that you get an exception, how do you know which ones succeeded? Wouldn't that be quite interesting to log? That `PolyStorageError` object could perhaps take in a list of stores that *did work* so that that can be listed in the final error exception message.

socorro/external/crashstorage_base.py
peterbe: Perhaps also include the `crash_id`. 

Also, what you're doing here is printing the `repr` of the crashstorage class. Not the name. The name is `store_name.__name__`. So, if you intended to write the name, use `.__name__` and/or change the name of the variable from `store_name` to something else. 

Or if you really want the `repr` of the module, which I suspect actually carries more information, you should use `%r` to include it in the logging string (and not bother with the `__name__`).

socorro/external/crashstorage_base.py
peterbe: Great. I didn't see that.

socorro/external/crashstorage_base.py
adngdb: We actually agreed on leaving that line so that live debugging would be easier.

adngdb:
So the thing about the PolyCrashStorage is that all storages that did not fail succeeded. This issue is actually fixing a bug where it was not going to try after the first failure.

socorro/external/crashstorage_base.py
adngdb: Yes, I do want to show the representation. I'm going to change the variable name to say ``repr`` instead of ``name``.

adngdb:
@peterbe updated

socorro/external/crashstorage_base.py
```diff
@@ -589,13 +590,30 @@ def save_processed(self, processed_crash):
...
+            self.quit_check()
+            try:
+                a_store.save_raw_and_processed(
+                    raw_crash,
+                    dump,
+                    processed_crash,
+                    crash_id
+                )
+            except Exception:
+                store_repr = getattr(
```
peterbe: No, it's not `repr` yet. At the moment, this variable is a pointer to a class. It *becomes* repr when you later use the `%r`. I suggest you simply call this variable `store_class` and on line 609 (and line 610) change it to:
```python
self.logger.error(
    '%r failed (crash ID: %s)',
    store_class,
    crash_id,
    exc_info=True
)
```
```

peterbe:
r+ with one last nit on the variable name and the use of `%r` instead of `%s`.

PR #3192, opened 2016-02-09
Bug 1246253 one productversions api to rule them all

peterbe:
@adngdb I know. It's huge. Let me explain...

What you need to play with it start it and use the new `/api/ProductVersions`. It's basically just a tool to read directly from the new class `socorro.external.postgresql.products.ProductVersions`. It works, right?

Next, because this is now the second implementation class that is directly implemented in the webapp, I had to do some surrounding changes...

**1)** Usually the implementation classes return something like this:
```python
things = self.query(sql, sql_params).zipped()
return {
    'hits': things,
    'total': len(things)
}
```
The problem with that is that `.zipped()` return a list of `DotDict` instances. Because `DotDict` is a fancy instance and not just a plain `dict` two very important classes struggled to deal with them. `django.core.cache.backends.locmem.LocMemCache` can't unpickle an object that has itself a method called `__setstate__`. And with `django.core.cache.backends.memcached.MemcachedCache` any attempt to do `cache.set('key', [DotDict({'x': 'y'})], 60)` will *work* but ultimately and silently failing meaning that everything we currently cache on `crashstats.crashstats.models.Bugs` *always* is a cache miss. 
So, because of that I refactored away that fancy use of `DocDict` in `self.query(...).zipped()`. Now, `.zipped()` just returns a list of pure python `dict`s. 
Hence, a hell of a lot of changes to tests related to `socorro.external.postgresql`.

**2)** The list of classes listed whose `.implementation` class attribute had to be replaced with a `mock.MagicMock` was growning. [Remember this?](https://github.com/mozilla/socorro/commit/df1738c950a1b45316bd238f097a17d65a3c69a8#diff-81f36efff7ced5436bb26a38f0c53b2eR33)
So I wrote that stuff in `testbase.py` that uses the `inspect` module. 

**3)** The big change to `001_product_info_view.sql` you may ignore. I just opened the file in Atom and ran the "Beautify Editor" plugin on it which does a half-decent job refactoring the file but doesn't actually change anything. Now the file is somewhat readable. Important because I'll later want to put that query inside `socorro.external.postgresql.products.ProductVersions`. 

**4)** I was tired of seeing `filters = [('somekey', None, ["list"]), ...]` etc. It makes much more sense to let that be real python types instead of strings that describe types. The code (e.g. `modern=False` by default) might be a bit messy right now but given time we should change all places where this happens.

### DIFF
--- a/socorro/external/postgresql/products.py
+++ b/socorro/external/postgresql/products.py
@@ -2,10 +2,14 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import datetime
 import logging
+import warnings
+import itertools
+
 import psycopg2
 
-from socorro.external import BadArgumentError
+from socorro.lib.datetimeutil import string_to_datetime
 from socorro.external.postgresql.base import add_param_to_dict, PostgreSQLBase
 from socorro.lib import datetimeutil, external_common
 from .dbapi2_util import single_row_sql
@@ -14,11 +18,129 @@ from .dbapi2_util import single_row_sql
 logger = logging.getLogger("webapi")
 
 
-class Products(PostgreSQLBase):
+class SmartDate(object):
+
+    def clean(self, value):
+        if any(itertools.imap(value.startswith, ('>=', '<='))):
+            op = value[:2]
+            value = value[2:]
+        elif any(itertools.imap(value.startswith, ('=', '>', '<'))):
+            op = value[:1]
+            value = value[1:]
+        else:
+            op = '='
+        return (op, string_to_datetime(value).date())
+
+
+class ProductVersions(PostgreSQLBase):
+
+    def get(self, **kwargs):
+        filters = [
+            ("version", None, [str]),
+            ("product", None, [str]),
+            ("is_featured", None, bool),
+            ("start_date", None, SmartDate()),
+            ("end_date", None, SmartDate()),
+            ("active", None, bool),
+            ("is_rapid_beta", None, bool),
+        ]
+        params = external_common.parse_arguments(filters, kwargs, modern=True)
+        where = []
+        sql_params = {}
+        for param, value in params.items():
+            if value is None:
+                continue
+            param = {
+                'product': 'product_name',
+                'version': 'version_string',
+                'featured': 'is_featured',
+            }.get(param, param)
+
+            if param == 'active':
+                # This is a convenient exception. It makes it possible
+                # to query for only productversions that are NOT sunset
+                # without having to do any particular date arithmetic.
+                param = 'end_date'
+                operator_ = value and '>=' or '<'
+                value = datetime.datetime.utcnow().date().isoformat()
+            elif isinstance(value, list):
+                operator_ = 'IN'
+                value = tuple(value)
+            elif isinstance(value, tuple):
+                assert len(value) == 2
+                operator_, value = value
+            else:
+                operator_ = '='
+            where.append('{} {} %({})s'.format(
+                param,
+                operator_,
+                param
+            ))
+            sql_params[param] = value
+
+        # rewrite it to a string
+        if where:
+            sql_where = 'WHERE ' + ' AND '.join(where)
+        else:
+            sql_where = ''
+
+        # XXX TODO
+        # RE-write this to use the same SQL select that's used by the
+        # the `product_info` view.
+        sql = """
+            /* socorro.external.postgresql.products.ProductVersions.get */
+            SELECT
+                product_name AS product,
+                version_string AS version,
+                start_date,
+                end_date,
+                throttle::REAL,
+                is_featured,
+                build_type,
+                has_builds,
+                is_rapid_beta
+            FROM product_info
+            {}
+            ORDER BY product_sort, version_sort DESC, channel_sort
+        """.format(sql_where)
+        results = self.query(sql, sql_params).zipped()
+
+        return {
+            'hits': results,
+            'total': len(results),
+        }
+
+    def post(self, **kwargs):
+        """adding a new product"""
+        filters = [
+            ("product", None, "str"),
+            ("version", None, "str"),
+        ]
+        params = external_common.parse_arguments(filters, kwargs)
+        with self.get_connection() as connection:
+            try:
+                result, = single_row_sql(
+                    connection,
+                    "SELECT add_new_product(%s, %s)",
+                    (params['product'], params['version']),
+                )
+            except psycopg2.Error:
+                connection.rollback()
+                return False
+            else:
+                connection.commit()
+            return result
+
+
+class Products(ProductVersions):
 
     def get(self, **kwargs):
         """ Return product information, or version information for one
-         or more product:version combinations """
+        or more product:version combinations """
+        warnings.warn(
+            'This class is deprecated. Use ProductVersions instead.',
+            DeprecationWarning
+        )
         filters = [
             ("versions", None, ["list", "str"]),  # for legacy, to be removed
         ]
@@ -49,21 +171,21 @@ class Products(PostgreSQLBase):
 
         for version in results.zipped():
             try:
-                version.end_date = datetimeutil.date_to_string(
-                    version.end_date
+                version['end_date'] = datetimeutil.date_to_string(
+                    version['end_date']
                 )
             except TypeError:
                 pass
             try:
-                version.start_date = datetimeutil.date_to_string(
-                    version.start_date
+                version['start_date'] = datetimeutil.date_to_string(
+                    version['start_date']
                 )
             except TypeError:
                 pass
 
-            version.throttle = float(version.throttle)
+            version['throttle'] = float(version['throttle'])
 
-            product = version.product
+            product = version['product']
             if product not in products:
                 products.append(product)
 
@@ -104,9 +226,11 @@ class Products(PostgreSQLBase):
             products_list.append(params["products_versions"][x])
             versions_list.append(params["products_versions"][x + 1])
 
-        sql_where = ["(product_name = %(product" + str(x) +
-                     ")s AND version_string = %(version" + str(x) + ")s)"
-                                  for x in range(len(products_list))]
+        sql_where = [
+            "(product_name = %(product" + str(x) +
+            ")s AND version_string = %(version" + str(x) + ")s)"
+            for x in range(len(products_list))
+        ]
 
         sql_params = {}
         sql_params = add_param_to_dict(sql_params, "product", products_list)
@@ -128,11 +252,11 @@ class Products(PostgreSQLBase):
 
         products = []
         for product in results.zipped():
-            product.start_date = datetimeutil.date_to_string(
-                product.start_date
+            product['start_date'] = datetimeutil.date_to_string(
+                product['start_date']
             )
-            product.end_date = datetimeutil.date_to_string(
-                product.end_date
+            product['end_date'] = datetimeutil.date_to_string(
+                product['end_date']
             )
             products.append(product)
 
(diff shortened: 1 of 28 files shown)


peterbe:
This doesn't actually change much. At this point we're only going to get another API class added. (One that is a bit more flexible.)

But I wanted to get started with a PR (it's big already) and we can build from here. The next things to work on are:

1) Replace all use of CurrentProducts, CurrentVersions and ProductsVersions (2 s'es) in the webapp with just calls to ProductVersions (1 s). 
2) Make it so that CurrentProducts, CurrentVersions and ProductsVersions still return the same result they used to return, on the public web API, but based on basically subclassing `ProductVersions`. 
3) Get rid of the `socorro.external.postgresql.products.Products` class. 
4) Stop using the view `product_info` and go straight to the "source". I.e. the table `product_versions`.

peterbe:
One important change that's going on here too is the ability to do things like 
```
curl .../api/MyClass?someboolfield=false
```
That means that if the implementation takes in a boolean, the value of that would become `False` if the key wasn't pretty. Now, with this PR you'll be able to set `somebool=false` and if you don't, i.e. omit it, then it'll arrive in the implementation class as `None`.

peterbe:
@adngdb I'm sorry. This PR is growing and growing. 
I'm really trying to make it stop. But I also want to make a PR that is complete and works. 

The latest big change is that now that I've replaced the hardcoded list of which models to mock the `.implementation` to a dynamic list loaded by `inspect` it broke a bunch of tests where we used `SuperSearchUnredacted` and had only set up mocking for `SuperSearch`.

socorro/external/postgresql/products.py
```diff
@@ -14,11 +18,131 @@
...
+        if any(itertools.imap(value.startswith, ('>=', '<='))):
+            operator_ = value[:2]
+            value = value[2:]
+        elif any(itertools.imap(value.startswith, ('=', '>', '<'))):
+            operator_ = value[:1]
+            value = value[1:]
+        else:
+            operator_ = '='
+        return (operator_, string_to_datetime(value).date())
+
```
peterbe: @adngdb So this is the new cool thing. I changed `external_common.parse_arguments` so that one of the "types" you can pass in can instead be a class instance that has a `clean` method. 
Its job is to return the value (doesn't have to be a tuple) or raise a `ValueError` (or `BadArgumentError`).

socorro/external/postgresql/raw_sql/views/001_product_info_view.sql
```diff
@@ -1,2 +1,22 @@
...
+       release_channels.sort AS channel_sort,
+       product_versions.has_builds,
+       product_versions.is_rapid_beta
+FROM ( ( ( product_versions
+          JOIN product_release_channels ON ( ( ( product_versions.product_name = product_release_channels.product_name )
+                                              AND ( product_versions.build_type = product_release_channels.release_channel ) ) ) )
+        JOIN products ON ( ( product_versions.product_name = products.product_name ) ) )
+      JOIN release_channels ON ( ( product_versions.build_type = release_channels.release_channel ) ) )
+ORDER BY product_versions.product_name,
+         product_versions.version_string;
```
peterbe: @adngdb This file **has not changed**. I just reformatted it so I could better *read* what it's doing. And I thought, why not save that change because there's no good reason for why it should be a file with 2 lines that is 1,000 characters wide.

socorro/external/postgresql/report.py
```diff
@@ -318,7 +318,7 @@ def generate_sql_from(self, params):
         """
         sql_from = ["FROM reports r"]
 
-        ## Searching through plugins
+        # Searching through plugins
```
peterbe: Some pep8 fixes that was the last remaining to make the file fully fixed.

socorro/external/postgresql/products.py
```diff
@@ -14,11 +18,131 @@
 logger = logging.getLogger("webapi")
 
 
-class Products(PostgreSQLBase):
+class SmartDate(object):
+
+    def clean(self, value):
+        if any(itertools.imap(value.startswith, ('>=', '<='))):
+            operator_ = value[:2]
```
adngdb: Why ``operator_``? As far as I know it's not a reserved word in python.

socorro/external/postgresql/products.py
```diff
@@ -14,11 +18,131 @@
...
+        return (operator_, string_to_datetime(value).date())
+
+
+class ProductVersions(PostgreSQLBase):
+
+    def get(self, **kwargs):
+        filters = [
+            ("version", None, [str]),
+            ("product", None, [str]),
+            ("featured", None, bool),
```
adngdb: Why not expose ``is_featured``? That would make it consistent with ``is_rapid_beta`` and probably easier to understand.

socorro/external/postgresql/products.py
```diff
@@ -14,11 +18,131 @@
...
+                'version': 'version_string',
+                'featured': 'is_featured',
+            }.get(param, param)
+
+            if param == 'active':
+                # This is a convenient exception. It makes it possible
+                # to query for only productversions that are NOT sunset
+                # without having to do any particular date arithmetic.
+                param = 'end_date'
+                operator_ = value and '>' or '<='
```
adngdb: ``active``, to me, means that the end date is greater than or equal to today's date, not strictly greater. If the end date is today, it is still active until we hit midnight. 

So I would do ``value and '>=' or '<'``.

adngdb: But maybe you just applied to same logic we have everywhere else, in which case please discard my comment. :)

socorro/external/postgresql/products.py
```diff
@@ -14,11 +18,131 @@
...
+                value = datetime.datetime.utcnow().date().isoformat()
+            elif isinstance(value, list):
+                operator_ = 'IN'
+                value = tuple(value)
+            elif isinstance(value, tuple):
```
adngdb: I find it kind of dangerous to have different behaviors for ``list`` and ``tuple``. To me they should be interchangeable as much as possible, and something being a tuple should not have a specific meaning other than it's immutable. 

I think this is trying to be too smart. It might be fine in this context because you, the developer, know what you wrote, but in the future that might be a cause for bugs.

socorro/external/postgresql/products.py
```diff
@@ -14,11 +18,131 @@
...
+            where.append('{} {} %({})s'.format(
+                param,
+                operator_,
+                param
+            ))
+            sql_params[param] = value
+
+        # rewrite it to a string
+        if where:
+            sql_where = 'WHERE ' + ' AND '.join(where) + '\n'
```
adngdb: That newline is not needed, you already have one in your sql string.

socorro/external/postgresql/products.py
```diff
@@ -14,11 +18,131 @@
...
+                is_featured AS featured,
+                build_type,
+                has_builds,
+                is_rapid_beta
+            FROM product_info
+            {}
+            ORDER BY product_sort, version_sort DESC, channel_sort
+        """.format(sql_where)
+        # print sql
+        # print sql_params
```
adngdb: debug leftovers

socorro/external/postgresql/products.py
```diff
@@ -14,11 +18,131 @@
...
+        results = self.query(sql, sql_params).zipped()
```
adngdb: No error message?

adngdb: ``self.query`` has an ``error_message`` argument that is used to show better errors in case of failure.

socorro/external/postgresql/report.py
```diff
@@ -84,7 +84,7 @@ def get_list(self, **kwargs):
                     '%s is not a recognized sort order key' % sort_order['key']
                 )
             sort_order['direction'] = 'ASC'
-            if 'reverse' in kwargs:
+            if str(kwargs.get('reverse', '')).lower() == 'true':
```
adngdb: That looks like a non-trivial change to the API of report/list/. Why did you make it?

socorro/unittest/external/postgresql/test_dbapi2_util.py
```diff
@@ -195,13 +196,14 @@ def __init__(self, name):
...
         )
-        zipped = sequence.zipped()
+        zipped = sequence.zipped(wrapper=DotDict)
         ok_(isinstance(zipped, list))
         eq_(zipped[0], {'first_name': 'Peter', 'last_name': 'Bengtsson'})
         eq_(zipped[1], {'first_name': 'Lars', 'last_name': 'Lohn'})
 
         # Also, because the zipped list actually is made up for DotDict
         # instances you can access this like a proper ORM thing.
+        # This is thanks to us passing `wrapper=DotDict` above.
```
adngdb: What is the value of this? If in all real life cases you do not use ``DotDict``, then why not just remove that option?

webapp-django/crashstats/api/views.py
```diff
@@ -145,6 +146,29 @@ def __new__(cls, name, bases, attrs):
...
+        for field in self.fields:
+            # Because the context for all of this is the API,
+            # and we're using django forms there's a mismatch to how
+            # boolean fields should be handled.
+            # Django forms are meant for HTML forms. A key principle
+            # functionality of a HTML form and a checkbox is that
+            # if the user choses to NOT check a checkbox, the browser
+            # will not send `mybool=false` or `mybool=''`. It will simply
+            # not send anything and then the server has to assume the
+            # chose to NOT check it because it was offerend.
```
adngdb: Mmm... What? I can't parse that sentence.  :)

socorro/external/postgresql/products.py
peterbe: No, but it's a common builtin module that we might import. However, I think it's ugly like that. I'll change it to `op`.

socorro/external/postgresql/products.py
peterbe: You're right. That does sound better.

socorro/external/postgresql/products.py
peterbe: It's making my brain melt. Oh Friday. 

(I actually wrote a long reply comment here arguing for something, that when I took a closer look, was exactly what you said. Sigh)

socorro/external/postgresql/products.py
peterbe: It's a fair point. I see what you're saying. To me, this was the only way I'd be able to have that `SmartDate` type-checker thing. With it, you'd basically be able to have a single field become two things. E.g. `input='>2016-01-01'`, `output=('>', '2016-01-01')`. 
I don't know how else to solve this.

socorro/external/postgresql/products.py
peterbe: thanks

socorro/external/postgresql/products.py
peterbe: If there's an error I'll consume it with sentry or something. I won't help to get a message like "there was an error". Can you tell I don't agree with the excessively verbose logging done in places? :)

socorro/external/postgresql/report.py
peterbe: Because before, if you do `reverse=false` in the URL this would mean that `'reverse'` would not exist in `kwargs`. Now it does, but it's `False`. What a mess, eh?

socorro/unittest/external/postgresql/test_dbapi2_util.py
peterbe: This was to test that that optional functionality still works. 
You're right, let's remove it.

webapp-django/crashstats/api/views.py
peterbe: Missing the word "user".

socorro/external/postgresql/products.py
adngdb: Haha, fair enough!

socorro/external/postgresql/products.py
adngdb: It seems you changed that name in the return value but not in the filters.

adngdb:
@peterbe rebasing is needed.

adngdb:
r+

PR #3193, opened 2016-02-10
Fixes bug 1247211 - Added a way to try examples in Super Search documentation.

adngdb:


webapp-django/crashstats/documentation/jinja2/documentation/supersearch/examples.html
```diff
@@ -14,7 +14,7 @@ <h2 id="list-of-signatures">List of signatures</h2>
 
     <h3>Solution</h3>
     <div class="example">
-        <pre><code><span class="url-domain">{{ full_url(request, 'api:model_wrapper', model_name='SuperSearch') }}</span><span class="query-string">?product={{ product }}&amp;version={{ releases[product][0].version }}&amp;_facets=signature&amp;_facets_size=20</span></code></pre>
+        <pre><code data-url="{{ full_url(request, 'api:model_wrapper', model_name='SuperSearch') }}?product={{ product }}&amp;version={{ releases[product][0].version }}&amp;_facets=signature&amp;_facets_size=20"><span class="url-domain">{{ full_url(request, 'api:model_wrapper', model_name='SuperSearch') }}</span><span class="query-string">?product={{ product }}&amp;version={{ releases[product][0].version }}&amp;_facets=signature&amp;_facets_size=20</span></code></pre>
```
peterbe: Maybe this `data-url` actually better belongs on the `<div class="example">` block-level element instead.

webapp-django/crashstats/documentation/static/documentation/css/documentation.less
```diff
@@ -153,6 +153,58 @@
...
+            margin: 0;
+            min-width: 0;
+        }
+
+        .results {
+            background-color: black;
+            border-top: 1px solid #999;
+
+            .loader {
+                background-image: url('../../img/ajax-loader16x16.gif');
```
peterbe: This is the loader gif we're supposed to use for things with a white background. That's why it looks a bit odd when it's over the soon-to-be-expanded Results panel thing.

peterbe:
It looks great! (with the exception of the ajax gif)
Would it be possible to change the color scheme to be one with white background?
If not, r+

webapp-django/crashstats/documentation/jinja2/documentation/supersearch/examples.html
adngdb: Nope, there can be several ``<code>`` tags in the same example, see the "Many results" example below for example.

adngdb:
I'm personally much more comfortable with code being on a black background. And I really think that the loader not being the most beautiful isn't really that important... :)

I'm going to refrain merging now since we want to push to prod soon, so I'll let you press the green button whenever you feel like doing it.

peterbe:
Ok. I'll merge right after the next prod push.

PR #3194, opened 2016-02-10
Fixes bug 1247259 - Override styling of code tags in API documentation.

adngdb:


peterbe:
I don't like CSS that exists just to undo other CSS. 
I would much rather we go back to where that CSS is defined (for the purpose of the new User Documentation) I believe and make it specific there. 
CSS that is very specific (i.e. specific CSS selectors) has the advantage that it's not re-used automatically when new HTML becomes available. However, being specific means that you're less likely to accidentally negatively affect other new parts. 

Also, having quite specific CSS selectors makes them easier to debug. That's valuable when the project grows so big that it's impossible for one 1 human to remember all parts.

adngdb:
@peterbe I have taken a new approach based on your comment. r?

peterbe:
Sure. We could have put it in `screen.less` and given it a class but then you have to mess with that every time :)

PR #3195, opened 2016-02-10
Fixes bug 1114591 - Turned remnant of `$` as `starts_with` to the expected `^`.

adngdb:


peterbe:
r+ if Travis is happy

PR #3197, opened 2016-02-12
Add 'TextureUsage' to report metadata whitelist

snorp:


adngdb:
Hi @snorp, we do not handle whitelisting for the RawCrash service in our code anymore, but in our admin panel, via bugs (for example: https://bugzilla.mozilla.org/show_bug.cgi?id=1185400 ). I've added ``TextureUsage`` to the whitelist just now, so you should be all set.

PR #3202, opened 2016-02-16
fixes bug 1248743 - Deprecate all mention of "dataservice"

peterbe:
In prod and stage, I think, we don't use these keys. Instead we use envconsul and the namespaces.


adngdb:
This looks good, but... Should we wait until we have changed our database password before merging this? That way we will see the config failing, if it fails.

peterbe:
Ok. That's a good idea. There's absolutely no rush to land this anyway. Just cleanup.

adngdb:
needs rebasing

webapp-django/crashstats/settings/base.py
```diff
@@ -764,11 +696,11 @@ def path(*dirs):
     'secrets': {
         'postgresql': {
             'database_password': config(
-                'DATASERVICE_DATABASE_PASSWORD',
+                'DATABASE_PASSWORD',
                 'aPassword'
             ),
             'database_username': config(
-                'DATASERVICE_DATABASE_USERNAME',
+                'DATABASE_USERNAME',
```
willkg: Should this be `DATABASE_USER`? I see instances of `DATABASE_USER`, but not `DATABASE_USERNAME` in the socorro and socorro-infra repositories. Would having the key wrong here affect anything?

peterbe: First of all, I don't know. 
There is a bit of a mismatch between the non-django work and the django world. It'd be nice to have a "common nomenclature" but not a must. There are some benefits of it being different too. 

The thing is, I think, thanks to `configman`, the value of this comes in from envconsul and by the fact that there's a config called `secrets.postgresql.database_username`.

adngdb:
No, this is blocked on us changing our stage db password. We had an issue recently where a change required a config change, but we forgot about it because stage uses the default password.

peterbe:
@adngdb Rebased. 

The stage password *has* changed now. I fumbled when I did that. The reason I fumbled was because we still use `DATASERVICE_DATABASE_PASSWORD` but this PR fixes that. Thus there are only 2 configs with the password in it for the webapp.

PR #3203, opened 2016-02-17
Fixes bug 1248928 - Experimental new home page based on the public API.

adngdb:
cc @lonnen 

Let's discuss whether this is a good idea or not. As far as I can tell it has the same features as the current home page, minus a few bugs. It makes 3 HTTP requests to get the data (and 2 more if you change the duration since we only call the ``ProductBuildTypes`` service once) instead of 1 for the current home. The logic is in a single JS file of less than 400 lines with comments (the logic of the current home page is, well, quite complex, with lots of function calls, passing data around... ).

I have not built this with the crashes per day page in mind, so if we want to share some code between those two pages we'll have to do some refactoring.


peterbe:
Remember; we need to solve the case where a product has **no ADI** for any of its versions.

webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -92,6 +92,11 @@ def json_dumps(data):
     return jinja2.Markup(json.dumps(data))
 
 
+@library.filter
+def to_json(data):
+    return json.dumps(data)
```
peterbe: I don't like this. It's dangerous. If you do something like this in a jinja template:
```
<script>var DATA = {{ some_view_python_data | to_json }};</script>
```
**and** somewhere inside that `some_view_python_data` there's a `</script>` it'll actually render. E.g. Suppose that you change your name from "Adrian" to `</script><script>alert('xss')</script>`, then you'd get this:
```html
<script>var DATA = {"name": "</script><script>alert('xss')</script>"};</script>
```
and you have an XSS hole.

peterbe: I know we're not using this `to_json` to in a inline script block. We're using inside a data tag. However, there might be other vulnerabilities to do with data tags. Or perhaps we forget and use it in some dangerous way in some other jinja template.

peterbe: The solution is to do this:
```python
def to_json(data):
    return json.dumps(data).replace('</', '<\\/')
```

peterbe:
Are you aware how different they look?
screen shot 2016-02-22 at 3 48 49 pm
screen shot 2016-02-22 at 3 49 26 pm

peterbe:
By the way, looking at the new graph, it appears like there was no crashes on Feb 21 (which was "yesterday" at the time of writing). But there were.

webapp-django/crashstats/home/static/home/js/home.js
```diff
@@ -0,0 +1,386 @@
...
+    var versions = container.data('versions');
+    var platforms = container.data('platforms');
+    var duration = container.data('duration');
+
+    var pageTitle = $('title').text();
+
+    // Versions ending in -b are "magic" versions that will be split into all
+    // of the beta versions starting with that same version number.
+    var betaVersions = [];
+    for (var i = versions.length - 1; i >= 0; i--) {
```
peterbe: Why not `versions.forEach`?

webapp-django/crashstats/home/static/home/js/home.js
```diff
@@ -0,0 +1,386 @@
...
+            betaVersions.push(versions[i]);
+        }
+    }
+
+    // Set default dates.
+    var endDate = new Date();
+    endDate.setDate(endDate.getDate() - 1);  // We start with yesterday.
+    var startDate = new Date();
+    startDate.setDate(endDate.getDate() - duration);
+
```
peterbe: I was allowed mess with the duration to some number that is NOT one of the duration options
screen shot 2016-02-22 at 4 20 18 pm
Should we not guard against that?

webapp-django/crashstats/home/static/home/js/home.js
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    /**
+     * History change handler. Reloads the graph's data.
+     */
+    window.onpopstate = function (e) {
+        var days = e.state.duration;
+
+        $('#duration .selected').removeClass('selected');
+        $('#duration .days-' + days).addClass('selected');
```
peterbe: Here we should perhaps do something like:
```javascript
if (!$('#duration .days-' + days).length) {
    alert("Invalid duration choice");
}
```

webapp-django/crashstats/home/static/home/js/home.js
```diff
@@ -0,0 +1,386 @@
...
+
+        updateDates(days);
+        loadData();
+    };
+
+    /**
+     * Click event handler for the duration menu. Changes the history and
+     * reloads the graph's data.
+     */
+    $('#duration a').click(function (e) {
```
peterbe: Change this to
```javascript
$('#duration').on('click', 'a', function(e) {
```
That's more performant. It's better to have 1 event handler listening and within it, delegate based on what was clicked.

webapp-django/crashstats/home/static/home/js/home.js
```diff
@@ -0,0 +1,386 @@
...
+            '_histogram.date': 'version',
+            _results_number: 0,
+        };
+
+        var url = superSearchApiUrl + '?' + $.param(params, true);
+        $.ajax({
+            url: url,
+            success: onSuperSearchData,
+            error: onQueryError,
+        });
```
peterbe: Cool!

I'm almost wondering if we should hardcode the value for `superSearchApiUrl`. Since it's a fully exposed thing. A URL we're committed to not changing.

webapp-django/crashstats/home/static/home/js/home.js
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    /**
+     * Get count of average daily installs per day and per version.
+     */
+    function queryAdi() {
```
peterbe: "adi" is not a word. It's an abbreviation. So, shouldn't this be `queryADI`?

webapp-django/crashstats/home/static/home/js/home.js
peterbe: By the way, it has little functional difference in this context but I suggest you use promises instead. 
Unlike basic callbacks they can be chained. So if you change this to:
```
return $.ajax({url: url})
.done(onSuperSearchData)
.fail(onQueryError);
```
Careful if you rely on `this` on this functions. Then you might need to bind the original `this`.

webapp-django/crashstats/home/static/home/js/home.js
```diff
@@ -0,0 +1,386 @@
...
+     * Format data received from SuperSearch to make it simpler and closer to
+     * what we will pass to the graph library.
+     */
+    function onSuperSearchData(data) {
+        var i, ln;
+
+        // Make a simple structure to formalize the data, and make it easy to
+        // handle beta versions.
+        var dataStruct = {};
+        for (i = 0, ln = versions.length; i < ln; i++) {
```
peterbe: Again, why not `versions.forEach`?

webapp-django/crashstats/home/static/home/js/home.js
```diff
@@ -0,0 +1,386 @@
...
+            while (currentDay <= +endDate) {
+                var day = dateIsoFormat(new Date(currentDay));
+                dataStruct[versions[i]][day] = 0;
+                currentDay += 24 * 60 * 60 * 1000;
+            }
+        }
+
+        var dates = data.facets.histogram_date;
+        for (i = 0, ln = dates.length; i < ln; i++) {
+            var date = dates[i];
```
peterbe: ditto

webapp-django/crashstats/home/static/home/js/home.js
```diff
@@ -0,0 +1,386 @@
...
+                    for (var k = betaVersions.length - 1; k >= 0; k--) {
+                        if (version.indexOf(betaVersions[k]) > -1) {
+                            baseVersion = betaVersions[k];
+                        }
+                    }
+                }
+
+                dataStruct[baseVersion][date.term.substring(0, 10)] += versionsCounts[j].count;
+            }
+        }
```
peterbe: I read it but my head's spinning. What does this whole thing do? I know it's about filling up `dataStruct` but some human-readable summary would be nice.

webapp-django/crashstats/home/static/home/js/home.js
```diff
@@ -0,0 +1,386 @@
...
+        container
+            .empty()
+            .append('There was an error processing the request: Status: ' + textStatus + ' Error: ' + errorThrown);
+    }
+
+    /**
+     * Called every time a service has stored some data. When all expected
+     * data is received, proceed with mixing that data and showing the graph.
+     */
+    function onDataComplete() {
```
peterbe: This needs to be entirely replaced with promises and deferreds. jQuery's built-in is quite nice. 
You're currently relying on `dataStore.superSearch` to exist. Which is kinda weird. Unlike Python, an array is always true. Even if there's nothing in there. Same for a dictionary. So if `dataStore.superSearch` is `[]` or `{}` this'll be true. 

I was going to say something stupid like "Check the documentation on $.when". But to back up my feedback, I put together a little test to make sure I understand this myself. 

http://codepen.io/peterbe/pen/wMLLwe?editors=0010

That way you get a promise call back once all the **post-processing** has been done **after** the AJAX has finished, for each AJAX query.

webapp-django/crashstats/home/static/home/js/home.js
```diff
@@ -0,0 +1,386 @@
...
+        if (dataStore.superSearch && dataStore.adi && dataStore.builds) {
+            // All data is there, we can proceed.
+            for (var i = 0, ln = dataStore.superSearch.length; i < ln; i++) {
+                var versionData = dataStore.superSearch[i];
+                var version = versions[i];
+                var invalidData = [];
+                for (var j = 0, lm = versionData.length; j < lm; j++) {
+                    var day = dateIsoFormat(versionData[j].date);
+                    if (typeof(dataStore.adi[day]) === 'undefined' || typeof(dataStore.adi[day][version]) === 'undefined') {
```
peterbe: Why not just `if (!dataStore.adi[day] || !dataStore.adi[day])`?

webapp-django/crashstats/home/static/home/js/home.js
```diff
@@ -0,0 +1,386 @@
+/*global MG: true, socorro: true */
+
+$(function () {
+    'use strict';
+
+    // A polyfill for `endsWith`
+    if (!String.prototype.endsWith) {
```
adngdb: @peterbe What do you think about polyfills in Socorro? I guess you don't mind them since you didn't comment here. But should we maybe centralize them? I could see us having a file called ``polyfills.js`` that we would include on every page, and which would contain all of our polyfills. That way it's easy to maintain by regularly verifying if something can be removed.

webapp-django/crashstats/home/static/home/js/home.js
adngdb: I've read different styles about this. Some say that it should be ``UUID``, others ``Uuid``. I'm not sold on anything, so if you prefer (and already use) the full uppercase for abbreviations, I don't mind switching to that.

peterbe: It's personal preference I guess. Personally I think I prefer `queryURL` or `productADI`.

webapp-django/crashstats/home/static/home/js/home.js
peterbe: Let's avoid globals as much as we possibly can. If a particular page needs a particular polyfill, we put in in that app's static dir. If we find we need them repeatedly, we move it to the `base`'s static dir. 

Global solutions are very hard to undo as the project grows. Exceptions might be jquery or d3 .

adngdb:
@peterbe This now works for products without ADIs. I have added validation for the duration. I have refactored the JS to use promises. And I have fixed the bug about the missing data for the last day of the covered time period.

webapp-django/crashstats/settings/bundles.py
```diff
@@ -228,8 +228,6 @@
             'crashstats/js/jquery/plugins/jquery.cookies.2.2.0.js',
             'crashstats/js/socorro/nav.js',
             'crashstats/js/socorro/analytics.js',
-            'browserid/api.js',
-            'browserid/browserid.js',
```
adngdb: Haha, I can't believe we didn't catch this... :) (I was working on this in a train with no internet connection. )

webapp-django/crashstats/home/static/home/js/home.js
```diff
@@ -0,0 +1,462 @@
...
+            product: product,
+            versions: versions,
+            platforms: platforms,
+            start_date: dateIsoFormat(startDate),
+            end_date: dateIsoFormat(endDate),
+        };
+
+        var url = adiApiUrl + '?' + $.param(params, true);
+        $.ajax({url: url})
+        .then(onADIData)
```
peterbe: uh?! I didn't even know you can do that. I thought `$.ajax` returns a promise that only has `done` and `fail`.

webapp-django/crashstats/home/static/home/js/home.js
```diff
@@ -0,0 +1,462 @@
...
+    /**
+     * Change the dates of the request depending on the duration.
+     */
+    function updateDates() {
+        endDate = new Date();
+        endDate.setDate(endDate.getDate() - 1);  // We start with yesterday.
+
+        startDate = new Date();
+        startDate.setDate(endDate.getDate() - duration);
+    }
```
peterbe: This is not working
http://codepen.io/peterbe/pen/EKaLWJ?editors=1010

For the record, if you look at this a day later and the logic breaks then. The startDate becomes March 22nd, when it's supposed to become FEB 22nd.

peterbe: I suspect it's a leapyear bug https://twitter.com/peterbe/status/704641995147444224

peterbe:
r+ **but** I'd like to view it one last time once you've fixed the date arithmetic bug.

adngdb:
@peterbe forgot to mention it but the date problem has been addressed.

peterbe:
Still r+ :)

We'll launch it gently and test it on stage and prod with @m8ttyB and @KaiRo-at

peterbe:
FYI, @m8ttyB and @KaiRo-at
When Adrian lands this it's going to be https://crash-stats.allizom.org/new_home/products/Firefox  (just `home` replaced by `new_home`)

PR #3204, opened 2016-02-17
Fixes bug 1247834 - Better mapping of _columns to returned fields in …

adngdb:
…Super Search.

### DIFF
--- a/socorro/unittest/external/es/base.py
+++ b/socorro/unittest/external/es/base.py
@@ -719,6 +719,22 @@ SUPERSEARCH_FIELDS = {
             'type': 'long'
         }
     },
+    # Add a synonym field.
+    'product_2': {
+        'data_validation_type': 'enum',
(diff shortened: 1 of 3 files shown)


socorro/unittest/external/es/base.py
```diff
@@ -719,6 +719,22 @@
             'type': 'long'
         }
     },
+    # Add a synonyme field.
```
peterbe: in English I think it's spelled "synonym".

peterbe:
r+ with the nit that I don't like that it's called `self.columns`. It's brief but confusing. I think it should be called `self.requested_columns` or `self.queried_columns` or `self.request_columns`. At the moment it's a bit too vague. 
That's by gut speaking when reading the code. Feel free to override me.

adngdb:
@peterbe I agree on the variable name!

PR #3207, opened 2016-02-19
fixes bug 1249623 - clean query args before calling reverse()

peterbe:
@adngdb I asked @rhelmer to review this because I wanted to land it today.


peterbe:
I took the URL that this happened to on prod and ran it locally, with this patch, and here's how it looks now:
screen shot 2016-02-19 at 10 25 52 am

rhelmer:
r+

peterbe:
@m8ttyB Can you write a new test based on this. I've CC'ed you on the bug. You can see the URL that was used (URL in bug). Using a crazy version string like that should not cause an ISE.

m8ttyB:
@peterbe sure :-) I can whip something together.

PR #3208, opened 2016-02-19
bug 1246253 - pivot productversions

peterbe:
There's a lot of work left to be done. But it works. 

I took some very few ideas from https://github.com/mozilla/socorro/pull/3192 but what you'll notice here is that this change is ALL in the webapp. No new messy class in `socorro/external/postgresql/`.

So the web API continues to work (might need some more testing). You can do things like `http://socorro.dev/api/ProductVersions/?product=Thunderbird`. But you can't add things like `&end_date__lt=2016-02-18` or `&end_date=<2016-02-18`. 

The idea is that any view code that needs the product versions would fetch it just like any other django ORM model. E.g. Instead of ...
```python
from crashstats.crashstats.models import ProductVersions
...
api = ProductVersions()
today = datetime.datetime.utcnow().date()
products = api.get(product='Firefox', end_date='>' + today.isoformat())['hits']
versions = [x['version'] for x in products]
```
We'd do this instead...
```python
from crashstats.legacy.models import ProductVersion
...
today = datetime.datetime.utcnow().date()
products = ProductVersion.objects.filter(product='Firefox', end_date__gte=today)
versions = products.values_list('version', flat=True)
```
So, "inside the web API" we'll have a much richer and better querying system/ORM. On the outside we'll have wrappers for these ORM models that we hack together. 

If you think this is an awesome idea and we go ahead with it, it means that not only can we take out `/products/` from the middleware app, we'd also be able to entirely delete `socorro/external/postgresql/products.py` and `socorro/unittest/external/postgresql/test_products.py`.
(There might be some processor, collector or something like that that relies on these, so we'll delete carefully)

### DIFF
--- a/webapp-django/crashstats/api/views.py
+++ b/webapp-django/crashstats/api/views.py
@@ -119,6 +119,7 @@ TYPE_MAP = {
     datetime.date: forms.DateField,
     datetime.datetime: forms.DateTimeField,
     int: forms.IntegerField,
+    bool: forms.BooleanField,
 }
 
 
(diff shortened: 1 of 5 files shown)


peterbe:
We decided not to use this. We'll keep the idea/technique and play with it in some other less "critical" areas.

PR #3209, opened 2016-02-22
Fixes bug 1248809 - Added js::AutoEnterOOMUnsafeRegion::crash to the …

adngdb:
…prefix list.


peterbe:
I saw some discussion between dbaron and you on IRC. Was that about this one?
It's hard to review. In terms of python, it's [obviously] correct.

peterbe:
:cake:

PR #3212, opened 2016-02-22
Fixes bug 1249805 - Added mozilla::CondVar::.* to the prefix list.

adngdb:
cc @dbaron


peterbe:
This time you included the `.*` but you didn't in https://bugzilla.mozilla.org/attachment.cgi?id=8722017
David's bugs looked the same with regards to thoughts on the matter.

adngdb:
@peterbe yes, because this time I want to catch every function that starts with ``mozilla::CondVar::``. So the regex is ``mozilla::CondVar::.*``. In #3211 I just want to catch that precise function name, so there's no need for a regex.

peterbe:
:tada: Sounds like you've really thought about it and double-checked.

PR #3215, opened 2016-02-24
fixes bug 1250645 - revoke some permissions when groups change

peterbe:
The loop in there is maybe not ultra-performant (for example, fixing the users' permissions for each token) but this operation is extremely rare.


adngdb:
Awesome! r+

PR #3216, opened 2016-02-24
fixes bug 1250695 - always return JSON objects, even on non-200 errors

peterbe:


adngdb:
That looks good, but I think it doesn't solve the initial problem: loading https://crash-stats.mozilla.com/api/SuperSearchUnredacted/ while not logged in will return a 403 error, which is not covered by your patch. 

Maybe we can merge this as a first step and work on actually fixing the bug in a different PR?

peterbe:
Ugh!? I'm surprised. I thought it was a 404 error. Perhaps I had some sort of debug on in my local instance. 

Either way, I think the 403 error should be accompanied with a JSON body in the response. I'll add that functionality to this PR. The work on returning JSON bodies in 404 and 500 cases I think is still good to have around. Although it's poorly attached to the bug that was actually about a 403 error. Sorry about that.

peterbe:
Ahh.. That's much better https://gist.github.com/peterbe/34c3f358042727d585d9

adngdb:
:rocket:

PR #3217, opened 2016-02-24
fixes bug 1251035 - Enable threads in webapp uwsgi workers

peterbe:


rhelmer:
The bug and PR don't explain why - do you need to allow user code (e.g. collector) to create background threads that outlive HTTP requests? Is the collector app configured to do this?

rhelmer:
I am asking because I wonder if you mean `--threads` instead ("run each worker in prethreaded mode with the specified number of threads") - still not sure why you'd want to do this though :)

If that's the goal I'd think one of the async options would outperform threads, and be safer to boot (uWSGI is C code and you're betting that they don't have race conditions or other errors in the code).

peterbe:
First of all, this is not for collector. Only for the webapp. The reason is that the newrelic python agent uses a sub-thread to process what just happened and send that to newrelic's server. It does that in a thread to avoid delaying the wsgi response. 

We SUSPECT that our working newrelic stopped worked when we upgraded the uWSGI version mentioned in requirements.txt. The old one stopped building on OSX so I upgraded. 

I have yet to understand how raven manages to send crashes from the webapp if it's not allowed to spawn threads. Perhaps it sends them blocking'ly.

rhelmer:
@peterbe d'oh, I thought I saw collector for some reason, not webapp. Yes this is the right fix to get newrelic to work correctly: https://docs.newrelic.com/docs/agents/python-agent/hosting-mechanisms/python-agent-uwsgi

Sorry for the noise!

peterbe:
ping @jdotpz

PR #3218, opened 2016-02-25
Fixes bug 1247300 - Simplified throttling rules of collectors.

adngdb:
@peterbe This is a config change. I'm not sure what the testing strategy should be. I can see 2 different paths: 

1. Write a functional test that will send a bunch of crashes into a freshly started collector and verify that what got processed was what we expected. 
2. Test on stage, like we normally do for configuration changes. 

What do you think?

### DIFF
--- a/socorro/collector/throttler.py
+++ b/socorro/collector/throttler.py
@@ -10,6 +10,7 @@ from configman import Namespace, RequiredConfig
 
 from socorro.lib.ver_tools import normalize
 
+
 Compiled_Regular_Expression_Type = type(re.compile(''))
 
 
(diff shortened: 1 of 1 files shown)


adngdb:
Let's go with option 1, write a test for this.

socorro/collector/throttler.py
```diff
@@ -32,20 +33,11 @@ class LegacyThrottler(RequiredConfig):
...
-        ("Version", r'''re.compile(r'\..*?[a-zA-Z]+')''', 100),
-        # 100% of Thunderbird & SeaMonkey
-        ("ProductName", '''lambda x: x[0] in "TSC"''', 100),
-        # reject everything else
-        (None, True, 0)
+        # 10% of Firefox release
+        ("*", '''lambda d: d.get("ProductName") == "Firefox" and
+            d.get("ReleaseChannel") == "release"''', 10),
+        # accept everything else
+        (None, True, 100)
```
twobraids: while I admit that the protection was weak before, this really open Socorro wide for a DOS attack. With this change the collectors will accept any junk submitted and send it on to the processors.

peterbe: What do you propose we do?

I think the original reasoning behind this was the prospect of using Socorro for a new Mozilla product that was being planned/prototyped and they asked about using Breakpad to dump the crashes with us. It was that discussion to caused us to take a second look at the list of rules and they we spun into a discussion about refactoring/simplifying. 

We **could** hard code the names of projects we expect and if a new project wants to start sending us stuff (using breakpad presumably) we change the code, and have it in production within 48 hours. 
Or we could do it in config. But without a good config management system and having to rely on ssh and consulate I strongly oppose.
Or we could do it in a database which we manage with a admin web app. 

By the way, should be chat about DOS attacks in a private channel? I have some more thoughts on the matter.

peterbe: For the record. I too get nervous at the thought of accepting everything.

lonnen:
I worry about this landing until we understand the garbage data coming in a little better. I'm not clear how we would catch it, but I'd love to see a metric tracking processor jobs that are caught in the filter that _are not_ in Firefox Release.

peterbe:
For the record, the discussion about measuring what's going through that throttler, see https://bugzilla.mozilla.org/show_bug.cgi?id=1257360 instead.

lonnen:
This may hang open a while if we wait for 1256360. I'm going to close it for now and we can reopen after.

PR #3219, opened 2016-02-25
Fixes bug 1241512 - Show all requested versions in crashes per day.

adngdb:


peterbe:
It works!
screen shot 2016-02-25 at 9 57 25 am
I took the URL Kairo showed in the bug and tried it locally. Yay!

peterbe:
It's still a little bit weird that the graphs start down on 0. The first number of 44.0b8 is 0.176 (on the 13th Feb) but the graph seems to start at 0 on 12th and go up to 0.176 on the 13th.

peterbe:
Note, in the screenshot, the version dropsdowns don't reflect what I have chosen (in the query string). The primary explanation is that those versions (those in the query string) don't exist as options in the drop-downs. Something we should worry about? I'm inclined to say no.

peterbe:
r+ but if you have headspace to figure out how to make the graph start "a day earlier" that would be nice. Perhaps that's for another day.

adngdb:
@peterbe yeah, that's definitely out of scope, and based on the experience I got from working on the new home page, I can tell that it would not be trivial. I think at some point we can try and rewrite the crashes per day page using the same technique as the new home page.

PR #3222, opened 2016-02-29
[ready] integrate socorrolib into the parent repo

lonnen:
I'm going to use this branch to try and work out getting socorrolib into the main repo. This is using [socorrolib 0.1.1](https://github.com/mozilla/socorrolib/releases/tag/0.1.1) right now, which is pinned to socorro revision 8e9b1ca1c33f385d7c12fc844c7aa60075b33da4. I will update as necessary.

Looking for early feedback on what the integration looks like and when we can consider this ready to land. Right now I'll consider it ready when I can replace all duplicate code from the repo with references to socorrolib in order to avoid ambiguity in where things need to be updated.


lonnen:
I rebased this to remove some of the noise once I got it passing. It has been updated to use socorrolib 2.2.0, which is current with socorro 1c485ce3e6e6b2839834ad8a9f184e407dc8c825 (master at the time I'm posting this)

lonnen:
I rebased this to remove some of the noise once I got it passing. It has been updated to use socorrolib 2.2.0, which is current with socorro 1c485ce3e6e6b2839834ad8a9f184e407dc8c825 (master at the time I'm posting this)

The diff is 'too large to be displayed' which can make this seem intimidating to review. The first four commits represent a find and replace operation or removing redundant code from the library. Then I removed an empty test module. The final two commits are where the real review work lies -- the first relocated `search_common.py` to the search app without changing it in a substantive way, and the second moves `buildutil` to the cron namespace.

peterbe:
There's a merge conflict now. Care to rebase?

alembic/README
```diff
@@ -35,7 +35,7 @@ We use a few types not currently supported by SQLAlchemy, so you may
 need to modify the migration slightly.
 
 
-== Helper functions in socorro.lib.migrations
+== Helper functions in socorrolib.lib.migrations
```
peterbe: Probably a rogue search-and-replace. There is no `socorrolib.lib.migrations` or `socorro.lib.migrations` for that matter.

peterbe: Blah! Ignore me. It's still there.

requirements.txt
```diff
@@ -248,3 +248,5 @@ pyinotify==0.9.6 ; sys_platform == 'linux' or sys_platform == 'linux2' \
     --hash=sha256:9c998a5d7606ca835065cdabc013ae6c66eb9ea76a00a1e3bc6e0cfe2b4f71f4
 django-jinja==2.1.2 \
     --hash=sha256:fc2b6a65201b27711e800eb2e1262ff42a32f027bd2267e34121f1bd2bd6b933
+socorrolib==0.2.2 \
```
peterbe: I kinda wish it was called `socorro-lib`? Then it'd make more sense to build `socorro-external` or `socorro-cron` some day.

setup.py
```diff
@@ -42,7 +42,7 @@ def read(fname):
     install_requires=[], # use pip -r requirements.txt instead
     entry_points={
         'console_scripts': [
-                'socorro = socorro.app.socorro_app:SocorroWelcomeApp.run'
+                'socorro = socorrolib.app.socorro_app:SocorroWelcomeApp.run'
```
peterbe: Apps have always been confusing to me. But it's kinda strange to put apps in a package called "somethinglib". An app is usually the last thing to write when you have libraries and dependencies sorted out. 
Perhaps it's hair-splittingly distracting to obsess about what should be moved and what should stay. But it's almost always a good idea to stop and think.

socorro/processor/hybrid_processor.py
```diff
@@ -20,15 +20,15 @@
...
 from socorro.external.postgresql.dbapi2_util import (
     execute_query_fetchall,
 )
 from socorro.external.postgresql.connection_context import ConnectionContext
-from socorro.lib.transform_rules import TransformRuleSystem
-from socorro.lib.datetimeutil import datetimeFromISOdateString, UTC
-from socorro.lib.ooid import dateFromOoid
-from socorro.lib.util import (
+from socorrolib.lib.transform_rules import TransformRuleSystem
+from socorrolib.lib.datetimeutil import datetimeFromISOdateString, UTC
```
peterbe: We should at some point replace all of this with [arrow](http://crsmithdev.com/arrow/).

peterbe:
The only thing I now object to is that this is called `socorrolib` but it handles the apps. The apps feels very much like something that belongs in socorro (the final put-together thing).

requirements.txt
lonnen: I had the same thought when I started. Python conventions prevent me from doing that, though. Module and package names are lower case letters with underscores allowed sometimes. Putting in hyphens breaks setup().

I suspect -external will dissolve into individual apps, with just a few common things ending up socorrolib. -cron and -database will likely live together in some kind of admin app.

setup.py
lonnen: I haven't moved all apps into that namespace, merely the common ancestor code for the other apps that will be necessary for downstream apps to live in their own repos. It may not survive long as an entry point here, depending on the resolution of the app name shortcut shim compromise. We need it for now to keep automation and operations code working

lonnen:
I've made some compromises to enable splitting out apps I to their own repo later. This necessarily means removing the centralized handling of applications from the Socorro repo. 

Consider - the alternative would be

requirements.txt
peterbe: The repo name can be `socorro-lib` and within there's a python package (what gets installed in `virtualenv/lib/python/site-packages`) called `socorrolib`. E.g. https://github.com/jsocol/django-ratelimit

lonnen:
Whoops, posted mid edit. 

I don't think Socorro will keep a centralized app runner. I could factor the bootstrap and app runner from the base classes to leave it in this namespace, but shortly it will either need to be moved to the common lib or removed completely as apps break out. 

This is not as clean, academically, but it is operationally safer to avoid changing the base class and app runner too much right now.

peterbe:
Ugh. This whole app thing is confusing to me. I never really got it under my skin. I got the concept but that's about it. Therefore, it's a bit blurry how it all ties together. 

At least I think I forced you to stop and think about it one last time. If you think it's good now, feel free to proceed and and land when things work.

peterbe:
:mushroom:

PR #3223, opened 2016-02-29
fixes bug 1246256 - Deprecate ancestors of the new ProductVersions API

peterbe:
[lonnen] - lets wait until peterbe is back from leave to land this, but keep reviewing plz&thnx

SORRY! That this PR is so big. But trust me, I can't think of any other way to make it smaller. 

I've been knee-deep in this code for two days intensively. Basically, what I've done is remove ALL use of `CurrentProducts`, `ProductsVersions` or `CurrentVersions` and replaced it with just 1; `ProductVersions`.

For the general views, there's a `@pass_default_context` which did some messy stuff to get a bunch of `current_versions` and `current_products` and stuff. Now that decorator does just one thing; make sure you have a dict called `active_versions`. This `active_versions` takes this query, `models.ProductVersions(active=True)['hits']` and turns it into a dict so each key is the product name. This thing is re-used heavily. 

Another major difference is that now, there's no more `end_date < datetime.datetime.utcnow()` stuff in for loops. Because it, (almost) everywhere, uses `models.ProductVersions().get(active=True)` there's no old versions sent into the webapp. There is **one** exception for this and that is the monster list of versions and products for the supersearch fields. 

Note; the `/api/CurrentProducts/` and `/api/CurrentVersions/` and `/api/ProductsVersions/` (note the extra s) all continue to work as is. My next task is to decouple them from the middleware/implementations and instead, in their `get` method basically call `models.ProductVersions()` and re-juggle the data so it looks like the old one. Along with that I'll put some deprecation warnings around those now deprecated APIs. 

**UPDATE**

Before, the `CurrentProducts` was used in every view. That would download, from the middleware a **606 Kbytes** JSON blob every time (99% of the time via memcache). (It would then internall chop that up into those that are active and those that are not)

Now, the new equivalent `models.ProductVersions().get(active=True)` gets equally cached but instead it's only **17 Kbytes** JSON blob. That means we don't have to ask memcache for **589 KBytes** on every single rendered view.


.gitignore
```diff
@@ -44,3 +44,4 @@ webapp-django/.env
 # exclude filesystem storage directories
 primaryCrashStore/
 processedCrashStore/
+/tags
```
peterbe: See http://www.peterbe.com/plog/ctags-in-atom-on-osx

This is a common convention file to put in a project root. I just happened to be messing around with this in Atom whilst I was working on this branch. Please forgive :)

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -290,43 +290,44 @@ def mocked_get(url, params, **options):
...
+    #     ok_('NightTrain' in dump)
+    #     versions = dump['WaterWolf']
+    #     version = versions[0]
+    #     ok_('product' in version)
+    #     ok_('version' in version)
+    #     ok_('throttle' in version)
+    #     ok_('start_date' in version)
+    #     ok_('end_date' in version)
+    #     ok_('featured' in version)
+    #     ok_('release' in version)
```
peterbe: These are dying. See main comment on this PR. 

We want to "delete" these endpoints very gracefully. When I do so, very little will change from "the outside". 

The reason I couldn't keep these un-commented is because these tests depend on there being a bunch of cache-priming of `CurrentProducts` in the test base's `setUp`.

webapp-django/crashstats/base/jinja2/crashstats_base.html
```diff
@@ -49,7 +49,7 @@
             <li>
                 <label for="products_select">Product:</label>
                 <select id="products_select">
-                {% for p in releases %}
+                {% for p in active_versions %}
```
peterbe: This, and line 64, are thee most important uses of `ProductVersions`. It drives the nav bar on every page.

webapp-django/crashstats/crashstats/jinja2/crashstats/crashes_per_day.html
```diff
@@ -43,7 +43,7 @@
                         <td>
                             <label class="visually-hidden" for="daily_search_version_form_products">Select Product</label>
                             <select id="daily_search_version_form_products" name="p">
-                            {% for product_name in releases %}
+                            {% for product_name in products %}
```
peterbe: We should avoid the word "releases" and "releases" as much as possible. We should just talk about "product versions" or "products".

webapp-django/crashstats/crashstats/tests/test_middleware.py
```diff
@@ -1,38 +0,0 @@
...
-    def test_propagate_400_error_disabled(self, rget):
-
-        def mocked_get(**options):
-            return Response('Naughty boy!', status_code=400)
-
-        rget.side_effect = mocked_get
-
-        with self.settings(PROPAGATE_MIDDLEWARE_400_ERRORS=False):
-            url = reverse('crashstats:home', args=('WaterWolf',))
-            assert_raises(BadStatusCodeError, self.client.get, url)
```
peterbe: The reason I simply deleted this file was because the `TestPropagate400Errors` depended on that, on every page, there's a call to `requests.get` to the middleware (to get products). That's no longer the case. 

We still have a couple of places where we use the middleware but they're all on their way out.

webapp-django/crashstats/crashstats/tests/test_utils.py
```diff
@@ -1,7 +1,7 @@
 import datetime
 from cStringIO import StringIO
 from unittest import TestCase
-from ordereddict import OrderedDict
+from collections import OrderedDict
```
peterbe: Fly-by-cleanup https://bugzilla.mozilla.org/show_bug.cgi?id=1252220

webapp-django/crashstats/crashstats/views.py
```diff
@@ -421,53 +392,56 @@ def get_product_versions_for_crashes_per_day(facets, product):
...
+def get_build_type_for_product_versions(product_versions):
+    """product_versions is a list that looks something like this:
+      ['Firefox:47.0a1']
+    Return the build type on matches
+    """
+    api = models.ProductVersions()
+    for product_version in product_versions:
+        product, version = product_version.split(':')
+        for hit in api.get(product=product, version=version)['hits']:
+            return hit['build_type']
```
peterbe: Don't worry. This ugly thing is part of a the graphs tab (sometimes) of the old reports list. So it's really not important.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -524,21 +498,36 @@ def home(request, product, versions=None,
...
 
-    context['versions'] = versions
-    context['product'] = product
-    if len(versions) == 1:
-        context['version'] = versions[0]
+        # Kinda strange, but legacy inherited code.
+        # If you look at the homepage without specifying a specific
+        # version, it will always assume that the product does not
+        # contain builds.
+        contains_builds = False
```
peterbe: So this is something we ought to talk about. Especially since you're soon replacing the home page. 

Note!
On prod, with no particular version selected:
screen shot 2016-02-29 at 4 33 08 pm

Now, if you pick a particular version for the home page. One whose `has_builds` column in `product_versions` is `true`
screen shot 2016-02-29 at 4 33 50 pm
(note the extra "Date Range:" widget thing)

peterbe: What my code does is the exact same thing as on existing prod. Just doing it differently with `ProductVersions`.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -2545,7 +2515,7 @@ def gccrashes(request, product, version=None, default_context=None):
         # No version was passed get the latest nightly
         version = get_latest_nightly(context, product)
 
-    current_products = context['currentproducts']['products']
+    current_products = context['active_versions'].keys()
```
peterbe: Because `context['active_versions']` is an `OrderedDict` object, this list of keys will thus forth be in the magical order as dictated by this SQL: https://github.com/mozilla/socorro/blob/master/socorro/external/postgresql/products.py#L104
which ultimately comes from this: https://github.com/mozilla/socorro/blob/master/socorro/external/postgresql/raw_sql/views/001_product_info_view.sql#L12

webapp-django/crashstats/base/jinja2/crashstats_base.html
```diff
@@ -61,22 +61,22 @@
...
-                    {% for release in releases[product] %}
-                        {% if release.featured %}
+                    {% for version in active_versions[product] %}
+                        {% if version.is_featured %}
                         <option
-                            value="{{ release.version }}"
-                            {% if release.version == version %}selected{% endif %}
-                        >{{ release.version }}</option>
+                            value="{{ version.version }}"
+                            {% if version.version == version %}selected{% endif %}
```
adngdb: Huh, that makes little sense... How can it ever be true?

webapp-django/crashstats/base/jinja2/crashstats_base.html
```diff
@@ -61,22 +61,22 @@
...
-                    {% for release in releases[product] %}
-                        {% if not release.featured %}
+                    {% for version in active_versions[product] %}
+                        {% if not product.is_featured %}
                         <option
-                            value="{{ release.version }}"
-                            {% if release.version == version %}selected{% endif %}
-                        >{{ release.version }}</option>
+                            value="{{ version.version }}"
+                            {% if version.version == version %}selected{% endif %}
```
adngdb: Same here. You have variable conflicts.

webapp-django/crashstats/crashstats/utils.py
```diff
@@ -254,41 +254,35 @@ def build_default_context(product=None, versions=None):
...
-        # e.g. like /query
-        if not context.get('product'):
-            context['product'] = settings.DEFAULT_PRODUCT
-    elif product != context.get('product'):
-        raise http.Http404("Not a recognized product")
-
-    if product and versions:
-        raise http.Http404("Not a recognized version for that product")
-
-    context['releases'] = build_releases(context['currentversions'])
```
adngdb: I believe function ``build_releases`` can be removed in this PR.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -421,53 +392,56 @@ def get_product_versions_for_crashes_per_day(facets, product):
...
 def get_latest_nightly(context, product):
     version = None
-    for release in context['currentversions']:
-        if release['product'] == product:
-            rel = release['release']
-            if rel.lower() == 'nightly' and release['featured']:
-                version = release['version']
-                break
+    for version in context['active_versions'][product]:
+        rel = version['build_type']
```
adngdb: I don't understand what ``rel`` means here.

adngdb: Why not call it ``build_type`` or ``channel``?

webapp-django/crashstats/manage/views.py
```diff
@@ -110,38 +109,26 @@ def home(request, default_context=None):
...
+    api = ProductVersions()
+    api.cache_seconds = 0
+    productversions = api.get(active=True)['hits']
 
+    releases = collections.OrderedDict()
+    for pv in productversions:
+        if pv['product'] not in releases:
+            releases[pv['product']] = []
+        releases[pv['product']].append(pv)
+    context['releases'] = releases
```
adngdb: I thought you were against the use of the word ``releases``?

webapp-django/crashstats/supersearch/forms.py
```diff
@@ -33,8 +33,7 @@ class SearchForm(forms.Form):
     def __init__(
         self,
         all_fields,
-        current_products,
-        current_versions,
+        productversions,
```
adngdb: Why not ``product_versions``?

webapp-django/crashstats/supersearch/views.py
```diff
@@ -56,16 +56,14 @@ def get_allowed_fields(user):
 
 
 def get_supersearch_form(request):
-    products = models.ProductsVersions().get()
-    versions = models.CurrentVersions().get()
     platforms = models.Platforms().get()
+    productversions = models.ProductVersions().get()['hits']
```
adngdb: Same here, ``product_versions`` is more pythonic.

adngdb:
As expected the version drop down is broken (current version is never selected), but otherwise it seems to be working fine. I'll review again after you have addressed my comments.

peterbe:
Good catches! I believe I've addressed all the comments you posted.

webapp-django/crashstats/api/tests/test_views.py
adngdb: So why not just remove those lines entirely?

webapp-django/crashstats/crashstats/utils.py
adngdb: You did not answer this. Did you investigate if it could be removed? As far as I can tell nothing is using that function anymore.

adngdb:
When trying to add a new product, I get the same error as for Graphics Devices: 

```
Request Method: POST
Request URL: http://localhost:8000/admin/products/

Traceback:
File "/home/adrian/mozilla/workspace/socorro/socorro-virtualenv/local/lib/python2.7/site-packages/django/core/handlers/base.py" in get_response
  132.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/home/adrian/mozilla/workspace/socorro/webapp-django/crashstats/manage/views.py" in inner
  98.         return view_func(request, *args, **kwargs)
File "/home/adrian/mozilla/workspace/socorro/webapp-django/crashstats/manage/views.py" in products
  706.                 version=form.cleaned_data['initial_version']
File "/home/adrian/mozilla/workspace/socorro/webapp-django/crashstats/crashstats/models.py" in post
  761.         return self.implementation().post(**data)
File "/home/adrian/mozilla/workspace/socorro/socorro/external/postgresql/base.py" in __init__
  46.             self.database = self.context.database_class(

Exception Type: AttributeError at /admin/products/
Exception Value: 'NoneType' object has no attribute 'database_class'
```

webapp-django/crashstats/api/tests/test_views.py
peterbe: Because I'm soon going to uncomment when when I re-write those classes to be "proxies" for `ProductVersions`.

webapp-django/crashstats/crashstats/utils.py
peterbe: You're right. Good catch. It's gone now.

peterbe:
I fixed the `post()` and I was able to actually create a new product. It got stored in various postgres tables on the stage RDS. I manually deleted them afterwards in psql.

adngdb:
Needs rebasing! :)

peterbe:
Rebased up and incorporated the new `home` app. 

Let's please get this landed soon as it's getting big and I don't want to have more conflicts as I'm killing the middleware :)

adngdb:
r+! 

Please merge after I have tagged the next release. :)

PR #3224, opened 2016-03-01
fixes Bug 1252461 allow for optionally saving processed json in PG

twobraids:
add a config parameter to the Socorro PG crashstorage class to optionally save the processed crash.  'save_processed_crash' is a boolean, with this behavior:
    True: the processed crash json is saved to a table PG
    False: the processed crash json is not saved directly [DEFAULT]

this has no effects on the regular 'reports' table.  It only toggles the save behavior on the the json table for processed crashes

This change does not prevent the creation of the processed crash table in PG nor stop socorro from trying to fetch a processed crash from that table.

### DIFF
--- a/socorro/external/postgresql/crashstorage.py
+++ b/socorro/external/postgresql/crashstorage.py
@@ -2,6 +2,16 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+# this file defines two crash storage classes:
+# PostgreSQLBasicCrashStorage & PostgreSQLCrashStorage.
+
+# PostgreSQLBasicCrashStorage defines no 'get' methods and saves processed
+# crashes in a lossey form - picking only certain fields to save in the
+# 'reports' table.
+
+# PostgreSQLCrashStorage is a more complete crashstore. It saves processed
+# crashes completely as json in the 'processed_crashes' table.
+
 import datetime
 import json
 from psycopg2 import ProgrammingError
@@ -24,7 +34,7 @@ from socorro.external.postgresql.dbapi2_util import (
 
 
 #==============================================================================
-class PostgreSQLCrashStorage(CrashStorageBase):
+class PostgreSQLBasicCrashStorage(CrashStorageBase):
     """this implementation of crashstorage saves processed crashes to
     an instance of Postgresql.  It only saves certain key values to the
     partitioned reports table, therefore it is not a source for fetching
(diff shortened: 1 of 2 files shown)


peterbe:
What do you think @adngdb ? 
I like it. I don't like the idea of storing processed crashes in 3 places (S3, ES and PG). (Also, the processed crash in ES isn't really its own thing. Every record *is* thee processed crash)

Every little helps to make PG less needed. 

@twobraids Any idea why the test fails?

twobraids:
I've changed the way that this works to make it less intrusive.  There is no longer a  config parameter to test if it should use the processed crash json table.  Instead there is a second PG crashstorage class that just doesn't do the json table save.    Accepting this PR into the codebase, does not change the behavior of the library.  If you wish to have the new behavior in the processors, then just change the configuration of the processor for the PG destination storage class from ```PostgreSQLCrashStorage``` to ```PostgreSQLBasicCrashStorage```

I chose to make the change this way so that we don't break other users of the library.  I know of at least one instance of Socorro that uses PG as the primary crashstore.  Had this PR changed the behavior, they would have been broken by an upgrade.

peterbe:
Assigning to @adngdb because he's leading the processors related stuff more than me. And I'm planning some away-time. 

THANK YOU @twobraids 
I think with lean'ness comes quality and reliability.

peterbe:
An interesting side-effect of stopping this is that I think the processed crashes stored in PG is the only place where we store non-ISO formatted dates. 
I was looking at the [`JsonDTEncoder` class](https://github.com/mozilla/socorrolib/blob/936b11add97cdd0f9a7921b13fe3d5e696d9d966/socorrolib/lib/datetimeutil.py#L149) and it bothered me that not only does it not support `datetime.date` objects, it also stores the date in a non-ISO format. 

If we can stop using this class we can confidently drop this class from socorrolib and everywhere else.

lonnen:
:r+:

peterbe:
> just change the configuration of the processor for the PG destination storage class from PostgreSQLCrashStorage to PostgreSQLBasicCrashStorage

Do you know what config that is by any chance?

adngdb:
@peterbe in processors, it's the ``destination.storage_classes`` key. You want to replace ``socorro.external.postgresql.crashstorage.PostgreSQLCrashStorage`` with ``socorro.external.postgresql.crashstorage.PostgreSQLBasicCrashStorage``.

peterbe:
@adngdb Can you do it? On stage and prod? Perhaps we should file a new bug for tracking that.

adngdb:
https://bugzilla.mozilla.org/show_bug.cgi?id=1257531

PR #3225, opened 2016-03-01
Fixes bug 1241488 - Added a NewSignatures endpoint to the public API.

adngdb:


webapp-django/crashstats/api/models.py
```diff
@@ -1 +1,88 @@
...
+from crashstats.crashstats import models
+from crashstats.supersearch.models import SuperSearch
+
+
+class NewSignatures(models.SocorroMiddleware):
+
+    API_WHITELIST = None
+
+    possible_params = (
+        'start_date',
```
peterbe: This should be a tuple like `('start_date', datetime.date),` I think.

webapp-django/crashstats/api/models.py
```diff
@@ -1 +1,88 @@
...
+
+        # Now we want to verify whether those signatures appeared or not during
+        # some previous period of time.
+        params['date'] = [
+            '>=' + not_after.isoformat(),
+            '<' + start_date.isoformat(),
+        ]
+
+        # Filter exactly the signatures that we have.
+        params['signature'] = ['=' + x for x in signatures]
```
peterbe: Why isn't it that operation by default?

peterbe:
Code is good but there's elements of this that makes me nervous. 

First of all, it's a bespoke ad-hoc query that one individual asked for and we don't know if that individual "speaks" on behalf of many people who need this feature.  

Secondly, this could should go in a new django app. The `api` one is not about business logic. It's just about exposing other business logic (defined elsewhere). 

Thirdly, this expands the surface area of socorro in regards to features. I know it's just 1 little extra endpoint but there's a certain pattern of progression here that scares me. 

Ideal for this is if this business logic and reporting could be stored as user data. I.e. storing entered code in a database instead of storing it in git. 

Seriously, how hard would it be to store this code in a database and execute it in a safe way? Is that something we'd be interested in investing into? (possibly outside this repo)

webapp-django/crashstats/api/models.py
adngdb: Why should it? I think ``contains`` is a better candidate, and in the SuperSearch world ``has term`` is the *de facto* default for most fields.

adngdb:
@peterbe I talked to @lonnen about this, and he says it's fine as is. We cannot spend too much time working on a perfect system for this, and it's fine to have those little, quite isolated, services. 

I moved the endpoint to a new app called ``tools`` and addressed the issue about validation you pointed out. r?

peterbe:
r+

lonnen:
chiming in to follow up on adrian's comment -- 

Part of the plan has always been to incorporate reports back into crash-stats, though we often talk about it in the framework of spectateur or another self-serve interface. Since we haven't been able to deliver on the interface (shrinking team, production instability, shifting priorities) we should accept a few hardcoded new reports. I can't see right now where we will get back to a spectateur interface in this team, but the data team is working on investigative tools that we can learn from or repurpose to this end.

PR #3226, opened 2016-03-01
fixes bug 1252593 - TransformRuleSystem too eager to swallow Attribut…

peterbe:
…eErrors

We could do some `getattr(a_rule, 'close', None)` with an if statement but generally, it's better to ask for forgiveness than seek permission in python.


twobraids:
looks good to me 
r+

peterbe:
@lonnen Note that this might need to be ported to socorrolib since I think you copied `transform_rules.py` over to it.

PR #3231, opened 2016-03-02
Bug 1252887 - Update forced versions for correlations for Firefox cycle starting 2016-03-07

KaiRo-at:
This adds current aurora and nightly versions as well as all planned beta versions for this cycle, and should land on prod early next week (2016-03-07 or closely after).

Note that once again we need to update the crash-analysis node with that code (and make sure this version of the script is actually being run) to actually make it live.

### DIFF
--- a/scripts/crons/cron_libraries.sh
+++ b/scripts/crons/cron_libraries.sh
@@ -69,7 +69,7 @@ do
   techo "Phase 1: end"
 done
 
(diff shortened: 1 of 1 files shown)


peterbe:
Out of curiosity, why don't we do 47 on correlations? Too few crashes?

peterbe:
r+ 
There's no point waiting for Travis on this.

KaiRo-at:
> Out of curiosity, why don't we do 47 on correlations? Too few crashes?

Not sure what you mean 47.0a2 is in there, and that's where the 47 train will live for the next 6 weeks.

peterbe:
Oops. I searched poorly.

PR #3234, opened 2016-03-07
fixes bug 1252623 - rremove graphics_devices from middleware

peterbe:


adngdb:
Grmbl, rebase needed, but r+ of course.

PR #3236, opened 2016-03-07
fixes bug 1254124 - Crontabber state pure implementation

peterbe:
Note how I took the liberty to delete the `test_carry_mware_error_codes` test. 
That one was about testing how the middleware response HTTP error codes. It had to use something so it used the `CrontabberState` (because it's simple). 

As of this PR, the middleware isn't gone. But it's dying. The `CrontabberState` no longer comes from the middleware and I **could** change the test to instead use some other implementation that's still in `middleware_app` but at this rate, that'd just have to be change and be changed as we continue to dismantle the middleware.

### DIFF
--- a/webapp-django/crashstats/api/tests/test_views.py
+++ b/webapp-django/crashstats/api/tests/test_views.py
@@ -5,6 +5,7 @@ import re
 from django.contrib.auth.models import User, Permission
 from django.core.urlresolvers import reverse
 from django.conf import settings
+from django.utils import timezone
 
 import mock
 import pyquery
@@ -20,7 +21,10 @@ from crashstats.supersearch.tests.common import (
     SUPERSEARCH_FIELDS_MOCKED_RESULTS
 )
 from crashstats.supersearch.models import SuperSearch, SuperSearchUnredacted
-from crashstats.crashstats.models import ProductVersions
+from crashstats.crashstats.models import (
+    ProductVersions,
+    CrontabberState,
+)
 from crashstats.tokens.models import Token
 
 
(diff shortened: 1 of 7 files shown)


adngdb:
needs rebasing

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -1347,69 +1348,6 @@ def mocked_get(url, params, **options):
         ok_(dump['total'])
 
     @mock.patch('requests.get')
-    def test_carry_mware_error_codes(self, rget):
```
adngdb: You know, I think instead of deleting this test, we should replace it with something that makes sure that those same kinds of errors in the implementation will return appropriate errors to the client.

peterbe: Don't we have that already?
Perhaps not. I'll take a look. NB, this is specifically for the web api.

peterbe:
I don't know what happened to your comment but I think it disappeared when I rebased. 

The file already has tests that makes sure that  any `raise BadArgumentError` (or `MissingArgumentError`) is transformed into a `400 Bad Request`. 
See [this test](https://github.com/mozilla/socorro/blob/2bf12c0d668d3fcc394f4840c9bf36eb0ade6946/webapp-django/crashstats/api/tests/test_views.py#L1851-L1858).

peterbe:
Huh! It *was* a bug in GitHub. Now that I've posted my comment your original comment re-appeared. 

Anyway, this is now good to go for a final review.

PR #3238, opened 2016-03-10
1188892 centos

willkg:
Fixes the vagrant environment:

* switch to CentOS 7 (and implicitly python 2.7) and puppet 4
* switch to postgres 9.4
* nix devtools
* add motd

r?


willkg:
With these changes you can do this:

1. `vagrant up` (then wait a while)
2. `vagrant ssh`
3. `cd socorro`
4. `make test`

and the tests pass.

What else should I test the vagrant environment with to make sure it's set up right?

lonnen:
running `make`, even after a successful `make test`, seems to fail on pipstrap:

```
/home/vagrant/socorro/socorro-virtualenv/bin/python tools/pipstrap.py
/home/vagrant/socorro/socorro-virtualenv/bin/python: error while loading shared libraries: libpython2.6.so.1.0: cannot open shared object file: No such file or directory
make: *** [bootstrap] Error 127
```

I'd test to make sure you can access to the webapp after the VM spins up, maybe run a smoke test by grabbing a crash from prod and using submitter app to fire at the VM. Optional tests, both. This seems like an improvement over where it was (totally broken), and we haven't had an out of the box working Vagrant since we moved to S3 as a data store.

It's valid to question the utility of Vagrant at this point.

willkg:
This comment is long, so I added section headers!

# Is vagrant helpful?

I'd like to simplify the dev/prod environment situation, but right now I don't know enough socorro to do that. Unless you or @rhelmer want to decide now, I want to keep working with vagrant and push off the decision until later. Maybe London?

My big reason for fixing vagrant now is that I need a dev environment and having one I can throw away without hosing my host machine is pretty convenient. Fixing the vagrant environment seemed like the shortest path to getting going.

I definitely don't want to spend too much time here right now. I mostly just need something to work with. I think I've got it mostly working now. We/I can fix issues that pop up in future PRs.

# "make dev" then "make" is sad

Going back to your issue, `~/socorro/` is a shared folder and that's the folder vagrant does all the things in, so it spans vagrant environment. I think the `socorro-virtualenv` on your machine is a leftover from the centos 6/python 2.6 based box and thus doesn't work with the centos 7/python 2.7 based box. Running `make test` then `make` on my machine which never used the centos 6/python 2.6 based vagrant box works fine.

Looking at make targets, I would think `make clean` would help here, but it doesn't look like `make clean` removes the `socorro-virtualenv` directory, so I think you'll need to do it manually. Maybe `make clean` should be fixed to remove `socorro-virtualenv`, too?

# other thoughts that are probably out of scope for this PR, but would make a lovely conversation over a cup of coffee some day

I'm not sure I really understand these make targets. Also, it seems like some of them should be idempotent, but it's not clear they are. Further they're not keeping state, so the ones that depend on bootstrap run bootstrap every time. Is that necessary? Some of them seem broken--there's no pylintrc file that I can see, so `make lint` doesn't work right. `make json_enhancements_pg_extension` seems to only apply to postgres 9.2--do we need that anymore? Could it be better marked?

I haven't updated any docs yet figuring that'll be easier and faster to do in a couple of weeks. I am keeping notes, though.

# outstanding

I found a couple of issues while going through make targets and I'll land fixes for those. I also haven't tried the webapp--I'll do that and see if that unearths issues.

lonnen:
I've run this a few times, burning the venv and the whole vagrant box between runs. The shared obejct problem went away, and I suspect you're right about it being a mismatch lingering from an old build.

Currently consistently hitting a migration error during `make test`. Provisions goes off without a hitch, then boom:

```
INFO  [alembic.runtime.migration] Running stamp_revision  -> 9371b45451b
+ PYTHONPATH=.
+ /home/vagrant/socorro/socorro-virtualenv/bin/alembic -c config/alembic.ini downgrade -1
  File "/home/vagrant/socorro/socorro-virtualenv/bin/alembic", line 11, in <module>
    sys.exit(main())
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/alembic/config.py", line 471, in main
    CommandLine(prog=prog).main(argv=argv)
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/alembic/config.py", line 465, in main
    self.run_cmd(cfg, options)
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/alembic/config.py", line 448, in run_cmd
    **dict((k, getattr(options, k)) for k in kwarg)
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/alembic/command.py", line 202, in downgrade
    script.run_env()
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/alembic/script/base.py", line 397, in run_env
    util.load_python_file(self.dir, 'env.py')
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/alembic/util/pyfiles.py", line 81, in load_python_file
    module = load_module_py(module_id, path)
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/alembic/util/compat.py", line 79, in load_module_py
    mod = imp.load_source(module_id, path, fp)
  File "alembic/env.py", line 91, in <module>
    run_migrations_online()
  File "alembic/env.py", line 74, in run_migrations_online
    connection = engine.connect()
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 1778, in connect
    return self._connection_cls(self, **kwargs)
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 60, in __init__
    self.__connection = connection or engine.raw_connection()
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 1847, in raw_connection
    return self.pool.unique_connection()
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/sqlalchemy/pool.py", line 280, in unique_connection
    return _ConnectionFairy._checkout(self)
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/sqlalchemy/pool.py", line 644, in _checkout
    fairy = _ConnectionRecord.checkout(pool)
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/sqlalchemy/pool.py", line 440, in checkout
    rec = pool._do_get()
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/sqlalchemy/pool.py", line 1057, in _do_get
    return self._create_connection()
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/sqlalchemy/pool.py", line 285, in _create_connection
    return _ConnectionRecord(self)
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/sqlalchemy/pool.py", line 411, in __init__
    self.connection = self.__connect()
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/sqlalchemy/pool.py", line 538, in __connect
    connection = self.__pool._creator()
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/sqlalchemy/engine/strategies.py", line 96, in connect
    connection_invalidated=invalidated
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/sqlalchemy/util/compat.py", line 199, in raise_from_cause
    reraise(type(exception), exception, tb=exc_tb)
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/sqlalchemy/engine/strategies.py", line 90, in connect
    return dialect.connect(*cargs, **cparams)
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/sqlalchemy/engine/default.py", line 377, in connect
    return self.dbapi.connect(*cargs, **cparams)
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/psycopg2/__init__.py", line 164, in connect
    conn = _connect(dsn, connection_factory=connection_factory, async=async)
sqlalchemy.exc.OperationalError: (OperationalError) FATAL:  role "vagrant" does not exist
 None None
make: *** [test] Error 1
```
Dev/Prod parity isn't great. We've swung back and forth on how much to enforce it, with varying levels and subsystems being in and outside the repo over the years. It seems like once a year we re-optimize for developer ease (onboarding someone) and parity with prod (something blew up, react by beefing up local dev envs). I truly think this won't resolve until we decompose the monorepo, since now we already tend to run only a subset of the system and hack on the part we need to get our thing working.

The build systems and Make targets in particular have been prone to this muddying [[1](https://github.com/mozilla/socorro/commit/41acdb025578e0f71ee677eab84c32061bba9ab2)]. It's not a strong criticism of individual devs; the setup is a lot to load into one brain. 

We've been using a mix of bash and Make for years because Make is different enough from bash to introduce a bunch of subtle, unexpected behavior when we put everything in there but the relationship between targets was more complex than we wanted to manage with bash only (especially with needing to build stackwalker). Over time Make expanded and we did a cleanup a few ears ago to rely mostly on bash script. See [[2](https://github.com/mozilla/socorro/blob/be3102a8f9f88a9e3bbf2434a1cfccaa93229d51/Makefile)] for what it used to look like.

At the same time I was pushing a convention I borrowed from a GH, where regardless of language I wanted to be able to use:
```
./script/bootstrap # set it up
./script/cibuild   # bootstrap + run tests, capture what used to be in a jenkins textfield
```
We used it for the django rewrite and it infected the rest of the repo when we folded that rewrite into the monorepo. For several years we were using it on anything new I was involved in. It's nice for when you switch languages with every project so you don't have to remember every build tool and syntax, and it was much simpler than Make. Not sure I'd stick with it still, but I digress. Here It kind of got away from me as we had to merge the build systems when we put django back into the monorepo.

lonnen:
To get to that error:

    vagrant destroy # Y
    vagrant up && say 'done' && vagrant ssh
    cd socorro
    make test

rhelmer:
@lonnen iirc vagrant only does the steps up to http://socorro.readthedocs.org/en/latest/development/build-from-source.html#add-a-new-superuse3r-account-to-postgresql - you need to add a superuser account to PG and either do:

```
make test database_username=test database_password=aPassword
```

Or, if you name the superuser account `vagrant` then a simple `make test` should work.

rhelmer:
@lonnen oh, I am wrong, one of the vagrant puppet manifests is doing this: https://github.com/mozilla/socorro/pull/3238/files#diff-b699bab5c9f73165579a449515ea4e05L129

However that's creating a user `test` not `vagrant`, so you probably either need to change that or do:

```
make test database_username=test database_password=aPassword
```

lonnen:
@rhelmer I included exactly what I was doing to match https://github.com/mozilla/socorro/pull/3238#issuecomment-195067239

There may be some manual steps; I certainly remember some being necessary at some point. There's probably a dev ergonomics piece there where we either do this manual stuff once or check to see if it has been done and warn the user if they need to do stuff. A ~/.profile script or something. Not sure exactly how to raise the need for manual user work.

lonnen:
and, as always, it's best to avoid additional talking monkey work where we can

rhelmer:
Hm yeah I don't see how `made test` worked for @willkg I suspect there is something missing.

Just having puppet create a `vagrant` superuser is probably the easiest thing to do, then `make test` will probably work without anything additional.

willkg:
I've got a completely clean slate since I've never done socorro development before. I suspect @lonnen is still hitting non-clean-slate issues.

I think the migrations use `config/alembic.ini` in which case, mine has:

```
sqlalchemy.url = postgresql://test:aPassword@localhost/socorro_migration_test
```

`test` is the role that vagrant creates:

https://github.com/mozilla/socorro/blob/master/puppet/modules/socorro/manifests/init.pp#L125

I'm guessing @lonnen has a different line.

One way to get a completely clean slate is to do this which will **delete all non-tracked files including ignored ones** (don't do this if you have non-tracked-files you want to keep!):

```
$ git clean -fdx
$ vagrant destroy -f
```

That gives you a completely clean slate to build a new vagrant environment from. I don't know offhand if that's what you really want to do, though.

Anyhow, I think that explains why it works for me and not @lonnen.

I'm not sure what to do here. I think a win here will be the minimal amount of work such that "@willkg and @lonnen can get their stuff done". Given that, I think I don't want to do anything to fix this and instead adjust the docs when I do a docs pass.

I know that's not great, but is it ok for now?

peterbe:
I'm not working but I couldn't help but to take a peek. 
I would happily let Will be aggressive and do whatever he wants to make it work. Just like we let intern Trevor revamp our testing scripts. 
Be bold! 

If Vagrant could become stable and something you can run without needing to read documentation I think even I would consider running it.

lonnen:
@willkg you correctly nailed it. My alembic config was statle and since we ignore it to prevent password check ins I wasn't purging it appropriately.

rhelmer:
Oh! That makes sense. Alembic didn't used to allow you to override that URL on the command-line so it was a pain to do from puppet, although I think it has since added that feature, so that might be a way to make this nicer (instead of having to depend on the contents of that `alembic.ini` to be correct and not stale)

Nice work @willkg !

lonnen:
Cool. So I purged all the .pyc files, stale configs, and dead directories, and it passes!

lonnen:
ok -- this is fine to land. it is at least better than what we had

lonnen:
:+1:

willkg:
Thank you!

I think a lot of the discussion we had here and on IRC about roles of the vagrant environment, Makefile and scripts is covered in other bugs:

* bug 1150797 covers fixing the vagrant puppet scripts so that they install socorro and run the various pieces
* bug 1039752 covers tweaking some things so that `make test` does some of the work that the vagrant puppet scripts are currently doing
* bug 1030241 could be a duplicate of 1150797 depending on how we think about them
* myriad of documentation bugs

I think it's worth talking about them in London in a smokey tavern. We could call ourselves The Breaklings.

I'll plan to do a docs pass in the next month which probably covers a few bugs.

PR #3240, opened 2016-03-16
fixes bug 1256949 - ADI pure implementation

peterbe:


socorro/external/postgresql/adi.py
```diff
@@ -63,7 +63,7 @@ def get(self, **kwargs):
 
         sql = """
             SELECT
-                SUM(adu_count) AS adi_count,
+                SUM(adu_count)::BIGINT AS adi_count,
```
peterbe: If we don't do this, the `adi_count` summation will become a `decimal.Decimal` object when piped via pyscopg2.

adngdb:
r+ but you'll need to rebase.

PR #3242, opened 2016-03-17
Bug 1253360 - Added a semicolon_keywords analyzer.

adngdb:


socorro/external/es/super_search_fields.py
```diff
@@ -14,6 +14,19 @@
 from socorrolib.lib import datetimeutil, external_common
 
 
+ES_CUSTOM_ANALYZERS = {
+    "analyzer": {
+        "semicolon_keywords": {
+            "type": "pattern",
+            "pattern": ";"
+        }
```
peterbe: call me silly but there's a mix of `"` and `'` here. This one uses `"` the `ES_QUERY_SETTINGS` uses `'`.

peterbe: actually. I don't care. but I noticed sometimes you tidy these things up so perhaps you would appreciate the finding.

socorro/unittest/external/es/base.py
```diff
@@ -719,6 +721,24 @@
...
+        'is_exposed': True,
+        'is_mandatory': False,
+        'is_returned': True,
+        'name': 'app_init_dlls',
+        'namespace': 'processed_crash',
+        'permissions_needed': [],
+        'query_type': 'string',
+        'storage_mapping': {
+            'type': 'string',
+            'analyzer': 'semicolon_keywords',
```
peterbe: It's amazing that we've not needed a custom analyzer for anything yet.

socorro/unittest/external/es/base.py
```diff
@@ -774,6 +794,15 @@ def test_with_version(self):
     return decorated
 
 
+class SuperSearchWithFields(SuperSearch):
```
peterbe: I'm not convinced about this. Every new class comes with a cost of noise. I know this is just a part of tests, but it's part of the base module so it might get reused. 

It might be more of an advanced pattern but at least it's clear what it does when you view it and you don't have to go and look up the class definition:
```pyhon
self.supersearch = SuperSearchWithFields(config=self.config)
self.api = functools.partial(self.supersearch.get, _fields=self.supersearch.get_fields())
```

peterbe: actually, you might not need to do `self.supersearch =` here. Just `supersearch = `

socorro/unittest/external/es/test_analyzers.py
```diff
@@ -0,0 +1,56 @@
...
+        res = self.api.get(
+            app_init_dlls='/path/to/dll',
+            _facets=['app_init_dlls'],
+        )
+        eq_(res['total'], 2)
+        ok_('app_init_dlls' in res['facets'])
+        facet_terms = [x['term'] for x in res['facets']['app_init_dlls']]
+        ok_('/path/to/dll' in facet_terms)
+        ok_('c:\\bar\\boo' in facet_terms)
+        ok_('foo' in facet_terms)
```
peterbe: Great test!

socorro/unittest/external/es/base.py
peterbe: Oh. I see. You just moved this from elsewhere. Anyway, I think it's always good to keep things simple because it's easier to debug and refactor. But you're free to keep it.

peterbe:
What does this `default_field` do? I'm not familiar with that. 

r+ with a questionable nit.

socorro/external/es/super_search_fields.py
adngdb: Yeah, I do appreciate it, since I favor single quotes. :)

socorro/unittest/external/es/base.py
adngdb: I actually think it's simpler to have this class in the ``base.py`` file. To me it's the same it terms of instant readability (I'm not used to the ``functools`` module, so I would have to go read the doc of ``partial`` anyway). And with my class, each file using ``SuperSearch`` only has one import to do instead of 3 (functools, SuperSearch and SuperSearchFields).

adngdb:
``default_field`` is a feature that says, if the user does not specify a field when doing a search, that default field will be used. We actually don't really use that feature, since SuperSearch always associates a field to each value. But I think it was best practice to have that, especially with such big data.

socorro/unittest/external/es/base.py
peterbe: Like I alluded to, it's a personal choice. Like PHP vs Go.
Also, you have to excuse me because I was on a high of reducing
CurrentProducts, CurrentVersions and ProductsVersions down to just one
thing (...but used differently). :)

On Monday, March 21, 2016, Adrian Gaudebert <notifications@github.com>
wrote:

> In socorro/unittest/external/es/base.py
> <https://github.com/mozilla/socorro/pull/3242#discussion_r56846125>:
>
> > @@ -774,6 +794,15 @@ def test_with_version(self):
> >      return decorated
> >
> >
> > +class SuperSearchWithFields(SuperSearch):
>
> I actually think it's simpler to have this class in the base.py file. To
> me it's the same it terms of instant readability (I'm not used to the
> functools module, so I would have to go read the doc of partial anyway).
> And with my class, each file using SuperSearch only has one import to do
> instead of 3 (functools, SuperSearch and SuperSearchFields).
>
> —
> You are receiving this because you were assigned.
> Reply to this email directly or view it on GitHub
> <https://github.com/mozilla/socorro/pull/3242/files/7d6d6a3594d03be0b32aaa2eaf0b750211389f0e#r56846125>
>


-- 
Peter Bengtsson
Mozilla Tools & Services

PR #3243, opened 2016-03-17
Bug 1257572 - Add js::AutoEnterOOMUnsafeRegion::crash to our markers for OOM signatures

KaiRo-at:
From what I understand, js::AutoEnterOOMUnsafeRegion::crash is either a replacement or at least a similar thing to js::CrashAtUnhandlableOOM which we already handle, and from the name, I'm pretty sure that all of AutoEnterOOMUnsafeRegion is pretty surely OOM handling, so if we crash there, we should assign an "OOM" signature.


adngdb:
r+. Thanks KaiRo!

PR #3244, opened 2016-03-17
Add 'RefPtr.*' to prefixSignatureRegEx.

dbaron:
This is since nsRefPtr has been renamed to RefPtr.  I'm leaving the equivalent nsRefPtr.* line for older versions.


adngdb:
Is there a bug for this? If not, could you please file one? We try to always have a bug associated to each commit. https://bugzilla.mozilla.org/enter_bug.cgi?product=Socorro&component=General

Otherwise this looks good. :)

adngdb:
Thanks @dbaron !

PR #3246, opened 2016-03-17
fixes bug 1257655 - Remove memoize function for crashstats models

peterbe:


willkg:
I verified that this memoize isn't used anywhere. So this looks good to me. r+ if Travis is happy.

PR #3249, opened 2016-03-23
Fixes bug 1258420 - Added a by build date option to Topcrashers.

adngdb:
OK, this is ready now. Incidentally, I used the opportunity to make that Topcrashes page support several versions (which it should already have). 

@willkg how do you feel about reviewing this?

### DIFF
--- a/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/topcrash.js
+++ b/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/topcrash.js
@@ -16,6 +16,11 @@ var Correlations = (function() {
     var all_signatures = {};
 
     function loadCorrelations(types, callback) {
+        // If there is no osname, that means there is no results, and thus no
+        // need to bother loading correlations.
+        if (!Object.keys(osnames).length) {
+            return;
+        }
         var data = {
             correlation_report_types: types,
             product: SocReport.product,
--- a/webapp-django/crashstats/topcrashers/jinja2/topcrashers/topcrashers.html
+++ b/webapp-django/crashstats/topcrashers/jinja2/topcrashers/topcrashers.html
@@ -6,7 +6,7 @@
 {% endblock %}
 
 {% block page_title %}
-Top Crashers for {{ product }} {{ version }}
+Top Crashers for {{ product }} {{ ', '.join(query.versions) }}
 {% endblock %}
 
 {% block content %}
@@ -16,8 +16,8 @@ Top Crashers for {{ product }} {{ version }}
                 <a href="{{ url(
                     'crashstats:topcrasher',
                     product=query['product'],
-                    versions=query['versions'],
-                    date_range_type='report',
+                    versions=';'.join(query['versions']),
+                    date_range_type=query.range_type,
                     crash_type=query['crash_type'].replace('any', 'all'),
                     os_name=query['os_name'],
                     result_count=query['result_count'],
@@ -27,12 +27,12 @@ Top Crashers for {{ product }} {{ version }}
             </p>
             <h2>
                 Top Crashers for <span id="current-product">{{ product }}</span>
-                <span id="current-version">{{ version }}</span>
+                <span id="current-version">{{ ', '.join(query.versions) }}</span>
             </h2>
         </div>
 
         <div class="panel">
-            <div class="body">
+        <div class="body">
 
             {% if total_crashing_signatures %}
             <div>
@@ -59,6 +59,24 @@ Top Crashers for {{ product }} {{ version }}
                         >By Day</a>
                     </li>
                 </ul>
+
+                {% if versions_have_builds %}
+                <ul class="tc-filter">
+                    <li class="tc-selector-heading">Range Type:</li>
+                    <li>
+                        <a
+                            href="{{ change_query_string(_range_type=None) }}"
+                            {% if query.range_type == 'report' %} class="selected"{% endif %}
+                        >By Report Date</a>
+                    </li>
+                    <li>
+                        <a
+                            href="{{ change_query_string(_range_type='build') }}"
+                            {% if query.range_type == 'build' %} class="selected"{% endif %}
+                        >By Build Date</a>
+                    </li>
+                </ul>
+                {% endif %}
             </div>
 
             <ul class="tc-duration-type tc-filter">
@@ -90,6 +108,7 @@ Top Crashers for {{ product }} {{ version }}
                 <li><a href="{{ change_query_string(_facets_size=res_count) }}" {% if result_count == res_count %} class="selected" {% endif %}>{{ res_count }}</a></li>
                 {% endfor %}
             </ul>
+
             {% if total_crashing_signatures %}
             <table id="{% if not os_name %}signature-list{% else %}peros-tbl{% endif %}" class="data-table tablesorter">
                 <thead>
@@ -201,18 +220,22 @@ Top Crashers for {{ product }} {{ version }}
                             </div>
                         </td>
                     </tr>
-    {% endfor %}
+                {% endfor %}
                 </tbody>
             </table>
-        </div>
+            {% elif not versions_have_builds %}
+            <p class="no-results">
+                Some versions do not support the by build date range option.
+                <a href="{{ change_query_string(_range_type=None) }}">Range by Report Date instead?</a>
+            </p>
             {% else %}
-              <p class="no-results">
+            <p class="no-results">
                 No crashing signatures found for the period
                 <time class="start-date">{{ query.start_date }}</time> to
                 <time class="end-date">{{ query.end_date }}</time>.
-              </p>
+            </p>
             {% endif %}
-            </div>
+        </div>
         </div>
     </div>
 {% endblock %}
--- a/webapp-django/crashstats/topcrashers/tests/test_views.py
+++ b/webapp-django/crashstats/topcrashers/tests/test_views.py
@@ -51,6 +51,10 @@ class TestViews(BaseTestViews):
             if '_columns' not in params:
                 params['_columns'] = []
 
+            # By default we range by date, so there should be no filter on
+            # the build id.
+            ok_('build_id' not in params)
+
             if 'hang_type' not in params['_aggs.signature']:
                 # Return results for the previous week.
                 results = {
@@ -162,7 +166,7 @@ class TestViews(BaseTestViews):
         # Test that several versions do not raise an error.
         response = self.client.get(self.base_url, {
             'product': 'WaterWolf',
-            'version': '19.0;20.0',
+            'version': ['19.0', '20.0'],
         })
         eq_(response.status_code, 200)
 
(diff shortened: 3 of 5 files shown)


webapp-django/crashstats/topcrashers/jinja2/topcrashers/topcrashers.html
```diff
@@ -16,8 +16,8 @@
                 <a href="{{ url(
                     'crashstats:topcrasher',
                     product=query['product'],
-                    versions=query['versions'],
-                    date_range_type='report',
+                    versions=';'.join(query['versions']),
```
willkg: Can `query['versions']` ever contain non-ascii data?

webapp-django/crashstats/topcrashers/jinja2/topcrashers/topcrashers.html
```diff
@@ -201,18 +220,22 @@
...
                     </tr>
-    {% endfor %}
+                {% endfor %}
                 </tbody>
             </table>
-        </div>
+            {% elif not versions_have_builds %}
+            <p class="no-results">
+                Some versions do not support the by build date range option.
+                <a href="{{ change_query_string(_range_type=None) }}">Range by report date instead? </a>
```
willkg: Nit: There's a space after the ?.

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -66,11 +66,19 @@ def get_date_boundaries(parameters):
...
+    return date.strftime('%Y%m%d%H%M%S')
+
+
 def get_topcrashers_results(**kwargs):
-    '''Return the results of a search. '''
+    """Return the results of a search. """
     results = []
 
     params = kwargs
+    build_mode = params.pop('_range_type') == 'build'
```
willkg: Why does `_range_type` start with an underscore? Does that indicate something about the param?

willkg: Also, `build_mode` ends up with a boolean value. I think the related code reads better if it was called `is_build`.

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -193,11 +214,14 @@ def topcrashers(request, days=None, possible_days=None, default_context=None):
...
     product = request.GET.get('product')
-    versions = request.GET.get('version')
+    versions = request.GET.getlist('version')
     crash_type = request.GET.get('process_type')
     os_name = request.GET.get('platform')
     result_count = request.GET.get('_facets_size')
     tcbs_mode = request.GET.get('_tcbs_mode')
+    range_type = request.GET.get('_range_type')
+
+    range_type = range_type == 'build' and 'build' or 'report'
```
willkg: This looks wrong, but assuming the intention here is to restrict the domain of values to 'build' and 'report' then I think it's correct. To fix the "it looks wrong aspect", I'd probably just add a set of parens like this:

```
range_type = (range_type == 'build') and 'build' or 'report'
```

Though, I think I would have done this with a ternary:

```
range_type = 'build' if range_type == 'build' else 'report'
```

willkg:
I did a minor code review. I'm working on getting the webapp running so I can test it out. I'm probably close.

webapp-django/crashstats/topcrashers/jinja2/topcrashers/topcrashers.html
adngdb: Yes, and it breaks on prod already: https://crash-stats.mozilla.com/topcrashers/?product=Firefox&version=%E2%81%BB (that's a ``NoReverseMatch`` error, because the unicode symbol is not supported in the old topcrasher app's URL). But that only happens if a user manually enters an incorrect character here, so I guess we can leave that as is. What do you recommend?

webapp-django/crashstats/topcrashers/views.py
adngdb: Yes, I try to make this page's URL compatible with Super Search. In Super Search all parameters that are not filters start with an underscore (like ``_results_number`` or ``_sort`` for example). 

I think it's a mistake to even pass a boolean and not just the same variable. I'll change that.

webapp-django/crashstats/topcrashers/views.py
adngdb: Ah yeah, the ternary looks a lot better!

webapp-django/crashstats/topcrashers/jinja2/topcrashers/topcrashers.html
willkg: I'm going to think out loud here for a bit.

I think there are two questions here:

1. What's the requirement?
2. Does the application handle junk data well?

Towards the requirement question, it sounds like versions will not contain non-ascii data. So we can treat them like ascii in the code.

Towards the junk data question, it depends on a bunch of things like what's the ramifications of junk data? Who's using it? How hard is it to put in junk data? How hard is it to deal with? In this particular case, we do string joining in a bunch of places so to handle junk data better, we could write a function that does that work for us and handles possible junk data better (probably just drop non-ascii versions) and then use that function everywhere. It sounds like the audience is a "more technical" audience. It sounds like the ramification is that the search doesn't work which is fine. Contrast that with the case where junk data causes junk data to be in the db which is usually not fine and also not the case here.

We could wait until it's a problem and then write up a bug and address it then. I think that's what I'd do.

adngdb: I think that's basically what we have been doing up to now, based on the same arguments that you wrote (not a priority, technical audience... ). :)

willkg:
The code changes you made look fine.

I'm still trying to look at it in my local dev environment. I've got the webapp running now. But when the front page goes to fetch `http://localhost:5000/home/frontpage_json?product=WaterWolf`, then it kicks up an HTTP 500. I'm also not seeing any top crasher data for the fake data I have in the db. I suspect I missed a setup step somewhere or the fake data isn't sufficient to generate top crasher data for the window I'm looking at or something something.

willkg:
I read through the code, but I think I'm going to pass this to @peterbe for verifying that it's working in the ui. It's either that or wait a few days for me to figure out how to get the webapp working with access to data that makes it do things.

peterbe:
I'd be more than happy to take over the review but I'd really like @willkg to get some decent local dev set up and the current blocker is @jdotpz not being in air airplane or in a group meeting.

willkg:
Passing this off to @peterbe. I've got a bunch of setup problems, of which this is just one. As I figure out how to put the pieces together, it'll be easier to do reviews. But in the meantime I don't think it makes sense to block reviews for a week.

willkg:
I talked with @adngdb and @peterbe earlier today about getting data into Elasticsearch that would enable the top crasher reports to work. I spent a fair amount of time on that today, but I'm still not seeing it.

First off, I'm still using fake data set up by the `socorro setupdb` command. I migrate that to Elasticsearch using the script @adngdb mentioned. I've got tons of data in the 201612 and 201613 indexes in Elasticsearch. That seems fine.

When I load this page:

http://crash-stats:8000/topcrashers/?product=WaterWolf&version=5.0a1&days=7&_range_type=build

it eventually does this request:

http://crash-stats:8000/correlation/signatures?correlation_report_types=core-counts&correlation_report_types=interesting-addons&correlation_report_types=interesting-modules&product=WaterWolf&version=5.0a1

That kicks up an HTTP 400 with this HTTP response content:

```
<ul class="errorlist"><li>platforms<ul class="errorlist"><li>This field is required.</li></ul></li></ul>
```

This is all with a fresh vagrant vm built using this branch. I'm not sure what I could be doing wrong to see this.

I'll track down the view that's returning the HTTP 400 and try to understand what it's expecting and why and why it's not getting it. But, if someone has ideas, I'm all ears. :ear:

adngdb:
@willkg Ah, there's a compatibility bug between postgres' fakedata and what Elasticsearch expects, no data is marked as "browser" crash in ES when you use the migration app. So you have to change the "Type" filter to "All" to see any results. 

Good catch for the correlations error, I'll fix it!

peterbe:
The correlations stuff almost never works. Besides it's being almost entirely re-written. We currently rely on doing [HTTP GETs to get the stuff](https://github.com/mozilla/socorro/blob/master/socorro/external/http/correlations.py). That's going to entirely change to be an ES query instead. So the outputs will be different. Ideally there should be almost no visual difference.

adngdb:
Correlations issue fixed.

willkg:
Got it working, verified that "by build date" works and that it's not running correlations if there are no results. r+

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/topcrash.js
```diff
@@ -16,6 +16,11 @@ var Correlations = (function() {
     var all_signatures = {};
 
     function loadCorrelations(types, callback) {
+        // If there is no osname, that means there is no results, and thus no
+        // need to bother loading correlations.
+        if (Object.keys(osnames).length === 0) {
```
peterbe: Could have `if (!Object.keys(osnames).length)`

webapp-django/crashstats/topcrashers/jinja2/topcrashers/topcrashers.html
```diff
@@ -201,18 +220,22 @@
...
                     </tr>
-    {% endfor %}
+                {% endfor %}
                 </tbody>
             </table>
-        </div>
+            {% elif not versions_have_builds %}
+            <p class="no-results">
+                Some versions do not support the by build date range option.
+                <a href="{{ change_query_string(_range_type=None) }}">Range by report date instead?</a>
```
peterbe: "Range by Report Date instead?"

webapp-django/crashstats/topcrashers/tests/test_views.py
```diff
@@ -162,7 +166,7 @@ def mocked_supersearch_get(**params):
         # Test that several versions do not raise an error.
         response = self.client.get(self.base_url, {
             'product': 'WaterWolf',
-            'version': '19.0;20.0',
+            'version': ['19.0', '20.0'],
```
peterbe: ahhh finally!

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -86,6 +94,12 @@ def get_topcrashers_results(**kwargs):
     if params.get('process_type') in ('any', 'all'):
         params['process_type'] = None
 
+    if range_type == 'build':
+        params['build_id'] = [
+            '>=' + datetime_to_build_id(dates[0]),
+            '<' + datetime_to_build_id(dates[1])
+        ]
```
peterbe: That simple?! How awesome!

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -158,6 +171,14 @@ def get_topcrashers_results(**kwargs):
         params['_aggs.signature'] = [
             'platform',
         ]
+        params['_facets_size'] *= 2
```
peterbe: This could do with a comment explaining the reasoning for doing this doubling.

peterbe:
I looked too. Looks great. r+
I'm stunned how simple the solution is now that you've figured it out. Well done.

PR #3250, opened 2016-03-24
fixes bug 1257146 - Product build types pure implementation

peterbe:


webapp-django/crashstats/crashstats/models.py
```diff
@@ -26,6 +26,7 @@
 import socorro.external.postgresql.gccrashes
 import socorro.external.postgresql.crontabber_state
 import socorro.external.postgresql.adi
+import socorro.external.postgresql.product_build_types
```
adngdb: Maybe at some point we'll want to import the whole ``socorro.external.postgresql``, don't you think?

adngdb:
r+

webapp-django/crashstats/crashstats/models.py
peterbe: That would only be possible if `socorro/external/postgresql/__init__.py` imports all those modules.

PR #3251, opened 2016-03-25
Bug 1249359 store correlations in es

peterbe:
The way to run this is like this:

```
socorro correlations  \
--resource.postgresql.database_port=5433 \
--resource.postgresql.database_name=breakpad \
--secrets.postgresql.database_username=breakpad_rw \
--secrets.postgresql.database_password=*********** \
--resource.boto.access_key="**********************" \
--secrets.boto.secret_access_key="*********************" \
--resource.boto.resource_class=socorro.external.boto.connection_context.RegionalS3ConnectionContext \
--resource.boto.bucket_name="org.allizom.crash-stats.rhelmer-test.crashes" \
--resource.postgresql.transaction_executor_class=socorro.database.transaction_executor.TransactionExecutor \
--global.correlations.core.output_class=socorro.external.es.correlations.CoreCounts \
--global.correlations.interesting.output_class=socorro.external.es.correlations.InterestingModules \
--new_crash_source.elasticsearch.elasticsearch_urls=localhost:9222 \
--source.crashstorage_class=socorro.analysis.correlations.correlations_app.LocallyCachedBotoS3CrashStorage \
--producer_consumer.number_of_threads=5 \
--new_crash_source.cap=2000 \
$@
```

That'll start 5 threads that jointly download 2000 processed crashes and then generates the summaries which it inserts into a local ES. 

Note: I have my ssh tunnel set to so that localhost:9222 is the stage ES. That's where I get a days worth of UUIDs from.

### DIFF
--- a/socorro/analysis/correlations/correlations_app.py
+++ b/socorro/analysis/correlations/correlations_app.py
@@ -2,37 +2,24 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import json
-import re
+import datetime
+import os
+import tempfile
 
-from itertools import (
-    product as iter_product,
-    ifilter,
-)
-from functools import partial
+import ujson as json
 
-from collections import defaultdict, Sequence, MutableMapping
-from contextlib import contextmanager
-
-from configman import Namespace, RequiredConfig, class_converter
-from configman.dotdict import DotDict as ConfigmanDotDict
-from configman.converters import list_converter, to_str
-
-from socorro.analysis.correlations import macdebugids
-from socorro.analysis.correlations import addonids
+from configman import Namespace, class_converter
+from configman.dotdict import DotDict
 
 from socorrolib.app.fetch_transform_save_app import (
     FetchTransformSaveWithSeparateNewCrashSourceApp
 )
 
-from socorrolib.lib.transform_rules import Rule
-from socorrolib.lib.util import DotDict as SocorroDotDict
-from socorrolib.lib.converters import change_default
+from socorrolib.lib.datetimeutil import UTC
+from socorro.external.crashstorage_base import CrashIDNotFound
+from socorro.external.postgresql.products import ProductVersions
 from socorro.processor.processor_2015 import rule_sets_from_string
-from socorro.processor.processor_app import ProcessorApp
-from socorro.external.crashstorage_base import (
-    CrashIDNotFound
-)
+from socorro.external.boto.crashstorage import BotoS3CrashStorage
 
 #------------------------------------------------------------------------------
 correlation_rule_sets = [
@@ -42,18 +29,25 @@ correlation_rule_sets = [
         "socorrolib.lib.transform_rules.TransformRuleSystem",
         "apply_all_rules",
         "socorro.analysis.correlations.core_count_rule"
-            ".CorrelationCoreCountRule, "
+        ".CorrelationCoreCountRule, "
         "socorro.analysis.correlations.interesting_rule"
-            ".CorrelationInterestingModulesRule,"
+        ".CorrelationInterestingModulesRule,"
         "socorro.analysis.correlations.interesting_rule"
-            ".CorrelationInterestingModulesVersionsRule,"
+        ".CorrelationInterestingModulesVersionsRule,"
         "socorro.analysis.correlations.interesting_rule"
-            ".CorrelationInterestingAddonsRule,"
+        ".CorrelationInterestingAddonsRule,"
         "socorro.analysis.correlations.interesting_rule"
-            ".CorrelationInterestingAddonsVersionsRule,"
+        ".CorrelationInterestingAddonsVersionsRule,"
     ],
 ]
 correlation_rule_sets_as_string = json.dumps(correlation_rule_sets)
+
+
+def date_with_default_yesterday(value):
+    if not value:
+        return datetime.datetime.utcnow().date() - datetime.timedelta(days=1)
+    y, m, d = [int(x) for x in value.split('-')]
+    return datetime.date(y, m, d)
 
 
 #==============================================================================
@@ -73,19 +67,80 @@ class CorrelationsApp(FetchTransformSaveWithSeparateNewCrashSourceApp):
         likely_to_be_changed=True,
     )
 
+    required_config.add_option(
+        'transaction_executor_class',
+        default="socorro.database.transaction_executor."
+        "TransactionExecutorWithInfiniteBackoff",
+        doc='a class that will manage transactions',
+        from_string_converter=class_converter,
+        reference_value_from='resource.postgresql',
+    )
+    required_config.add_option(
+        'database_class',
+        default=(
+            'socorro.external.postgresql.connection_context'
+            '.ConnectionContext'
+        ),
+        doc='the class responsible for connecting to Postgres',
+        from_string_converter=class_converter,
+        reference_value_from='resource.postgresql',
+    )
+
+    required_config.add_option(
+        name='date',
+        doc='Specific date to run this for',
+        default='',
+        from_string_converter=date_with_default_yesterday,
+    )
+
+    required_config.add_option(
+        name='product',
+        doc='Product name',
+        default='Firefox',
+    )
+
+    #--------------------------------------------------------------------------
+    def __init__(self, config, quit_check_callback=None):
+        super(CorrelationsApp, self).__init__(config)
+
     #--------------------------------------------------------------------------
     @staticmethod
     def get_application_defaults():
         return {
             "number_of_submissions": 'all',
-            "source.crashstorage_class":
-                'socorro.external.boto.crashstorage.BotoS3CrashStorage',
-            "destination.crashstorage_class":
-                'socorro.external.crashstorage_base.NullCrashStorage',
-            "new_crash_source.new_crash_source_class":
-                'socorro.external.postgresql.new_crash_source'
-                '.PGPVNewCrashSource',
+            "source.crashstorage_class": (
+                'socorro.external.boto.crashstorage.BotoS3CrashStorage'
+            ),
+            "destination.crashstorage_class": (
+                'socorro.external.crashstorage_base.NullCrashStorage'
+            ),
+            "new_crash_source.new_crash_source_class": (
+                'socorro.external.es.new_crash_source.ESNewCrashSource'
+            ),
         }
+
+    #--------------------------------------------------------------------------
+    def _create_iter(self):
+        hits = ProductVersions(config=self.config).get(
+            active=True,
+            product=self.config.product
+        )['hits']
+        versions = [
+            x['version'] for x in hits if not x['version'].endswith('b')
+        ]
+        assert versions, "No active versions"
+
+        # convert a datetime.date object to datetime.datetime
+        dt = datetime.datetime(
+            self.config.date.year,
+            self.config.date.month,
+            self.config.date.day,
+        ).replace(tzinfo=UTC)
+        return self.new_crash_source.new_crashes(
+            dt,
+            product=self.config.product,
+            versions=versions,
+        )
 
     #--------------------------------------------------------------------------
     def _transform(self, crash_id):
@@ -117,20 +172,11 @@ class CorrelationsApp(FetchTransformSaveWithSeparateNewCrashSourceApp):
                 meta_data
             )
             self.quit_check()
-        try:
-            self.destination.save_processed(processed_crash)
-            self.config.logger.info('saved - %s', crash_id)
-        except Exception as x:
-            self.config.logger.error(
-                "writing raw: %s",
-                str(x),
-                exc_info=True
-            )
 
     #--------------------------------------------------------------------------
     def _setup_source_and_destination(self):
         super(CorrelationsApp, self)._setup_source_and_destination()
-        self.rule_system = ConfigmanDotDict()
+        self.rule_system = DotDict()
         for a_rule_set_name in self.config.rules.rule_sets.names:
             self.config.logger.debug(
                 'setting up rule set: %s',
@@ -149,9 +195,53 @@ class CorrelationsApp(FetchTransformSaveWithSeparateNewCrashSourceApp):
         self.config.logger.debug('CorrelationsApp closes')
         for a_rule_set_name, a_rule_set in self.rule_system.iteritems():
             self.config.logger.debug('closing %s', a_rule_set_name)
-            try:
-                a_rule_set.close()
-            except AttributeError:
-                # guess we don't need to close that rule
-                pass
+            a_rule_set.close()
         self.config.logger.debug('done closing rules')
+
+
+#==============================================================================
+class LocallyCachedBotoS3CrashStorage(BotoS3CrashStorage):  # pragma: no cover
+    """This class is dedicated only for local development, but checked in.
+    Never enabled by default and has no test coverage.
+
+    When you attempt to run correlations repeatedly on your laptop,
+    you have to download lots and lots of processed crashes from S3.
+    That's fine once but if you need to re-run it again, you'll have
+    to download the same stuff again. It's faster then to just use
+    this class instead of the default BotoS3CrashStorage class.
+
+    To enable this class instead, set:
+
+    --source.crashstorage_class=socorro.analysis.correlations.correlations_app\
+    .LocallyCachedBotoS3CrashStorage
+
+    in your call to `socorro correlations ...`
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(LocallyCachedBotoS3CrashStorage, self).__init__(
+            *args, **kwargs
+        )
+        self.temp_dir = os.path.join(
+            tempfile.gettempdir(),
+            'locally_downloaded_processed_crashes'
+        )
+        if not os.path.isdir(self.temp_dir):
+            os.mkdir(self.temp_dir)
+
+    def get_unredacted_processed(self, crash_id):
+        file_path = os.path.join(self.temp_dir, '{}.json'.format(crash_id))
+        try:
+            with open(file_path, 'r') as f:
+                return json.load(f)
+        except IOError:
+            crash = (
+                super(LocallyCachedBotoS3CrashStorage, self)
+                .get_unredacted_processed(crash_id)
+            )
+            with open(file_path, 'w') as f:
+                json.dump(crash, f)
+            self.config.logger.debug(
+                'Cache MISS downloading {}'.format(crash_id)
+            )
+            return crash
--- a/socorro/external/es/correlations.py
+++ b/socorro/external/es/correlations.py
@@ -2,19 +2,37 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import datetime
 import json
 import os
+import hashlib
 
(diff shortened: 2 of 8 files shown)


peterbe:
@adngdb I know the test is failing but I can't figure out what the error means. It works locally. Does it make any sense to you?

socorro/analysis/correlations/correlations_app.py
```diff
@@ -149,9 +195,55 @@ def close(self):
...
+    That's fine once but if you need to re-run it again, you'll have
+    to download the same stuff again. It's faster then to just use
+    this class instead of the default BotoS3CrashStorage class.
+
+    To enable this class instead, set:
+
+    --source.crashstorage_class=socorro.analysis.correlations.correlations_app.LocallyCachedBotoS3CrashStorage
+
+    in your call to `socorro correlations ...`
+    """
```
adngdb: Wow, that is awesome!

socorro/analysis/correlations/correlations_app.py
```diff
@@ -42,20 +29,27 @@
...
     ],
 ]
 correlation_rule_sets_as_string = json.dumps(correlation_rule_sets)
 
 
+def date_with_default_yesterday(value):
+    if not value:
+        return datetime.datetime.utcnow().date() - datetime.timedelta(days=1)
+    y, m, d = [int(x) for x in value.split('-')]
+    return datetime.date(y, m, d)
```
adngdb: We have ``isodate`` in our requirements, why not use it here? This code doesn't look very solid.

socorro/analysis/correlations/correlations_app.py
```diff
@@ -73,21 +67,82 @@ class CorrelationsApp(FetchTransformSaveWithSeparateNewCrashSourceApp):
...
+
+    required_config.add_option(
+        name='product',
+        doc='Product name',
+        default='Firefox',
+    )
+
+    #--------------------------------------------------------------------------
+    def __init__(self, config, quit_check_callback=None):
+        super(CorrelationsApp, self).__init__(config)
```
adngdb: Why are you not passing down the ``quit_check_callback`` param, or doing something with it? As far as I understand it, it's pretty useful, notably if you want to kill a running app.

socorro/analysis/correlations/correlations_app.py
```diff
@@ -73,21 +67,82 @@ class CorrelationsApp(FetchTransformSaveWithSeparateNewCrashSourceApp):
...
+            x['version'] for x in hits if not x['version'].endswith('b')
+        ]
+        assert versions, "No active versions"
+
+        # convert a datetime.date object to datetime.datetime
+        dt = datetime.datetime(
+            self.config.date.year,
+            self.config.date.month,
+            self.config.date.day,
+        ).replace(tzinfo=UTC)
```
adngdb: Why do you make it a date in the config if it's to turn it into a datetime here?

socorro/external/es/correlations.py
```diff
@@ -2,19 +2,41 @@
...
 
+
+# Needed for the ability to find the
+# mappings/correlations_index_settings.json file whose path is relative
+# to this file.
 HERE = os.path.dirname(os.path.abspath(__file__))
 
 
-class Correlations(RequiredConfig):
+def string_to_list(input_str):
```
adngdb: Already exists in configman: https://github.com/mozilla/configman/blob/master/configman/converters.py#L300

Use ``from configman import class_converter, str_to_list`` above.

socorro/external/es/correlations.py
```diff
@@ -51,6 +84,124 @@ def create_correlations_index(self, es_index):
...
+                    'key': name,
+                    'notes': notes,
+                }
+                with self.es_context() as conn:
+                    conn.index(
+                        index=index,
+                        # see correlations_index_settings.json
+                        doc_type='correlations',
+                        body=doc,
+                        id=self.make_id(doc)
```
adngdb: Do you intend to ever call a document by its ID? If not, then I don't think you need to bother creating it. You can just remove this line and let Elasticsearch create a new UUID for each document you send.

adngdb: Maybe what you want to solve is the problem of the job failing midway. If that happens, and you re-run it from the start, and you use random IDs, then the same document might be indexed twice. However, if that's what you're chasing, I think there is another problem: what happens if you re-run the correlations and the results are different? Then all of your IDs will be different, and you will have a lot of duplicated results. 

I think that, in that case, you want to generate the ID based on keys that actually make a good primary key. That would be ``product``, ``version``, ``platform``, ``date`` and ``signature``. ``count``, ``notes`` and ``payload``, notably, shouldn't be part of the ID since they could potentially change.

socorro/external/es/correlations.py
```diff
@@ -51,6 +84,124 @@ def create_correlations_index(self, es_index):
...
+        name,
+        key,
+    ):
+        date = self._prefix_to_datetime_date(prefix)
+        index = self.get_index_for_date(date)
+        self.create_correlations_index(index)
+
+        notes = counts_summary_structure['notes']
+        product = key.split('_')[0]
+        version = key.split('_')[1]
```
adngdb: Isn't this better: ``product, version = key.split('_', 1)``?

socorro/external/es/correlations.py
```diff
@@ -51,6 +84,124 @@ def create_correlations_index(self, es_index):
...
+                    conn.index(
+                        index=index,
+                        # see correlations_index_settings.json
+                        doc_type='correlations',
+                        body=doc,
+                        id=self.make_id(doc)
+                    )
+
+    def close(self):
+        pass
```
adngdb: Do you have to define this in all sub-classes? Can you not define it in the ``Correlations`` class instead?

socorro/external/es/correlations.py
```diff
@@ -51,6 +84,124 @@ def create_correlations_index(self, es_index):
...
+        name,
+        key,
+    ):
+        date = self._prefix_to_datetime_date(prefix)
+        index = self.get_index_for_date(date)
+        self.create_correlations_index(index)
+
+        notes = counts_summary_structure['notes']
+        product = key.split('_')[0]
+        version = key.split('_')[1]
```
adngdb: Same as above. :)

socorro/external/es/new_crash_source.py
```diff
@@ -0,0 +1,83 @@
...
+    )
+    required_config.elasticsearch = Namespace()
+    required_config.elasticsearch.add_option(
+        'elasticsearch_class',
+        default='socorro.external.es.connection_context.ConnectionContext',
+        from_string_converter=class_converter,
+        reference_value_from='resource.elasticsearch',
+    )
+
+    #--------------------------------------------------------------------------
```
adngdb: You're mixing Lars style and Peter style in this file. :)

socorro/external/es/new_crash_source.py
```diff
@@ -0,0 +1,83 @@
...
+
+        query = {
+            'filter': {
+                'bool': {
+                    'must': [
+                        {
+                            'range': {
+                                'processed_crash.date_processed': {
+                                    'gte': date.isoformat(),
+                                    'lte': next_day.isoformat(),
```
adngdb: This needs to be ``lt`` instead of ``lte``.

socorro/unittest/external/es/base.py
```diff
@@ -963,7 +969,7 @@ def index_many_crashes(
         )
         self.refresh_index()
 
-    def refresh_index(self):
+    def refresh_index(self, index=None):
```
adngdb: That variable is usually called ``es_index`` throughout the app, I think it would be beneficial to keep that convention.

adngdb: Meh, that's a small nit, really. It doesn't really matter.

socorro/unittest/external/es/test_correlations.py
```diff
@@ -3,25 +3,89 @@
...
+    InterestingModules,
+)
 from socorro.unittest.external.es.base import ElasticsearchTestCase
 
 # Uncomment these lines to decrease verbosity of the elasticsearch library
 # while running unit tests.
-# import logging
-# logging.getLogger('elasticsearch').setLevel(logging.ERROR)
+import logging
+logging.getLogger('elasticsearch').setLevel(logging.ERROR)
```
adngdb: You forgot to re-comment those. We want the noise in travis! :)

socorro/unittest/external/es/test_correlations.py
```diff
@@ -3,25 +3,89 @@
...
 # logging.getLogger('requests').setLevel(logging.ERROR)
 
 
 ES_CORRELATIONS_INDEX = 'integration_test_correlations_%Y%m'
 
 
-class IntegrationTestCorrelations(ElasticsearchTestCase):
+sample_core_counts_summary = {
```
adngdb: Shouldn't those be all caps? They look like constants to me.

socorro/unittest/external/es/test_correlations.py
```diff
@@ -31,12 +95,19 @@ def __init__(self, *args, **kwargs):
         # deleted in tearDown.
         self.indices_for_deletion = []
 
+        self.es_client = elasticsearch.Elasticsearch(
+            hosts=self.config.elasticsearch.elasticsearch_urls
+        )
```
adngdb: That is the same thing as ``self.connection``. Why create two connections?

socorro/unittest/external/es/test_correlations.py
```diff
@@ -53,3 +124,90 @@ def test_create_correlations_index(self):
...
+        eq_(_source['date'], today.strftime('%Y-%m-%d'))
+        eq_(_source['notes'], sample_core_counts_summary['notes'])
+        eq_(_source['key'], 'core-counts')
+        eq_(_source['platform'], 'Windows NT')
+        eq_(_source['product'], 'Firefox')
+        eq_(_source['version'], '43.0.1')
+        payload, = (
+            sample_core_counts_summary['Windows NT']['signatures'].values()
+        )
+        eq_(_source['payload'], json.dumps(payload))
```
adngdb: Those tests are quite difficult to read. I think they could do with a bit more spacing. :)

socorro/unittest/external/es/test_correlations.py
```diff
@@ -53,3 +124,90 @@ def test_create_correlations_index(self):
...
+        today = datetime.datetime.utcnow()
+        correlations.store(
+            sample_core_counts_summary,
+            today.strftime('%Y%m%d'),
+            'core-counts',
+            'Firefox_43.0.1'
+        )
+        es_index = correlations.get_index_for_date(today.date())
+        correlations.create_correlations_index(es_index)
+        self.indices_for_deletion.append(es_index)
```
adngdb: Unless I missed it, you do not set a special value for this field in unit tests. You should, otherwise you risk deleting an actual index in your local database. Or mine. Maybe that's not the most terrible thing that can happen, but it's still annoying to see tests delete your data. :)

You can set a different default for tests here: https://github.com/peterbe/socorro/blob/bug-1249359-store-correlations-in-es/socorro/unittest/external/es/base.py#L22

socorro/unittest/external/es/test_new_crash_source.py
```diff
@@ -0,0 +1,199 @@
...
+
+        try:
+            self.index_client.delete(
+                self.config.elasticsearch.elasticsearch_index
+            )
+
+        except elasticsearch.exceptions.NotFoundError:
+            # It's fine it's fine; 404 means the test didn't create any
+            # indices, therefore they can't be deleted.
+            pass
```
adngdb: This is already done in ``ElasticsearchTestCase.tearDown``.

socorro/unittest/external/es/test_new_crash_source.py
```diff
@@ -0,0 +1,199 @@
...
+    def __init__(self, *args, **kwargs):
+        super(IntegrationTestESNewCrashSource, self).__init__(*args, **kwargs)
+
+        self.config = self.get_tuned_config(ESNewCrashSource)
+
+        # Helpers for interacting with ES outside of the context of a
+        # specific test.
+        self.es_client = elasticsearch.Elasticsearch(
+            hosts=self.config.elasticsearch.elasticsearch_urls
+        )
```
adngdb: Again, I think you can use ``self.connection`` instead.

socorro/unittest/external/es/test_new_crash_source.py
```diff
@@ -0,0 +1,199 @@
...
+            )
+
+        except elasticsearch.exceptions.NotFoundError:
+            # It's fine it's fine; 404 means the test didn't create any
+            # indices, therefore they can't be deleted.
+            pass
+
+    def test_no_new_crashes(self):
+        new_crash_source = ESNewCrashSource(self.config)
+        self.refresh_index()
```
adngdb: I bet you don't need this anymore, since you run ``health_check``.

socorro/analysis/correlations/correlations_app.py
peterbe: Looks pretty solid to me. It's never supposed to be any other format. Isodate is for when you have unpredictable formats coming in.

socorro/analysis/correlations/correlations_app.py
peterbe: The parent class does not accept it. 
https://github.com/mozilla/socorrolib/blob/master/socorrolib/app/fetch_transform_save_app.py#L110

socorro/analysis/correlations/correlations_app.py
peterbe: Good question. The dt, when passed into ESNewCrashSource NEEDS to have a tzinfo and I can't set that on a `datetime.date` object. 
Can you see any other way to do it?

socorro/external/es/correlations.py
peterbe: Thanks.

socorro/external/es/correlations.py
peterbe: Yeah, I'll make the `make_id()` function only generate based on certain keys.

socorro/external/es/correlations.py
peterbe: Much better.

socorro/external/es/correlations.py
peterbe: thanks. fixed.

socorro/external/es/new_crash_source.py
peterbe: What should it be? Consistency is nice.

peterbe: Personally I don't like the excessive comment lines he prefers.

peterbe: Actually. This particular new file can be without these lines.

socorro/unittest/external/es/base.py
peterbe: fixed.

socorro/unittest/external/es/test_correlations.py
peterbe: Stole it from `test_crashstorage.py`

socorro/unittest/external/es/test_correlations.py
peterbe: I get the idea but what do you suggest? The defaults have the word "test" in them.

socorro/unittest/external/es/test_correlations.py
peterbe: I don't understand. Put an extra linebreak between each? But this segment of tests belongs to the search done on line 144.

socorro/unittest/external/es/base.py
```diff
@@ -167,7 +167,9 @@
         'is_mandatory': False,
         'storage_mapping': {
             'type': 'date',
-            'format': 'yyyy-MM-dd\'T\'HH:mm:ssZZ||yyyy-MM-dd\'T\'HH:mm:ss.SSSSSSZZ'
+            'format': (
+                'yyyy-MM-dd\'T\'HH:mm:ssZZ||yyyy-MM-dd\'T\'HH:mm:ss.SSSSSSZZ'
+            )
```
peterbe: I changed this because it was the only thing missing from making this file pep8 compliant and git hooks kept nagging me.

peterbe:
@adngdb See the extra [new commit](https://github.com/mozilla/socorro/pull/3251/commits/b409588c5ae17930ab5da5d129cdcd0656b616fd).
It solves that problem I emailed you about. It used to work like this:
```
1) start scan()
2) yield crash_id1
3) download crash_id1 from S3
4) yield crash_id1
5) download crash_id2 from S3
6) yield crash_id1
7) download crash_id3 from S3
...
999) close the new_crashes() iterator and calculate the summaries. 
```
That meant the scroll scan has to be open for a crazy long time. Instead now, it does this:
```
1) start scan()
2) list.append(crash_id1)
3) list.append(crash_id2)
4) list.append(crash_id3)
5) close the scroll scan()
6) yield crash_id1
4) yield crash_id2
6) yield crash_id3
7) download crash_id1 from S3
8) download crash_id2 from S3
9) download crash_id3 from S3
...
999) close the new_crashes() iterator and calculate the summaries. 
```
Now the scroll scan can close as soon as all crash IDs are extracted out and we can return to the FTS machinery.

socorro/analysis/correlations/correlations_app.py
adngdb: Well I would just make it a datetime right from the start, and use ``dt.replace()`` to make sure it has hour, minute and second to 0.

socorro/external/es/new_crash_source.py
adngdb: Yeah, I guess at some point we'll remove those all-dashes lines. Right now I mostly care about consistency inside files.

socorro/unittest/external/es/test_correlations.py
adngdb: Well, I think you could add line breaks so that there are more logical blocks. Unit tests are not fun to read, so I think it's extra important to make them as readable as possible.

socorro/unittest/external/es/test_correlations.py
adngdb: I guess something like ``socorro_integration_test_correlations`` would be coherent with the other names? Also note that it means there won't be weekly indices, but you likely don't need them except for specific tests, and you'll want to remove that index in the tearDown of your tests.

socorro/external/es/correlations.py
```diff
@@ -96,9 +92,21 @@ def _prefix_to_datetime_date(self, prefix):
...
-    def make_id(document):
+    def make_id(document, only_keys=None):
+        only_keys = only_keys or (
+            'platform',
+            'product',
+            'version',
+            'date',
+            'key',
+            'signature',
+            'count',
```
adngdb: I think ``count`` should not be in there. It could change when you reload the job, if new crashes came in in the mid-time for example. I can happen if someone triggers the processing of a crash that has been throttled. Or say we had a problem with processing, and we want to re-run correlations a bit later when all crashes made it into the database.

adngdb:
It seems to have worked great locally! 

r+ when the ``make_id`` function is fixed.

socorro/analysis/correlations/correlations_app.py
peterbe: I kinda like that the config is all about a date. But the internal implementation is a datetime. 
If you do `socorro correlations --date=2016-03-28 --help` it will display what value we have for the date and if we let the config be a datetime it'd now display a datetime which isn't what I put in.

socorro/external/es/correlations.py
peterbe: You're right. I've changed it.

PR #3252, opened 2016-03-29
No bug, doc only - Added instruction to migrate data from psql to Ela…

adngdb:
…sticsearch.


willkg:
Looking at this now...

willkg:
I ran this most of today just to see what happens. I definitely see tons of data in ES and it's in week-based indexes. So I think this works.

I had seen this material in the Elasticsearch chapter, but didn't really understand what it was doing or that it wasn't transactional and could be interrupted, so I'm not sure this would have helped me much.

However, I do think this is a good change to this doc because without this step, the webapp is kind of silly. I'll do a pass on docs at some point soon and we can work out clarification issues there.

Given all that, r+!

PR #3253, opened 2016-03-29
fixes bug 1260216 - Use requests in ftpscraper.py

peterbe:
To actually run:
```
time python socorro/cron/jobs/ftpscraper.py --product=thunderbird,firefox,mobile
```
(Takes about 5 minutes and instead of writing into PG, it just prints what it would insert)

I'm not convinced about the way the test classes are set up with the mocking. Not sure it's particularly important either.

### DIFF
--- a/socorro/unittest/cron/jobs/test_ftpscraper.py
+++ b/socorro/unittest/cron/jobs/test_ftpscraper.py
@@ -3,11 +3,10 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import datetime
-import urllib2
 import json
 from functools import wraps
-from cStringIO import StringIO
 
+import requests
 import mock
 from nose.tools import eq_, ok_, assert_raises
 from crontabber.app import CronTabber
@@ -22,10 +21,16 @@ from socorro.unittest.cron.setup_configman import (
 )
 
 
-def stringioify(func):
+class Response(object):
+    def __init__(self, content, status_code=200):
+        self.content = content
+        self.status_code = status_code
+
+
+def responsify(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
-        return StringIO(func(*args, **kwargs))
+        return Response(func(*args, **kwargs))
     return wrapper
 
 
@@ -39,10 +44,14 @@ class TestFTPScraper(TestCaseBase):
         super(TestFTPScraper, self).setUp()
         self.psycopg2_patcher = mock.patch('psycopg2.connect')
         self.psycopg2 = self.psycopg2_patcher.start()
-        self.urllib2_patcher = mock.patch('urllib2.urlopen')
-        self.urllib2 = self.urllib2_patcher.start()
+        self.requests_session_patcher = mock.patch('requests.Session')
+        self.mocked_session = self.requests_session_patcher.start()
+
+        def download(url):
+            return self.mocked_session.get(url)
 
         self.scrapers = ftpscraper.ScrapersMixin()
+        self.scrapers.download = download
         self.scrapers.config = DotDict({
             'logger': mock.Mock()
         })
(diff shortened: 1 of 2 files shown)


willkg:
I hit a 503 which the assertion dies on:

```
  File "/home/vagrant/socorro/socorro-virtualenv/lib/python2.7/site-packages/crontabber/transaction_executor.py", line 46, in __call__
    result = function(connection, *args, **kwargs)
  File "/home/vagrant/socorro/socorro/cron/jobs/ftpscraper.py", line 365, in _scrape_json_releases_and_nightlies
    self.scrape_json_releases(connection, product_name)
  File "/home/vagrant/socorro/socorro/cron/jobs/ftpscraper.py", line 400, in scrape_json_releases
    for info in self.get_json_release(release, dirname):
  File "/home/vagrant/socorro/socorro/cron/jobs/ftpscraper.py", line 176, in get_json_release
    kvpairs = self.parse_build_json_file(json_url)
  File "/home/vagrant/socorro/socorro/cron/jobs/ftpscraper.py", line 69, in parse_build_json_file
    content = self.download(url)
  File "/home/vagrant/socorro/socorro/cron/jobs/ftpscraper.py", line 335, in download
AssertionError: 503
```

It's this url:

```
2016-03-30 16:59:13,503 WARNING - SocorroAppBaseClass -  - MainThread - warning, unsupported JSON file: https://archive.mozilla.org/pub/thunderbird/candidates/42.0b2-candidates/build1/linux-i686/en-US/thunderbird-42.0b2.mozinfo.json
```

When I ran ftpscraper on master, I didn't hit this. However, with this branch, I hit it three times in a row.

Couple of possibilities:

1. the issue is ephemeral--so we wait a bit and it goes away and yay!
2. there's some difference between the old code and the new code--I'll look into this
3. something changed with those files or that server in the last hour--I'll look into this, too

That's the update.

I still have to look at the code changes.

willkg:
I waited a while and am running it now and it's moving along nicely. Ergo, I suspect the issue was ephemeral. Those are THE BEST! :trophy:

peterbe:
It's very interesting actually. 503 means archives.mozilla.org is temporarily down. Which is fine. That's why we have so much retries. 
The old `patient_urlopen` (written by yours truly) it would just, on any 5xx error sleep for 20 seconds. And after 4x20 seconds it would eventually complain.
With the `requests` approach, it retries (I think) 5 times but if the site is down, there's no sleeping going on at all. It tries 5 times immediately one after another and it can do that in about 1-2 seconds. 

If, however, the archives.mozilla.org isn't immediately serving up a 503 error. I.e. if it takes long time to respond then our new requests approach will wait 10 seconds between each 5 attempts. 

Perhaps I need to go back and add a lot more "higher level" wrapper around this that is a lot more forgiving.

peterbe:
In all fairness, ftpscraper is a crontabber app and if it ever fails the whole job it'll wait 5 minutes and try the whole thing again. Yeah, that's perhaps a better thing to rely on instead of making the job itself more complicated. 

Ergo, if it eventually works, that's good enough for me. crontabber will have to handle the "patience" aspect.

willkg:
I think that makes sense: if crontabber is already dealing with retries, then no sense in having other retrying infrastructure.

When this fails with the HTTP 503 error, does it do it in a way that will cause crontabber to retry? Is crontabber looking for process exit codes or something else?

willkg:
When the assertion fails, then the Python process returns an exit code of 1. I'm assuming crontabber pays attention to exit codes, but I can't actually find the code where it does.

peterbe:
crontabber just wraps every execution of apps in a big fat `try: except Exception:`. It doesn't execute the jobs from the command line (i.e. subprocess). ftpscraper.py is an exception of a crontabber app in that you can run it on the command line.

socorro/unittest/cron/jobs/test_ftpscraper.py
```diff
@@ -22,10 +21,17 @@
...
+        self.content = content
+        self.status_code = status_code
+
+
+def responsify(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
-        return StringIO(func(*args, **kwargs))
+        res = func(*args, **kwargs)
+        return Response(res)
```
willkg: Nit: You could nix the `res` intermediary variable.

willkg:
The code seems fine to me as far as I can tell. r+

PR #3255, opened 2016-03-30
fixes bug 1064968 - Deprecate CurrentVersions, CurrentProducts and Pr…

peterbe:
…oductsVersions api endpoints

### DIFF
--- a/webapp-django/crashstats/crashstats/tests/test_models.py
+++ b/webapp-django/crashstats/crashstats/tests/test_models.py
@@ -198,110 +198,103 @@ class TestModels(DjangoTestCase):
         info = api.get('747238', 'product')
         eq_(info['bugs'], [{u'product': u'DIFFERENT'}])
 
-    @mock.patch('requests.get')
-    def test_current_versions(self, rget):
+    def test_current_versions(self):
         model = models.CurrentVersions
         api = model()
 
         def mocked_get(**options):
-            assert '/products' in options['url']
-            return Response("""
-                {"hits": {
-                   "SeaMonkey": [{
-                     "product": "SeaMonkey",
-                     "throttle": "100.00",
-                     "end_date": "2012-05-10 00:00:00",
-                     "start_date": "2012-03-08 00:00:00",
-                     "featured": true,
-                     "version": "2.1.3pre",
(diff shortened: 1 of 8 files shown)


peterbe:
By the way, I should explain something. Note how I put in those...
```python
if settings.DEBUG:  # pragma: no cover
    raise DeprecatedModelError("you're not supposed to use this")
```
That's to prevent us for ever using these in other views when in development. So we don't ever accidentally start depending on them again. 

Before I put those in, using the master branch, I ran a bunch of queries on `/api/CurrentProducts/`, '`/api/CurrentVersions/` and `/api/ProductsVersions/` and saved as `.json` files. Then after this patch was done, I ran those again and compared the results. They matched. 

After checking that I added the extra deprecation warning to the JSON output (when it's a dict) and the extra Deprecation-Warning hedaer. 

God forbid we have a bug in this legacy-bridge-hack. If we break someones code, it's a great opportunity to inform them of the new and much better `/api/ProductVersions/`.

webapp-django/crashstats/crashstats/tests/test_models.py
```diff
@@ -198,110 +198,103 @@ def new_mocked_get(**options):
...
                 }
-                """)
 
-            raise NotImplementedError(url)
+        models.ProductVersions.implementation().get.side_effect = mocked_get
 
-        rget.side_effect = mocked_get
         info = api.get()
-        eq_(info['hits'][0]['product_name'], 'NightTrain')
+        eq_(info['hits']['NightTrain'][0]['product'], 'NightTrain')
```
adngdb: Was this not supposed to stay the same? As far as I understand if you had to change this test that means the output of ``CurrentProducts`` changed, right?

adngdb:
r+

webapp-django/crashstats/crashstats/tests/test_models.py
peterbe: Uh. Good point. I forgot the logic here. I'll take a more careful look.

peterbe: Ha! No. the old test was just wrong. :)

PR #3261, opened 2016-03-31
bug 1261111 - Only DATABASE_HOST

peterbe:
See the bug for the story on this.

### DIFF
--- a/webapp-django/crashstats/settings/base.py
+++ b/webapp-django/crashstats/settings/base.py
@@ -723,7 +723,7 @@ SOCORRO_IMPLEMENTATIONS_CONFIG = {
         },
         'postgresql': {
             'database_hostname': config(
-                'DATABASE_HOSTNAME',
+                'DATABASE_HOST',
                 'localhost'
             ),
             'database_name': config(


webapp-django/crashstats/settings/base.py
```diff
@@ -723,7 +723,7 @@ def path(*dirs):
         },
         'postgresql': {
             'database_hostname': config(
-                'DATABASE_HOSTNAME',
+                'DATABASE_HOST',
                 'localhost'
             ),
```
willkg: This seems messy to me because it's changed the environment variable, but the configuration variable is still `database_hostname` so the two don't match. Matching is nice since it's less cognitive load.

Is there a reason we can't change the configuration variable, too? Or am I misunderstanding what's going on here?

willkg:
When I suggested we go with HOST vs. HOSTNAME, I didn't realize the context.

Now that I think I understand the context more, I wouldn't make this change. If I'm reading the rest of this file right, DATABASE_HOST is deprecated for DATABASE_URL, so we shouldn't be using DATABASE_HOST anymore. DATABASE_HOSTNAME as an environment variable matches the configuration variable it affects.

Seems like it'd be better to finish the work for ditching DATABASE_HOST in favor of DATABASE_URL and leave DATABASE_HOSTNAME as is.

I say that completely unsure if I'm missing something or not.

webapp-django/crashstats/settings/base.py
peterbe: The `database_hostname` is a configman/socorro.external.postgresql.connection_context thing. That's the only place where it's called "..._hostname". This variable `socorro/webapp-django/DATABASE_HOST` (that I propose) serves two purposes:

1) Be the hostname for the django ORM
2) Be the hostname for the implementations

I think it's not worth it changing all the socorro code to call this `database_host` too. It'd affect the processors, crontabber and possibly other bits and pieces like the crashmover. 

However, I kinda like that this value is only something that matters is "django land". It needs to be "handed over" to the implementations under a different name.

peterbe:
I've been thinking about that but it poses an immediate challenge. 

In my local set up I use two different database.
1) one for my local django ORM thing likes users, tokens, sessions, log events, etc. 
2) one for talking to hardcode socorro postgresql tables on the stage DB. 

This is not applicable for stage and prod and travis where it's not interesting to make that distinguishing. 

However, during the course of writing this reply I'm convinced there's a better way. We'll simply make it so you can define, in your local `.env` a `DATABASE_URL` **and** a `SOCORRO_DATABASE_URL`. 
The latter is optional. 

Yeah. Let's do that!

willkg:
If you were using DATABASE_URL, then you're all set because DATABASE_URL covers item 1 and DATABASE_HOSTNAME (and friends) cover item 2.

I think we should ditch the code that sets up the Django database backend using the old DATABASE_* variables.

https://github.com/mozilla/socorro/blob/master/webapp-django/crashstats/settings/base.py#L519

It's been almost a year since that if block was added. That seems like a sufficient amount of time to nix it.

(Again, I could be horribly misunderstanding things.)

peterbe:
Yes. We are in agreement. 
https://github.com/mozilla/socorro/pull/3262

PR #3262, opened 2016-03-31
bug 1261145 - Only use DATABASE_URL everywhere in webapp

peterbe:
So now I get the best of both worlds:
* MUCH fewer things
* On local dev, I can set up, in my `.env` a `IMPLEMENTATIONS_DATABASE_URL` to make it different.

### DIFF
--- a/webapp-django/crashstats/settings/base.py
+++ b/webapp-django/crashstats/settings/base.py
@@ -516,48 +516,16 @@ CACHES = {
 
 TIME_ZONE = config('TIME_ZONE', 'UTC')
 
-# Only use the old way of settings DATABASES IF you haven't fully migrated yet
-if (
-    not config('DATABASE_URL', None) and (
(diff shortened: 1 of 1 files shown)


willkg:
There's nothing in the vagrant setup regarding `DATABASE_*` vars, so we're ok there.

There's some minor documentation in `docs/configuring-crash-stats.rst` that should be updated.

I think this looks ok otherwise. Having said that, I'm not very familiar with `config`, so I might be missing things.

This will go through Travis. How else will we know if there are issues here? Will it be obvious on -stage? Are there other ways we could know?

peterbe:
Vagrant is the only place I'm not super comfortable with. I'm pretty confident it'll just work in stage (and prod), now thanks to me preparing the DATABASE_URL set up in consul there preemptively.

willkg:
I think this looks good. r+

PR #3264, opened 2016-03-31
fixes bug 1261164 - Skip bogus .json files in ftpscraper

peterbe:
I ran this a bunch of times before and after. The difference is small. I think it's because the real difference is in the latency between me and that S3 bucket. 

Before it was, on average 1.46 minutes. After it was 1.20 minutes. 

Either way, now you don't get a lot of those warning log messages at least.

### DIFF
--- a/socorro/cron/jobs/ftpscraper.py
+++ b/socorro/cron/jobs/ftpscraper.py
@@ -4,13 +4,15 @@ import re
 import os
 import json
 import urlparse
+import fnmatch
 
 import mock
 import lxml.html
 import requests
 from requests.adapters import HTTPAdapter
 
-from configman import Namespace, class_converter
+from configman import Namespace
+from configman.converters import class_converter, str_to_list
 from crontabber.base import BaseCronApp
 from crontabber.mixins import (
     as_backfill_cron_app,
@@ -26,6 +28,7 @@ from socorrolib.lib.datetimeutil import string_to_datetime
 class ScrapersMixin(object):
     """
     Mixin that requires to be able to call `self.download(some_url)`
+    and `self.skip_json_file(json_url)`.
     """
 
     def get_links(self, url, starts_with=None, ends_with=None):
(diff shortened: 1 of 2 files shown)


socorro/cron/jobs/ftpscraper.py
```diff
@@ -335,6 +345,16 @@ def download(self, url):
         assert response.status_code == 200, response.status_code
         return response.content
 
+    def skip_json_file(self, json_url):
+        basename = os.path.basename(json_url)
+        for rule in self.config.json_files_to_ignore:
+            # If the rule was, for example, `*.foo.bar` then,
+            # the regex becomes `.\.foo\.bar`.
+            regex_string = re.escape(rule).replace('\\*', '.')
```
willkg: I'm pretty sure this isn't right. This replaces a `*` with a regex `.` which matches a character of length 1.

I think what you're trying to do here is convert from a file-globbing style syntax where `*` is a wildcard for 0 or more characters. If that's the case, then you probably want to convert `*` to `.*`.

Further, I claim you also want to add anchors to the regexp and use `.match` instead of `.findall`. So then you're converting something like `*.foo.bar` to `^.*\.foo\.bar$`. This prevents substring matching so `*.foo` doesn't accidentally match `test.foo.baz` which I don't think you want.

Does that sound right? If so, then I think this line would be:

```
regex_string = '^' + re.escape(rule).replace(r'\*', '.*') + '$'
```

I'm a little iffy on the placement of that re.escape. I'm pretty sure that creates weird edge cases with escaping. I'm not sure if it matters.

Regardless, I suggest another approach which has the nicety of also supporting `?` and `[]` notation: use the fnmatch module. https://docs.python.org/2/library/fnmatch.html

peterbe: I was lazy about the `^` and `$`. Didn't think it's particularly plausible to have a file called `something.mozinfo.json.other.json` (they have to be something `.json` [anyway to get near this](https://github.com/peterbe/socorro/blob/bug-1261164-skip-bogus-json-files-in-ftpscraper/socorro/cron/jobs/ftpscraper.py#L175).

But this new fancy `fnmatch` looks perfect! I'll just use that instead.

peterbe:
Updated to use fnmatch now instead. Thanks for sharing.

willkg:
The changes look good to me. If Travis is happy, I'm happy. r+

PR #3265, opened 2016-04-04
fixes bug 598608 - Add a way to reprocess reports

peterbe:
cc @twobraids

### DIFF
--- a/socorro/external/rabbitmq/crashstorage.py
+++ b/socorro/external/rabbitmq/crashstorage.py
@@ -314,11 +314,6 @@ class ReprocessingOneRabbitMQCrashStore(ReprocessingRabbitMQCrashStore):
         'socorro.reprocessing'
     )
 
-    def __init__(self, config, quit_check_callback=None):
-        super(ReprocessingOneRabbitMQCrashStore, self).__init__(
-            config, quit_check_callback=quit_check_callback
-        )
-
     def reprocess(self, crash_id):
         return self.save_raw_crash(
             DotDict({'legacy_processing': 0}),
--- a/socorro/external/rabbitmq/crashstorage.py
+++ b/socorro/external/rabbitmq/crashstorage.py
@@ -15,15 +15,12 @@ from configman import (
     class_converter
 )
 from configman.dotdict import DotDict
+from socorrolib.lib.converters import change_default
 from socorro.external.rabbitmq.connection_context import (
     ConnectionContext,
     ConnectionContextPooled,
 )
-from socorro.external.crashstorage_base import (
-    Redactor,
-    CrashStorageBase,
-)
-from socorrolib.lib.converters import change_default
+from socorro.external.crashstorage_base import CrashStorageBase
 
 
 #==============================================================================
@@ -78,7 +75,6 @@ class RabbitMQCrashStorage(CrashStorageBase):
         doc='percentage of the time that rabbit will try to queue',
         reference_value_from='resource.rabbitmq',
     )
-
 
     #--------------------------------------------------------------------------
     def __init__(self, config, quit_check_callback=None):
@@ -319,10 +315,6 @@ class ReprocessingOneRabbitMQCrashStore(ReprocessingRabbitMQCrashStore):
     )
 
     def __init__(self, config, quit_check_callback=None):
-        config.forbidden_keys = ''
-        config.redactor_class = Redactor
-        config.throttle = 100
-        config.filter_on_legacy_processing = False  # must be set
         super(ReprocessingRabbitMQCrashStore, self).__init__(
             config, quit_check_callback=quit_check_callback
         )
--- a/socorro/external/rabbitmq/crashstorage.py
+++ b/socorro/external/rabbitmq/crashstorage.py
@@ -315,11 +315,11 @@ class ReprocessingOneRabbitMQCrashStore(ReprocessingRabbitMQCrashStore):
     )
 
     def __init__(self, config, quit_check_callback=None):
-        super(ReprocessingRabbitMQCrashStore, self).__init__(
+        super(ReprocessingOneRabbitMQCrashStore, self).__init__(
             config, quit_check_callback=quit_check_callback
         )
 
-    def post(self, crash_id):
+    def reprocess(self, crash_id):
         return self.save_raw_crash(
             DotDict({'legacy_processing': 0}),
             [],
--- a/socorro/external/rabbitmq/crashstorage.py
+++ b/socorro/external/rabbitmq/crashstorage.py
@@ -14,10 +14,13 @@ from configman import (
     Namespace,
     class_converter
 )
+from configman.dotdict import DotDict
 from socorro.external.rabbitmq.connection_context import (
-    ConnectionContextPooled
+    ConnectionContext,
+    ConnectionContextPooled,
 )
 from socorro.external.crashstorage_base import (
+    Redactor,
     CrashStorageBase,
 )
 from socorrolib.lib.converters import change_default
@@ -135,6 +138,7 @@ class RabbitMQCrashStorage(CrashStorageBase):
                 'RabbitMQCrashStorage saving crash %s', crash_id
             )
             self.transaction(self._save_raw_crash_transaction, crash_id)
+            return True
         else:
             self.config.logger.debug(
                 'RabbitMQCrashStorage not saving crash %s, legacy processing '
@@ -234,10 +238,10 @@ class RabbitMQCrashStorage(CrashStorageBase):
             while True:
                 crash_id_to_be_acknowledged = \
                     self.acknowledgment_queue.get_nowait()
-                #self.config.logger.debug(
-                    #'RabbitMQCrashStorage set to acknowledge %s',
-                    #crash_id_to_be_acknowledged
-                #)
+                # self.config.logger.debug(
+                #     'RabbitMQCrashStorage set to acknowledge %s',
+                #     crash_id_to_be_acknowledged
+                # )
                 try:
                     acknowledgement_token = \
                         self.acknowledgement_token_cache[
@@ -258,7 +262,7 @@ class RabbitMQCrashStorage(CrashStorageBase):
                         crash_id_to_be_acknowledged,
                         exc_info=True
                     )
-                except Exception as x:
+                except Exception:
                     self.config.logger.error(
                         'RabbitMQCrashStorage unexpected failure on %s',
                         crash_id_to_be_acknowledged,
@@ -299,3 +303,33 @@ class ReprocessingRabbitMQCrashStore(RabbitMQCrashStorage):
         False
     )
 
+
+#==============================================================================
+class ReprocessingOneRabbitMQCrashStore(ReprocessingRabbitMQCrashStore):
+    required_config = Namespace()
+    required_config.rabbitmq_class = change_default(
+        RabbitMQCrashStorage,
+        'rabbitmq_class',
+        ConnectionContext,
+    )
+    required_config.routing_key = change_default(
+        RabbitMQCrashStorage,
+        'routing_key',
+        'socorro.reprocessing'
+    )
+
+    def __init__(self, config, quit_check_callback=None):
+        config.forbidden_keys = ''
+        config.redactor_class = Redactor
+        config.throttle = 100
+        config.filter_on_legacy_processing = False  # must be set
+        super(ReprocessingRabbitMQCrashStore, self).__init__(
+            config, quit_check_callback=quit_check_callback
+        )
+
+    def post(self, crash_id):
+        return self.save_raw_crash(
+            DotDict({'legacy_processing': 0}),
+            [],
+            crash_id
+        )
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -394,6 +394,7 @@ class SocorroCommon(object):
                 config = config_from_configman()
                 if self.implementation_config_namespace:
                     config = config[self.implementation_config_namespace]
+
                 _implementations[key] = self.implementation(
                     config=config
                 )
@@ -420,9 +421,9 @@ class SocorroMiddleware(SocorroCommon):
     # by default, assume the class to not have an implementation reference
     implementation = None
 
-    # The default, when configman has set up the implementation config
-    # we can use a sub-select of that.
-    implementation_config_namespace = None
+    # The default, is 'database' which means it's to do with talking
+    # to a PostgreSQL based implementation.
+    implementation_config_namespace = 'database'
 
     base_url = settings.MWARE_BASE_URL
     http_host = settings.MWARE_HTTP_HOST
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -1811,4 +1811,4 @@ class Reprocessing(SocorroMiddleware):
     )
 
     def post(self, **data):
-        return self.get_implementation().post(**data)
+        return self.get_implementation().reprocess(**data)
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -49,9 +49,9 @@ class DeprecatedModelError(DeprecationWarning):
 def config_from_configman():
     return configuration(
         definition_source=[
-            ElasticsearchConfig.required_config,
-            PostgreSQLCrashStorage.required_config,
-            ReprocessingOneRabbitMQCrashStore.required_config,
+            ElasticsearchConfig.get_required_config(),
+            PostgreSQLCrashStorage.get_required_config(),
+            ReprocessingOneRabbitMQCrashStore.get_required_config(),
 
             # This required_config defines the logger aggregate
             socorro_app.App.required_config,
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -51,11 +51,68 @@ def config_from_configman():
     # print "PostgreSQLCrashStorage", PostgreSQLCrashStorage.get_required_config().keys()
     # print "ReprocessingOneRabbitMQCrashStore", ReprocessingOneRabbitMQCrashStore.get_required_config().keys()
     # print "socorro_app.App.get_required_config()", socorro_app.App.get_required_config().keys()
-    definition_source = socorro_app.App.get_required_config()
+    # definition_source = socorro_app.App.get_required_config()
+    definition_source = Namespace()
+    # definition_source = socorro_app.App
 
     # definition_source.add_option(
     #     'app',
     #     default=socorro_app.App,
+    # )
+    definition_source.namespace('logging')
+
+    definition_source.logging.add_option(
+        'syslog_host',
+        doc='syslog hostname',
+        default='localhost',
+        reference_value_from='resource.logging',
+    )
+    definition_source.logging.add_option(
+        'syslog_port',
+        doc='syslog port',
+        default=514,
+        reference_value_from='resource.logging',
+    )
+    definition_source.logging.add_option(
+        'syslog_facility_string',
+        doc='syslog facility string ("user", "local0", etc)',
+        default='user',
+        reference_value_from='resource.logging',
+    )
+    definition_source.logging.add_option(
+        'syslog_line_format_string',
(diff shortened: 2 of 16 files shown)


peterbe:
@adngdb This is how you test this:

1) Set up a ssh tunnel to stage:
```
echo "localhost:5772 --> RMQ on stage"
ssh -Ag -i ~/.ssh/socorro__us-west-2.pem \
  -L 5772:***FIND ON CONSUL YOURSELF***:5672 \
  -p 22123 centos@ec2-54-200-76-190.us-west-2.compute.amazonaws.com
```
That domain you need you'll be able to find from doing:
`consulate kv get resource.rabbitmq.host` on a stage node.

2) Set things in your `.env`:
```
RABBITMQ_USER=***FIND ON CONSUL YOURSELF***
RABBITMQ_PASSWORD=***FIND ON CONSUL YOURSELF***
RABBITMQ_VIRTUAL_HOST=localhost
RABBITMQ_VIRTUAL_HOST=***FIND ON CONSUL YOURSELF***
RABBITMQ_PORT=5772
```

3) Start your webapp, find a crash ID from a Reports tab somewhere. Look at it's "Processor Notes" under the main Report Index page. 

4) Go the admin. Click Re-processing. Paste in the crash ID. (Also, try to paste in junk)

5) Wait a bit. Then reload the Report Index page and see how the Processor Notes should have changed.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -234,10 +238,10 @@ def _consume_acknowledgement_queue(self):
...
                 crash_id_to_be_acknowledged = \
                     self.acknowledgment_queue.get_nowait()
-                #self.config.logger.debug(
-                    #'RabbitMQCrashStorage set to acknowledge %s',
-                    #crash_id_to_be_acknowledged
-                #)
+                # self.config.logger.debug(
+                #     'RabbitMQCrashStorage set to acknowledge %s',
+                #     crash_id_to_be_acknowledged
+                # )
```
peterbe: Oops. Ignore this change. I couldn't help myself but I later realized the whole file is full of pep8 violations that I don't want to fix now.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -258,7 +262,7 @@ def _consume_acknowledgement_queue(self):
                         crash_id_to_be_acknowledged,
                         exc_info=True
                     )
-                except Exception as x:
+                except Exception:
```
peterbe: Same point here. I started pep8 fixing but abdandoned after two small changes.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -299,3 +303,33 @@ class ReprocessingRabbitMQCrashStore(RabbitMQCrashStorage):
         False
     )
 
+
+#==============================================================================
+class ReprocessingOneRabbitMQCrashStore(ReprocessingRabbitMQCrashStore):
```
peterbe: Note that 20 lines above there's another class called `ReprocessingOneRabbitMQCrashStore` that is used by [`reprocessingjobs.py`](https://github.com/mozilla/socorro/blob/master/socorro/cron/jobs/reprocessingjobs.py#L20-L23) and done with some magic I don't understand. 
So I added a new class so I can go into the `__init__` and mess with the `config` object.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -299,3 +303,33 @@ class ReprocessingRabbitMQCrashStore(RabbitMQCrashStorage):
...
+        RabbitMQCrashStorage,
+        'routing_key',
+        'socorro.reprocessing'
+    )
+
+    def __init__(self, config, quit_check_callback=None):
+        config.forbidden_keys = ''
+        config.redactor_class = Redactor
+        config.throttle = 100
+        config.filter_on_legacy_processing = False  # must be set
```
peterbe: These 4 lines I just don't understand why they're not set by default. Grrr...
You'd think that making an instance of this class would create a `config` that has the values of its parent class by default. 
If you delete any of these lines you get subtle `KeyError`s further down.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -299,3 +303,33 @@ class ReprocessingRabbitMQCrashStore(RabbitMQCrashStorage):
...
+        super(ReprocessingRabbitMQCrashStore, self).__init__(
+            config, quit_check_callback=quit_check_callback
+        )
+
+    def post(self, crash_id):
```
peterbe: Note the use of the word "post" here. It lends itself nicely to one day directly linking this to the web API.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -48,6 +51,8 @@ def config_from_configman():
         definition_source=[
             ElasticsearchConfig.required_config,
             PostgreSQLCrashStorage.required_config,
+            ReprocessingOneRabbitMQCrashStore.required_config,
```
peterbe: here's the magic so that it picks up the host, password etc from `SOCORRO_IMPLEMENTATIONS_CONFIG`.

twobraids: _ah ha!_  these are the problem, not the change_default that we discussed earlier.  

1) In each one of these lines, you're referencing the class var ```required_config``` of the derived class.  This ignores the ```required_config``` of any base classes.  The quick fix is to replace the ```.required_config``` with ```.get_required_config()```.  That latter function walks the inheritance hierarchy to get a union of all the required config from every level.

2) however, you've got a second problem.  By having both PG and RMQ at the same level you're going to have name collisions in the options (both have 'transaction_executor_class', for example)  Each of these ought to be in namespaces.  Do something like this:

``` python
definition_source = Namespace()
definition_source.namespace('search')
definition_source.search.add_option(
     'search_class',
     default='socorro.external.es.base.ElasticsearchConfig',
)
definition_source.namespace('database')
definition_source.database.add_option(
     'database_storage_class',
     default='socorro.external.postgresql.crashstorage.PostgreSQLCrashStorage',
)
definition_source.namespace('queuing')
definition_source.queuing.add_option(
     'queuing_class',
     default='socorro.external.rabbitmq.crashstorage.ReprocessingOneRabbitMQCrashStore',
)
```
Anywhere that you reference these values, you then should refer to them with their namespaces:

``` python
# to create the ES object
search_object = config.search.search_class(config.search)
...
```

socorro/external/rabbitmq/crashstorage.py
peterbe: @twobraids This problem went away when I switch to `.get_required_config()`

webapp-django/crashstats/crashstats/models.py
peterbe: That's unfortunate. Doesn't appear to be a problem. SuperSearch works, the reprocessing works, and I can talk to postgres just fine. Maybe we're just lucky at the moment. 

The problem with referencing these is the way that config gets used. We essentially do this:
```python
implemenation = some_class(config_from_configman())
return implemenation.get(**request.GET)
```
Basically, we don't know if `some_class` is postgres, es or rabbitmq. 
We could mark up each class to tell which namespace to seek inside. E.g. instead of
```python
class SuperSearch(SocorroMiddleware):
    implemenation = socorro.external.es.supersearch.SuperSearch
```
We do
```python
class SuperSearch(SocorroMiddleware):
    implemenation = socorro.external.es.supersearch.SuperSearch
    implementation_namespace = 'es'
```

peterbe: But should I bother? All is working.

peterbe: I can't get this to work. So much fail. Instead of 
```
implemenation = some_class(config_from_configman())
```
I tried to do 
```
implemenation = some_class(config_from_configman()['database'])
```
but then I get 
```
  File "/opt/boxen/homebrew/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/wsgiref/handlers.py", line 85, in run
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/crashstats/decorators.py", line 51, in inner
    versions
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/crashstats/utils.py", line 247, in build_default_context
    for pv in api.get(active=True)['hits']:
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/crashstats/models.py", line 445, in get
    return self._get(expect_json=expect_json, **kwargs)
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/crashstats/models.py", line 486, in _get
    implementation = self.get_implementation()
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/crashstats/models.py", line 400, in get_implementation
    config=config_from_configman()['database']
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/socorro/external/postgresql/base.py", line 47, in __init__
    self.context
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/socorro/external/postgresql/crashstorage.py", line 106, in __init__
    quit_check_callback=quit_check_callback
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/socorro/external/crashstorage_base.py", line 206, in __init__
    self.logger = config.logger
  File "/Users/peterbe/virtualenvs/socorro/lib/python2.7/site-packages/configman/dotdict.py", line 335, in __getattr__
    return getattr(self._parent, key)
ReferenceError: weakly-referenced object no longer exists
```

The defintion_source is set up like this:
```
    definition_source = Namespace()
    definition_source.namespace('es')
    definition_source.es.add_option(
         'es',
         default=ElasticsearchConfig,
    )
    definition_source.namespace('database')
    definition_source.database.add_option(
         'database_class',
         default=PostgreSQLCrashStorage,
    )
    definition_source.namespace('rabbitmq')
    definition_source.rabbitmq.add_option(
         'rabbitmq',
         default=ReprocessingOneRabbitMQCrashStore,
    )
    # definition_source.add_option(
    #     'app',
    #     default=socorro_app.App,
    # )
    return configuration(
        definition_source=[definition_source,socorro_app.App.get_required_config()],
        # definition_source=[
        #     ElasticsearchConfig.get_required_config(),
        #     PostgreSQLCrashStorage.get_required_config(),
        #     ReprocessingOneRabbitMQCrashStore.get_required_config(),
        #
        #     # This required_config defines the logger aggregate
        #     socorro_app.App.get_required_config(),
        # ],
        values_source_list=[
            settings.SOCORRO_IMPLEMENTATIONS_CONFIG,
        ]
    )
```

socorro/external/rabbitmq/crashstorage.py
adngdb: I don't like it. ``crashstorage`` classes should never be directly exposed in the API. In the new world of separate repos, this will be a part of ``socorrolib``, not of ``webapp`` or ``middleware``.

webapp-django/crashstats/manage/views.py
```diff
@@ -1033,3 +1034,32 @@ def crash_me_now(request):
...
+            crash_id = form.cleaned_data['crash_id']
+            url = reverse('manage:reprocessing')
+            if Reprocessing().post(crash_id=crash_id):
+                url += '?crash_id={}'.format(crash_id)
+                messages.success(
+                    request,
+                    '{} sent in for reprocessing.'.format(crash_id)
+                )
+            else:
+                messages.success(
```
adngdb: ``success`` for showing an error?

webapp-django/crashstats/manage/views.py
```diff
@@ -1033,3 +1034,32 @@ def crash_me_now(request):
...
+                    request,
+                    'Currently unable to send in the crash ID '
+                    'for reprocessing.'
+                )
+            return redirect(url)
```
adngdb: You always redirect. Don't you want to stay on the same page when the reprocessing failed?

peterbe: Yes because it's not our fault here. The pattern of not redirecting is used when there's something wrong with the form validation.

webapp-django/crashstats/manage/views.py
peterbe: fixed.

peterbe:
@adngdb So some big changes since you last looked at this. In particular, in the `config_from_configman` method. The big deal now is that that config is **namespaced**. Before it just to be all "in the flat root" so to say. But it also means that we need to know which part of that namespaced config to send into the implementation classes. You'll notice that now every "model" needs to be aware of which "namespace" it belongs to. It's just a string. The default one is here: https://github.com/mozilla/socorro/pull/3265/files#diff-0aa95732d6b7aebcee83eabf22da5dc7R426

Without this change we can't add more things such as `socorro.external.boto.BotoS3CrashStorage` easily.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -299,3 +299,29 @@ class ReprocessingRabbitMQCrashStore(RabbitMQCrashStorage):
...
+    required_config.routing_key = change_default(
+        RabbitMQCrashStorage,
+        'routing_key',
+        'socorro.reprocessing'
+    )
+
+    def __init__(self, config, quit_check_callback=None):
+        super(ReprocessingRabbitMQCrashStore, self).__init__(
+            config, quit_check_callback=quit_check_callback
+        )
```
adngdb: What's the point of skipping the ``ReprocessingRabbitMQCrashStore.__init__`` method? If there is a good reason, can you add a comment to explain it?

socorro/external/rabbitmq/crashstorage.py
adngdb: I still don't like this name. I think it doesn't make sense in the context of a crash storage class. Someone from the outside reading this will wonder what this function does.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -44,17 +47,37 @@ class DeprecatedModelError(DeprecationWarning):
...
+        default=ElasticsearchConfig,
+    )
+    definition_source.namespace('database')
+    definition_source.database.add_option(
+        'database_storage_class',
+        default=PostgreSQLCrashStorage,
+    )
+    definition_source.namespace('queuing')
+    definition_source.queuing.add_option(
+        'rabbitmq',
```
adngdb: That's quite vague. Why not something more precise like ``rabbitmq_reprocessing_class``? When I see just ``rabbitmq`` I expect a ConnectionContext.

webapp-django/crashstats/manage/jinja2/manage/reprocessing.html
```diff
@@ -0,0 +1,40 @@
...
+        {% endif %}
+
+        <form action="" method="post">
+            {% csrf_token %}
+            <table class="data-table">
+                {{ form }}
+                <tr>
+                    <th>&nbsp;</th>
+                    <td>
+                        <input type="submit" value="Let's do this!">
```
adngdb: That's not a very good button value. Why not use something more straightforward like ``Reprocess``?

adngdb: I know it's bikeshedding because that's the admin panel and there's only a handful of people using it, but it's generally a good idea to keep button labels very explicit.

webapp-django/crashstats/manage/tests/test_views.py
```diff
@@ -1888,3 +1888,51 @@ def test_symbols_uploads_data_pagination_bad_request(self):
...
+        def mocked_post(crash_id):
+            if crash_id == good_crash_id:
+                return True
+            elif crash_id == bad_crash_id:
+                return
+            raise NotImplementedError(crash_id)
+
+        models.Reprocessing.implementation().post = mocked_post
+
+        self._login()
```
adngdb: Is there value in having a test when not logged in? Or maybe that's already covered by other tests?

adngdb:
Tested locally, it worked! Still a few questions though. :)

socorro/external/rabbitmq/crashstorage.py
peterbe: Skipping? Hmm... I don't see that. What I do see is that there's no reason for this local `__init__` to exist since it just calls the parent class without anything special.

socorro/external/rabbitmq/crashstorage.py
peterbe: What do you suggest instead? The `ReprocessingRabbitMQCrashStore` class, which I'm subclassing, was already there.

webapp-django/crashstats/crashstats/models.py
peterbe: You're right. I don't like the name either. Let's see if we can change to `queuing`.

socorro/external/rabbitmq/crashstorage.py
adngdb: It actually calls the parent's parent, since the class is named ``ReprocessingOneRabbitMQCrashStore`` and its parent is named ``ReprocessingRabbitMQCrashStore``. Note that the first argument to ``super`` is the latter, so what you're calling is the ``__init__`` of the parent of ``ReprocessingRabbitMQCrashStore`` (without 'One'). 

Or I completely misunderstood how ``super`` works, in which case you have to tell me because that would be quite a flow in my education. :)

socorro/external/rabbitmq/crashstorage.py
adngdb: It's just ``post`` that I dislike. Something like ``reprocess`` would be a lot more accurate to me.

webapp-django/crashstats/manage/jinja2/manage/reprocessing.html
peterbe: Done. You're right.

webapp-django/crashstats/manage/tests/test_views.py
peterbe: There's not really any code that does it. We just have a django decorator. But it's not bad to have a little bit of basic testing to double-check that we really have that decorator on.

webapp-django/crashstats/crashstats/models.py
peterbe: Corrected.

socorro/external/rabbitmq/crashstorage.py
adngdb: Still an issue imho. :)

socorro/external/rabbitmq/crashstorage.py
adngdb: And you did not answer this.

adngdb:
There are 2 questions in the way of your r+! :)

adngdb:
Oops, sorry, my reworking of the admin home page created a conflict with this. I recommend you use the icon ``refresh`` for that Reprocessing page you're adding.

socorro/external/rabbitmq/crashstorage.py
peterbe: Most likely my oversight. Will tend to.

socorro/external/rabbitmq/crashstorage.py
peterbe: Yeah, `reprocess` is a better name. Changed.

peterbe:
Review addressed and rebased. Just waiting for travis (hint hint @adngdb ) now.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -315,11 +315,11 @@ class ReprocessingOneRabbitMQCrashStore(ReprocessingRabbitMQCrashStore):
     )
 
     def __init__(self, config, quit_check_callback=None):
-        super(ReprocessingRabbitMQCrashStore, self).__init__(
+        super(ReprocessingOneRabbitMQCrashStore, self).__init__(
             config, quit_check_callback=quit_check_callback
         )
```
adngdb: Unless I'm missing something, you can actually delete this ``__init__`` method, it does nothing the default doesn't do.

adngdb:
Still failing sadly :(

socorro/external/rabbitmq/crashstorage.py
peterbe: I love code reviews! Thank you!

peterbe:
Sigh. Hopefully fixed all tests now.

PR #3266, opened 2016-04-04
fixes bug 1214033 - Session CSRF cookies not needed for API views

peterbe:
@willkg you're pretty good at these kinds of deeper Django security things. r?

### DIFF
--- a/webapp-django/crashstats/api/tests/test_views.py
+++ b/webapp-django/crashstats/api/tests/test_views.py
@@ -11,6 +11,8 @@ import mock
 import pyquery
 from nose.tools import eq_, ok_
 
+from session_csrf import ANON_COOKIE
+
 from socorro.external import BadArgumentError, MissingArgumentError
 from crashstats.base.tests.testbase import TestCase
 from crashstats.crashstats.tests.test_views import (
@@ -170,6 +172,28 @@ class TestViews(BaseTestViews):
         response = self.client.get(url)
         eq_(response.status_code, 200)
         eq_(response['Access-Control-Allow-Origin'], '*')
+
+    def test_no_cookie_set(self):
+
+        def mocked_get(**options):
+            return {}
+
+        # Doesn't matter which model we use, as long as we use one of them.
+        CrontabberState.implementation().get.side_effect = mocked_get
+        url = reverse('api:model_wrapper', args=('CrontabberState',))
+        # self.client.handler.enforce_csrf_checks = True
+        response = self.client.get(url)
+        eq_(response.status_code, 200)
+        # The test client sucks up all set cookies. This one should not
+        # be set.
+        ok_(ANON_COOKIE not in response.cookies)
+
+        # But load the regular web page.
+        # The documentation for the API
+        url = reverse('api:documentation')
+        response = self.client.get(url)
+        eq_(response.status_code, 200)
+        ok_(ANON_COOKIE in response.cookies)
 
     @mock.patch('requests.get')
     def test_CrashesPerAdu_too_much(self, rget):
--- a/webapp-django/crashstats/base/middleware.py
+++ b/webapp-django/crashstats/base/middleware.py
@@ -1,0 +1,24 @@
+from session_csrf import CsrfMiddleware
+
+
+class ExtendedCsrfMiddleware(CsrfMiddleware):
+    """The only difference between this class and the base class
+    from session_csrf is that it's possible to make it not set a
+    cookie on the response. That basically makes it impossible to use
+    CSRF but for certain requests that's perfectly fine.
+
+    Note: The inbound cookie on the request is still valid and might
+    help the user get additional access.
+
+    The companion decorator to leverage this is
+    crashstats.base.decorators.no_csrf_cookie
+    Use that on your views and no anonymous CSRF cookie will be set
+    on the response.
+    """
+
+    def process_response(self, request, response):
+        if getattr(request, 'no_session_csrf_cookie', None):
+            return response
+        return super(ExtendedCsrfMiddleware, self).process_response(
+            request, response
+        )
--- a/webapp-django/crashstats/api/views.py
+++ b/webapp-django/crashstats/api/views.py
@@ -17,6 +17,7 @@ from waffle.decorators import waffle_switch
 
 import crashstats
 from socorro.external import BadArgumentError, MissingArgumentError
+from crashstats.base.decorators import no_csrf_cookie
 from crashstats.crashstats import models
 from crashstats.crashstats import utils
 from crashstats.tokens.models import Token
@@ -195,6 +196,7 @@ def has_permissions(user, permissions):
     return True
 
 
+@no_csrf_cookie
 @waffle_switch('!app_api_all_disabled')
 @ratelimit(
     key='ip',
--- a/webapp-django/crashstats/base/decorators.py
+++ b/webapp-django/crashstats/base/decorators.py
@@ -1,0 +1,12 @@
+import functools
+
+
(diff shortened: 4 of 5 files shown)


peterbe:
The way to test this is perhaps to run `curl -v https://socorro.dev/api/CrontabberState/` and there should be no `Set-Cookie` response header.

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -171,6 +173,28 @@ def mocked_get(url, params, **options):
...
 
+    def test_no_cookie_set(self):
+
+        def mocked_get(**options):
+            return {}
+
+        # Doesn't matter which model we use, as long as we use one of them.
+        CrontabberState.implementation().get.side_effect = mocked_get
+        url = reverse('api:model_wrapper', args=('CrontabberState',))
+        # self.client.handler.enforce_csrf_checks = True
```
willkg: Should this line get removed?

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -171,6 +173,28 @@ def mocked_get(url, params, **options):
...
+        response = self.client.get(url)
+        eq_(response.status_code, 200)
+        # The test client sucks up all set cookies. This one should not
+        # be set.
+        ok_(ANON_COOKIE not in response.cookies)
+
+        # But load the regular web page.
+        # The documentation for the API
+        url = reverse('api:documentation')
```
willkg: This PR is about nixing the anon csrf stuff for API views, but isn't this an API view? It reads confusingly.

webapp-django/crashstats/base/middleware.py
```diff
@@ -0,0 +1,24 @@
...
+    Use that on your views and no anonymous CSRF cookie will be set
+    on the response.
+    """
+
+    def process_response(self, request, response):
+        if getattr(request, 'no_session_csrf_cookie', None):
+            return response
+        return super(ExtendedCsrfMiddleware, self).process_response(
+            request, response
+        )
```
willkg: I think we should implement this "don't session-csrf-me plz!" thing in django-session-csrf. I would have liked to have had something like this in Input, but I never tracked the problem down far enough to figure out where to fix it. I'm pretty sure we had similar problems with SUMO, too.

webapp-django/crashstats/api/tests/test_views.py
peterbe: Ah. Confusing perhaps. The URL `api:documentation` is a regular django HTML view function. It's the documentation page.

webapp-django/crashstats/api/tests/test_views.py
peterbe: Certainly should.

webapp-django/crashstats/base/middleware.py
peterbe: Could do. Apparently I'm good friends with the last guy who committed on the project: https://github.com/mozilla/django-session-csrf/commits/master
However, I'm not entirely sure I'm that eager to work on it.

willkg: I think it's a small change, though then we have to push out a new release. I think it's worth it, though.

I'm happy to do the leg work, but I don't think I can push new releases to PyPI for django-session-csrf.

willkg:
I've thought about this and looked at the django-session-csrf code. This fixes the code so that we nix the cookie in the HTTP Response, but django-session-csrf has already done a bunch of work and prepped a csrf token and cached it and all that. Seems like for these views, we want to disable the whole anonymous csrf thing rather than just nixing the cookie at the end.

Maybe a better approach is to change django-session-csrf to either blacklist all the urls that we shouldn't be doing anonymous csrf tokens for or whitelist the urls that we should?

webapp-django/crashstats/base/middleware.py
peterbe: I have pypi owner access to django-session-csrf. If you're willing to do the legwork, I'd be happy to review and make a release.

peterbe:
> Seems like for these views, we want to disable the whole anonymous csrf thing rather than just nixing the cookie at the end.

django-session-csrf has some decorators that exempts CSRF from being checked. But that doesn't prevent a cookie from being set on the response. Only a change to its code would do that, I think. 

> Maybe a better approach is to change django-session-csrf to either blacklist all the urls that we shouldn't be doing anonymous csrf tokens for or whitelist the urls that we should?

I'd MUCH rather set an explicit decorator on the views that I want to stay the heck out of CSRF middleware. 

By the way, I think we'll soon need to use that built in django `@csrf_excempt` decorator since we'll soon have a Web API endpoint that you'll be able to PUT/POST to (with certain auth-token permissions).

willkg:
I spent some *quality time* reading through the django-session-csrf code and I think this change isn't something we want to do as is.

django-session-csrf generates a csrf token and caches it during `process_request` if there is no cookie that indicates there's already a csrf token.

https://github.com/mozilla/django-session-csrf/blob/master/session_csrf/__init__.py#L59

In `process_response`, it throws the csrf token in a cookie if it isn't already there.

https://github.com/mozilla/django-session-csrf/blob/master/session_csrf/__init__.py#L114

If we make the change in this PR as it currently stands, then we end up with a situation where every API HTTP request generates a csrf token and caches it as opposed to the just the first API HTTP request per user. I don't think we really want to do that.

Assuming all that's right (a second pair of eyes would be super), then I think our options are these:

1. drop django-session-csrf altogether--does socorro need it? what problem is it really solving?
2. fix django-session-csrf so that it's doing all the work in `process_view` and pays attention to Django's `csrf_exempt` and related decorators
3. learn to be ok with anonymous API HTTP requests having cookies--no one will starve or die from this (probably)
4. give up this whole software developer business and become shepherds of cats
5. some fifth option that didn't occur to me

peterbe:
I read the [Django documentation](https://docs.djangoproject.com/en/1.8/ref/csrf/) again and I've concluded that option 1 is not an option. 
Using [CSRF_COOKIE_DOMAIN](https://docs.djangoproject.com/en/1.8/ref/settings/#csrf-cookie-domain) we can be strict and hardcode it to specifically `crash-stats.mozilla.com`. That means that a CSRF cookie generate on our site can't be reused at `othersite.mozilla.com`. 
However, Django (1.8) can NOT protect us from the fact that a user managed to get a cookie over at `beenhacked.mozilla.com` and leverage that here on our site. That's what `session_csrf` solves by controlling the cookie values in *our* [session] server. 

Option 3 is tempting. I filed [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1214033) not because we were in pain or security weak or anything like that. 
The more I think about it, the more I realize that the `Set-Cookie` header in the `/api/*` responses is just harmless noise. 

What matters to me more is excess work done on our server when it's never going to lead to anything. 
I.e. unnecessary entries in django_sessions. However, I don't think they get generated at all for anonymous GETs. I think they only get generated whenever Django needs to send a `Set-Cookie: sessionid=xxxx` in the response. [More about that here](https://bugzilla.mozilla.org/show_bug.cgi?id=1262424).

We COULD redirect our energy into session_csrf and make it better so that it becomes possible to decorate certain functions that should just be left alone. 

So if you don't mind. Let's shut this down and work on https://bugzilla.mozilla.org/show_bug.cgi?id=1262424 instead. What do you think?

willkg:
I like the cut of your jib! I would eagerly subscribe to your podcast.

peterbe:
Bigger fish to fry in https://bugzilla.mozilla.org/show_bug.cgi?id=1262424

PR #3272, opened 2016-04-06
Fixes bug 1262137 - Moved topcrashers static files to new app.

adngdb:


webapp-django/crashstats/settings/bundles.py
```diff
@@ -222,6 +228,12 @@
         ),
         'output_filename': 'js/topcrash.min.js',
     },
+    'new_topcrashers': {
```
peterbe: I don't like this name. We should change the old one from `topcrashers` to `old_topcrashers`. 
Otherwise, when you one day, remove the old top crasher, calling what's left "new" doesn't make sense.

peterbe:
r+ with the nit about calling the current thing "new".

webapp-django/crashstats/settings/bundles.py
adngdb: Much much smarter!

PR #3276, opened 2016-04-07
color in the webqspassing tool

peterbe:
Not an important change at all. I just wanted to see how easy it was to add color to the output
screen shot 2016-04-07 at 1 30 03 pm

### DIFF
--- a/tools/webqapassing.py
+++ b/tools/webqapassing.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
+import os
 import xml.etree.ElementTree as ET
 
 import requests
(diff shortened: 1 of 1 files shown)


willkg:
So, terminal emulation, shells and colors is a HUGE can of worms. I highly encourage you to use a library to do it that will handle the variety of terminal types, what they support, tty modes and all the other things.

I've seen colorama used a lot. I used blessings. There's a blessings fork called blessed that is like blessings but has some differences but I think is better maintained at this point. There are probably a bunch of others.

peterbe:
I looked at [termcolor](https://pypi.python.org/pypi/termcolor) and copied the critical parts. I don't want to pollute requirements.txt for such a small lib and for a tool that's not part of production.

willkg:
Yeah... that lib is completely lacking all the stuff to deal with the complexities of ansi escape code support.

Look at blessings for better inspiration:

https://github.com/erikrose/blessings/blob/master/blessings/__init__.py

That handles terms, ttys, encodings and other things.

So the risk here is that you've built something that works on your machine with your specific shell setup the way you've tested it. But it won't work using shell redirection and may not work on other peoples' machines as they differ in shells, locales, terminal emulation support, etc.

peterbe:
Fair enough. I didn't know that was so complex.

PR #3277, opened 2016-04-07
filtering on signaturesummary api results

peterbe:
cc @lonnen

### DIFF
--- a/webapp-django/crashstats/api/tests/test_views.py
+++ b/webapp-django/crashstats/api/tests/test_views.py
@@ -1855,3 +1855,97 @@ class TestViews(BaseTestViews):
         response = self.client.get(url, {'product': 'foobaz'})
         eq_(response.status_code, 400)
         ok_('foobaz' in response.content)
+
+    @mock.patch('requests.get')
+    def test_SignatureSummary_exploitability(self, rget):
+
+        def mocked_get(url, params, **options):
+            assert 'signaturesummary' in url, url
+            reports = {}
+            if 'exploitability' in params['report_types']:
+                reports['exploitability'] = [
+                    {
+                        'high_count': 0,
+                        'low_count': 1,
+                        'medium_count': 0,
+                        'none_count': 0,
+                        'null_count': 0,
+                        'report_date': '2016-04-02'
+                    },
+                    {
+                        'high_count': 0,
+                        'low_count': 0,
+                        'medium_count': 0,
+                        'none_count': 1,
+                        'null_count': 0,
+                        'report_date': '2016-03-31'
+                    }
+                ]
+            if 'uptime' in params['report_types']:
+                reports['uptime'] = [
+                    {
+                        'version_string': '12.0',
+                        'percentage': '48.440',
+                        'report_count': 52311,
+                        'product_name': 'WaterWolf',
+                        'category': 'XXX'
+                    },
+                    {
+                        'version_string': '13.0b4',
+                        'percentage': '9.244',
+                        'report_count': 9983,
+                        'product_name': 'WaterWolf',
+                        'category': 'YYY'
+                    }
+                ]
+            return Response({'reports': reports})
+
+        rget.side_effect = mocked_get
+
+        url = reverse('api:model_wrapper', args=('SignatureSummary',))
+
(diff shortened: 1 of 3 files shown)


webapp-django/crashstats/api/views.py
```diff
@@ -258,7 +259,7 @@ def model_wrapper(request, model_name):
     form = FormWrapper(model, request.REQUEST)
     if form.is_valid():
         try:
-            result = function(**form.cleaned_data)
+            result = function(**dict(form.cleaned_data, _user=request.user))
```
willkg: Pretty sure you can just do:

```
result = function(_user=request.user, **form.cleaned_data)
```

You don't need to form a dict and then de-dict it.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -1376,6 +1376,15 @@ class SignatureSummary(SocorroMiddleware):
         'reports',
     )
 
+    def get(self, *args, **kwargs):
+        # You're not allowed to view the exploitability report if you
+        # don't have permission to do so.
+        user = kwargs['_user']
```
willkg: If you don't want `_user` to stay in kwargs, you can use `.pop`:

```
user = kwargs.pop('_user')
```

I mention that in case you don't want to pass it up to the superclass `.get()`.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -1376,6 +1376,15 @@ class SignatureSummary(SocorroMiddleware):
...
+        if not user.has_perm('crashstats.view_exploitability'):
+            if 'exploitability' in kwargs['report_types']:
+                kwargs['report_types'].remove('exploitability')
```
willkg: I applaud your ability to spell "exploitability" twice in a row exactly the same. :clap:

peterbe: Blessed is he who haths good auto-completion in Atom.

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -1855,3 +1855,97 @@ def mocked_supersearch_get(**params):
         response = self.client.get(url, {'product': 'foobaz'})
         eq_(response.status_code, 400)
         ok_('foobaz' in response.content)
+
+    @mock.patch('requests.get')
+    def test_SignatureSummary_exploitability(self, rget):
+
+        def mocked_get(url, params, **options):
```
willkg: This is mocking requests.get which is called to do an HTTP GET from the middleware, right? Or is it for something else?

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -1855,3 +1855,97 @@ def mocked_supersearch_get(**params):
...
+            'start_date': '2012-1-1',
+            'end_date': '2013-1-1',
+        })
+        eq_(response.status_code, 200)
+        dump = json.loads(response.content)
+        ok_(dump['reports'])
+        ok_('exploitability' not in dump['reports'])
+        ok_('uptime' in dump['reports'])
+
+        # let's sign in
```
willkg: If I were writing this, I'd have the authenticated-but-lacking-permission situation as a separate test.

Also, is there something that tests the authenticated-and-has-permission situation?

webapp-django/crashstats/api/tests/test_views.py
peterbe: That is correct.

webapp-django/crashstats/api/tests/test_views.py
peterbe: I know it's a bit lazy to do multiple things in one test but I actually don't mind here in this case. The mocking business is quite cumbersome. So having to replicate that for the various access right states feels unnecessary. 

Here I can test the following states
1) completely anonymous user calling it
2) signed in user without perms calling it
3) signed in user with permission calling it.

webapp-django/crashstats/crashstats/models.py
peterbe: I did this differently now.

willkg:
I think this looks ok. r+

PR #3278, opened 2016-04-08
fixes bug 1263242 - Allow api/ProductVersions/ to filter by build_type

peterbe:


willkg:
The tests pass and this looks fine to me as far as I can tell.

Do we have documentation for these APIs that needs to get updated?

peterbe:
> Do we have documentation for these APIs that needs to get updated?

No. We have the opportunity to write a doc string on any of the classes in `crashstats/models.py` but if I were to attempt to document and explain what the different values can be for this field, it opens a can of worms in that there are many many other places where the possible values aren't explained.

willkg:
How do people who use these APIs figure out how they work? Are these "unofficial" API endpoints?

willkg:
Anyhow, this looks ok as far as I can tell. r+

peterbe:
> How do people who use these APIs figure out how they work? Are these "unofficial" API endpoints?

Our APIs are an afterthought. I'm not sure if you want to waste 50 min of your life on this: https://air.mozilla.org/using-the-crash-stats-api/

But the story is that we used to have 2 things. The PHP website and a web.py "middleware" service. The PHP site would get its data by doing phpcurl (or whatever it's called). Later, basically all the stuff that was in that middleware service was exposed to the web. We added a cute [autogenerated documentation](https://crash-stats.mozilla.com/api/) that had the benefit that you can sort of test the API by typing in stuff into the boxes and press the "Run Test Drive!" button. 

Basically, the API was built to serve a website. Not to serve people who want access to the data. 

Another mistake we made was that we never versioned the URLs. 

I'd be interested in making a v2 whereby we only expose a couple of few ones that we really want external users to use. But the win isn't huge. The only pain we're in right now is that it looks "kinda ugly" that we have an API *without documentation*. I think you felt that, didn't you? :)

PR #3280, opened 2016-04-09
Modify Breakpad build scripts to build in TaskCluster

luser:
This is bug 1230659. There are two parts to this:
1) Fix build-breakpad.sh to check Breakpad out of the git repo that is its canonical home nowadays using gclient, since that's how you do it.
2) Add a breakpad-taskcluster.sh script that can be run in Taskcluster that configures the environment properly (needs to download a newer GCC etc) and runs build-breakpad.sh.

I split breakpad-taskcluster out as a separate script because the Socorro build scripts will try to invoke build-breakpad.sh for a local build that's not on Linux, so I figured I'd keep the Taskcluster-specific bits separate.

This isn't going to be set up to run as CI anywhere right now, but at least having a one-click way to get a new build of Breakpad is useful.
r? @lonnen or @rhelmer

### DIFF
--- a/scripts/breakpad-taskcluster.sh
+++ b/scripts/breakpad-taskcluster.sh
@@ -1,0 +1,78 @@
+#!/bin/bash
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This script can be used to build a new breakpad.tar.gz for Socorro
+# under Taskcluster. It is not currently used in an automated fashion,
+# you must manually submit a task to get a new build. The simplest way
+# is to paste the contents of the following here document into
+# the task-creator tool: https://tools.taskcluster.net/task-creator/
+#
+# This task will update a Taskcluster index if it succeeds, such that
+# the most recent tarball can be fetched from:
+# https://index.taskcluster.net/v1/task/project.socorro.breakpad.v1.builds.linux64.latest/artifacts/public/breakpad.tar.gz
+#
+# You must be a member of the `socorro` project in Taskcluster for this
+# task to work properly.
+: <<'EOF'
+{
+	"created": "2016-04-09T00:49:49.410Z",
+	"deadline": "2016-04-09T01:49:49.410Z",
+	"provisionerId": "aws-provisioner-v1",
+	"workerType": "emulator-jb",
+	"retries": 0,
+	"expires": "2017-04-08T22:33:21.202Z",
+        "routes": [
+            "index.project.socorro.breakpad.v1.builds.linux64.latest",
+        ],
+        "scopes": [
+            "queue:route:index.project.socorro.breakpad.v1.builds.linux64.latest"
+        ],
+	"payload": {
+		"image": "taskcluster/desktop-build:0.1.11",
+		"command": [
+			"/bin/sh",
+			"-c",
+			"curl \"https://raw.githubusercontent.com/${SOCORRO_FORK:=mozilla/socorro}/${SOCORRO_BRANCH:=master}/scripts/breakpad-taskcluster.sh\" | bash"
+		],
+		"artifacts": {
+			"public/breakpad.tar.gz": {
+				"type": "file",
+				"path": "/home/worker/breakpad.tar.gz",
+			}
+		},
+		"maxRunTime": 7200
+	},
+	"metadata": {
+		"name": "Build Breakpad",
+		"description": "Build Breakpad for Socorro consumption",
+		"owner": "ted@mielczarek.org",
+		"source": "http://tools.taskcluster.net/task-creator/"
+	}
+}
+EOF
+
+set -v -e -x
+
+# This script runs in the desktop-build image, but its default python is
+# ancient. It does have Python 2.7 installed.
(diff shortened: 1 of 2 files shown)


scripts/breakpad-taskcluster.sh
```diff
@@ -0,0 +1,61 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# TaskCluster build script for building Breakpad.
+# Submit a task like this to build:
+: <<'EOF'
+{
+	"created": "2016-04-09T00:49:49.410Z",
+	"deadline": "2016-04-09T01:49:49.410Z",
+	"provisionerId": "aws-provisioner-v1",
+	"workerType": "emulator-jb",
```
rhelmer: Is `emulator-jb` the right `workerType`?

scripts/build-breakpad.sh
```diff
@@ -4,42 +4,39 @@
...
 echo "PREFIX: ${PREFIX:=`pwd`/build/breakpad}"
-svn co http://google-breakpad.googlecode.com/svn/trunk google-breakpad
-cd google-breakpad
+mkdir breakpad
+cd breakpad
+fetch breakpad
+cd src
 mkdir -p ${PREFIX}
-rsync -a --exclude="*.svn" ./src ${PREFIX}/
+rsync -a --exclude="*.git" --exclude="*.svn" ./src ${PREFIX}/
```
rhelmer: `*.svn` exclude still needed? breakpad was the only thing coming from SVN AFAIK

lonnen:
@luser -- what references this file outside of our repo? @willkg may partition the processor out to it's own repository and all this stackwalker would go with it

scripts/breakpad-taskcluster.sh
```diff
@@ -0,0 +1,61 @@
...
+		"command": [
+			"/bin/sh",
+			"-c",
+			"curl \"https://raw.githubusercontent.com/${SOCORRO_FORK:=mozilla/socorro}/${SOCORRO_BRANCH:=master}/scripts/breakpad-taskcluster.sh\" | bash"
+		],
+		"artifacts": {
+			"public/breakpad.tar.gz": {
+				"type": "file",
+				"path": "/home/worker/breakpad.tar.gz",
+				"expires": "2016-12-04T18:46:14.355Z"
```
lonnen: If we remove the value it will default to 1 year from the deadline date, which is probably good? Alternatively, consider hardcoding it into the far future.

scripts/breakpad-taskcluster.sh
```diff
@@ -0,0 +1,61 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# TaskCluster build script for building Breakpad.
+# Submit a task like this to build:
+: <<'EOF'
+{
+	"created": "2016-04-09T00:49:49.410Z",
+	"deadline": "2016-04-09T01:49:49.410Z",
```
lonnen: should this be dynamically set? since this deadline is now in the past, I expect future jobs to insta-fail.

scripts/breakpad-taskcluster.sh
luser: It's totally arbitrary. There isn't a workerType for "run random crap", and configuring one for things like this is silly, so I just tend to use that one because it has lots of capacity.

scripts/breakpad-taskcluster.sh
luser: When you edit jobs through the task-creator interface it will update those for you if you paste this in. (Also there's a button to update them.)

scripts/breakpad-taskcluster.sh
luser: Removing it should be fine.

scripts/build-breakpad.sh
luser: I *think* when I originally wrote this they still had some deps in SVN but they're all git now.

luser:
@lonnen Nothing, the reason I stuck the task definition in there verbatim is because I don't have anything configured to actually run this task, so I figured I ought to at least document how to run it.

luser:
Fixed a few of the issues mentioned in that commit. I think obviously it would be better to have this task defined somewhere that it could actually be executed directly, where the timestamps would be filled in by whatever was executing it, but getting a CI job working in Taskcluster given the other constraints is more work than I want to deal with right now (I spent some time on it and gave up). Having the example task here is better than nothing, I think, so that we can produce new Breakpad builds in the future without having to rediscover how I did it.

lonnen:
It was not clear to me that this was an example task. If you rename the file and modify the header comment to reflect that I am +1 on merging it.

luser:
Rename the file how?

lonnen:
Rename the file to indicate that it's an example file. If I came across it now I would be uncertain if it was actively being used by an external system or not.

luser:
Oh, OK. My intention here is that if we need to rebuild Breakpad, this is what we'd use, so I don't really think of the *script* as an example. The task definition (the JSON inline in the here document) is an example, just because it's not wired up to anything but it's silly to reinvent it from first principles.

lonnen:
Thanks Ted!

PR #3282, opened 2016-04-11
Fix stackwalk build against newer Breakpad

luser:
This small change is necessary to build the stackwalker against the latest Breakpad code, which is https://bugzilla.mozilla.org/show_bug.cgi?id=1262135 .


peterbe:
Hard to review this but I can see that that change was what made it possible to use the other breakpad you had from TaskCluster. 

Will landing this change conflict with the current breakpad tarball we have on S3?

peterbe:
Oh. I see [the comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1262135#c7) now. 

So r+

I can upload the new tarball file to S3 too. A little bit dim here on the necessary timings but perhaps we you and I meet on IRC we can hack this out quickly.

luser:
No, it ought to be harmless.

PR #3283, opened 2016-04-11
fixes bug 1263247 - Web API doesn't convert BadArgumentErrors to 400 …

peterbe:
…requests

CC @adngdb It is strange that we didn't notice this earlier. Basically, there's a `class BadArgumentError` in socorro and one in socorrolib. Same name but totally different classes.


peterbe:
Instead https://github.com/mozilla/socorro/pull/3284 solves this "fundamentally". I.e. the same fix everywhere where it occurs.

PR #3284, opened 2016-04-11
fixes bug 1263689 - Use exceptions from socorrolib where appropriate

peterbe:
CC @lonnen 

The first thing I did was that I wiped `socorro/external/__init__.py` because every class there was also the same in https://github.com/mozilla/socorrolib/blob/master/socorrolib/lib/__init__.py

Then I just grepped and ran tests till things worked.


willkg:
I'll work through this either tonight or tomorrow.

peterbe:
As you'll find it's just a big fat, but mostly manual, search-and-replace of replacing `from socorro.external import ` to `from socorrolib.lib import `.
As I went through it, I corrected the import lines a little bit.

lonnen:
lgtm

PR #3285, opened 2016-04-13
Pull the Breakpad tarball from a Taskcluster index

luser:
This immediately gets us a newer Breakpad build, and will also
streamline the process for updating Breakpad in the future, since
new runs of the same Taskcluster task will update the index to
point at the most recent build.

@peterbe: this is the better solution I was talking about this morning.


peterbe:
I don't think I've ever seen a "303 See Other" HTTP code before. But clearly it works. Travis passes, so I'm happy. 
:balloon:

PR #3286, opened 2016-04-14
sanitize 400 error responses

peterbe:


webapp-django/crashstats/base/views.py
```diff
@@ -3,6 +3,16 @@
 from django.shortcuts import render
 
 
+class HttpResponseBadRequest(http.HttpResponseBadRequest):
```
willkg: This is a Socorro-enhanced version, so it should have a different name to make it easier to know in the code when you're using the right one especially where this has security-related consequences.

webapp-django/crashstats/base/views.py
```diff
@@ -3,6 +3,16 @@
...
+    """Override of the Django HttpResponseBadRequest that makes sure
+    to set the content_type to "text/plain" if it hasn't already been
+    set."""
+    def __init__(self, *args, **kwargs):
+        if 'content_type' not in kwargs:
+            kwargs['content_type'] = 'text/plain; charset=UTF-8'
```
willkg: Would it be better to raise an assertion error thus forcing the caller to always set it?

webapp-django/crashstats/crashstats/utils.py
```diff
@@ -54,6 +54,11 @@ def wrapper(request, *args, **kw):
         response = f(request, *args, **kw)
 
         if isinstance(response, http.HttpResponse):
+            if (
+                isinstance(response, http.HttpResponseBadRequest) and
+                not response['Content-Type'].startswith('application/json')
+            ):
+                response['Content-Type'] = 'application/json; charset=UTF-8'
```
willkg: This seems weird. If this is a JSON view, then shouldn't it always return JSON?

At a bare minimum, this should get a comment explaining why you're picking this specific response out for special treatment and what the consequences of not doing it are.

webapp-django/crashstats/base/views.py
peterbe: There are just too many places for that. And all the places where we have just a short `return HttpResponseBadRequest(str(form.errors))` would have to be quite a lot expanded to have to write `...., content_type='text/plain; charset=UTF-8')`

I'm also wondering, should perhaps

webapp-django/crashstats/base/views.py
peterbe: Yeah. I was thinking about that too. If that was your almost immediate reaction too, it's an indicator that you're right.

webapp-django/crashstats/crashstats/utils.py
willkg: Also, what happens here if the content actually isn't JSON? Then the content and content-type don't match.

webapp-django/crashstats/base/views.py
willkg: The assertion would improve our confidence that we didn't miss something and that all the places this is used have the appropriate content-type which seems to vary depending on the situation in ways that don't seem easy to skim (at least not for me).

You're right that it probably results in more typing and it'll take longer to do.

webapp-django/crashstats/base/views.py
peterbe: addressed now.

webapp-django/crashstats/crashstats/utils.py
peterbe: Note taken about the necessary comment. 
Also, I think you're right. We should ALWAYS set the content-type to `application/json` on views with this decorator. A common use-case is this:

```python
@json_view
def myview(request):
    if some_reason:
        return redirect(reverse('url'))
```
I think the `Location` header trumps the `Content-Type` header when the browser gets one of these. E.g.
```
HTTP 302 Moved
Content-Type: application/json; charset=UTF-8
Location: http://localhost:8000/other/url
```

peterbe:
Solution is over zealous. A simpler patch is coming.

PR #3287, opened 2016-04-14
sanitize bad requests on the web api

peterbe:


peterbe:
So, with this change, if anything goes wrong with the input, on `/api/*`, it will ALWAYS return content type of `application/json; charset=UTF-8`. The only exception was if you're querying for binary data. E.g. `https://crash-stats.mozilla.com/api/RawCrash/?crash_id=b17c1891-ac12-4306-b45b-98e852160407&format=raw`. Suppose there's a Bad Request that happens on a request like this, then you'll get `Content-Type: text/plain` and `HTTP 400 Bad Request`.

peterbe:
Oh. And the other major change is that, if we're going to be 99% JSON in the content type, on all things `/api/*`, then we should also return valid JSON. That's why I changed `return http.HttpResponseBadRequest(exception, ...)` to `return http.HttpResponseBadRequest(json.dumps({'error': exception}), ...)`.

webapp-django/crashstats/api/views.py
```diff
@@ -303,13 +307,16 @@ def model_wrapper(request, model_name):
...
-                    'Not a valid JSON response',
-                    status=400
+                return http.HttpResponseBadRequest(
+                    json.dumps({'error': 'Not a valid JSON response'}),
+                    content_type='application/json; charset=UTF-8'
                 )
             raise
-        except BAD_REQUEST_EXCEPTIONS as e:
-            return http.HttpResponseBadRequest(e)
+        except BAD_REQUEST_EXCEPTIONS as exception:
```
willkg: Much nicer!

willkg:
Travis is happy. This looks good to me, so r+. Merging so it goes to -stage.

PR #3288, opened 2016-04-15
fixes bug 1264940 - validate date range on crashes per day

peterbe:
This just prevents people from using the date parameter wrong and triggering a internal server error.

### DIFF
--- a/webapp-django/crashstats/crashstats/forms.py
+++ b/webapp-django/crashstats/crashstats/forms.py
@@ -241,7 +241,7 @@ class DailyFormBase(BaseForm):
             # Check the invariant. Make sure the start < end.
             if cleaned_data['date_start'] > cleaned_data['date_end']:
                 raise forms.ValidationError(
-                    'date start > date end'
+                    'Start date greater than end date'
                 )
         return cleaned_data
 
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/crashstats/forms.py
```diff
@@ -235,6 +235,16 @@ def clean_v(self):
             )
         return versions
 
+    def clean(self):
+        cleaned_data = super(DailyFormBase, self).clean()
+        if cleaned_data.get('date_end') and cleaned_data.get('date_start'):
+            # Check the invariant. Make sure the start < end.
+            if cleaned_data['date_start'] > cleaned_data['date_end']:
+                raise forms.ValidationError(
+                    'date start > date end'
```
adngdb: That's a pretty obscure error. Why not writing something a bit more explicit, like ``start date has to be lower than end date``?

adngdb:
No tests for this?

webapp-django/crashstats/crashstats/forms.py
peterbe: I'll rewrite it to say "Start date greater than end date". I think that's more useful to whoever has to see this.

peterbe:
> No tests for this? 

That's why reviews are so useful! :)

adngdb:
r+

PR #3289, opened 2016-04-15
fixes bug 1264960 - Super searching for future dates

peterbe:
This is quite an important fix. First of all, if you do a `/api/SuperSearch/` query with a future date, instead of getting an error from ElasticSearch about the index not existing, you now get this:
screen shot 2016-04-15 at 4 29 26 pm

Secondly, when we talk to ES from the webapp, we used to do that via the middleware using `requests.get`. We don't any more. Instead we use the `get_implementation()` and a pure python solution. That means that any errors happening in the implementation will be pure raised exception in the django views where this is used. Note the lack of `try:except:` [here](https://github.com/peterbe/socorro/blob/8e513792f3474b55fcfb16908283f1c5dffc07ba/webapp-django/crashstats/crashstats/models.py#L350).
That means that all the places where we do things like `models.SuperSearch().get(**params)` if anything went wrong, it would ultimately become a internal server error. Now, with my change, some errors (namely `BadArgumentError`) will at least be re-wrapped and returned as a 400 Bad Request.

### DIFF
--- a/socorro/middleware/search_common.py
+++ b/socorro/middleware/search_common.py
@@ -272,10 +272,16 @@ class SearchBase(object):
             for param in parameters['date']:
                 if not param.operator:
                     # dates can't be a specific date
-                    raise BadArgumentError('date must have a prefix operator')
+                    raise BadArgumentError(
+                        'date',
+                        msg='date must have a prefix operator'
+                    )
                 today = datetimeutil.utc_now()
                 if param.value > today:
-                    raise BadArgumentError('future date')
+                    raise BadArgumentError(
+                        'date',
+                        msg="date can't be in the future"
+                    )
                 if (
                     '<' in param.operator and (
                         not lower_than or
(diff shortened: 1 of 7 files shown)


socorro/middleware/search_common.py
```diff
@@ -273,6 +273,9 @@ def fix_date_parameter(self, parameters):
                 if not param.operator:
                     # dates can't be a specific date
                     raise BadArgumentError('date must have a prefix operator')
+                today = datetimeutil.utc_now()
+                if param.value > today:
+                    raise BadArgumentError('future date')
```
adngdb: This, and the previous ones as well, is a bad use of the ``BadArgumentError`` class: the first argument must be the field name (here, ``date``). By default the error will say ``Bad value for parameter(s) 'field_name'``. If you want to specify a different error message, use ``msg='my error message'``. So something like this: 

``raise BadArgumentError('date', msg='date most not be in the future')``

(Note that you can fix the previous use, on line 275, as well: ``raise BadArgumentError('date', 'date must have a prefix operator')``)

adngdb:
r+ when the exception is fixed

PR #3290, opened 2016-04-18
Fixes bug 1264153 - Show install_time as a human-readable datetime.

adngdb:


webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -57,17 +57,23 @@ def recursive_state_filter(state, root):
 
 
 @library.filter
-def js_date(dt, format='ddd, MMM D, YYYY, h:mma UTCZZ', enable_timeago=True):
```
adngdb: I've hit two birds with one stone: this helper was not used anywhere.

webapp-django/crashstats/signature/jinja2/signature/signature_reports.html
```diff
@@ -26,6 +26,8 @@
                     {% if crash[column] %}
                         {% if column == 'date' %}
                             {{ crash[column] | human_readable_iso_date }}
+                        {% elif column == 'install_time' %}
+                            {{ crash[column] | timestamp_to_date }}
```
peterbe: Is this **ALWAYS** a value that can be turned into a datetime? I mean, is this value something we can trust to always be sane? If it comes from breakpad is could be weird can cause problems with `datetime.datetime.fromtimestamp`

peterbe:
Now it looks like this:
screen shot 2016-04-18 at 10 44 13 am

webapp-django/crashstats/signature/jinja2/signature/signature_reports.html
peterbe: The value is not always there. How did jinja not get an error here?
screen shot 2016-04-18 at 10 47 46 am

If `{{ crash[column] }}` becomes an empty string, I suspect that's because we're lucky about how jinja handles empty values. 
Either way, a safer approach (and more transparent) would be something like this:
```
{% elif column == 'install_time' %}
  {% if crash[column] %}
    {{ crash[column] | timestamp_to_date }}
  {% endif %}
{% else %}
```

webapp-django/crashstats/supersearch/jinja2/supersearch/search_results.html
```diff
@@ -53,6 +53,10 @@
                                 {{ val }}
                             </a>
                             {% endfor %}
+                        {% elif column == 'install_time' %}
+                        <a href="{{ current_url }}&amp;{{ column }}=%3D{{ crash[column] }}" class="term" data-field="{{ column }}" data-content="{{ crash[column] }}" title="Add this term to the search form">
+                            {{ crash[column] | timestamp_to_date }}
```
peterbe: ditto here. It makes me uncomfortable that we might be sending in `None` or an empty string to that function.

webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -57,17 +57,23 @@ def recursive_state_filter(state, root):
...
+    timeago = ' timeago' if enable_timeago else ''
+    return jinja2.Markup(
+        '<time datetime="{}" class="jstime{}" data-format="{}">{}</time>'
+        .format(
+            dt.isoformat(),
+            timeago,
+            format,
+            dt.strftime(format)
+        )
+    )
```
peterbe: As discussed in IRC. We don't want "2 weeks and 3 days ago" in the tabular form. (signature report and supersearch results).

adngdb:
@peterbe problems addressed!

webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -57,17 +57,25 @@ def recursive_state_filter(state, root):
...
-    formatted_datetime = ' '.join([dt_date, dt_time, dt_tz])
-    timeago = 'timeago ' if enable_timeago else ''
-    return jinja2.Markup('<time datetime="%s" class="%sjstime"'
-                         ' data-format="%s">%s</time>'
-                         % (dt.isoformat(), timeago,
-                            format, formatted_datetime))
+    try:
+        timestamp = float(timestamp)
+    except (TypeError, ValueError):
+        return
```
peterbe: awesomely simple!

peterbe:
r+ with one nit. 
On the `report_index.html`  and its view, the `context['raw']['InstallTime']`, gets almost the same treatment as what your jinja helper function does. Meaning, all that code there in the `views.py` can be dropped and since your helper function can cope with None or other invalid junk you can just use `<td>{{ raw.InstallTime  | timestamp_to_date }}</td>` or whatever it is.

PR #3292, opened 2016-04-18
fixes bug 1126776 - Analytics for the public API

peterbe:
CC @adngdb 

To test this manually, I ran this: https://gist.github.com/ccc7d2f3a4329febfd777f4441baa2f0
Then I edited my `.env` like this:
```
GOOGLE_ANALYTICS_API_URL=http://localhost:5000/?timeout=2
GOOGLE_ANALYTICS_DOMAIN=what.ever.org
GOOGLE_ANALYTICS_ID=ANY-THING
```

Things to note;

I'm still not 100% sure what to put for the `cid` value. In `fjord` the `cid` was usually made around a particular Django ORM object. That doesn't really exist in this context. The documentation on `cid` is both vague and clear at the same time. Generally, it's supposed to be like a session ID for the client. That doesn't exist when people do `curl https://crash-stats.m.c/api/SomeModel/` since "curlers" wouldn't re-use a cookie. We could do something like this:
```
cid = request.session.get('ga_cid')
if not cid:
    cid = uuid.uuid4().hex
    request.session['ga_cid'] = cid
params['cid'] = cid
```
But I'm not sure how I feel about modifying the `request` object in a tracking function like this. Also, our goal is to find out how many hits we get per Web API endpoint. Not who's surfing from one to another. 

~~I've tried to fire this against our stage domain, but nothing shows up when I'm logged in to Google Analytics Real-time Overview. Not sure if I'm expected to. The only way to know is to use it I suspect. Or wait a day and see if it appears in the aggregates.~~
Using my local curl and editing my `.env` to simulate stage's I really was able to trigger pageviews on API calls. See:
screen shot 2016-04-18 at 4 15 01 pm 
In other words! Unless google analytics later filter these out; it works!

I have tested this code when running my local django via uwsgi as a socket. Things worked just fine.

### DIFF
--- a/webapp-django/crashstats/api/views.py
+++ b/webapp-django/crashstats/api/views.py
@@ -18,7 +18,7 @@ from waffle.decorators import waffle_switch
 from socorrolib.lib import BadArgumentError, MissingArgumentError
 
 import crashstats
-from crashstats.crashstats.decorators import track_api_pageview_decorator
+from crashstats.crashstats.decorators import track_api_pageview
 from crashstats.crashstats import models
 from crashstats.crashstats import utils
 from crashstats.tokens.models import Token
@@ -205,7 +205,7 @@ def has_permissions(user, permissions):
     rate=utils.ratelimit_rate,
     block=True
 )
-@track_api_pageview_decorator
+@track_api_pageview
 @utils.add_CORS_header  # must be before `utils.json_view`
 @utils.json_view
 def model_wrapper(request, model_name):
--- a/webapp-django/crashstats/api/views.py
+++ b/webapp-django/crashstats/api/views.py
@@ -18,6 +18,7 @@ from waffle.decorators import waffle_switch
 from socorrolib.lib import BadArgumentError, MissingArgumentError
 
 import crashstats
+from crashstats.base.ga import track_api_pageview
 from crashstats.crashstats import models
 from crashstats.crashstats import utils
 from crashstats.tokens.models import Token
@@ -317,6 +318,9 @@ def model_wrapper(request, model_name):
                 json.dumps({'error': str(exception)}),
                 content_type='application/json; charset=UTF-8'
             )
+
+        # Execution was successful, inform Google Analytics
+        track_api_pageview(request)
 
         # Some models allows to return a binary reponse. It does so based on
         # the models `BINARY_RESPONSE` dict in which all keys and values
--- a/webapp-django/crashstats/api/views.py
+++ b/webapp-django/crashstats/api/views.py
@@ -18,7 +18,7 @@ from waffle.decorators import waffle_switch
 from socorrolib.lib import BadArgumentError, MissingArgumentError
 
 import crashstats
-from crashstats.base.ga import track_api_pageview
+from crashstats.crashstats.decorators import track_api_pageview_decorator
 from crashstats.crashstats import models
 from crashstats.crashstats import utils
 from crashstats.tokens.models import Token
@@ -205,6 +205,7 @@ def has_permissions(user, permissions):
     rate=utils.ratelimit_rate,
     block=True
 )
+@track_api_pageview_decorator
 @utils.add_CORS_header  # must be before `utils.json_view`
 @utils.json_view
 def model_wrapper(request, model_name):
@@ -318,9 +319,6 @@ def model_wrapper(request, model_name):
                 json.dumps({'error': str(exception)}),
                 content_type='application/json; charset=UTF-8'
             )
-
-        # Execution was successful, inform Google Analytics
-        track_api_pageview(request)
 
         # Some models allows to return a binary reponse. It does so based on
         # the models `BINARY_RESPONSE` dict in which all keys and values
--- a/webapp-django/crashstats/base/ga.py
+++ b/webapp-django/crashstats/base/ga.py
@@ -91,15 +91,9 @@ def track_pageview(
         params['uid'] = str(request.user.id)
 
     params['dp'] = request.path  # Page
-    if request.META.get('QUERY_STRING'):
-        params['dp'] += '?{}'.format(request.META.get('QUERY_STRING'))
+    params['dl'] = request.build_absolute_uri()
 
     params['dt'] = page_title
-
-    if request.META.get('REMOTE_ADDR'):
-        ip = request.META['REMOTE_ADDR']
-        if ip != '127.0.0.1':
-            params['uip'] = ip
 
     if request.META.get('HTTP_USER_AGENT'):
         params['ua'] = request.META['HTTP_USER_AGENT']
--- a/webapp-django/crashstats/base/ga.py
+++ b/webapp-django/crashstats/base/ga.py
@@ -104,7 +104,7 @@ def track_pageview(
     )
 
     def success_cb():
-        # Note! This will trigger as long as there's not python exception
+        # Note! This will trigger as long as there's no python exception
         # happening inside the send.
         # Meaning, if the requests.post(...).status_code != 200, this
         # callback is still called.
@@ -114,7 +114,13 @@ def track_pageview(
         )
 
     def failure_cb(exception):
-        logger.error('Failed to send GA page tracking (%s)', exception)
+        # This can happen if it fails to make the connection to the
+        # remote. E.g. ssl.google-analytics.com
+        # If the HTTP connection is made but, for some reason, GA
+        # refuses the call or they timeout or any other 5xx error,
+        # then this will NOT be a failure callback. In fact, the
+        # success callback will be executed.
+        logger.exception('Failed to send GA page tracking')
 
     transporter.async_send(
         params,
--- a/webapp-django/crashstats/base/ga.py
+++ b/webapp-django/crashstats/base/ga.py
@@ -115,7 +115,7 @@ def track_pageview(
         # Meaning, if the requests.post(...).status_code != 200, this
         # callback is still called.
         logger.info(
-            'Successfully attempted to sent pageview to Google Analytics (%s)',
+            'Successfully attempted to send pageview to Google Analytics (%s)',
             params,
         )
 
--- a/webapp-django/crashstats/base/ga.py
+++ b/webapp-django/crashstats/base/ga.py
@@ -1,0 +1,135 @@
+"""
+Sending Google Analytics events from the webapp.
+
+See https://developers.google.com/analytics/devguides/collection/protocol/\
+v1/devguide#page
+"""
+
+import logging
+import urlparse
+import uuid
+
+from raven.transport.threaded_requests import ThreadedRequestsHTTPTransport
+
+from django.contrib.sites.requests import RequestSite
+from django.conf import settings
+
+logger = logging.getLogger('crashstats:ga')
+
+# Uncomment for local debugging. This will print all logging on stdout.
+# logger.setLevel(logging.DEBUG)
+# import sys
+# logger.addHandler(logging.StreamHandler(sys.stdout))
+
+
+def track_api_pageview(
+    request,
+    page_title=None,
+    data_source='api',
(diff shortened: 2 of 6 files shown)


adngdb:
:dancers:

webapp-django/crashstats/settings/test.py
```diff
@@ -72,3 +72,9 @@
 
 MIDDLEWARE_RETRIES = 10
 MIDDLEWARE_MIDDLEWARE_RETRY_SLEEPTIME = 3
+
+# During testing, if mocking isn't done right, we never want to
+# accidentally send data to Google Analytics.
+GOOGLE_ANALYTICS_API_URL = 'https://fake.guugle-analytics.fake/collect'
```
willkg: It's always better to use "example.com" for fake domain names. That's what it's for.

webapp-django/crashstats/api/views.py
```diff
@@ -318,6 +319,9 @@ def model_wrapper(request, model_name):
                 content_type='application/json; charset=UTF-8'
             )
 
+        # Execution was successful, inform Google Analytics
+        track_api_pageview(request)
```
willkg: Why do this here as opposed to as a decorator on the view function?

webapp-django/crashstats/base/ga.py
```diff
@@ -0,0 +1,135 @@
...
+# import sys
+# logger.addHandler(logging.StreamHandler(sys.stdout))
+
+
+def track_api_pageview(
+    request,
+    page_title=None,
+    data_source='api',
+    client_id=None,
+    async=True,
```
willkg: Why implement sync and async versions when you only want to use the async form?

webapp-django/crashstats/settings/test.py
peterbe: fixed.

webapp-django/crashstats/api/views.py
peterbe: My original intent was to not trigger a tracking on failed things. But in afterthought I guess we can still use a decorator and trigger on the response if its status_code is 200.

willkg: Is it interesting to know about failed uses of the api?

webapp-django/crashstats/settings/test.py
willkg: This is a nit of mine because it's bit me several times.

In case anyone doesn't know and is reading along, there are a handful of top-level domains that are reserved for specific purposes like documenting and testing. Using these appropriately helps when reading the relevant code since the reader knows exactly what the thing is. Further, you can't accidentally have the code work with unintended consequences.

* http://www.iana.org/domains/reserved
* http://tools.ietf.org/html/rfc2606

Fun thing I just discovered: if you type "example.com" in Firefox, it goes to a nice page that talks about this.

peterbe:
CC @adngdb @lonnen 
One interesting challenge is; do we want to track usage WITHIN the webapp too? (I think not) For example, the Home page uses the API in client-side but ultimately the user only loads `/home/product/Firefox` so it would be weird to say that also loading `/api/SuperSearch` is part of that pageview. 

I can accomplish this distinction by looking at the HTTP_REFERER which things like jQuery sets. 

If we don't do this, for every loading of the home page (with javascript enabled) would yield a count of 4 page views.

webapp-django/crashstats/base/ga.py
peterbe: I thought I couldn't mock the async nature and since I know that tests would always need to be "eager" I thought this could come in handy. 
But I'll remove it now. It's silly to have code checked in until it's needed.

peterbe: removed now.

webapp-django/crashstats/crashstats/decorators.py
```diff
@@ -65,3 +68,21 @@ def inner(request, *args, **kwargs):
             raise
         return view(request, *args, **kwargs)
     return inner
+
+
+def track_api_pageview_decorator(view):
```
adngdb: Isn't it weird to put ``decorator`` in the name of a decorator, in a ``decorators.py`` file?

peterbe: It is. I guess I could do `from crashstats.base.ga import track_api_pageview as _track_api_pageview` or something and call the decorator `track_api_pageview` instead. What do you think?

webapp-django/crashstats/base/ga.py
```diff
@@ -0,0 +1,130 @@
...
+        timeout=settings.GOOGLE_ANALYTICS_API_TIMEOUT
+    )
+
+    def success_cb():
+        # Note! This will trigger as long as there's not python exception
+        # happening inside the send.
+        # Meaning, if the requests.post(...).status_code != 200, this
+        # callback is still called.
+        logger.info(
+            'Successfully attempted to sent pageview to Google Analytics (%s)',
```
willkg: Nit: sent -> send.

webapp-django/crashstats/crashstats/decorators.py
willkg: You could make `track_api_pageview` work as both a function and a decorator but that's probably more exotic than it's worth.

I think it's good to have two things that do different things have two different names. Beyond that, I don't think this matters much in the grand scheme of things.

webapp-django/crashstats/base/ga.py
```diff
@@ -0,0 +1,130 @@
...
+    params['dp'] = request.path  # Page
+    if request.META.get('QUERY_STRING'):
+        params['dp'] += '?{}'.format(request.META.get('QUERY_STRING'))
+
+    params['dt'] = page_title
+
+    if request.META.get('REMOTE_ADDR'):
+        ip = request.META['REMOTE_ADDR']
+        if ip != '127.0.0.1':
+            params['uip'] = ip
```
willkg: One thing this does is it subverts anyone's preferences for being tracked. If I don't want to be tracked, then I can prevent GA from loading in my browser and doing things. However, since we're doing this server-side, even if I do that, I'm still getting tracked.

Thus I'm not wildly excited about using peoples' ip addresses here.

With Input, I "faked" the id by using the request object's id. In that way, it's not tied to someone in a way that's identifying.

Do you have something similar you can use? A Django session id, perhaps?

willkg: Er, scratch that--Input doesn't provide a uid at all. It's optional. I suggest you not provide one either.

webapp-django/crashstats/base/ga.py
```diff
@@ -0,0 +1,130 @@
...
+    #   The value of this field should be a random UUID (version 4) as
+    #   described in http://www.ietf.org/rfc/rfc4122.txt
+    params['cid'] = client_id or uuid.uuid4().hex
+
+    if not request.user.is_anonymous():
+        params['uid'] = str(request.user.id)
+
+    params['dp'] = request.path  # Page
+    if request.META.get('QUERY_STRING'):
+        params['dp'] += '?{}'.format(request.META.get('QUERY_STRING'))
```
willkg: I think this is wrong. I think dp is just the path and shouldn't include the querystring portion. I think the querystring is garnered from dl by GA.

https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#dp

webapp-django/crashstats/base/ga.py
```diff
@@ -0,0 +1,130 @@
...
+        return
+
+    domain = settings.GOOGLE_ANALYTICS_DOMAIN
+    if not domain or domain == 'auto':
+        domain = RequestSite(request).domain
+
+    params = {}
+    params['v'] = 1  # version
+    params['tid'] = settings.GOOGLE_ANALYTICS_ID  # Tracking ID / Property ID
+    params['dh'] = domain
```
willkg: We might as well send dl since we're sending all the parts.

https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#dl

willkg:
The changes look fine, so I spent some time working through what we're sending to GA in regards to PII and making sure we're tracking the information we intend to. I made comments accordingly.

I'm still a little puzzled regarding dl and dp. Seems like we shouldn't be sending query params with dp and instead send them with dl. If that's true, I'm not sure why it seems to work as is.

webapp-django/crashstats/crashstats/decorators.py
peterbe: I think I'll go for `from crashstats.base.ga import track_api_pageview as _track_api_pageview` or something such.

webapp-django/crashstats/base/ga.py
peterbe: Removed.

peterbe:
@willkg Addressed. I removed the `uip` and added `dl` accordingly.

webapp-django/crashstats/base/ga.py
```diff
@@ -0,0 +1,124 @@
...
+    if not request.user.is_anonymous():
+        params['uid'] = str(request.user.id)
+
+    params['dp'] = request.path  # Page
+    params['dl'] = request.build_absolute_uri()
+
+    params['dt'] = page_title
+
+    if request.META.get('HTTP_USER_AGENT'):
+        params['ua'] = request.META['HTTP_USER_AGENT']
```
willkg: So, I know with Input, I would run PII-relevant things by @lonnen to get his :+1: because he was the all-powerful all-seeing data privacy sauron in the sky.

I don't know how that works with socorro--when we do PII-relevant things, does that require sign-off from someone who is one with the elemental nature of PII? If so, the above data we'd like to send to GA probably falls into that category of things.

peterbe:
Oops. Deleted the wrong branch,.

webapp-django/crashstats/base/ga.py
willkg: I talked with @lonnen and he said this is "probably fine". He can either chime in or I'll take the responsibility of assuming that means "completely fine" and we can move forward.

webapp-django/crashstats/base/ga.py
```diff
@@ -0,0 +1,124 @@
...
+        # happening inside the send.
+        # Meaning, if the requests.post(...).status_code != 200, this
+        # callback is still called.
+        logger.info(
+            'Successfully attempted to send pageview to Google Analytics (%s)',
+            params,
+        )
+
+    def failure_cb(exception):
+        logger.error('Failed to send GA page tracking (%s)', exception)
```
willkg: Is this actually an exception instance? Do you know offhand if the rest of `sys_excinfo` is available at this point or has it been corrupted by whatever else happened between the exception being thrown and this callback being called?

I ask because if everything is intact, you could do `logger.exception('Failed to send GA page tracking.')` and get better data including a traceback.

willkg:
I had a question about the exception argument of `failure_cb`. I think everything else is covered. The `failure_cb` thing probably isn't a big deal, so if you want to land this today, it's fine as is.

What do you want to do?

willkg:
To clarify, r+ with comments about the `failure_cb`.

peterbe:
I need to play around with that `failure_cb` and see if `logger.exception` would work (because that assumes it gets something from `sys.exc_info()` right?)

willkg:
During lunch, I thought about this some and I think we should just leave it as is.  Yes, we could look into this further to see if we could get traceback information, but honestly, I doubt we're ever going to look at the logs in this case and I'm pretty sure there's nothing that's going to trigger something that pokes us into doing so. So, it seems pretty moot and probably not worth spending time on right now. We can always go figure it out later.

Probably worth tossing in a FIXME comment, though.

peterbe:
I checked. If you do something that triggers a `ConnectionError` for example (which I did by setting `GOOGLE_ANALYTICS_API_URL` to something that isn't running) it does call the `failure_cb` within the context of an exception. 

One final r?

PR #3293, opened 2016-04-19
Bug 1121605 - Added icons to the admin home page.

adngdb:
cc @lonnen @KaiRo-at 

Today I felt like making that admin home page a bit prettier, so here's a proposal: 

If we decide to merge this in, I'll email the folks at https://icons8.com/ to get the approval on us freely using their icon set. 

Note that this won't fix the associated bug, because we still need to figure out how to deal with navigation. But that's a bit more complex.


lonnen:
@adngdb conceptually +1

As it was written, that bug could stay open indefinitely. I'm not sure what you mean by 'deal with navigation' , unless you mean the list of menu items  above the panel and to the right of the page title ("Admin Home"). That's a bit larger than the admin page, since its a pattern re-used over the whole site.

peterbe:
Let's not merge until we get an approval from icons8.com about using it.

peterbe:
I like, nay, I LOVE the separation of groups like that. Data, Users, etc. 
What I don't like is the use of PNG icons. They're taking up space, fixed size and always black. 

Can we not use [Font Awesome](https://fortawesome.github.io/Font-Awesome/icons/) instead?

peterbe:
My other objection is one of nitpickery which you're free to ignore. I think it's bad to mix text transform. In the nav (and in the header of each page) it's titelized. e.g. Graphics Devices. But under the icons you uppercase it. Then visually it looks like two different words.

adngdb:
New proposal: 



This is now using FontAwesome instead of PNG images from font8. 

@peterbe I don't like the look of that page if I remove the ``text-transform: uppercase;``. It makes it very hard to read. I'm in favor of keeping it. Plus we'll have to change the admin navigation at some point anyway. :)

peterbe:
Using `#467086`  instead it looks much nicer. 
screen shot 2016-04-21 at 10 08 39 am

webapp-django/crashstats/manage/jinja2/manage/home.html
```diff
@@ -4,54 +4,86 @@
...
+{% block site_css %}
+{{ super() }}
+{% stylesheet 'manage:home' %}
+{% endblock %}
+
 {% block mainbody %}
 
 <div class="panel">
     <div class="body">
+        <h1>Data</h1>
```
peterbe: Not `h1` please. It looks so wrong when the page header is "Admin Home" in a smaller font and suddenly there's these monster headings.

peterbe:
Perhaps also `#467086` for those uppercased fonts too. 

r+ with or without my bikeshed on the color. But I really think we need to correct the heading tags.

PR #3296, opened 2016-04-20
Fixes bug 1257382 - Added a regex operator for strings in Super Search.

adngdb:
Example test case: [http://localhost:8000/search/?product=Firefox&signature=%40%22OOM+%7C+%22.*%22+%7C+%22.*](http://localhost:8000/search/?product=Firefox&signature=%40%22OOM+%7C+%22.*%22+%7C+%22.*)


socorro/middleware/search_common.py
```diff
@@ -80,7 +81,7 @@ def __init__(self, name, default=None, data_type='enum', mandatory=False):
 
 
 class SearchBase(object):
-    meta_filters = [
+    meta_filters = (
```
adngdb: This and related changes are a fix for an issue I discovered while debugging this, where fields were added to ``self.meta_filters`` every time a call to SuperSearch was made, making it increasingly big for nothing.

peterbe: Ha! The old mutation danger. Yeah, tuples are awesome!

peterbe:
Sorry for being an ass but this is such a cool feature that it just HAS to not be forgotten in the documentation.

peterbe:
As noted in IRC, I followed your link and it worked but the operator drop-down for `signature` became `has terms`, not `matches regex`.

peterbe:
Here I would expect to find only signatures that have a `@` in it. 
screen shot 2016-04-21 at 8 43 15 am

Can I search using `^` and `$` too? E.g. to find signatures that only *start* with `@`

peterbe:
r+ if you provide some guidance on how to use this in the documentation. Perhaps a link to the [ES documentation](https://www.elastic.co/guide/en/elasticsearch/reference/1.4/query-dsl-regexp-query.html#regexp-syntax) too.

Also, you still have a bug there to do with the operator drop-down not becoming `matches regex`.

adngdb:
Note to self:

 <KaiRo> my specific case is that in perl-style regex I'd like to do /^(@0x0 \| )?[^@\|]+@0x/

adngdb:
I filed this bug for the problem with operators in values: https://bugzilla.mozilla.org/show_bug.cgi?id=1267707

@peterbe mind quickly reviewing my latest changes?

webapp-django/crashstats/documentation/jinja2/documentation/supersearch/examples.html
```diff
@@ -67,6 +67,19 @@ <h2 id="installations">Installations</h2>
...
     <p>It is possible to use special aggregations as parameters of an aggregation, like here we use a "cardinality" aggregation inside a nested aggregation. This query will perform an aggregation on products, and for each product it will aggregate on versions, and for each version it will count the distinct number of install times. </p>
 
+    <h2 id="regular-expression">Regular expression</h2>
+
+    <h3>Goal</h3>
+    <p>Find signatures that start with <code>OOM | </code> and have at least another pipe (<code>|</code>). </p>
+
+    <h3>Solution</h3>
+    <div class="example">
+        <pre><code data-url="{{ full_url(request, 'api:model_wrapper', model_name='SuperSearch') }}?signature=@&quot;OOM | &quot;.*&quot; | &quot;.*&amp;_facets=signature&amp;_results_number=0"><span class="url-domain">{{ full_url(request, 'api:model_wrapper', model_name='SuperSearch') }}</span><span class="query-string">?signature=@\"OOM | \".*\" | \".*&amp;_facets=signature&amp;_results_number=0</span></code></pre>
```
peterbe: Hmm... Those spaces should probably be replaced with `%20`. Either way, we really need a jinja helper function that does this computer-work for us humans.

peterbe:
r+ on the latest changes.

webapp-django/crashstats/base/templatetags/jinja_helpers.py
```diff
@@ -76,7 +76,7 @@ def change_query_string(context, **kwargs):
         base = ''
     else:
         base = context['request'].META['PATH_INFO']
-    qs = cgi.parse_qs(context['request'].META['QUERY_STRING'])
+    qs = urlparse.parse_qs(context['request'].META['QUERY_STRING'])
```
adngdb: This was deprecated, see https://docs.python.org/2/library/cgi.html#cgi.parse_qs

adngdb:
@peterbe aaand here's a new commit for you to review! :)

peterbe:
r+ again

PR #3297, opened 2016-04-20
fixes bug 1264153 - show install time in report index

peterbe:
The diff is big. Sorry. But the only real change is how it gathers `install_time` in the view before rendering the template. 

Whilst I was in there I tidied up the tests.

### DIFF
--- a/webapp-django/crashstats/crashstats/views.py
+++ b/webapp-django/crashstats/crashstats/views.py
@@ -1425,6 +1425,18 @@ def report_index(request, crash_id, default_context=None):
         ]
     context['raw_keys'].sort(key=unicode.lower)
 
+    context['install_time'] = None
+    # add install_time if the value is valid
+    if context['raw'].get('InstallTime'):
+        try:
+            datetime.datetime.fromtimestamp(
+                float(context['raw']['InstallTime'])
+            )
+            context['install_time'] = context['raw']['InstallTime']
+        except ValueError:
+            # not a validtimestamp
+            pass
+
     if request.user.has_perm('crashstats.view_rawdump'):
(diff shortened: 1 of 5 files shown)


webapp-django/crashstats/crashstats/views.py
```diff
@@ -1425,6 +1425,18 @@ def handle_middleware_404(crash_id, error_code):
         ]
     context['raw_keys'].sort(key=unicode.lower)
 
+    context['install_time'] = None
```
adngdb: Actually, this should just be ``context['install_time'] = context['raw'].get('InstallTime')``. The conversion to a date is done in the template. 

I feel terrible for merging something that obviously broken in master. Thanks for catching it!

peterbe: So, the old behavior was different. It used to be that if the InstallTime is something that is clearly NOT a valid timestamp, the whole `<th>Install Time</th>...` row was removed/hidden. That was something we wrote code for because it happened. I.e. `"InstallTime": "junk",` in the raw crash. 

Now that I think about it, perhaps we just display this, in the rendered HTML, if the `InstallTime` is corrupted:
```html
<tr>
  <th>InstallTime</th>
  <td></td>
</tr>
```
What do you think?

adngdb: Yeah, that's what I was thinking we were already doing, actually. I'm all in favor of it!

peterbe:
@adngdb Aforementioned problem addressed. Now you get 
```
<tr>
  <th>Install Time</th>
  <td></td>
</tr>
```
...if the `raw.InstallTime` value is garbage. And no messing around in the view function.

PR #3298, opened 2016-04-21
Bug 1265015 - Added a new crashstore to put subset of the json_dump in ES.

adngdb:
@peterbe this should take priority over everything else, we'd like to have this turned on in prod by Monday. Let's work together to have it in stage asap so we can test it on real data.

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -44,9 +44,6 @@ class ESCrashStorage(CrashStorageBase):
 
     #--------------------------------------------------------------------------
     def __init__(self, config, quit_check_callback=None):
-        """Init, you know.
-        """
-
         super(ESCrashStorage, self).__init__(
             config,
             quit_check_callback
@@ -180,17 +177,15 @@ class ESCrashStorageRedactedSave(ESCrashStorage):
 
     #--------------------------------------------------------------------------
     def __init__(self, config, quit_check_callback=None):
-        """Init, you know.
-        """
         super(ESCrashStorageRedactedSave, self).__init__(
             config,
             quit_check_callback
         )
         self.redactor = config.es_redactor.redactor_class(config.es_redactor)
         self.config.logger.warning(
-            "beware, this crashstorage class is destructive to the "
+            "Beware, this crashstorage class is destructive to the "
             "processed crash - if you're using a polycrashstore you may "
-            "find the modified processed crash saved to the other crashstores"
+            "find the modified processed crash saved to the other crashstores."
         )
 
     #--------------------------------------------------------------------------
@@ -199,7 +194,6 @@ class ESCrashStorageRedactedSave(ESCrashStorage):
         """This is the only write mechanism that is actually employed in normal
         usage.
         """
-        self.reconstitute_datetimes(processed_crash)
         self.redactor.redact(processed_crash)
 
         super(ESCrashStorageRedactedSave, self).save_raw_and_processed(
@@ -211,13 +205,13 @@ class ESCrashStorageRedactedSave(ESCrashStorage):
 
 
 #==============================================================================
-class ESCrashStorageRestrictedDump(ESCrashStorage):
+class ESCrashStorageRedactedJsonDump(ESCrashStorageRedactedSave):
     """This class stores redacted crash reports into Elasticsearch, but instead
     of removing the entire `json_dump`, it keeps only a subset of its keys.
     """
     required_config = Namespace()
     required_config.add_option(
-        name="json_dump_keys",
+        name="json_dump_whitelist_keys",
         doc="keys of the json_dump field to keep in the processed crash",
         default=[
             "largest_free_vm_block",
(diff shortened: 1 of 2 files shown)


socorro/external/es/crashstorage.py
```diff
@@ -211,6 +211,77 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
...
+                               crash_id):
+        """This is the only write mechanism that is actually employed in normal
+        usage.
+        """
+        self.reconstitute_datetimes(processed_crash)
+        self.redactor.redact(processed_crash)
+
+        # Now replace the `json_dump` with a subset.
+        restricted_json_dump = {
+            k: processed_crash['json_dump'].get(k)
```
peterbe: Does every crash always have a `json_dump`? E.g. fennec or some other non-Firefox crashes.

socorro/external/es/crashstorage.py
```diff
@@ -211,6 +211,77 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
...
+        default='socorro.external.crashstorage_base.Redactor',
+        from_string_converter=class_converter,
+    )
+    required_config.es_redactor.forbidden_keys = change_default(
+        Redactor,
+        "forbidden_keys",
+        "upload_file_minidump_flash1.json_dump, "
+        "upload_file_minidump_flash2.json_dump, "
+        "upload_file_minidump_browser.json_dump"
+    )
```
peterbe: There's something "unnatural" about this. "Redacting" and "forbidden keys" is something that fits with the definition of the word "redact" ("censor or obscure (part of a text) for legal or security purposes.")

It just feels like we're mixing things a bit too wildly. I thought we used this redacting business for the sake of security. Not for the sake of dieting. Is that a concern you share?

socorro/external/es/crashstorage.py
```diff
@@ -211,6 +211,77 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
...
+        super(ESCrashStorageRestrictedDump, self).__init__(
+            config,
+            quit_check_callback
+        )
+        self.redactor = config.es_redactor.redactor_class(config.es_redactor)
+        self.config.logger.warning(
+            "beware, this crashstorage class is destructive to the "
+            "processed crash - if you're using a polycrashstore you may "
+            "find the modified processed crash saved to the other crashstores"
+        )
```
peterbe: Nice!
(but I suggest you write it as a full sentence (capitalized and full stop) because it's so long and it'll appear a bit weird in the log files.

socorro/external/es/crashstorage.py
```diff
@@ -211,6 +211,77 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
 
 
 #==============================================================================
+class ESCrashStorageRestrictedDump(ESCrashStorage):
+    """This class stores redacted crash reports into Elasticsearch, but instead
+    of removing the entire `json_dump`, it keeps only a subset of its keys.
+    """
+    required_config = Namespace()
+    required_config.add_option(
+        name="json_dump_keys",
```
peterbe: Maybe call it `json_dump_whitelist_keys` instead?

socorro/external/es/crashstorage.py
```diff
@@ -211,6 +211,77 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
 
 
 #==============================================================================
+class ESCrashStorageRestrictedDump(ESCrashStorage):
```
peterbe: s/Restricted/Redacted

To be "restricted" means when your covering or show less. But what it does is that it doesn't just show you a restricted crash, it modifies it. 

Also, "Dump"? To me, in socorro, "dump" refers to the binary blobs and not around the JSON crashes.

socorro/external/es/crashstorage.py
```diff
@@ -211,6 +211,77 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
...
+    """
+    required_config = Namespace()
+    required_config.add_option(
+        name="json_dump_keys",
+        doc="keys of the json_dump field to keep in the processed crash",
+        default=[
+            "largest_free_vm_block",
+            "tiny_block_size",
+            "write_combine_size",
+        ],
```
peterbe: One thing I've never understood about configman, and how we use configman in socorro, is when do you write the list as a string and when you actually write a python list. For example, the `forbidden_keys` thing is a string that gets converted to a python list by configman.

socorro/external/es/crashstorage.py
```diff
@@ -211,6 +211,77 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
...
+        Redactor,
+        "forbidden_keys",
+        "upload_file_minidump_flash1.json_dump, "
+        "upload_file_minidump_flash2.json_dump, "
+        "upload_file_minidump_browser.json_dump"
+    )
+
+    #--------------------------------------------------------------------------
+    def __init__(self, config, quit_check_callback=None):
+        """Init, you know.
```
peterbe: Unnecessary.

socorro/external/es/crashstorage.py
```diff
@@ -211,6 +211,77 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
...
+            "find the modified processed crash saved to the other crashstores"
+        )
+
+    #--------------------------------------------------------------------------
+    def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
+                               crash_id):
+        """This is the only write mechanism that is actually employed in normal
+        usage.
+        """
+        self.reconstitute_datetimes(processed_crash)
```
peterbe: This is a bit of a surprise. The class docstring mentions nothing about changing datetimes. A comment is needed here. Or can we use the superclass instead?

socorro/external/es/crashstorage.py
```diff
@@ -211,6 +211,77 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
...
+        self.redactor.redact(processed_crash)
```
peterbe: This could do with a comment too.

peterbe:
Ugh! I just don't understand how this works. And it's not your fault @adngdb. 
We override the redactor with extra keys or something. But then we call the redactor AND we modify the processed json too. 
Perhaps there's nothing we can do about that. It's just the way it is. I just don't thing it feels natural.

socorro/external/es/crashstorage.py
adngdb: In this case we redact for the sake of disk space consumption in Elasticsearch, not for the sake of security. It might not have been the first intent of those tools, but that's pretty convenient for this use case.

socorro/external/es/crashstorage.py
adngdb: That's a copy/paste from another class just above this one, but sure, I'll edit it. :)

socorro/external/es/crashstorage.py
adngdb: Changed to ``ESCrashStorageRedactedJsonDump``. But note that the JSON dump is simply a JSON version of the dump.

socorro/external/es/crashstorage.py
adngdb: I don't really get it either, but this works. I didn't dig into the ``Redactor`` class to see how the forbidden_keys thing works.

socorro/external/es/crashstorage.py
adngdb: Actually, this is not needed, it's already applied in a subsequent method.

adngdb:
@peterbe updated. I've changed the class a bit to inherit from the previous redacting class, and changed its default ``forbidden_keys`` to ignore ``json_dump``. This way I think it's a bit clearer how this works.

peterbe:
r+ but I'm still not a 100% confident how stuff works. I can't find anything wrong with the code. Just saying I don't want to take the heat with the prod data melts through the earth core.

PR #3302, opened 2016-04-21
fixes bug 1266529 - Crashes per Day with future date 500 error

peterbe:


webapp-django/crashstats/crashstats/forms.py
```diff
@@ -235,6 +235,16 @@ def clean_v(self):
             )
         return versions
 
+    def clean_date_end(self):
+        value = self.cleaned_data['date_end']
+        if value:
+            now = datetime.datetime.utcnow()
+            if not isinstance(value, datetime.datetime):
```
adngdb: Why not just ``if isinstance(value, datetime.date):``? It's simpler to understand and I think strictly equivalent here.

adngdb: (I'm saying that because I was confused by this. )

webapp-django/crashstats/crashstats/tests/test_forms.py
```diff
@@ -271,6 +271,22 @@ def get_new_form(cls, data):
...
+             'date_end': today + datetime.timedelta(days=1)},
+        )
+        ok_(not form.is_valid())
+        form = get_new_form(
+            forms.DailyFormByVersion,
+            {'p': 'WaterWolf',
+             'date_start': today,
+             'date_end': today + datetime.timedelta(days=1)},
+        )
+        ok_(not form.is_valid())
```
adngdb: I think you just wrote the same test twice.

adngdb:
r+ with nits

webapp-django/crashstats/crashstats/forms.py
peterbe: Ugh. This whole thing is a mess. I think I discovered some weird things about django's `form.DateField()` and all was well until I had to hack around the tests. I'll take another stab at it. 

By the way, did you know that...:
```
>>> import datetime
>>> now = datetime.datetime.now()
>>> isinstance(now, datetime.datetime)
True
>>> isinstance(now, datetime.date)
True
```
...but...
```
>>> now = datetime.datetime.now().date()
>>> isinstance(now, datetime.date)
True
>>> isinstance(now, datetime.datetime)
False
```

adngdb: Fun, I didn't know, but I guess that makes sense. Good to know though!

peterbe:
@adngdb addressed. Merge when happy.

PR #3304, opened 2016-04-25
fixes bug 1267287 - SuperSearch ValueError on bad _results_number

peterbe:


socorro/external/es/supersearch.py
```diff
@@ -214,6 +214,8 @@ def get(self, **kwargs):
                         results_number = param.value[0]
                         if results_number > 1000:
                             raise BadArgumentError('_results_number too large')
+                        if results_number < 0:
+                            raise BadArgumentError('_results_number negative')
```
adngdb: While you're at it, do you mind fixing the use of ``BadArgumentError`` once again? And I would be more verbose in the error, too.

```python
raise BadArgumentError('_results_number', msg='_results_number cannot be negative')
```

Also, just above on line 216: 

```python
raise BadArgumentError('_results_number', '_results_number cannot be greater than 1,000')
```

adngdb:
r+

peterbe:
Fixed and rebased. Just needs to be merged.

PR #3305, opened 2016-04-25
fixes bug 1267301 - Postgres table for storing correlations

peterbe:
This was tricky to make. But I think I figured it out and I have no, in my local psql, a nice looking table. 

Here's how I created it:

```
$ dropdb 
$ createdb socorro_integration_test
$ PYTHONPATH=. python socorro/external/postgresql/setupdb_app.py --database_name=socorro_integration_test --database_username=peterbe --database_password=secret --database_superusername=peterbe --database_superuserpassword=secret
$ alembic -c config/alembic.ini revision --autogenerate -m 'correlations table'
$ jed alembic/versions/373ef4569b93_correlations_table.py
```

### DIFF
--- a/alembic/versions/373ef4569b93_correlations_table.py
+++ b/alembic/versions/373ef4569b93_correlations_table.py
@@ -1,0 +1,71 @@
+"""correlations table
+
+Revision ID: 373ef4569b93
+Revises: 335c2bfd99a6
+Create Date: 2016-04-25 14:11:28.373859
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+from socorrolib.lib import jsontype
+
+# revision identifiers, used by Alembic.
+revision = '373ef4569b93'
+down_revision = '335c2bfd99a6'
+
+
+def upgrade():
+    op.create_table(
+        'correlations',
+        sa.Column('id', sa.INTEGER(), nullable=False),
+        sa.Column(
+            'product_version_id',
+            sa.INTEGER(),
+            autoincrement=False,
+            nullable=False
+        ),
+        sa.Column('platform', sa.TEXT(), nullable=False),
+        sa.Column('signature_id', sa.INTEGER(), nullable=False),
+        sa.Column('key', sa.TEXT(), nullable=False),
+        sa.Column(
+            'count',
+            sa.INTEGER(),
+            server_default=sa.text(u'0'),
+            nullable=False
+        ),
+        sa.Column('notes', sa.TEXT(), server_default=u'', nullable=False),
+        sa.Column('date', sa.DATE(), nullable=False),
+        sa.Column('payload', jsontype.JsonType(), nullable=True),
+        sa.PrimaryKeyConstraint('id', 'platform')
+    )
+    op.create_index(
+        'correlations_signature_idx',
+        'correlations',
+        [
+            'product_version_id',
+            'platform',
+            'key',
+            'date',
+            'signature_id'
+        ],
+        unique=True
+    )
+    op.create_index(
+        'correlations_signatures_idx',
+        'correlations',
+        [
+            'product_version_id',
+            'platform',
+            'key',
+            'date'
+        ],
+        unique=True
+    )
+
+
+def downgrade():
+    op.drop_index('correlations_signatures_idx', table_name='correlations')
+    op.drop_index('correlations_signature_idx', table_name='correlations')
+    op.drop_table('correlations')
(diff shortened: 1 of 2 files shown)


peterbe:
@adngdb I know I said to hold off a little bit on this but I'm now pretty sure this is good to go. I've tested it locally and it seems to create a table that will work. If I'm wrong we can write another migration later that upgrades it.

socorro/external/postgresql/models.py
```diff
@@ -1425,6 +1425,54 @@ class MissingSymbols(DeclarativeBase):
...
+    #  signature
+    #
+    __table_args__ = (
+        Index(
+            'correlations_signatures_idx',
+            product_version_id,
+            platform,
+            key,
+            date,
+            unique=True
```
adngdb: This one cannot be unique, because for each of those combinations you can have multiple signatures.

alembic/versions/373ef4569b93_correlations_table.py
```diff
@@ -0,0 +1,71 @@
...
+
+from socorrolib.lib import jsontype
+
+# revision identifiers, used by Alembic.
+revision = '373ef4569b93'
+down_revision = '335c2bfd99a6'
+
+
+def upgrade():
+    op.create_table(
```
adngdb: I don't know much about alembic, but isn't it possible to import the model you created and use that to create the table instead of repeating the same information twice?

alembic/versions/373ef4569b93_correlations_table.py
```diff
@@ -0,0 +1,71 @@
...
+    op.create_index(
+        'correlations_signatures_idx',
+        'correlations',
+        [
+            'product_version_id',
+            'platform',
+            'key',
+            'date'
+        ],
+        unique=True
```
adngdb: Same here, this cannot be unique.

alembic/versions/373ef4569b93_correlations_table.py
peterbe: Not according the their documentation. 
This is how to make them. You create the class in models.py and then run the `alembic --autogenerate` thing. Sad but reality. 

I for one wouldn't mind that we migrate ALL postgres things to Django's ORM and entirely ditch SQLAlchemy and alembic. (CC @willkg )

socorro/external/postgresql/models.py
peterbe: Hmm... the code was autogenerated by alembic and then I tidied up the indentation. I guess I missed these uniqueness things. Removed now.

alembic/versions/373ef4569b93_correlations_table.py
willkg: It's important that migrations continue to work over time as the rest of the codebase changes. Given that, the migration explicitly can't depend on the rest of the codebase for information or functionality. The same is true for Django migrations and South migrations.

peterbe:
@adngdb r? again.

alembic/versions/373ef4569b93_correlations_table.py
peterbe: Yeah, when you point it out. It makes sense that the model can't be trusted. Since it changes between when the migration was created and when you run it.

peterbe: I'm not trolling but I am stirring. 
Django can talk SQL to any PG table without necessarily have created its table. 

Adrian's point is that it's mixing the streams into one giant pot. I.e. we're using Django for too much where as things like socorro.external.postgres.* is all about data and the stuff that django does is all about presentation. 
Also, I'm such a Django-nut that I wonder if sometimes I see Django as the hammer that hammers any kind of nail. 

But think about it. You get a really powerful ORM query language. You can generate forms from the models. You can use django migrations. 

I'd love to take this, as an idea, seriously again late in 2016 or some time next year.

adngdb: Will's argument is exactly what I wanted to read. What you have now seems like the right thing to do.

socorro/external/postgresql/models.py
adngdb: You messed your last commit. This ``unique=True`` should be removed...

alembic/versions/373ef4569b93_correlations_table.py
```diff
@@ -0,0 +1,69 @@
...
+    op.create_index(
+        'correlations_signature_idx',
+        'correlations',
+        [
+            'product_version_id',
+            'platform',
+            'key',
+            'date',
+            'signature_id'
+        ],
```
adngdb: ... and the one that was here should still be there. At the moment your migration doesn't match the model, and the model cannot work (because a non-unique index is marked as unique).

peterbe:
@adngdb Ugh! I'm failing at this PR. Hopefully I get it right this time.

PR #3306, opened 2016-04-27
bug 1187852 - make stackwalker expose symbol URLs for modules with loaded symbols

luser:
This will let us link to symbol files from the report page (bug 1184101), which would be useful.

This is mostly just refactoring things to reuse the "build a symbol URL" logic, and then storing the URL the existing SymbolStats struct.

### DIFF
--- a/minidump-stackwalk/http_symbol_supplier.cc
+++ b/minidump-stackwalk/http_symbol_supplier.cc
@@ -123,18 +123,32 @@ HTTPSymbolSupplier::StoreSymbolStats(const CodeModule* module,
   }
 }
 
+string
+HTTPSymbolSupplier::GuessURL(const CodeModule* module)
+{
+  string path, url;
+  if (!server_urls_.empty()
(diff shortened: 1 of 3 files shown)


minidump-stackwalk/http_symbol_supplier.cc
```diff
@@ -123,18 +123,32 @@ HTTPSymbolSupplier::StoreSymbolStats(const CodeModule* module,
...
+  }
+  return "";
+}
+
 void
 HTTPSymbolSupplier::StoreCacheHit(const CodeModule* module)
 {
-  SymbolStats stats = {true, 0.0f};
+  // We don't know the original URL, so guess that it would have
+  // been from the first URL we would have tried.
```
rhelmer: It would be nice if we didn't have to guess here, I suppose it'd need to keep state about where it downloaded files (or have a separate cache area for different source URLs).

I think this is fine for now but something to consider if we end up using multiple symbol URLs much - I don't expect that we will in the near term.

rhelmer:
r+

PR #3308, opened 2016-04-27
fixes bug 1268043 - "More Reports" link on report index links to old report list

peterbe:
We could maybe add to the URL, the `&date=>...&date=<...` like it does on `topcrashers.html` but it feels slightly out of context. What do you think? We'd have to arbitrarily pick today and today minus 7 days.

### DIFF
--- a/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
+++ b/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
@@ -77,7 +77,7 @@
                                 <th scope="row">Signature</th>
                                 <td>
                                     {{ report.signature }}
-                                    <a href="{{ url('signature:signature_report') }}?product={{ report.product }}&amp;signature={{ report.signature|urlencode }}&amp;version={{ report.version|urlencode }}" class="sig-overview" title="View more reports of this type">More Reports</a>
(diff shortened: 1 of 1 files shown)


adngdb:
@peterbe that's not needed, as those values are the default anyway.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -77,7 +77,7 @@
                                 <th scope="row">Signature</th>
                                 <td>
                                     {{ report.signature }}
-                                    <a href="{{ url('crashstats:report_list') }}?product={{ report.product }}&amp;signature={{ report.signature|urlencode }}" class="sig-overview" title="View more reports of this type">More Reports</a>
+                                    <a href="{{ url('signature:signature_report') }}?product={{ report.product }}&amp;signature={{ report.signature|urlencode }}&amp;version={{ report.version|urlencode }}" class="sig-overview" title="View more reports of this type">More Reports</a>
```
adngdb: Why do you add the version if it wasn't there before? I don't think we want to change the way this works, just link to the new Signature report.

peterbe: Fair point. I'll remove the version thing.

peterbe:
Version parameter removed now.

PR #3311, opened 2016-04-28
Fixes bug 1260988 - Added product and version filters to NewSignature…

adngdb:
… service.


webapp-django/crashstats/tools/forms.py
```diff
@@ -0,0 +1,25 @@
...
+    start_date = forms.DateField(required=False)
+    not_after = forms.DateField(required=False)
+    product = MultipleValueField(required=False)
+    version = MultipleValueField(required=False)
+
+    def clean_product(self):
+        """By default, we want the product to be set to something, it should
+        not be empty. """
+        value = self.cleaned_data['product']
+        value = filter(bool, value)  # Remove all empty values.
```
peterbe: Neat but if you need to write a comment perhaps `value = [x for x in value if x]` is better :)

webapp-django/crashstats/tools/forms.py
```diff
@@ -0,0 +1,25 @@
...
+
+    end_date = forms.DateField(required=False)
+    start_date = forms.DateField(required=False)
+    not_after = forms.DateField(required=False)
+    product = MultipleValueField(required=False)
+    version = MultipleValueField(required=False)
+
+    def clean_product(self):
+        """By default, we want the product to be set to something, it should
+        not be empty. """
```
peterbe: This should not be done in the form. Defaults are best dealt with in the view. E.g.
```python
product = form.cleaned_data['product'] or settings.DEFAULT_PRODUCT
```

peterbe:
r+ if you move the default product thing out of the form.

webapp-django/crashstats/tools/forms.py
adngdb: Using ``filter`` is faster, and I'm happy I learned about it, it's quite a convenient built-in!

PR #3314, opened 2016-04-29
Fixes bug 1268029 - Added a Rule to replace a signature if there is a JIT classification.

adngdb:
cc @luser

### DIFF
--- a/socorro/processor/signature_utilities.py
+++ b/socorro/processor/signature_utilities.py
@@ -970,3 +970,31 @@ class SignatureRunWatchDog(SignatureGenerationRule):
             "shutdownhang | %s" % processed_crash['signature']
         )
         return result
+
+
+#==============================================================================
+class SignatureJitCategory(Rule):
+    """replaces the signature if there is a JIT classification in the crash"""
+
+    #--------------------------------------------------------------------------
+    def version(self):
+        return '1.0'
+
+    #--------------------------------------------------------------------------
+    def _predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
+        return (
+            processed_crash.get('classifications') and
+            processed_crash.classifications.get('jit') and
+            processed_crash['classifications']['jit'].get('category')
+        )
+
+    #--------------------------------------------------------------------------
+    def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
+        processor_meta.processor_notes.append(
+            'Signature replaced with a JIT Crash Category, '
+            'was: {}'.format(processed_crash.signature)
+        )
+        processed_crash.signature = "jit | {}".format(
+            processed_crash.classifications['jit']['category']
+        )
+        return True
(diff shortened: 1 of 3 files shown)


socorro/processor/signature_utilities.py
```diff
@@ -970,3 +970,31 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
+    #--------------------------------------------------------------------------
+    def version(self):
+        return '1.0'
+
+    #--------------------------------------------------------------------------
+    def _predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
+        return (
+            processed_crash.get('classifications') and
+            processed_crash.classifications.get('jit') and
+            processed_crash['classifications']['jit'].get('category')
```
peterbe: you could have done `processed_crash.get('classifications', {}).get('jit', {}).get('category')` :)
or 
```python
try:
    return processed_crash.classifications.jit.category
except KeyError:
    return 
```

socorro/processor/signature_utilities.py
```diff
@@ -970,3 +970,31 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
+        )
+
+    #--------------------------------------------------------------------------
+    def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
+        processor_meta.processor_notes.append(
+            'Signature replaced with a JIT Crash Category, '
+            'was: {}'.format(processed_crash.signature)
+        )
+        processed_crash.signature = "jit | {}".format(
+            processed_crash.classifications['jit']['category']
```
peterbe: Is it a DotDict or a `dict`?
I for one, prefer pure python `dict` objects everywhere. I think DotDict is asking for trouble.

socorro/processor/signature_utilities.py
```diff
@@ -970,3 +970,31 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
+            processed_crash.get('classifications') and
+            processed_crash.classifications.get('jit') and
+            processed_crash['classifications']['jit'].get('category')
+        )
+
+    #--------------------------------------------------------------------------
+    def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
+        processor_meta.processor_notes.append(
+            'Signature replaced with a JIT Crash Category, '
+            'was: {}'.format(processed_crash.signature)
```
peterbe: Pardon my lack of experience with processors but wouldn't it be better to use `...was: "{}"'.format(...)`?

peterbe:
Thanks for making the pep8 cleanup and nice that it's a separate commit!
As you can tell I'm against using DotDict. I base that on years of experience of weird DotDict errors that are hard to debug, the fact that you can't easily transpose them as JSON and the years of using MongoDB and trying to mess with a DotDict equivalent. Not a nit. Just a drive-by opinion. 

I do think we you might want to at least look at the nit I mentioned in the `_predicate` method. Otherwise r+

socorro/processor/signature_utilities.py
adngdb: ``processed_crash`` I know for sure is a DotDict, but I think ``classifications`` is not. Hence the syntax. But you're right, and I agree about DotDict being asking for trouble, even though it's a nicer syntax.

PR #3317, opened 2016-05-02
Fixes bug 1268867 - Removed dead link to Plugins bugzilla product.

adngdb:


peterbe:
Should we perhaps keep the link but clicking it should instead set the product to "Core" and the component to "Plug-ins". As per [marco's point](https://bugzilla.mozilla.org/show_bug.cgi?id=1268867#c1)

adngdb:
@peterbe updated to the new product.

PR #3318, opened 2016-05-02
Fixes bug 1269240 - Display pretty platform version in report/index.

adngdb:


webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -149,7 +149,7 @@
                             </tr>
                             <tr>
                                 <th scope="row">OS</th>
-                                <td>{{ report.os_name }}</td>
+                                <td>{{ report.os_pretty_version or report.os_name }}</td>
```
peterbe: Because this becomes a python expression, what happens if `report.os_pretty_version` does not exist? E.g.
```python
>>> class F:
...   pass
...
>>> report = F()
>>> report.os_name = 'Windows NT'
>>> print report.os_pretty_name or report.os_name
AttributeError: F instance has no attribute 'os_pretty_name'
>>>
```

Jinja is pretty smart when it comes to rendering things that are not defined but whichever way we do it's better to be explicit. I know you can do something like `{% if report.os_pretty_version %}` and it becomes `False` if there is no key called `os_pretty_version` on the `report` object/dict. However, I think you get a Jinja error if you try to `{% if report.os_pretty_version %}` if `report` does not exist (i.e. `Undefined`). 

To best test this, make a test where the `os_pretty_version` simply doesn't exist. But `os_name` does. And the result should be that it prints the value of the `os_name`.

adngdb: I have verified that the code works as it is now. I removed the ``os_pretty_version`` field from the report and verified that no error was raised. Writing a test for that would be cumbersome and quite useless because it would just be testing a jinja feature.

adngdb: Also, ``report`` cannot be undefined. And if it ever is, there will be plenty of errors on that page. We can't be over protective with everything, we sometimes have to trust that some value will be there. And since ``os_pretty_version`` is set by the processors, it should always be there. Otherwise we have a different problem to solve.

peterbe: Cool. So you've tested it. I just got nervous after the problems we had with `{{ report.InstallTime | print_date }}`

PR #3319, opened 2016-05-02
fixes bug 1269453 - Improve cachability for the first signature dates

peterbe:
I put in a print statement [here](https://github.com/peterbe/socorro/blob/6e89e6ac4dab66c622e1c229780c5f4312d11a5b/socorro/external/postgresql/signature_first_date.py#L19) that said `print "Querying for", len(params['signatures']), "signatures"`. 
Then I opened the Top Crashers for each version of Firefox. Before this patch, it said:
```
Querying for 50 signatures
Querying for 50 signatures
Querying for 50 signatures
Querying for 50 signatures
```
Then, after writing this patch, same thing and this time it printed this out:
```
Querying for 50 signatures
Querying for 47 signatures
Querying for 29 signatures
Querying for 40 signatures
```
In other words, it stops querying PG for signatures' first date if it's already done that for some other version. 

So this patch makes it less SQL select work on PG but that's the small part. 
What really makes a difference is that now EACH signature has its own first_date cache and it's 24 hours. It means that we'll only really need something like a total of 150ish queries against PG in a day.  Instead of 200x24. 

That's going to make the Top Crashers page faster.

### DIFF
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -1352,6 +1352,15 @@ class SignaturesByBugs(SocorroMiddleware):
 
 class SignatureFirstDate(SocorroMiddleware):
 
+    # Set to a short cache time because, the only real user of this
+    # model is the Top Crasher page and that one uses the highly
+    # optimized method `.get_dates()` which internally uses caching
+    # for each individual signature and does so with a very long
+    # cache time.
+    # Making it non-0 is to prevent the stampeding herd on this endpoint
+    # alone when exposed in the API.
+    cache_seconds = 5 * 60  # 5 minutes only
+
     implementation = (
         socorro.external.postgresql.signature_first_date.SignatureFirstDate
     )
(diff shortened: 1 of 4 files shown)


webapp-django/crashstats/crashstats/models.py
```diff
@@ -1368,6 +1377,51 @@ class SignatureFirstDate(SocorroMiddleware):
...
+                # remember the cache keys of those we need to look up
+                missing[signature] = cache_key
+
+        if missing:
+            hits = super(SignatureFirstDate, self).get(
+                signatures=missing.keys()
+            )['hits']
+
+            for hit in hits:
+                signature = hit.pop('signature')
```
adngdb: What's the point of using ``pop`` here? Is there any performance gain compared to ``get``? I can't see any difference locally, and ``pop`` is more confusing in my opinion.

adngdb: Damn, I should read the rest of the code before commenting! Forget this! :)

adngdb:
r+

PR #3323, opened 2016-05-04
Fixes bug 1269817 - Added a rule for signature if crash contains an I…

adngdb:
…PC channel error.

### DIFF
--- a/socorro/processor/signature_utilities.py
+++ b/socorro/processor/signature_utilities.py
@@ -998,3 +998,35 @@ class SignatureJitCategory(Rule):
             processed_crash['classifications']['jit']['category']
         )
         return True
+
+
+#==============================================================================
+class SignatureIPCChannelError(Rule):
+    """replaces the signature if there is a JIT classification in the crash"""
+
+    #--------------------------------------------------------------------------
+    def version(self):
+        return '1.0'
+
+    #--------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


socorro/processor/signature_utilities.py
```diff
@@ -998,3 +998,35 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
+    """replaces the signature if there is a JIT classification in the crash"""
+
+    #--------------------------------------------------------------------------
+    def version(self):
+        return '1.0'
+
+    #--------------------------------------------------------------------------
+    def _predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
+        try:
+            return bool(raw_crash['ipc_channel_error'])
```
peterbe: Why not just `return raw_crash.get('ipc_channel_error')`?

peterbe:
r+ with a nit question.

socorro/processor/signature_utilities.py
adngdb: Because I like that ``predicate`` returns a true boolean value. And I don't like it exposing values. I don't know. It's just me being a bit too much of a purist, I guess. :)

peterbe: `return bool(raw_crash.get('ipc_channel_error'))` then :)

PR #3326, opened 2016-05-04
Bug 1005224 - Swallow error when the report type does not exist.

adngdb:


peterbe:
r+ if travis is happy.

PR #3327, opened 2016-05-04
fixes bug 1005224 - log failed correlations reports

peterbe:
Cosmetic small improvement.

### DIFF
--- a/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/report.js
+++ b/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/report.js
@@ -14,7 +14,7 @@ var SignatureCorrelations = (function() {
                     if (response.count) {
                         $('li.correlations').show();
                     }
-                    if (response.errors) {
+                    if (response.errors && response.errors.length) {
                         console.warn(
                             'There were some errors fetching correlations:'
                         );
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/crashstats/static/crashstats/js/socorro/report.js
```diff
@@ -14,6 +14,12 @@ var SignatureCorrelations = (function() {
                     if (response.count) {
                         $('li.correlations').show();
                     }
+                    if (response.errors) {
```
adngdb: An empty array evaluates to true in JavaScript (try ``if ([]) console.log('yes');`` in your console), so this test should be different. Maybe ``if (response.errors.length) {`` instead?

adngdb:
I'm not sure there is much value in doing this, but it's still better than doing nothing. :)

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/report.js
peterbe: The server doesn't include the key `errors` if there are no errors. So the code works as is. Arguably, perhaps it should. And then the if statement would need to be smarter.

adngdb: It actually does. You changed that behavior in this PR. :)

peterbe: Good catch! Pushed a fix for this now.

PR #3329, opened 2016-05-05
bug 1091732 - No Response as JSON strings in crashstats/test_views.py

peterbe:
This is purely stylistic changes. No code or functionality change. 
I just replaced all places where it does things like:
```python
def mocked_get(...):
    return Response("""
         {"foo": null}
    """)
```
...with...:
```python
def mocked_get(...):
    return Response({
        'foo': None,
    })
```

Whilst I was in there I also replaced all (I hope) use of `"` in favor of `'`.

### DIFF
--- a/webapp-django/crashstats/api/tests/test_views.py
+++ b/webapp-django/crashstats/api/tests/test_views.py
@@ -90,30 +90,28 @@ class TestViews(BaseTestViews):
     def test_CrashesPerAdu(self, rget):
         def mocked_get(url, params, **options):
             if 'crashes/daily' in url:
-                return Response("""
-                    {
-                      "hits": {
+                return Response({
+                    "hits": {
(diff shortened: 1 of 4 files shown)


willkg:
Grabbing this to look at today.

willkg:
I skimmed it and nothing stuck out at me.

It passes Travis, so we can be confident the following are true:

1. the changes don't create new Python parsing errors or Python -> JSON errors
2. the tests continue to test what they were testing before because we're only making changes to Response data

You mentioned changing all " to '. There's a bunch of " in there, still. You can see it in the top part of the "Files changed" tab. I think it's worth landing this as is and then maybe doing another pass if you feel strongly about it.

r+

willkg:
@peterbe told me to merge this PR with all due haste. Mergificationalyzing now!

PR #3330, opened 2016-05-05
fixes bug 1270498 - ! should be URL encoded in the SuperSearch API URL

peterbe:
Seem's jQuery's `$.params({'foo': ['!']})` fails us. It manages to get other characters right. 

See http://codepen.io/peterbe/pen/reovmx?editors=1010

### DIFF
--- a/webapp-django/crashstats/supersearch/static/supersearch/js/socorro/search.js
+++ b/webapp-django/crashstats/supersearch/static/supersearch/js/socorro/search.js
@@ -130,6 +130,7 @@ $(function () {
     function updatePublicApiUrl(params) {
         // Update the public API URL.
         var queryString = $.param(params, true);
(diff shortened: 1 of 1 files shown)


adngdb:
Fun fact: your branch name is breaking my git commands in zsh. :D

adngdb:
See my comment in the bug. I don't think we should merge this.

peterbe:
> Fun fact: your branch name is breaking my git commands in zsh. :D

mine too! 

But I have [a fix](https://www.peterbe.com/plog/gg-prototype) coming.

peterbe:
I didn't know about using the the single quote to escape it on the command line. Do many people know about that? What's the harm? Isn't it kinda nice to be able to just take the URL from the UI and paste it into the command line?

PR #3332, opened 2016-05-05
Fixes bug 1266827: Add keybuilder class to s3 crash storage [DO NOT MERGE]

willkg:
This breaks out key building into a separate class allowing us to
specify which keybuilder class to use.

It also creates a new DatePrefixKeyBuilder which will pull the date
portion off the end of a raw_crash ooid and add that to the s3
pseudo-filename before the ooid. This new pseudo-filename form lets us
list all the crashes that came in on a specific date.

Further, it does this in a backwards compatible way by changing
build_key() to build_keys() which returns a list of keys to try in order
of "goodness". That way .fetch() can try the keys in order until it
finds the object in question.

r?

### DIFF
--- a/socorro/external/boto/connection_context.py
+++ b/socorro/external/boto/connection_context.py
@@ -14,6 +14,7 @@ import boto.exception
 from configman import Namespace, RequiredConfig, class_converter
 
 from socorrolib.lib.converters import change_default
+from socorrolib.lib.ooid import dateFromOoid
 
 
 class KeyNotFound(Exception):
@@ -21,11 +22,71 @@ class KeyNotFound(Exception):
 
 
 class JSONISOEncoder(json.JSONEncoder):
-
     def default(self, obj):
         if isinstance(obj, datetime.date):
             return obj.isoformat()
         raise NotImplementedError("Don't know about {0!r}".format(obj))
+
+
+class KeyBuilderBase(object):
+    """Base key builder for s3 pseudo-filenames"""
+    def build_keys(self, prefix, name_of_thing, id):
+        """
+        Use S3 pseudo-directories to make it easier to list/expire later
+        {prefix}/v1/{name_of_thing}/{id}
+        """
+        return [
+            '%s/v1/%s/%s' % (prefix, name_of_thing, id)
+        ]
+
+
+class DatePrefixKeyBuilder(KeyBuilderBase):
+    """s3 pseudo-filename key builder with date prefixes
+
+    This adds the date to the prefix for raw_crash objects making it easier to
+    list all the raw_crash objects for a specific date.
+
+    """
+    def build_keys(self, prefix, name_of_thing, id):
+        """Use S3 pseudo-directories to make it easier to list/expire.
+
+        For "raw_crash" things, the id is an ooid/crash_id. This uses the
+        first three characters of the ooid for entropy and extracts the
+        crash submission date from the last 6 charactes. Then it mooshes
+        that all together into this structure::
+
+            {prefix}/v2/{name_of_thing}/{entropy}/{date}/{id}
+
+        This makes it possible to list all the raw_crashes submitted on a
+        given date.
+
+        It also returns the keys that KeyBuilderBase builds so that we can
+        fetch objects that were saved with older keys.
+
+        """
+        keys = []
+
+        if name_of_thing == 'raw_crash':
+            datestamp = dateFromOoid(id)
+
+            if datestamp is not None:
+                # We insert the first 3 chars of the ooid/crash_id providing
+                # some entropy earlier in the key so that consecutive s3
+                # requests get distributed across multiple s3 partitions.
+                first_chars = id[:3]
+                date = datestamp.strftime('%Y%m%d')
+                keys.append(
+                    '%s/v2/%s/%s/%s/%s' % (
+                        prefix, name_of_thing, first_chars, date, id
+                    )
+                )
+
+        keys.extend(
+            super(DatePrefixKeyBuilder, self).build_keys(
+                prefix, name_of_thing, id
+            )
+        )
+        return keys
 
 
 #==============================================================================
@@ -57,6 +118,17 @@ class ConnectionContextBase(RequiredConfig):
         'prefix',
         doc="a prefix to use inside the bucket",
         default='',
+        reference_value_from='resource.boto',
+        likely_to_be_changed=True,
+    )
+    required_config.add_option(
+        'keybuilder_class',
+        default='socorro.external.boto.connection_context.KeyBuilderBase',
+        doc=(
+            'fully qualified dotted Python classname to handle building s3 '
+            'pseudo-filenames'
+        ),
+        from_string_converter=class_converter,
         reference_value_from='resource.boto',
         likely_to_be_changed=True,
     )
@@ -92,6 +164,7 @@ class ConnectionContextBase(RequiredConfig):
             boto.exception.StorageResponseError,
             KeyNotFound
         )
+        self.keybuilder = config.keybuilder_class()
 
         self._bucket_cache = {}
 
@@ -112,13 +185,11 @@ class ConnectionContextBase(RequiredConfig):
         raise NotImplementedError
 
     #--------------------------------------------------------------------------
-    @staticmethod
-    def build_key(prefix, name_of_thing, id, version='v1'):
+    def build_keys(self, prefix, name_of_thing, id):
+        """Builds an s3 pseudo-filename using the specified keybuilder class.
+
         """
-        Use S3 pseudo-directories to make it easier to list/expire later
-        {prefix}/{version}/{name_of_thing}/{id}
-        """
-        return '%s/%s/%s/%s' % (prefix, version, name_of_thing, id)
+        return self.keybuilder.build_keys(prefix, name_of_thing, id)
 
     #--------------------------------------------------------------------------
     def _get_bucket(self, conn, bucket_name):
@@ -146,7 +217,9 @@ class ConnectionContextBase(RequiredConfig):
         conn = self._connect()
         bucket = self._get_or_create_bucket(conn, self.config.bucket_name)
 
-        key = self.build_key(self.config.prefix, name_of_thing, id)
+        all_keys = self.build_keys(self.config.prefix, name_of_thing, id)
+        # Always submit using the first key
+        key = all_keys[0]
         key_object = bucket.new_key(key)
         key_object.set_contents_from_string(thing)
 
@@ -157,17 +230,20 @@ class ConnectionContextBase(RequiredConfig):
         conn = self._connect()
         bucket = self._get_bucket(conn, self.config.bucket_name)
 
-        key = self.build_key(self.config.prefix, name_of_thing, id)
-        key_object = bucket.get_key(key)
-        if key_object is None:
-            raise KeyNotFound(
-                '%s (bucket=%r key=%r) not found, no value returned' % (
-                    id,
-                    self.config.bucket_name,
-                    key,
-                )
+        all_keys = self.build_keys(self.config.prefix, name_of_thing, id)
+        for key in all_keys:
+            key_object = bucket.get_key(key)
+            if key_object is not None:
+                return key_object.get_contents_as_string()
+
+        # None of the keys worked, so raise an error
+        raise KeyNotFound(
+            '%s (bucket=%r keys=%r) not found, no value returned' % (
+                id,
+                self.config.bucket_name,
+                all_keys,
             )
-        return key_object.get_contents_as_string()
+        )
 
     #--------------------------------------------------------------------------
     def _convert_mapping_to_string(self, a_mapping):
--- a/socorro/unittest/external/boto/test_connection_context.py
+++ b/socorro/unittest/external/boto/test_connection_context.py
@@ -8,6 +8,9 @@ import mock
 
 from socorrolib.lib.util import DotDict
 from socorro.external.boto.connection_context import (
+    DatePrefixKeyBuilder,
+    KeyBuilderBase,
+    KeyNotFound,
     S3ConnectionContext,
     RegionalS3ConnectionContext,
 )
@@ -46,8 +49,7 @@ a_thing = {
 thing_as_str = json.dumps(a_thing)
 
 
-class TestCase(socorro.unittest.testbase.TestCase):
-
+class ConnectionContextTestCase(socorro.unittest.testbase.TestCase):
     def setup_mocked_s3_storage(
         self,
         executor=TransactionExecutor,
@@ -66,6 +68,7 @@ class TestCase(socorro.unittest.testbase.TestCase):
             'access_key': 'this is the access key',
             'secret_access_key': 'secrets',
             'bucket_name': 'silliness',
+            'keybuilder_class': KeyBuilderBase,
             'prefix': 'dev',
             'calling_format': mock.Mock()
         })
@@ -96,10 +99,15 @@ class TestCase(socorro.unittest.testbase.TestCase):
         prefix = 'dev'
         name_of_thing = 'dump'
         crash_id = 'fff13cf0-5671-4496-ab89-47a922141114'
-        good = connection_source.build_key(prefix, name_of_thing, crash_id)
+        all_keys = connection_source.build_keys(prefix, name_of_thing, crash_id)
+
+        # This uses the default key builder which should produce a single
+        # key--so we verify that there's only one key and that it's the one
+        # we're looking for.
+        self.assertEqual(len(all_keys), 1)
         self.assertEqual(
-            'dev/v1/dump/fff13cf0-5671-4496-ab89-47a922141114',
-            good
+            all_keys[0],
+            'dev/v1/dump/fff13cf0-5671-4496-ab89-47a922141114'
         )
 
     def test_submit(self):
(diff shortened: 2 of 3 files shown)


willkg:
The test plan for this is something along these lines:

1. Merge this.
2. Update the configuration on stage and see how it does
  1. Are there errors in the log?
  2. Sentry?
  3. Send some crashes to stage and then check the s3 stage bucket to see if they get a date prefix.
  4. Make sure those crashes get processed. If they do, then the `.fetch()` code is working with the new-style paths.
  5. See if the crashes we're siphoning off from prod process. If they do, then the `.fetch()` code is working with the old-style paths.

If all that goes well, I think this should work out.

Things I'm not sure about:

First, there are three things in the socorro repo that access s3 directly:

* `./webapp-django/crashstats/symbols/management/commands/reencode-bad-sym-files.py`
* `./webapp-django/crashstats/symbols/views.py`
* `./scripts/cmp_hbase_s3.py`

The third of those, I wrote up a bug for. The first two look like they doing something with s3 that's limited to symbol something or other and doesn't involve raw crash data at all.

Are there other scripts or anything else that access s3 directly that is hard-coded for the old paths? Maybe things that live outside the socorro repo?

Second, I'm pretty sure that even if I screwed up, this will fail ok and that we can back it out or fail forward. Does that sound right?

socorro/external/boto/connection_context.py
```diff
@@ -14,20 +14,72 @@
...
+        Then it also returns the old-style key so that we can fetch objects
+        that were saved with the old key::
+
+            {prefix}/{version}/{name_of_thing}/{id}
+
+        """
+        keys = []
+
+        if name_of_thing == 'raw_crash':
+            datestamp = dateFromOoid(id)
```
willkg: `dateFromOoid` returns `None` if it can't extract a date. That code looks pretty solid, so I rely on that here.

socorro/external/boto/connection_context.py
```diff
@@ -14,20 +14,72 @@
...
+
+            if datestamp is not None:
+                date = datestamp.strftime('%Y%m%d')
```
willkg: I switch it to a 4-digit year because that's what I think that's the format of the least surprise.

Does that sound right? Is there a precedent for 2-digit years in Socorro other than the ooid which has space restrictions?

socorro/external/boto/connection_context.py
```diff
@@ -157,17 +221,20 @@ def fetch(self, id, name_of_thing):
...
+
+        # None of the keys worked, so raise an error
+        raise KeyNotFound(
+            '%s (bucket=%r key=%r) not found, no value returned' % (
+                id,
+                self.config.bucket_name,
+                all_keys,
             )
-        return key_object.get_contents_as_string()
+        )
```
willkg: I think I got these changes right, but double-checking them and thinking about scenarios where this fails would be helpful.

socorro/external/boto/connection_context.py
```diff
@@ -14,20 +14,72 @@
...
+
+        Then it also returns the old-style key so that we can fetch objects
+        that were saved with the old key::
+
+            {prefix}/{version}/{name_of_thing}/{id}
+
+        """
+        keys = []
+
+        if name_of_thing == 'raw_crash':
```
willkg: I'm pretty sure we want this change *only* for `raw_crash`, but I'm not 100% positive about that. Does the collector create other s3 object types?

socorro/unittest/external/boto/test_connection_context.py
```diff
@@ -218,3 +226,294 @@ def test_fetch_with_regional_s3connection_context(self):
...
+    ):
+        config = DotDict({
+            'resource_class': resource_class,
+            'logger': mock.Mock(),
+            'host': host,
+            'port': port,
+            'access_key': 'this is the access key',
+            'secret_access_key': 'secrets',
+            'bucket_name': 'silliness',
+            'keybuilder_class': DatePrefixKeyBuilder,
```
willkg: This method is exactly the same as the original one except for this line because we need it to use a `DatePrefixKeyBuilder`. Figured I'd  mention that for easier review.

willkg:
We talked on IRC where we determined we want to make the following changes:

1. change it to a YYYYmmddHHMM timestamp
2. add an additional 3-character prefix *before* the timestamp taken from the first three characters of the ooid so we don't overwhelm s3 partitions
3. change "version" in the various method signatures so that it always gets determined by the KeyBuilder class rather than passed by the caller of `.build_keys()`
4. the new-style key should be v2
5. given all that, we should rename it to TimestampPrefixKeyBuilder

I'll do that tomorrow.

willkg:
Relevant s3 docs:

* https://docs.aws.amazon.com/AmazonS3/latest/dev/request-rate-perf-considerations.html#workloads-with-mix-request-types
* http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-keys

willkg:
Nix item 1 and 5 because the date is coming from the end of the ooid which doesn't include a time.

Still working on items 2, 3, and 4.

willkg:
I spent some time re-reading the links. I think we don't need the additional 3-character prefix at all.

Here's my thinking:

1. The "request-rate-perf-considerations" document has two examples. The first example uses date + time down to the second and because seconds change once a second. a series of consecutive PUTs will end up with keys where the first differing character for a set of keys will be a sequence. The second example uses a sequence id.
2. The document talks about grouping keys using the prefix and putting that grouping before the randomized part.
3. In our situation, we use a grouping (e.g. `raw_crash`) and we're using date--not date + time. Given that, consecutive PUTs all have the same prefix up to the ooid which is random.

Given that, seems like we don't need the 3-character randomness before the date.

I still have some doubt. Does s3 look at the first differing character to determine the partition? Does s3 look at the first n characters to determine the partition? If it's the first n characters, how big is n?

I'll follow up with jp.

willkg:
I'm still working on this... I'll have it finished up on Monday.

socorro/external/boto/connection_context.py
lonnen: The collector does not. Further into the ingestion pipeline we store the `processed_crash` and we may want a similar naming convention change there.

socorro/external/boto/connection_context.py
```diff
@@ -14,20 +14,78 @@
...
+            datestamp = dateFromOoid(id)
+
+            if datestamp is not None:
+                # We insert the first 3 chars of the ooid which are hex
+                # so that we don't overwhelm s3 partitions by blasting
+                # a bunch of objects with the exact same prefix.
+                first_chars = id[:3]
+                date = datestamp.strftime('%Y%m%d')
+                keys.append(
+                    '%s/v2/%s/%s/%s/%s' % (
```
willkg: Version is now hard-coded in the schema because the version is about the schema.

willkg:
Changes ^^^.

I changed my mind again. Everything I'm reading about s3 and partitions can get watered down to "put something random as left-most as you can to reduce the likelihood you bottleneck on partitions for large loads". Our current scheme doesn't have anything random until pretty late in the string. Given that, I'm going to put something random before the date and thus maintain equivalent to what we've got now with `raw_crash`.

Old:

```
<prefix>/<version>/raw_crash/<ooid>
                             ^ first random character
```

New:

```
<prefix>/<version>/raw_crash/<random-prefix>/<date>/<ooid>
                             ^ first random character
```

Further, the "random-prefix" is really the first three characters of the ooid, so it should behave almost the same except under loads where three characters isn't enough.

r?

socorro/external/boto/connection_context.py
```diff
@@ -14,20 +14,78 @@
...
+        """
+        Use S3 pseudo-directories to make it easier to list/expire later
+        {prefix}/v1/{name_of_thing}/{id}
+        """
+        return [
+            '%s/v1/%s/%s' % (prefix, name_of_thing, id)
+        ]
+
+
+class DatePrefixKeyBuilder(KeyBuilderBase):
```
peterbe: Whenever I read that I ready "dateprefix key builder" :)
It's note a dateprefix. It's "date key builder with prefix" or something.

socorro/external/boto/connection_context.py
```diff
@@ -14,20 +14,78 @@
 from configman import Namespace, RequiredConfig, class_converter
 
 from socorrolib.lib.converters import change_default
+from socorrolib.lib.ooid import dateFromOoid
```
peterbe: The fact that it's called "ooid" (and the case style of the function) tells me this one is old. Feel free to evaluate if it's worth writing a new one. By the way, the name of a crash ID has changed over the years from ooid to uuid to crash ID.

peterbe:
Is the "prefix" the bucket or something else? Or is it just legacy that we never ever use? 
Can we re-evaluate it? I also noticed it's the first (and not optional) parameter on things like `build_keys()`. 
I think it's weird that the version (e.g. `v1` or `v2`) is not the very first thing.

peterbe:
The files `./webapp-django/crashstats/symbols/management/commands/reencode-bad-sym-files.py` and `./webapp-django/crashstats/symbols/views.py` have nothing to do with crashes. It's S3 but that's just a "coincidence".

peterbe:
r+ on the general idea! Now I just need to dig into the code. 

I really like this idea of turning the GET into a seek by trial-and-error. That means we don't need a particular config switch-over date timing. It'll just work. 

I'm pessimistic towards using configuration to enable this. Instead, I think we should aim to get it right and if it turns out we made a mistake, we either revert or fail forwards. If you want to, you get "carve out some time" on stage by informing the team "Lemme brew something in stage for a day or two. No deployments please". We're small enough team to be able to do that. Also, that avoid having to remember to clean up the config when it's proven to work.

socorro/external/boto/connection_context.py
```diff
@@ -14,20 +14,78 @@
...
+            {prefix}/v1/{name_of_thing}/{id}
+
+        """
+        keys = []
+
+        if name_of_thing == 'raw_crash':
+            datestamp = dateFromOoid(id)
+
+            if datestamp is not None:
+                # We insert the first 3 chars of the ooid which are hex
```
peterbe: s/ooid/crash ID

willkg:
@peterbe "I'm pessimistic towards using configuration to enable this." I'm not sure what you mean by that. What's the alternative with the current architecture?

socorro/external/boto/connection_context.py
```diff
@@ -14,20 +14,78 @@
...
+                # so that we don't overwhelm s3 partitions by blasting
+                # a bunch of objects with the exact same prefix.
```
peterbe: I don't understand this. S3 is not like a filesystem where you have to worry about not putting too many files in a directory. (...which I believe is bad because of `ls` freaking out amongst other things).
Please elaborate.

socorro/external/boto/connection_context.py
willkg: Is it "crash_id" for all things stored in s3 that uses this component?

socorro/external/boto/connection_context.py
```diff
@@ -14,20 +14,78 @@
...
+                first_chars = id[:3]
```
peterbe: Isn't the first 3 randomish? How then am I supposed to be able to list all S3 keys by date? E.g. all for June 2016.

socorro/external/boto/connection_context.py
```diff
@@ -60,6 +118,17 @@ class ConnectionContextBase(RequiredConfig):
...
+    required_config.add_option(
+        'keybuilder_class',
+        default='socorro.external.boto.connection_context.KeyBuilderBase',
+        doc=(
+            'fully qualified dotted Python classname to handle building s3 '
+            'pseudo-filenames'
+        ),
+        from_string_converter=class_converter,
+        reference_value_from='resource.boto',
+        likely_to_be_changed=True,
```
peterbe: Why?

socorro/external/boto/connection_context.py
```diff
@@ -112,13 +182,11 @@ def _get_credentials(self):
         raise NotImplementedError
 
     #--------------------------------------------------------------------------
-    @staticmethod
-    def build_key(prefix, name_of_thing, id, version='v1'):
```
peterbe: Would it be value in keeping this? It could easily be:
```python
def build_key(self, prefix, name_of_thing, id, version='v1'):
    return [self.keybuilder.build_keys(prefix, name_of_thing, id, version=version)]
```
Note, if you think this is a good idea, we'd need to have `version='v2'` in the new fancy `build_keys()` method.

peterbe:
Yeah, I'm still dubious about the prefix. It's weird that we do `self.build_keys(self.config.prefix, ...)`
If we indeed want to keep it, I think we should stop having it as a method parameter.

socorro/external/boto/connection_context.py
```diff
@@ -157,17 +227,20 @@ def fetch(self, id, name_of_thing):
...
-                    key,
-                )
+        all_keys = self.build_keys(self.config.prefix, name_of_thing, id)
+        for key in all_keys:
+            key_object = bucket.get_key(key)
+            if key_object is not None:
+                return key_object.get_contents_as_string()
+
+        # None of the keys worked, so raise an error
+        raise KeyNotFound(
```
peterbe: Was that what it raised before? Isn't there code that does something like this:
```python
try:
    crash_storage.get(id, name_of_thing)
except CrashNotFound:
    ...
```

socorro/external/boto/connection_context.py
```diff
@@ -157,17 +227,20 @@ def fetch(self, id, name_of_thing):
...
+            '%s (bucket=%r key=%r) not found, no value returned' % (
```
peterbe: ought to perhaps have ` prefix=%r` too in the bracket.

socorro/unittest/external/boto/test_connection_context.py
```diff
@@ -8,8 +8,11 @@
 
 from socorrolib.lib.util import DotDict
 from socorro.external.boto.connection_context import (
+    DatePrefixKeyBuilder,
+    KeyBuilderBase,
+    KeyNotFound,
     S3ConnectionContext,
-    RegionalS3ConnectionContext,
+    RegionalS3ConnectionContext
```
peterbe: A trailing comma is a good thing. It means, next time someone adds to that list, their diff is less.

socorro/unittest/external/boto/test_connection_context.py
```diff
@@ -218,3 +226,294 @@ def test_fetch_with_regional_s3connection_context(self):
...
+        prefix = 'dev'
+        name_of_thing = 'raw_crash'
+        crash_id = 'fff13cf0-5671-4496-ab89-47a922141114'
+        all_keys = connection_source.build_keys(prefix, name_of_thing, crash_id)
+
+        # The new- and old-style keys are returned
+        self.assertEqual(len(all_keys), 2)
+        self.assertEqual(
+            all_keys[0],
+            'dev/v2/raw_crash/fff/20141114/fff13cf0-5671-4496-ab89-47a922141114'
```
peterbe: Yeah, that `fff` makes it impossible to list all S3 objects from November 2014.

socorro/external/boto/connection_context.py
peterbe: Can't answer with confidence. But what I can say with confidence is that we uses to refer to it as "ooid" and later changed it to "uuid" and later changed it again to "crash ID". There was a point where we pondered to rename the postgres column to be called "crash_id" (instead of "uuid") but an ALTER TABLE command like that would be deemed possibly too time expensive on such a crucial table. New postgres things calls it "crash_id" instead. 

Just to make things even more confusing, see https://github.com/mozilla/socorro/blob/master/socorro/external/postgresql/models.py#L1056 :)

But basically, if it's a comment of English it should ideally be "crash ID".

willkg:
I talked with @peterbe about a bunch of his comments:

1. I incorrectly thought that prefix was the bucketname--they're two different things. I'll go through the code changes to make sure I'm not conflating the two.
2. We talked about the 3-character thing and how it affects getting a listing of all crashes for a given date plus how it affects s3 partitions. This was important because in this PR I flip-flopped on my thoughts on this several times making my comments somewhat incoherent.
3. We talked about uuid -> ooid -> crash_id nomenclature. One problem we have with socorro is that a given thing has different names depending on where you are in the codebase and that's somewhat confusing. I'll go through the code changes and make sure I'm not making this problem worse in regards to usage of uuid/ooid/id/crash_id.

I think that amicably covers everything.

I'll work on double-checking that now.

peterbe:
Just to be clear to others looking at this other than Will and I. This new key naming convention will NOT "prefix" only on the date. If you want to download/list all crashes for a specific day, you need to try one query per every combination of 3 characters random 0-9a-f. Will believes this makes for faster queries on the S3 side but you have to do that query more times.

socorro/external/boto/connection_context.py
willkg: The key includes the prefix here, so we should be all set.

willkg: Er, it should be changed to "keys", though. I'll fix that now.

socorro/external/boto/connection_context.py
willkg: `.build_keys()` returns a list, so I think if we were to do this, it'd return the first item of what `.build_keys()` returned.

Regardless, I looked through the codebase and the only thing that uses `.build_key()` is the classes defined in this directory plus the tests that test it. I couldn't think of a compelling reason to have `.build_key()` other than backwards compatibility. Given that, I decided it was better to remove it altogether.

socorro/unittest/external/boto/test_connection_context.py
willkg: Totally my fault--I didn't intend to remove this.

willkg:
^^^ That fixes the outstanding things.

I'm going to run this by @lonnen and then coordinate with socorro devs on landing this.

socorro/unittest/external/boto/test_connection_context.py
lonnen: It's not impossible. You'd have to try all `16 ^ 3` combinations. This is easy enough to script though. You can see how the GETs to retrieve a day scale here: https://www.wolframalpha.com/input/?i=table+30+*+(16+%5E+z)+from+0+to+4

willkg:
I talked with @lonnen and we decided to keep this schema and keep 3 characters of entropy, but switch it to 2 if @jdotpz tells us that we're rate-limited on s3 bucket listing requests.

Waiting on learning the outcome of that.

lonnen:
There are no hard limits, but everything has a soft limit. I'm waiting to hear back from our solutions architect.

PR #3333, opened 2016-05-06
fixes bug 1270611 - Split up reprocessing_jobs to commit delete after…

peterbe:
… each save

cc @twobraids

### DIFF
--- a/socorro/cron/jobs/reprocessingjobs.py
+++ b/socorro/cron/jobs/reprocessingjobs.py
@@ -5,18 +5,19 @@
 
 from crontabber.base import BaseCronApp
 from crontabber.mixins import (
-    with_postgres_transactions,
-    with_single_postgres_transaction,
-    with_transactional_resource
+    with_transactional_resource,
+    using_postgres,
+    with_postgres_connection_as_argument,
 )
-from socorro.external.postgresql.dbapi2_util import execute_query_iter
+from socorro.external.postgresql.dbapi2_util import (
+    execute_query_iter,
+    execute_no_results,
+)
 from socorrolib.lib.util import DotDict
 
-_reprocessing_sql = """ DELETE FROM reprocessing_jobs RETURNING crash_id """
 
-
-@with_postgres_transactions()
-@with_single_postgres_transaction()
+@using_postgres()
+@with_postgres_connection_as_argument()
 @with_transactional_resource(
     'socorro.external.rabbitmq.crashstorage.ReprocessingRabbitMQCrashStore',
     'queuing'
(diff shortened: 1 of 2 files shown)


peterbe:
I snook in a semi-separate change. I capped the number of crash_ids (out of the `reprocessing_jobs` table) to 10,000. That's just a safe guard. It means, if there's more than 10,000 in the table, we're only sending RabbitMQ 10,000 per every 5 minutes. That'll help RabbitMQ avoid "ever" running out of memory. 
And again, if it was to happen that RabbitMQ refuses to receive, crontabber will just calmly try again till it works.

socorro/cron/jobs/reprocessingjobs.py
```diff
@@ -30,10 +31,24 @@ class ReprocessingJobsApp(BaseCronApp):
...
+        for crash_id in crash_ids:
             self.queuing_connection_factory.save_raw_crash(
                 DotDict({'legacy_processing': 0}),
                 [],
                 crash_id
             )
+            execute_no_results(connection, delete_sql, {
+                'crash_id': crash_id
+            })
+            connection.commit()
```
adngdb: Wouldn't it be a problem to commit 1000 times in a row?

adngdb: I think I would encapsulate this in a ``try`` block and do the commit on the ``finally``. This way you still only delete what has been reprocessed, but you commit only once.

peterbe: Good idea!

twobraids: if you move to a single commit on the deletes, you've returned the method to the same behavior that it had before your refactoring.

peterbe: @twobraids No. If you do:
```
> BEGIN;
> DELETE FROM reprocessing_jobs RETURNING crash_id;
> # send 50% of those returned crash_ids to a third party
> ROLLBACK;
> # EVERYthing will be rolled back. 
```

I think what adrian is suggesting is this:
```
# inside the crontabber app with manual transaction handling
cursor = conn.cursor()
crash_ids = (get python list of 10,000 crash_ids for `cursor`)
try:
    for crash_id in crash_ids:
      (send `crash_id` to rabbit)
      conn.execute("delete from reprocessing_jobs where crash_id = %s", crash_id)
finally:
    conn.commit()  # commit all those deletes we've amassed even if there's an exception raised
```

twobraids: I misunderstood.

so what happens if ```conn.execute``` raises an exception?

peterbe: > so what happens if conn.execute raises an exception? 

That **one** crash ID would be re-submitted for processing.

twobraids: but the exception will break out of the loop.  the commit will fail because you can't commit on a connection that has an error state.  I think the whole transaction will be lost even though the rabbit inserts succeeded.

peterbe: You're right. I forgot about that. If the `.execute()` fails, the `.commit()` will also fail. 

So I think it's best, and safe to leave it as is. 10,000 commits. Solid and safe. And rare too. And, best of all, if 10,000 is too much and it causes pain on the PG side, crontabber will simply try again 5 minutes later.

adngdb: I think we can find a compromise. Can we try committing every 100 submitted crashes? This way, if the postgres connection fails, we let that bubble up, and next time crontabber re-runs this job we know that only at most 100 crashes will be re-reprocessed. 

Note: this is based on a gut feeling that committing 10,000 times to postgres is a bad idea, but maybe it isn't such a big deal. I don't know. :)

peterbe: I don't think it's worth it. It's so rare and 10,000 commits is unlikely to be a big problem. If it is, we can revise in the future.

PR #3335, opened 2016-05-06
fixes bug 1270662 - Supersearch drops special characters from URL

peterbe:
This certainly fixes the problem. 
Before this fix, loading [this url](http://socorro.dev/search/?product=Firefox&app_notes=~WGL%2B&_facets=signature&_columns=date&_columns=signature&_columns=product&_columns=version&_columns=build_id&_columns=platform#facet-signature) would insert into the `app_notes` search this value: `WGL`. After this change that value becomes `WGL+`. 

But I suspect this line that I'm removing might be there for a good reason. Some other reason. Perhaps an old reason.


adngdb:
SuperSearch and Signature report are the only two pages that use this function, so it should be safe. Tested locally, it worked fine.

PR #3336, opened 2016-05-06
Fixes bug 1270598: Remove cmp_hbase_s3 and sync_hbase_s3 scripts

willkg:
This removes two scripts which we're not using anymore.

Quick r?


peterbe:
💋

PR #3337, opened 2016-05-09
fixes bug 1271267 - Re-wrap APIWhitelistError as 404

peterbe:


adngdb:
I think this is the wrong solution. The initial problem is that we expose something that should not be (the ``ESSocorroMiddleware`` class). I think it is a fine thing to have a 500 if something exposed doesn't have a white list, because when that happens it means we have done a poor job.

peterbe:
I later discovered that that's a different problem. The problem still exists that you can get a 500 error by going to `/api/SocorroMiddleware`.

adngdb:
I still think that, if we get to the point where we check the API white list, that means the service is expected to be user facing. And thus, it not having a white list is an error on our side, which should raise a 500 anywhere, so we can get an alert in sentry. 

If we want a service not to be exposed, we should add it to the black list.

PR #3338, opened 2016-05-09
Fixes bug 1271281 - Hide API base classes from public eyes.

adngdb:
What do you think about this?

### DIFF
--- a/webapp-django/crashstats/api/tests/test_views.py
+++ b/webapp-django/crashstats/api/tests/test_views.py
@@ -86,6 +86,15 @@ class TestViews(BaseTestViews):
         response = self.client.get(url)
         eq_(response.status_code, 404)
 
+    def test_base_classes_raise_bad_request(self):
+        url = reverse('api:model_wrapper', args=('SocorroMiddleware',))
+        response = self.client.get(url)
+        eq_(response.status_code, 400)
+
+        url = reverse('api:model_wrapper', args=('ESSocorroMiddleware',))
+        response = self.client.get(url)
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -86,6 +86,15 @@ def test_invalid_url(self):
         response = self.client.get(url)
         eq_(response.status_code, 404)
 
+    def test_base_classes_raise_bad_request(self):
+        url = reverse('api:model_wrapper', args=('SocorroMiddleware',))
+        response = self.client.get(url)
+        eq_(response.status_code, 400)
```
peterbe: Should be 404. 400 is when the URL is valid but there's something wrong with the parameters. 
Equally, if you do `curl .../api/cm49b7aktgja7v` that should definitely be a 404.

webapp-django/crashstats/api/views.py
```diff
@@ -223,6 +223,16 @@ def model_wrapper(request, model_name):
     else:
         raise http.Http404('no model called `%s`' % model_name)
 
+    if (
+        not issubclass(model, models.SocorroMiddleware) or
+        model is models.SocorroMiddleware or
+        model is crashstats.supersearch.models.ESSocorroMiddleware
+    ):
```
peterbe: Please turn this into a function and re-use that around like 433.

peterbe:
r+ with nits but I wouldn't mind taking a second last look.

adngdb:
@peterbe edited!

PR #3340, opened 2016-05-10
fixes bug 1271621 - Add reason foreign key to correlations table

peterbe:


peterbe:
I was able to run this locally:
```
 PYTHONPATH=. alembic -c config/alembic.ini upgrade head
```
and it did add the column to my local PG.

adngdb:
I believe you forgot to update the associated model.

peterbe:
@adngdb Fixed now.

alembic/versions/a44fb7695c10_add_reasons_foreign_key_to_correlations.py
```diff
@@ -0,0 +1,52 @@
...
+            'date',
+            'signature_id',
+            'reason_id',
+        ],
+        unique=True,
+    )
+
+
+def downgrade():
+    op.drop_column(
```
adngdb: Don't you need to drop the index here as well? Or will it get removed automatically when removing one of its columns?

peterbe: That's what I had but it broke the tests. I'm trying to comprehend why.

PR #3342, opened 2016-05-10
Bug 1267783 store correlations in postgres [DO NOT MERGE]

peterbe:
This works on my laptop. I get correlations stored in my local PG.


adngdb:
@peterbe shall we close this PR for the time being?

PR #3344, opened 2016-05-11
fixes bug 1272053 - Refactor how we cache individual bug data cache

peterbe:
cc @adngdb 

The bug description is pretty extensive. 

The **tl;dr** is that we don't change anything on the outside. Instead, when calling the URL `/buginfo/bug?bugs_ids=123,124,125` that whole request does not get cached. Instead it individually caches `123`, `124` and `125`. 

Also, I refactored it so that with that view, you can't specify the `include_fields` any more. It's hardcoded. It's no longer trying to be a cached proxy to Bugzilla's API. It tries to be smarter than that. 

One cool benefit with this is that we can now write a Pulse consumer (I have a working prototype of this on my laptop) that connects to the same memcache as we use in production. Then it can immediately invalidate the cache so we'd be able to set the cache time to 24h or 7days AND never ever have stale information for the bug info. 

A possibly easier alternative is to make a crontabber app that queries bugzilla for ALL bugs that have changed in the last, say, 30 minutes and again connect to the same memcache, and invalidate them all.

But first things first, let's explode this model.

### DIFF
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -1518,19 +1518,41 @@ class BugzillaAPI(SocorroCommon):
 
 class BugzillaBugInfo(BugzillaAPI):
 
-    def get(self, bugs, fields):
+    cache_seconds = 0
+
+    @staticmethod
+    def make_cache_key(bug_id):
+        # This is the same cache key that we use in show_bug_link()
+        # the jinja helper function.
+        return 'buginfo:{}'.format(bug_id)
+
+    def get(self, bugs):
         if isinstance(bugs, basestring):
             bugs = [bugs]
-        if isinstance(fields, basestring):
-            fields = [fields]
-        params = {
-            'bugs': ','.join(bugs),
-            'fields': ','.join(fields),
-        }
-        headers = {'Accept': 'application/json',
-                   'Content-Type': 'application/json'}
-        url = ('/bug?id=%(bugs)s&include_fields=%(fields)s' % params)
-        return self.fetch(url, headers)
+        fields = ('summary', 'status', 'id', 'resolution')
+        results = []
+        missing = []
+        for bug in bugs:
+            cache_key = self.make_cache_key(bug)
+            cached = cache.get(cache_key)
+            if cached is None:
+                missing.append(bug)
+            else:
+                results.append(cached)
+        if missing:
+            params = {
+                'bugs': ','.join(missing),
+                'fields': ','.join(fields),
+            }
+            headers = {'Accept': 'application/json',
+                       'Content-Type': 'application/json'}
+            url = ('/bug?id=%(bugs)s&include_fields=%(fields)s' % params)
+            fetched = self.fetch(url, headers)
+            for each in fetched['bugs']:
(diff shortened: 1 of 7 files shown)


webapp-django/crashstats/crashstats/models.py
```diff
@@ -1518,19 +1518,41 @@ class BugzillaAPI(SocorroCommon):
 
 class BugzillaBugInfo(BugzillaAPI):
 
-    def get(self, bugs, fields):
+    cache_seconds = 0
```
willkg: I don't see this used anywhere. What's it for?

webapp-django/crashstats/crashstats/models.py
```diff
@@ -1518,19 +1518,41 @@ class BugzillaAPI(SocorroCommon):
...
+                'bugs': ','.join(missing),
+                'fields': ','.join(fields),
+            }
+            headers = {'Accept': 'application/json',
+                       'Content-Type': 'application/json'}
+            url = ('/bug?id=%(bugs)s&include_fields=%(fields)s' % params)
+            fetched = self.fetch(url, headers)
+            for each in fetched['bugs']:
+                cache_key = self.make_cache_key(each['id'])
+                cache.set(cache_key, each, 60 * 60)
```
willkg: I'd toss this `60 * 60` into a class attribute or a module attribute so it's easier to find and change later.

willkg:
I think this is ok.

Is there a follow-up bug for invalidating the cache when bug status changes, yet?

webapp-django/crashstats/crashstats/models.py
peterbe: Every model has a `cache_seconds`. So if you do `SomeClassInThisFile().get(**parameters)` it will cache the results for `self.cache_seconds`. That number is often not specified in the classes in this file. Because they all lazily inherit from the base class which sets the default (which I believe is 60). 

By setting this to `0` it means this model's internal `get()` doesn't read from or write to the cache. That's because I instead cache the individual bits inside the `.get()` method.

webapp-django/crashstats/crashstats/models.py
peterbe: The value is never used anywhere else. So there's no need to re-use it as a variable. By putting it in the class as a class attribute for example (or in `settings/base.py`) doesn't add much value because it's then further away from where it's being used. So moving it further away from where it's used and matter isn't a big win. What do you think?

peterbe:
No plan to do invalidation. It might not matter.

webapp-django/crashstats/crashstats/models.py
willkg: I think something like:

```
cache.set(cache_key, each, self.CACHE_LIFESPAN)
```

is clearer in its intention.

willkg:
I thought you talked about plans to add cache invalidation in the PR description which is why I asked if you were tracking it in a bug, yet.

Anyhow, this is fine. r+

peterbe:
@willkg I've updated it. Please feel free to merge when you and travis is happy.

willkg:
Looks good! r+

PR #3345, opened 2016-05-11
fixes bug 1272050 - Report index breaks on failing counting correlations

peterbe:


webapp-django/crashstats/crashstats/static/crashstats/js/socorro/report.js
```diff
@@ -5,6 +5,11 @@ var SignatureCorrelations = (function() {
     var container = $('#mainbody');
     return {
         showSignatureCorrelationsTab: function() {
+            if (!container.data('product')) {
```
adngdb: You're verifying that ``product`` is false, but in the comment you talk about ``platform``. Which one holds the truth?

webapp-django/crashstats/crashstats/views.py
```diff
@@ -69,6 +69,10 @@
 
 
 def make_correlations_count_cache_key(product, version, platform, signature):
+    assert product
+    assert version
+    assert platform
+    assert signature
```
adngdb: I don't really like using ``assert`` to much in the code. You already make sure all of those have a value before calling the function, what's the point of verifying that twice?

peterbe: I used it to debug. I took care of the real problem elsewhere later but forgot to delete these lines.

peterbe:
@adngdb Nits addressed.

PR #3346, opened 2016-05-11
parallelize travis build [DO NOT MERGE, PROTOTYPE]

peterbe:


peterbe:
Here's the email I sent to the list concluding why this is not feasible::

Because I'm curious I wanted to try running our travis test suite in parallel. Turns out, it's not so easy. https://github.com/mozilla/socorro/pull/3346

It's not hard to get parallelizing set up. You just set up, using environments, but the problem is that all the `install` and `before_script` and `before_install` gets executed **on each environment**.

In Socorro's case, the ./scripts/bootstrap.sh takes about 3m35s. So multiply that by 4.
The ./scripts/install.sh and ./scripts/package.sh parts take about 11s and 15s each. Also, multiply that by 4.
The meat of the testing, for socorro takes about 51s and the webapp 28s.
So reusing the environment would be the big time-saver. But that leaves the only jobs that don't need the environment (the work of ./scripts/bootstrap.sh) to just be silly (e.g. puppet lint and webapp flake8).

The conclusion is, it's not worth it for Socorro. We can't parallelize it without also having to pay for the expensive setup "costs". If it was possible to post-fork the environment instead, it might be a different story.

PR #3348, opened 2016-05-17
Fixes bug 1273538 - Correctly encode white spaces in URLs.

adngdb:
``$.param`` uses the ``+`` sign to encode white spaces. So I added a new library to Socorro, called ``Qs``, to deal with query string creation.

### DIFF
--- a/webapp-django/crashstats/api/static/api/js/testdrive.js
+++ b/webapp-django/crashstats/api/static/api/js/testdrive.js
@@ -78,7 +78,7 @@
         // The second parameter (`true`) is so that things like
         // `{products: ["Firefox", "Thunderbird"]}`
         // becomes: `products=Firefox&products=Thunderbird`
-        var qs = $.param(form.serializeExclusive(), true);
+        var qs = Qs.stringify(form.serializeExclusive(), { indices: false });
         if (qs) {
             url += '?' + qs;
         }
--- a/webapp-django/crashstats/base/templatetags/jinja_helpers.py
+++ b/webapp-django/crashstats/base/templatetags/jinja_helpers.py
@@ -86,6 +86,9 @@ def change_query_string(context, **kwargs):
             # change it
             qs[key] = value
     new_qs = urllib.urlencode(qs, True)
+
+    # We don't like + as the encoding character for spaces. %20 is better.
+    new_qs = new_qs.replace('+', '%20')
     if new_qs:
         return '%s?%s' % (base, new_qs)
     return base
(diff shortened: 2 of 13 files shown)


peterbe:
We chatted about it briefly on IRC, and the question was; why is this Qs thing better than `$.param`?
I saw that you're doing the string replace of `+` for `%20` anyway, which was the beef we had with `$.param` and I believe that was entirely fixed with that one string replace.

adngdb:
There are two different things happening here. The first one is that I replaced ``$.param`` with ``Qs.stringify``, because the latter uses ``%20`` as the encoding character for white spaces (as opposed to ``+``, which causes bugs with the rest of our code). The second thing is that Python's ``urllib.urlencode`` also uses ``+`` to encode white spaces, so I am replacing all ``+`` characters with the "correct" ``%20`` string. This way we are consistent everywhere (I hope) in our encoding, and we do not use the ``+`` character as an encoding character.

peterbe:
I think I get what this PR does. You consolidate so it's `%2B` instead of `+` BOTH in python AND in all javascript. In other words, I understand the code in the PR. But...

I don't want to stand in your way for progress but I'm also interested in learning. So...

What's the problem we're trying to fix? Like, what is broken/breaking?
One possible answer is that maybe [this pr](https://github.com/mozilla/socorro/pull/3335) wouldn't have been necessary if all `+` gone.

peterbe:
By the way, instead of adding a big new js library you could just define a global function like
```javascript
window.urlencode = function(params) {
    return $.param(params, {indices: false}).replace(/\+/g, '%2B');
};
```

adngdb:
#3335 was indeed the root of this problem. It fixed an issue when users were writing things like ``WebGL+``, but it broke cases when they used phrases like ``foo bar``. Those are currently turned into ``foo+bar``, and that's then interpreted as such by Super Search because we removed the replacement of ``+`` with white spaces. If you want to see this for yourself, go to Super Search on prod and search for ``cpu info`` ``starts with`` ``AuthenticAMD family 20``. You should see quite a few crash reports. Now click any of the signatures to go to Signature report, and it will have no results, because the query became ``cpu info`` ``starts with`` ``AuthenticAMD+family+20``.

peterbe:
Ah! So this solves something I broke. Hehe. Thanks for the clear example. 

r+

But I still think you should consider taking a look at [my `window.urlencode`](https://github.com/mozilla/socorro/pull/3348#issuecomment-220027057) function first.

adngdb:
I understand the value of not adding too much JS to our pages, but I also like using focused libraries instead of big, do-it-all ones like jQuery. ``Qs`` takes care of query strings and does it well for what I have seen. 

Also, I find it a lot cleaner and less confusing than your hack. It's easier to find documentation about how this library works than to have to search our code to find out where our hack is.

PR #3356, opened 2016-05-25
fixes bug 1275359 - Reprocess from report index page

peterbe:
See:
screen shot 2016-05-25 at 4 27 36 pm

This is quite a lot a bigger change than I had planned. It's because we don't have any exposed API endpoints that take only POST. So, apart from regular code reviews, here's what you ought to also test. 

First of all, make sure you have run `./manage.py migrate` so that you have the new reprocessing permission. 

Then, go to a report index page. Notice the new Reprocessing tab. It should just work. To trigger the AJAX error state you have to temporarily "mess" with the code or just switch off your django server (so you'd get a http error). 

You should now also be able to reprocess crashes from the command line. E.g.:
```
curl -v -X POST -H "Auth-token: 8b9bfcb0872d4b1881a61e5d7247d617" -d 'crash_ids=9a715a26-02d0-4fe8-856c-4cdcd2160524' -d 'crash_ids=7f6cfce8-0c6b-4d9c-a59a-562da2160524' http://socorro.dev/api/Reprocessing/
```
(If you run the same with `-X GET` you should get a 405 error)

The last thing to test is that the rest of `testdrive.js` still works. Just execute any endpoint from the API documentation page. Like the ProductVersions.

### DIFF
--- a/socorro/external/rabbitmq/crashstorage.py
+++ b/socorro/external/rabbitmq/crashstorage.py
@@ -317,12 +317,12 @@ class ReprocessingOneRabbitMQCrashStore(ReprocessingRabbitMQCrashStore):
     def reprocess(self, crash_ids):
         if not isinstance(crash_ids, (list, tuple)):
             crash_ids = [crash_ids]
-        all_ = True
+        success = bool(crash_ids)
         for crash_id in crash_ids:
             if not self.save_raw_crash(
                 DotDict({'legacy_processing': 0}),
                 [],
                 crash_id
             ):
-                all_ = False
-        return all_
+                success = False
+        return success
--- a/socorro/external/rabbitmq/crashstorage.py
+++ b/socorro/external/rabbitmq/crashstorage.py
@@ -314,9 +314,15 @@ class ReprocessingOneRabbitMQCrashStore(ReprocessingRabbitMQCrashStore):
         'socorro.reprocessing'
     )
 
-    def reprocess(self, crash_id):
-        return self.save_raw_crash(
-            DotDict({'legacy_processing': 0}),
-            [],
-            crash_id
-        )
+    def reprocess(self, crash_ids):
+        if not isinstance(crash_ids, (list, tuple)):
+            crash_ids = [crash_ids]
+        all_ = True
+        for crash_id in crash_ids:
+            if not self.save_raw_crash(
+                DotDict({'legacy_processing': 0}),
+                [],
+                crash_id
+            ):
+                all_ = False
+        return all_
--- a/webapp-django/crashstats/api/tests/test_views.py
+++ b/webapp-django/crashstats/api/tests/test_views.py
@@ -1965,8 +1965,6 @@ class TestViews(BaseTestViews):
             return True
 
         Reprocessing.implementation().reprocess = mocked_reprocess
-        # func = ReprocessingOneRabbitMQCrashStore.implementation().reprocess
-        # func.side_effect = mocked_reprocess
 
         url = reverse('api:model_wrapper', args=('Reprocessing',))
         response = self.client.get(url)
@@ -1984,7 +1982,7 @@ class TestViews(BaseTestViews):
         perm = Permission.objects.get(
             codename='reprocess_crashes'
         )
-        # but make a token that only has the 'view_pii'
+        # but make a token that only has the 'reprocess_crashes'
         # permission associated with it
         token = Token.objects.create(
             user=user,
--- a/webapp-django/crashstats/api/tests/test_views.py
+++ b/webapp-django/crashstats/api/tests/test_views.py
@@ -25,6 +25,7 @@ from crashstats.crashstats.models import (
     ProductVersions,
     CrontabberState,
     CurrentProducts,
+    Reprocessing,
 )
 from crashstats.tokens.models import Token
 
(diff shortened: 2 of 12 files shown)


peterbe:
For context, my goal here is to expand on this some more. 
Remember that idea of Ted's where you'd do "something more advanced" before you press that Reprocess button. I.e. the possibility to change something so that next time it goes in for processing (note that I didn't say "re-processing") the processing does certain things differently.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -314,9 +314,15 @@ class ReprocessingOneRabbitMQCrashStore(ReprocessingRabbitMQCrashStore):
...
-    def reprocess(self, crash_id):
-        return self.save_raw_crash(
-            DotDict({'legacy_processing': 0}),
-            [],
-            crash_id
-        )
+    def reprocess(self, crash_ids):
+        if not isinstance(crash_ids, (list, tuple)):
+            crash_ids = [crash_ids]
+        all_ = True
```
adngdb: ``success`` seems to be a better name here.

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -1956,3 +1957,44 @@ def mocked_get(url, params, **options):
...
+
+    def test_Reprocessing(self):
+
+        def mocked_reprocess(crash_ids):
+            assert crash_ids == ['xxxx']
+            return True
+
+        Reprocessing.implementation().reprocess = mocked_reprocess
+        # func = ReprocessingOneRabbitMQCrashStore.implementation().reprocess
+        # func.side_effect = mocked_reprocess
```
adngdb: debug leftover

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -1956,3 +1957,44 @@ def mocked_get(url, params, **options):
...
+        response = self.client.get(url, params, HTTP_AUTH_TOKEN='somecrap')
+        eq_(response.status_code, 403)
+
+        user = User.objects.create(username='test')
+        self._add_permission(user, 'reprocess_crashes')
+
+        perm = Permission.objects.get(
+            codename='reprocess_crashes'
+        )
+        # but make a token that only has the 'view_pii'
```
adngdb: ``s/'view_pii'/'reprocess_crashes'``

webapp-django/crashstats/api/views.py
```diff
@@ -273,6 +273,8 @@ def model_wrapper(request, model_name):
         function = instance.post
     else:
         function = instance.get
+        if not function:
+            return http.HttpResponseNotAllowed(['POST'])
```
adngdb: Why make it something more generic? For example, unindent it one level and do:

```python
if not function:
    return http.HttpResponseNotAllowed([request.method])
```

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -644,6 +651,64 @@
                     </ul>
                 </div>
                 <!-- /correlation -->
+
+                {% if request.user.has_perm('crashstats.reprocess_crashes') %}
+                <div id="reprocess" class="ui-tabs-hide" data-crash-id="{{ report.uuid }}">
```
adngdb: It would make more sense to me to have that crash id in the form in a hidden input instead of in a data attribute. That would probably allow this to work without any JS. Progressive enhancement and stuff. ;-)

webapp-django/crashstats/crashstats/models.py
```diff
@@ -1858,17 +1858,25 @@ class GraphicsReport(SocorroMiddleware):
 
 
 class Reprocessing(SocorroMiddleware):
-    """Return true the supplied crash ID
+    """Return true if all supplied crash IDs
     was sucessfully submitted onto the reprocessing queue.
```
adngdb: s/was/were

adngdb:
The admin page for reprocessing is broken, it requires that you update it to use ``crash_ids`` instead of ``crash_id`` when calling the ``Reprocessing`` model.

socorro/external/rabbitmq/crashstorage.py
peterbe: I actually changed it. Now it does this instead:
```python
    def reprocess(self, crash_ids):
        if not isinstance(crash_ids, (list, tuple)):
            crash_ids = [crash_ids]
        success = bool(crash_ids)
        for crash_id in crash_ids:
            if not self.save_raw_crash(
                DotDict({'legacy_processing': 0}),
                [],
                crash_id
            ):
                success = False
        return success
```
That means that if the list of crash_ids is length 0, you get a `False` returned.

webapp-django/crashstats/api/views.py
peterbe: you are right! I'll do that.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
peterbe: I've gotten so used to `data-` attributes because they're so convenient with jQuery because they support types. 
Also, a `<input>` tag would add to the DOM which is odd for something that will never be displayed. 

Also, the point of putting it in the form would only be to be able to use `$(form).serialize()` or something. I'm not convinced it's worth it.

peterbe:
Admin reprocessing fixed now. Sad how the tests were too mockery. But it's just an admin page thing that we don't expect to use much.

PR #3359, opened 2016-05-27
fixes bug 889491 - Improve error messages for invalid queries

peterbe:
Don't ask why I dedicated an hour of my Friday afternoon to this. The short answer was that I was just curious. I wanted to play with the [`handler400`](https://docs.djangoproject.com/en/1.8/ref/urls/#django.conf.urls.handler400) but that's not at all what that's for. :)

Before:
screen shot 2016-05-27 at 3 55 55 pm
screen shot 2016-05-27 at 3 56 15 pm
screen shot 2016-05-27 at 3 56 32 pm

After:
screen shot 2016-05-27 at 3 54 18 pm
screen shot 2016-05-27 at 3 54 45 pm
screen shot 2016-05-27 at 3 55 13 pm

### DIFF
--- a/webapp-django/crashstats/base/jinja2/400.html
+++ b/webapp-django/crashstats/base/jinja2/400.html
@@ -1,0 +1,43 @@
+{% extends "crashstats_base.html" %}
+
+{% block page_title %}Bad Request{% endblock %}
+
+{% block product_nav_filter %}&nbsp;{% endblock %}
+
+{% block site_css %}
+{{ super() }}
+<style type="text/css">
+ul.errorlist li {
+    list-style: unset;
+    margin-left: 30px;
+}
+.body p {
+    color: red;
+}
+</style>
+{% endblock %}
+
+{% block content %}
+<div id="mainbody">
+    <div class="page-heading">
+        <h2>Bad Request</h2>
+    </div>
+    <div class="panel">
+        <div class="title">
+            <h2>The request's parameters could not be understood.</h2>
+        </div>
+        <div class="body">
+            {% if 'ul class="errorlist"' in error %}
+                {{ error | safe }}
+            {% else %}
+                <p>
+                    Error Message:
+                </p>
+                <pre>{{ error }}
+                </pre>
+            {% endif %}
+        </div>
+    </div>
+</div>
+
+{% endblock %}
--- a/webapp-django/crashstats/crashstats/middleware.py
+++ b/webapp-django/crashstats/crashstats/middleware.py
@@ -1,6 +1,8 @@
 from django import http
 from django.conf import settings
 from django.core.exceptions import MiddlewareNotUsed
+from django.shortcuts import render
+
 from crashstats.crashstats.models import BadStatusCodeError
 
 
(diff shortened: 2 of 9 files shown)


webapp-django/crashstats/base/jinja2/400.html
```diff
@@ -0,0 +1,43 @@
...
+    <div class="page-heading">
+        <h2>Bad Request</h2>
+    </div>
+    <div class="panel">
+        <div class="title">
+            <h2>The request's parameters could not be understood.</h2>
+        </div>
+        <div class="body">
+            {% if 'ul class="errorlist"' in error %}
+                {{ error | safe }}
```
adngdb: Are you sure this is really safe? Could you add tests that verify it can't lead to an XSS attack for example?

webapp-django/crashstats/crashstats/decorators.py
```diff
@@ -4,11 +4,28 @@
 from django.http import HttpResponseBadRequest, Http404
 from django.shortcuts import redirect
 from django.core.urlresolvers import reverse
+from django.shortcuts import render
 
 from . import utils
 from crashstats.base import ga
 
 
+def decorate_bad_request(view):
```
adngdb: A decorator named ``decorate`` is not very helpful to me. What about ``handle`` or ``catch`` instead?

adngdb:
I see you applied this new behavior to a handful of views. Is there not a way to apply to every view we have? Any 400 page should behave like that imo, and having to add another decorator to every view sounds painful.

webapp-django/crashstats/crashstats/decorators.py
peterbe: What about "dressup" or "prettify" or "htmlify" or "redress"?
To me, "decorate" sounds like you're making something pretty. And it's very much about making a plain text thing into HTML, but "htmlify" just sounds like jargon :)

webapp-django/crashstats/base/jinja2/400.html
peterbe: Will do.

peterbe:
>  Is there not a way to apply to every view we have?

Yes, mayhaps if we write a middleware. If we just make sure to not meddle if the request is an AJAX one it might just work.

webapp-django/crashstats/crashstats/decorators.py
adngdb: Ah, OK, I see what you want to express. ``prettify_bad_request`` or ``better_bad_request`` sound good. 

But to be honest, the fact that we struggle finding a good name for this is another argument in favor of not making this a decorator. :)

webapp-django/crashstats/base/jinja2/400.html
peterbe: I couldn't think of a way to test that. So I hijacked one of the form validation functions and put in `raise forms.ValidationError('<script> ...`
This is what it looked like;
http://jmp.sh/CRDYEwV

peterbe:
@adngdb See the new update. I managed to do it with a piece of middleware. Now, if any view does something like...
```python
def myview(request):
    if condition:
        return http.HttpResponseBadRequest('Bad!')
```
it will be rendered as the `400.html` template. 

I've tested this...
* on the report index page, by replacing one of the characters in the crash ID with an x
* on the topcrashers page, by setting `&days=xxx`
* on the signature report, removing the `&signature=...`
* on the home page, by setting `&days=xxx`
* on the supersearch page, by entering a date that isn't a date (Note! Here it should NOT use `400.html`)

adngdb:
This looks good, but it needs to be rebased!

PR #3367, opened 2016-06-08
Fixes bug 803779 - Added a rule for Abort signatures.

adngdb:
@bsmedberg Does this look good to you? I think it covers all the different ``abort_message`` formats I could find in [Super Search](https://crash-stats.allizom.org/search/?product=Firefox&abort_message=abort&_columns=abort_message#crash-reports).


peterbe:
@adngdb How is this different from before?

adngdb:
@peterbe the latest commit is the only one you have never seen before. The first 2 made it into master at some point in our recent history.

PR #3370, opened 2016-06-10
document to only use ./manage.py migrate

peterbe:
`./manage.py syncdb` doesn't do anything any more. If it does, it's just doing the same as `migrate`.

### DIFF
--- a/docs/development/build-from-source.rst
+++ b/docs/development/build-from-source.rst
@@ -103,7 +103,6 @@ Django needs to write its ORM tables:
 ::
   export SECRET_KEY="..."
   cd webapp-django
-  ./manage.py syncdb --noinput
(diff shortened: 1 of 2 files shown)


willkg:
I hit a problem when I started with @adngdb helped me through. It was some problem with `auth_user`--I forget the specifics. Anyhow, @adngdb suggested doing this:

```
$ ./manage.py migrate auth
$ ./manage.py migrate
```

And that fixes it.

It's the last step in my socorro-zero steps:

https://github.com/willkg/socorro-zero/blob/master/docs/dev_vagrant.rst

I think you need to do that here, too.

adngdb:
What Will says. Otherwise, r+!

PR #3373, opened 2016-06-17
fixes bug 1280579 - Stop using product_info view in products implementation

peterbe:
My goal is to remove all use of the product_info view. One thing at a time. 
What I did was I just copied the definition of the view into the current query and did some juggling of column names.

### DIFF
--- a/socorro/external/postgresql/products.py
+++ b/socorro/external/postgresql/products.py
@@ -94,7 +94,7 @@ class ProductVersions(PostgreSQLBase):
                 pv.version_string AS version,
                 pv.build_date AS start_date,
                 pv.sunset_date AS end_date,
-                ((prc.throttle * (100)::numeric))::numeric(5,2) AS throttle,
+                ((prc.throttle * (100)::numeric))::REAL AS throttle,
                 pv.featured_version AS is_featured,
                 pv.build_type,
                 pv.has_builds,
@@ -117,6 +117,7 @@ class ProductVersions(PostgreSQLBase):
         """.format(sql_where)
         results = self.query(sql, sql_params).zipped()
 
+        print results[0]
         return {
             'hits': results,
             'total': len(results),
--- a/socorro/external/postgresql/products.py
+++ b/socorro/external/postgresql/products.py
@@ -52,16 +52,18 @@ class ProductVersions(PostgreSQLBase):
             if value is None:
                 continue
             param = {
-                'product': 'product_name',
+                'product': 'pv.product_name',
                 'version': 'version_string',
-                'featured': 'is_featured',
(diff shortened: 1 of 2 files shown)


willkg:
I looked through the `product_info` view definition and went through the changes in this PR and verified field names and tables. Everything looks ok to me.

Are there tests that use `ProductVersions` that would balk if things weren't right?

peterbe:
Now you make me nervous. The unit tests clearly tests the query and they continued to work. I'll have another play using the website.

peterbe:
Wow. That was interesting. 
For one, I found a big glaring bug. Without properly converting the `NUMERIC` to `REAL` it broke the ability to convert to JSON in the webapp. Very subtle. Glad I poked around. 

Then I ran a bunch of queries using the web API and saved as variours .json files. 
Switched branch and ran the same queries again and started comparing the .json files. 
Turns out that 3 of the 6 different queries yielded different results. Weird. 

So I started digging into the difference. After a lot of scripting and poking around, the only difference is the sort order. Really weird. Weird because the sort on the SQL is well defined and quite easy to reason with. Also, every time I looked at the .json files to compare it never made sense. 
It was not until I started comparing ALL versions. Especially all OLD versions. 

For example, when querying for just `?product=Firefox` and nothing else, it *will* include version 7.0a1 and 7.0a2. It will also include version 7.0b1, 7.0b2, 7.0b3, etc. 
Now if you look closely, the sort on the version number is not done by a string but by a "binary" value which is the "version_sort" column. 

Turns out, the version_sort for "7.0a1" and "7.0a2" is the same!!
```
breakpad=> select version_string, version_sort from product_versions where version_string like '7.0a%' and product_name='Firefox';
 version_string | version_sort
----------------+---------------
 7.0a1          | 007000000b000
 7.0a2          | 007000000b000
(2 rows)
```
That's not the case for other versions later in life:
```
breakpad=> select version_string, version_sort from product_versions where version_string like '31.0a%' and product_name='Firefox';
 version_string | version_sort
----------------+---------------
 31.0a1         | 031000000a001
 31.0a2         | 031000000a002
(2 rows)
```

So all this work for some buggy implementation in how `version_sort` is generated back in 2011!

willkg:
That's pretty intense! Good sleuthing!

How do you feel about this code now? It still looks ok to me, but I don't know any of the code around it so I wouldn't know if it was wrong in the greater scheme of things.

peterbe:
*Now* I feel very confident about the code. Also, we have the stage server to catch things. 
I'm glad you stopped me in my tracks a bit. I'm happy. If you're happy, please merge.

willkg:
Sounds good! r+

PR #3374, opened 2016-06-21
Bug 1274345 - Added more DLLs to the siglists.

adngdb:


willkg:
Jeff said this looked good to him in the bug. Looks good to me, too, as far as I can tell. r+

PR #3375, opened 2016-06-21
fixes bug 1278937 - Implement downloading missing symbols CSV

peterbe:


peterbe:
Actually. The more I look at this the more I wonder if I should re-write it. 
The new implementation class is only ever used by the crontabber app. It's never used by the webapp. 
So, the code (the SQL part and iterating over a cursor) could ALL sit inside the crontabber app. That would mean only two files need to change instead of 4.

peterbe:
At the moment, one crontabber job introduces 4 files. 1 for getting the data, one for testing getting the data, 1 for using the data and 1 for testing using the data. 
Reminds me of that blog post about separating the IO from the algorithm.

peterbe:
Or, perhaps I just don't bother re-writing it. It seems to work. And I'm not sure how important it is to spend more time on it.

socorro/external/postgresql/missing_symbols.py
```diff
@@ -0,0 +1,71 @@
...
+        WHERE
+            date_processed = %(date)s AND
+            debug_file != '' AND
+            debug_id != ''
+        """
+        if params['limit'] is not None:
+            sql += '\nLIMIT %(limit)s'
+        return sql, params
+
+    def get(self, **kwargs):
```
adngdb: As far as I can tell this method is never used. Why keep it around?

socorro/cron/jobs/missingsymbols.py
```diff
@@ -0,0 +1,115 @@
...
+        key_object = bucket.new_key('latest.csv')
+        key_object.set_contents_from_string(buf.getvalue())
+        self.config.logger.info(
+            'Generated ' +
+            key_object.generate_url(expires_in=0, query_auth=False) +
+            ' ({} bytes, {:.2f} Mb)'.format(
+                format(key_object.size, ','),
+                key_object.size / 1024.0 / 1024.0
+            )
+        )
```
adngdb:
```python
self.config.logger.info(
    'Generated {} ({} bytes, {:.2f} Mb)'.format(
        key_object.generate_url(expires_in=0, query_auth=False),
        format(key_object.size, ','),
        key_object.size / 1024.0 / 1024.0
    )
)
```

This seems more readable. I struggled to understand that logger. :)

socorro/unittest/cron/jobs/test_missingsymbols.py
```diff
@@ -0,0 +1,156 @@
...
+            tab.run_all()
+
+        # Basic assertion test of stored procedure.
+        information = self._load_structure()
+        assert information['missing-symbols']
+        assert not information['missing-symbols']['last_error']
+        assert information['missing-symbols']['last_success']
+
+        self.mock_boto_class()._connect.assert_called_with()
+        self.mock_boto_class.close.assert_called_with()
```
adngdb: Why is there no parenthesis here after ``mock_boto_class``? My "mockery" is rusty but I feel like the first line, with parenthesis, doesn't test anything.

socorro/external/postgresql/missing_symbols.py
peterbe: good point!

socorro/cron/jobs/missingsymbols.py
peterbe: I agree. It grew over time and I didn't see that I had to refactor it back to sanity.

socorro/unittest/cron/jobs/test_missingsymbols.py
peterbe: Good point. I just looked at `print self.mock_boto_class.mock_calls` and wrote the assert statements accordingly. I'll take another look.

adngdb:
You'll want to remove the tests for the ``get`` method you removed, too.

socorro/unittest/cron/jobs/test_missingsymbols.py
adngdb: One good way to verify a test is working is to see if it can fail. Can you make that precise line fail in any way?

PR #3377, opened 2016-06-22
make it not a tuple

peterbe:
sigh


peterbe:
Tests didn't notice because the actual value was overwritten in tests so the default didn't kick in.

willkg:
Looks good! r+

PR #3379, opened 2016-06-27
Fixes bug 1281161 - Added server-side sorting to Super Search and Signature report.

adngdb:


webapp-django/crashstats/signature/static/signature/js/signature_tab_reports.js
```diff
@@ -97,11 +103,20 @@ SignatureReport.ReportsTab.prototype.getParamsForUrl = function () {
...
 
+    // Get the sort for the input.
+    var sortArr = this.$sortInputHidden.val().split(',');
+    if (sortArr.length === 1 && !sortArr[0]) {
+        sortArr = [];
+    }
+
+    if (sortArr) {
+        params._sort = sortArr;
+    }
```
peterbe: All of the above could just be:
```javascript
if (this.$sortInputHidden.val().trim()) {
    params._sort = this.$sortInputHidden.val().trim().split(',');
}
```

webapp-django/crashstats/signature/static/signature/js/signature_tab_reports.js
```diff
@@ -115,5 +130,47 @@ SignatureReport.ReportsTab.prototype.buildUrl = function (params) {
 
     // Return the whole URL.
     return this.dataUrl + queryString;
+};
+
+SignatureReport.ReportsTab.prototype.onAjaxSuccess = function (contentElement, data) {
+    SignatureReport.Tab.prototype.onAjaxSuccess.apply(this, arguments);
+    var that = this;
```
peterbe: Not super helpful :)
Perhaps `var tab = this;`?

webapp-django/crashstats/signature/static/signature/js/signature_tab_reports.js
```diff
@@ -115,5 +130,47 @@ SignatureReport.ReportsTab.prototype.buildUrl = function (params) {
...
+
+    $('.sort-header', contentElement).each(function() {
+        $(this).click(function(event) {
```
peterbe: Why not just `$('.sort-header', contentElement).on('click', function(event) {`?
Now you have to create an event handler inside each loop of `.each()`.

webapp-django/crashstats/signature/static/signature/js/signature_tab_reports.js
```diff
@@ -115,5 +130,47 @@ SignatureReport.ReportsTab.prototype.buildUrl = function (params) {
...
+            event.preventDefault();
+
+            var thisElt = $(this);
+
+            function removeFromArray(elt, arr) {
```
peterbe: What does eslint or jslint say about this? It's a function defined inside an iterator. It ought to be moved out to a more "global" scope.

webapp-django/crashstats/signature/static/signature/js/signature_tab_reports.js
```diff
@@ -115,5 +130,47 @@ SignatureReport.ReportsTab.prototype.buildUrl = function (params) {
...
+                var index = arr.indexOf(elt);
+                if (index > -1) {
+                    arr.splice(index, 1);
+                }
+                return arr;
```
peterbe: A much simpler version would be:
```
function removeFromArray(elt, arr) {
    return arr.filter(function(each) { 
        return each !== elt;
    });
}
```

peterbe:
I haven't yet fully understood why but here's a case of it *not* working. 
1) Go to SuperSearch
2) Add "version", "has terms", "50.0a1"
3) Press search
4) Change to the "Crash Reports" tab
5) Note that it's already sorted by date. Most recent crash first. There's a little down arrow to indicate this. 
6) Click the Date column header. The page reloads. 

The page reloads but the Crash Reports don't change. It's still sorted by Date most recent first. 

In other words, I can't make it show me the oldest crash first.

peterbe:
Actually, I **can** see the oldest crash first. I just have to first click to sort by, for example, "Product" and *then* click the Date column. Non-ascending defaults are hard.

peterbe:
Another strange thing. I click on "Date" or "Product" and the tab reloads. But when I click on "CrashID" it does a in-javascript client sort.

peterbe:
I'm not sure how I did this but now it's sorted by Date AND Signature. 
screen shot 2016-06-27 at 2 10 30 pm

webapp-django/crashstats/signature/views.py
```diff
@@ -127,16 +135,27 @@ def signature_reports(request, params):
 
     context['params'] = current_query.copy()
 
+    if '_sort' in context['params']:
+        del context['params']['_sort']
+
     if '_columns' in context['params']:
         del context['params']['_columns']
 
+    context['sort'] = request.GET.getlist('_sort') or DEFAULT_SORT
```
peterbe: In `pass_validated_params` you validate this list. (by making sure it's in `fields`). Here you don't. I don't think that's an inroad to XSS because the `context['sort']` in this case is never printed in HTML. 
But I don't know what craziness a hacker could trigger by sorting on something we don't first check to be a valid field. [For example](http://socorro.dev/search/results/?product=Firefox&version=50.0a1&_sort=CRAZYCRAZYCRAZY&_sort=date&_sort=product&_facets=signature&_columns=date&_columns=signature&_columns=product&_columns=version&_columns=build_id&_columns=platform).

webapp-django/crashstats/signature/views.py
```diff
@@ -127,16 +135,27 @@ def signature_reports(request, params):
...
     context['columns'] = request.GET.getlist('_columns') or DEFAULT_COLUMNS
 
-    # Make sure only allowed fields are used
+    # Make sure only allowed fields are used.
+    context['sort'] = [
+        x for x in context['sort']
+        if x in allowed_fields or
+        (x.startswith('-') and x[1:] in allowed_fields)
+    ]
```
peterbe: arhg! My comment 10 lines above was wrong. You do check the sort value later. But what happens if I pass `?_sort=JUNK`. Then it doesn't get a chance to set the `context['sort']` to be `DEFAULT_SORT`.

webapp-django/crashstats/supersearch/static/supersearch/js/socorro/search.js
```diff
@@ -67,6 +68,44 @@ $(function () {
                 });
                 $('.tablesorter').tablesorter();
 
+                $('.sort-header', contentElt).each(function() {
+                    $(this).click(function(event) {
```
peterbe: Here's the same code as in the other block of JS. It sets an individual event handler *inside* a loop.

webapp-django/crashstats/supersearch/static/supersearch/js/socorro/search.js
```diff
@@ -67,6 +68,44 @@ $(function () {
...
+
+                        var thisElt = $(this);
+
+                        function removeFromArray(elt, arr) {
+                            var index = arr.indexOf(elt);
+                            if (index > -1) {
+                                arr.splice(index, 1);
+                            }
+                            return arr;
+                        }
```
peterbe: Ditto what I mentioned last time :)

webapp-django/crashstats/supersearch/static/supersearch/js/socorro/search.js
```diff
@@ -107,6 +146,14 @@ $(function () {
         var i;
         var len;
 
+        var sortArr = sortInput.select2('data');
+        if (sortArr) {
+            params._sort = [];
+            for (i = 0, len = sortArr.length; i < len; i++) {
+                params._sort[i] = sortArr[i].id;
+            }
```
peterbe: Or just `params._sort = sortArr.map(function(x) { return x.id });`

webapp-django/crashstats/supersearch/views.py
```diff
@@ -89,14 +96,29 @@ def get_params(request):
...
+    ]
     params['_facets'] = [
         x for x in params['_facets'] if x in allowed_fields
     ]
+    params['_columns'] = [
+        x for x in params['_columns'] if x in allowed_fields
+    ]
+
+    # The uuid is always displayed in the UI so we need to make sure it is
+    # always returned by the model.
```
peterbe: Why?

peterbe:
I struggled to understand all the code. There seemed to be quite a lot of duplication. For example, that code that converts `request.GET.getlist('_sort')` to a list with or with items starting with a `-`. That happens 3 times I think. 

Perhaps it doesn't matter since these features aren't going to last more than a year :)

I'm happy to take a closer look when you've fixed it so that it works.

peterbe:
What would be really cool is to disable all the fancy AJAX-sorting-reloading if there is only 1 page in the table.

webapp-django/crashstats/signature/static/signature/js/signature_tab_reports.js
adngdb: Heh, now that you say it that sounds obvious. I probably copied this from somewhere else without thinking...

webapp-django/crashstats/signature/views.py
adngdb: It's fine if there is not sorting done. That's what happens at the moment anyway.

webapp-django/crashstats/supersearch/views.py
adngdb: I'm not sure I understand your concern here. I find it hard to make that comment clearer. What exactly is not clear to you?

peterbe:
As discussed, let's make it so that clicking on a header changes that header to the best first sort.

adngdb:
@peterbe I believe all issues are addressed now!

peterbe:
Sorting by date doesn't work. 
1) Make a search for Firefox, version 50.0a1
2) Note that the default sort is Date with most recent first (aka. descending order)
3) Note that the Date column header has the down arrow. 
4) Click the date column. 
5) Note that on the Date column header, it's no longer a down arrow. Now it's an up & down. 
6) URL became: `http://socorro.dev/search/?product=Firefox&version=50.0a1&_facets=signature&_columns=date&_columns=signature&_columns=product&_columns=version&_columns=build_id&_columns=platform#crash-reports`

I think what happens is that if you clicked the Date header, it removes any sort order.

peterbe:
Clicking the Date header 1 more time, then it sorts by date again, in descending order.

peterbe:
I think we need to reconsider how the query string works entirely. Sorry. 
If I give you a URL like this: `http://socorro.dev/search/?product=Firefox&version=50.0a1&_sort=-platform&_sort=signature&_sort=date&_facets=signature&_columns=date&_columns=signature&_columns=product&_columns=version&_columns=build_id&_columns=platform#crash-reports` 
How should it be sorted? There's no order. 
You can never trust that `_sort=-platform` comes *before* `_sort=signature` in the query string. 

What we need is something like this: `_sort=-platform,signature,-date`

webapp-django/crashstats/signature/static/signature/js/signature_tab_reports.js
```diff
@@ -104,14 +104,7 @@ SignatureReport.ReportsTab.prototype.getParamsForUrl = function () {
...
     // Get the sort for the input.
-    var sortArr = this.$sortInputHidden.val().split(',');
-    if (sortArr.length === 1 && !sortArr[0]) {
-        sortArr = [];
-    }
-
-    if (sortArr) {
-        params._sort = sortArr;
-    }
+    params._sort = this.$sortInputHidden.val().trim().split(',');
```
peterbe: This is different from the code you had before. If the `$sortInputHidden` is empty, the `params._sort` will become `['']`.

webapp-django/crashstats/signature/static/signature/js/signature_tab_reports.js
```diff
@@ -133,44 +126,44 @@ SignatureReport.ReportsTab.prototype.buildUrl = function (params) {
...
-    var that = this;
+    var tab = this;
 
-    $('.sort-header', contentElement).each(function() {
-        $(this).click(function(event) {
-            event.preventDefault();
+    contentElement.empty().append($(data));
+    $('.tablesorter').tablesorter({
+        headers: {
+            0: {  // disable the first column, `Crash ID`
```
peterbe: ...or `Rank` in the facet tables.

peterbe:
I have now clicked on 6 different column headers to sort. None of the headers (except Crash ID) have either an up or a down arrow. But I have no idea which one it sorts by. Perhaps we can display what the sort order is in text. Or put a number near the down/up arrow. Bugzilla "solves" this by using different shades of gray.

adngdb:
> I think what happens is that if you clicked the Date header, it removes any sort order. 
> Clicking the Date header 1 more time, then it sorts by date again, in descending order. 

This is expected behavior. The sorting cycle when you click is this: none -> asc -> desc -> none. Date is by default desc, so when you click the header, the next step in that cycle is "none". If this is too confusing, I can remove the default sorting.

peterbe:
Hmm... you're not wrong or right. And neither am I. I think I let my gut speak. 
I'm so used to that when I click on a header that it changes the direction if that's already the header that is used for sorting. That's how things work in apps like Finder for example. You can't unselect sort. 
Unselecting sort might make s

peterbe:
Hmm... you're not wrong or right. And neither am I. I think I let my gut speak. 
I'm so used to that when I click on a header that it changes the direction if that's already the header that is used for sorting. That's how things work in apps like Finder for example. You can't unselect sort. 
Unselecting sort might make sense if you have multiple sort headers.

webapp-django/crashstats/signature/static/signature/js/signature_tab_reports.js
```diff
@@ -97,11 +103,13 @@ SignatureReport.ReportsTab.prototype.getParamsForUrl = function () {
         });
     }
 
+    // Get the sort for the input.
+    params._sort = this.$sortInputHidden.val().trim().split(',') || [];
```
peterbe: That `|| []` is never going to be needed. `['']` or `['one','two']` is always true.

webapp-django/crashstats/supersearch/views.py
peterbe: What I meant was that it never says why we need the 'uuid'. You're just saying that you're making sure it's returned, which is what the code does.

webapp-django/crashstats/signature/static/signature/js/signature_tab_reports.js
adngdb: That is incorrect. In JavaScript, ``[''] == false``. (Not in Python, though. )

webapp-django/crashstats/supersearch/views.py
adngdb: Well "the uuid is always displayed in the UI" is the real explanation of it. I guess it could be slightly more precise by saying "the uuid is always displayed in the results in the UI", but I hope you don't mind if I don't bother. :)

webapp-django/crashstats/signature/static/signature/js/signature_tab_reports.js
peterbe: Now I'm confused. 
if `this.$sortInputHidden.val().trim().split(',')` is exactly `['']` the end result is that `params._sort` will be `['']`. 
In node...
```
> [''] || []
[ '' ]
```
If you want `params._sort` to be `[]` when the `.val().trim()` is nothing, you need a different solution.

peterbe: I.e.
```javascript
> input = ''
''
> sort = input.trim().split(',') || []
[ '' ]
> sort
[ '' ]
```

Perhaps this is what you wanted:
```javascript
> input.trim() && input.trim().split(',') || []
[]
> input = 'one,two'
'one,two'
> input.trim() && input.trim().split(',') || []
[ 'one', 'two' ]
```

webapp-django/crashstats/supersearch/views.py
peterbe: Ah! My "Why?" was about "WHY is it always displayed in the UI?"
But now when I see it all again, it's clear that that's not for this code to understand or have opinions on what the UI does.

PR #3381, opened 2016-06-29
fixes bug 1283214 - Check SuperSearch in healthcheck monitoring

peterbe:


webapp-django/crashstats/monitoring/tests/test_views.py
```diff
@@ -278,3 +280,42 @@ def fake_es_instance(**config):
         )
 
         eq_(es_instance.info.call_count, 2)
+
+    def test_assert_supersearch_counts(self):
+
+        searches = []
```
adngdb: You don't use the content of ``searches`` anywhere, might as well make it a counter?

webapp-django/crashstats/monitoring/tests/test_views.py
```diff
@@ -278,3 +280,42 @@ def fake_es_instance(**config):
...
+
+        def mocked_supersearch_get(**params):
+            searches.append(params)
+            eq_(params['product'], [settings.DEFAULT_PRODUCT])
+            # print params
```
adngdb: debug leftover

webapp-django/crashstats/monitoring/views.py
```diff
@@ -167,4 +168,56 @@ def healthcheck(request):
...
+    This test has been tested in our production database, using the
+    /api/SuperSearch endpoint and the epsilon hovers around 0.005 roughly
+    and that's *with* webapp caching at play.
+    """
+    supersearch = SuperSearch()
+    # We don't want any caching this time
+    supersearch.cache_seconds = 0
+    results = supersearch.get(
+        product=settings.DEFAULT_PRODUCT,
+        _results_number=1,
```
adngdb: Why not ``0``?

webapp-django/crashstats/monitoring/tests/test_views.py
```diff
@@ -278,3 +280,42 @@ def fake_es_instance(**config):
...
+                        {'uuid': '3969175'},
+                    ],
+                    'facets': [],
+                    'total': 30004
+                }
+
+        SuperSearch.implementation().get.side_effect = (
+            mocked_supersearch_get
+        )
+        assert_supersearch_counts()
```
adngdb: You don't test the output of this function. Why not?

adngdb: My bad, I read ``return`` instead of ``assert`` in the function. Still, I would have expected to see a test for when the check fails.

adngdb:
A bunch of nits, but it's looking good overall. Most importantly, I think we want to have a test for the failure case, since it's quite difficult to reproduce locally.

PR #3390, opened 2016-07-15
bug 1283285 - Google Sign-in

peterbe:
CC @willkg 

I've really tried to make it small but it's still pretty big. A lot of it is just tests on django views. 

Basically, what this does is that it introduces a **second** log in button next to the Persona button. If you successfully sign in you, all the Persona stuff will disappear. There's a dependent bug about later removing Persona entirely. I'm just trying to do one thing at a time. 

**IMPORTANT** To be able to test it locally you need to go to the [API Manager](https://console.developers.google.com/projectselector/apis/library), (sign in with your Google Mozilla LDAP thing) and click on the "Socorro Crash-stats" project. 
Click "Crash-stats" to see the credentials (note the email I sent a few minutes ago with subject line "Google APIs"). 
As you can see I've prepared 3 URLs (3 for Origin and 3 for redirect URI). Locally I use `http://socorro.dev` to do dev work. If you use something different, add that. I suspect you'll be simply able to add `http://localhost:8000` if that's what you use.


peterbe:
I should explain how it works:

The Google Sign-in is a "layer" on top of OAuth2. It's a client-side solution, but we're making it work on the server-side too. So, the user uses the Google API (only the auth API) to get a token. Once the token is received, in javascript, you have to use it together with the secret. So, in my AJAX code I take that token and send it to the server. The server then uses the libraries from `google-api-python-client` (token + our secret) to check that the user exists. If so, you get an email. Once you have an email you forcibly log the user in to Django. Now you're signed in on the server-side (eg. `request.user.is_authenticated()`).

webapp-django/crashstats/authentication/views.py
```diff
@@ -1,16 +1,36 @@
...
 from crashstats.crashstats.utils import json_view
 
 
+User = get_user_model()
+
+
+def default_username(email):
+    # Store the username as a base64 encoded sha1 of the email address
+    # this protects against data leakage because usernames are often
+    # treated as public identifiers (so we can't use the email address).
```
adngdb: :+1:

webapp-django/crashstats/authentication/views.py
```diff
@@ -49,3 +69,85 @@ def debug_login(request):
...
+            token,
+            settings.OAUTH2_CLIENT_ID
+        )
+    except crypt.AppIdentityError:
+        return http.HttpResponseForbidden(
+            'Invalid Identity token'
+        )
+
+    # We should probably look at idinfo['email_verified'] and
+    # return a Forbidden if it's not verified.
```
adngdb: What about this? Should we do it or not?

webapp-django/crashstats/authentication/views.py
```diff
@@ -49,3 +69,85 @@ def debug_login(request):
...
+            user.save()
+    except User.DoesNotExist:
+        user = User.objects.create(
+            username=default_username(idinfo['email']),
+            email=idinfo['email'],
+            first_name=idinfo.get('given_name', ''),
+            last_name=idinfo.get('family_name', ''),
+        )
+    user.backend = 'django.contrib.auth.backends.ModelBackend'
+    request.user = user
```
adngdb: Can you explain these last two lines? In particular, I don't understand why you want to set the user in ``request`` before calling ``auth.login(request, user)``.

webapp-django/crashstats/base/jinja2/crashstats_base.html
```diff
@@ -172,8 +177,22 @@
                     <a href="{{ url('manage:home') }}">Admin</a>
                     {% endif %}
                     |
-                    {{ browserid_logout(text='Sign out') }}
+                    {% if request.session.get('signin_method') %}
```
adngdb: I would use `` == 'google'`` here. Just because it makes it more explicit, and more resilient to future changes.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/oauth2.js
```diff
@@ -0,0 +1,134 @@
...
+        var data = {
+            csrfmiddlewaretoken: csrfmiddlewaretoken,
+        };
+        $.post(signoutURL, data)
+        .done(function() {
+            // XXX could this be smarter
+            // https://bugzilla.mozilla.org/show_bug.cgi?id=1283296
+            // For example, you might be on a sensitive URL
+            // so redirect will just become a redirect fest.
+            document.location = '/';
```
adngdb: I believe this is what we do at the moment with persona sign out as well, so this is fine by me. It's a lot of trouble to redirect somewhere different than home when signing out, for very low value.

adngdb: Plus no-one signs out anyway.

webapp-django/crashstats/authentication/tests/test_views.py
```diff
@@ -52,6 +58,175 @@ def test_successful_redirect(self):
...
+        # because you're not signed in any more
+        eq_(response.status_code, 302)
+
+    def test_oauth2_signin(self):
+
+        def mocked_verify_id_token(token, client_id):
+            return {
+                'aud': settings.OAUTH2_CLIENT_ID,
+                'iss': settings.OAUTH2_VALID_ISSUERS[0],
+                'email': 'test@example.com',
```
adngdb: That is the same email address as the one you use in ``_login``. Isn't there a risk of overlap here? I would use a different one to be really sure that a new user is being created.

adngdb:
Alright, code looks good, but I haven't tested yet. There's a few nits. I'll finish testing first thing tomorrow!

webapp-django/crashstats/authentication/tests/test_views.py
peterbe: "test@example.com" doesn't get created until you *call* `self._login()`.

webapp-django/crashstats/authentication/views.py
peterbe: Good point. Taken care of.

webapp-django/crashstats/base/jinja2/crashstats_base.html
peterbe: Good point. Easier to understand/debug that way.

adngdb:
There is a ``Sign out`` link in the profile page that you did not take care of. It always links to browserid logout.

adngdb:
This is great! r+ when the other "Sign out" link is taken care of. :)

adngdb:
Huh, there are conflicts with master.

webapp-django/crashstats/authentication/views.py
```diff
@@ -49,3 +69,87 @@ def debug_login(request):
...
+@json_view
+def oauth2_signout(request):
+    """If people click the "Sign out" link, an AJAX post will trigger
+    a post to this view. But if the user right-clicks to, for example,
+    "Open Link in New Tab" they'll arrive here.
+
+    If they arrive here, in any way, and are NOT signed in, redirect
+    out.
+    """
+    if not request.user.is_authenticated():
```
adngdb: Shouldn't this be ``is_active`` instead?

adngdb: I guess it doesn't matter. We don't really care about the user being active here.

webapp-django/crashstats/authentication/views.py
adngdb: Up?

adngdb:
r++

peterbe:
@adngdb I've rebased this up. Feel free to merge when happy.

PR #3391, opened 2016-07-18
Bug 1287548 migrate users by ldap alias email

peterbe:
cc @adngdb 

When we switch from Persona to Google Sign-in, people who used to sign in with `alias@example.com` will no longer be able to do so. They simply HAVE to use `flastname@example.com`. If they used to use `alias@...` their permissions etc would be lost. So we'll "merge" those and transfer all the goodies that `alias@...` had to `flastname@...`. 

The day we switch over from Persona to Google Sign-in we need to generate a CSV file that looks like this:
```
ALIAS,CORRECT
alias@example.com,flastname@example.com
alias2@example.com,flastname2@example.com
aliasN@example.com,flastnameN@example.com
```
Then we need to scp that file into any of the webheads and run this command:

```
$ envconsul -prefix socorro/common -prefix socorro/webapp-django \
./manage.py migrate-alias-users --dry-run test.csv
```
(If it works, remove the `--dry-run`)

The CSV file I'll generate using my Mozilla LDAP certificate. 

Once we've done this we can decommission all of this code.

### DIFF
--- a/webapp-django/crashstats/authentication/migration.py
+++ b/webapp-django/crashstats/authentication/migration.py
@@ -1,0 +1,53 @@
+from django.contrib.auth import get_user_model
+from django.db.models import Q
+
+User = get_user_model()
+
+
+def migrate_users(combos, dry_run=False):
+    # Because the list of comboes is possibly very large, we don't want
+    # to query the ORM for each and every one since most of the emails
+    # we expect to search for are not going to be found.
+    size = 100
+    groups = [
+        combos[i:i + size] for i in range(0, len(combos), size)
+    ]
+    for group in groups:
+        transform = dict((x.lower(), y) for x, y in group)
+
+        q = Q()
+        for email in transform:
+            q = q | Q(email__iexact=email)
+        # Most likely the list of emails we search for is MUCH larger
+        # than the number of users we find. For those few found,
+        # we'll deal with it one at a time.
+        users = User.objects.filter(q)
+        for user in users:
+            print 'NEED TO MIGRATE', user.email.ljust(30),
+            print 'TO', transform[user.email]
+            try:
+                destination = User.objects.get(
+                    email__iexact=transform[user.email]
+                )
+                if user.is_staff:
+                    print "\tTransferring 'is_staff'"
+                    destination.is_staff = True
+                if user.is_superuser:
+                    print "\tTransferring 'is_superuser'"
+                    destination.is_superuser = True
+                for group in user.groups.all():
+                    print '\tTransferring group membership %r' % group.name
+                    destination.groups.add(group)
+                user.is_active = False
+                if not dry_run:
+                    user.save()
+                    destination.save()
+            except User.DoesNotExist:
+                # then it's easy peasy!
+                print '\tSimply changing email from %r to %r' % (
+                    user.email,
+                    transform[user.email]
+                )
+                user.email = transform[user.email]
+                if not dry_run:
+                    user.save()
--- a/webapp-django/crashstats/authentication/migration.py
+++ b/webapp-django/crashstats/authentication/migration.py
@@ -56,8 +56,8 @@ def migrate_users(combos, dry_run=False):
                 # then it's easy peasy!
                 print '\tSimply changing email from %r to %r' % (
                     user.email,
-                    transform[user.email]
+                    transform[user.email.lower()]
                 )
-                user.email = transform[user.email]
+                user.email = transform[user.email.lower()]
                 if not dry_run:
                     user.save()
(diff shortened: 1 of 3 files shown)


peterbe:
**NOTE the last sentence. This code only needs to work once. And it will not be kept indefinitely.**

webapp-django/crashstats/authentication/migration.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def migrate_users(combos, dry_run=False):
+    # Because the list of comboes is possibly very large, we don't want
+    # to query the ORM for each and every one since most of the emails
+    # we expect to search for are not going to be found.
+    size = 100
+    groups = [
+        combos[i:i + size] for i in range(0, len(combos), size)
+    ]
```
willkg: I always found it puzzling that this isn't in the Python itertools module. Because of that, I keep writing it:

https://github.com/mozilla/fjord/blob/master/fjord/search/index.py#L249

Erik wrote one, too:

https://github.com/erikrose/more-itertools/blob/master/more_itertools/more.py#L15

Yours is fine, too. It just always puzzled me.

webapp-django/crashstats/authentication/migration.py
```diff
@@ -0,0 +1,63 @@
...
+        q = Q()
+        for email in transform:
+            q = q | Q(email__iexact=email)
+        # Most likely the list of emails we search for is MUCH larger
+        # than the number of users we find. For those few found,
+        # we'll deal with it one at a time.
+        users = User.objects.filter(q)
+        for user in users:
+            print 'NEED TO MIGRATE', user.email.ljust(30),
+            print 'TO', transform[user.email]
```
willkg: Unless we know for a fact that all the email addresses in the db are lowercased, you probably want to lowercase the email address to confidently match the transform dict keys.

webapp-django/crashstats/authentication/migration.py
```diff
@@ -0,0 +1,63 @@
...
+            try:
+                destination = User.objects.get(
+                    email__iexact=transform[user.email]
```
willkg: Might want to lowercase the email here, too.

webapp-django/crashstats/authentication/migration.py
```diff
@@ -0,0 +1,63 @@
...
+                if not dry_run:
+                    user.save()
+                    destination.save()
+            except User.DoesNotExist:
+                # then it's easy peasy!
+                print '\tSimply changing email from %r to %r' % (
+                    user.email,
+                    transform[user.email]
+                )
+                user.email = transform[user.email]
```
willkg: Might want to lowercase the email here, too.

webapp-django/crashstats/authentication/migration.py
```diff
@@ -0,0 +1,63 @@
...
+                    destination.date_joined = user.date_joined
+                user.is_active = False
+                if not dry_run:
+                    user.save()
+                    destination.save()
+            except User.DoesNotExist:
+                # then it's easy peasy!
+                print '\tSimply changing email from %r to %r' % (
+                    user.email,
+                    transform[user.email]
```
willkg: Might want to lowercase the email here, too.

willkg:
Can we assume the email addresses in the db are all lowercase? If not, then this needs fixing. Other than that, this looks good to me.

webapp-django/crashstats/authentication/migration.py
peterbe: It's `__iexact` so it's case-insensitive.

webapp-django/crashstats/authentication/migration.py
peterbe: The querying is done case-INsensitively.

willkg: This isn't a query--this is a dict key which is not case insensitive.

webapp-django/crashstats/authentication/migration.py
willkg: This is a dict key--so it's case sensitive. For clarification, I'm saying you should do this:

```
email__iexact=transform[user.email.lower()]
```

peterbe: Ohhh! Now I get it. Sorry. 
I modified the test so it broke then I added those `.lower()` to make it work.

webapp-django/crashstats/authentication/migration.py
peterbe: Yeah. Thanks.

webapp-django/crashstats/authentication/migration.py
peterbe: Yours is certainly prettier than mine.

webapp-django/crashstats/authentication/migration.py
willkg: I should have been clearer--sorry about that.

willkg:
This looks good to me. If Travis is happy, I'm happy. r+

willkg:
If Travis is cool with this, do you want me to merge it and squash the commits?

peterbe:
> If Travis is cool with this, do you want me to merge it and squash the commits?

Yes please.

PR #3392, opened 2016-07-19
Fixes bug 1282529 - Removed field name from truthy Elasticsearch operator.

adngdb:
@peterbe Don't ask me why this works, because I have no clue. The current code works for fields like ``accessibility`` but not for ``submitted_from_infobar``, and I do not have the slightest idea as to why. But, I have confirmed in my testing that removing this (unneeded and buggy) line fixed the problem. 

Of course, since I don't understand the problem, I have no idea how to reproduce it in unit tests. :(

### DIFF
--- a/socorro/external/es/supersearch.py
+++ b/socorro/external/es/supersearch.py
@@ -310,7 +310,6 @@ class SuperSearch(SearchBase):
                 elif param.operator == '__true__':
                     filter_type = 'term'
                     filter_value = True
-                    args['field'] = name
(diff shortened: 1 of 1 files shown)


peterbe:
How can it be "unneeded" and "buggy" at the same time?

adngdb:
It is unneeded because it is not required for this feature to work (it was a copy/pasting error) and it is buggy because it causes the bug we are seeing (though, as I said, I cannot explain why).

PR #3394, opened 2016-07-20
No bug - Fixed a display bug in the API reference of Super Search.

adngdb:


adngdb:
See the bug live here: https://crash-stats.mozilla.com/documentation/supersearch/api/#section-list%20of%20fields

PR #3395, opened 2016-07-20
Fixes bug 1273538 - Fixes white spaces encoding in pagination URLs.

adngdb:


webapp-django/crashstats/base/templatetags/jinja_helpers.py
```diff
@@ -7,6 +7,8 @@
 from django.contrib.staticfiles.storage import staticfiles_storage
 from django.core.urlresolvers import reverse
 
+from crashstats.base.utils import urlencode_dict
```
peterbe: 👍

webapp-django/crashstats/signature/views.py
```diff
@@ -195,7 +193,7 @@ def signature_reports(request, params):
 
     context['current_url'] = '%s?%s' % (
         reverse('signature:signature_report'),
-        current_query.urlencode()
+        current_query.urlencode().replace('+', '%20')
```
peterbe: Perhaps consider doing `urlencode_dict(current_query)` and inside `urlencode_dict` you do a 
```python
if hasattr(thing, 'urlencode'):
    thing = urllib.urlencode()
else:
    thing = urllib.urlencode(thing, True)
# comment here WHY we have to do this
return thing.replace('+', '%20')
```
that way, the hacky `+` replace thing is only in one place.

peterbe:
It's tempting to merge because it DOES solve the problem but I don't like that you've sprinkled the "hacky solution" in 3 places.

adngdb:
> It's tempting to merge because it DOES solve the problem but I don't like that you've sprinkled the "hacky solution" in 3 places. 

+1

peterbe:
Sorry for being a butt but could you do one last thing? 
Can you rename the function? `urlencode_dict` sounds like it only takes Python dict objects. 
How about simply `urlencode`?

adngdb:
I thought about that, but I find ``urlencode`` even more confusing. Maybe ``urlencode_obj`` or ``urlencode_object``? What do you think?

peterbe:
Or `encode_url`? Actually, this is up to you now. You decide. I'll approve whatever you prefer. Just wanted to "make you stop and think" :)

peterbe:
r+

adngdb:
I have to go, feel free to merge when it's green.

PR #3398, opened 2016-07-25
fixes bug 1289196 - observe user.is_active

peterbe:
I know this one is huge but it's actually really simple. Lemme explain:

* `request.user.is_active` is better than `request.user.is_authenticated` when you want to decide if the user should be given something. If an anonymous user makes a request, django automatically sets `request.user = AnonymousUser()`. That user is [always `is_active=False`](https://github.com/django/django/blob/master/django/contrib/auth/models.py#L386) so by using `request.user.is_active` you're asking two questions at the same time. "Is the user  authenticated?" (I.e. does the user have a session cookie) AND "Is the user active?". 

* Some places we were using `django.contrib.auth.decorators.login_required` and that one relies only on `request.user.is_authenticated()`. I made a new one, with the only difference that it ALSO checks that the user is active. 

* A bunch of small extensions of existing view tests to see what would happen if the user ceases to be active AFTER they have logged in. 

Our site is difference because we might have an interest in centrally, ceasing someones access AFTER they have successfully signed in. E.g. ceasing to be a Mozilla LDAP staff member.

### DIFF
--- a/webapp-django/crashstats/crashstats/tests/test_views.py
+++ b/webapp-django/crashstats/crashstats/tests/test_views.py
@@ -7,13 +7,13 @@ import json
 import urllib
 import random
 import urlparse
+from cStringIO import StringIO
 
 import pyquery
 import mock
-
-from cStringIO import StringIO
 from nose.tools import eq_, ok_, assert_raises
 from nose.plugins.skip import SkipTest
+
 from django.test.client import RequestFactory
 from django.test.utils import override_settings
 from django.conf import settings
@@ -1157,6 +1157,15 @@ class TestViews(BaseTestViews):
         # if you try to mess with the paginator it should just load page 1
         response = self.client.get(url, {'page': 'meow'})
         eq_(response.status_code, 200)
+
+        # If you cease to be active, access should be revoked automatically
+        user = User.objects.get(id=user.id)
+        user.is_active = False
+        user.save()
+        assert not user.has_perm('crashstats.view_exploitability')
+        response = self.client.get(url)
+        ok_(settings.LOGIN_URL in response['Location'] + '?next=%s' % url)
+        eq_(response.status_code, 302)
 
     @mock.patch('crashstats.crashstats.models.Bugs.get')
     @mock.patch('requests.get')
@@ -2441,6 +2450,14 @@ class TestViews(BaseTestViews):
         # Ensure fields have their description in title.
         ok_('No description for this field.' in response.content)
         ok_('Description of the signature field' in response.content)
+
+        # If the user ceases to be active, these PII fields should disappear
+        User.objects.filter(id=user.id).update(is_active=False)
+        response = self.client.get(url)
+        eq_(response.status_code, 200)
+        ok_('peterbe@mozilla.com' not in response.content)
+        ok_(_SAMPLE_META['Email'] not in response.content)
+        ok_(_SAMPLE_META['URL'] not in response.content)
 
     @mock.patch('crashstats.crashstats.models.Bugs.get')
     @mock.patch('requests.get')
--- a/webapp-django/crashstats/crashstats/tests/test_views.py
+++ b/webapp-django/crashstats/crashstats/tests/test_views.py
@@ -464,10 +464,10 @@ class BaseTestViews(DjangoTestCase):
         # once like they are in unit test running.
         SocorroCommon.clear_implementations_cache()
 
-    def _login(self):
-        user = User.objects.create_user('test', 'test@mozilla.com', 'secret')
-        assert self.client.login(username='test', password='secret')
-        return user
+    def _login(self, username='test', password='s', email='test@mozilla.com'):
+        User.objects.create_user(username, email, password)
+        assert self.client.login(username=username, password=password)
+        return User.objects.get(username=username)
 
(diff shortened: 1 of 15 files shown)


webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -1192,6 +1192,15 @@ def mocked_get(url, params, **options):
         response = self.client.get(url, {'page': 'meow'})
         eq_(response.status_code, 200)
 
+        # If you cease to be active, access should be revoked automatically
+        user = User.objects.get(id=user.id)
+        user.is_active = False
+        user.save()
+        assert not user.has_perm('crashstats.view_exploitability')
+        response = self.client.get(url)
+        ok_(settings.LOGIN_URL in response['Location'] + '?next=%s' % url)
```
adngdb: So that means, if a user is made inactive, they will be forever redirected to the login page? What happens when an inactive user tries to log in?

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -2442,6 +2451,14 @@ def mocked_get(url, params, **options):
         ok_('No description for this field.' in response.content)
         ok_('Description of the signature field' in response.content)
 
+        # If the user ceases to be active, these PII fields should disappear
+        User.objects.filter(id=user.id).update(is_active=False)
+        response = self.client.get(url)
+        eq_(response.status_code, 200)
+        ok_('peterbe@mozilla.com' not in response.content)
```
adngdb: You shouldn't put real email addresses in code. Why not use an ``@example.org`` fake one? That's what they're for.

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -1192,6 +1192,15 @@ def mocked_get(url, params, **options):
         response = self.client.get(url, {'page': 'meow'})
         eq_(response.status_code, 200)
 
+        # If you cease to be active, access should be revoked automatically
+        user = User.objects.get(id=user.id)
+        user.is_active = False
+        user.save()
```
adngdb: Why not use ``User.objects.filter(id=self.user.id).update(is_active=False)`` here as well?

webapp-django/crashstats/symbols/tests/test_views.py
```diff
@@ -166,6 +161,15 @@ def test_home(self):
             ok_(link['url'] in response.content)
             ok_(link['label'] in response.content)
 
+        # The access should disappar if you cease to be active
```
adngdb: ``disappear``

webapp-django/crashstats/symbols/tests/test_views.py
```diff
@@ -166,6 +161,15 @@ def test_home(self):
...
+        User.objects.filter(id=user.id).update(is_active=False)
```
adngdb: Can you not do ``user.update(is_active=False)`` instead? You already have the ``user`` object, why do you need to get it again?

adngdb:
If you sign in while being inactive, you do not get signed in, but you have no clue about what happened. That looks a lot like a bug. We should at least have a warning like the one we have for when you don't have superuser access and you try to hit the admin page.

webapp-django/crashstats/crashstats/tests/test_views.py
peterbe: They can't. They're locked out. The reason why we would ever bother to make a user inactive in the admin is because we don't want that user to be able to sign in. It could be because they have stopped being member of staff or because they've abused their access or something.

webapp-django/crashstats/crashstats/tests/test_views.py
peterbe: Good point. It was already there though.

webapp-django/crashstats/symbols/tests/test_views.py
peterbe: Glad you asked. The reason is because the `_login` method works like this:
```python
user = User.objects.create_user(...)
# at this point, the user.last_login is None
self.client.login(username=user.username, ...)
# it's now the same user from above. The self.client.login() updates the user's last_login
return user
```
So even though the `_login` method does both a user creation AND a login, the user it returns is one that doesn't have a `last_login` set.

peterbe: I think what I'm going to do instead is to re-write the `_login` method so that it returns a user **with** a `last_login` set.

peterbe:
@adngdb Pushed in another fix. Now, the `self._login()` method is improved so that it **returns** a user that has a `.last_login`. 
Also, when I was in there I noticed that the `manage/tests/test_views.py` had its own `def _login(...)` so I refactored that.

adngdb:
I still think we need to show something to inactive user when they sign in. Maybe not here nor in the other PR, but there should at least be a bug about it. It is way too confusing to see nothing happen after you have finished logging in. 

Other than that, r+.

peterbe:
@adngdb I think it handles the error handling quite nicely for inactive users. 
screen shot 2016-08-03 at 12 10 36 pm
It's not amazing but the use case rare and it works.

PR #3399, opened 2016-07-25
fixes bug 1283214 - healthcheck monitor shards too

peterbe:
I'm not sure I know what I'm doing in there but is it a good start?

### DIFF
--- a/socorro/external/es/supersearch.py
+++ b/socorro/external/es/supersearch.py
@@ -461,7 +461,7 @@ class SuperSearch(SearchBase):
 
         # Query and compute results.
         hits = []
-        shards = []
+        shards = None
 
         if params['_return_query'][0].value[0]:
             # Return only the JSON query that would be sent to elasticsearch.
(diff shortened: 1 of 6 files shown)


peterbe:
@adngdb I can't figure out why this test fails. It's `nosetests socorro/unittest/external/es/test_supersearch.py:IntegrationTestSuperSearch.test_get_against_nonexistent_index  -s -x`
When I print the `params` variable around line 517 in `supersearch.py` I see that `params['_return_shards']` becomes a list instead of a `socorro.middleware.search_common.SearchParam` instance. Any idea why?

adngdb:
I'm working on something very similar, but slightly different. I'd suggest you put this on hold for now, and we'll revisit when I'm done. The things I need will be compatible with your needs, but different enough that it will conflict.

peterbe:
oops. I closed the wrong one. 

@adngdb Don't forget to rebase this

PR #3400, opened 2016-07-26
Bug 1277301 - Capture and expose failing shards errors in Super Search.

adngdb:
@peterbe this is step 1. It will unblock #3399 and will allow me to give better explanations of what's wrong to our users on the web site.


socorro/external/es/supersearch.py
```diff
@@ -505,12 +523,30 @@ def get(self, **kwargs):
...
+                failed_indices[failure.index] += 1
+
+            for index, shards_count in failed_indices.items():
+                errors.append({
+                    'type': 'shards',
+                    'index': index,
+                    'shards_count': shards_count,
+                })
+
+            # TODO: send an error to sentry?
```
peterbe: Nope. We only want to send to sentry when there's a exception "frame". E.g. something in `sys.exc_info()`.

socorro/external/es/supersearch.py
```diff
@@ -481,7 +482,19 @@ def get(self, **kwargs):
                     hits.append(self.format_fields(hit.to_dict()))
 
                 total = search.count()
-                aggregations = self.format_aggregations(results.aggregations)
+
+                try:
+                    aggregations = self.format_aggregations(
+                        results.aggregations
+                    )
+                except AttributeError:
```
peterbe: Is this AttributeError for `results.aggregations`? Right now this gets used to capture that AND any potential `AttributeError` inside `self.format_aggregations`. I think it should be like this:

```python
try:
    aggregations = results.aggregations
except AttributeError:
    aggregations = None
if aggregations:
    aggregations = self.format_aggregations(aggregations)
```

peterbe: Actually, it's probably more Pythonic to do:
```python
aggregations = getattr(results, 'aggregations', None)
```

socorro/external/es/supersearch.py
```diff
@@ -481,7 +482,19 @@ def get(self, **kwargs):
...
+                    aggregations = {}
+
+                try:
+                    shards = results._shards
+                except AttributeError:
+                    shards = {}
```
peterbe: These four lines are probably nicer with just `shards = getattr(results, '_shards', {})`

socorro/unittest/external/es/test_supersearch.py
```diff
@@ -1781,3 +1783,109 @@ def test_get_with_too_many(self):
...
+            },
+            {
+                "status": 500,
+                "index": "other_index",
+                "reason": "foo bar gone bad",
+                "shard": 1
+            }
+        ]
+    }
+}'''
```
peterbe: I prefer that we write python dicts instead. Then you can do
```python
mock_requests.get(
    ...,
    text=json.dumps(es_results)
)
```
When it's python dicts you get syntax highlighting, you can put commas at the last item of a list, you can easily inject interesting values an you immediately get a `SyntaxError` instead of some later cryptic hard-to-debug JSON parse error.

socorro/external/es/supersearch.py
```diff
@@ -493,6 +506,11 @@ def get(self, **kwargs):
                     # it does, better know it.
                     raise
 
+                errors.append({
+                    'type': 'missing_index',
+                    'index': missing_index,
+                })
+
```
peterbe: Nice!

socorro/external/es/supersearch.py
```diff
@@ -505,12 +523,30 @@ def get(self, **kwargs):
                     hits = []
                     total = 0
                     aggregations = {}
+                    shards = None
                     break
 
+        if shards and shards.failed:
```
peterbe: Why not `if shards.get('failed'):` ? Is shards not a dict?

adngdb: It's an ``AttrDict`` returned by ``elasticsearch-dsl``. It's basically like our ``DotDict`` objects.

socorro/unittest/external/es/test_supersearch.py
```diff
@@ -1781,3 +1783,109 @@ def test_get_with_too_many(self):
...
+        "failures": [
+            {
+                "status": 500,
+                "index": "fake_index",
+                "reason": "foo bar gone bad",
+                "shard": 3
+            }
+        ]
+    }
+}'''
```
peterbe: Here too. Using a proper python dict has many advantages.

peterbe:
It works locally but there are a few nits.

socorro/external/es/supersearch.py
```diff
@@ -481,7 +482,15 @@ def get(self, **kwargs):
                     hits.append(self.format_fields(hit.to_dict()))
 
                 total = search.count()
-                aggregations = self.format_aggregations(results.aggregations)
+
+                aggregations = getattr(results, 'aggregations', {})
+                if aggregations:
```
peterbe: No need. If it *is* an empty dict, the `format_aggregations` should just be iterating of an empty dict.

peterbe:
Come on https://bugzilla.mozilla.org/show_bug.cgi?id=1289704

peterbe:
So isn't this potentially flawed. Earlier today we noticed a strange thing in the `_shards` struct returned when I did a Custom search and looked at the raw output. 
```
    "_shards": {
        "successful": 9,
        "failed": 0,
        "total": 10
    }
```
Will the code in this PR cope with that and call that an error?

adngdb:
That case you are describing is something very different than the case I am trying to address. It happens when the cluster is reallocating shards. That is a temporary problem, and it is very hard to know what is happening in the results the cluster returns. It isn't a problem that we can expect to happen, like the memory issue we have had recently. 

Since we won't be able to automatically explain what's wrong with it in the webapp, I'm in favor of not considering that sort of issue here, and taking care of it by other means (email, IRC... ).

PR #3401, opened 2016-07-26
validate_and_test relativity

peterbe:
now I can run this from anywhere

### DIFF
--- a/schemas/validate_and_test.py
+++ b/schemas/validate_and_test.py
@@ -1,14 +1,18 @@
+#!/usr/bin/env python
+
 import json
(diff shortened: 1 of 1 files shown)


peterbe:
Before, the script wasn't doing anything. Because the API expect the parameter to be `crash_id` (not `uuid`) the 100 requests ALL returned something that looked like this:
```
{u'errors': {u'crash_id': [u'This field is required.']}}
```
And `jsonschema.validate` didn't complain :)

adngdb:
r+ (when travis starts working again)

PR #3402, opened 2016-07-27
Fixes bug 1277301 - Added error messages in home and daily when shard…

adngdb:
…s are failing.

This adds a big red warning to the home page and the crashes per day page when shards are failing in Elasticsearch. It requires #3400 to work. 

Here's what it looks like: 

The simplest way to test it is to edit the ``supersearch.py`` middleware file and manually add errors right before returning the results.

### DIFF
--- a/webapp-django/crashstats/crashstats/views.py
+++ b/webapp-django/crashstats/crashstats/views.py
@@ -1127,7 +1127,7 @@ def crashes_per_day(request, default_context=None):
         week = int(error['index'][-2:])
         year = int(error['index'][-6:-2])
         day = isoweek.Week(year, week).monday()
-        percent = error['shards_count'] * 100 / 5
+        percent = error['shards_count'] * 100 / settings.ES_SHARDS_PER_INDEX
         errors.append(
             'The data for the week of {} is ~{}% lower than expected.'.format(
                 day, percent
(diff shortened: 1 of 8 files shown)


webapp-django/crashstats/crashstats/views.py
```diff
@@ -1118,6 +1119,22 @@ def crashes_per_day(request, default_context=None):
...
 
+    errors = []
+    for error in results.get('errors', []):
+        if not error['type'] == 'shards':
+            continue
+
+        week = int(error['index'][-2:])
+        year = int(error['index'][-6:-2])
+        day = isoweek.Week(year, week).monday()
+        percent = error['shards_count'] * 100 / 5
```
peterbe: why integer arithmetic here? Is that really safe?

peterbe:
I would replace the word "We" with "The team".

webapp-django/crashstats/crashstats/views.py
willkg: Does the following look right to you for what you're expecting? 

```
>>> 1 * 100 / 5
20
>>> 2 * 100 / 5
40
>>> 3 * 100 / 5
60
>>> 5 * 100 / 5
100
```

Where's the 5 come from? That seems like a magic number.

adngdb: 5 is the number of shards per index in our Elasticsearch database. It is kind of a magical number. I'll move to configuration, it will make more sense like that.

peterbe:
Conflicting.

PR #3403, opened 2016-07-27
Fixes bug 1289704 - Use ruby 2 in travis build process.

adngdb:
Trying to see if that fixes the issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ services:
   - memcached
 
 before_install:
+  - rvm install 2.0.0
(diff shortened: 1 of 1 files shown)


adngdb:
\o/

Thanks @luser! @peterbe r?

adngdb:
cc @jdotpz too, in case there are consequences to using rvm 2 that I do not know of.

peterbe:
r+

peterbe:
But is it travis that we use to build the build artifact?? I thought it was Jenkins that does that.

PR #3404, opened 2016-07-27
fixes bug 1289833 - Cache-Control on API responses

peterbe:
First load of the home page (or refreshing the page):
screen shot 2016-07-27 at 12 53 49 pm

Second load of the home page (by click, not reloading):
screen shot 2016-07-27 at 12 54 55 pm


peterbe:
It's actually important for the upcoming new Google Sign-In. 
If you try to load the home page and the Google sign-in code says it's time to refresh you get an AJAX error on the homepage because the AJAX request to load the graph data gets interrupted. So if we can load it MUCH faster we won't get that ugly AJAX loading every time you hit the home page.

adngdb:
r+

PR #3407, opened 2016-08-01
fixes bug 1283214 - healthcheck monitor shards too

peterbe:


adngdb:
Let's remove the pagination check, it's a bit too heavy on the database for its value.

webapp-django/crashstats/monitoring/views.py
```diff
@@ -169,57 +169,20 @@ def healthcheck(request):
...
-    If some nodes in the cluster are unhealthy, the whole
-    SuperSearch query might succeed but the number of records
-    might not match the "total" count provided in every result page.
-
-    This test has been tested in our production database, using the
-    /api/SuperSearch endpoint and the epsilon hovers around 0.005 roughly
-    and that's *with* webapp caching at play.
-    """
+def assert_supersearch_no_errors():
+    """Makre sure an uncached SuperSearch query doesn't have any errors"""
```
adngdb: ``Make``

webapp-django/crashstats/monitoring/views.py
```diff
@@ -169,57 +169,20 @@ def healthcheck(request):
...
     supersearch = SuperSearch()
     # We don't want any caching this time
     supersearch.cache_seconds = 0
     results = supersearch.get(
         product=settings.DEFAULT_PRODUCT,
-        _results_number=0,
+        _results_number=1,
         _columns=['uuid'],
+        _facets_size=1,
     )
```
adngdb: What's the point of those ``1``s?

webapp-django/crashstats/monitoring/tests/test_views.py
```diff
@@ -295,75 +280,31 @@ def mocked_supersearch_get(**params):
...
-                eq_(params['_columns'], ['uuid'])
-                return {
-                    'hits': [
-                        {'uuid': '12345'},
-                    ],
-                    'facets': [],
-                    'total': 320,
-                    'errors': ['bad'],
-                }
+            # this is the first one
```
adngdb: You can remove this comment now.

webapp-django/crashstats/monitoring/tests/test_views.py
```diff
@@ -259,32 +259,17 @@ def test_healthcheck(self, mocked_elasticsearch):
...
-                    'hits': [
-                        {'uuid': '0128940'},
-                        {'uuid': '9156826'},
-                        {'uuid': '3969175'},
-                    ],
-                    'facets': [],
-                    'total': 30004,
-                    'errors': [],
-                }
+            # this is the first one
```
adngdb: Same here, obsolete comment.

webapp-django/crashstats/monitoring/views.py
peterbe: So that it queries at least one. Or else ES might not be exercised enough. If it seems that you only need 0 rows, perhaps it won't start the query engine. 
And for `_facets_size` we don't need them. But setting it to 0 is like not setting it at all I vaguely remember.

PR #3409, opened 2016-08-02
fixes bug 1290926 - no facets at all if _facets_size=0

peterbe:
Seems to work with curl.


adngdb:
This is a good start but incomplete. I made a PR to your branch to include sub-aggregations and histograms as well: https://github.com/peterbe/socorro/pull/16

PR #3412, opened 2016-08-03
Bug 1289572 write processed json subset to s3

peterbe:


peterbe:
@adngdb I took your PR (on my fork) and merged it without looking very deeply. 
I also rebased this up to avoid the conflict.

socorro/external/boto/crashstorage.py
```diff
@@ -293,12 +304,37 @@ def save_raw_and_processed(
         processed_crash,
         crash_id
     ):
-        schema = PROCESSED_CRASH_JSON_SCHEMA
-        crash = json_schema_reducer.make_reduced_dict(schema, processed_crash)
+        all_fields = SuperSearchFields(config=self.config).get()
```
peterbe: Can this be moved to the `__init__` so it only needs to download those once?

peterbe:
One big change here is that the `processed_crash.json` file moved. Are there people outside the project who "depend" on it being where it was?

peterbe:
Can we rename the JSON Schema file from "processed_crash.json" to "crash_report.json" or something? People in the telemetry don't really care but we have a hard notion of "processed" being a sub-set of information about a crash.

socorro/external/boto/crashstorage.py
```diff
@@ -272,6 +276,68 @@ class BotoS3CrashStorage(BotoCrashStorage):
...
+
+        # Rename fields in processed_crash.
+        processed_fields_map = dict(
+            (x['in_database_name'], x['name'])
+            for x in all_fields.values()
+            if x['namespace'] == 'processed_crash'
+        )
+        for key, val in processed_crash.items():
+            crash_report[processed_fields_map.get(key, key)] = val
+
```
peterbe: Hmm... we're looping over every field known in SuperSearchFields but then on line 332 we remove many of them. 

How about this:
```python
all_possible_json_schema_keys = find_all_properties(PROCESSED_CRASH_JSON_SCHEMA)

# Rename fields in raw_crash.
 raw_fields_map = dict(
    ....
for key in raw_crash:      
    name = raw_fields_map.get(key, key)
    if name in all_possible_json_schema_keys:
        crash_report[name] = raw_crash[key] 
# (same idea for processed_fields_map and processed_crash)
```

socorro/schemas/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+import json
+
+from pkg_resources import resource_stream
+
+
+def _get_file_content(filename):
+    with resource_stream(__name__, filename) as f:
+        return json.load(f)
+
+PROCESSED_CRASH_JSON_SCHEMA = _get_file_content('processed_crash.json')
```
peterbe: So @adngdb what do you think about renaming this variable and the filename to `CRASH_REPORT_JSON_SCHEMA` and `crash_report.json` respectively?

adngdb: Yep, that'd be good!

adngdb:
> One big change here is that the processed_crash.json file moved. Are there people outside the project who "depend" on it being where it was?

As far as I know, that's not the case. We're good to move / rename it.

socorro/external/boto/crashstorage.py
adngdb: Sure, that works too. I'm not sure it has a huge impact on performance though, so maybe it's not worth it, but do it if you think it's better. :)

socorro/external/boto/crashstorage.py
adngdb: It's not that simple. If processors were to cache that value, we would have to restart them every time we change any value in the SuperSearchFields list. I agree that querying this for every single crash is too much though. 

I think this is the same problem that we had with having the signature lists in the database. I don't remember us finding a good solution (and I do not really consider having to kill the processor threads to be a good solution).

peterbe: ah! that old problem again. 
So, does introducing 1 ES query for every single crash processed introduce a big overhead. Also, ES is notoriously flaky so I'm afraid that every time it fails this read, we either have to skip or it breaks the whole processing. 

What about an optimizing like:

```
all_fields = None
if hasattr(self, 'all_fields') and hasattr(self, 'all_fields_timestamp'):
    if time.time() - self.all_fields_timestamp > 60 * 60:
        # cached but too old
        self.all_fields = None
 if not self.all_fields:
    self.all_fields = SuperSearchFields(config=self.config).get()
    self.all_fields_timestamp = time.time()
all_fields = self.all_fields
```
What do you think?

socorro/external/boto/crashstorage.py
peterbe: I'll make a note about a possibility for optimization.

peterbe:
What's next? I removed the print statement.

Is the next thing to deploy this, in consul, in stage?

socorro/external/boto/crashstorage.py
```diff
@@ -272,6 +276,70 @@ class BotoS3CrashStorage(BotoCrashStorage):
...
+    #--------------------------------------------------------------------------
+    def save_raw_and_processed(
+        self,
+        raw_crash,
+        dumps,
+        processed_crash,
+        crash_id
+    ):
+        all_fields = SuperSearchFields(config=self.config).get()
```
adngdb: I do believe your proposed optimization, with hand-made caching, will be better than calling Elasticsearch every time we process a crash.

socorro/external/boto/crashstorage.py
```diff
@@ -272,6 +276,70 @@ class BotoS3CrashStorage(BotoCrashStorage):
...
+        crash_report = {}
+
+        # TODO Opportunity of optimization;
+        # We could inspect CRASH_REPORT_JSON_SCHEMA and get a list
+        # of all (recursive) keys that are in there and use that
+        # to limit the two following loops to not bother
+        # filling up `crash_report` with keys that will never been
```
adngdb: s/been/be

socorro/unittest/external/boto/test_crashstorage.py
```diff
@@ -39,9 +39,9 @@
 
 # Uncomment these lines to decrease verbosity of the elasticsearch library
 # while running unit tests.
-# import logging
-# logging.getLogger('elasticsearch').setLevel(logging.ERROR)
-# logging.getLogger('requests').setLevel(logging.ERROR)
+import logging
+logging.getLogger('elasticsearch').setLevel(logging.ERROR)
+logging.getLogger('requests').setLevel(logging.ERROR)
```
adngdb: Please comment these again. :)

socorro/unittest/external/boto/test_crashstorage.py
```diff
@@ -1167,3 +1167,116 @@ def test_save_raw_and_processed(self):
...
+        )
+        # NOTE the lack of 'plugin_name=PerfectlyFine' in this S3 write.
+        # It's because the SuperSearchFields renames it to "UNRECOGNIZED_JUNK"
+        # which the JSON Schema is going to reject, so it gets filtered
+        # out before being stored in S3.
+        mocked_set_function.assert_called_with(
+            json.dumps({
+                "uuid": "3f267a63-2f8c-407d-8683-adc452160804",
+            })
+        )
```
adngdb: Those tests are really good!

socorro/unittest/external/boto/test_crashstorage.py
peterbe: done.

PR #3413, opened 2016-08-04
Update platforms for Android

nthomas-mozilla:
Firefox 48.0 dropped android-api-9, and 46.0 changed api-11 to api-15, so look for those when scraping. 

We may have been mis-identifying crashes - eg https://crash-stats.mozilla.com/report/index/b8b4676f-64af-497f-84c2-73e0f2160804 is '48.0 fennec on release', but report a buildID of 20160727164219 instead of 20160726133114. Meanwhile other crashes are fine - eg https://crash-stats.mozilla.com/report/index/99f5e6ce-f435-4e09-8953-baa122160804

### DIFF
--- a/socorro/cron/jobs/ftpscraper.py
+++ b/socorro/cron/jobs/ftpscraper.py
@@ -164,7 +164,7 @@ class ScrapersMixin(object):
 
         possible_platforms = (
             'linux', 'mac', 'win', 'debug',  # for Firefox
-            'android-api-11', 'android-api-9', 'android-x86',  # for mobile
+            'android-api-15', 'android-x86',  # for mobile
         )
 
         for platform in possible_platforms:


peterbe:
When did this change happen? Or will it happen?

I'm trying to gauge the urgency of updating our ftpscraper deployment on stage and prod.

peterbe:
@nthomas-mozilla ^

nthomas-mozilla:
The change has happened already, sorry for not catching it sooner. On the basis of your gist (https://gist.github.com/peterbe/c016866d460984899d8d3eb264ec1b6e) we're still getting new buildIDs for nightly/aurora. I think we'll be missing beta/release though.

peterbe:
Cool. This has now been applied on prod too now. 
Does this look good? https://crash-stats.mozilla.com/api/ProductVersions/?active=true&product=FennecAndroid

nthomas-mozilla:
The versions there look legit but I can't verify the buildID is correct. Wondering how the start_date is determined - if it's related to submission date into the db then perhaps something else is populating the db ? Also interesting that has_builds is True for nightly/aurora but False for beta/release.

PR #3415, opened 2016-08-04
Fixes bug 1290687 - Added a status message to all pages.

adngdb:
cc @lonnen 

Info:

Warning:

Critical:


adngdb:
@peterbe 

```
======================================================================
ERROR: test_create_status_message (crashstats.manage.tests.test_views.TestViews)
----------------------------------------------------------------------
  File "/home/travis/build/mozilla/socorro/webapp-django/crashstats/manage/tests/test_views.py", line 1626, in test_create_status_message
    eq_(event.user, self.user)
AttributeError: 'TestViews' object has no attribute 'user'

======================================================================
ERROR: test_disable_status_message (crashstats.manage.tests.test_views.TestViews)
----------------------------------------------------------------------
  File "/home/travis/build/mozilla/socorro/webapp-django/crashstats/manage/tests/test_views.py", line 1654, in test_disable_status_message
    eq_(event.user, self.user)
AttributeError: 'TestViews' object has no attribute 'user'
```

I can't reproduce that locally. Any idea what it might be? If not I'll investigate on Monday.

peterbe:
I can't reproduce the travis error either. It speaks of an error on line 1626
```
File "/home/travis/build/mozilla/socorro/webapp-django/crashstats/manage/tests/test_views.py", line 1626, in test_create_status_message
```
But 1626 is 
```
eq_(response.status_code, 302)
```
NOT 
```
eq_(event.user, self.user)
```
like it speaks of in the traceback. 

How bizarre! I'll clear the Travis cache and see if that fixes things.

peterbe:
When I click "Disable" it disappears :(
I'd expect it to remain but be disabled.

peterbe:
I think it should be possible to have **multiple** messsages. It's not totally unrealistic. And it's generally better to build things in arrays rather than binary.

webapp-django/crashstats/manage/jinja2/manage/status_message.html
```diff
@@ -0,0 +1,62 @@
+{% extends "manage/base.html" %}
+
+{% block page_title %}{{ super() }} - Status Message{% endblock %}
+
+{% block site_js %}
+{{ super() }}
+{% javascript 'moment' %}
+{% javascript 'pagination' %}
+{% javascript 'api_tokens' %}
```
peterbe: why?

webapp-django/crashstats/manage/jinja2/manage/status_message.html
```diff
@@ -0,0 +1,62 @@
...
+{% endblock %}
+
+{% block site_css %}
+{{ super() }}
+{% stylesheet 'api_tokens' %}
```
peterbe: again, why?

webapp-django/crashstats/manage/jinja2/manage/status_message.html
```diff
@@ -0,0 +1,62 @@
...
+                    <th>Severity</th>
+                    <th>&nbsp;</th>
+                </tr>
+            </thead>
+            <tbody>
+                {% for status in statuses %}
+                <tr>
+                    <td>{{ status.message }}</td>
+                    <td>{{ status.severity }}</td>
+                    <td><a href="{{ url('manage:status_message_disable') }}?id={{ status.id }}">Disable</a></td>
```
peterbe: This should be `{{ url('manage:status_message_disable', status.id) }}` instead. The ID is not optional and that takes care of dealing with any 404s.

webapp-django/crashstats/manage/tests/test_views.py
```diff
@@ -1576,6 +1577,76 @@ def test_api_tokens_delete(self):
...
+
+        # Test while logged out.
+        response = self.client.get(url)
+        eq_(response.status_code, 302)
+
+        self._login()
+        response = self.client.get(url)
+        eq_(response.status_code, 200)
+
+        # expects some permissions to be available as dropdowns
```
peterbe: permissions?

webapp-django/crashstats/manage/urls.py
```diff
@@ -89,6 +89,12 @@
     url('^api-tokens/data/$',
         views.api_tokens_data,
         name='api_tokens_data'),
+    url('^status/$',
+        views.status_message,
+        name='status_message'),
+    url('^status/disable/$',
```
peterbe: Again, this ought to be `'^status/(?P<id>\d+)/disable/$'` instead.

webapp-django/crashstats/manage/views.py
```diff
@@ -1002,3 +1003,53 @@ def reprocessing(request):
...
+@superuser_required
+def status_message_disable(request):
+    if not request.GET.get('id'):
+        return http.HttpResponseBadRequest('No id')
+    status = get_object_or_404(StatusMessage, id=request.GET['id'])
+
+    log(request.user, 'status_message.disable', {})
+
+    status.enabled = False
+    status.save()
```
peterbe: Or `status.update(enabled=False)`

webapp-django/crashstats/status/context_processors.py
```diff
@@ -0,0 +1,18 @@
...
+    if not len(statuses):
+        return {}
+
+    status = statuses[0]
+    return {
+        'status_message': {
+            'text': status.message,
+            'severity': status.severity,
+        },
+    }
```
peterbe: I think this whole thing should be made into a list. E.g.
```python
def status_message(request):
    qs = StatusMessage.objects.filter(enabled=True).order_by('-created_at')
    messages = []
    for status in qs:
        messages.append({ ... })
    return {'status_messages': messages}
```
That takes care of making it possible to have more than one at the same time.

webapp-django/crashstats/status/context_processors.py
```diff
@@ -0,0 +1,18 @@
+from crashstats.status.models import StatusMessage
+
+
+def status_message(request):
+    statuses = (
+        StatusMessage.objects.filter(enabled=True).order_by('-created_at')
+    )
+
+    if not len(statuses):
```
peterbe: By the way, a more efficient way is to do:
```python
if not statuses.count():
```
otherwise it needs to be converted into a list of ORM instance objects in a list but we're only using the length of the list. 

In fact, an even more efficient thing to use is `.exists()`. It's faster in SQL since it returns True as soon as anything exists. 
I.e. 

```
if not statuses.exists():
    return {}
``` 
Not really applicable in this context but might be good to be aware of.

webapp-django/crashstats/status/tests/test_views.py
```diff
@@ -0,0 +1,36 @@
...
+        ok_('#status-message' not in response.content)
+
+        status = StatusMessage.objects.create(
+            message='an incident is ongoing',
+            severity='critical',
+        )
+
+        response = self.client.get(url)
+        eq_(response.status_code, 200)
+        ok_('id="status-message' in response.content)
```
peterbe: This I think should be `class="status-message"` instead.

webapp-django/crashstats/status/static/status/css/status.less
```diff
@@ -0,0 +1,41 @@
+
+#status-message {
```
peterbe: To support multiple, change this to `.status-message`.

peterbe:
Why did you create `webapp-django/crashstats/tokens/tests/__init__.py` and `webapp-django/crashstats/topcrashers/tests/__init__.py`. Pretty sure nosetests didn't need that.

webapp-django/crashstats/manage/views.py
```diff
@@ -1002,3 +1003,53 @@ def reprocessing(request):
...
+            messages.success(
+                request,
+                'Status Message created. '
+            )
+            return redirect('manage:status_message')
+    else:
+        form = forms.StatusMessageForm()
+
+    statuses = (
+        StatusMessage.objects.filter(enabled=True).order_by('-created_at')
```
peterbe: Remove the filter and instead, when you loop over them do something like
```
{% for status in statuses %}
  <tr {% if not status.enabled %}style="text-decoration: line-through" title="Disabled"{% endif %}
```

webapp-django/crashstats/manage/views.py
```diff
@@ -1002,3 +1003,53 @@ def reprocessing(request):
...
+    return render(request, 'manage/status_message.html', context)
+
+
+@superuser_required
+def status_message_disable(request):
+    if not request.GET.get('id'):
+        return http.HttpResponseBadRequest('No id')
+    status = get_object_or_404(StatusMessage, id=request.GET['id'])
+
+    log(request.user, 'status_message.disable', {})
```
peterbe: This is useless without leaving some metadata in there. E.g.
```python
log(request.user, 'status_message.disable', {
    'id': status.id
})
```

peterbe:
Why doesn't travis trigger on https://github.com/adngdb/socorro/pull/5

peterbe:
> Why doesn't travis trigger on adngdb#5
Oh because it's against your fork

peterbe:
Same crazy travis error here: https://travis-ci.org/mozilla/socorro/builds/149840175

webapp-django/crashstats/status/context_processors.py
adngdb: Why would that not be applicable here? Does it not take filters into consideration?

webapp-django/crashstats/status/tests/test_views.py
adngdb: What's wrong with an ID?

webapp-django/crashstats/status/context_processors.py
peterbe: Actually, it is applicable. It's always a good idea to do things right :)
What you're *currently* doing, if you deep log in the database is:
`select * from status_messages where ...`

If you first do a `.exists()` and then a select you're going to trigger 2 database queries:
`select 1 from status_messages where ...` AND `select * from status_messages where ...`.

However, the first query is fast and 99% of the time, the second query won't even have to happen. 

Also, from a performance point of view, what often takes time is the python code to "dress up" the results. I.e. making instances of the `StatusMessage` model in a list-list structure. 

MongoKit was an ORM for MongoDB. It too would have a model and when you make a query, it talks to the database and takes each python `dict` and instantiates it from a model. So, if your model has methods, you can do something like:
```python
for model_instance in model.query(...):
    model_instance.some_method()
```
Anyway, in [this blog post](https://www.peterbe.com/plog/mongokit-dict_plus) I played with opting out of model instance creation and used a DotDict-like thing instead and for a large recordset it made a certain big query 50 times faster.

webapp-django/crashstats/status/tests/test_views.py
peterbe: Because of my previous point that I think you should support MULTIPLE status messages. Then it no longer makes sense with `id="status-message"`. (It might make sense with `id="status-messages"` as a wrapper).

Also, there's a general best practice that using IDs in DOM is bad. 

I try to ALWAYS use `class="..."` instead of `id="..."` even though there's only 1. Who knows, perhaps some day you want to put one in the footer too. 
The exception is when you really want to point out that there can be only one. Kinda like an `assert` statement in non-test python code. It's useless but indicates to the code-reader something of value.

webapp-django/crashstats/manage/jinja2/manage/status_message.html
adngdb: That's not what you did for your other admin apps. And I'm getting more and more reluctant to putting arguments in the URL. It makes sense for some things, but not all, and I find there is a beauty in query strings.

adngdb:
@peterbe the only thing I haven't done from your comments is changing the URL for disabling. Everything else should be addressed. Wanna take another look?

webapp-django/crashstats/manage/jinja2/manage/status_message.html
```diff
@@ -25,7 +25,13 @@
                 <tr>
                     <td {% if not status.enabled %}class="strike" title="Disabled"{% endif %}>{{ status.message }}</td>
                     <td>{{ status.severity }}</td>
-                    <td><a href="{{ url('manage:status_message_disable') }}?id={{ status.id }}">Disable</a></td>
+                    <td>
+                        {% if status.enabled %}
+                        <a href="{{ url('manage:status_message_disable') }}?id={{ status.id }}">Disable</a>
+                        {% else %}
+                        disabled
```
peterbe: why? Don't you want to ever re-enable it?

peterbe:
Hehe on the pagination. You could have just let it be massive. My browser doesn't struggle to cope with 500 rows of plain HTML those once-a-month times that I need to use this :)

Will take a new fresh look tomorrow morning.

webapp-django/crashstats/manage/jinja2/manage/status_message.html
adngdb: Not really, no. If there's something that needs to be said again, I guess you can just create a new message with the same content. But I don't expect problems to happen repeatedly. Or we're not doing a very good job... :) 

Also, that gives us a nice history of what happened on the website over time. You know, for when we're depressed.

PR #3417, opened 2016-08-04
Tests for 1290687

peterbe:


peterbe:
This is an extension of https://github.com/mozilla/socorro/pull/3417 but with some `__init__.py` files deleted.

adngdb:
I don't see the point of _removing_ those files. They don't do any harm. And I think they are actually needed if we want to import some tests stuff from other tests. These are regular python modules. They should have their ``__init__.py`` files.

peterbe:
Every time I do an import from another test I throw up in my mouth a little. It shouldn't be needed. 
However, there are annoyingly many exceptions where you want to import base classes and some constants. [Like this](https://github.com/mozilla/airmozilla/blob/ff80f42aeca420c79a8e9225666175441c6c8690/airmozilla/manage/tests/views/test_closedcaptions.py#L16)

Actually, I wanted to know if it had something to do with the strange travis error. Apparently not. Also, it's often a good idea to not do something unless it's necessary. 

Consider this PR my test bed for trying to figure out what the heck is going on with your PR's tests.

PR #3418, opened 2016-08-09
Fixes bug 1290326 - Do not truncate signatures in Topcrashers.

adngdb:
@peterbe this doesn't break layout for me locally, using stage data. Does it for you? I think that's what we want to do anyway. It's simple and solves the problem. I don't think we care too much if sometimes the data overflows to the right.


peterbe:
I have no idea what I did wrong when [I tried that](https://bugzilla.mozilla.org/attachment.cgi?id=8777343). But now it works wonderfully!
screen shot 2016-08-09 at 8 52 47 am

nnethercote:
I grep'd for `truncate(` and found another similar occurrence in crashstats/crashstats/jinja2/crashstats/topcrasher.html. I'm not sure what part of the webapp that file implements, but would it be worth changing there too?

(Oh, I see now it's the old_topcrashers and includes the text "Dear user, this page is deprecated and will be removed soon". So probably doesn't matter)

adngdb:
@nnethercote Indeed, that's part of the old topcrasher, which should be deleted soon. No need to worry. :)

PR #3423, opened 2016-08-10
cleanup code around boto

peterbe:
Stylistic cleanup and change from `self.assertEqual` to `eq_`.

### DIFF
--- a/socorro/external/boto/crashstorage.py
+++ b/socorro/external/boto/crashstorage.py
@@ -21,7 +21,6 @@ from socorro.external.es.super_search_fields import SuperSearchFields
 from socorro.schemas import CRASH_REPORT_JSON_SCHEMA
 
 
-#==============================================================================
 class BotoCrashStorage(CrashStorageBase):
     """This class sends processed crash reports to an end point reachable
     by the boto S3 library.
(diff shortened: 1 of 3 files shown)


socorro/external/boto/crashstorage.py
```diff
@@ -364,19 +360,18 @@ class SupportReasonAPIStorage(BotoS3CrashStorage):
...
+    #     BotoS3CrashStorage,
+    #     'bucket_name',
+    #     'mozilla-support-reason'
+    # )
+    # required_config.prefix = change_default(
+    #     BotoS3CrashStorage,
+    #     'prefix',
+    #     'mozilla-support-reason'
+    # )
+
```
adngdb: I suspect the comment above is wrong and we should really just delete this. :)

socorro/unittest/external/boto/test_crashstorage.py
```diff
@@ -647,16 +642,16 @@ def temp_failure_fn(*args):
...
             mock.call("936ce666-ff3b-4c7a-9674-367fe2120408", "raw_crash"),
             mock.call("936ce666-ff3b-4c7a-9674-367fe2120408", "raw_crash"),
             mock.call("936ce666-ff3b-4c7a-9674-367fe2120408", "raw_crash"),
         ])
 
-        self.assertEqual(result, self._fake_unredacted_processed_crash())
+        eq_(result, self._fake_unredacted_processed_crash())
 
-    def test_get_unredacted_processed_crash_with_consistency_trouble_no_recover(self):
+    def test_get_unredacted_processed_crash_with_consistency_trouble_no_recover(self):  # NOQA
```
adngdb: Why not just shrink down the name of that function?

adngdb:
r+

socorro/unittest/external/boto/test_crashstorage.py
peterbe: Did not dare to change too much.

PR #3424, opened 2016-08-11
fixes bug 1294437 - Public API is failing to start new threads

peterbe:


peterbe:
I actually don't know if this will actually work because it's impossible to reproduce the error. 
Even if I run uwsgi on my laptop it won't happen. After all, it doesn't happen in prod, unless very rare few occasions.

adngdb:
I think this is good enough. We really just want that external resource to not be able to break our website, and this commit does just that.

PR #3428, opened 2016-08-15
1295206 boto fakes3

willkg:
Implements boto HostPortS3ConnectionContext so that we can connect to a local fake s3. This will make it easier to do development work without requiring that we have S3 buckets and a network connection.

r?

### DIFF
--- a/socorro/external/boto/connection_context.py
+++ b/socorro/external/boto/connection_context.py
@@ -12,6 +12,7 @@ import boto.s3.connection
 import boto.exception
 
 from configman import Namespace, RequiredConfig, class_converter
+from configman.converters import str_to_boolean
 
 from socorrolib.lib.converters import change_default
 from socorrolib.lib.ooid import dateFromOoid
@@ -406,7 +407,7 @@ class HostPortS3ConnectionContext(S3ConnectionContext):
         'secure',
         doc='Whether to connect securely or not (true/false)',
         reference_value_from='resource.boto',
-        from_string_converter=lambda x: x.lower().startswith('t'),
+        from_string_converter=str_to_boolean,
         default=True,
     )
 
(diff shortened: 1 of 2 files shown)


socorro/external/boto/connection_context.py
```diff
@@ -399,3 +379,44 @@ def _connect(self):
...
+        'port',
+        doc='The network port',
+        reference_value_from='resource.boto',
+        from_string_converter=int
+    )
+    required_config.add_option(
+        'secure',
+        doc='Whether to connect securely or not (true/false)',
+        reference_value_from='resource.boto',
+        from_string_converter=lambda x: x.lower().startswith('t'),
```
willkg: configman says it has a bool parser, but its bool parser is the built-in Python type which means an empty string is False and any non-empty string is True. I think that kind of sucks for configuration, so I wrote this which results in more readable configuration values True and False.

What we should probably do is what other configuration libraries do and implement a bool parser that handles a variety of different positive and negative values in configman. I wrote up an issue along those lines:

https://github.com/mozilla/configman/issues/165

Until then, I'm just going to leave this here.

willkg:
^^^ That switches to configman's `str_to_boolean`.

socorro/external/boto/connection_context.py
```diff
@@ -399,3 +380,43 @@ def _connect(self):
...
+        'port',
+        doc='The network port',
+        reference_value_from='resource.boto',
+        from_string_converter=int
+    )
+    required_config.add_option(
+        'secure',
+        doc='Whether to connect securely or not (true/false)',
+        reference_value_from='resource.boto',
+        from_string_converter=str_to_boolean,
```
peterbe: Does it not work without this?

peterbe:
r+

Kinda sucky that configman requires a converter (from strings) on options with a default that is a bool.

socorro/external/boto/connection_context.py
willkg: I'm pretty sure it doesn't, but didn't spend much time looking into it.

willkg:
Thank you!

PR #3429, opened 2016-08-15
Fixes bug 1269471 - Add more lock related signatures to the irrelevan…

amccreight:
…t signature list.


amccreight:
It looks like some unit tests use some of the signatures I wanted to skip. I guess I'll update those (and remove them from the skip list as I assume they must be on there).

adngdb:
Thank you!

PR #3431, opened 2016-08-17
fixes bug 1296017 - default SimpleDatePrefixKeyBuilder on TelemetryBo…

peterbe:
…toS3CrashStorage

### DIFF
--- a/socorro/external/boto/crashstorage.py
+++ b/socorro/external/boto/crashstorage.py
@@ -17,6 +17,9 @@ from socorro.external.crashstorage_base import (
     CrashIDNotFound,
     MemoryDumpsMapping,
 )
+from socorro.external.boto.connection_context import (
+    SimpleDatePrefixKeyBuilder
+)
 from socorro.external.es.super_search_fields import SuperSearchFields
 from socorro.schemas import CRASH_REPORT_JSON_SCHEMA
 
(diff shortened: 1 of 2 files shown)


socorro/external/boto/crashstorage.py
```diff
@@ -292,6 +292,20 @@ class TelemetryBotoS3CrashStorage(BotoS3CrashStorage):
         reference_value_from='resource.elasticsearch',
     )
 
+    required_config.add_option(
+        'keybuilder_class',
```
willkg: I don't understand what you're doing here. The keybuilder class is part of the connection context--not the crash storage. I don't see how the changes you're making has an effect.

socorro/external/boto/crashstorage.py
```diff
@@ -292,6 +292,20 @@ class TelemetryBotoS3CrashStorage(BotoS3CrashStorage):
...
+        default=(
+            'socorro.external.boto.connection_context'
+            '.SimpleDatePrefixKeyBuilder'
+        ),
+        doc=(
+            'fully qualified dotted Python classname to handle building s3 '
+            'pseudo-filenames'
+        ),
+        from_string_converter=class_converter,
+        reference_value_from='resource.boto',
```
willkg: Given that this class very specifically needs the SimpleDatePrefixKeyBuilder, I wouldn't have it pull the key builder from `resource.boto`.

socorro/external/boto/crashstorage.py
willkg: Why not instead override `__init__` and change the config there explicitly before calling `super().__init__()` which instantiates the connection context?

socorro/external/boto/crashstorage.py
peterbe: I suspect you're right. I might have rushed this. 

Now that you mention it, it's true that we probably never want to have anything other than SimpleDatePrefixKeyBuilder so it could be moved from configuration to hardcoded in `__init__`.

peterbe:
Now I'm really not sure how to do it. @adngdb do you know a better way. 
Instead of supplanting the config, I just put it in but disallowed it to read the value from `resource.boto`.

Is there a better way?

PS. I tried to mess with setting it "by force" in the `__init__` but it doesn't work because I can't change the config of `self.config.resource_class.config` because that doesn't exist.

willkg:
I think you want to do something like this:

```
def __init__(self, config, quit_check_callback=None):
    # TelemetryS3BlahBlahBlah requires that we use
    # SimpleDatePrefixKeyBuilder, so we stomp on the configuration
    # here.
    config.keybuilder_class = 'socorro.external.boto.connection_context.SimpleDatePrefixKeyBuilder'
    super(TelemetryS3BlahBlahBlah, self).__init__(config, quit_check_callback)
```

Did you try something like that?

adngdb:
You know, I think this is actually just something we could solve with a configuration key. It is intended to work this way, and trying to do it differently will only lead to hacks. 

Otherwise, Will's suggestion might just be the best one.

peterbe:
@willkg Thank you for not giving up on this. 
It's kinda nuts to hardcode things in a framework full of config but that crashstorage is really dependent on the right folder structure. 
Let's have this merged in on stage as soon as @jdotpz gives us the go-ahead to deploy on stage.

willkg:
Travis is happy and I think this looks good to me. r+

PR #3438, opened 2016-08-22
fixes bug 1296208 - super search error when dates are illogical

peterbe:
I'm not super familiar with this code but I think I have a decent solution. It's decent because it's not hard coded specifically to "date" values.

Basically, if you have multiple prefixed values, each and every value is compared and checks that the operator function for each makes sense.

### DIFF
--- a/webapp-django/crashstats/supersearch/tests/test_form_fields.py
+++ b/webapp-django/crashstats/supersearch/tests/test_form_fields.py
@@ -1,7 +1,8 @@
 import datetime
-from nose.tools import eq_
+from nose.tools import eq_, assert_raises
 
 from django.utils.timezone import utc
+from django.forms import ValidationError
 
 from crashstats.base.tests.testbase import TestCase
 from crashstats.supersearch import form_fields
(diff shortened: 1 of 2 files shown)


peterbe:
What is looks like:
screen shot 2016-08-22 at 2 38 41 pm
I'm not convinced the error message is particularly easy to understand. We could be more "human" and say "One of the values of multiple doesn't make sense with that operator."

webapp-django/crashstats/supersearch/tests/test_form_fields.py
```diff
@@ -61,3 +62,95 @@ def test_several_fields(self):
...
+        # more than two fields
+        assert_raises(
+            ValidationError,
+            field.clean,
+            ['>2016-08-01', '<=2016-08-02', '>=2016-08-02']
+        )
+        assert_raises(
+            ValidationError,
+            field.clean,
+            ['<2016-08-01', '<2016-08-02', '<2016-08-03']
```
adngdb: I don't see why this should error out. It's stupid, certainly, but definitely not invalid.

peterbe: Yeah, tricky. 
If a human typed this into the UI in Super search and you leaned over her shoulder, would you say she's done it wrong? After all, it's like saying "I want something less than 10 and something less than 11 and something less than 12" which you "should" be saying "I want something less than 12".

PR #3439, opened 2016-08-22
Bug 1293776 initial pass on StatsdMetrics

willkg:
This creates a new kind of external component called Metrics. It gets used in the breakpad collector after saving the crash to log data about the incoming crash at the point we've collected it.

r?

### DIFF
--- a/socorro/collector/wsgi_breakpad_collector.py
+++ b/socorro/collector/wsgi_breakpad_collector.py
@@ -91,14 +91,15 @@ class BreakpadCollectorBase(GenericCollectorBase):
         else:
             raw_crash.legacy_processing = int(raw_crash.legacy_processing)
 
-        # We want to capture the crash report size, but need to differentiate
-        # between compressed vs. uncompressed data as well as accepted vs.
-        # rejected data.
-        crash_report_size = self._get_content_length()
-        is_compressed = self._is_content_gzipped()
-        is_accepted = (raw_crash.legacy_processing in (ACCEPT, DEFER))
 
         try:
+            # We want to capture the crash report size, but need to
+            # differentiate between compressed vs. uncompressed data as well as
+            # accepted vs. rejected data.
+            crash_report_size = self._get_content_length()
+            is_compressed = self._is_content_gzipped()
+            is_accepted = (raw_crash.legacy_processing in (ACCEPT, DEFER))
+
             metrics_data = {}
             size_key = '_'.join([
                 'crash_report_size',
@@ -109,12 +110,11 @@ class BreakpadCollectorBase(GenericCollectorBase):
                 size_key: crash_report_size
             }
             self.metrics.capture_stats(metrics_data)
-        except Exception as exc:
+        except Exception:
             # We *never* want metrics reporting to prevent saving a crash, so
             # we catch everything and log an error.
             self.logger.error(
-                'metrics kicked up exception: %s',
-                str(exc),
+                'metrics kicked up exception',
                 exc_info=True
             )
 
--- a/socorro/collector/wsgi_breakpad_collector.py
+++ b/socorro/collector/wsgi_breakpad_collector.py
@@ -5,7 +5,7 @@
 import time
 
 from socorrolib.lib.ooid import createNewOoid
-from socorro.collector.throttler import DISCARD, IGNORE
+from socorro.collector.throttler import ACCEPT, DEFER, DISCARD, IGNORE
 from socorrolib.lib.datetimeutil import utc_now
 from socorro.collector.wsgi_generic_collector import GenericCollectorBase
 
@@ -91,6 +91,33 @@ class BreakpadCollectorBase(GenericCollectorBase):
         else:
             raw_crash.legacy_processing = int(raw_crash.legacy_processing)
 
+        # We want to capture the crash report size, but need to differentiate
+        # between compressed vs. uncompressed data as well as accepted vs.
+        # rejected data.
+        crash_report_size = self._get_content_length()
+        is_compressed = self._is_content_gzipped()
+        is_accepted = (raw_crash.legacy_processing in (ACCEPT, DEFER))
+
+        try:
+            metrics_data = {}
+            size_key = '_'.join([
+                'crash_report_size',
+                'accepted' if is_accepted else 'rejected',
+                'compressed' if is_compressed else 'uncompressed',
+            ])
+            metrics_data = {
+                size_key: crash_report_size
+            }
+            self.metrics.capture_stats(metrics_data)
+        except Exception as exc:
+            # We *never* want metrics reporting to prevent saving a crash, so
+            # we catch everything and log an error.
+            self.logger.error(
+                'metrics kicked up exception: %s',
+                str(exc),
+                exc_info=True
+            )
+
         if raw_crash.legacy_processing == DISCARD:
             self.logger.info('%s discarded', crash_id)
             return "Discarded=1\n"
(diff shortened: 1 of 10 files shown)


willkg:
This can land as is. After it lands, we'll want to make a configuration change to switch to the `StatsdMetrics` class so that it can log things to datadog.

willkg:
Mmm... It occurred to me that the placement of the metrics capture prevents us from capturing sizes for rejected crash reports. We probably want that, too. I'll fix it now.

willkg:
I'll have to track down that test bug, too. I've never seen that test before, so I'll have to go figure out how it gets set up.

willkg:
^^^ Travis is happy and I fixed the crash size metrics placement. So this is good to go.

To review:

1. Run the tests if you like
2. Read through the code and my pr comments

@peterbe Can you look at this tomorrow (Tuesday)? I'd love to land this Wednesday at the latest so we can collect data for the work week.

socorro/collector/wsgi_breakpad_collector.py
```diff
@@ -86,6 +91,33 @@ def POST(self, *args):
...
+            }
+            self.metrics.capture_stats(metrics_data)
+        except Exception as exc:
+            # We *never* want metrics reporting to prevent saving a crash, so
+            # we catch everything and log an error.
+            self.logger.error(
+                'metrics kicked up exception: %s',
+                str(exc),
+                exc_info=True
+            )
```
willkg: I'm pretty sure this section is safe. Does it look safe to you? If it goes sideways, does it tell us enough information that we can do something useful?

socorro/collector/wsgi_generic_collector.py
```diff
@@ -64,10 +64,18 @@ def _process_fieldstorage(self, fs):
             return fs
 
     #--------------------------------------------------------------------------
+    def _is_content_gzipped(self):
+        return web.ctx.env.get('HTTP_CONTENT_ENCODING') == 'gzip'
+
+    #--------------------------------------------------------------------------
+    def _get_content_length(self):
+        return web.ctx.env.get('CONTENT_LENGTH', 0)
```
willkg: I added these here because in the tests we mock the bejesus out of `web` in this module and it simplified everything if we kept web usage here.

socorro/collector/wsgi_breakpad_collector.py
peterbe: You might as well still putting the `exc` into the string. The `exc_info=True` will take care of that much better. 
Also, on the paranoid side of things, what if the `exc` object doesn't respond well to being `str()` called? E.g. some unicode object in there that's trouble.

socorro/collector/wsgi_breakpad_collector.py
```diff
@@ -86,6 +91,33 @@ def POST(self, *args):
...
+            size_key = '_'.join([
+                'crash_report_size',
+                'accepted' if is_accepted else 'rejected',
+                'compressed' if is_compressed else 'uncompressed',
+            ])
+            metrics_data = {
+                size_key: crash_report_size
+            }
+            self.metrics.capture_stats(metrics_data)
+        except Exception as exc:
```
peterbe: Perhaps only the `self.metrics.capture_stats(metrics_data)` line should be in the `try:except:`. 
Or perhaps the `self._get_content_length()` should be included too?

However, we might be best to put this into stage for a couple of days/weeks and see if we ever get any exceptions on any of the lines *other* than `self.metrics.capture_stats(metrics_data)`.

socorro/collector/wsgi_breakpad_collector.py
```diff
@@ -42,6 +42,7 @@ def __init__(self, config):
         self.accept_submitted_legacy_processing = \
             self._get_accept_submitted_legacy_processing()
         self.throttler = self._get_throttler()
+        self.metrics = self._get_metrics()
```
peterbe: Why not just `self.metrics = self.config.metrics`?
What's the purpose of the `_get_metrics` method?

socorro/collector/wsgi_breakpad_collector.py
```diff
@@ -86,6 +91,33 @@ def POST(self, *args):
         else:
             raw_crash.legacy_processing = int(raw_crash.legacy_processing)
 
+        # We want to capture the crash report size, but need to differentiate
+        # between compressed vs. uncompressed data as well as accepted vs.
+        # rejected data.
+        crash_report_size = self._get_content_length()
+        is_compressed = self._is_content_gzipped()
+        is_accepted = (raw_crash.legacy_processing in (ACCEPT, DEFER))
```
peterbe: excessive bracket :)

socorro/collector/wsgi_breakpad_collector.py
willkg: This is how most of the components are structured. I'm not entirely clear on why it's this way. In some cases, the `_get*` methods cache the results. One nice thing about it is that it's easy to mock out in tests, though I didn't mock out `_get_metrics` anywhere.

I think I'm hesitant to change this way of doing things at this juncture.

socorro/collector/wsgi_breakpad_collector.py
willkg: Does this go against the socorro style guide? I haven't seen evidence of that, yet.

I like this style because it cordons off comparison expressions and makes it clearer that it's an assignment and not a typo of `==`. So I do this often.

socorro/collector/wsgi_breakpad_collector.py
willkg: Good thoughts here. I'll move it all into the try/except block.

socorro/collector/wsgi_breakpad_collector.py
willkg: I was surprised our logger doesn't have a `.exception()` method. I thought about looking into what logger actually is here, but then never got around to it.

Anyhow, these are good points. I'll nix the `str(exc)` bit.

socorro/external/metrics_base.py
```diff
@@ -0,0 +1,44 @@
...
+    def capture_stats(self, data_items):
+        """Takes a dict of values to capture
+
+        This is for measuring the statistical distribution of a set of values
+        over time. For example, crash sizes or db query durations.
+
+        :arg data_items: dict of key/value pairs; values are always ints
+
+        """
+        pass
```
peterbe: Should this perhaps be `raise NotImplementedError`?

socorro/external/metrics_base.py
```diff
@@ -0,0 +1,44 @@
...
+class MetricsBase(RequiredConfig):
+    """Base class for Metrics as well as a no-op default
+
+    This is used by the collector on incoming crashes so that we can collect
+    metrics on them. This data can be logged, sent to statsd,
+    etc--functionality implemented by subclasses of this class.
+
+    Subclasses should implement::
+
+        def capture_data(self, raw_crash, dumps, crash_id)
```
peterbe: `capture_data` or `capture_stats`?

socorro/unittest/collector/test_wsgi_breakpad_collector.py
```diff
@@ -505,6 +530,68 @@ def test_no_x00_character(self):
...
+        mocked_utc_now.return_value = datetime(2012, 5, 4, 15, 10)
+        mocked_time.time.return_value = 3.0
+        c.throttler.throttle.return_value = (ACCEPT, 100)
+        r = c.POST()
+
+        # Verify capture_stats was called
+        eq_(config.metrics.capture_stats_calls, 1)
+        ok_(r.startswith('CrashID=bp-'))
+
+
```
peterbe: Very good!

socorro/external/metrics_base.py
peterbe: Reason being, the class *is* called SomethingSomethingBase.

peterbe:
There are some aspects of the whole functionality I don't understand but to understand it fully would be too big a time commitment and then you'd never have this into stage :)

r+

peterbe:
Not merging yet because of the question mark on the doc string `capture_data`.

willkg:
-stage is free again, so I'm going to merge. Thank you for reviewing this!

PR #3442, opened 2016-08-24
fixes bug 1297684 - force LocMemCache in tests

peterbe:
@adngdb ,
what's neat about this is that you can now have memcache in your `.env` (or leave it out and let it be the default from `settings/base.py`) so that when you do a lot of runserver work it's able to cache things after runserver restarts.


willkg:
Looks good to me. r+

PR #3445, opened 2016-08-25
running a local processor

peterbe:


peterbe:
@willkg This is what I needed to compliment that big email I sent about how to run a local processor. The code in this PR doesn't change anything existing and I don't really need a review but I also don't want to blatantly commit this into master without some pairs of eyes on it. 

cc @adngdb

adngdb:
This looks good. I would move the script to a different folder (``./tools``? ) and add a comment at the start of it to explain what it does.

PR #3446, opened 2016-08-25
fixes bug 1283296 - Redirect back after sign-in

peterbe:
The two major changes here are:

1) The oauth2.js code now seeks out a `?next=/some/url` if available and redirects to that instead of doing `document.location.reload()`.

2) The `superuser_required` decorator is rewritten. One of the reasons we wrote our our (a long time ago) was to do two things: A) check that you're signed in AND superuser and B) if you're not a superuser store a flash message. By instead using the well-tested one in django.contrib.auth.dectorators we get the `?next=/admin/` for example when you try to access a superuser protected view. However the flash message is lost but the login page already has a message for those who are signed in but not allowed all the way in.

### DIFF
--- a/webapp-django/crashstats/crashstats/decorators.py
+++ b/webapp-django/crashstats/crashstats/decorators.py
@@ -38,6 +38,27 @@ def login_required(
     return actual_decorator
 
 
+def superuser_required(
+    function=None,
+    redirect_field_name=REDIRECT_FIELD_NAME,
+    login_url=None
+):
+    """Same logic as in login_required() (see doc string above) but with
+    the additional check that we require you to be superuser also.
+    """
+
+    def check_user(user):
+        return user.is_active and user.is_superuser
+
+    actual_decorator = user_passes_test(
+        check_user,
+        login_url=login_url,
+        redirect_field_name=redirect_field_name
+    )
+    if function:
+        return actual_decorator(function)
+    return actual_decorator
+
 _marker = object()
 
 
--- a/webapp-django/crashstats/crashstats/decorators.py
+++ b/webapp-django/crashstats/crashstats/decorators.py
@@ -38,27 +38,6 @@ def login_required(
     return actual_decorator
 
 
-def superuser_required(
-    function=None,
-    redirect_field_name=REDIRECT_FIELD_NAME,
-    login_url=None
-):
-    """Same logic as in login_required() (see doc string above) but with
-    the additional check that we require you to be superuser also.
-    """
-
-    def check_user(user):
-        return user.is_active and user.is_superuser
-
-    actual_decorator = user_passes_test(
-        check_user,
-        login_url=login_url,
-        redirect_field_name=redirect_field_name
-    )
-    if function:
-        return actual_decorator(function)
-    return actual_decorator
-
 _marker = object()
 
 
--- a/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/oauth2.js
+++ b/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/oauth2.js
@@ -59,15 +59,21 @@ var OAuth2 = (function() {
     };
 
     /* http://stackoverflow.com/a/901144/205832 */
-    var getParameterByName = function getParameterByName(name, url) {
-        if (!url) url = window.location.href;
+    function getParameterByName(name, url) {
+        if (!url) {
+            url = window.location.href;
+        }
         name = name.replace(/[\[\]]/g, "\\$&");
-        var regex = new RegExp("[?&]" + name + "(=([^&#]*)|&|#|$)"),
-            results = regex.exec(url);
-        if (!results) return null;
-        if (!results[2]) return '';
+        var regex = new RegExp("[?&]" + name + "(=([^&#]*)|&|#|$)");
+        var results = regex.exec(url);
+        if (!results) {
+            return null;
+        }
+        if (!results[2]) {
+            return '';
+        }
         return decodeURIComponent(results[2].replace(/\+/g, " "));
-    };
+    }
 
     return {
         init: function() {
--- a/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/oauth2.js
+++ b/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/oauth2.js
@@ -58,7 +58,6 @@ var OAuth2 = (function() {
         });
     };
 
-
     /* http://stackoverflow.com/a/901144/205832 */
     var getParameterByName = function getParameterByName(name, url) {
         if (!url) url = window.location.href;
@@ -123,12 +122,14 @@ var OAuth2 = (function() {
                         .done(function(response) {
                             // It worked!
                             var next = getParameterByName('next');
+                            // only if ?next=/... exists on the current URL
                             if (next) {
                                 // A specific URL exits.
                                 // This is most likely the case when you tried
                                 // to access a privileged URL whilst being
                                 // anonymous and being redirected.
-                                document.location.href = next;
+                                // Make sure it's on this server
+                                document.location.pathname = next;
                             } else {
                                 document.location.reload();
                             }
--- a/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/oauth2.js
+++ b/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/oauth2.js
@@ -58,23 +58,6 @@ var OAuth2 = (function() {
         });
     };
 
-    /* http://stackoverflow.com/a/901144/205832 */
-    function getParameterByName(name, url) {
-        if (!url) {
-            url = window.location.href;
-        }
-        name = name.replace(/[\[\]]/g, "\\$&");
-        var regex = new RegExp("[?&]" + name + "(=([^&#]*)|&|#|$)");
-        var results = regex.exec(url);
-        if (!results) {
-            return null;
-        }
-        if (!results[2]) {
-            return '';
-        }
-        return decodeURIComponent(results[2].replace(/\+/g, " "));
-    }
-
     return {
         init: function() {
             // The "signin" meta tag, and its value being "signout"
@@ -127,7 +110,9 @@ var OAuth2 = (function() {
                         $.post(url, data)
                         .done(function(response) {
                             // It worked!
-                            var next = getParameterByName('next');
+                            var next = Qs.parse(
+                                document.location.search.slice(1)
+                            ).next;
                             // only if ?next=/... exists on the current URL
                             if (next) {
                                 // A specific URL exits.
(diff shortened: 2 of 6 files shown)


webapp-django/crashstats/crashstats/static/crashstats/js/socorro/oauth2.js
```diff
@@ -110,8 +122,16 @@ var OAuth2 = (function() {
...
                             // It worked!
-                            // TODO: https://bugzilla.mozilla.org/show_bug.cgi?id=1283296
-                            document.location.reload();
+                            var next = getParameterByName('next');
+                            if (next) {
+                                // A specific URL exits.
+                                // This is most likely the case when you tried
+                                // to access a privileged URL whilst being
+                                // anonymous and being redirected.
+                                document.location.href = next;
```
willkg: It doesn't look like we're sanitizing or validating the url before redirecting someone. If we are, I'm totally missing it.

peterbe: We don't. Here. Django's `user_passes_test` or whatever it's called does that.

willkg: I don't see how you can sanitize it in Django. Pretty sure you have to do it in JS.

For the record, I'm talking about this:

https://www.owasp.org/index.php/Unvalidated_Redirects_and_Forwards_Cheat_Sheet

peterbe: Those are all redirects in server-side. 
The `?next=...` that Django generates will always only be `/...` or `https://crash-stats.mozila.com/...`

willkg: URLs are considered user-provided values. So any user can click on a url that has a next param and maybe that next param was generated by Django, but maybe it wasn't.

It sounds like the answer to my question is "no, it's not sanitized" in which case we should add some code to do that here.

peterbe: What did you have in mind? Check that it starts with a `/`? Or check that it doesn't contain any query string characters like `?` or `&`?
If I can trick your browser to visit `https://crash-stats.m.c/login/?next=/admin/some/dangerouspage/that/does/writes` then your browser would go to that after you sign in. But it's only doing a GET to get there. 

We could also whitelist a bunch of paths that are OK. Like `['/admin/', '/api/tokens/', '/profile/', '/symbols/']`.

willkg: One thing we want to do is make sure we're redirecting to a url that's explicitly on this server and not a url on some other server. My JS is lacking, so I asked @Osmose and he suggested something like this:

```
document.location.href = URL(next).pathname;
```

If you need to pull in the query params, those are in `.searchParams`.

Further, he pointed out you can pull out the next param from this url's query string with this:

```
next = URL(window.location.href).searchParams.next;
```

If that works (I haven't tested it), then you can ditch the function you pulled from StackOverflow.

peterbe: That's just a half-solution. A good start. We still have the concern that you trick people to go to a page like `https://crash-stats.m.c/login/?next=/admin/some/dangerouspage/that/does/writes`

However, if we have URLs that does writes on a HTTP GET, we ought to fix that first :)

willkg: We shouldn't have pages that do dangerous writes via a GET request. If we do, that's a separate set of problems from this one.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/oauth2.js
```diff
@@ -110,8 +121,18 @@ var OAuth2 = (function() {
...
+                            var next = getParameterByName('next');
+                            // only if ?next=/... exists on the current URL
+                            if (next) {
+                                // A specific URL exits.
+                                // This is most likely the case when you tried
+                                // to access a privileged URL whilst being
+                                // anonymous and being redirected.
+                                // Make sure it's on this server
+                                document.location.pathname = next;
+                            } else {
```
peterbe: @willkg Annoyingly, all the previous discussions disappeared when I made this new commit. 

First of all, `URL` doesn't work for us. If you do `new URL('/foo/', document.location.href).href` it nicely takes the base from that second parameter and puts `/foo/` as the pathname. 
But if you do `new URL('http://www.example.com/something', document.location.href).href` the output becomes `http://www.example.com/something`. 

Also, this `new URL(...).searchParams` I don't see anywhere. Certainly not in Firefox. Perhaps we're talking about different things. 

Lastly, one important change I made here is that I switched to using `document.location.PATHNAME = next`. So if someone tries to be funny and tricks me to go to `https://crash-stats.m.c/login/?next=http://pr0n.net/spam` it will attempt to redirect to `https://crash-stats.m.c/http:/pr0n.net/spam` which will simply 404.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/oauth2.js
```diff
@@ -58,6 +58,17 @@ var OAuth2 = (function() {
...
+    /* http://stackoverflow.com/a/901144/205832 */
+    var getParameterByName = function getParameterByName(name, url) {
+        if (!url) url = window.location.href;
+        name = name.replace(/[\[\]]/g, "\\$&");
+        var regex = new RegExp("[?&]" + name + "(=([^&#]*)|&|#|$)"),
+            results = regex.exec(url);
+        if (!results) return null;
+        if (!results[2]) return '';
+        return decodeURIComponent(results[2].replace(/\+/g, " "));
+    };
```
adngdb: I know it comes from stackoverflow and you didn't write it, but I would still make this pass our linter. Notably, no inlined ``if`` statement and not multiple ``var`` declarations.

webapp-django/crashstats/crashstats/decorators.py
```diff
@@ -38,6 +38,27 @@ def login_required(
     return actual_decorator
 
 
+def superuser_required(
```
adngdb: Why move it to the ``crashstats`` app if it's only ever used in the ``manage`` app? I'd rather have it in ``manage/decorators.py`` than here.

adngdb:
I feel like Will is much more knowledgeable about security issues than I am, so I'd like to get his approval before we merge this. Regarding the rest of the code, I have only nits.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/oauth2.js
mythmon: I've convinced myself that `searchParams` is a decently safe way to do this, if a bit nerve wracking. It only works in Firefox though. I don't know if that's a concern for you. I'd put some sort of fallback, in case of other browsers. At least a `return` afterwards, in case the browser falls through.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/oauth2.js
peterbe: done.

webapp-django/crashstats/crashstats/decorators.py
peterbe: Will do!

peterbe:
Pretty confident this ready for merge @adngdb

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/oauth2.js
willkg: @peterbe: I don't really understand what you tried to do. I was thinking you'd create a new `URL` object using the next value and then set the pathname. Something like this:

```
var thisURL = URL(document.location.href);
var newURL = thisURL.searchParams['next'];

if (next) {
    documentation.location.pathname = newURL.pathname;
} else {
    ....
}
```

Why wouldn't that work here?

peterbe: That doesn't work in Firefox.

willkg: Mmm... Can you elaborate? What specifically doesn't work? Can I see the code you tried?

peterbe: screen shot 2016-09-12 at 3 16 09 pm

peterbe: i.e. you can't extract the query string parameters from the `URL` native object thing.

willkg: I read the docs and `.searchParams` is a `URLSearchParams` object. This works locally:

```
>> var url = new URL('https://example.com/?next=bar');
< undefined
>> url.searchParams.get('next');
< "bar"
```

peterbe: What does "locally" mean?
screen shot 2016-09-12 at 3 36 45 pm

Osmose: `searchParams` isn't a mapping, is an object with methods you use to modify and read the parameters.

`searchParams.get` works in both Firefox and Chrome (I tested Dev Edition and whatever the latest Chrome my machine has): http://codepen.io/anon/pen/RGAAQq?editors=0012

```js
var url = new URL('https://www.mozilla.org/foo/bar?param1=4&param2=7');
console.log(url.searchParams.get('param1'));
```

peterbe: Pardon my terrible testing, but I did test this on Safari and there it fails:
screen shot 2016-09-12 at 4 15 57 pm
That means iOS users most likely wouldn't be able to be redirected to where they wanted to go.

adngdb: We have the ``Qs`` library in Socorro, so you can use it like this: 

```javascript
Qs.parse(document.location.search.slice(1));
// Object { product: "Firefox", _dont_run: "1" }
```

This is compatible with most browsers, and is already in our base bundle, so it's accessible anywhere in the webapp. 

Documentation: https://github.com/ljharb/qs/

adngdb: After re-reading the code, yes, please do use ``Qs`` instead of your current solution. It will require less code and will be more robust. 

I like your solution of putting the ``next`` value into the pathname, that's great.

peterbe: Done.

adngdb:
r+

PR #3449, opened 2016-08-26
fixes bug 1137376 - symbols upload page public

peterbe:
cc @luser 

Sorry that this is a bit messy Adrian. I even had to put in html comments. 
There are a bunch of difference states but ultimately this PR is about being able to visit `http://socorro.dev/symbols/upload/api/` without being signed in. That way the documentation can be available for people who have an auth token but just aren't signed in on the webapp. 
screen shot 2016-08-26 at 3 56 17 pm

Beyond that, there are states that try to guide the user. 
1. You're signed in but haven't generated a API Token valid for symbol uploads
2. You have an API Token valid for symbol uploads
3. You're signed in but don't even have permission to generate an API Token for the necessary permission.

### DIFF
--- a/webapp-django/crashstats/symbols/jinja2/symbols/api_upload.html
+++ b/webapp-django/crashstats/symbols/jinja2/symbols/api_upload.html
@@ -24,12 +24,23 @@
         <h2>Scripted API Upload</h2>
     </div>
     <div class="body">
-        {% if has_possible_token %}
-        <p>
-            Now that you have an actively working
-            <a href="{{ url('tokens:home') }}">API Token</a> that has the
-            <code>{{ required_permission.name }}</code> permission you can use the API to upload your symbols files.
-        </p>
+        {% if has_possible_token or not request.user.is_active %}
+            {% if has_possible_token %}
+                <!-- Signed in, has a possible API token to use -->
+                <p>
+                    Now that you have an actively working
+                    <a href="{{ url('tokens:home') }}">API Token</a> that has the
+                    <code>{{ required_permission.name }}</code> permission you can use the API to upload your symbols files.
+                </p>
+            {% else %}
+                <!-- Not signed in -->
+                <p>
+                    <b>Note!</b>
+                    To be able to upload symbols you need to have an
+                    <a href="{{ url('tokens:home') }}"><b>API Token</b></a>.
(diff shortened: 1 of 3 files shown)


adngdb:
The logic of the HTML template doesn't make sense to me. I tried to reduce as much as possible, to this: 

```python
if user has token or user is not signed in:
    if user has token:
        NOTIFICATION ABOUT BEING ALL SET
    else if user is not signed in:
        NOTIFICATION THAT YOU NEED A TOKEN

    INSTRUCTIONS ON HOW TO USE THE API
else if user is signed in and user has no token:
    if user has permission:
        NOTIFICATION ABOUT NEEDING A TOKEN
    else if user doesn't have permission:
        NOTIFICATION ABOUT NEEDING A PERMISSION
```

I don't understand why signed in users do not get the instructions in some cases while signed out users always get them. I would greatly simplify that logic, doing something like this: 

```python
if user is not signed in:
    NOTIFICATION THAT YOU NEED A TOKEN
else if user is signed in:
    if user has token:
        NOTIFICATION ABOUT BEING ALL SET
    else if user has permission:
        NOTIFICATION ABOUT NEEDING A TOKEN
    else if user doesn't have permission:
        NOTIFICATION ABOUT NEEDING A PERMISSION
        
INSTRUCTIONS ON HOW TO USE THE API
```

peterbe:
> I don't understand why signed in users do not get the instructions in some cases while signed out users always get them. 

That's the whole point. The full instructions should appear for people who are not signed in because it's very plausible that they possess an auth key but is to "lazy" to sign in on the website. 

And when you do sign in on the website, but don't have the permission it's just a tease about showing instructions that will "never" be useful.

adngdb:
I disagree with your comments. 

> it's very plausible that they possess an auth key

I don't think there's any way to prove that. They might just have wandered there for any reason, or they might be interested but not have the permission yet. 

To me there are only two good ways of doing this: either you show the instructions to everyone, or you show them only to people that are ready to use them. Any other option doesn't make sense to me.

peterbe:
Ok. I've simplified it. See two screenshots:

**Anonymous**
screen shot 2016-09-26 at 3 05 17 pm

**Signed in and has permission**
screen shot 2016-09-26 at 3 05 06 pm

PR #3451, opened 2016-08-30
fixes bug 1298969

lonnen:
removes the earlier attempt to use min-height and replaces it with a flexbox based approach that ensures the footer appears either below the content or at the bottom of the page

My local env is hosed so this needs a little extra testing. I've checked this change against the main page and search page using dev tools. If something is broken it will not be subtle -- after landing, load the main page, search, topcrasher, daily, report list, and report index to check. You may want to use devtools to remove a div panel or two in order to make room.

Before:
screenshot 2016-08-29 15 05 03-2

After:
screenshot 2016-08-29 15 09 52

### DIFF
--- a/webapp-django/crashstats/crashstats/static/crashstats/css/screen.less
+++ b/webapp-django/crashstats/crashstats/static/crashstats/css/screen.less
@@ -215,6 +215,7 @@ html.production {
 }
 .page-header {
     position: relative;
+    min-height: 64px;
     height: 64px;
     background-color: #0a2533;
     background-image: url(../../img/2.0/header-bg.png);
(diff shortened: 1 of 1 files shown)


adngdb:
It doesn't seem to work for me, maybe something is missing? 

socorro-flex

lonnen:
Good catch! See my commit message on the second commit. This should resolve it.

I initially wrote this between meetings during the workday, noticed an issue on search with the `height: 100%` approach, and then forgot about it in my hustle to make a patch before bed at the end of the day.

adngdb:
Much better, the footer is always at the bottom, but now the header breaks when the page is too height.

socorro-search-broken

adngdb:
It worked everywhere but on the GC crashes page, because that page works differently than every other page. I've addressed the issue here: https://github.com/lonnen/socorro/pull/20

r+

lonnen:
Patch accepted. That was extremely helpful. Thanks!

PR #3452, opened 2016-08-31
fixes bug 1262252 - Crash data pure implementation

peterbe:
Again, sorry for the size of this. I'll try to break it down. 

### Implementations
It used to be that crash data got looked up via the middleware_app and ultimately it would be `socorro.external.crash_data_base.CrashDataBase` that would do the wrapping of the `S3BotoCrashStorage` to pull down raw dumps, raw crashes, processed crashes from S3. That class would also do too much. It would listen for boto StorageError exceptions and if that happens it would put the crash ID onto the priority jobs queue in rabbitmq.

What I've done is break that apart. And simplified. Instead of re-writing `socorro.external.crash_data_base.CrashDataBase` I created a fresh new class called `socorro.external.boto.crash_data.SimplifiedCrashData`. First of all, it's put in `boto` folder which is very explicit. Second, if this works out and goes into production I might rename it back to plain `CrashData` once I've cleaned up old legacy code. 

Then there's a another brand new class that deals specifically with putting jobs into the priority queue. That's all it does. 

[Deleting more old legacy code will be done later](https://bugzilla.mozilla.org/show_bug.cgi?id=1299465).

### Webapp
[Before it used be weird](https://github.com/mozilla/socorro/blob/344979da66b185fc61ba660e9848591bf6a1ef14/webapp-django/crashstats/crashstats/views.py#L1359-L1378). When loading the report index it would first try to load the [unredacted] processed crash *then* later try to load the raw crash. With some magical juggling of types of exceptions from the middleware it would either display a page that says "Crash Not Found", "Crash Too Old" or "Crash Pending, fetching it now". 

I just vastly simplified that. So **first** it loads the raw crash. If that fails (e.g. `CrashIDNotFound`) it plain and simply says "Crash Not Found". Secondly, when it loads the processed crash, if it's not found it puts the crash ID on the priority job AND returns a "Crash Pending" template (error code 200) and that would load some JavaScript code that checks `/api/ProcessedCrash/?crash_id=:crash_id` repeatedly until it exists and then reloads the page. 
[I hope to improve the "Crash Not Found" error page later](https://bugzilla.mozilla.org/show_bug.cgi?id=1299460).

## How to test it
Load up a report index page (using stage data). Go into the AWS management console and go to the `org.allizom.crash-stats.rhelmer` bucket and click into the [v1/processed_crash directory](https://console.aws.amazon.com/s3/home?region=us-west-2#&bucket=org.allizom.crash-stats.rhelmer-test.crashes&prefix=v1/processed_crash/). Look up the crash ID (that you loaded in the report index) and find it in AWS with the "Search by prefix" search. Once found, simply delete the processed crash json file. Then **restart** your webapp (to clear the LocMemCache) and refresh the report index URL you were on. (Make sure you have the Web Console open!) It should now load the "pending template" and it should just work :)

## The mocking
The major reason this PR is so HUGE is because `crashstats/crashstats/tests/test_views.py` does a lot of tests on top of `report_index`. Every single line of mocking (implementation directly instead of middleware JSON) had to be re-written. As I was doing that I refactored a couple of places where it had a big hardcoded python dict and use the `_SAMPLE_META` or `_SAMPLE_UNREDACED` dicts.


socorro/external/boto/crash_data.py
```diff
@@ -24,3 +28,53 @@ def get_storage(self):
...
+    makes this easier to work with from the webapp's model bridge.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(SimplifiedCrashData, self).__init__(*args, **kwargs)
+        # Forcibly set this to override the default in the base
+        # crash storage class for boto. We're confident that at this
+        # leaf point we want to NOT return a DotDict but just a plain
+        # python dict.
+        self.config.json_object_hook = dict
```
peterbe: Hopefully this can be removed once we get rid of the old CrashData classes and stuff.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -326,3 +327,31 @@ def reprocess(self, crash_ids):
             ):
                 success = False
         return success
+
```
peterbe: All the stuff above was just some random pep8 fixes.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -326,3 +327,31 @@ def reprocess(self, crash_ids):
...
+    required_config.rabbitmq_class = change_default(
+        RabbitMQCrashStorage,
+        'rabbitmq_class',
+        ConnectionContext,
+    )
+    required_config.add_option(
+        'routing_key',
+        default='socorro.priority',
+        doc='the name of the queue to recieve crashes',
+    )
```
peterbe: here, I tried to use that `change_default` function but it didn't work. It clashed with the `ReprocessOneRabbitMQCrashStore` class.

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -28,6 +28,9 @@
     Reprocessing,
```
peterbe: Almost all the changes in this file is for changing the mocking now that the crash_data comes from an implementation instead of the middleware (json).

webapp-django/crashstats/crashstats/tests/test_models.py
```diff
@@ -735,17 +735,15 @@ def mocked_get(url, params, **options):
         ok_(r['hits'])
```
peterbe: Almost all the changes in this file is for changing the mocking now that the crash_data comes from an implementation instead of the middleware (json).

I bet you can ignore everything in this file.

peterbe:
I removed ALL use of middleware in `integration-test.sh`. Bye bye!

socorro/external/boto/crashstorage.py
```diff
@@ -70,6 +69,12 @@ class BotoCrashStorage(CrashStorageBase):
         default='.dump',
         reference_value_from='resource.boto',
     )
+    required_config.add_option(
+        'json_object_hook',
+        default='configman.dotdict.DotDict',
+        from_string_converter=class_converter,
+        reference_value_from='resource.boto',
```
adngdb: I don't think that line should be there. This option is not strictly related to ``boto`` as a resource, it's related to this CrashStorage class.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -170,7 +171,7 @@ def new_crashes(self):
         # queues the crash_id. The '_consume_acknowledgement_queue' function
         # is run to send acknowledgments back to RabbitMQ
         self._consume_acknowledgement_queue()
-        conn = self.rabbitmq.connection()
+        self.rabbitmq.connection()  # XXX is this ever used and needed?
```
adngdb: Can you not just remove that line? It looks pretty useless, since it doesn't return anything. That would be a major design error of the ``rabbitmq`` library if it was needed.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -326,3 +327,31 @@ def reprocess(self, crash_ids):
...
+    required_config = Namespace()
+    required_config.rabbitmq_class = change_default(
+        RabbitMQCrashStorage,
+        'rabbitmq_class',
+        ConnectionContext,
+    )
+    required_config.add_option(
+        'routing_key',
+        default='socorro.priority',
+        doc='the name of the queue to recieve crashes',
```
adngdb: ``receive``

socorro/unittest/external/boto/test_crash_data.py
```diff
@@ -0,0 +1,156 @@
...
+            app_name='testapp',
+            app_version='1.0',
+            app_description='',
+            values_source_list=[values_source],
+            argv_source=[],
+        )
+
+        return config_manager.get_config()
+
+    def get_s3_store(self, storage_class=SimplifiedCrashData):
```
adngdb: As far as I can see, you are never overwriting ``storage_class`` when calling this method. I think you can just hard-code ``SimplifiedCrashData`` in here. That's what you're testing in this test case anyway. 

"Don't do things until they're needed. "
-- Adrian Gaudebert, 2016

scripts/integration-test.sh
```diff
@@ -131,7 +131,7 @@ function cleanup() {
   echo "INFO: Terminating background jobs"
   echo "  any kill usage errors below may be ignored"
 
-  for p in collector processor middleware
+  for p in collector processor
```
adngdb: We still use the middleware in production. I think we should keep this around until we can entirely remove it.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index_pending.html
```diff
@@ -1,32 +1,33 @@
...
-    </div>
-
-    <div id="fail" class="pendingStatus" style="display:none;">
-        <h3>Oh Noes!</h3>
-        <p>This archived report could not be located.</p>
+    <div class="pending">
+        <div class="searching">
+            <h3>Please wait...</h3>
+            <p>
+                Processing this crash only takes a couple of seconds.
```
adngdb: I would write ``Processing this crash report should only take a few seconds. `` instead. I find ``a couple of seconds`` to be slightly too accurate. :)

webapp-django/crashstats/crashstats/models.py
```diff
@@ -28,6 +31,7 @@
 import socorro.external.postgresql.product_build_types
 import socorro.external.postgresql.signature_first_date
 import socorro.external.postgresql.server_status
+import socorro.external.boto.crash_data
```
adngdb: Why do you import both this module and ``SimplifiedCrashData`` from that same module?

webapp-django/crashstats/crashstats/models.py
```diff
@@ -1239,11 +1256,17 @@ class RawCrash(SocorroMiddleware):
     )
 
     def get(self, **kwargs):
-        format = kwargs.get('format', 'meta')
-        if format == 'raw_crash':
-            format = kwargs['format'] = 'raw'
-        kwargs['expect_json'] = format != 'raw'
-        return super(RawCrash, self).get(**kwargs)
+        fmt = kwargs.get('format', 'meta')
```
adngdb: ``fmt`` confused me for a bit. How about ``format_`` or ``data_type`` or ``crash_format``?

webapp-django/crashstats/crashstats/models.py
```diff
@@ -1239,11 +1256,17 @@ class RawCrash(SocorroMiddleware):
...
+        if fmt == 'raw_crash':
+            # legacy
+            fmt = kwargs['format'] = 'raw'
+        expect_dict = fmt != 'raw'
+        result = super(RawCrash, self).get(**kwargs)
+        # This 'result', will either be a binary blob or a python dict.
+        # Unless kwargs['format']==raw, this has to be a python dict.
+        if expect_dict and not isinstance(result, dict):
+            raise BadArgumentError(fmt)
```
adngdb: This will show an error saying that the value of ``fmt`` is a field that has a bad value. Instead I think this should say ``raise BadArgumentError('format')``.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -1881,6 +1904,25 @@ def post(self, **data):
...
+
+    implementation_config_namespace = 'priority'
+
+    required_params = (
+        ('crash_ids', list),
+    )
+
+    get = None
+
+    def post(self, **data):
```
adngdb: Why using ``data`` instead of the usual ``kwargs``?

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/pending.js
```diff
@@ -1,47 +1,38 @@
+var Checker = (function() {
```
adngdb: This exposes ``Checker`` in the global namespace. If that's not what you deliberately intended, I guess it should be moved into a closure.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/pending.js
```diff
@@ -1,47 +1,38 @@
...
-            seconds = original_seconds;
-            $('#counter').html(original_seconds);
-            setTimeout("pendingReportTimer(\""+url+"\")",1000);
+    return {
+        startChecking: function(crashID) {
+            checkInterval = setInterval(function() {
+                $.get('/api/ProcessedCrash/', {crash_id: crashID})
+                .then(function() {
+                    clearInterval(checkInterval);
+                    // If it exists, we can reload the page we're one.
```
adngdb: ``s/one/on``

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/pending.js
```diff
@@ -1,47 +1,38 @@
...
+                    $('.pending .searching').hide();
+                    $('.pending .found').fadeIn(500, function() {
```
adngdb: I've read it's good practice to make animations 350ms at most. I'm actually using ``300`` in most places, so I'd recommend you use the same value here.

webapp-django/crashstats/crashstats/tests/test_models.py
```diff
@@ -1642,3 +1628,19 @@ def mocked_reprocess(crash_id):
...
+                return True
+            elif crash_id == 'bad-crash-id':
+                return
+            raise NotImplementedError(crash_id)
+
+        models.Priorityjob.implementation().process = mocked_process
+        ok_(api.post(crash_id='some-crash-id'))
+        # Note that it doesn't raise an error if
+        # the PriorityjobRabbitMQCrashStore choses NOT to queue it.
+        ok_(not api.post(crash_id='bad-crash-id'))
```
adngdb: I believe this test shows something is wrong, or is not testing anything. Your ``Priorityjob`` model requires the ``crash_ids`` argument, and it's missing in both cases here.

scripts/integration-test.sh
peterbe: The integration tests here only tested crash_data. If I remove crash_data from the middleware, there's no point keeping this in the tests because it never uses the middleware, because it can't work.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index_pending.html
peterbe: Good point!

webapp-django/crashstats/crashstats/models.py
peterbe: oops

webapp-django/crashstats/crashstats/models.py
peterbe: I renamed it to `format_`.

webapp-django/crashstats/crashstats/models.py
peterbe: thanks

webapp-django/crashstats/crashstats/tests/test_models.py
peterbe: I actually didn't matter because it was mocked.

peterbe:
All nits addressed.

socorro/external/boto/crashstorage.py
peterbe: What should it be then?

socorro/external/rabbitmq/crashstorage.py
peterbe: It makes me nervous. I don't exactly know how to test this in this context.

peterbe: Ok, to test it I ran my local processor, removed the line and processing still works. I see stackwalking happening just fine. 
When I put in a `raise Exception` here, it did raise which means that this code definitely is used in my local process. 

I'm confident I can delete it now.

socorro/unittest/external/boto/test_crash_data.py
peterbe: I totally agree with that principle!
What tripped me up was that it got confused with what `SimplifiedCrashData` was. There's the implementation class and this test class. So I renamed this test class to `TestSimplifiedCrashData`.

peterbe:
I believe I've addressed all review comments now.

socorro/external/boto/crashstorage.py
adngdb: It should just not be set. :)

peterbe: Done!

scripts/integration-test.sh
adngdb: Mmm, I see. Damn, that happened to me before and it's a pain. I guess the easiest thing to do is to just remove it indeed. But then we should hurry and remove all of the middleware asap.

webapp-django/crashstats/settings/base.py
```diff
@@ -715,6 +718,16 @@ def path(*dirs):
...
+        'boto': {
+            'access_key': config('resource.boto.access_key', ''),
+            'bucket_name': config(
+                'resource.boto.bucket_name', 'crashstats'),
+            'prefix': config('resource.boto.prefix', ''),
+            'keybuilder_class': config(
+                'resource.boto.keybuilder_class',
+                'socorro.external.boto.connection_context.DatePrefixKeyBuilder'
+            ),
+        }
```
adngdb: I'm just seeing that these config keys are different from every other one we have (the others are all caps). Why is that?

adngdb:
This is tested locally, it works well. I haven't been able to delete a document from S3 sadly, I guess I don't have the right permission, but I trust that you have tested it. 

r+ pending the configuration inconsistency.

webapp-django/crashstats/settings/base.py
peterbe: This is to take advantage of what's in consul under the `socorro/common` prefix. 
But now that you mention it, why didn't we do the same with the key for rabbitmq for example?!

My mind is a bit frayed with this and I'll like just deploy and test this first and foremost. Perhaps iterate thereafter.

peterbe:
OK. Let's merge this as soon as we've pushed a deploy to prod. Then I'll keep a keen eye on it and make sure it works as expected on stage.

webapp-django/crashstats/settings/base.py
adngdb: That doesn't really make sense, because you'll have to rewrite them to the django namespace anyway. I can see beauty in the consistency between namespaces, but I think we should favor consistency inside the namespace more. Everything in the django namespace is uppercase, we should stick to that. 

But, I'm not going to block merging this PR on that. :)

peterbe: Hmm... Now I'm getting tired but I thought my changes were so that nothing needs to be configured once this PR lands. ...because we use `socorro/common` as well as `socorro/webapp`.

PR #3454, opened 2016-09-01
fixes bug 1290967

lonnen:
Some linux crashes are crashing in RunWatchdog without calling it from or otherwise including the '::' operator in the frame, so these crashes are not triggering the appropriate processor rule.

This fixes it by removing the '::' requirement from the predicate. There are no Firefox crashes in super search at the time of this writing that will be adversely caught by this -- it should catch the same crashes as the previous rule plus the missing reports.

We may need to reprocess after this. Check the bug.


lonnen:
We probably want to reprocess at least the `[@ RunWatchdog ]` entries after this is out.

PR #3455, opened 2016-09-01
Fixes bug 1299769 - Ensute RunWatchdogRule re-generates signature usi…

adngdb:
…ng thread 0.


adngdb:
This is just going back to the previous state of things (plus some comments). 

r=me

PR #3456, opened 2016-09-06
1300367 remove server status

lonnen:
This fixes 1300367.

I got all the relevant references to `server_status` out of the repo. A few remain because we're re-using endpoints, or to maintain api names that are no longer backed by this data.

This is now passing but needs migrations to drop existing tables.


peterbe:
I can add the migration if you like.

peterbe:
Note-to-lonnen; either get alembic running locally or reach out to peterbe for help.

lonnen:
adrian landed my remove `scripts/staging` patch so I had to remove the `minidb` update commit from here (now deleted) and rebase from master. I also created and checked in the migration script.

lonnen:
This is ready for review. We'll need to run the migration after it lands.

peterbe:
🎆 
But stage is currently not updating on new commits. At the time of writing @jdotpz is looking into it. I dare not merge yet but will do when the coast is clear.

PR #3459, opened 2016-09-08
Bug 1296621 let polycrashstorage process cloned objects each time

peterbe:
The cool thing is that I now have a fully functional local processor. 
It is configured to run that S3TelemetryCrashStorage as its nth class and that means it runs AFTER the `ESCrashStorageRedactedJsonDump` class. 
Then I run the processor, and it does upload two json files. Both contain the full `json_dump` thing. Also, I load ES to see what was stored there:
screen shot 2016-09-07 at 5 33 35 pm
As you can see, it's the redacted `json_dump` key.

### DIFF
--- a/socorro/unittest/external/test_crashstorage_base.py
+++ b/socorro/unittest/external/test_crashstorage_base.py
@@ -52,6 +52,17 @@ class B(A):
                               )
 
 
+class MutatingProcessedCrashCrashStorage(CrashStorageBase):
+    def save_raw_and_processed(
+        self,
+        raw_crash,
+        dump,
+        processed_crash,
+        crash_id
+    ):
+        del processed_crash['foo']
+
+
 def fake_quit_check():
     return False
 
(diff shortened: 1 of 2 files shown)


socorro/unittest/external/test_crashstorage_base.py
```diff
@@ -270,6 +281,44 @@ def test_poly_crash_storage(self):
...
 
+    def test_poly_crash_storage_processed_crash_immutability(self):
+        n = Namespace()
+        n.add_option(
+          'storage',
+          default=PolyCrashStorage,
+        )
+        n.add_option(
+          'logger',
+          default=mock.Mock(),
```
adngdb: Wrong indentation here.

adngdb:
This is great! We will want to monitor the memory consumption of our stage processors to ensure we do not run into troubles when getting this to prod, though.

socorro/unittest/external/test_crashstorage_base.py
peterbe: All indentation fixed in that file.

peterbe:
As of this change, can we delete [this](https://github.com/peterbe/socorro/blob/beb3dc6b340c396609f88ff3efcfae1ca77f8230/socorro/external/es/crashstorage.py#L185-L189) now?

adngdb:
@peterbe yes!

PR #3460, opened 2016-09-08
fixes bug 1294088 - JSON Schema ambiguous types on certain keys

peterbe:
They're always strings. They're strings that contain numbers. 
See https://github.com/mozilla/socorro/blob/3877c90cfcbf3e12ca031321959a1964b0b9efe0/minidump-stackwalk/stackwalker.cc#L760-L761 and https://github.com/mozilla/socorro/blob/3877c90cfcbf3e12ca031321959a1964b0b9efe0/minidump-stackwalk/stackwalker.cc#L122


adngdb:
r+

PR #3461, opened 2016-09-09
nobug - make breakpad build idempotent

lonnen:
This enhances the breakpad script with some prechecks so that the breakpad make target will successfully run idempotently. Additionally it fixes some shellcheck warnings.


lonnen:
r=ted

the tests do not execute this code so I'm not waiting for test output

PR #3463, opened 2016-09-11
fix 1301926 -- remove hbase

lonnen:
This purges deprecated hbase classes and prunes the second order dead code. It also modifies documentation to use more representative classes without too much thought about the larger context (I probably corrected a few facts in otherwise mostly wrong pages).

It passes tests but will need manual verification on stage and prod to ensure I'm not stepping on any run time deps. I'm especially wary given [this comment](https://github.com/mozilla/socorro/compare/1301926-remove-hbase?expand=1#diff-1c662ce0ed6275baf5d2bdeeeea6c33eL16) and [this line](https://github.com/mozilla/socorro/compare/1301926-remove-hbase?expand=1#diff-1623cbaf81fefd3c37aa1297b41a1f06L168).

I'm also looking for pieces I may have missed that are not easy to find with grep, and second opinions on if this should happen at all.


peterbe:
I only started reviewing. The way hbase config is used is very much likely affecting how crash_data works. I would sleep much better at night if we **wait** with this PR until **after** https://github.com/mozilla/socorro/pull/3452 has landed and made it to at least stage. 
In other words, let's return to **this** PR when crash data (S3) is done directly from django.

willkg:
FYI, the config in prod that has "hb" in it (case insensitive check) looks like this:

```
socorro/middleware/hbase.hbase_class	socorro.external.boto.crashstorage.BotoS3CrashStorage
socorro/middleware/implementations.implementation_list	psql: socorro.external.postgresql, hbase: socorro.external.boto, es: socorro.external.es, fs: socorro.external.filesystem, http: socorro.external.http, rabbitmq:socorro.external.rabbitmq
socorro/middleware/implementations.service_overrides	CrashData: hbase, Correlations: http, CorrelationsSignatures: http, SuperSearch: es, Priorityjobs: rabbitmq, Search: es, Query: es
```

peterbe:
Yeah, that `socorro/middleware/implementations.service_overrides` thing is what scares me. 
That's why I want my PR landed first. Then we can breath easy.

lonnen:
@peterbe agreed. This should probably go through on its own.

@willkg -- thanks for checking that. We should do a similar check for stage. I modified those defaults, at least the implementation_list, in the patch.

From the comments it seems the hardcoded bits lars was worried about are in the middleware, so our front end testing on stage should catch it. We can manually verify by loading any individual report on stage.

PR #3465, opened 2016-09-15
Fixes bug 1302930 - Upgraded cryptography package to version 1.5.

adngdb:


peterbe:
Where is this package used? I'm pretty sure it was added as a dependency of something else. But I can't remember what.

rhelmer:
@peterbe `pipdeptree` is nice for showing this - in this case `cryptography` being pulled in by `ndg-httpsclient` originally added in bug 1102567.

adngdb:
I didn't test signing in after updating ``cryptography``, silly me. Good think njn had me set it up locally today, because there was a compatibility problem between ``pyOpenSSL`` and ``cryptography``. I've added a commit to update the former as well.

peterbe:
After installing this in *my* virtualenv logging in with Google **still** worked. 
I can't imagine any other thing this'd break. r+

...but, beyond waiting till we can merge things again. I think this **could** affect things like crontabber like the ftpscraper. So when we put this on stage, let's let it fester and run for a while to make sure all non-webapp aspects continue to work.

peterbe:
Let's work on https://github.com/mozilla/socorro/pull/3485 instead. It upgrades pyOpenSSL, crytopgraphy, ndg-httpsclient and pyasn1. Same as this PR but more. OK with that?

PR #3466, opened 2016-09-15
fixes bug 1302997 - Crash Not Found on every crash in stage

peterbe:


jdotpz:
r+ let's give it a run in stage

PR #3470, opened 2016-09-16
Fixes bug 1303250 - Use a JSON-serializable DotDict object.

adngdb:
The default value of `configman.dotdict.DotDict` broke processing, because configman DotDict cannot be JSON-serialized. The DotDict object from socorrolib is a very simple extension of the default dict type, so it is what we should use for our crash documents.

### DIFF
--- a/socorro/external/boto/crashstorage.py
+++ b/socorro/external/boto/crashstorage.py
@@ -71,7 +71,7 @@ class BotoCrashStorage(CrashStorageBase):
     )
     required_config.add_option(
         'json_object_hook',
-        default='configman.dotdict.DotDict',
+        default='socorrolib.lib.util.DotDict',
         from_string_converter=class_converter,
     )
 


adngdb:
Note that I haven't changed that config option on stage, so processing will not work there until this PR is merged. I know we have a stage-freeze window ongoing, so I am not making the call to merge or not. If you want to fix stage without merging this, apply the commands here: https://bugzilla.mozilla.org/show_bug.cgi?id=1303250#c2

peterbe:
r+ but waiting for @jdotpz until we can merge/push to stage again.

peterbe:
By the way, I've manually overwritten this by consulate on stage. So it's urgent to land this. But when we do, we have to run:
```
consulate kv rm socorro/processor/destination.storage1.json_object_hook
consulate kv rm socorro/processor/source.json_object_hook
```

peterbe:
Following up with removing the added consulate settings in https://bugzilla.mozilla.org/show_bug.cgi?id=1304493

PR #3475, opened 2016-09-19
fixes bug 1303853 - Error templates depend on request

peterbe:
Check the [bug description](https://bugzilla.mozilla.org/show_bug.cgi?id=1303853#c0).

This patch totally simplifies how error template rendering happens. Now crashstats_base.html is left to healthy requests. No more subtle errors during rendering the `500.html`. 

To test this, make sure you have `DEBUG=False` and `DEBUG_PROPAGATE_EXCEPTIONS=False` and to trigger the `DisallowedHost` I use my nginx but under a domain that Django was not allowing. See:
screen shot 2016-09-19 at 4 23 15 pm

I also checked that it continues to work for other errors:
Page Not Found:
screen shot 2016-09-19 at 4 23 57 pm

Bad Request:
screen shot 2016-09-19 at 4 24 25 pm
(note how I messed with the `date_end` parameter in the URL)

JSON errors still work as expected:
screen shot 2016-09-19 at 4 24 59 pm

SuperSearch errors still work:
screen shot 2016-09-19 at 4 26 03 pm

It also continues to work when I put in a temporary `raise NameError('test!')` in one of the view functions:
screen shot 2016-09-19 at 4 27 01 pm

### DIFF
--- a/webapp-django/crashstats/base/jinja2/500.html
+++ b/webapp-django/crashstats/base/jinja2/500.html
@@ -1,24 +1,19 @@
-{% extends "crashstats_base.html" %}
-
-{% block page_title %}
-Internal Server Error
-{% endblock %}
-
-{% block product_nav_filter %}&nbsp;{% endblock %}
-
+{% extends "error.html" %}
+{% block page_title %}Internal Server Error{% endblock %}</title>
 {% block content %}
 <div id="mainbody">
+    <div class="page-heading">
+        <h2>Internal Server Error</h2>
+    </div>
     <div class="panel">
         <div class="title">
-            <h2>Internal Server Error</h2>
+            <h2>Something went wrong, it is not you it's me.</h2>
         </div>
 
         <div class="body">
-            <p>Something went wrong, it is not you it's me</p>
-            <p>Please file an <a href="https://bugzilla.mozilla.org/enter_bug.cgi?{{ make_query_string(
+            <p>Please file an <a class="bugzilla-link" href="https://bugzilla.mozilla.org/enter_bug.cgi?{{ make_query_string(
                 product='Socorro',
                 component='Webapp',
-                bug_file_loc=request.build_absolute_uri(),
             ) }}">issue in Bugzilla</a> describing what happened, and please include the URL for this page.</p>
         </div>
     </div>
--- a/webapp-django/crashstats/base/views.py
+++ b/webapp-django/crashstats/base/views.py
@@ -1,5 +1,4 @@
 from django import http
-from django.conf import settings
 from django.shortcuts import render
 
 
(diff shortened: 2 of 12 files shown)


webapp-django/crashstats/base/jinja2/500.html
```diff
@@ -1,24 +1,19 @@
...
+            <h2>Something went wrong, it is not you it's me.</h2>
         </div>
 
         <div class="body">
-            <p>Something went wrong, it is not you it's me</p>
-            <p>Please file an <a href="https://bugzilla.mozilla.org/enter_bug.cgi?{{ make_query_string(
+            <p>Please file an <a class="bugzilla-link" href="https://bugzilla.mozilla.org/enter_bug.cgi?{{ make_query_string(
                 product='Socorro',
                 component='Webapp',
-                bug_file_loc=request.build_absolute_uri(),
```
adngdb: What is wrong with this? Why can you not use it, and need to rely on JS code instead? The document implies this is exactly what we are looking for, so I'm curious.

webapp-django/crashstats/base/views.py
```diff
@@ -27,5 +26,5 @@ def handler404(request):
             'path': request.path,
             'query_string': request.META.get('QUERY_STRING'),
         }, status=404)
-    context = {'product': settings.DEFAULT_PRODUCT}
+    context = {}
     return render(request, '404.html', context, status=404)
```
adngdb: You can rewrite this as ``return render(request, '404.html', status=404)`` (and same for the ``handler500`` view).

willkg: I'd leave this as is. The 404 and 500 views are different--the 404 view can take a context, but the 500 view is all "OH DESPAIR!" and can't take a context.

* https://docs.djangoproject.com/en/1.8/ref/views/#http-not-found-view
* https://docs.djangoproject.com/en/1.8/ref/views/#http-internal-server-error-view

Having said that, I have no idea if the socorro 404 handler/view/template uses the context at all. It probably does.

willkg: Er, let me clarify. The 500 stuff always takes an empty context. The 404 stuff takes a context. I have no idea if the socorro 404 stuff cares about having the product in the context or not. If not, then it's moot, I suppose.

webapp-django/crashstats/base/jinja2/500.html
peterbe: You can't use `request` in this template. Django goes into a strange mode when you've ended up here. If you try to use anything from `request` (which I think comes from the context processor) you get an error (within the error). And that's what I'm trying to avoid.

adngdb: Ah yes, of course, that makes sense. Thanks for the explanation! :)

PR #3478, opened 2016-09-22
Fixes bug 1281540 - Added a new datetime picker for Super Search and Signature report.

adngdb:
cc @lonnen 

Here's what it looks like:


lonnen:
@adngdb -- that looks prettier. How does it handle a smaller window? I don't think we need to be fully responsive, but it's common to split a screen in half with browser + terminal or similar. The current layout looks pretty good until you add rules, and then they wrap with no space.

screenshot 2016-09-21 17 50 10

lonnen:
Also, moment.js is used all over the webapp and this is a major version bump. The reviewer should check to make sure this will not break other functionality. It may impact [Bug 1304108](https://bugzilla.mozilla.org/show_bug.cgi?id=1304108)

adngdb:
@lonnen It looks okay on a smaller screen: 




The dynamic rules are not styled using flex, and I'd love to revisit their style but I thought it was slightly out of scope for this PR.

webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -108,6 +108,37 @@ def time_tag(dt, format='%a, %b %d %H:%M %Z', future=False):
     )
 
 
+@library.global_function
+def datetime_picker(input_name, default_value=''):
+    """Return a datetime picker HTML element to be powered by a JS library.
+    """
+    return jinja2.Markup('''
+        <span
+            class="datetime-picker {0}"
```
peterbe: It's a nit but I would use named parameters. E.g
```
"""Foo {name} and {bar} and {name} again""".format(name='name', bar='bar')
```

webapp-django/crashstats/signature/jinja2/signature/signature_report.html
```diff
@@ -26,12 +26,35 @@
...
+                    {{ datetime_picker('date_from', query.start_date) }}
+                </div>
+                <div>
+                    <label for="date_to">To</label>
+                    {{ datetime_picker('date_to', query.end_date) }}
+                </div>
+            </fieldset>
+
+            <p class="date-shortcuts">
+                Shortcuts, last
```
peterbe: I would say "Shortcuts: " here instead of the comma.

webapp-django/crashstats/supersearch/jinja2/supersearch/search.html
```diff
@@ -25,49 +25,64 @@
...
+            <p class="date-shortcuts">
+                Shortcuts, last
+                <a href="#" data-range="1d">24 hours</a> -
+                <a href="#" data-range="3d">3 days</a> -
+                <a href="#" data-range="7d">7 days</a> -
+                <a href="#" data-range="2w">14 days</a> -
+                <a href="#" data-range="1M">1 month</a> -
+                <a href="#" data-range="3M">3 months</a> -
+                <a href="#" data-range="6M">6 months</a>
+            </p>
```
peterbe: Since this snippet is exactly the same as in `signature_report.html` you should move this into a file and do something like `{% include "crashstats/date-shortcuts.html" %}`

webapp-django/crashstats/supersearch/static/supersearch/js/socorro/date_filters.js
```diff
@@ -0,0 +1,93 @@
...
+        // Limit filters based on the other filter's value.
+        filters.to.set('minDate', getDate('from'));
+        filters.from.set('maxDate', getDate('to'));
+
+        filters.from.config.onChange = function (dateObj) {
+            filters.to.set('minDate', dateObj);
+        };
+        filters.to.config.onChange = function (dateObj) {
+            filters.from.set('maxDate', dateObj);
+        };
```
peterbe: Nice!

adngdb:
@peterbe comments addressed!

peterbe:
When I load the page, the "7 days" shortcut does not have `class="selected"`. 
When I click one of the shortcuts it does set `class="selected"` but there's no CSS that does anything with this.

adngdb:
@peterbe I've addressed the missing bits, and while I was at it, I improved the search form UX (it now become visible only when it is fully loaded) and I integrated the loader work of #3484 here. 

@m8ttyB This might change the selector you want to use to get the loader, because I have moved it out of the ``#advanced-search`` element into the ``#search-form`` element. Basically, it is higher up in the DOM tree. This makes the whole form UX much better, and also means that the loader will disappear and be replaced by the entire form only when all of it is ready to be used.

m8ttyB:
Adrian,

Thank you for heads up, and yes, I'm excited for this change set. It will
likely make the tests more reliable. I'll update the selectors and include
it in the e2e-tests pr.

~ Matt

peterbe:
I first clicked on "3 days" then changed my dates manually and it still indicates that it's "3 days"
screen shot 2016-09-29 at 3 01 35 pm

adngdb:
@peterbe Good catch! I fixed that in the latest commit.

PR #3484, opened 2016-09-22
No Bug - Hide loader only when loading is done.

adngdb:
QA tests rely on the loader gif element to disappear to start running their tests. As of this, that loader will get removed only when the dynamic_form library is done with everything it has to do. That will prevent weird race conditions to happen, and hopefully will make the tests more reliable.

### DIFF
--- a/webapp-django/crashstats/supersearch/static/supersearch/js/lib/dynamic_form.js
+++ b/webapp-django/crashstats/supersearch/static/supersearch/js/lib/dynamic_form.js
@@ -26,7 +26,7 @@
         '__null__': 'does not exist',
         '!__null__': 'exists',
         '__true__': 'is true',
-        '!__true__': 'is false'
+        '!__true__': 'is false',
     };
 
     // Order matters here, the first operator will be used as the default
@@ -120,10 +120,11 @@
                     setParams(initialParams);
                 }
 
+                $('.loader', container).remove();
+
                 if (onReadyCallback) {
                     onReadyCallback();
                 }
-                $('.loader', container).remove();
             }
         );
 
@@ -323,9 +324,10 @@
          * Reset this form by removing all lines.
          */
         function reset() {
-            var line = false;
-            while (line = lines.pop()) {
+            var line = lines.pop();
+            while (line) {
                 line.remove();
+                line = lines.pop();
             }
         }
 
@@ -348,7 +350,7 @@
             var deleteOption = $('<a>', {
                 'class': 'dynamic-line-delete',
                 'href': '#',
-                'text': 'x'
+                'text': 'x',
             }).click(function (e) {
                 e.preventDefault();
                 this.remove();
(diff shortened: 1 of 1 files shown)


webapp-django/crashstats/supersearch/static/supersearch/js/lib/dynamic_form.js
```diff
@@ -124,6 +123,7 @@
                 if (onReadyCallback) {
                     onReadyCallback();
                 }
+                $('.loader', container).remove();
```
peterbe: I don't like this. It just doesn't feel right. This function calls "onReadyCallback()" but it's clearly not ready if it does more. Just the name implies that we're lying :)

It also doesn't feel right to do more in a function after you've "exited" by calling the callback. 

What could happens is this:

```javascript
showLoader()
fetch(...).then(r => {
    someCallback(r)
    hideLoader()
})
function someCallback(data) {
    showTable(data)
}
```
Your DOM will then be...
```
<!-- frame 1 -->
<div class="loading">Please wait</div>
```
```
<!-- frame 2 -->
<div class="loading">Please wait</div>
<table>
  data
</table>
```
```
<!-- frame 3 -->
<table>
  data
</table>
```

adngdb:
Alright, I move the removal of the loader to right before calling the callback. This would avoid that specific case you showed. 

Regarding using that loader for tests, it is, as far as selenium is concerned, the only sensible way to know that the form is ready to be used. And it is the correct thing to do, because that is also what users use to know they can interact with it. Also, this is a commit I made while talking to Matt about this very problem. :)

peterbe:
Why can't the selenium tests instead depend on the form being **present**? 
```javascript
var wait = setInterval(function() {
    if ($('form.myclass:visible').length) {
      clearInterval(wait);
      proceedCallback();
   }
}, 500);
```

peterbe:
Also, I don't think the cognitive process of the human is "the loader is gone, now I can type in what I want". I think it's more likely to be "the input field is there, now I can type in what I want". But that's honestly just a gut feeling. 

Anyway this is less important for the context of this PR. I'm just hoping @m8ttyB is joining in :)

peterbe:
r+ by the way. I don't know if stage is "free".

m8ttyB:
@peterbe The way that the tests are engineered is to look for an element on the page that signals that the dynamic elements in the page are loaded (enough) and the tests can begin to do their work. 

@adngdb and I met last Friday and focused in on using the loader as that keystone. I'm sure there are multiple routes to this end, but we focused in on this solution.

peterbe:
Yeah, I'm not trying to block this. Just saying ... "have you considered ..."

adngdb:
I'm actually closing this because I have integrated in #3478 as part of commit https://github.com/mozilla/socorro/pull/3478/commits/01798187eefe3d53580625c13e271cdb97e2f5e4.

PR #3485, opened 2016-09-23
Bug 1305129 upgrade pyopenssl pyasn1 and ndg httpsclient

peterbe:
This makes https://github.com/mozilla/socorro/pull/3465/files obsolete because I couldn't upgrade pyOpenSSL without also upgrading crytography.


peterbe:
By the way, after installing this update in my local virtualenv I tested that the google analytics also worked. I also testing google sign in. I did not test ftpscraper but it's extremely unlikely to be any different from the requests.post that the google analytics ping does.

adngdb:
``pyOpenSSL`` is already being updated in https://github.com/mozilla/socorro/pull/3465/files.

adngdb:
Actually, why don't you update ``cryptography`` here as well, and then we can get rid of #3465 ?

peterbe:
> Actually, why don't you update cryptography here as well, and then we can get rid of #3465 ? 

It does. This PR is what #3465 is but more.

PR #3490, opened 2016-09-27
Fixes bug 1266601 - No memory_report key when erroring out.

adngdb:


peterbe:
r+

socorro/processor/mozilla_transform_rules.py
```diff
@@ -410,10 +410,16 @@ def _extract_memory_info(self, dump_pathname, processor_notes):
     def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
         pathname = raw_dumps['memory_report']
         with temp_file_context(pathname):
-            processed_crash.memory_report = self._extract_memory_info(
+            memory_report = self._extract_memory_info(
                 dump_pathname=pathname,
                 processor_notes=processor_meta.processor_notes
             )
+
+            if isinstance(memory_report, dict) and memory_report.get('ERROR'):
```
peterbe: Perhaps it's not important but the return value of `self._extract_memory_info` is ALWAYS a dict. 
You could replace these lines with:
```python
assert isinstance(memory_report, dict), type(memory_report)
if memory_report.get('ERROR'):
    ...
```

peterbe: Or just...:
```python
if memory_report.get('ERROR'):
    ...
```

PR #3491, opened 2016-09-28
Fixes bug 1305616 - Stop storing memory_report into Elasticsearch.

adngdb:
@lonnen this is more of a political decision than a code one, so I'm asking you for review.

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -231,9 +231,12 @@ class ESCrashStorageRedactedJsonDump(ESCrashStorageRedactedSave):
     required_config.es_redactor.forbidden_keys = change_default(
         Redactor,
         "forbidden_keys",
-        "upload_file_minidump_flash1.json_dump, "
-        "upload_file_minidump_flash2.json_dump, "
-        "upload_file_minidump_browser.json_dump"
+        (
+            "memory_report, "
+            "upload_file_minidump_flash1.json_dump, "
+            "upload_file_minidump_flash2.json_dump, "
+            "upload_file_minidump_browser.json_dump"
+        )
     )
 
     #--------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


socorro/external/es/crashstorage.py
```diff
@@ -231,9 +231,12 @@ class ESCrashStorageRedactedJsonDump(ESCrashStorageRedactedSave):
     required_config.es_redactor.forbidden_keys = change_default(
         Redactor,
         "forbidden_keys",
-        "upload_file_minidump_flash1.json_dump, "
-        "upload_file_minidump_flash2.json_dump, "
-        "upload_file_minidump_browser.json_dump"
+        (
+            "memory_report, "
```
peterbe: If this is set in consulate will this new default not take effect?

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -464,6 +469,7 @@ def test_success_with_limited_json_dump_class(self, espy_mock):
             k: a_processed_crash['json_dump'][k]
             for k in modified_config.json_dump_whitelist_keys
         }
+        del expected_processed_crash['memory_report']
```
peterbe: I don't understand what the point of this is

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -225,7 +229,7 @@ def test_bulk_index_crash(self):
                 x,
                 a_processed_crash['uuid'][2:]
             )
-            # Ensure that the document was indexed by attempting to retreive it.
+            # Ensure that the document was indexed by attempting to retreive it
```
peterbe: Shouldn't there be a test that tries to retrieve it afterwards that checks that `memory_report` did NOT get stored.

socorro/unittest/external/es/test_crashstorage.py
adngdb: That is the answer to your previous question. It's how I verify that the ``memory_report`` has not been saved in the database. A few lines below there is a comparison that what has been sent to ES is the same as ``expected_processed_crash``.

peterbe: Ok. I think I get it.

PR #3493, opened 2016-09-28
fixes bug 1306048 - ability to connect to S3 by region for symbol upload

peterbe:
I know this is probably pretty hard to test but here's what I did...

1) I have this in my `.env`
```
# Personal
AWS_ACCESS_KEY=...secret...
AWS_SECRET_ACCESS_KEY=...secret...

SYMBOLS_BUCKET_DEFAULT_LOCATION=us-west-2
#SYMBOLS_BUCKET_DEFAULT_NAME=peterbe-crash-stats-symbols
SYMBOLS_BUCKET_DEFAULT_NAME=peterbe.crash.stats.symbols
```

2) Tried to upload a .zip file and got this error:
```
CertificateError: hostname 'peterbe.crash.stats.symbols.s3.amazonaws.com' doesn't match either of '*.s3.amazonaws.com', 's3.amazonaws.com'
```

3) Applied the patch and uploaded again now now it worked!
screen shot 2016-09-28 at 3 08 36 pm

Once this PR lands there shouldn't be anything we need to do in config.

### DIFF
--- a/webapp-django/crashstats/symbols/views.py
+++ b/webapp-django/crashstats/symbols/views.py
@@ -69,17 +69,11 @@ def unpack_and_upload(iterator, symbols_upload, bucket_name, bucket_location):
                 "Setting %s must be set" % key
             )
 
-    conn = boto.connect_s3(
-        settings.AWS_ACCESS_KEY,
-        settings.AWS_SECRET_ACCESS_KEY,
-        # Deliberately commented out until we know a better way to do
-        # this. When connecting to S3 on a Python 2.7 on OSX, you can't
-        # get buckets that dots in the name. But applying this calling_format
-        # thing breaks on our Python 2.7 on production.
-        # So it's commented out, in a rush, until we discover a unified
-        # way of dealing with this on local dev environments as well
-        # as in production.
-        # calling_format=boto.s3.connection.OrdinaryCallingFormat(),
+    conn = boto.s3.connect_to_region(
+        bucket_location,
+        aws_access_key_id=settings.AWS_ACCESS_KEY,
+        aws_secret_access_key=settings.AWS_SECRET_ACCESS_KEY,
+        calling_format=boto.s3.connection.OrdinaryCallingFormat(),
     )
     assert bucket_name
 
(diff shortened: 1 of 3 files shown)


webapp-django/crashstats/symbols/views.py
```diff
@@ -69,17 +69,11 @@ def unpack_and_upload(iterator, symbols_upload, bucket_name, bucket_location):
...
-        # thing breaks on our Python 2.7 on production.
-        # So it's commented out, in a rush, until we discover a unified
-        # way of dealing with this on local dev environments as well
-        # as in production.
-        # calling_format=boto.s3.connection.OrdinaryCallingFormat(),
+    conn = boto.s3.connect_to_region(
+        bucket_location,
+        aws_access_key_id=settings.AWS_ACCESS_KEY,
+        aws_secret_access_key=settings.AWS_SECRET_ACCESS_KEY,
+        calling_format=boto.s3.connection.OrdinaryCallingFormat(),
```
willkg: In `S3ConnectionContext` we additionally pass `is_secure=True`. Should we do that here, too?

willkg:
I didn't test this manually--I don't have the pieces running to do that.

The changes here are pretty minimal and mirror what we do elsewhere in Socorro. I think it'd be ok to land this and then test in -stage after it gets pushed. If that doesn't sit well with you (which is fine), maybe @adngdb is in a better position to test? He'll be around in an hour or so.

webapp-django/crashstats/symbols/views.py
peterbe: is_secure is true by default. It shouldn't be explictly set there in `S3ConnectionContext`.

willkg: It doesn't say that in the docs:

http://boto.readthedocs.io/en/latest/ref/s3.html#boto.s3.connect_to_region

How do you know it's true by default?

peterbe: I did a bunch of grepping on the boto site-packages tree.

PR #3497, opened 2016-09-30
fixes bug 1306730 - Upgrade cssselect and pyquery

peterbe:


willkg:
These are typically used by tests and nothing else. Assuming that's the case here, if Travis is happy, we're good. r+

PR #3499, opened 2016-10-03
Remove scripts/orphanSubmitter.py

willkg:
This script refers to paths that don't work anymore, so it's almost
assuredly obsolete. It's not referred to in the documentation or the
code anywhere I can find.

Therefore, time to remove it.

Quick r?


willkg:
I bumped into this while looking for something else. There is no `socorro.storage` anymore, so this is definitely no longer functional code.

PR #3503, opened 2016-10-05
Fixes bug 1281540 - Ensure dates are UTC in links to signature report.

adngdb:
@peterbe Alright, I've got this! ``moment`` is actually a pretty good library to deal with these timezone problems, it just took me some time to find the right functions to use. 

So now, if a date has no timezone, UTC is used. If a date has a timezone, that timezone is used to convert the date to UTC. That is dealt with in the ``datetime_filters`` component directly, when dealing with dates from the URL. 

I also kept the topcrashers's timezone-aware date change, because it's better, but it is not needed to fix the problem.


peterbe:
All the ways to test it, now works for me.

PR #3505, opened 2016-10-06
Fixes bug 1308320 - Add nsDependentString to the prefix signature list.

amccreight:


adngdb:
That will probably hit prod early next week. Thanks for contributing! :)

PR #3507, opened 2016-10-07
Add warning about how production RPM is broken

willkg:
We create the RPM on a build box using a local install of Python
2.7.11 so the socorro-virtualenv/ that gets created points to Python
libs that don't exist on a standard CentOS install.

This adds a warning about that and a link to the bug in question. This
is a stopgap until it gets fixed.

This is a docs change, so quick r?


willkg:
I'm going to self-r+ this. It's a fix to documentation that's broken, so even if it's wrong, it can't make it any more broken.

PR #3508, opened 2016-10-07
Fixes bug 1297966 - Show startup crash field in report index page

marco-c:


webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -252,6 +252,14 @@
 
                             {# Crash circumstances #}
 
+                            {% if raw.StartupCrash %}
+                            <tr title="{{ fields_desc['raw_crash.StartupCrash'] }}">
+                                <th scope="row">Startup Crash</th>
+                                <td>
+                                    <pre>True</pre>
```
peterbe: This is not going to work. 
You mentioned in IRC that in the raw crash this value might be `'0'`, `'1'` or non-existant. If it's `'0'` the line `{% if raw.StartupCrash %}` will be true. 

I think a better approach would be to handle these values differently. 

The raw crash JSON will probably always be a string (that is a number 0 or 1) and not a fully fledged JSON boolean. 
So, the line `{% if raw.StartupCrash %}` will be true every time the *key* is present in the raw crash. 

Then, when it comes to displaying the value we could have a jinja helper function do a string conversion to something like this:

```python
def display_booleanish(value):
if str(value).lower() in ('1', 'true', 'yes'):
    return "true"
else:
    return "false"
```

marco-c: Turns out the crash I used to test the absence of the row was a content crash, so the 'StartupCrash' annotation was missing in that case. I'll fix the patch.

marco-c:
I've verified that the patch works with three crash reports:
- `7f156e79-e4f4-451e-a68a-d992f2161007` is a content crash, so the annotation is absent and the row is not displayed;
- `c461963d-bf94-4603-b6e5-609e42161005` is a parent non-startup crash, so the annotation is '0' and the row is not displayed;
- `dfbc9aca-3eba-4325-b65c-3746f2161007` is a parent startup crash, so the annotation is '1' and the row is displayed.

marco-c:
@peterbe the annotation has been added to SuperSearch.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -252,6 +252,14 @@
 
                             {# Crash circumstances #}
 
+                            {% if booleanish_to_boolean(raw.StartupCrash) %}
```
peterbe: I feel like I'm messing around with you when I'm not. I'm just poorly communicating it. 
What I originally wanted to say is that isn't it better to AVOID the VALUE when determining whether to show this table row. 

So, the PRESENCE of the value should determine if the table row should appear. Then within, in the `<td>` cell it should show the string "True" or "False" depending on the value. 

E.g. if `raw.StartupCrash == "1"` it should become:
```html
<tr>
  <th>Startup Crash</th>
  <td>True</td>
</tr>
```

And if `raw.StartupCrash == "0"`
```html
<tr>
  <th>Startup Crash</th>
  <td>False</td>
</tr>
```

And if `raw.StartupCrash` does not exist at all, it should simply not be any table row at all. 

What I suggest you do is keep everything as is except change this `if` statement to simply be: `{if raw.StartupCrash %}`. 

Jinja is smart and does not fail or raise an error if the key doesn't exist. In normal Python you'd get an `AttributeError` if the object `raw` doesn't have an attribute called `StartupCrash`. 
And because the value is always a non-empty string (e.g. "0" or "1") the if statement will pass if it's set to one of those and *not* pass if the key doesn't exist or is an empty string.

peterbe:
Hmm... Perhaps we disagree on something fundamental. **Do we want show the table row if the value is "0"?** 
It's hard to decide. Sometimes it's useful to show what is false and sometimes it's not appropriate. Hard to decide. 

My [previous comment](https://github.com/mozilla/socorro/pull/3508#discussion_r83002721) was based on the assumption that you'd want to display it (True or False) simply based on it being existing. 

Perhaps what you want is that if it's non-existent or `"0"` then the row should not appear. 

It's hard to decide, so I will let you decide. You pioneered this change. 

**If** you want the row NOT to appear if the value is "0" you'll need to change your code to only call `booleanish_to_boolean` once. It's no point calling it when it's time to display the cell value. 

Does any of that make sense?

marco-c:
> Hmm... Perhaps we disagree on something fundamental. Do we want show the table row if the value is "0"?

Yes, my original idea was to only show it when it is a startup crash 😄 
Perhaps it's less confusing to always show it though, as you were assuming.
I'll change the patch to always show it, we can reconsider in the future.

PR #3511, opened 2016-10-10
Bug 1308824 - add vcruntime140.dll to skiplist

makotokato:
Like https://bugzilla.mozilla.org/show_bug.cgi?id=1308795, there is no vcruntime140.dll symbol into socorro now.  So we should add it to prefix skiplist.

When we used old vc, we add MSVCRT120.DLL to skiplist by https://bugzilla.mozilla.org/show_bug.cgi?id=1161067.

### DIFF
--- a/socorro/siglists/prefix_signature_re.txt
+++ b/socorro/siglists/prefix_signature_re.txt
@@ -187,6 +187,7 @@ strstr
 __swrite
 TlsGetValue
 TouchBadMemory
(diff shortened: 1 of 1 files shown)


adngdb:
Do you also want to add that dll to the Trim DLL Signatures list? See https://github.com/mozilla/socorro/tree/master/socorro/siglists#trim-dll-signatures for explanations.

PR #3515, opened 2016-10-11
Fixes bug 1297760: handle OSError when symlinking

willkg:
We have a problem with the -stage collector where we can get two
crashes with the same crash id. The first one will come in and get
saved--no problem.  The second one comes in and then trouble
starts. First, save_raw_crash() saves the crash by calling into
_save_files(). _save_files() happily stomps on any existing file, so
if there was a file already, it's no big deal.

Then save_raw_crash() goes to create two symlinks. Before this patch,
it would raise an OSError creating the name symlink because it created
one already with the first crash. That wouldn't get handled, so it'd
percolate up all the way and cause the collector to return an HTTP 500
to the client. Blech.

This patch changes the code to ignore the OSError. The better thing to
do would be for FSTemporaryStorage to realize the crash already exists
and stop trying to save it again, but that would require non-trivial
refactoring of the code. Instead, what this does is put both symlink
creating method calls into a try/except to ignore the OSError that
would result if we try to create the same symlink a second time.

A while back, someone had a similar error and wrote a test to handle
the case where one symlink is created, but not the other one. This
nixes that and replaces it with a test for making sure
save_raw_crash() doesn't raise an OSError due to symlinks.

### DIFF
--- a/socorro/external/fs/crashstorage.py
+++ b/socorro/external/fs/crashstorage.py
@@ -408,10 +408,18 @@ class FSDatedRadixTreeStorage(FSRadixTreeStorage):
                 #parent_dir)
 
         with using_umask(self.config.umask):
-            # Bug 971496 reversed the order of these calls so that the one that
-            # can fail will fail first and not leave an orphan symlink behind.
-            self._create_date_to_name_symlink(crash_id, slot)
-            self._create_name_to_date_symlink(crash_id, slot)
+            # If we've saved a crash with this crash_id before, then we'll kick
+            # up an OSError when creating the name symlink, but not the date
+            # symlink. Thus it's important to have these two tied together and
+            # try creating the name symlink first so if that fails we don't end
+            # up with an extra date symlink.
+            try:
+                self._create_date_to_name_symlink(crash_id, slot)
+                self._create_name_to_date_symlink(crash_id, slot)
+            except OSError:
+                # We pass here, because an OSError indicates the symlink
+                # already exists and we don't need to do anything about that.
+                pass
 
     def remove(self, crash_id):
         dated_path = os.path.realpath(
@@ -767,4 +775,3 @@ class TarFileSequentialReadingCrashStore(CrashStorageBase):
         reconstituted_processed_crash_as_str = result_gzip_fp.read().strip()
         processed_crash = json.loads(reconstituted_processed_crash_as_str)
         return processed_crash
-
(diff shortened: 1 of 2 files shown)


willkg:
This changes one of the unittests, so if the unittests pass and the code looks sane, then we're probably fine.

I'm pretty sure this doesn't have consequences for other parts of the code. It's unclear what the contract for `save_raw_crash()` is, but I'm pretty sure "don't raise an OSError" is a good idea.

If this change isn't good, we can approach this a different way and make a change in `BreakpadCollectorBase` and have it wrap the `save_raw_crash()` call in some kind of exception handling code. That's trickier to do generally because it has no idea what exceptions could get thrown and no idea which ones are ok and which ones are bad. Seems like we should be doing this in the crashstorage classes instead.

r?

willkg:
@peterbe, @adngdb: Can either of you review this? Landing this will make it easier to figure out the -stage submitter problems since it'll reduce the red herrings. Further, it increases confidence in the fact that our -stage system is working right from change to change.

socorro/external/fs/crashstorage.py
```diff
@@ -408,10 +408,18 @@ def save_raw_crash(self, raw_crash, dumps, crash_id):
...
-            self._create_date_to_name_symlink(crash_id, slot)
-            self._create_name_to_date_symlink(crash_id, slot)
+            # If we've saved a crash with this crash_id before, then we'll kick
+            # up an OSError when creating the name symlink, but not the date
+            # symlink. Thus it's important to have these two tied together and
+            # try creating the name symlink first so if that fails we don't end
+            # up with an extra date symlink.
+            try:
+                self._create_date_to_name_symlink(crash_id, slot)
+                self._create_name_to_date_symlink(crash_id, slot)
```
willkg: I don't think there's a good way to do that without creating a race condition.

socorro/external/fs/crashstorage.py
```diff
@@ -408,10 +408,18 @@ def save_raw_crash(self, raw_crash, dumps, crash_id):
...
+            except OSError:
+                # We pass here, because an OSError indicates the symlink
+                # already exists and we don't need to do anything about that.
```
willkg: In both of those situations, we'll see errors in other places like in `_save_files()` and logging etc.

willkg:
To answer the review summary comment, I think you're asking a lot of good questions. Testing permissions is good, but in this case, we're working with filesystem semantics and race conditions (as you pointed out). I don't think it's helpful to test twice--handling the exception is sufficient and requires less code overall.

I could add a `logger.info` line with the exception message so that the issue isn't completely swept under the carpet. I wouldn't bother spitting out a traceback because we know the call stack already as well as the type of exception being handled. I'll do that in a follow-up PR.

PR #3516, opened 2016-10-11
Fixes bug 1308618 - Always show form in Custom Search.

adngdb:


webapp-django/crashstats/supersearch/static/supersearch/js/socorro/search_custom.js
```diff
@@ -1,8 +1,12 @@
+/* globals ace */
+
 $(function () {
     'use strict';
 
     // parameters
     var form = $('#search-form form');
+    form.show(); // The form is hidden by default, show it immediatly.
```
peterbe: s/immediatly/immediately

peterbe: Why is the form hidden at all? I found https://github.com/mozilla/socorro/blob/master/webapp-django/crashstats/supersearch/static/supersearch/css/search.less#L7-L10

I find it kinda strange altogether. First, it's hidden, then it's displayed and then it's turned into an ACE form. I just don't get the point of all this hiding.

adngdb:
@peterbe All things addressed!

PR #3519, opened 2016-10-11
fixes bug 1297760: Adjust logging for OSError

willkg:
Peter pointed out it'd be better to have *any* logging, so I added
a logging.info line.

@peterbe: Quick r?

### DIFF
--- a/socorro/external/fs/crashstorage.py
+++ b/socorro/external/fs/crashstorage.py
@@ -416,10 +416,8 @@ class FSDatedRadixTreeStorage(FSRadixTreeStorage):
             try:
                 self._create_date_to_name_symlink(crash_id, slot)
                 self._create_name_to_date_symlink(crash_id, slot)
-            except OSError:
-                # We pass here, because an OSError indicates the symlink
-                # already exists and we don't need to do anything about that.
-                pass
+            except OSError as exc:
+                self.logger.info('failed to create symlink: %s', str(exc))
 
     def remove(self, crash_id):
         dated_path = os.path.realpath(


socorro/external/fs/crashstorage.py
```diff
@@ -416,10 +416,8 @@ def save_raw_crash(self, raw_crash, dumps, crash_id):
             try:
                 self._create_date_to_name_symlink(crash_id, slot)
                 self._create_name_to_date_symlink(crash_id, slot)
-            except OSError:
-                # We pass here, because an OSError indicates the symlink
-                # already exists and we don't need to do anything about that.
-                pass
+            except OSError as exc:
+                self.logger.info('failed to create symlink: %s', str(exc))
```
peterbe: Why not 
```python
self.logger.info('failed to create symlink', exc_info=True)
```
?
Also, why not `self.logger.warning` (or whatever it's called). Or even, if the exception is worth seeing, why not `self.logger.error`?

willkg: Info seemed prudent because it's an informational thing and not indicative of an error we need to deal with. I don't really have a strong opinion here, though.

I'm surprised you're suggesting we spit out a traceback here given you've been on a warpath about not including tracebacks in logging for things that aren't problems in other bugs. "The exception" is actually a bunch of different bits of information. Here, I'm adding the type and message because those seem helpful--particularly the message. The traceback isn't helpful in this case.

Maybe it'd help if you explained why you think these changes are interesting or important rather than just listing options?

willkg:
I'm going to merge this and verify it on -stage in a couple of hours. It's pretty benign.

PR #3520, opened 2016-10-11
bug 1309266 - API Tokens form going nuts

peterbe:
This might not solve it. But perhaps. 

It's not valid HTML to have an empty `action` attribute like that. Don't know if that's what's causing the browser to do weird things like submitting the form. Kinda reminds me of how it was dangerous to generate HTML like this `` because some browsers thought it would be a valid URL and crazy things could happen. 

Also, the form is always going to be there. It's never non-existent. If you're signed in, with or without any permissions, you can use the form to generate tokens.

### DIFF
--- a/webapp-django/crashstats/tokens/jinja2/tokens/home.html
+++ b/webapp-django/crashstats/tokens/jinja2/tokens/home.html
@@ -108,8 +108,7 @@
             <h2>Generate a New Token</h2>
         </div>
         <div class="body">
-            {% if form %}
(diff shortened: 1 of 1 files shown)


adngdb:
It's funny, the spec indeed says that ``action=""`` is invalid (because empty string is an invalid URL) but MDN shows examples with just that: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#Examples

There's nothing wrong with this PR, but I suspect that won't solve the problem. See the bug for more on that. :)

PR #3523, opened 2016-10-12
Fixes bug 1308478 - Enforce the timezone in stored date of date filters.

adngdb:


peterbe:
I can confirm that the steps mentioned in https://bugzilla.mozilla.org/show_bug.cgi?id=1308478#c0 are no longer reproducible. Yay!

peterbe:
True also for https://bugzilla.mozilla.org/show_bug.cgi?id=1308478#c1
Double Yay!

PR #3530, opened 2016-10-13
fixes bug 1309731 - link to Google Auth wiki page

peterbe:
No matter how I tried, it always looked so untidy and misaligned when the link was there in the right-hand lower corner. 
Considering that a large majority of people don't struggle with Google Sign-In, I think it's better to put it on the left with the other footer links.

### DIFF
--- a/webapp-django/crashstats/crashstats/context_processors.py
+++ b/webapp-django/crashstats/crashstats/context_processors.py
@@ -1,0 +1,7 @@
+from django.conf import settings
+
+
+def help_urls(request):
+    return {
+        'GOOGLE_AUTH_HELP_URL': settings.GOOGLE_AUTH_HELP_URL,
+    }
--- a/webapp-django/crashstats/authentication/jinja2/auth/login_failure.html
+++ b/webapp-django/crashstats/authentication/jinja2/auth/login_failure.html
@@ -6,6 +6,13 @@
 
 <h4>Unfortunately unable to sign you in.</h4>
 
+<p>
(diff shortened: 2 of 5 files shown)


webapp-django/crashstats/crashstats/context_processors.py
```diff
@@ -0,0 +1,7 @@
+from django.conf import settings
+
+
+def help_urls(request):
+    return {
+        'GOOGLE_AUTH_HELP_URL': settings.GOOGLE_AUTH_HELP_URL,
+    }
```
willkg: With the socorro webapp, do we add `settings` to the context for templates? If so, then we don't need this context processor.

peterbe: We do not. At least not in the list we control:
```python
_CONTEXT_PROCESSORS = (
    'django.contrib.auth.context_processors.auth',
    'django.core.context_processors.debug',
    'django.core.context_processors.media',
    'django.core.context_processors.request',
    'session_csrf.context_processor',
    'django.contrib.messages.context_processors.messages',
    'django.core.context_processors.request',
    '%s.authentication.context_processors.oauth2' % PROJECT_MODULE,
    '%s.base.context_processors.google_analytics' % PROJECT_MODULE,
    '%s.base.context_processors.browserid' % PROJECT_MODULE,
    '%s.status.context_processors.status_message' % PROJECT_MODULE,
    '%s.crashstats.context_processors.help_urls' % PROJECT_MODULE,
)
```

PR #3531, opened 2016-10-17
Fixes bug 1306891 - Integrate new correlations results in Socorro

marco-c:
I'd like to get some feedback before continuing.

Don't mind the usage of ES6 features, I'll remove them.


marco-c:
@adngdb ping, I'd just like some quick feedback here, not a full review.

adngdb:
I meant to do this today and got distracted, sorry. I'll get to it tomorrow!

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
```diff
@@ -97,4 +97,169 @@ $(document).ready(function () {
...
+        return loadChannelsData(product)
+        .then(() => {
+            if (signature in correlationData[product][channel]['signatures']) {
+                return;
+            }
+
+            return sha1(signature)
+            .then(sha1signature => {
+              return sha1signature;
+            })
```
adngdb: This ``then`` does nothing, right?

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
```diff
@@ -97,4 +97,169 @@ $(document).ready(function () {
...
+            .then(sha1signature => fetch(getDataURL(product) + channel + '/' + sha1signature + '.json.gz'))
+            .then(response => response.json())
+            .then(data => {
+                correlationData[product][channel]['signatures'][signature] = data;
+            });
```
adngdb: I think every Promise chain should end with, at least, ``.catch(console.log.bind(console));`` or any way to handle errors. Logging is the simplest way of handling them but I would be satisfied with it for a start.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
```diff
@@ -97,4 +97,169 @@ $(document).ready(function () {
...
+    }
+
+    function toPercentage(num) {
+        let result = (num * 100).toFixed(2);
+
+        if (result == '100.00') {
+          return '100.0';
+        }
+
+        if (result.substring(0, result.indexOf('.')).length == 1) {
```
adngdb: ``result`` being a number, that should raise a ``TypeError`` because (number).substring is not a function.

webapp-django/crashstats/signature/static/signature/js/signature_tab_correlations.js
```diff
@@ -182,6 +182,18 @@ SignatureReport.CorrelationsTab.prototype.onAjaxSuccess = function () {
                 ['core-counts', 'interesting-addons', 'interesting-modules']
             );
         });
+    }
+
+    function load() {
+        var channel = $('#correlations_channel').val();
+        var product = $('#correlations_product').val();
+        var signature = $('#correlations_signature').val();
```
adngdb: You don't need to use a hidden input, the signature is already accessible with ``SignatureReport.signature``.

webapp-django/crashstats/signature/views.py
```diff
@@ -372,6 +372,21 @@ def signature_correlations(request, params):
...
+        elif all('a2' in version for version in params['version']):
+            context['channel'] = 'aurora'
+        elif all('a1' in version for version in params['version']):
+            context['channel'] = 'nightly'
+
+    context['product'] = 'Firefox'
+    if 'product' in params and 'FennecAndroid' in params['product'] and 'Firefox' not in params['product']:
+            context['product'] = 'FennecAndroid'
+
+    context['signature'] = signature
```
adngdb: And thus you can remove this line as well.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
marco-c: Yes, I was probably debugging something here :)

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
marco-c: IIRC they are logged anyway in the console if you don't catch them, but I'll add the `catch` anyway.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
marco-c: `result` is a string (generated with `toFixed`)

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
```diff
@@ -97,4 +97,169 @@ $(document).ready(function () {
...
+          return '0' + result;
```
peterbe: This is all odd. First of all, the if statement can simply be done with `if (num < 0.01)`
but why display the number with a leading padded 0? If there's a good reason, there needs to be a comment somewhere here. 

Also, I find painful to use 2 decimals for all numbers except for `100.0`.
If the purpose of all of this is to make sure the string is always 5 characters, I'd much much rather solve the display part of this using CSS.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
```diff
@@ -97,4 +97,169 @@ $(document).ready(function () {
...
+    let standard_error = Math.sqrt(prop1 * (1 - prop1) / total1 + prop2 * (1 - prop2) / total2);
+    let ci = [diff - 1.96 * standard_error, diff + 1.96 * standard_error];
+
+    // Yates continuity correction for the confidence interval.
+    let correction = 0.5 * (1.0 / total1 + 1.0 / total2);
+
+    return [ci[0] - correction, ci[1] + correction];
+  }
+
+  function sortCorrelationData(correlationData, total_reference, total_group) {
```
peterbe: I can't skim this function and understand it. It sorts number *number* of keys in each objects' `item` attribute??? And then if equal it sorts by some sort of comparison of confident intervals.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
```diff
@@ -97,4 +97,169 @@ $(document).ready(function () {
...
+      textElem.text(sortCorrelationData(correlationData, total_reference, total_group)
+      .reduce((prev, cur) =>
+        prev + '(' + toPercentage(cur.count_group / total_group) + '% in signature vs ' + toPercentage(cur.count_reference / total_reference) + '% overall) ' + itemToLabel(cur.item) + '\n'
+      , ''));
+    });
+  }
+
+  window.correlations = {};
+  window.correlations.loadCorrelationData = loadCorrelationData;
+  window.correlations.writeResults = text;
```
peterbe: Why these three lines? I don't get it. 
And perhaps it wasn't planned. 

So you're defining all of these functions AFTER the DOM has been loaded. Why can't the functions be define BEFORE or INDEPENDENT of the loading of the DOM?

Use the module pattern or just encapsulate all of these functions as methods in a class.

webapp-django/crashstats/signature/jinja2/signature/signature_correlations.html
```diff
@@ -1,5 +1,29 @@
 <div id="correlations-wrapper" data-correlations-source="{{ url('crashstats:correlations_json') }}">
 
+    <label for="correlations_product">Product:</label>
+    <select id="correlations_product">
+        {% for p in ['Firefox', 'FennecAndroid'] %}
```
peterbe: Why hardcode these?

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
```diff
@@ -97,4 +97,169 @@ $(document).ready(function () {
...
+        // Join all the hex strings into one
+        return hexCodes.join('');
+    }
+
+    function sha1(str) {
+        return crypto.subtle.digest('SHA-1', new TextEncoder('utf-8').encode(str))
+        .then(hash => hex(hash));
+    }
+
+    let correlationData = {};
```
marco-c: I'm caching the results in this variable, so they don't have to be requested multiple times (e.g. if in the correlations tab the user switches to another product and then switches back).

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
marco-c: I'll add some comments.

It sorts the results by number of attributes first (e.g. platform=Windows will come before platform=Windows && graphic_card=NVIDIA), because they are often the most interesting results.

Then, it sorts by percentage difference (e.g. if platform=Windows is 100% in the signature and 60% overall and graph_card=NVIDIA is 40% in the signature and 25% overall, platform=Windows will come first).

webapp-django/crashstats/signature/jinja2/signature/signature_correlations.html
marco-c: For now, the correlations are only generated for these two products.

marco-c:
> So, in conclusion test by running ./manage.py collectstatic --noinput then run with DEBUG=False in your .env and see if things still work.

I've had to remove arrow functions, `let` and `for ... of`.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
marco-c: I put the code in there just temporarily, for testing.

marco-c:
In another PR, I'll add the correlations results to the top crashers page, replacing the current ones.

adngdb:
@marco-c please don't rebase your commits before reviewing is finished, it makes my job a lot harder! Now I can't see what you have changed since the last time I reviewed. :(

Also, I thought you wanted to keep the old correlations around because modules and other stuff where missing from yours? I don't mind that you remove the current correlations, they are doomed to die anyway, but I guess I'm just confused.

marco-c:
> @marco-c please don't rebase your commits before reviewing is finished, it makes my job a lot harder! Now I can't see what you have changed since the last time I reviewed. :(

Oops, sorry. I thought at the beginning you only had a quick look. In general, most of the code is the same, I've just incorporated your feedback and Peter's, and I've cleaned up the old correlation stuff.

> Also, I thought you wanted to keep the old correlations around because modules and other stuff where missing from yours? I don't mind that you remove the current correlations, they are doomed to die anyway, but I guess I'm just confused. 

Talking with Peter, we concluded that we could remove the old correlation stuff because basically nobody is using it, it doesn't work most of the time, and the reports are still accessible anyway at crash-analysis.mozilla.org. Only the modules are missing for now, but hopefully they will be included soon.

adngdb:
I just did quite big review written down as code: https://github.com/marco-c/socorro/pull/1 :)

adngdb:
@marco-c Do you think this is ready to go? I haven't tested them but your last changes look good!

marco-c:
Yes, I think it's ready.
I've tested some signature pages, a report page for a signature with correlations, and a report page for a signature without correlations. All of them seem to be working correctly.

adngdb:
This is the only concern I have after testing this new feature: https://bugzilla.mozilla.org/show_bug.cgi?id=1315109

It can be a follow-up improvement. These new correlations rock! Let's ship them! 

Thanks for your impressive work @marco-c !!!

marco-c:
> This is the only concern I have after testing this new feature: https://bugzilla.mozilla.org/show_bug.cgi?id=1315109

Makes perfect sense. Sometimes signatures might have the same causes in Thunderbird and Firefox, so checking the correlations with the Firefox product might be useful, but it's better to avoid confusion.

Thank you for the reviews and the fixes!

PR #3534, opened 2016-10-17
Fixes bug 1305016 - Added the number of distinct install times to Top…

adngdb:
… Crashers.


philipp-sumo:
would that be something that may fairly easily be incorporated into super search as well (sorting/searching by install count)?

adngdb:
@philipp-sumo I'm much more reluctant to adding this kind of features to Super Search. Topcrashers is a pretty static page, the list of columns is controlled by us. Super Search is much more dynamic, so every little hack we add makes it more fragile. 

Ideally, I would like to make Super Search able to construct deeper aggregations, like the API can do. But that's not in my priorities right now sadly.

PR #3544, opened 2016-10-20
bug 1311410 - raise nginx file upload limit

peterbe:


peterbe:
Actually, what does this file mean? It's called `socorro-webapp.conf.sample`. "sample"? Is it not the real one we use?

peterbe:
I suspect this file is not "live". It's just a sample for when it gets created in socorro-infra. 
Seems https://github.com/mozilla/socorro-infra/pull/259 is more likely to work.

PR #3545, opened 2016-10-20
Up max body size in nginx for symbol uploads resolves bug 1311410

jdotpz:
TRIGGER


peterbe:
Use https://github.com/mozilla/socorro/pull/3544 instead

PR #3546, opened 2016-10-20
fixes bug 1311697 - deep copy SocorroDotDict better

peterbe:
I feel like I totally stressed through this. Lots of things going on today. 
This PR adds a test that does break unless you apply the fix. 
Also, this change does make the whole processor work. I've tested it by making the the crash gets sent in correctly (to my personal AWS S3 bucket). 

Let's get this to prod soon.

### DIFF
--- a/socorro/external/crashstorage_base.py
+++ b/socorro/external/crashstorage_base.py
@@ -20,6 +20,25 @@ from configman.converters import classes_in_namespaces_converter, \
 from configman.dotdict import DotDict as ConfigmanDotDict
 
 
+def socorrodotdict_to_dict(sdotdict):
+    """Takes a socorrolib.lib.util.DotDict and returns a dict
+
+    This does a complete object traversal converting all instances of the
+    things named DotDict to dict so it's deep-copyable.
+
+    """
+    def _dictify(thing):
+        if isinstance(thing, collections.Mapping):
+            return dict([(key, _dictify(val)) for key, val in thing.items()])
+        elif isinstance(thing, basestring):
+            return thing
+        elif isinstance(thing, collections.Sequence):
+            return [_dictify(item) for item in thing]
+        return thing
+
+    return _dictify(sdotdict)
+
+
 #==============================================================================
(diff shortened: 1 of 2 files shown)


socorro/external/crashstorage_base.py
```diff
@@ -603,7 +603,16 @@ def make_deep_copy(crash):
...
+            def dictify(thing):
+                for key, value in thing.items():
+                    if isinstance(value, SocorroDotDict):
+                        thing[key] = dict(value)
+                    if isinstance(value, collections.Mapping):
+                        dictify(value)
+
+            dictified = dict(crash)
+            dictify(dictified)
+            return SocorroDotDict(copy.deepcopy(dictified))
```
willkg: Why not extract dictify and then test it separately with unit tests?

willkg:
I'll try fixing my issue here and pushing a commit.

willkg:
^^^ I pushed stuff into your branch. That's the most amazing thing ever.

willkg:
I have no idea what we do with reviews that have two authors. Do we need a third separate pair of eyes? Can we review each others' stuff? Does @lonnen have to come down from his throne of skulls to bless us with new powers?

adngdb:
@willkg the ``dictify`` function can be removed now, and replaced with a call to ``deepcopy``. Do you want to do that? I can push to Peter's branch as well, and then this PRs becomes like a big Socorro party!

willkg:
@adngdb I don't understand what you mean by that. How can `dictify` or the new `socorrodotdict_to_dict` be replaced with a deepcopy?

willkg:
@adngdb Oh, my bad--I didn't finish. My sincerest apologies. I'll fix it now. Bad naughty me.

willkg:
@adngdb I think that fixes it. The tests pass locally. If you see any issues, you should push a commit, too. Then we can all spend tomorrow figuring out what the new process should be.

🎉 🎈 🍰

peterbe:
I'll merge this once it works running as my local processor.

PR #3548, opened 2016-10-21
bug 1305807 - move google analytics to not be inline

peterbe:
I don't know how much you know about Google Analytics. 
I had to do a bunch of research myself. Perhaps it's good if you have to do some research of your own to check that I'm doing the right thing. 

The most important change here is that the GA ID is now super hardcoded. 
Yes, it spells trouble for other companies that tries to use our webapp, but does that even exist any more? And also, GA will filter out any traffic that comes from domains not recognized in GA Admin. 

I did the same for Air Mozilla (which uses the non-360 GA still), because it's just not realistic that someone will run that site under a different domain. https://github.com/mozilla/airmozilla/blob/master/airmozilla/base/static/js/google-analytics.js#L2

### DIFF
--- a/webapp-django/crashstats/base/jinja2/crashstats_base.html
+++ b/webapp-django/crashstats/base/jinja2/crashstats_base.html
@@ -8,17 +8,16 @@
         {% block site_css %}
             {% stylesheet 'crashstats_base' %}
         {% endblock %}
-    {% if GOOGLE_ANALYTICS_ID %}
-        <script type="text/javascript">
-          (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-          (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-          m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-          })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
 
-          ga('create', '{{ GOOGLE_ANALYTICS_ID }}', '{{ GOOGLE_ANALYTICS_DOMAIN }}');
-          ga('send', 'pageview');
-        </script>
-    {% endif %}
+        {% if not DEBUG %}
+        {#
+            Use the preload technique which gives a performance boost for
+            modern browsers.
+            See https://developers.google.com/analytics/devguides/collection/analyticsjs/#alternative_async_tracking_snippet
+        #}
+        {% javascript 'google_analytics' %}
+        <script async src="https://www.google-analytics.com/analytics.js"></script>
+        {% endif %}
 
     <meta name="google-signin-client_id" content="{{ OAUTH2_CLIENT_ID }}">
     {% if OAUTH2_SIGNOUT %}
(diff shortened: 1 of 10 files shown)


adngdb:
I don't feel good about this. Hard-coding the GA tracking number is conceptually something that I think we should avoid. Plus that new code looks more complex than the current one to me. 

I also don't see the benefits here. My first reaction when seeing this was "you are being too religious about the removal of inline JS". I fail to see the security risks that this change avoids? The only values we load in there are the GA domain and ID, which both come from our settings. 

We can have exceptions to the "no-inline-js" rule and this is a good candidate to me. 

But, I might be completely wrong about this. If you think I am, please educate me! :)

peterbe:
The crusade for getting rid of all inline JS is to make the site secure. One of the most common forms of attacks are almost always avoided by having good Content Security Policies (CSP). April King told me this when I made Airmozilla secure with Observatory. 

You can't have a tight CSP if you have inline JS. It might be possible to have inline `script` if you set a hash nonce on it but then, equally, that has to be generated depending on the ID. 

By the way, the new way of setting the script I found out from reading the Google Analytics documentation. This new way has the advantage that the browser can prefetch the URL whilst loading the page. Even though it's set to load async. The async load just means it won't get executed until after the DOM has loaded. The only way, the browser didn't know that this URL was going to be needed until after the script snippet has been parsed and executed which can happen a lot later. 

The goal is to be 100% free of inline `style` and inline `script` tags and then set a Script CSP policy that disallows all inline script execution. It means, for example, that if we have an XSS leak, a hacker can't cause it to be triggered within the DOM. So suppose we have something like this:
```html
<tr>
  <th>Sorting by {{ _column | some_jinja_filter }}</th>
</tr>
```
And there's a bug in `some_jinja_filter` that doesn't escape properly, then the hacker would be able to maybe do `?_column=date<script>alert()</script>` or something. If we have a CSP that does NOT allow inline scripts, that execution would be stopped.

peterbe:
Let's do with a django-pipeline compiler
https://django-pipeline.readthedocs.io/en/latest/compilers.html#rd-party-compilers

peterbe:
I have to reopen otherwise I can't rebase and reopen later.

webapp-django/crashstats/settings/base.py
```diff
@@ -599,10 +600,11 @@ def path(*dirs):
         'release': SOCORRO_REVISION,
     }
 
-# Optional Google Analytics ID (UA-XXXXX-X)
+# Google Analytics ID (UA-XXXXX-X) used for sending
+# pings to Google Analytics about usage of the public API.
+# Note that the Mozilla Google Analytics ID is hardcoded
+# in the static file: google_analytics.js
```
willkg: This comment is out of date now since we're compiling it.

webapp-django/crashstats/base/jinja2/crashstats_base.html
```diff
@@ -8,17 +8,16 @@
...
-    {% endif %}
+
+        {% if not DEBUG %}
+        {#
+            Use the preload technique which gives a performance boost for
+            modern browsers.
+            See https://developers.google.com/analytics/devguides/collection/analyticsjs/#alternative_async_tracking_snippet
+        #}
+        {% javascript 'google_analytics' %}
+        <script async src="https://www.google-analytics.com/analytics.js"></script>
```
willkg: Nit: I think this block should get indented 4 spaces.

PR #3549, opened 2016-10-24
fixes bug 1237704 - django-pipeline 1.6.9

peterbe:
Nothing actually changes. There are [some changes](https://pypi.python.org/pypi/django-pipeline/#history) since 1.5 but nothing major that I can see affects us.


webapp-django/crashstats/settings/base.py
```diff
@@ -525,24 +525,36 @@ def path(*dirs):
...
+        'UGLIFYJS_BINARY',
+        path('node_modules/.bin/uglifyjs')
+    ),
+    'UGLIFYJS_ARGUMENTS': '--mangle',
+    'CSS_COMPRESSOR': 'pipeline.compressors.cssmin.CSSMinCompressor',
+    'CSSMIN_BINARY': config(
+        'CSSMIN_BINARY',
+        path('node_modules/.bin/cssmin')
+    ),
+    # Don't wrap javascript code in... `(...code...)();`
```
adngdb: You mean a closure? :)

PR #3553, opened 2016-10-24
bug 1231783 - Send all processor errors to sentry

peterbe:
I've tested this quite extensively. E.g.
https://sentry.prod.mozaws.net/operations/socorro-stage/issues/362120/
https://sentry.prod.mozaws.net/operations/socorro-stage/issues/362112/

The crucial part is the new lines inside `processor_app.py`. 
The most important thing is how the exception is juggled. 

### First key learning...

If anything goes wrong in any of the crashstoage classes, the processor collects all exceptions into an instance of a `PolyStorageError`. That's why it's important to loop over over this. It might be more than 1 exception. If you don't do this iteration **iff** it's a an error inside one of the crashstorage classes you get errors in Sentry like this: https://sentry.prod.mozaws.net/operations/socorro-stage/issues/362043/ (scroll down and look at the "Exception".

And because it might NOT be a `PolyStorageError` you'll see that 
```python
if isinstance(exception, PolyStorageError):
    # Then it's already an iterable!
    exceptions = exception
else:
    exceptions = [exception]
```

### Second key learning...

Python is tricky when it comes to doing `try:except` INSIDE an exception frame. 
Consider: https://gist.github.com/peterbe/c39548a3eec97634e7d2de28fa768f11
What's wrong with that gist is the Traceback points to "line 8" which is not where the error happened. It happened on line 2.

The only way to get this to report correctly is to capture the `sys.exc_info()` as local variable(s) and then at the bottom of the parent exception you raise it like this `raise exc_type, exc_value, exc_tb`. 

See for the working solution:  https://gist.github.com/peterbe/007c7acef30ff8bb0ba90e9a92350bae

####  Lastly...
This PR is two commits. First one with the juicy change. Second one with pep8 cleanups. You might want to click on the first commit in PR to avoid the pep8 cleanup noise. 
Also, this PR is not automatically closing the bug. That's because I want to configure sentry and test that it works first.

### DIFF
--- a/socorro/processor/processor_app.py
+++ b/socorro/processor/processor_app.py
@@ -57,15 +57,15 @@ class ProcessorApp(FetchTransformSaveWithSeparateNewCrashSourceApp):
         'companion_class',
         doc='a classname that runs a process in parallel with the processor',
         default='',
-        #default='socorro.processor.symbol_cache_manager.SymbolLRUCacheManager',
+        # default='socorro.processor.symbol_cache_manager.SymbolLRUCacheManager',
         from_string_converter=class_converter
     )
 
     ###########################################################################
-    ### TODO: implement an __init__ and a waiting func.  The waiting func
-    ### will take registrations of periodic things to do over some time
-    ### interval.  the first periodic thing is the rereading of the
-    ### signature generation stuff from the database.
+    # TODO: implement an __init__ and a waiting func.  The waiting func
+    # will take registrations of periodic things to do over some time
+    # interval.  the first periodic thing is the rereading of the
+    # signature generation stuff from the database.
     ###########################################################################
 
     required_config.namespace('sentry')
--- a/socorro/processor/processor_app.py
+++ b/socorro/processor/processor_app.py
@@ -6,7 +6,9 @@
 """the processor_app converts raw_crashes into processed_crashes"""
 
 import os
+import sys
 
+import raven
 from configman import Namespace
 from configman.converters import class_converter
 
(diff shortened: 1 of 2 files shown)


socorro/processor/processor_app.py
```diff
@@ -131,6 +142,45 @@ def _transform(self, crash_id):
...
+        except Exception as exception:
+            # Immediately capture this as local variables.
+            # During this error handling we're going to be using other
+            # try:except: constructs (e.g. swallowing raven send errors)
+            # so we can't reference `sys.exc_info()` later.
+            exc_type, exc_value, exc_tb = sys.exc_info()
+
+            if self.config.sentry and self.config.sentry.dsn:
+                try:
+                    if isinstance(exception, PolyStorageError):
```
adngdb: I think you can make this more generic, and not rely on ``PolyStorageError``. Maybe use ``collections.Sequence`` instead? That way you catch most iterables. 

```python
In [9]: isinstance([], collections.Sequence)
Out[9]: True

In [10]: isinstance(Exception('test'), collections.Sequence)
Out[10]: False
```

Note that you can't use ``iter`` because ``Exception`` is an iterable.

peterbe: Hmm... But you make a good point. If we use `collections.Sequence` instead, we might not need that explicit import from the `crashstorage_base.py`. That'd be nice.

peterbe:
For the record, I'm going to update the code and use `collections.Sequence` instead (to avoid the import). But I don't want to land this on prod today. That's because I want it to sit on stage a couple of hours to see what kind of flood it might open. 
(Later today we're going to enable the new crashstorage classes)

PR #3555, opened 2016-10-25
fixes bug 1312725 - Upgrade to google-api-python-client==1.5.4

peterbe:
The only way I know to test this is to simply sign in.


peterbe:
For the record, it worked for me on my mac.

PR #3557, opened 2016-10-25
fixes bug 1312727 - Upgrade to isoweek 1.3.1

peterbe:


peterbe:
The only place where this is used is in Crashes per Day and we have good test coverage of that. 
Also, I loaded this alongside stage and the output looks exactly the same.

PR #3558, opened 2016-10-26
Fixes bug 1311527 - Added a safety margin for date checking in Super …

adngdb:
…Search.

### DIFF
--- a/socorro/middleware/search_common.py
+++ b/socorro/middleware/search_common.py
@@ -278,7 +278,8 @@ class SearchBase(object):
                         msg='date must have a prefix operator'
                     )
                 today = datetimeutil.utc_now()
-                if param.value > today:
+                # We add a safety margin to not be too strict.
+                if param.value > today + datetime.timedelta(days=1):
                     raise BadArgumentError(
                         'date',
                         msg="date can't be in the future"
--- a/socorro/unittest/middleware/test_search_common.py
+++ b/socorro/unittest/middleware/test_search_common.py
@@ -228,10 +228,16 @@ class TestSearchBase(TestCase):
                 config=config,
             )
 
-        # the date parameter must always have a prefix operator
-        tomorrow = (
-            datetime.datetime.utcnow() + datetime.timedelta(days=1)
-        ).strftime('%Y-%m-%d')
+        # Test the safety margin.
+        in_a_few_hours = '<' + (
+            datetime.datetime.utcnow() + datetime.timedelta(hours=12)
+        ).isoformat()
+        search.get_parameters(date=in_a_few_hours)
+
+        # Test a date too far in the future.
+        tomorrow = '<' + (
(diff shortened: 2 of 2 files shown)


socorro/middleware/search_common.py
```diff
@@ -278,7 +278,8 @@ def fix_date_parameter(self, parameters):
                         msg='date must have a prefix operator'
                     )
                 today = datetimeutil.utc_now()
-                if param.value > today:
+                # We add a safety margin to not be too strict.
+                if param.value > today + datetime.timedelta(days=1):
```
peterbe: Why a whole day? Why not one minute. 
The granularity of the datepickers on SuperSearch is minute. I just tested it and it correctly set the time to the current hour and minute in UTC. 
Not possible to reproduce. 
If your computer's clock is a couple of seconds *ahead* it could happen that the minute number is 1 minute greater than the UTC standard time. So, it would be safe to just add the margin of 1 minute. 

If someone has a computer whose time is **more** than 60 seconds too fast, that would be super rare. 

The reason we have these validation guards is to that there aren't ES errors when trying to query an index that doesn't exist *yet*. If someone has a crazy date on their computer, and we allow it in the validation, they could easily have a hard 500 Internal Server Error because we queried an index based on a future date. 

So, in conclusion:
* Why not just 1 minute?
* Why not generate that default date from the server instead of letting `new Date()` in JS do it?

peterbe: uh? now I'm royally confused. 
So SuperSearch allows future dates? How can that be? We've had so many errors because it fails hard when the index doesn't exist because the user "accidentally" typed in a future date. So, is it that it allows it differently depending on how far in the future it is?

So how does Wayne get his errors? I've never seen them myself and I don't even know how to fake my OS's clock so I can't pretend "now" is in an hours time. 

I get a feeling that ALL of this can be avoided if avoid relying on the client to set the date. E.g. render the `<input name="start_date" value="{{ server_now_date.isodate() }}">` or something.

adngdb: SuperSearch has always been able to handle missing indices. I believe the reason why we put that in place was because there was a problem when dates where incoherent, like if a date range was exclusive (for example: ``> 2016-10-01 AND < 2016-09-30``).

Wayne's error does not come from Elasticsearch, it comes from our validation process. I believe his clock is off by a few seconds and that triggers the "no future" condition. You don't need to fake your clock to reproduce that. In the form, just enter a ``to`` date that is slightly ahead of the current time.

It is not possible to rely on a server-side date for this. Or rather, it is, but that would be very inconvenient, and would lead to weirdness (if you leave the page open for a day, come back to it, and click the shortcuts to make the date up-to-date, you would get yesterday's datetime). 

I am now convinced that we should remove the "no future" validation step from Super Search, and keep the rest (make sure the date range is coherent). I'm closing this and will work on something different.

PR #3560, opened 2016-10-27
Fixes bug 1305129: GA ssl handshake fix

willkg:
Certificate verification is failing. We debated several options (details
in the bug) and then decided to set verify_ssl=False for this specific
call. Because we're doing that, we nixed all possible PII, too.

r?

### DIFF
--- a/webapp-django/crashstats/base/tests/test_ga.py
+++ b/webapp-django/crashstats/base/tests/test_ga.py
@@ -1,3 +1,6 @@
+from functools import wraps
+from uuid import UUID
+
 import mock
 from nose.tools import eq_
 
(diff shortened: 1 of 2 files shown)


willkg:
@peter: Can you review this? We can test it out on -stage.

willkg:
I'll fix those errors now. Somehow I screwed up testing it--very puzzled.

willkg:
@peterbe ^^^ Tests pass, so it's set for review now.

webapp-django/crashstats/base/tests/test_ga.py
```diff
@@ -14,16 +16,30 @@
...
+
+def mock_uuid4(fun):
+    """Mock uuid.uuid4() decorator that returns a real non-random UUID"""
+    @wraps(fun)
+    def _mock_uuid4(*args, **kwargs):
+        with mock.patch('uuid.uuid4') as mocked_uuid:
+            mocked_uuid.return_value = UUID(hex=EXPECTED_CID, version=4)
+            return fun(*args, **kwargs)
+
+    return _mock_uuid4
```
willkg: This is sort of like what I'm doing with Antenna. The thinking is that mocking decorators and context managers provide a better testing infrastructure API than off-the-cuff mocking.

But if you don't like it, it's entirely in the last commit and we can nix that and go back to mocking uuid in each test that uses it.

PR #3561, opened 2016-10-28
Update and fix Vagrant environment

willkg:
- fixes Vagrant environment to build Python 2.7.11 and use it
- updates and tweaks documentation for setting up Vagrant environment
- tweaks setup to install vim and emacs to make life easier in
  the vm for Will

### DIFF
--- a/puppet/vagrant/modules/socorro/files/build_python.sh
+++ b/puppet/vagrant/modules/socorro/files/build_python.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 # This script downloads Python source tarball, builds it and does an altinstall.
 # An altinstall installs Python, but doesn't change any of the symlinks in
 # /usr/bin, so it doesn't affect anything else on the system that is expecting
@@ -7,7 +9,7 @@
 
 PYVERSION=2.7.11
 
-cd /tmp
+cd /tmp || exit 1
 wget "https://www.python.org/ftp/python/${PYVERSION}/Python-${PYVERSION}.tgz"
 tar -xzvf Python-${PYVERSION}.tgz
 cd Python-${PYVERSION}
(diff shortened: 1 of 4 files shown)


willkg:
Why maintain the vagrant environment? A couple of us still use it. It's the only environment we have to run all the tests. Given that, it seems worth maintaining until there's a viable alternative.

This update is important because it updates Python to the version we're using in -prod.

@lonnen: Do you want to review this? I think you and I are the only people who use Vagrant. If you don't (which is fine), I'll self-review it later and land it if it still looks ok. It's low impact--mostly I'm landing it to save me time.

puppet/vagrant/modules/socorro/files/build_python.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# This script downloads Python source tarball, builds it and does an altinstall.
+# An altinstall installs Python, but doesn't change any of the symlinks in
+# /usr/bin, so it doesn't affect anything else on the system that is expecting
+# Python 2.7.5.
+
+PYVERSION=2.7.11
+
+cd /tmp
```
lonnen: `SC2164: Use cd ... || exit in case cd fails.`

puppet/vagrant/modules/socorro/files/build_python.sh
```diff
@@ -0,0 +1,16 @@
...
+wget "https://www.python.org/ftp/python/${PYVERSION}/Python-${PYVERSION}.tgz"
+tar -xzvf Python-${PYVERSION}.tgz
+cd Python-${PYVERSION}
```
lonnen: `SC2164: Use cd ... || exit in case cd fails.`

puppet/vagrant/modules/socorro/files/build_python.sh
lonnen: for this and the next one, you could also consider `set -e`, which will exit on a failed cd _most_ of the time

willkg:
@lonnen I fixed the issues you brought up. Does it look ok now?

willkg:
Thank you!

PR #3564, opened 2016-10-28
bug 1312006 - Upload crash_reports.json to S3

peterbe:
Here's how I configured it when running locally:

```
[[class-UploadCrashReportJSONSchemaCronApp]]
    bucket_name=peterbe-crash-stats-telemetry
    access_key=AKIA...
    secret_access_key=SECRETYSECRET
```

I'm pretty sure we run crontabber with `socorro/common` as a prefix should it should automatically pick up the access_key and secret_access_key but I'm not so sure about the `bucket_name`. The only place this is elsewhere written down is for that `[[storage3]]` or whatever it's called in processor's crashstorage list. 

I'll update the bug with the necessary consulate instructions that I think we need.

### DIFF
--- a/socorro/cron/jobs/upload_crash_report_json_schema.py
+++ b/socorro/cron/jobs/upload_crash_report_json_schema.py
@@ -1,0 +1,58 @@
+from configman import Namespace
+from configman.converters import class_converter
+from crontabber.base import BaseCronApp
+
+from socorro.schemas import CRASH_REPORT_JSON_SCHEMA_AS_STRING
+
+
+class UploadCrashReportJSONSchemaCronApp(BaseCronApp):
+    app_name = 'upload-crash-report-json-schema'
+    app_description = """
+    We always send a copy of the crash (mainly processed crash) to a
+    S3 bucket meant for Telemetry to ingest.
+    When they ingest, they need a copy of our crash_report.json JSON
+    Schema file.
+
+    They use that not to understand the JSON we store but the underlying
+    structure (types, nesting etc.) necessary for storing it in .parquet
+    files in S3.
+
+    To get a copy of the crash_report.json they can take it from the
+    git repository but that's fragile since it depends on github.com
+    always being available.
+
+    By uploading it to S3 not only do we bet on S3 being more read-reliable
+    that github.com's server but by being in S3 AND unavailable that means
+    the whole ingestion process has to halt/pause anyway.
+    """
+
+    required_config = Namespace()
+    required_config.add_option(
+        'resource_class',
+        default=(
+            'socorro.external.boto.connection_context.S3ConnectionContext'
+        ),
+        doc=(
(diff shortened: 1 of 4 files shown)


socorro/cron/jobs/upload_crash_report_json_schema.py
```diff
@@ -0,0 +1,58 @@
...
+        doc=(
+            'fully qualified dotted Python classname to handle Boto '
+            'connections'
+        ),
+        from_string_converter=class_converter,
+        reference_value_from='resource.boto'
+    )
+    required_config.add_option(
+        'json_filename',
+        default='crash_report.json',
```
adngdb: I think I'd call it ``crash_report.schema.json`` in S3 to make sure folks stumbling upon that file know what it is immediately. In our code base it's in a ``/schemas`` folder so we don't need to extension.

socorro/cron/jobs/upload_crash_report_json_schema.py
```diff
@@ -0,0 +1,58 @@
...
+    They use that not to understand the JSON we store but the underlying
+    structure (types, nesting etc.) necessary for storing it in .parquet
+    files in S3.
+
+    To get a copy of the crash_report.json they can take it from the
+    git repository but that's fragile since it depends on github.com
+    always being available.
+
+    By uploading it to S3 not only do we bet on S3 being more read-reliable
+    that github.com's server but by being in S3 AND unavailable that means
```
adngdb: s/that github/than github

mreid-moz:
drive-by comment... could we this be set up as a github hook that uploads to S3 when a change to the crash_report schema is merged?

peterbe:
> drive-by comment... could we this be set up as a github hook that uploads to S3 when a change to the crash_report schema is merged?

Maybe. What I had in mind is a cron job that uploads it every hour. 
The advantage with that is that in our infra all cron job runners (it's based on something called [crontabber](https://github.com/mozilla/crontabber)) has all the config and credentials so that it can do the boto S3 upload. 

A webhook could potentially trigger a thing in our webapp that also as S3 config and credentials. 

Having it as one of these crontabber apps makes it quite explicit. And we have good monitoring and infrastructure to monitor these. [For example](https://crash-stats.mozilla.com/crontabber-state/)

mreid-moz:
Good monitoring sounds like a compelling reason to go with crontabber. WFM, just figured I'd check. Thanks!

socorro/cron/jobs/upload_crash_report_json_schema.py
peterbe: I disagree. I like that the name is the same. I think that keeps it cleaner than having different names. 

Also, it's clearly a JSON Schema file as soon as you open it.

peterbe:
@adngdb Description updated. It's hard to be accurate and brief at the same time.

PR #3566, opened 2016-11-01
bug 1314258 - run './manage.py collectstatic' with env vars

peterbe:
@willkg I honestly don't know how this script gets executed. Ideally, I'd like for it to be executed with the usual `env consul -prefix ....` stuff but I don't know where that happens and @jdotpz is on PTO today. 

I'm hopeful that this fix will unbreak our broken Google Analytics in stage and prod.

### DIFF
--- a/webapp-django/bin/bootstrap.sh
+++ b/webapp-django/bin/bootstrap.sh
@@ -10,13 +10,6 @@ source ${VIRTUAL_ENV:-"../socorro-virtualenv"}/bin/activate
 export PATH=$PATH:./node_modules/.bin/
 export SECRET_KEY="doesn't matter, tests"
 
-# See https://bugzilla.mozilla.org/show_bug.cgi?id=1314258
-# When the collectstatic command is run, it needs to have all the
-# config set from consulate. But because we don't know how to do
-# that here, we instead set it temporarily here in this build script.
-# NOTE! This is NOT a sensitive key/secret.
-export GOOGLE_ANALYTICS_ID="UA-35433268-50"
-
 echo "Install the node packages"
 npm install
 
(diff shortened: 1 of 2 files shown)


webapp-django/bin/bootstrap.sh
```diff
@@ -10,6 +10,13 @@ source ${VIRTUAL_ENV:-"../socorro-virtualenv"}/bin/activate
 export PATH=$PATH:./node_modules/.bin/
 export SECRET_KEY="doesn't matter, tests"
 
+# See https://bugzilla.mozilla.org/show_bug.cgi?id=1314258
+# When the collectstatic command is run, it needs to have all the
+# config set from consulate. But because we don't know how to do
+# that here, we instead set it temporarily here in this build script.
+# NOTE! This is NOT a sensitive key/secret.
+export GOOGLE_ANALYTICS_ID="UA-35433268-50"
+
```
willkg: I'd much rather we made this the default in the setting rather than feed it in this way.

willkg: My thinking is that we should limit the scope of hacks as much as possible. When we set the variable here, then it affects everything in `bootstrap.sh`. Further, if someone were trying to figure out why that google id were showing up in their file, they're much more likely to look at the settings file in django-land than here.

peterbe: So, you suggest we put our Mozilla ID in https://github.com/mozilla/socorro/blob/master/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/google_analytics.js instead?

peterbe: Oh. I misread. "default in the setting rather". Silly me. Will do.

willkg: https://github.com/mozilla/socorro/blob/master/webapp-django/crashstats/settings/base.py#L604

peterbe:
Moved to default now instead.

willkg:
We should have a bug to undo the hack and maybe a comment explaining why we're using that id.

Assuming that's the correct id, this looks good enough to me. r+

peterbe:
I've, hopefully, written a better comment now. The idea is that once JP comes back from PTO he can help us execute the `./manage.py collectstatic --noinput` command WITH consulate applied. 

Note that my commit in this PR does *not* close the bug.

PR #3567, opened 2016-11-02
Bug 1293776: Fix prefix

willkg:
The dogstatsd client uses the defaults (localhost and 8125 and no key
prefix), so that's why it's ignoring the key prefix we assigned.

I could fix that code, but that would affect all the metrics. Instead,
I'm manually adding the prefix.


willkg:
@peterbe Want to skim this? It's a really minor change.

willkg:
Thank you!

PR #3568, opened 2016-11-02
Fixes bug 946921 - Remove bogus fields from crash reports in Elastics…

adngdb:
…earch.

Make the Elasticsearch CrashStorage retry indexing documents in case of a failure that we know about, after "fixing" said document by removing the bogus field. Currently handles two cases, strings that are too long and integers that are too big.

cc @willkg, maybe you'll be interested in reviewing this as well.

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -18,12 +18,6 @@ from socorrolib.lib.datetimeutil import string_to_datetime
 from socorro.external.crashstorage_base import Redactor
 
 
-# These regex will catch field names from Elasticsearch exceptions. They have
-# been tested with Elasticsearch 1.4.
-FIELD_NAME_STRING_ERROR_REGEX = re.compile(r'field=\"([\w\-.]+)\"')
-FIELD_NAME_NUMBER_ERROR_REGEX = re.compile(r'\[failed to parse \[([\w\-.]+)]]')
-
-
 #==============================================================================
 class ESCrashStorage(CrashStorageBase):
     """This sends processed crash reports to Elasticsearch."""
@@ -48,6 +42,13 @@ class ESCrashStorage(CrashStorageBase):
     # This cache reduces attempts to create indices, thus lowering overhead
     # each time a document is indexed.
     indices_cache = set()
+
+    # These regex will catch field names from Elasticsearch exceptions. They
+    # have been tested with Elasticsearch 1.4.
+    field_name_string_error_re = re.compile(r'field=\"([\w\-.]+)\"')
+    field_name_number_error_re = re.compile(
+        r'\[failed to parse \[([\w\-.]+)]]'
+    )
 
     #--------------------------------------------------------------------------
     def __init__(self, config, quit_check_callback=None):
@@ -163,13 +164,13 @@ class ESCrashStorage(CrashStorageBase):
                 if 'MaxBytesLengthExceededException' in e.error:
                     # This is caused by a string that is way too long for
                     # Elasticsearch.
-                    matches = FIELD_NAME_STRING_ERROR_REGEX.findall(e.error)
+                    matches = self.field_name_string_error_re.findall(e.error)
                     if matches:
                         field_name = matches[0]
                 elif 'NumberFormatException' in e.error:
                     # This is caused by a number that is either too big for
                     # Elasticsearch or just not a number.
-                    matches = FIELD_NAME_NUMBER_ERROR_REGEX.findall(e.error)
+                    matches = self.field_name_number_error_re.findall(e.error)
                     if matches:
                         field_name = matches[0]
 
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import re
 from threading import Thread
 from Queue import Queue
 from contextlib import contextmanager
@@ -15,6 +16,10 @@ from socorro.external.es.index_creator import IndexCreator
 from socorrolib.lib.converters import change_default
 from socorrolib.lib.datetimeutil import string_to_datetime
 from socorro.external.crashstorage_base import Redactor
+
+
+FIELD_NAME_STRING_ERROR_REGEX = re.compile(r'field=\"([\w.]+)\"')
+FIELD_NAME_NUMBER_ERROR_REGEX = re.compile(r'\[failed to parse \[([\w.]+)]]')
 
 
 #==============================================================================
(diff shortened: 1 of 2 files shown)


socorro/external/es/crashstorage.py
```diff
@@ -138,22 +143,91 @@ def _submit_crash_to_elasticsearch(self, connection, crash_document):
...
+                )
+                success = True
+            except elasticsearch.exceptions.TransportError as e:
+                field_name = None
+
+                if 'MaxBytesLengthExceededException' in e.error:
+                    # This is caused by a string that is way too long for
+                    # Elasticsearch.
+                    matches = re.findall(
+                        FIELD_NAME_STRING_ERROR_REGEX, e.error
```
peterbe: Wow! I didn't even know this was possible!
I.e. 
```python
>>> re.findall('\d', 'n00dle')
['0', '0']
>>> re.findall(re.compile('\d'), 'n00dle')
['0', '0']
```

But you don't need to double-wrap it anyway. 
Just use `matches = FIELD_NAME_STRING_ERROR_REGEX.findall(e.error)`

peterbe: By the way, perhaps because I'm weird but the ALL_CAPS style is usually used for simple constants (e.g. strings, integers). 
If you want to define a global regex consider just making it a private looking thing like
```python
_field_name_string_error_regex = re.compile(...)


 class ESCrashStorage(CrashStorageBase):
    ...
```

socorro/external/es/crashstorage.py
```diff
@@ -17,6 +18,10 @@
 from socorro.external.crashstorage_base import Redactor
 
 
+FIELD_NAME_STRING_ERROR_REGEX = re.compile(r'field=\"([\w.]+)\"')
```
peterbe: What about underscores? Or hyphens?
You could consider perhaps...
```python
FIELD_NAME_STRING_ERROR_REGEX = re.compile('field="[^"]+"', re.DOTALL)
```
That way it's ANYTHING (even newlines and whitespace) it'll match just fine. Perhaps that not what you want but just a worthwhile thought.

socorro/external/es/crashstorage.py
```diff
@@ -138,22 +143,91 @@ def _submit_crash_to_elasticsearch(self, connection, crash_document):
...
+                if not field_name:
+                    # We are unable to parse which field to remove, we cannot
+                    # try to fix the document. Let it raise.
+                    self.config.logger.critical(
+                        'Submission to Elasticsearch failed for %s (%s)',
+                        crash_id,
+                        e,
+                        exc_info=True
+                    )
+                    raise
```
peterbe: Perhaps there's something I'm missing but isn't this repeated code?
Couldn't you just do...
```python
except elasticsearch.exceptions.TransportError as e:
    field_name = None
    if 'MaxBytesLengthExceededException' in e.error:
        ....
    elif 'NumberFormatException' in e.error:
        ...
    if not field_name:
        self.config.logger.critical(...)
        raise
```

socorro/external/es/crashstorage.py
```diff
@@ -138,22 +143,91 @@ def _submit_crash_to_elasticsearch(self, connection, crash_document):
...
+
+                if field_name.endswith('.full'):
+                    # Remove the `.full` at the end, that is a special mapping
+                    # construct that is not part of the real field name.
+                    field_name = field_name[:-5]
```
peterbe: Use `field_name = field_name.rstrip('.full')`

socorro/external/es/crashstorage.py
```diff
@@ -138,22 +143,91 @@ def _submit_crash_to_elasticsearch(self, connection, crash_document):
...
+                    if i == len(field_path) - 1:
+                        # This is the last level, so `field` contains the name
+                        # of the field that we want to remove from `parent`.
+                        del parent[field]
+                    else:
+                        parent = parent[field]
+
+                # Add a note in the document that a field has been removed.
+                if crash_document.get('removed_fields'):
+                    crash_document['removed_fields'] = '{} {}'.format(
```
peterbe: Perhaps it ought to be separated by a `; . Isn't that what we do for processor notes?

socorro/external/es/crashstorage.py
```diff
@@ -138,22 +143,91 @@ def _submit_crash_to_elasticsearch(self, connection, crash_document):
...
+                    )
+                else:
+                    crash_document['removed_fields'] = field_name
+            except elasticsearch.exceptions.ElasticsearchException as e:
+                self.config.logger.critical(
+                    'Submission to Elasticsearch failed for %s (%s)',
+                    crash_id,
+                    e,
+                    exc_info=True
+                )
```
peterbe: Why is the point of this? I'm not trying to convince you otherwise, just curious. And if there's a good reason to log, then it would be nice to have a comment why this logging has to happen. 
We have Sentry now so the exception will be there and it will have the crash_id in it too.

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -22,6 +22,12 @@
 from socorrolib.lib.datetimeutil import string_to_datetime
 
 
+# Uncomment these lines to decrease verbosity of the elasticsearch library
+# while running unit tests.
+#import logging
+#logging.getLogger('elasticsearch').setLevel(logging.ERROR)
```
peterbe: not pep8

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -615,3 +612,215 @@ def esindex_fn(*args, **kwargs):
             body=document,
             **additional
         )
+
+    @mock.patch('socorro.external.es.connection_context.elasticsearch')
```
peterbe: Why not `@mock.patch('elasticsearch.Elasticsearch')` instead? That's what we do, for example, in so many places where we mock on `requests`. E.g. `@mock.patch('requests.get')` (instead of `@mock.patch('crashstats.crashstats.models.requests')`)

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -615,3 +612,215 @@ def esindex_fn(*args, **kwargs):
...
+        Expected behavior is to remove that field and retry indexing.
+        """
+
+        # It's mocks all the way down.
+        sub_mock = mock.MagicMock()
+        espy_mock.Elasticsearch.return_value = sub_mock
+
+        # ESCrashStorage uses the "limited backoff" transaction executor.
+        # In real life this will retry operational exceptions over time, but
+        # in unit tests, we just want it to hurry up and fail.
```
peterbe: Nice comment!

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -615,3 +612,215 @@ def esindex_fn(*args, **kwargs):
...
+                    '(whose UTF8 encoding is longer than the max length 32766)'
+                    ', all of which were skipped.  Please correct the analyzer'
+                    ' to not produce such terms.  The prefix of the first '
+                    'immense term is: \'[124, 91, 48, 93, 91, 71, 70, 88, 49, '
+                    '45, 93, 58, 32, 65, 116, 116, 101, 109, 112, 116, 32, '
+                    '116, 111, 32, 99, 114, 101, 97, 116, 101]...\', original '
+                    'message: bytes can be at most 32766 in length; got 98489]'
+                    '; nested: MaxBytesLengthExceededException'
+                    '[bytes can be at most 32766 in length; got 98489]; '
+                )
```
peterbe: Nice!

socorro/external/es/crashstorage.py
```diff
@@ -17,6 +18,10 @@
 from socorro.external.crashstorage_base import Redactor
 
 
+FIELD_NAME_STRING_ERROR_REGEX = re.compile(r'field=\"([\w.]+)\"')
+FIELD_NAME_NUMBER_ERROR_REGEX = re.compile(r'\[failed to parse \[([\w.]+)]]')
```
peterbe: Nit: We might want to move these INTO the class `ESCrashStorage` as class attributes. 
There's no need to put them on a module level since we're not importing them from other places.

socorro/external/es/crashstorage.py
adngdb: Well, ``\w`` catches everything we care about except hyphens. We do have a bunch a very weird field names, but I don't think we actually want to capture these. I think it's a risk that we might mess the regex up and end up with some weird string matching the pattern.

socorro/external/es/crashstorage.py
adngdb: This is simply what we used to do before, but with a new indentation. I don't think it's a problem to have that error in sentry and in the processor logs. We can remove it later if we find it painful.

socorro/unittest/external/es/test_crashstorage.py
adngdb: Because that doesn't work here. I do not know why. ``mock`` is black magic to me, I usually just try patterns until I find one that works. This one works, the ``@mock.patch('elasticsearch.Elasticsearch')`` doesn't, so I'm sticking to mine. :)

However, if you want to educate me on the art of mocking and can explain what I am doing wrong, I would gladly learn!

socorro/external/es/crashstorage.py
peterbe: Ok. Let's just leave it.

socorro/external/es/crashstorage.py
peterbe: cool.

socorro/unittest/external/es/test_crashstorage.py
peterbe: See https://github.com/adngdb/socorro/pull/8
It requires mocking two things but it's better to go deeper so that it actually travels ConnectionContext "unmocked".

peterbe:
One weird thing...

* I found [a crash_id from sentry that had failed](https://sentry.prod.mozaws.net/operations/socorro-stage/issues/362962/).
* [You can load it as a report index page](https://crash-stats.allizom.org/report/index/697dd17b-422a-4be3-bdf7-896e62161107#tab-details)
* First I wanted to see it fail here in my local processor so I could see the glory of it *not* failing when I switch to your branch. But it did NOT fail. I tried to understand what's different about my local ES 1.7.5 compared to stage. 
Here's how I compared:
screen shot 2016-11-07 at 10 21 57 am
(my local ES)
screen shot 2016-11-07 at 10 22 56 am
(stage ES)
As you can see, my local ES index is different from stage. I have successfully run `python setup_supersearch_app.py`. Is there something else I need to do?

peterbe:
Oh I get it! The source of truth lies in SuperSearchFields which is only available in the actual database. There's no way I can have a database like that locally. 

Then what is `setup_supersearch_app.py` for?

And how can I get a local ES copy that has the same fields as on stage?

PR #3571, opened 2016-11-02
Fixes bug 1314789 - Added a gpu option for process_type in Topcrashers.

adngdb:


webapp-django/crashstats/supersearch/views.py
```diff
@@ -46,7 +46,6 @@
 # Facetting on those fields doesn't provide useful information.
 EXCLUDED_FIELDS_FROM_FACETS = (
     'date',
-    'dump',
```
adngdb: Completely unrelated to this work but the dump field doesn't exist anymore. So I just removed it from there. It's highly innocuous.

peterbe:
Maybe I'm a nut but I just think it's so weird to write real words in fake capitalization. E.g. "Gpu" is not a word. "GPU" is an abbreviation of the word "Graphical Processing Unit". 
I'll see if I can correct it myself.

peterbe:
Ahh... much better
screen shot 2016-11-03 at 11 37 17 am

peterbe:
See https://github.com/adngdb/socorro/pull/7

PR #3572, opened 2016-11-02
Fixes bug 1314822 - Removed ReportList API endpoint.

adngdb:


peterbe:
Is it time to deprecate this? Our deprecation warning just vaguely says "This endpoint is deprecated and will soon cease to exist."
Has there been any outcry/outcome of your communication about the deprecation?

adngdb:
I sent an email about 3 weeks ago, and no one answered it. Plus, the idea of removing those services one by one is that we can very easily turn them back on by reverting a small commit. 

I don't think people tend to answer emails, I'm not even sure they read emails, so I want to try something a bit more aggressive. We cannot be cautious forever, otherwise we'll never move forward.

PR #3577, opened 2016-11-03
Fixes bug 1311527 - Removed validation of dates in future in Super Search.

adngdb:


peterbe:
The only weird thing is that now you can search with a from date in the future too and there's no error. Just nothing found:
screen shot 2016-11-03 at 2 44 35 pm
Note, Nov 4 is in the future.

PR #3580, opened 2016-11-04
Fixes bug 1315324 - Add another heuristic rule for the signature correlations

marco-c:


adngdb:
Not merging yet because we are testing a new feature on stage.

PR #3581, opened 2016-11-09
Fixes bug 1316144 - Allow querying root fields in Super Search.

adngdb:
Fixes a bug with the way Super Search manages fields. A field that had no namespace (ie. that is in the root of the crash document) became `.field` instead of `field`, thus preventing any filtering or aggregating on it.

### DIFF
--- a/socorro/external/es/supersearch.py
+++ b/socorro/external/es/supersearch.py
@@ -73,16 +73,22 @@ class SuperSearch(SearchBase):
 
         return self.get_list_of_indices(start_date, end_date)
 
+    def get_full_field_name(self, field_data):
+        if not field_data['namespace']:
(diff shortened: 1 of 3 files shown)


adngdb:
The difference here lies in the context. The ``processed_crash`` is generated by the processor. It contains data that the rule system makes up. And it gets stored in all of our databases. We do remove some keys in some cases (mostly ES for disk space / error preventing), but I believe we should not be adding any data to it outside of the rule system. It's a matter of consistency. I believe ``CrashStorage`` classes should not be adding data to the ``raw_crash`` or the ``processed_crash``. Hence, for this case where we're adding data that is specific to Elasticsearch, I consider it is better to add it outside of these documents. Thus, I put it in the root of the crash document.

PR #3584, opened 2016-11-10
Fixes bug 1316185: fold socorrolib back in

willkg:
At the beginning of the year, we extracted socorrolib from socorro with
the thinking that we'd use socorrolib as a requirement for future
projects. Later, we decided to do those projects in other ways, so
then the value proposition of maintaining socorrolib as a separate
repository wasn't enthusing.

This merges that code back in. It was done in the following steps:

1. copy the relevant directories from socorrolib/ to socorro/
2. do a global search-and-replace of "socorrolib" to "socorro"

It's pretty straight-forward--nothing fancy.

### DIFF
--- a/docs/development/dumpingdumptables.rst
+++ b/docs/development/dumpingdumptables.rst
@@ -19,7 +19,7 @@ https://bugzilla.mozilla.org/show_bug.cgi?id=484032
 Library support
 ---------------
 
-'done' as of 2009-05-07 in socorrolib.lib.dmpStorage (Coding, testing is done; integration testing is done, 'go live' is today)
+'done' as of 2009-05-07 in socorro.lib.dmpStorage (Coding, testing is done; integration testing is done, 'go live' is today)
 Socorro UI
 
 /report/index/{uuid}
--- a/alembic/README
+++ b/alembic/README
@@ -35,7 +35,7 @@ We use a few types not currently supported by SQLAlchemy, so you may
 need to modify the migration slightly.
 
 
-== Helper functions in socorrolib.lib.migrations
+== Helper functions in socorro.lib.migrations
 
 We have a couple helper functions for doing cleanup tasks, and adding custom
 types. These functions and classes are imported by the default template for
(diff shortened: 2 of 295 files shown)


willkg:
I did the following to test things:

1. generate a fresh vagrant environment and run the tests inside that environment
2. run the socorro collector in a docker container with a prod-like configuration and send a test crash through

I don't have the processor dockerized, yet, so I can't test that easily. But I think we're probably fine.

I want to go through and re-review all the changes, then I'll mark this as ready to review.

willkg:
I went through and checked the changes. They're one of two things:

1. added file
2. changed "socorrolib" to "socorro"

The repository at this commit should have no instances of "socorrolib" in it.

Two things I didn't do:

1. fix the import statement order for any of the files adjusted--doing that would have increased the review work and possibility of error and it's easily done later
2. look into the DotDict situation--I want to look into that at some point, but will do that in a different bug on a rainy day

This is ready for review.

@peterbe: Can you skim it? Can you think of anything else that would need to get done that's not covered in https://bugzilla.mozilla.org/show_bug.cgi?id=1316185#c4 ?

docs/development/dumpingdumptables.rst
```diff
@@ -19,7 +19,7 @@ https://bugzilla.mozilla.org/show_bug.cgi?id=484032
 Library support
 ---------------
 
-'done' as of 2009-05-07 in socorrolib.lib.dmpStorage (Coding, testing is done; integration testing is done, 'go live' is today)
+'done' as of 2009-05-07 in socorro.lib.dmpStorage (Coding, testing is done; integration testing is done, 'go live' is today)
```
lonnen: ha, this becomes correct again

lonnen:
Spent the morning reviewing this.

I'm confident this restores the repository, more or less. 1 can be done later. 2 seems orthogonal. We had multiple dotdicts before socorrolib, we still have them now. I went through the diff specifically looking for DotDict issues and I don't think this will introduce any new ones.

r+

willkg:
I'm waiting on the following to land this:

1. verify that something runs the migrations to make sure they work
2. waiting for a -prod push so that -stage is clear to land this and test it

willkg:
Migrations are probably fine.

Merging now and then we'll do a system test.

PR #3585, opened 2016-11-14
Fixes bug 1317082 - Make loadChannelsData reject the promise when getDataURL throws an exception

marco-c:


peterbe:
Also, why don't we have correlations for Thunderbird? Do we not store any of them in Telemetry? I know we do send crashes over there. Is there *going* to be Thunderbird Telemetry data available? If so, we should put a link to the Bugzilla bug into the code in a comment.

peterbe:
Now might not be the time to talk about this but I don't like the pattern used here. `loadCorrelationData()` returns a promise but the promise doesn't come with data. I know you're using it for "caching" as it fills up an object outside the function. Why can't it be renamed to `getCorrelationData()` and return a promise with the data? You can use a closure to store the cache data. 

(This is not super important but just nerdy devs discussing for the sake of learning from each other)

peterbe:
One last thing, in case I can't convince you to write a more verbose `console.warn` message. 
When you load a report index, it just says "Unknown product: Thunderbird" in the web console. But it gives no context. The Report index page does a lot of other JS. When you're in development mode it's OK because you can see it happens in a file called "correlation.js" but in production the .js files are compacted into one `main.hashhash.min.js` or something.

marco-c:
> Also, why don't we have correlations for Thunderbird? Do we not store any of them in Telemetry? I know we do send crashes over there. Is there going to be Thunderbird Telemetry data available? If so, we should put a link to the Bugzilla bug into the code in a comment. 

I can calculate correlations for Thunderbird as well if needed. I haven't for now because I don't know if there's interest and because I suppose a lot of crashes would have similar correlations to Firefox.

P.S.: Currently I'm not using the Socorro Telemetry data because there are some things missing that I need, but calculating correlations for Thunderbird is independent from this.

marco-c:
> One last thing, in case I can't convince you to write a more verbose console.warn message. 

You had already convinced me :)

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
```diff
@@ -16,38 +16,49 @@ window.correlations = (function () {
...
-                correlationData[product][ch] = {
-                    'total': totals[ch],
-                    'signatures': {},
+            var dataURL = getDataURL(product);
+            if (!dataURL) {
+                console.warn('Correlation results unavailable for the "' + product + '" product.');
+                return null;
+            }
+
+            return fetch(getDataURL(product) + 'all.json.gz')
```
peterbe: Called twice

marco-c: Fixed in afdc9210489613e710d757fed0e46695a3a96315.

peterbe:
I'll test one more time in my local dev and then I'll merge if Travis is OK.

peterbe:
Argh! My bad. For some reason the green merge button had turned to one that doesn't squash and I didn't notice. Oh well. Too late now.

PR #3587, opened 2016-11-14
fixes bug 1230285 - Crontabber status page link in footer

peterbe:


lonnen:
Did you consider alternatives to this? It seems like we're piling so much stuff into the footer.

peterbe:
> Did you consider alternatives to this? It seems like we're piling so much stuff into the footer.

No. We had a length discussion about this and I think we removed the Status link and was going to replace it with a Status+CrontabberState but then we decided there's nothing from the Status page to keep so we removed it. So this is basically just a replacement of the Status link in there.

PR #3588, opened 2016-11-14
fixes bug 1136371 - enable HSTS

peterbe:


willkg:
This looks ok to me. We do the same thing with Standup, but we have the middleware higher in the middleware list. I looked at the docs and didn't see anything suggesting whether placement is meaningful, but @pmclanahan says it's a response middleware, so you want to put it as close to the top as possible so that it's responding to all requests.

One thing the docs do mention is that you should implement HSTS in the server rather than in django-land:

https://docs.djangoproject.com/en/1.8/ref/middleware/#module-django.middleware.security

I don't know offhand if our situation would benefit from implementing it in the server or not.

That's everything I can think of. This looks fine to me. r+

peterbe:
Been thinking, (thank you dog-walk breaks!) we should do this in Nginx instead. 
It doesn't make sense not to. Doing these kinds of things in Django only makes sense in hosted things like Heroku. If we do it in Nginx it's faster, covers non-django URLs (static assets) and just fewer moving parts in Django.

PR #3591, opened 2016-11-16
fixes bug 1013976 - logged in users to see their own PII

peterbe:
Note that it does NOT grant you access to download Unredacted JSON or minidumps. 
See https://bugzilla.mozilla.org/show_bug.cgi?id=1013976#c3 *"a desired feature in terms of data transparency (not so much an engineering ask)"*

### DIFF
--- a/webapp-django/crashstats/base/tests/testbase.py
+++ b/webapp-django/crashstats/base/tests/testbase.py
@@ -47,7 +47,16 @@ class DjangoTestCase(django.test.TestCase):
     def shortDescription(self):
         return None
 
-    def _login(self):
-        user = User.objects.create_user('test', 'test@example.com', 'secret')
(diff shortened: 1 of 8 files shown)


peterbe:
By the way, I took the liberty to clean up a bit too. No more realistic mozilla.com emails and I removed the repeated `_login()` method and improved the one in `testbase.py`.

adngdb:
Sorry, missed that one today. I'll review and take care of merging / pushing next week. (And if anything needs to be fixed, I can push to your branch and ask Will for further review :) ).

peterbe:
> Just to be overly cautious, I want to challenge this assertion: a logged in user necessarily has a not-empty email address (otherwise they could have access to most of the PII). Is that correct?

I don't understand this question. 
Every signed in persona always has an email address.

PR #3592, opened 2016-11-16
A large uint64_t can take more than 17 chars

bsergean:
To be serialized ...
This was crashing for us in sprintf on Linux (not on Mac weirdly).

```
#include <limits>
#include <iostream>

int main()
{
   std::cout << std::numeric_limits<uint64_t>::max() << std::endl;
   return 0;
}

./a.out | wc -c
     21
```


adngdb:
@luser mind taking a look at this?

luser:
Oops, that is silly. Thanks for the fix!

bsergean:
No, thank you guys for writing the stack walker in the first place :) !!

And thanks for merging, I just discovered this but it must have been a pretty rare case or maybe I wasn't looking at our logs well enough in the past.

PR #3593, opened 2016-11-17
Fixes bug 1315116 - Integrate new correlation results in the top crashers page

marco-c:


marco-c:
> I am so amazingly grateful for you stepping up Marco and doing the engineering approach of building instead of just filing bugs. Keep up all your hard work!

Thanks 😄 

> There's a "Load" link in every cell in the Correlations column. I have to manually click each one to find out if there's any correlation there. A computer ought to be able to do this for me.

I didn't want to perform so many requests as soon as the top crashers page was open. I could generate a file containing the list of available signatures on the server that generates the correlation results, this way it would be a single request.

> The overlay needs to be width constraint. 

Right! I hadn't noticed it.

> Why are we using \<pre\> anyway? :) http://jmp.sh/YQQesc8

We're using it to make the results aligned (in a way that keeps them aligned also when you copy & paste them to Bugzilla).

> What's the point of having this column at all? It's a kinda terrible UX (the overlay) compared to the fuller Correlations tab when you click into the Signature Report for each signature.

> Do we really need it as a column? What are the chances that people, who are hunting for correlations, don't do that on a signature by signature basis? Also, with one less column the TopCrasher page is going to get slightly less noisy and slightly more pleasant.

Yes, removing the column altogether might make sense. It was there, so I modified it to use the new correlations, but I'm not sure it is actually useful.

peterbe:
So here I'd love to hear from @adngdb  who might have some insight into the Correlations column's need. Also, @lonnen might have some interesting thoughts.

marco-c:
> When I click "Load" the spinner icon appears and it still has the "Load" link there.

I've just noticed that this is already the case with the current implementation. I'm not sure how you handle these things on Socorro (can PRs fix several issues, or do they have to be really self-contained?).

adngdb:
I would be in favor of simply removing the correlations column from topcrashers. To me it doesn't really make sense there. But, we'll want to have confirmation from our users that it is indeed not useful to them.

peterbe:
Did we manage to conclude this? We touched on it in Hawaii but no decision made.

adngdb:
Decision has been made to remove the correlations from Top Crashers. Closing this. 

Thanks for your work on this Marco!

PR #3594, opened 2016-11-17
Fixes bug 1318448: reduces deepcopies we do in the processor

willkg:
This changes the PolyCrashStorage code such that it only does a deep
copy of the processed crash if the crash storage says it mutates the
crash data.

Currently, we have 5 crash storage classes configured in production.
This change will reduce the number of deepcopies we're doing from 5 to
1.

r?

### DIFF
--- a/socorro/external/crashstorage_base.py
+++ b/socorro/external/crashstorage_base.py
@@ -232,6 +232,14 @@ class CrashStorageBase(RequiredConfig):
         """some implementations may need explicit closing."""
         pass
 
+    def is_mutator(self):
+        """Whether this storage class mutates the crash or not
+
+        By default, crash storage classes don't mutate the crash.
+
+        """
+        return False
+
     #--------------------------------------------------------------------------
     def save_raw_crash(self, raw_crash, dumps, crash_id):
         """this method that saves  both the raw_crash and the dump, must be
@@ -621,15 +629,20 @@ class PolyCrashStorage(CrashStorageBase):
         for a_store in self.stores.itervalues():
             self.quit_check()
             try:
-                a_store.save_raw_and_processed(
-                    raw_crash,
-                    dump,
+                if a_store.is_mutator():
                     # We do this because `a_store.save_raw_and_processed`
                     # expects the processed crash to be a DotDict but
                     # you can't deepcopy those, so we deepcopy the
                     # pure dict version and then dress it back up as a
                     # DotDict.
-                    SocorroDotDict(copy.deepcopy(processed_crash_as_dict)),
+                    crash = SocorroDotDict(copy.deepcopy(processed_crash_as_dict))
+                else:
+                    crash = processed_crash
+
+                a_store.save_raw_and_processed(
+                    raw_crash,
+                    dump,
+                    crash,
                     crash_id
                 )
             except Exception:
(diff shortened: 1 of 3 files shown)


socorro/external/es/crashstorage.py
```diff
@@ -303,6 +303,10 @@ class ESCrashStorageRedactedJsonDump(ESCrashStorageRedactedSave):
         )
     )
 
+    def is_mutator(self):
+        # This crash storage mutates the crash, so we mark it as such.
+        return True
```
willkg: This is the class that we use that mutates the data. I didn't go and check to see if other classes that we don't use mutate the data. We can write up another bug for that.

socorro/unittest/external/test_crashstorage_base.py
```diff
@@ -51,7 +51,23 @@ class B(A):
...
+    def save_raw_and_processed(
+        self,
+        raw_crash,
+        dump,
+        processed_crash,
+        crash_id
+    ):
+        # This is sort of a lie, but we lie so that we can verify the code went
+        # through the right path.
+        del processed_crash['foo']
```
willkg: I piggy-backed on the awesome tests that existed already. Thank you!

socorro/unittest/external/test_crashstorage_base.py
```diff
@@ -405,6 +421,45 @@ def test_poly_crash_storage_processed_crash_immutability(self):
...
+            default=PolyCrashStorage,
+        )
+        n.add_option(
+            'logger',
+            default=mock.Mock(),
+        )
+        value = {
+            'storage_classes': (
+                'socorro.unittest.external.test_crashstorage_base'
+                '.NonMutatingProcessedCrashCrashStorage'
```
willkg: This test is identical to one of the existing ones except this one uses `NonMutatingProcessedCrashCrashStorage` and has a different assert at the end.

willkg:
@peterbe Can you look at this?

socorro/external/crashstorage_base.py
```diff
@@ -621,15 +629,20 @@ def save_raw_and_processed(self, raw_crash, dump, processed_crash,
         for a_store in self.stores.itervalues():
             self.quit_check()
             try:
-                a_store.save_raw_and_processed(
-                    raw_crash,
-                    dump,
+                if a_store.is_mutator():
```
peterbe: Doesn't it need to be `if a_store.wrapped_object.is_mutator():`?

willkg: Oo... You're right--this is wrong.

willkg:
^^^ That should fix the wrapped problem.

peterbe:
Note-to-self;
I added `print (store_class, store_class.is_mutator())` and it prints:
```
(<socorro.external.boto.crashstorage.BotoS3CrashStorage object at 0x107b22c90>, False)
(<socorro.external.es.crashstorage.ESCrashStorageRedactedJsonDump object at 0x107b224d0>, True)
(<socorro.external.boto.crashstorage.TelemetryBotoS3CrashStorage object at 0x107b22350>, False)
```

willkg:
^^^ That fixes the fix. Bah.

PR #3596, opened 2016-11-17
Fix is_mutator() call

willkg:
The StatsdCounter is a duck-typed crash storage so it doesn't have an
is_mutator().

So what we're saying now is "if the crash storage has an is_mutator and
that returns True, then deepcopy". The underlying assumption is that
crash storages without an is_mutator aren't mutating the crash data.

### DIFF
--- a/socorro/external/crashstorage_base.py
+++ b/socorro/external/crashstorage_base.py
@@ -629,11 +629,9 @@ class PolyCrashStorage(CrashStorageBase):
         for a_store in self.stores.itervalues():
             self.quit_check()
             try:
-                store_instance = getattr(
-                    a_store, 'wrapped_object', a_store
-                )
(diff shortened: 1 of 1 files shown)


willkg:
@adngdb Can you look this over?

The problem we're seeing in -stage is that the StatsdCounter crash storage looks like this:

https://github.com/mozilla/socorro/blob/f409048a12c7d4159383b880394da8bcdc9b6643/socorro/external/statsd/statsd_base.py#L80

The `storage_instance` or whatever I called it is the `_dummy_do_nothing` method of that instance and thus doesn't have an `is_mutator`.

There are a couple of options:

1. fix StatsdCounter -- somewhat invasive change
2. do this hack where maybe things don't have an `is_mutator` and lower our expectations for crash storage API stuff

willkg:
I'll wait until Travis is happy, then land this.

PR #3598, opened 2016-11-18
Fixes bug 1316455 - Removed all deprecated /report/list implementations.

adngdb:


peterbe:
Re-open so that github will try to send this to travis again.

PR #3602, opened 2016-11-23
Fixes bug 1313283 - Strip signatures from leading or trailing white s…

adngdb:
…paces.

@willkg What do you think about this?

Fun fact: we can easily use SuperSearch to find all signatures starting or ending with a white space, e.g. https://crash-stats.mozilla.com/api/SuperSearch/?signature=%24%20&_results_number=0

### DIFF
--- a/socorro/processor/mozilla_processor_2015.py
+++ b/socorro/processor/mozilla_processor_2015.py
@@ -63,6 +63,7 @@ mozilla_processor_rule_sets = [
         "socorro.processor.signature_utilities.SignatureRunWatchDog, "
         "socorro.processor.signature_utilities.SignatureIPCChannelError, "
         "socorro.processor.signature_utilities.SignatureIPCMessageName, "
+        "socorro.processor.signature_utilities.SigTrim, "
         "socorro.processor.signature_utilities.SigTrunc, "
     ],
     [   # a set of classifiers for support
--- a/socorro/processor/signature_utilities.py
+++ b/socorro/processor/signature_utilities.py
@@ -631,6 +631,25 @@ class AbortSignature(Rule):
 
 
 #==============================================================================
+class SigTrim(Rule):
+    """ensure that the signature never has any leading or trailing white
+    spaces"""
+
+    #--------------------------------------------------------------------------
+    def version(self):
+        return '1.0'
+
+    #--------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


willkg:
I haven't done much with processor rules, but this seems fine to me.

I'm curious why we create signatures with spaces at the beginning and end in the first place. Maybe we should adjust the thing that's creating signatures to `strip()` when it's done?

adngdb:
There's few places where we replace the signature with something like ``Marker | some relevant text``, but we don't want signatures to be too long, so we cut them. I figured the easiest and safest way to ensure we never have any trailing spaces is to just have a rule that ensures that cannot happen. I _think_ the cost of doing that is small enough that we can afford doing it for every crash, but I could be proven wrong.

socorro/processor/mozilla_processor_2015.py
```diff
@@ -63,6 +63,7 @@
         "socorro.processor.signature_utilities.SignatureRunWatchDog, "
         "socorro.processor.signature_utilities.SignatureIPCChannelError, "
         "socorro.processor.signature_utilities.SignatureIPCMessageName, "
+        "socorro.processor.signature_utilities.SigTrim, "
```
peterbe: I really don't understand the logic of putting this transform *before* `SigTrunc`. Suppose it's 

1) `Some signature that is really long `
2) `Some signature that is really long`
2) `Some signature that ` (i.e. truncating away the last couple of words)

peterbe:
By the way, I wholeheartedly approve the idea of ALWAYS whitespace stripping all signatures.

socorro/processor/mozilla_processor_2015.py
willkg: I think it makes sense to strip it before truncating so the extra spaces don't affect truncation. Having said that, maybe `SigTrunc` should take the signature, strip it and then truncate it and make sure it's stripped?

adngdb: I like you proposal Will, and I think it makes sense. I'll merge this new rule into the ``SigTrunc`` one.

adngdb:
Actually! The ``SigTrunc`` rule doesn't just "cut" the signature, it adds ``...`` at the end. So it is impossible for a signature to have a trailing white space after passing ``SigTrunc``. I this think this is good as is. Better separation of concerns and all.

PR #3605, opened 2016-11-25
Fixes bug 1320394 - Take into account priors when printing and sorting correlation results

marco-c:


webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
```diff
@@ -136,10 +136,36 @@ window.correlations = (function () {
...
+                var count_group_b = b.count_group;
+                var total_group_b = total_group;
+                var count_reference_b = b.count_reference;
+                var total_reference_b = total_reference;
+            } else {
+                var count_group_b = b.prior.count_group;
+                var total_group_b = b.prior.total_group;
+                var count_reference_b = b.prior.count_reference;
+                var total_reference_b = b.prior.total_reference;
+            }
```
peterbe: There's a lot of repetition here and I don't know what the latest linting rules think about declaring the same named variable in the same scope. 
I think all of these could be refactored to something like `var count_group_a = a.prior && a.prior.count_group || a.count_group`
Or you can rewrite `a` to so something like this:
```javascript
if (a.prior) {
    a = a.prior;
}
var count_group_a = a.count_group;
...
```

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
```diff
@@ -169,7 +195,17 @@ window.correlations = (function () {
...
                 var percentRef = toPercentage(line.count_reference / total_reference);
-                return '(' + percentGroup + '% in signature vs ' + percentRef + '% overall) ' + itemToLabel(line.item);
+
+                var result = '(' + percentGroup + '% in signature vs ' + percentRef + '% overall) ' + itemToLabel(line.item);
+
+                // If the rule has a prior that alters its distribution significantly, print it after the rule.
+                if (line.prior) {
+                    var percentGroupGivenPrior = toPercentage(line.prior.count_group / line.prior.total_group);
+                    var percentRefGivenPrior = toPercentage(line.prior.count_reference / line.prior.total_reference);
+                    result += ' [' + percentGroupGivenPrior + '% vs ' + percentRefGivenPrior + '% if ' + itemToLabel(line.prior.item) + ']'
```
peterbe: Missing semicolon. 
You might want to configure a git hook that runs jslint or eslint to check of easy mistakes.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
marco-c: I wanted to clean this up a bit, but I couldn't find a way that I really liked. I like your second suggestion though.

marco-c: Actually, I can't use the second suggestion, because for example `total_group_a` is either `total_group` (when a doesn't have a prior) or `a.prior.total_group`.

marco-c: I'll do this:
```
var count_group_a = a.prior ? a.prior.count_group : a.count_group;
```
which was what I initially did but then discarded because I didn't like evaluating the same condition (`a.prior`) multiple times.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
marco-c: I thought Travis was running a linter for me. Fixed in 541464e.

peterbe:
By the way, Travis only lints the .py code I think. And only for certain sections. 
The reason we never set up JS linting was because of all the `jquery.somelib.min.js` things. 
But honestly, it wouldn't be impossible to have some sort of list of things to ignore.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
```diff
@@ -136,10 +136,21 @@ window.correlations = (function () {
...
+            var total_group_b = b.prior ? b.prior.total_group : total_group;
+            var count_reference_b = b.prior ? b.prior.count_reference : b.count_reference;
+            var total_reference_b = b.prior ? b.prior.total_reference : total_reference;
+
             // Then, sort by percentage difference between signature and overall (using the lower endpoint
             // of the confidence interval of the difference).
-            var ciA = confidenceInterval(a.count_group, total_group, a.count_reference, total_reference);
-            var ciB = confidenceInterval(b.count_group, total_group, b.count_reference, total_reference);
+            var ciA = confidenceInterval(count_group_a, total_group_a, count_reference_a, total_reference_a);
+            var ciB = confidenceInterval(count_group_b, total_group_b, count_reference_b, total_reference_b);
```
adngdb: This big chunk of ``a = b ? c : d`` lines makes me unhappy. I think this would be easier to read and understand: 

```javascript
var ciA = null;
if (a.prior) {
    ciA = confidenceInterval(
        a.prior.count_group,
        a.prior.total_group, 
        a.prior.count_reference, 
        a.prior.total_reference
    );
}
else {
    ciA = confidenceInterval(a.count_group, total_group, a.count_reference, total_reference);
}
// Same for `b`.
```

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
```diff
@@ -169,7 +180,17 @@ window.correlations = (function () {
...
                 var percentGroup = toPercentage(line.count_group / total_group);
                 var percentRef = toPercentage(line.count_reference / total_reference);
-                return '(' + percentGroup + '% in signature vs ' + percentRef + '% overall) ' + itemToLabel(line.item);
+
+                var result = '(' + percentGroup + '% in signature vs ' + percentRef + '% overall) ' + itemToLabel(line.item);
+
+                // If the rule has a prior that alters its distribution significantly, print it after the rule.
+                if (line.prior) {
+                    var percentGroupGivenPrior = toPercentage(line.prior.count_group / line.prior.total_group);
+                    var percentRefGivenPrior = toPercentage(line.prior.count_reference / line.prior.total_reference);
```
adngdb: You might want to make sure that ``line.prior.total_group`` and ``line.prior.total_reference`` are different than zero, to prevent possible mistakes.

adngdb:
See https://github.com/marco-c/socorro/pull/2 about my concerns.

PR #3606, opened 2016-11-29
Fixes bug 1320922 - Add cpu_microcode_version from /proc/cpuinfo to json_dump

marco-c:


peterbe:
This PR needs a reviewer.

marco-c:
@luser r?

minidump-stackwalk/stackwalker.cc
```diff
@@ -795,6 +795,35 @@ static string stripquotes(const string& s) {
...
+static void ConvertCPUInfoToJSON(const string& cpuinfo,
+                                 Json::Value& root) {
+  for (string& line : split(cpuinfo, '\n')) {
+    vector<string> bits = split(line, ':');
+    if (bits.size() != 2 || !startswith(bits[0], "microcode")) {
+        continue;
+    }
+
+    try {
+        root["system_info"]["cpu_microcode_version"] = static_cast<Json::UInt>(std::stoi(trim(bits[1]), 0, 0));
```
luser: two nits:
1. You should pass `nullptr` as the second parameter here
2. I think you should explicitly pass `16` for the third parameter, since this should always be a hex string. 

Adding C++ exception usage to this file makes me a little uncomfortable, since we're not currently using them elsewhere, but this is only ever run as a standalone binary anyway so I think it should be fine. (We just don't use C++ exceptions anywhere in Gecko so I have a bit of an aversion to them.)

luser:
r=me with those tiny nits fixed

marco-c:
Nits fixed!

PR #3609, opened 2016-11-29
fixes bug 1308218 - remove caching on supersearch and single crash mo…

peterbe:
…dels


adngdb:
I think this is a no-go. You have identified a problem, but this solution is not guaranteed to provide improvements. It could make some situations go really wrong, for example a link that is heavily hit. We would need to have much better metrics to ensure that we actually improve the situation, and that is not something we want to invest into right now. So, in fear of making things worse, I'm going to close this.

PR #3610, opened 2016-11-29
fixes bug 1319483 - ignore platform if not selective

peterbe:
The change is small but important. What it does now is that instead, by default, it selects NO platforms. I.e. the 3 checkboxes are not checked. Only if you select one, does the platform parameter go to supersearch. 

This solves the problem for platforms like FennecAndroid that has no crashes under either Windows, Mac OS X or Linux.

### DIFF
--- a/webapp-django/crashstats/crashstats/views.py
+++ b/webapp-django/crashstats/crashstats/views.py
@@ -724,6 +724,11 @@ def crashes_per_day(request, default_context=None):
     for version in context['active_versions'][params['product']]:
         context['available_versions'].append(version['version'])
 
+    if not params.get('platforms'):
+        params['platforms'] = [
+            x['name'] for x in platforms if x.get('display')
+        ]
+
     context['platforms'] = params.get('platforms')
 
     end_date = params.get('date_end') or datetime.datetime.utcnow()
@@ -779,12 +784,17 @@ def crashes_per_day(request, default_context=None):
     # to be called `version`
     supersearch_params = copy.deepcopy(params)
     supersearch_params['version'] = supersearch_params.pop('versions')
-    if 'platforms' in supersearch_params:
-        supersearch_params['platform'] = supersearch_params.pop('platforms')
-        # in SuperSearch it's called 'Mac' not 'Mac OS X'
-        if 'Mac OS X' in supersearch_params['platform']:
-            supersearch_params['platform'].append('Mac')
-            supersearch_params['platform'].remove('Mac OS X')
+    supersearch_params['platform'] = supersearch_params.pop('platforms')
+    # in SuperSearch it's called 'Mac' not 'Mac OS X'
+    if 'Mac OS X' in supersearch_params['platform']:
+        supersearch_params['platform'].append('Mac')
+        supersearch_params['platform'].remove('Mac OS X')
+
+    if params['product'] == 'FennecAndroid':
+        # FennecAndroid only has one platform and it's "Android"
+        # so none of the options presented in the crashes_per_day.html
+        # template are applicable.
+        del supersearch_params['platform']
 
     try:
         graph_data, results, adi_by_version = _get_crashes_per_day_with_adu(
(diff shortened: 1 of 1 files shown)


webapp-django/crashstats/crashstats/views.py
```diff
@@ -723,10 +723,6 @@ def crashes_per_day(request, default_context=None):
     for version in context['active_versions'][params['product']]:
         context['available_versions'].append(version['version'])
 
-    if not params.get('platforms'):
-        params['platforms'] = [
-            x['name'] for x in platforms if x.get('display')
-        ]
```
willkg: Looks like this used to check to see if the user was specifying platforms and if not, it'd reduce the search to all displayed platforms. (I'm not clear on what "display" means in this context, but guessing it's what's in a drop-down somewhere.)

webapp-django/crashstats/crashstats/views.py
```diff
@@ -782,11 +778,12 @@ def crashes_per_day(request, default_context=None):
...
-    # in SuperSearch it's called 'Mac' not 'Mac OS X'
-    if 'Mac OS X' in supersearch_params['platform']:
-        supersearch_params['platform'].append('Mac')
-        supersearch_params['platform'].remove('Mac OS X')
+    if 'platforms' in supersearch_params:
+        supersearch_params['platform'] = supersearch_params.pop('platforms')
+        # in SuperSearch it's called 'Mac' not 'Mac OS X'
+        if 'Mac OS X' in supersearch_params['platform']:
+            supersearch_params['platform'].append('Mac')
+            supersearch_params['platform'].remove('Mac OS X')
```
willkg: This changes the behavior such that if the user didn't specify platforms, then it doesn't reduce on platforms at all.

This seems like a big backwards-incompatible behavioral change. Is this really what we want to do? Any idea how it'll affect the crashes-per-day graphs? Should we tell someone that this is happening before we land it?

Would it be better to add a "Ignore platform" option instead?

webapp-django/crashstats/crashstats/views.py
peterbe: If you go to `https://crash-stats.mozilla.com/crashes-per-day/?p=Firefox` it used to automatically added `{'platform': ['Windows', 'Linux', 'Mac OS X']}` to the parameters sent to SuperSearch. But if you go to https://crash-stats.mozilla.com/crashes-per-day/?p=FennecAndroid it would do that same thing and because the platform for all FennecAndroid crashes is "Android" that extra supersearch parameter would excluded all crashes. 

Secondly, the UX *used* to be that all checkboxes are checked by default so once you, for example, change the dates on that page and click the "Generate" button it would add `&platforms=Windows&platforms=Linux&platforms=Mac%20OS%20X` to the query string. 

What I'm changing is basically that if nothing is selected (or all three options are selected) the whole `platform` parameter is totally ignored in terms of fetching the data.

webapp-django/crashstats/crashstats/views.py
peterbe: An alternative, but arguably more hackish, solution is to leave all the checkboxes on by default but then do something like this:
```python
if params['product'] == 'FennecAndroid':
    # the 'platforms' field no longer makes sense
    del params['platform']
```

willkg: I prefer your hackish option since that's clear in its intentions and has a comment.

willkg:
I prefer the hackish option, so I think we should change to that because it's clearer.

Further, I think this could use a test.

peterbe:
I opted for not having another test because all the products are mocked so when I tried with `FennecAndroid` it fails because that's that's not a recognized product. All tests strive for avoiding real product names. 

I certainly could extend that an continue the exception for FennecAndroid in the fixtures too but doesn't seem important enough at the moment. Especially since we're soon going to phase out Crashes per Day for something based on Telemetry data.

willkg:
Ok. Looks good enough. r+

PR #3613, opened 2016-11-30
Auth-Token and raven

peterbe:
@willkg I don't know how environmentally equipped you are to review and test this. 
Testing is actually kinda messy. Enabled the raven DSN we use for socorro-stage. Then I put in a temporary `raise` and tested it with:
```
curl -H "Auth-token: e85659f64c734013a39df04a74a716b4" http://socorro.dev/api/UploadedSymbols/\?end_date\=2016-11-29\&start_date\=2016-11-01
```
(note token, it just for my laptop)
Here's what it looks like:
https://sentry.prod.mozaws.net/operations/socorro-stage/issues/370785/

The most important change is that now the `Auth-Token` is scrubbed. 
screen shot 2016-11-30 at 1 18 08 pm

### DIFF
--- a/webapp-django/crashstats/tokens/utils.py
+++ b/webapp-django/crashstats/tokens/utils.py
@@ -4,4 +4,7 @@ from raven.processors import SanitizePasswordsProcessor
 class RavenSanitizeAuthTokenProcessor(SanitizePasswordsProcessor):
     """Extend the standard SanitizePasswordsProcessor with one more key
     which is 'Auth-Token' which is what we use in our API."""
-    FIELDS = tuple(SanitizePasswordsProcessor.FIELDS) + ('auth-token',)
+    # Make it a frozenset because that's what the base class has.
+    FIELDS = frozenset(
+        tuple(SanitizePasswordsProcessor.FIELDS) + ('auth-token',)
+    )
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/tokens/utils.py
```diff
@@ -0,0 +1,7 @@
+from raven.processors import SanitizePasswordsProcessor
+
+
+class RavenSanitizeAuthTokenProcessor(SanitizePasswordsProcessor):
+    """Extend the standard SanitizePasswordsProcessor with one more key
+    which is 'Auth-Token' which is what we use in our API."""
+    FIELDS = tuple(SanitizePasswordsProcessor.FIELDS) + ('auth-token',)
```
willkg: `FIELDS` is a frozenset:

https://github.com/getsentry/raven-python/blob/908f3711b175c8184244ee9d648d625e446007a9/raven/processors.py#L75

I think I'd probably make sure it stays a frozenset.

Relatedly, it's unenthusing that it's this difficult to add other things to the sanitization list with raven.

peterbe:
I did at first write a class with its own `process` method and there I could do something like:
```python
def process(self, data, **kwargs):
    if data['request']['headers'].get('Auth-Token'):
        data['request']['headers']['Auth-Token'] = '*' * 10
```
But I don't know how case sensitivity works. Also, I'd need to mimic how other passwords are scrubbed (technically they display 8 `*` chars). 

I'll make it a `frozenset` too.

peterbe:
Updated.

PR #3614, opened 2016-11-30
fixes bug 1321367 - BadArgumentError in Crashes per Day on bad date

peterbe:
Now it looks like this:
screen shot 2016-11-30 at 2 20 02 pm

At first I thought this a bit ugly and we should instead re-render with error messages in the form. 
However, that's [not what we do with any other form validation errors](https://github.com/mozilla/socorro/blob/7a0329ba25ee2c306f783cf4484fbaa7d2a3387f/webapp-django/crashstats/crashstats/views.py#L700).

### DIFF
--- a/webapp-django/crashstats/crashstats/tests/test_views.py
+++ b/webapp-django/crashstats/crashstats/tests/test_views.py
@@ -26,6 +26,7 @@ from django.core.cache import cache
 from django.core.urlresolvers import reverse
 from django.contrib.contenttypes.models import ContentType
 
+from socorro.lib import BadArgumentError
 from socorro.external.crashstorage_base import CrashIDNotFound
 
 from crashstats.base.tests.testbase import DjangoTestCase
@@ -1548,6 +1549,44 @@ class TestViews(BaseTestViews):
 
         ok_('Our database is experiencing troubles' in response.content)
         ok_('week of 2010-01-04 is ~40% lower' in response.content)
+
+    @mock.patch('requests.get')
+    def test_crashes_per_day_bad_argument_error(self, rget):
+        url = reverse('crashstats:crashes_per_day')
+
+        def mocked_adi_get(**options):
+            return {
+                'total': 0,
+                'hits': []
+            }
+
+        models.ADI.implementation().get.side_effect = mocked_adi_get
+
+        def mocked_product_build_types_get(**options):
+            return {
+                'hits': {
+                    'release': 0.1,
+                    'beta': 1.0,
+                }
+            }
+
+        models.ProductBuildTypes.implementation().get.side_effect = (
+            mocked_product_build_types_get
+        )
+
+        def mocked_supersearch_get(**params):
+            raise BadArgumentError('Someone is wrong on the Internet')
+
+        SuperSearchUnredacted.implementation().get.side_effect = (
+            mocked_supersearch_get
+        )
+
+        response = self.client.get(url, {
+            'p': 'WaterWolf',
+            'v': ['20.0', '19.0']
+        })
+        eq_(response.status_code, 400)
+        ok_('Someone is wrong on the Internet' in response.content)
 
     def test_crashes_per_day_legacy_by_build_date(self):
         url = reverse('crashstats:crashes_per_day')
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/crashstats/views.py
```diff
@@ -788,13 +789,16 @@ def crashes_per_day(request, default_context=None):
...
+    try:
+        graph_data, results, adi_by_version = _get_crashes_per_day_with_adu(
+            supersearch_params,
+            start_date,
+            end_date,
+            platforms,
+            _date_range_type
+        )
+    except BadArgumentError as exception:
+        return http.HttpResponseBadRequest(unicode(exception))
```
willkg: Why pass a unicode here? Methinks this should be a str. At a minimum, it'll get converted to a str so if you don't do the encoding yourself, then you leave it to Django to do in a way you don't control.

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -1549,6 +1550,44 @@ def mocked_supersearch_get(**params):
...
+        SuperSearchUnredacted.implementation().get.side_effect = (
+            mocked_supersearch_get
+        )
+
+        response = self.client.get(url, {
+            'p': 'WaterWolf',
+            'v': ['20.0', '19.0']
+        })
+        eq_(response.status_code, 400)
+        ok_('Someone is wrong on the Internet' in response.content)
```
willkg: The error that bug 1321367 covers is bad dates, but this doesn't seem to test bad dates at all. Plus there's so much mocked out here, I'm not clear on what's actually getting tested.

webapp-django/crashstats/crashstats/views.py
peterbe: The reason for using unicode is to avoid UnicodeEncodeErrors on the stringify on an exception object. This happens when hackerbots try passing crazy and invalid unicode on query string parameters. See https://github.com/mozilla/socorro/pull/3612

webapp-django/crashstats/crashstats/tests/test_views.py
peterbe: The point of this test is to capture `BadArgumentError` exceptions inside the implementation class. 

Back in the day when we had a middleware service, if there was anything wrong with the data the middleware server would respond with a `400 Bad Request` and we would sometimes let that bubble and sometimes re-wrap that. Then we did away with the middleware service and let django "talk directly" to the python code and now we could get `BadArgumentError` exceptions in the django view functions.

webapp-django/crashstats/crashstats/views.py
willkg: Ouch--I wouldn't have fixed #3612 that way, either. If you read through the Django code, it converts everything down to str. By passing in a unicode here, you're just passing the buck and who knows what happens in Django-land.

I think it's a mistake. If you really think that calling unicode is the right thing here, then you should do:

```
return http.HttpResponseBadRequest(unicode(exception).encode('utf-8'))
```

webapp-django/crashstats/crashstats/tests/test_views.py
willkg: So why mock everything out? Why not actually send in a bad argument?

peterbe: All implementation classes are mocked within the context of the django tests.

peterbe: The tests that actually use postgres and ES are in `socorro/unittest/external/es/` etc.

webapp-django/crashstats/crashstats/views.py
peterbe: How odd. But thanks for sharing!
So what if you had something like this:
```python
  return http.HttpResponseBadRequest(u'Skät också')
```
Would you change that to:
```python
  return http.HttpResponseBadRequest('Skät också')
```

willkg: No, because that's not a str. You could use Django's `make_bytes`, if that helps.

willkg:
I think this underscores some serious problematic ways of doing things in the Socorro codebase. However, the changes in this particular PR aren't making things worse and they seem to fix the problem specified in the bug. So r+.

PR #3617, opened 2016-12-02
fixes bug 1316407 - Create a suite of e2e tests for the API

peterbe:
CC @m8ttyB

### DIFF
--- a/e2e-tests/tests/test_api.py
+++ b/e2e-tests/tests/test_api.py
@@ -1,0 +1,121 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from collections import defaultdict
+
+import requests
+import pytest  # noqa
+
+
+class TestAPI:
+
+    def test_public_api_navigation(self, base_url):
+        response = requests.get(base_url + '/api/ProductVersions/', {
+            'active': True,
+        })
+        assert response.status_code == 200
+
+        product_versions = defaultdict(list)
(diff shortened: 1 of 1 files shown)


m8ttyB:
Since these are read-only tests, add `@pytest.mark.nondestructive` to each test. Other than that tests look good.

e2e-tests/tests/test_api.py
```diff
@@ -0,0 +1,121 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from collections import defaultdict
+
+import requests
+import pytest  # noqa
+
+
+class TestAPI:
+
```
m8ttyB: add `@pytest.mark.nondestructive` to each test method

peterbe:
Done.

e2e-tests/tests/test_api.py
```diff
@@ -0,0 +1,125 @@
...
+            })
+            assert response.status_code == 200
+            assert response.json()['hits']
+
+    @pytest.mark.nondestructive
+    def test_supersearch_fields(self, base_url):
+        response = requests.get(base_url + '/api/SuperSearchFields/')
+        assert response.status_code == 200
+        fields = response.json()
+        assert fields['uuid'], fields
```
willkg: I don't understand this line. pytest generally produces good assertion error messages. Seems like this is testing to see if `fields['uuid']` is truthy, but I suspect it's really just checking to see if `'uuid'` is in fields.

I'm going to tweak this in a followup commit.

willkg:
This looks good to me.

@adngdb Can you check to see if the tests are helpful and/or if they're missing anything? If it's good, we should land it.

PR #3618, opened 2016-12-07
IAM roles

ezolenko:
boto supports IAM roles as long as no keys are passed when creating connections. This works in my v254 deployment for S3 crash storage and symbol uploads from EC2 instances with IAM role that has S3 access. I patched rpm install instead of building from source though, but I think this covers everything.


ezolenko:
Hold on, didn't get all of them...

PR #3619, opened 2016-12-07
fixes bug 1322406 - Enable SetRemoteAddrFromForwardedFor middleware

peterbe:


willkg:
I looked to see if we had tests for this and I don't see any. It gets added as a middleware for `crashstats/api/tests/test_views.py:TestViews`, but I don't see any tests that populate the relevant header and/or test ratelimiting.

I think we should have a test verifying the middleware is doing the right thing or a test that verifies ratelimiting works with the middleware or something along those lines.

peterbe:
Test added @willkg

willkg:
Looks good to me. r+

PR #3620, opened 2016-12-07
ES_SHARDS_PER_INDEX => 10 (was 5)

milescrabill:
@adngdb r?

This should help allow us to add more nodes to the elasticsearch cluster, which will increase its resilience.

### DIFF
--- a/socorro/external/es/super_search_fields.py
+++ b/socorro/external/es/super_search_fields.py
@@ -25,6 +25,8 @@ ES_CUSTOM_ANALYZERS = {
 ES_QUERY_SETTINGS = {
     'default_field': 'signature'
 }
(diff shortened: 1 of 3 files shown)


lonnen:
I'm pretty sure this is a cosmetic configuration variable kept manually in sync with a cluster setting. It will be necessary to land after we change the cluster but will not effect the desired change on its own.

lonnen:
Build failure looks like a temp problem in travis infra. kicking those

lonnen:
I'm good with this. @adngdb -- thoughts?

adngdb:
I'd rather we merge #3632 but if that is not possible (r- there, or whatever reason), I am fine with merging this and making it better later on.

PR #3621, opened 2016-12-08
unblocking support for AWS IAM roles instead of hardcoding access keys

ezolenko:
boto supports IAM roles as long as no keys are passed when creating connections. This works in my v254 deployment for S3 crash storage and symbol uploads from EC2 instances with IAM role that has S3 access. I patched rpm install instead of building from source though, but I think this covers everything.

### DIFF
--- a/socorro/external/boto/connection_context.py
+++ b/socorro/external/boto/connection_context.py
@@ -124,13 +124,13 @@ class ConnectionContextBase(RequiredConfig):
     required_config.add_option(
         'access_key',
         doc="access key",
-        default="",
+        default=None,
         reference_value_from='resource.boto',
     )
     required_config.add_option(
         'secret_access_key',
         doc="secret access key",
-        default="",
+        default=None,
         secret=True,
         reference_value_from='secrets.boto',
         likely_to_be_changed=True,
--- a/webapp-django/crashstats/settings/base.py
+++ b/webapp-django/crashstats/settings/base.py
@@ -577,8 +577,8 @@ ANALYZE_MODEL_FETCHES = config('ANALYZE_MODEL_FETCHES', False, cast=bool)
 
 
 # Credentials for being able to make an S3 connection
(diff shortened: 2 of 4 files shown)


willkg:
I understand the issue you're having. When I wrote Antenna (the next iteration of the Socorro collector), I wrote it so that it worked with the situation you're referring to.

However, this PR has no tests and I'm not sure I have a good way of verifying correctness in all the parts that you've changed.

Instead of adjusting the base classes, can you instead create a connection class that covers the situation you have? That would cover the collector and processor. I'm not sure what to do with the webapp since it seems to eschew the component architecture elsewhere in Socorro.

ezolenko:
Yeah, webapp would be a problem, not sure how to avoid modifying settings/base.py... 

As for correctness, the only changing behavior is the case where AWS keys are not provided in consul, so that defaults are used. This can happen in 3 cases -- non AWS deployment, AWS deployment using instance roles or deployment using non-AWS S3 compatible service that doesn't need credentials somehow.

First case can fail if None as a default value cause side effects (crashes in code not expecting None, etc). This is low probability, because boto will not be used at all.

Second case is the reason for the PR. It seems to work for me, unless I didn't stumble on an area that fails yet. Anyway, this case doesn't work as it is, so no regression here. 

Third case is possible (if there are such deployments anywhere), but a work around that restores old behavior is providing empty configuration values. Don't know how relevant this case might be.

The change is fairly simple -- use None instead of empty string as default value for AWS key/secret and make them optional in the unittest.

If this is unmergable, maybe just leave it for reference in case somebody needs it or if Mozilla's own deployment switches to IAM roles and this can get proper testing.

willkg:
We discussed this a bit yesterday and I don't have any good ways to test it, so I think I'm going to land it and we'll push it to -stage and if it works, we'll keep it. If it doesn't work, we'll back it out and figure out what to do then.

PR #3622, opened 2016-12-08
Bug 1305821 implement content security policies throughout webapp

peterbe:


g-k:
🎉 💯

can we add?

```python
CSP_REPORT_URI = (
    "/__cspreport__"
)
```

g-k:
@peterbe my bad that should be a string instead of a tuple (i.e. `CSP_REPORT_URI = "/__cspreport__"`). The report uri setting will send CSP violations to an nginx endpoint then the logging pipeline and a kibana dashboard.  We should be able to see XSS without CSP bypass there.

Some Addons inject CSP or trigger erroneous CSP reports.  For example, I have some violations without a violated directive for location leaderboard from noscript.

If you're worried about breaking things, you can set `CSP_REPORT_ONLY = True` to only send the violations but not block the requests or see if including some of the srcs from my ZAP session on prod would fix the problem: https://github.com/mozilla/socorro/compare/master...g-k:add-report-only-csp#diff-10b748dce4f615dd8e2dac2be124d8cc (this branch is better since csp_update lets you lock things down further by endpoint).

peterbe:
@adngdb It's sad that we have to allow [unsafe-eval](https://github.com/mozilla/socorro/pull/3622/files#diff-10b748dce4f615dd8e2dac2be124d8ccR750). It's because of our jquery tablesorter plugin. 
What I might do is to pause on this PR a bit and see if I can upgrade jquery tablesorter and maybe newer versions don't use `eval`. 

To begin with, it appears we have two different `jquery.tablesorter.js`:
https://github.com/mozilla/socorro/blob/master/webapp-django/crashstats/manage/static/manage/tablesorter/jquery.tablesorter.min.js and https://github.com/mozilla/socorro/blob/master/webapp-django/crashstats/crashstats/static/crashstats/js/jquery/plugins/jquery.tablesorter.js

The minified one doesn't have a version but the not-minified one is version `2.0.5b` and according to https://mottie.github.io/tablesorter/docs/ the latest version is 2.28.1. 

Lemme see if I can clean that up first.

g-k:
@peterbe It would definitely be good to remove the unsafe-eval src.  If this PR is close, it'd still be good to land it to protect against XSS from inline scripts and other srcs.

peterbe:
@g-k See https://github.com/mozilla/socorro/pull/3629#issue-195586493
*"This new version appears to not use eval so that's going to make our CSP headers happy. Haven't tested this yet."*

g-k:
@peterbe 👍 awesome!

adngdb:
Testing locally, this raises a few errors. One with Super Search and Signature report, about Google Analytics, which I believe is related to our pings when a user changes tab or runs a search (since that's sent from one of our JS files), though I'm not sure, and the other in /report/index about telemetry. 

So, r- for now until I fix these issues.

willkg:
The changes here seem fine as far as I can tell. This PR doesn't add any tests and our unit tests don't test these changes.

@adngdb has done some manual testing which brought up some issues.

Have we run the e2e tests? How else can we verify this is good?

adngdb:
I did not run the e2e tests, but they will be run once this hits stage. I think it's pretty hard to fully test this, as the 2 errors that I caught would definitely not have been noticed by our e2e tests. One was removing some non-critical functionalities of one of our JS libraries, and the other caused correlations to never be shown on the /report/index page. But it's pretty difficult to have a test case for that, since correlation data comes from an external source over which we have no control. 

Thus, I am not sure how we can automatically test this, and believe we might be good with just manual testing.

willkg:
@adngdb Ok--good to know.

I'm with you--let's land this and then clean up the mess that results if there is one.

PR #3623, opened 2016-12-09
fixes bug 1322615 - Don't look for status messages in JSON endpoints

peterbe:
Now, when I do `curl http://socorro.dev/monitoring/healthcheck/?elb=true` it doesn't need to do any SQL queries.

### DIFF
--- a/webapp-django/crashstats/status/context_processors.py
+++ b/webapp-django/crashstats/status/context_processors.py
@@ -2,19 +2,22 @@ from crashstats.status.models import StatusMessage
 
 
 def status_message(request):
-    statuses = StatusMessage.objects.filter(enabled=True)
 
-    if not statuses.exists():
-        return {}
+    def inner():
+        statuses = StatusMessage.objects.filter(enabled=True)
 
-    messages = []
-    for status in statuses.order_by('-created_at'):
-        messages.append({
-            'text': status.message,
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/status/context_processors.py
```diff
@@ -2,19 +2,22 @@
...
+            return {}
+
+        messages = []
+        for status in statuses.order_by('-created_at'):
+            messages.append({
+                'text': status.message,
+                'severity': status.severity,
+                'date': status.created_at,
+            })
+        return messages
```
willkg: Seems like we really want to call `statuses.order_by('-created_at').values('message', 'severity', 'created_at')` and skip this for loop. Then the SQL query is only performed if it's used in the template and you don't need to create an `inner()` function that only gets called in the template and all that.

Does that sound right?

peterbe: In fact, we don't even need the `.values()`. It could just return a regular QuerySet. 
The idea of doing a closure is generally a good one because often not everything inside the context processor is as lazy as a QuerySet.

peterbe:
Updated.

willkg:
I like this--looks good to me. r+

adngdb:
This is so much better!

PR #3625, opened 2016-12-13
fixes bug 1323123 - unable to switch result size in top crasher

peterbe:
Now this happens instead:

screen shot 2016-12-13 at 1 05 04 pm

### DIFF
--- a/webapp-django/crashstats/topcrashers/forms.py
+++ b/webapp-django/crashstats/topcrashers/forms.py
@@ -1,0 +1,14 @@
+from django import forms
+
+from crashstats.supersearch.form_fields import MultipleValueField
+from crashstats.crashstats.forms import BaseForm
+
+
+class TopCrashersForm(BaseForm):
+    product = forms.CharField()
+    version = MultipleValueField(required=False)
+    process_type = forms.CharField(required=False)
+    platform = forms.CharField(required=False)
+    _facets_size = forms.IntegerField(required=False)
+    _tcbs_mode = forms.CharField(required=False)
+    _range_type = forms.CharField(required=False)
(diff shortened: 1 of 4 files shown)


webapp-django/crashstats/topcrashers/forms.py
```diff
@@ -0,0 +1,14 @@
+from django import forms
+
+from crashstats.supersearch.form_fields import MultipleValueField
```
peterbe: Perhaps we should refactor this and move this class into `crashstats/base/` since it's share between at least two apps (`supersearch` and `topcrashers`)

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -180,21 +182,25 @@ def get_topcrashers_results(**kwargs):
...
+    range_type = form.cleaned_data['_range_type']
 
     range_type = 'build' if range_type == 'build' else 'report'
 
     if not tcbs_mode or tcbs_mode not in ('realtime', 'byday'):
         tcbs_mode = 'realtime'
 
     if product not in context['active_versions']:
-        raise http.Http404('Unrecognized product')
+        return http.HttpResponseBadRequest('Unrecognized product')
```
peterbe: This was slightly off-topic but I saw it and it was easy to fix. 
We shouldn't say that the page doesn't exist when it did but with the parameters being wrong.

webapp-django/crashstats/topcrashers/forms.py
peterbe: Perhaps we can do that later so as to not cloud the PR with more refactoring.

PR #3626, opened 2016-12-13
fixes bug 1323235 - Validate input on _histogram_interval.date

peterbe:
I first thought I could validate the input early in the webapp but when I saw [the documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#time-units) and decided it's too risky to get it right. Better to ask for forgiveness than "seek permission". 

Here's what you get now:
screen shot 2016-12-13 at 2 03 28 pm

### DIFF
--- a/socorro/unittest/external/es/test_supersearch.py
+++ b/socorro/unittest/external/es/test_supersearch.py
@@ -1486,6 +1486,19 @@ class IntegrationTestSuperSearch(ElasticsearchTestCase):
         )
 
     @minimum_es_version('1.0')
+    def test_get_with_date_histogram_with_bad_interval(self):
+        kwargs = {
+            '_histogram.date': ['product', 'platform'],
+            'signature': '!=crash_me_I_m_famous',
+            '_histogram_interval.date': 'xdays',  # Note! It's just wrong
+        }
+        assert_raises(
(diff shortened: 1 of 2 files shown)


adngdb:
I think this is the wrong way of doing it. We have a list of all the fields and should use that for sanitation instead of waiting for Elasticsearch to fail. 

I'm keeping this open until I come up with a better solution.

adngdb:
Scratch that, I'm stupid. I misunderstood the problem. Let me revisit this.

socorro/unittest/external/es/test_supersearch.py
```diff
@@ -1486,6 +1486,19 @@ def dt_to_midnight(date):
...
+        kwargs = {
+            '_histogram.date': ['product', 'platform'],
+            'signature': '!=crash_me_I_m_famous',
+            '_histogram_interval.date': 'xdays',  # Note! It's just wrong
+        }
+        assert_raises(
+            BadArgumentError,
+            self.api.get,
+            **kwargs
+        )
```
adngdb: Can you also verify that the exception has the correct key?

peterbe:
@adngdb Test updated now.

PR #3627, opened 2016-12-13
fixes bug 1304128 - eliminate inline JS on supersearch_fields

peterbe:


peterbe:
Note: Seems we moved this into `supersearch_fields.js` but forgot to remove the inline part.

PR #3628, opened 2016-12-13
fixes bug 1323314 - Excessive whitespace from pagination.html

peterbe:
Now the search result page is only ~300KB instead of ~700KB each time. 
(Only applicable when there's a LOT of pages possible)


willkg:
Looks good to me! r+

PR #3629, opened 2016-12-14
fixes bug 1323278 - Upgrade jQuery Tablesorter

peterbe:
I bet there's very little code changes to review. Best to just test it by clicking around. 

Notes...:

* There is now only 2 files. I chose the un-minified .js version. Makes it easier to debug, grep and it has a version number in the first comment. Also, our infra is good for doing this with yuglify. 
* I tested with and without DEBUG to make sure it minifies correctly. 
* Annoyingly there are CSS hacks in both `screen.less` and `tables.less`. I didn't want to tackle that right now but I don't like that it's fragmented so much. 
* We used to have 2 tablesorter. One in `crashstats` and one in `manage`. Now it's just 2 files and I put them in `base`. The new `.css` uses data urls to get the little sort arrows. 
* I renamed it in `bundles.py` to avoid accidentally using the old one. Also, the bundle used to be called `jquery_tablesorter` and don't think the name `jquery` is important enough to have in the name. Or else we'd need to do that to all other libs we use that require jQuery. 
* This new version appears to **not** use `eval` so that's going to make our CSP headers happy. Haven't tested this yet. 
* Right or wrong, I tried to make the new tablesorter CSS appear like it used to appear. For example, I had to set `margin: 0 !important` so that we get the margin we used to have around the tables.


adngdb:
This is conflicting now.

adngdb:
:+1:

peterbe:
Rebased now.

PR #3631, opened 2016-12-14
fixes bug 1323519 - Remove correlations from Top Crashers

peterbe:
Now, the only thing remaining is the `/api/Correlations/` endpoint. It still works but see no reason to keep it.


webapp-django/crashstats/crashstats/forms.py
```diff
@@ -173,82 +172,6 @@ def __init__(self, *args, **kwargs):
         ]
 
 
-class CorrelationsJSONFormBase(BaseForm):
-    correlation_report_types = forms.MultipleChoiceField(
-        required=True,
-        choices=make_choices(settings.CORRELATION_REPORT_TYPES),
```
adngdb: You can remove that setting now.

peterbe:
And rebased.

PR #3632, opened 2016-12-15
Fixes bug 1319896 - Increased the number of shards per index to 10.

adngdb:
This is a "better" version of #3620, see my commit for more explanations. 

I suggest that, if this PR is fine and gets an r+, we merge it and proceed with it, and close #3620. If this needs more work, then we should merge #3620 instead and land this later on as an improvement. 

cc @lonnen @milescrabill

### DIFF
--- a/socorro/external/es/base.py
+++ b/socorro/external/es/base.py
@@ -39,6 +39,12 @@ class ElasticsearchConfig(App):
     )
     # shared and not specifically in the elasticsearch config
     required_config.add_option(
+        'index_creator_class',
+        doc='a class that can create Elasticsearch indices',
+        default='socorro.external.es.index_creator.IndexCreator',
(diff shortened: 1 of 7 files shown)


willkg:
This looks ok to me as far as I can tell.

Do the existing tests cover the fundamental changes here? For example, if we typoed the `number_of_shards` key or provided a bad value, would a test fail?

adngdb:
We do test creating an index. However I don't think that we verify that the index is correctly configured, so we don't check that the number of shards is correct.

willkg:
If we test index creation, that's probably fine. If ES didn't like the parameter or value, pretty sure it'll complain at that point.

r+ on this with the caveat that when we push this to -stage, we verify correctness there.

PR #3635, opened 2016-12-15
Bug 1291173 - Added a rule to extract key measurements from memory_report.

adngdb:
This is based on work by @nnethercote, see [bug 1291173](https://bugzilla.mozilla.org/show_bug.cgi?id=1291173) for more context. 

Assigning to Will because you seem to be willing to review processor stuff. Feel free to bounce it to Peter if you think that's incorrect. :)

### DIFF
--- a/socorro/processor/rules/memory_report_extraction.py
+++ b/socorro/processor/rules/memory_report_extraction.py
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from socorrolib.lib.transform_rules import Rule
+from socorro.lib.transform_rules import Rule
 
 
 UNITS_BYTES = 0
--- a/socorro/processor/rules/memory_report_extraction.py
+++ b/socorro/processor/rules/memory_report_extraction.py
@@ -19,9 +19,16 @@ class MemoryReportExtraction(Rule):
 
     def _predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
         try:
+            # Verify that...
             return (
+                # ... we have a pid...
+                'pid' in processed_crash['json_dump'] and
+                # ... we have a memory report...
                 bool(processed_crash['memory_report']) and
-                'pid' in processed_crash['json_dump']
+                # ... and that memory report is recognisable.
+                'version' in processed_crash['memory_report'] and
+                'reports' in processed_crash['memory_report'] and
+                'hasMozMallocUsableSize' in processed_crash['memory_report']
             )
         except KeyError:
             return False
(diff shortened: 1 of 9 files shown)


socorro/processor/rules/memory_report_extraction.py
```diff
@@ -0,0 +1,132 @@
...
+        processed_crash['memory_measures'] = measures
+        return True
+
+    def _get_memory_measures(self, memory_report, pid):
+        if (
+            'version' not in memory_report or
+            'reports' not in memory_report or
+            'hasMozMallocUsableSize' not in memory_report
+        ):
+            raise ValueError('not a recognisable memory reports')
```
willkg: At this point, we know the rule should be running because this is the kind of crash it should run on. Why would we run into this situation? Does this mean that the memory report is malformed?

Why wouldn't we test to see if the memory report has all the things we need in the `_predicate`?

socorro/processor/rules/memory_report_extraction.py
```diff
@@ -0,0 +1,132 @@
...
+        except KeyError:
+            return False
+
+    def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
+        pid = processed_crash['json_dump']['pid']
+        memory_report = processed_crash['memory_report']
+
+        try:
+            measures = self._get_memory_measures(memory_report, pid)
+        except ValueError as e:
```
willkg: `_get_memory_measures` looks like it's doing a lot of dict lookups. Given we don't control what is in the memory report and it could theoretically change at any time without us knowing, I we should catch `KeyError` here, too.

socorro/processor/rules/memory_report_extraction.py
```diff
@@ -0,0 +1,132 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from socorro.lib.transform_rules import Rule
+
+
+UNITS_BYTES = 0
+KIND_NONHEAP = 0
+KIND_HEAP = 1
```
willkg: `KIND_NONHEAP` and `KIND_HEAP` appear to be special values in the memory report. Is there a definition of those values or a specification somewhere? If there is, I'd toss the link in the module docstring or something like that so it's easier to find later.

socorro/unittest/processor/rules/memory_reports/bad_units.json
```diff
@@ -0,0 +1,15 @@
+{
+ "version": 1,
+ "hasMozMallocUsableSize": true,
+ "reports": [
+  {
+   "process": "Main Process (pid 11620)",
+   "path": "explicit/foo",
+   "kind": 2,
```
willkg: I think this has a bad units and a bad kind.

willkg: To clarify, if that's correct, then we should fix the kind so that this test case only has bad units.

willkg:
Also, sorry about not getting to this earlier. I plum forgot.

socorro/processor/rules/memory_report_extraction.py
adngdb: @nnethercote do you know?

adngdb: I sent an email to Nicholas, I'm expecting to have docs here soon/ :)

adngdb:
This is now ready to be merged, just waiting for travis to go green.

PR #3643, opened 2016-12-21
fixes bug 1325183 - Remove happybase requirement

peterbe:


willkg:
I double-checked to see if we're using this anywhere and we're not. Ergo, r+.

PR #3650, opened 2017-01-17
Fixes bug 1331659 - Upgraded elasticsearch libraries to latest compat…

adngdb:
…ible versions.


socorro/unittest/external/es/base.py
```diff
@@ -123,14 +117,11 @@
         'is_returned': True,
         'is_mandatory': False,
         'storage_mapping': {
-            'type': 'multi_field',
+            'type': 'string',
             'fields': {
-                'os_name': {
-                    'type': 'string'
-                },
```
willkg: Can you talk about this change? It looks like we're switching it from a `multi_field` to a `string`, but I'm not sure what this went from and to.

socorro/external/es/super_search_fields.py
```diff
@@ -380,9 +380,6 @@ def add_field_to_properties(properties, namespaces, field):
                 '_all': {
                     'enabled': False,
                 },
-                '_source': {
-                    'compress': True,
-                },
```
willkg: Can you talk a bit about this? Looks like we want to store the source data compressed--probably to save space. Do the newer versions default to that? Does this affect our space estimations?

adngdb: This was a leftover from my first tests with ES 5.x, where that feature has been removed. It still exists in 2.x, so I'm going to put it back. Thanks for spotting that!

socorro/unittest/external/es/base.py
adngdb: Yes, this is a mapping change that happened in 1.x and has been removed from 2.x forward. This was also some leftover from my work on 5.x, but I believe it is a fine thing to keep. 

So, the new mapping format for multi-fields is this: define the default field in the root, and then add a ``fields`` key in which you define other sub-fields. The two syntaxes are strictly equivalent, but the new one is future-proof.

willkg:
Ok--this sounds good to me.

What's the plan for merging this? Is it something like:

1. merge it, let it sit on -stage for a few days and an index changeover
2. verify things are working
3. if they are, push to -prod

If so, then we should wait to merge this until after the -prod deploy today and let it sit on -stage for the next week.

Does that sound right?

adngdb:
We don't need to wait for a new index to be created. The changes in mapping only happened in unit tests, so our ES database shouldn't be impacted at all.

PR #3653, opened 2017-01-18
[DO NOT MERGE] Bug 1311647 - Add more fields to telemetry crash reports

marco-c:
I've tested the schema with the `validate_and_test.py` script (also with 1000 reports instead of 100, just to be sure).

### DIFF
--- a/socorro/schemas/crash_report.json
+++ b/socorro/schemas/crash_report.json
@@ -43,6 +43,30 @@
         }
     },
     "properties": {
+        "abort_message": {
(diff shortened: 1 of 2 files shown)


willkg:
Adding a "do not merge" thing so that we don't accidentally merge it until the work specified in https://bugzilla.mozilla.org/show_bug.cgi?id=1311647#c8 occurs.

lonnen:
subset of 3663, recommend closing

adngdb:
Closing in favor of #3663.

PR #3656, opened 2017-01-19
Fixes bug 1325086 - Removed CrashesByExploitability model from the pu…

adngdb:
…blic API.


willkg:
The tests pass and the code changes look fine as far as I can tell.

What kinds of things do we do to review this kind of middleware removal change?

adngdb:
I don't really know. I personally just verify on stage that nothing breaks (but that's pretty much what the e2e tests do), and rely on unit tests otherwise. This is removing a feature that is not being used anywhere, so it's hard to test anything other than unexpected side effects.

willkg:
Ok. This looks good. r+!

PR #3660, opened 2017-01-24
Fixes bug 1331726 - fix assertion in test_public_api_navigation

willkg:
This fixes the assertion that we should have at least one signature first date
for each signature--that turns out to not necessarily be the case. I changed it
to make sure it's > 0. Hopefully we don't eventually find ourselves in a
situation where it's 0, though I suspect that's possible.

While doing that, I changed some variable names to make the code easier to read
and I added comments to summarize what's about to happen so it's easier to
verify the test is doing what it should be doing.

r?

### DIFF
--- a/e2e-tests/tests/test_api.py
+++ b/e2e-tests/tests/test_api.py
@@ -10,9 +10,9 @@ import pytest
 
 class TestAPI:
 
-    @pytest.mark.skip('Bug 1331726 - test_public_api_navigation is sporadically failing')
     @pytest.mark.nondestructive
     def test_public_api_navigation(self, base_url):
+        # Request a list of all the products and versions
         response = requests.get(base_url + '/api/ProductVersions/', {
             'active': True,
         })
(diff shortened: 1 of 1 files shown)


e2e-tests/tests/test_api.py
```diff
@@ -73,32 +76,33 @@ def test_public_api_navigation(self, base_url):
...
         results = response.json()
-        first_dates = {}
+
+        signature_to_first_date = {}
         for hit in results['hits']:
-            first_dates[hit['signature']] = hit['first_date']
+            signature_to_first_date[hit['signature']] = hit['first_date']
 
-        assert len(first_dates) == len(signatures)
+        assert 0 < len(signature_to_first_date) <= len(uuid_to_signature)
```
willkg: This is the assertion that was failing and a little bit above, I tweaked the comment about what this section is doing and looking for.

@adngdb Is this in line with what you were thinking?

PR #3661, opened 2017-01-25
Bug 1333577 - skip multiline search test

m8ttyB:


willkg:
Looks good and works right. r+!

PR #3662, opened 2017-01-26
Fixes bug 1333577 - Fixed Super Search e2e test.

adngdb:


e2e-tests/README.md
```diff
@@ -25,11 +25,18 @@ rest of Socorro so you're not mixing requirements:
 	$ # your dev virtualenv.
 	$ pip install -r requirements.txt
 
-An additional constraint for Firefox users, with the release of Firefox 48, Webdriver support is currently broken until GeckoDriver is out of Alpha. We suggest using an older version of Firefox which can be [downloaded here][firefoxdownloads].
+An additional constraint for Firefox users, with the release of Firefox 48, Webdriver support is currently broken until GeckoDriver is out of Alpha. We suggest using an older version of Firefox which can be [downloaded here][firefoxdownloads]. Note that on Linux, you don't need to downgrade your Firefox install in order to use an older version.
```
m8ttyB: We should correct our docs, GeckoDriver is not alpha and is ready for use. If we come across any issues we should be filing issues against that project.

e2e-tests/README.md
```diff
@@ -25,11 +25,18 @@ rest of Socorro so you're not mixing requirements:
...
 
 If you have multiple versions of Firefox installed, you can specifiy a specific one by using the `--firefox-path <path to firefox binary>` flag.
 
-___Running the tests on stage___
+___Using an older version of Firefox on Linux___
```
m8ttyB: I think we should remove this section. We should default to testing against Firefox latest when possible.

e2e-tests/README.md
adngdb: I'm surprised by that. What version of Firefox / Selenium are compatible now? I'm testing on Ubuntu 16.04 with Firefox 50.0.1 and Selenium 3.0.2, and I get this error: 

```bash
$ py.test --driver="Firefox" --base-url="http://localhost:8000" -x tests/test_search.py
======================================= test session starts =======================================
platform linux2 -- Python 2.7.12, pytest-3.0.4, py-1.4.31, pluggy-0.4.0
[...]
WebDriverException: Message: Can't load the profile. Profile Dir: /tmp/tmpAg7sH3 If you specified a log_file in the FirefoxBinary constructor, check it for details.
```

m8ttyB: Do you have GeckoDriver installed?

m8ttyB:
There's a known behavior with the selenium client where we'll need to add `--capability marionette true` as a cmd line arg with when using pytest. Alternatively we can add this to conftest.py. In theory this won't affect Firefox < 48 or other drivers but we may wish to validate that assertion.

```@pytest.fixture
def capabilities(request, capabilities):
    driver = request.config.getoption('driver')
    if capabilities.get('browserName', driver).lower() == 'firefox':
        # In order to run these tests in Firefox 48, marionette is required
        capabilities['marionette'] = True
    return capabilities```

e2e-tests/README.md
m8ttyB: Selenium is compatible with most versions of Firefox. With non-legacy versions, Firefox 48 and above, the new goodness is in using geckodriver. For older versions we can continue to use the selenium client without geckodriver

In our instance, we likely only care to test crash-stats using the latest version of Firefox.

willkg:
@m8ttyB I'd never heard of geckodriver. Knowing about that would have saved a lot of work over the last month on getting the e2e tests usable for @peterbe, @adngdb and myself.

@adngdb The geckodriver releases are here: https://github.com/mozilla/geckodriver/releases I'll try it out on Monday.

adngdb:
I have refactored this to be focused on the problem at hand. Further work about GeckoDriver will be done in [bug 1335027](https://bugzilla.mozilla.org/show_bug.cgi?id=1335027).

willkg:
Looks good and tests pass. r+

m8ttyB:
@willkg apologies, I noted it to @peterbe but at the time the focus was on phantomjs and not driving the tests using Firefox (or another browser for that matter).

PR #3669, opened 2017-02-14
Bug 1338565 - webapp: increase uwsgi buffer-size 4096 (default) => 16384

milescrabill:
@adngdb r?

This is a little bit of a funky way of doing this, but we aren't already using a config file for uwsgi, so there's nowhere else to add it without creating and using one.


adngdb:
It looks good to me, I guess. I don't know anything about those files so I trust that you know the syntax. And that number looks good. :)

ezolenko2:
Adding socorro/webapp-django/UWSGI_BUFFER_SIZE=number in consul should work too btw (I'm using it for other uwsgi params)

milescrabill:
Oh, that's a good call @ezolenko2. Probably a better way to do things, though hard to document.

PR #3670, opened 2017-02-15
stackwalker: Fix compilation with upstream breakpad

ePirat:
Recently on breakpad upstream there was a parameter added to the
StackFrameSymbolizer::FillSourceLineInfo method, therefore compilation
would fail.

https://chromium.googlesource.com/breakpad/breakpad/+/0924d424e444d57dd95c647652a11f2d655c11a0

This commit passes NULL for the unloaded_modules parameter to fix the
compilation failure. It might be useful to actually use that parameter
in the future though.

### DIFF
--- a/minidump-stackwalk/get-minidump-instructions.cc
+++ b/minidump-stackwalk/get-minidump-instructions.cc
@@ -415,6 +415,7 @@ int main(int argc, char** argv)
       if (p && (p - line == 8 || p - line == 16) && !strstr(line, "<.data>:")) {
         frame.instruction = strtoull(line, nullptr, 16);
         symbolizer.FillSourceLineInfo(module_list,
+              /* unloaded_modules= */ NULL,
                                       &system_info,
                                       &frame);
         print_frame(frame, last_frame);
--- a/minidump-stackwalk/stackwalker.cc
+++ b/minidump-stackwalk/stackwalker.cc
@@ -138,6 +138,7 @@ public:
                                               StackFrame* stack_frame) {
     SymbolizerResult res =
       StackFrameSymbolizer::FillSourceLineInfo(modules,
+                       /* unloaded_modules= */ NULL,
                                                system_info,
                                                stack_frame);
     RecordResult(stack_frame->module, res);


adngdb:
It seems your work is breaking our build, see the logs: https://travis-ci.org/mozilla/socorro/builds/201991842

ePirat:
Yes indeed, thats because the CI is using an older Breakpad snapshot, that does not has the change I mentioned.

willkg:
I'm going to r- this and close it out.

The stackwalker in Socorro is specifically targeted for Socorro/Mozilla use and it's tied to the breakpad snapshot that we use to build it. Ergo, if you wanted to make the changes you're making, we'd need to update the snapshot and a bunch of other stuff across Firefox development. I think that should start off with a bug in Bugzilla because it requires coordination, etc.

willkg:
I talked with @luser briefly. Turns out we don't need to coordinate updating the snapshot with anyone else.

So this is breaking CI. I don't know what other effects it has. I don't know how to move forward with it at the moment.

I have to jet for a bit. I'll follow up later today or this week.

luser:
So yeah, we do need to update the snapshot to take this patch, since otherwise our CI will be broken. The last snapshot update happened in [bug 1262135](https://bugzilla.mozilla.org/show_bug.cgi?id=1262135). Unfortunately we don't have a fully-automated process for doing that.

@willkg and I talked about this on IRC--I'm going to submit a different PR that will let us avoid this situation for now. The problem is that our CI builds from the snapshot, which is an older rev of Breakpad, but local builds that don't use the snapshot build from the tip of Breakpad, so breaking changes upstream are bad for local builds. I'm going to change the build-breakpad script to pin the revision of Breakpad in use to the one in the snapshot, so local builds should at least work the same as CI builds. This is good dev hygiene in general.

Updating the snapshot is still a manual process, and we can still do that, at which point we'll also need to take this patch, but we can do that at another time.

luser:
That PR is https://github.com/mozilla/socorro/pull/3672

willkg:
I just landed #3672, so I think we can close this out.

PR #3672, opened 2017-02-16
Pin the revision of Breakpad to build in build-breakpad.sh to the revision used in the snapshot.

luser:
This fixes local builds that don't use the Breakpad snapshot because
upstream Breakpad made a breaking API change.


willkg:
Wow--this took ages to test for a bunch of lousy reasons on my part.

It looks good to me. r+

PR #3676, opened 2017-02-23
[DNM] Fixes bug 1342081 - Upgraded ES libraries and related code.

adngdb:
This commit upgrades our ES modules, ``elasticsearch`` and ``elasticsearch-dsl``, as well as all code related to them. It makes sure that unit tests pass using ES 5.1 as the database.


scripts/start-travis-elasticsearch.sh
```diff
@@ -4,12 +4,9 @@ set -ex
...
-tar zxf elasticsearch-1.4.5.tar.gz
-echo "script.disable_dynamic: false" > elasticsearch.yml
-echo "index.number_of_shards: 1" >> elasticsearch.yml
-echo "index.number_of_replicas: 0" >> elasticsearch.yml
-# the -d flag starts it as a daemon in the background
-./elasticsearch-1.4.5/bin/elasticsearch -d --config=elasticsearch.yml
+wget --no-check-certificate 'https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-5.1.1.deb'
+sudo dpkg -i --force-confnew elasticsearch-5.1.1.deb
+
+sudo service elasticsearch start
```
willkg: Do we want to stomp on values in the yml file at all before running tests?

socorro/external/es/connection_context.py
```diff
@@ -129,6 +130,12 @@ def indices_client(self, name=None):
             elasticsearch.client.IndicesClient(self.connection())
         )
 
+    def cat_client(self):
```
willkg: What does this do? Can you add a docstring?

The elasticsearch-py docs aren't helpful at all:

https://elasticsearch-py.readthedocs.io/en/master/api.html?highlight=catclient#cat

socorro/unittest/external/es/base.py
```diff
@@ -172,12 +170,17 @@
         'form_field_choices': None,
         'permissions_needed': [],
         'default_value': None,
-        'has_full_version': False,
+        'has_full_version': True,
```
willkg: Is this a memory address? If so, why would we want to store it as `text`?

socorro/unittest/external/es/base.py
```diff
@@ -135,12 +128,17 @@
...
         'is_exposed': True,
         'is_returned': True,
         'is_mandatory': False,
         'storage_mapping': {
-            'type': 'string'
+            'type': 'text',
```
willkg: Why would we want to store the `release_channel` as text?

socorro/unittest/external/es/base.py
```diff
@@ -284,7 +289,12 @@
         'permissions_needed': [],
         'query_type': 'enum',
         'storage_mapping': {
-            'type': 'string'
+            'type': 'text',
```
willkg: Why would we want to store this as text?

socorro/unittest/external/es/base.py
```diff
@@ -301,11 +311,10 @@
         'permissions_needed': [],
         'query_type': 'string',
         'storage_mapping': {
-            'type': 'string',
+            'type': 'text',
             'fields': {
                 'full': {
-                    'type': 'string',
-                    'index': 'not_analyzed',
+                    'type': 'keyword',
```
willkg: Why would we want to store this as keyword?

socorro/unittest/external/es/base.py
```diff
@@ -377,7 +384,12 @@
         'permissions_needed': [],
         'query_type': 'enum',
         'storage_mapping': {
-            'type': 'string'
+            'type': 'text',
```
willkg: Why would we want to store this as text?

socorro/unittest/external/es/base.py
```diff
@@ -463,11 +474,10 @@
         'permissions_needed': [],
         'query_type': 'enum',
         'storage_mapping': {
-            'type': 'string',
+            'type': 'text',
```
willkg: Why would we want to store this as text?

socorro/unittest/external/es/base.py
```diff
@@ -509,11 +518,10 @@
...
```
willkg: Why would we want to store this as text?

socorro/unittest/external/es/base.py
```diff
@@ -654,14 +664,19 @@
         ],
         'query_type': 'enum',
         'storage_mapping': {
-            'type': 'string'
+            'type': 'text',
```
willkg: Why would we want to store this as text?

socorro/unittest/external/es/test_supersearch.py
```diff
@@ -1569,33 +1550,33 @@ def test_get_with_number_histogram(self):
...
                             'term': 'Linux',
                             'count': 1
                         }
                     ],
+                    'product': [
+                        {
+                            'term': 'NightTrain',
+                            'count': 1
+                        },
+                    ],
```
willkg: Why did this move? Is it a cosmetic change?

scripts/start-travis-elasticsearch.sh
adngdb: I think we're fine using the default values. Anyway, we can't set the config options we use to anymore, because ES won't accept those values in the yml file.

socorro/external/es/mappings/correlations_index_settings.json
```diff
@@ -4,9 +4,6 @@
             "_all": {
                 "enabled": false
             },
-            "_source": {
-                "compress": true
-            },
```
adngdb: Yes. It's not just that we can, it's that the option is not accepted anymore, so we have to remove it. :)

socorro/external/es/new_crash_source.py
```diff
@@ -53,34 +54,30 @@ def new_crashes(self, date, product, versions):
...
+        filter_args = {
+            'processed_crash.date_processed': {
+                'gte': date.isoformat(),
+                'lt': next_day.isoformat(),
+            },
+        }
+        search = search.filter('range', **filter_args)
+
+        # Turn that search body into a dict that we can send to Elasticsearch.
+        query = search.to_dict()
```
adngdb: Yes, by default the library applies an ``AND`` boolean operator between filters. So all of the filters above must be true in the query.

socorro/external/es/new_crash_source.py
```diff
@@ -92,11 +89,11 @@ def new_crashes(self, date, product, versions):
                 scroll='2m',  # keep the "scroll" connection open for 2 minutes
                 index=es_index,
                 doc_type=es_doctype,
-                fields=['crash_id'],
                 query=query,
+                _source=False,
             )
             for hit in res:
-                crash_ids.append(hit['fields']['crash_id'][0])
+                crash_ids.append(hit['_id'])
```
adngdb: It's always been the case, and I used this opportunity to make this a little more efficient by not needing any field to be returned.

socorro/middleware/search_common.py
```diff
@@ -135,7 +135,7 @@ def build_filters(self, fields):
                 ))
 
                 # Add an interval field.
-                default_interval = 1
+                default_interval = 50
```
adngdb: That's the Elasticsearch default now. This change is a security against the "matrix" problem I had while testing. If we set it to 1, then any histogram over a field that contains big numbers will return a huge number of results and will cause out-of-memory problems. By increasing the default interval we can mitigate that problem and ensure users won't accidentally hit it.

socorro/unittest/external/es/test_supersearch.py
adngdb: No, that's just because I was stupid when debugging a test failure and thought it was an array and not a dict. :)

socorro/external/es/supersearch.py
```diff
@@ -99,19 +99,20 @@ def format_field_names(self, hit):
...
-                    hit[field] = hit[field][0]
+        def flatten(d, parent_key='', sep='.'):
+            items = []
+            for k, v in d.items():
+                new_key = parent_key + sep + k if parent_key else k
+                if isinstance(v, collections.MutableMapping):
+                    items.extend(flatten(v, new_key, sep=sep).items())
+                else:
+                    items.append((new_key, v))
+            return dict(items)
```
adngdb: You're right, so I started moving it to ``socorro/lib/util.py``, and I feel like I've open Pandora's box... That file is a huge mess, I'm trying to clean it up, that will add a bunch of code changes to this PR but it will be in a separate commit.

socorro/unittest/external/es/base.py
adngdb: This multi-fields storage serves a simple purpose: when searching, we want to have access to the content separated as terms, and when aggregating, we want to see it as a single block. In this case, take a signature like ``oom | ns::something<T> | randomFunc``. We want it to be stored as ``[oom, ns, something, randomfunc]`` to be able to retrieve it when doing ``<signature> <has terms> <oom>`` (that's why we store it as ``text``) but when showing a list of top signatures, we want to see it as ``oom | ns::something<T> | randomFunc`` (that's why we have a second storage as a ``keyword``, under the ``.full`` key name). Super Search uses the ``has_full_version`` attribute to see if there's a ``.full`` sub-key, and use it in aggregation if so.

adngdb: Also, all of these are only ever used for unit tests. The actual mapping that we use in production are inside our database, in the Super Search Fields list. See https://crash-stats.mozilla.com/admin/supersearch-fields/

adngdb: However you were right that some of those choices I made where dumb, so I went through all of them and fixed a few. Now all of them should be coherent with the data we have in production.

willkg: Awesome--this makes total sense.

scripts/start-travis-elasticsearch.sh
willkg: Makes sense!

socorro/middleware/search_common.py
willkg: Makes sense!

socorro/external/es/new_crash_source.py
willkg: Makes sense!

socorro/lib/util.py
```diff
@@ -2,225 +2,229 @@
...
+                    "e": 3
+                }
+            }
+        }
+        OUT:
+        {
+            "a": 1,
+            "b.c": 2,
+            "b.d.e": 3
+        }
```
willkg: I tend to do this either using code blocks or doctests. One advantage of doctests is that you can test them and thus know when your examples are wrong. Both code blocks and doctests will get syntax highlighting when rendered in Sphinx.

I have no idea if this is rendered in Sphinx or not. If it is, it's worth changing.

Something like this:

```
For example:

>>> flatten({'a': 1, 'b': {'c': 2, 'd': {'e': 3}}})
{'a': 1, 'b.c': 2, 'b.d.e': 3}

```

willkg:
Everything here seems fine to me. r+

adngdb:
Just found out that we don't have any unit tests for the ``missing`` feature of Super Search. I need to add that (and fix the feature). Also, the schema still needs to be tested and validated more. Right now basic functionalities work, but I need to do further testing.

adngdb:
I have been testing locally with data from downloaded from prod, and with an ES 5.1 database and the latest version of the schema that is in this PR. Everything seems to be working fine now!

@willkg mind taking a look at the last commit I added? It turned out we were missing unit tests for the ``__null__`` operator.

scripts/data/supersearch_fields.json
```diff
@@ -338,9 +336,7 @@
         "is_returned": false,
         "query_type": "enum",
         "permissions_needed": [],
-        "storage_mapping": {
-            "type": "string"
-        },
+        "storage_mapping": null,
```
willkg: What does it mean to change the `storage_mapping` to null?

scripts/data/supersearch_fields.json
```diff
@@ -1207,26 +1195,6 @@
...
-        },
-        "namespace": "processed_crash",
-        "name": "dump",
-        "has_full_version": false,
-        "data_validation_type": "str",
-        "in_database_name": "dump",
-        "is_exposed": false,
-        "form_field_choices": [],
-        "description": ""
-    },
```
willkg: This dumps the `dump` field. Why? Does anything use the `dump` field?

scripts/data/supersearch_fields.json
```diff
@@ -2865,29 +2811,27 @@
                     "dynamic": "true",
                     "properties": {
                         "classification_version": {
-                            "type": "string",
-                            "analyzer": "keyword"
+                            "type": "keyword"
                         },
                         "classification_data": {
-                            "type": "string"
+                            "type": "text"
```
willkg: What does the `classification_data` field hold?

scripts/data/supersearch_fields.json
```diff
@@ -3450,7 +3386,7 @@
         "query_type": "enum",
         "permissions_needed": [],
         "storage_mapping": {
-            "type": "string"
+            "type": "integer"
```
willkg: This switched from `string` to `integer`. Does that sound right? Are all arm version values strictly integers?

socorro/external/es/supersearch.py
```diff
@@ -324,7 +324,9 @@ def get(self, **kwargs):
                         operator_range[param.operator]: param.value
                     }
                 elif param.operator == '__null__':
-                    filter_type = 'missing'
+                    filter_type = 'exists'
+                    # Need to reverse the 'not' operator.
+                    param.operator_not = not param.operator_not
```
willkg: Clever!

willkg:
Ugh. The GH interface for reviews totally sucks if you're looking at specific commits. That's frustrating.

I have questions in the second-to-last commit.

scripts/data/supersearch_fields.json
adngdb: That this field will not have a storage mapping in Elasticsearch. Basically the impact is very low, we will just rely on ES's default mechanism. I'm doing this here because that field is not exposed or returned in any way, it's an old field that we don't use anymore.

scripts/data/supersearch_fields.json
adngdb: That field doesn't exist anymore. It was a bit stupid to remove it there, it's out of scope, sorry about that.

scripts/data/supersearch_fields.json
adngdb: According to https://crash-stats.mozilla.com/search/?_facets=support_classification that field contains only integers.

scripts/data/supersearch_fields.json
adngdb: All I remember is that it has text about the classification. But that is not exposed anywhere, so it doesn't really matter.

willkg:
This looks good to me. r+

requirements.txt
```diff
@@ -59,12 +59,12 @@ psycopg2==2.6.2 \
...
-elasticsearch==1.9.0 \
-    --hash=sha256:cbee23c29f9947b1b78be5c0f741c1f8b504d3c53163b1355ea08f69cbd0fc0c \
-    --hash=sha256:eb0d4b5b20a4ccf9e49bd20498b6215d5e960d8a63a0eae44cd4431e3f303a24
-elasticsearch-dsl==0.0.11 \
-    --hash=sha256:663fb62ad39200c7d903e973aa0aa693578613264d83796455cbf4cd172bd878 \
-    --hash=sha256:59a76c4142478a1952bba6f9a9ca4fc7b029afb619e8ffcf0d135ce37ea692da
+elasticsearch==5.1.0 \
+    --hash=sha256:ee95918958508fd34e533f78d0790e0e5ee03d0942b8988d34e6b6c4467edf66 \
+    --hash=sha256:a0bc4508764aa14c0cd15840a0ab38020b68e17d699274417986e3da0cee82fa
+elasticsearch-dsl==5.1.0 \
```
peterbe: There is now a `5.2.0` of both `elasticsearch` and `elasticsearch-dsl` in pypi. Is now maybe a good time to push to that too?

adngdb:
Rebased and updated to be compatible with Elasticsearch 5.3, cc @milescrabill.

willkg:
This has conflicts. I'm going to see if I can address them now. Then I'll update the branch and we can squash and merge.

🎉

PR #3678, opened 2017-02-28
Fixes bug 1325087 - Removed implementation of the obsolete exploitabi…

adngdb:
…lity service.


willkg:
I read through this and it seems fine. I have a couple of questions because I don't know anything about this:

1. should this remove the postgres model `ExploitabilityReport`, too?
2. should this remove the webapp view `exploitability_report` which uses the `ExploitabilityReportForm`?

adngdb:
> should this remove the postgres model ExploitabilityReport, too?

Yes, ultimately, but I was going to do that afterwards, when I start attacking our database schema. Up until now I have been removing a lot of reports from the webapp / middleware without removing the corresponding data sources. Do you think that's a mistake? 

> should this remove the webapp view exploitability_report which uses the ExploitabilityReportForm?

Nope, that view still exists, it simply uses SuperSearch as a source of data now. The old view, that was based on postgres, has been removed in https://github.com/mozilla/socorro/commit/0dc0c2a8e9d234dc0481534ec9528fa9ad743433

willkg:
That sounds fine to me. r+!

PR #3681, opened 2017-03-01
Fixes bug 1343540 - Removed postgres implementations of TCBS services.

adngdb:


willkg:
This seems fine. I went through and made sure the modules and functions removed aren't used anywhere. Beyond that, I don't really know enough about this bug to review more. r+ with that caveat.

PR #3687, opened 2017-03-03
fixes bug 1344217 - let validate_and_test.py accept Super Search queries

peterbe:
I've tried it locally with a bunch of queries and it seems to work. 
One thing to look out for, the default number of hits returned by the SuperSearch API is 100. So if you paste in 5 URLs that all yield lots of hits, you're going to be testing the JSON Schema against 500 crashes. 

CC @adngdb

### DIFF
--- a/socorro/schemas/validate_and_test.py
+++ b/socorro/schemas/validate_and_test.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 from __future__ import print_function
 
+from urlparse import urlparse
 import json
 import os
 
@@ -32,7 +33,7 @@ class MockedTelemetryBotoS3CrashStorage(TelemetryBotoS3CrashStorage):
         self.combined = crash
 
 
-def main():
+def main(*urls):
     file_path = os.path.join(HERE, 'crash_report.json')
     with open(file_path) as f:
         schema = json.load(f)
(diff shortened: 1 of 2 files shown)


socorro/schemas/validate_and_test.py
```diff
@@ -32,50 +33,82 @@ def save_processed(self, crash):
...
+        assert '://' in url, url
+
+        # To make it easy for people, if someone pastes the URL
+        # of a regular SuperSearch page (and not the API URL), then
+        # automatically convert it for them.
+        parsed = urlparse(url)
+        if parsed.path == '/search/':
+            parsed = parsed._replace(path='/api/SuperSearch/')
+            parsed = parsed._replace(fragment=None)
+            url = parsed.geturl()
```
adngdb: This is convenient but might not work in some cases. Sadly the URLs between API and SuperSearch are not commutable. I don't think it's too bad, but it might happen that someone is going to get confused by a weird error or unexpected results here.

socorro/schemas/validate_and_test.py
```diff
@@ -32,50 +33,82 @@ def save_processed(self, crash):
...
+            params={
+                'product': 'Firefox',
+                '_columns': ['uuid'],
+                '_facets_size': 0
+            }
+        )
+        search = r.json()
+        uuids = [x['uuid'] for x in search['hits']]
+
+    assert len(uuids) == len(set(uuids))
```
adngdb: Huh? What's the point of this? Why not just make ``uuids`` a set to begin with? It's going to be very painful if you need to make searches that do not collide.

socorro/schemas/validate_and_test.py
peterbe: Interesting. I didn't know that. Care you enlighten me with some examples of that? 
I do vaguely remember we have some bugs where the API URL doesn't correctly reflect what you searched for but I thought that was related to facetting. 

Should we reopen this bug?

socorro/schemas/validate_and_test.py
peterbe: The point is self-check. I added it early and should probably have removed it later. Unless SuperSearch returned repeated UUID we should always have unique UUIDs here.

socorro/schemas/validate_and_test.py
marco-c: We could print an error instead of converting the URL for them, but I think the script will usually be used for simple searches (e.g. I used it only to find cases where a given field had a value).

PR #3688, opened 2017-03-03
Fixes bug 1291173 and bug 1311647 by readding the new fields back with correct types

marco-c:


peterbe:
Is this a bug in SuperSearch @adngdb ?
screen shot 2017-03-03 at 1 52 49 pm

peterbe:
I looked it up in processed crash. It's `0` so perhaps the column display in Super Search thinks it's falsy and chooses not to display it. 
screen shot 2017-03-03 at 1 57 02 pm

peterbe:
Also, it seems I can make a facet on "memory gfx textures"
screen shot 2017-03-03 at 2 00 01 pm

peterbe:
(Attention @adngdb)

Perhaps I'm being Friday but I can't figure out how to test this. 
I thought I'd make a SuperSearch query that gives me a bunch of crashes that has something in this new `memory_measures` field. 

See https://crash-stats.mozilla.com/search/?memory_vsize=%3E2030030848&product=Firefox&date=%3E%3D2017-02-24T18%3A51%3A00.000Z&date=%3C2017-03-03T18%3A51%3A00.000Z&_sort=-date&_facets=memory_vsize&_columns=memory_gfx_textures&_columns=memory_vsize#crash-reports

That should only include crashes with `memory_vsize > 2030030848`. But the first UUID returned by that query is: https://crash-stats.mozilla.com/api/ProcessedCrash/?crash_id=6c92e56e-8f1a-4e95-82d5-b7ed82170224 and that doesn't even have a `memory_measures` container. 

In other words, I can't figure out how to use SuperSearch to get crashes out that would check all these `memory_measures` keys. Please halp. (also, I'm done for the day as of now)

adngdb:
@peterbe Your first screenshot is a bug in the Web UI of Super Search, see the corresponding API call: https://crash-stats.mozilla.com/api/SuperSearch/?memory_gfx_textures=%3E%3D0&product=Firefox&date=%3E%3D2017-02-28T16%3A15%3A00.000Z&date=%3C2017-03-07T16%3A15%3A00.000Z&_columns=memory_gfx_textures

I cannot reproduce your second problem. The query you link to returns correct results for me. You can see that a crash report has memory measures in the report/index/ page, they are the ``MR: xxx`` lines in the Details tab. I think that query is fine to test the new memory fields.

peterbe:
@marco-c I don't suspect there's anything wrong with your schema changes. The types seem to be correct. However, either I'm being really dim or we have a bug in how send crashes to Telemetry. Let me debug this with @adngdb.

@adngdb I dug into the validate_and_test.py to see if I can list all the keys in the schema and compare that to ALL keys in all crashes it tests. It found that certain keys like `memory_measures.memory_private` is never tested. Hmm... Why, I wonder. 

Look at this super search field: https://crash-stats.mozilla.com/admin/supersearch-field/?name=memory_private This field doesn't get included in the combined crash object that we reduce by the JSON schema. The crash https://crash-stats.mozilla.com/api/ProcessedCrash/?crash_id=07794e51-0de1-4d3a-8c07-33ea72170309 does NOT contain a dict called `memory_measures` (Note if you open that URL with a logged in session, the whitelisting is ignored). I believe this is because it's not listed in Super Search Fields. But it's checked on both "Is returned" and "Is exposed". You can test this with:
```
curl https://crash-stats.mozilla.com/api/ProcessedCrash/\?crash_id\=07794e51-0de1-4d3a-8c07-33ea72170309 | grep private
``` 
or 
```
curl https://crash-stats.mozilla.com/api/ProcessedCrash/\?crash_id\=07794e51-0de1-4d3a-8c07-33ea72170309 | grep memory
```
**Is that simply because we don't have the root key `memory_measures` returned?** 

That's what we do with [`json_dump`](https://crash-stats.mozilla.com/admin/supersearch-field/?name=json_dump) (note it's "Is returned" but not "Is exposed")

Let's tackle that first. 

The other thing I don't understand is [this code](https://github.com/mozilla/socorro/blob/2b551320c35a49ba8cc628630d3122fb1df1f9a4/socorro/external/boto/crashstorage.py#L364-L369):
```python
        # Rename fields in processed_crash.
        processed_fields_map = dict(
            (x['in_database_name'], x['name'])
            for x in all_fields.values()
            if x['namespace'] == 'processed_crash'
)
```

Doesn't that mean we NEVER include any fields that are called things like `processed_crash.memory_measures` or `processed_crash.json_dump` or `processed_crash.whatever`?

Ping me when time allows and we'll debug it together.

adngdb:
@peterbe you were right, without the ``memory_measures`` key added to Super Search Fields, results did not come out of the processed crash API. I've added it to stage and prod, your script should now use these values.

adngdb:
You are also correct that the piece of code you quoted is problematic with sub-dicts inside the crash document. I am not sure how best we can solve it. Adding the first level field works, as seen with ``processed_crash.json_dump`` (and now hopefully ``processed_crash.memory_measure``) but it doesn't sound easy to maintain, as our current struggles prove. It is likely that we should change the code to make it consider all namespaces instead of just ``processed_crash`` and ``raw_crash`` to ensure that we won't forget this again in the future. What do you think?

peterbe:
I've had a chance to sleep on this. Better clarity now. 

The biggest problem is that people across the org might expect/think/suspect/hope that nested structures get flattened in Telemetry. They don't. But **we** do in Socorro (aka. Super Search). You can search for "memory private >= 123456" but in Redash etc. you'd have to do something like "WHERE memory_measures IS NOT NULL and memory_measures->private >= 123456". 

The guiding light here is that people (I hope, marco did) will understand that the stuff we send from Socorro is modelled after the structure of `crash_reports.json` and in that file the things like `memory_measures` *is* a nested structure.

adngdb:
@peterbe we can be smarter in the way we create that new crash_report document, and copy all keys that we actually know. That means that we would have, in the document, the entire JSON dump in ``json_dump`` and the few things we extract from it in the root under their SuperSearch names.

peterbe:
For the record, the problem is that ProcessedCrash isn't exposing memory_measures so I can't test if the JSON Schema matches what we intend to send to Telemetry. 
See https://bugzilla.mozilla.org/show_bug.cgi?id=1346851

PR #3689, opened 2017-03-06
fixes bug 1344838 - Upgrade to crontabber >= 0.18 [don't merge til thursday]

peterbe:
I actually don't know how to really test this. 
I've been doing manual testing of the new crontabber locally a lot but it would be really hard to test it with socorro's config. I propose we land this, upgrade the stage admin node and see how it behaves over (days'ish) time.


willkg:
@lonnen mentioned not wanting to merge this until Thursday giving us a week of it sitting on -stage so as to give us some more confidence. That seems prudent to me, too.

I'm hoping to work through this by end-of-Wednesday.

lonnen:
I wanted to make sure this has a long bake time on stage, and to ensure that once we switch to the new lock its _the_ lock.

willkg:
I went through the crontabber changes and the changes here and I had a bunch of thoughts. I decided to put them in the bug since many of them are about what we do before and after we land these changes and that seems like a better conversation for the bug rather than this PR.

https://bugzilla.mozilla.org/show_bug.cgi?id=1344838#c2

I think I want to get those resolved before continuing here since landing this triggers a -stage deploy and then we're committed.

adngdb:
We are past Thursday (or are we?). Should we merge this?

willkg:
@peterbe Want to merge this today (Monday) and keep tabs on it for the next few days?

PR #3690, opened 2017-03-07
fixes bug 1344862 - remember preferred panes on Summary tab

peterbe:


webapp-django/crashstats/signature/jinja2/signature/signature_summary.html
```diff
@@ -1,4 +1,4 @@
-<div class="panel tab-inner-panel">
+<div class="panel tab-inner-panel" id="tab-platform">
```
adngdb: Can you replace ``tab`` with ``panel`` in all these ids? They aren't tabs but panels, and it might conflict with the actuals tabs of that page.

PR #3692, opened 2017-03-08
Fixes bug 1345504 - Made better use of space in Signature report.

adngdb:
This uses the column-count CSS property to display inner panels in columns, thus leading to a better use of white space when there are long results. It also reorders summary panels to keep the same display order than now.

Assigned to @lonnen for he is the Grand Master of UI.

### DIFF
--- a/webapp-django/crashstats/signature/jinja2/signature/signature_summary.html
+++ b/webapp-django/crashstats/signature/jinja2/signature/signature_summary.html
@@ -25,33 +25,6 @@
     </div>
 </div>
 
-<div class="panel tab-inner-panel" id="panel-uptime">
-    <header class="title" title="Toggle visibility">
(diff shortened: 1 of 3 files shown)


willkg:
Can we get a screenshot on this?

adngdb:
Here are a few:

willkg:
Oo--nice!

lonnen:
I'm late to this party. This is a nice improvement. When I interact with the panels the reflow between columns is jarring, which I suspect is better solved by reducing the need to interact with the panels and reducing the amount of space that changes when someone does. Good patch.

I collected my thoughts on a broader overhaul here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1346640

They're not comprehensive and I didn't look at panels beyond these two. I may add to it later.

PR #3703, opened 2017-03-24
bug 1349661 - bump $target_version and document new version bump requ…

peterbe:
Awaiting review by Anthony Miyaguchi but he's not in this repo's team. 

cc @adngdb


acmiyaguchi:
It might be nice to associate `json_dump.pid` with the version bump, but the bug number should be enough context.

Looks good to me.

PR #3709, opened 2017-03-30
Fixes bug 1351302 - Removed the GCCrashes page from the web app.

adngdb:


adngdb:
Damn, wrong bug number...

PR #3711, opened 2017-03-30
Fixes bug 1352095 - Removed /releases/channels middleware service.

adngdb:


socorro/external/postgresql/releases.py
```diff
@@ -23,38 +23,6 @@ class Releases(PostgreSQLBase):
...
-
-        channels = {}
-        for res in sql_results.zipped():
-            if res['product'] not in channels:
-                channels[res['product']] = [res['channel']]
-            else:
-                channels[res['product']].append(res['channel'])
-
-        return channels
-
```
peterbe: Good riddance! One less use of the `product_info` view function.

adngdb:
Nope, the other two endpoints related to Releases are still being used by the webapp, so they'll have to be moved to django. This one was unused.

PR #3713, opened 2017-03-31
fixes bug 1352385 - Use active versions if 0 featured versions

peterbe:
This basically makes the home page, Top Crashers, Crashes Per Day work again if the product *has* active versions but *no featured* versions.

### DIFF
--- a/webapp-django/crashstats/base/jinja2/crashstats_base.html
+++ b/webapp-django/crashstats/base/jinja2/crashstats_base.html
@@ -59,26 +59,30 @@
                     <optgroup label=" ">
                         <option value="Current Versions">Current Versions</option>
                     </optgroup>
+
+                    {% set pvs = filter_featured_versions(active_versions[product]) %}
+                    {% if pvs %}
                     <optgroup label=" ">
-                    {% for pv in active_versions[product] %}
-                        {% if pv.is_featured %}
+                    {% for pv in pvs %}
                         <option
                             value="{{ pv.version }}"
                             {% if pv.version == version %}selected{% endif %}
                         >{{ pv.version }}</option>
-                        {% endif %}
-                    {% endfor  %}
+                    {% endfor %}
                     </optgroup>
+                    {% endif %}
+
+                    {% set pvs = filter_not_featured_versions(active_versions[product]) %}
+                    {% if pvs %}
                     <optgroup label=" ">
-                    {% for pv in active_versions[product] %}
-                        {% if not pv.is_featured %}
+                    {% for pv in pvs %}
                         <option
                             value="{{ pv.version }}"
                             {% if pv.version == version %}selected{% endif %}
                         >{{ pv.version }}</option>
-                        {% endif %}
-                    {% endfor  %}
+                    {% endfor %}
                     </optgroup>
+                    {% endif %}
                 </select>
             </li>
             <li>
--- a/webapp-django/crashstats/crashstats/tests/test_views.py
+++ b/webapp-django/crashstats/crashstats/tests/test_views.py
@@ -323,7 +323,7 @@ class BaseTestViews(DjangoTestCase):
                     'throttle': '99.00',
                     'end_date': yesterday,
                     'start_date': '2012-03-08',
-                    'is_featured': True,
+                    'is_featured': False,
                     'version': '9.5',
                     'build_type': 'Alpha',
                     'has_builds': True,
@@ -333,7 +333,7 @@ class BaseTestViews(DjangoTestCase):
                     'throttle': '99.00',
                     'end_date': end_date,
                     'start_date': '2012-03-08',
-                    'is_featured': True,
+                    'is_featured': False,
                     'version': '10.5',
                     'build_type': 'nightly',
                     'has_builds': True,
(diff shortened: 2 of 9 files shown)


peterbe:
screen shot 2017-03-31 at 8 58 11 am
It works again. Yay!

webapp-django/crashstats/base/jinja2/crashstats_base.html
```diff
@@ -59,26 +59,30 @@
...
+                    {% for pv in pvs %}
                         <option
                             value="{{ pv.version }}"
                             {% if pv.version == version %}selected{% endif %}
                         >{{ pv.version }}</option>
-                        {% endif %}
-                    {% endfor  %}
+                    {% endfor %}
                     </optgroup>
+                    {% endif %}
```
peterbe: All of these changes were necessary so that you don't get this:
screen shot 2017-03-31 at 9 56 27 am

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -369,7 +369,7 @@ def mocked_product_versions(**params):
                     'throttle': '99.00',
                     'end_date': end_date,
                     'start_date': '2012-03-08',
-                    'is_featured': True,
+                    'is_featured': False,
```
peterbe: I changed both of these ^ first to see if that broke any tests. It didn't. So I reused this piece of fixture for my new changes.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -510,9 +510,15 @@ def crashes_per_day(request, default_context=None):
...
         params['versions'] = []
-        for pv in context['active_versions'][context['product']]:
+        active_versions = context['active_versions'][context['product']]
+        for pv in active_versions:
             if pv['is_featured']:
                 params['versions'].append(pv['version'])
+        if not params['versions'] and active_versions:
+            # There were no featured versions, but there were active
+            # versions. Use the top 4 active versions instead.
+            for pv in active_versions[:4]:
```
peterbe: This number `4` *could* be moved to `settings` but I figured we'd never go into consul and change that. If we decide that 3 is the right number instead, we might as well just make another PR.

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -213,6 +213,15 @@ def topcrashers(request, days=None, possible_days=None, default_context=None):
...
                 return redirect(url)
+        if context['active_versions'][product]:
+            # Not a single version was featured, but there were active
+            # versions. In this case, use the first available
+            # *active* version.
+            for pv in context['active_versions'][product]:
+                url = '%s&version=%s' % (
+                    request.build_absolute_uri(), urlquote(pv['version'])
+                )
+                return redirect(url)
```
peterbe: There is a bit of repeated code here but I think it's really really easy to read and there isn't much actual business logic that is repeated.

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -1102,6 +1102,59 @@ def is_percentage(thing):
...
+                    'release': 0.1,
+                    'beta': 1.0,
+                }
+            }
+
+        models.ProductBuildTypes.implementation().get.side_effect = (
+            mocked_product_build_types_get
+        )
+
+        models.ADI.implementation().get.side_effect = mocked_adi_get
```
adngdb: Nit: this would be better right after the ``mocked_adi_get`` function imo.

webapp-django/crashstats/crashstats/views.py
adngdb: It doesn't make sense to have that in ``settings``, but we have a weird way of handling constants (basically: we don't use them). This could very well go into a ``NUMBER_OF_FEATURED_VERSIONS`` constant at the top of this file.

webapp-django/crashstats/home/tests/test_views.py
```diff
@@ -26,6 +26,20 @@ def test_home(self):
         ok_('WaterWolf 4.0.1' in response.content)
         ok_('WaterWolf 19.0' not in response.content)
 
+    def test_home_product_without_featued_versions(self):
```
adngdb: ``featued`` -> ``featured``

webapp-django/crashstats/home/views.py
```diff
@@ -26,6 +26,15 @@ def home(request, product, default_context=None):
...
         ]
+        # If there are no featured versions but there are active
+        # versions, then fall back to use that instead.
+        if not context['versions'] and context['active_versions'][product]:
+            # But when we do that, we have to make a manual cut-off of
+            # the number of versions to return. So make it max 4.
+            context['versions'] = [
+                x['version']
+                for x in context['active_versions'][product]
+            ][:4]
```
adngdb: This is another place the ``NUMBER_OF_FEATURED_VERSIONS`` constant could be of use.

adngdb:
Also, another rebase is needed.

peterbe:
I decided to make the number `4` a setting. It was just too nuts that it was written in the code twice in two different files.

PR #3714, opened 2017-03-31
fixes bug 1352440 - Cleanup old use of mocking

peterbe:
I know this looks like a lot but it's not. 

It's a leftover legacy from when we switched from mocking the middleware to implementations. I also did some cleaning up for stuff that was made global and didn't need to be. 

Sooo much less mocking now.


willkg:
GH is saying this PR has conflicts.

peterbe:
Rebased and fixed.

PR #3715, opened 2017-04-03
Fixes bug 1352094 - Moved Query service to django.

adngdb:


willkg:
GH is saying this PR has conflicts.

adngdb:
The 'json' name is just a name, behind the scene it turns the data into a dict, but it uses the ``json`` module to do so. I think this name makes sense. What you are expected to pass in is some JSON data, be it as a string or a dict.

PR #3719, opened 2017-04-04
fixes bug 1353413 - Django 1.8.18

peterbe:
After upgrading `http://localhost:8000/contribute.json` still worked. 
And I was able to go to `http://localhost:800/admin` but redirect to the login page, log in and successfully be redirected back to where I wanted to go.


willkg:
To clarify, I verified the version number and the hashes. I didn't verify that the code works fine figuring if the tests pass, then it's probably fine.

peterbe:
Will merge tomorrow when stage is no longer owned by the antenna deploy thing.

PR #3721, opened 2017-04-05
Fixes bug 1353372 - Removed middleware healthcheck.

adngdb:
And that would be the very last one!


willkg:
@adngdb Definitely worth talking to @milescrabill before merging this so we know what it's going to affect.

peterbe:
@willkg The middleware healthcheck is not something that miles is monitoring. His monitoring points to the healthcheck endpoint in the webapp. And the webapp (now) *used* to include/wrap the middleware's healthcheck.

PR #3722, opened 2017-04-05
Add object-src 'none' to CSP

g-k:
Assuming we don't use flash, java and other plugins, we shouldn't run them at all even from our origin.

Our CSP object-src currently falls back to the default-src of 'self' allowing clients to run plugins from our origin. 

This PR locks things down further to prevent plugin execution in the unlikely event that someone can inject object/embed html and upload a plugin to our domain.


peterbe:
Welcome to the git commit logs!!

PR #3724, opened 2017-04-06
Update python-memcached

Tayamarn:
My main concern is about cPickle was added instead of basic pickle:
https://github.com/linsomniac/python-memcached/blob/master/memcache.py#L61

Sometimes I have really large dicts to try to cache and when I do, I want caching to be fast. =)


willkg:
This looks good to me. Thank you!

PR #3727, opened 2017-04-13
DO NOT MERGE: Fixes bug 1250132 - Added a Rule to rewrite addons.

adngdb:

@marco-c What do you think about that new format? Is that going to work fine in Telemetry?

### DIFF
--- a/socorro/processor/rules/addons_rewrite.py
+++ b/socorro/processor/rules/addons_rewrite.py
@@ -1,0 +1,61 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from socorro.lib.transform_rules import Rule
+
+
+class AddonsRewriteRule(Rule):
+    """Rewrite the `addons` field so make it easier to query.
+
+    The `addons` field contains a list of add-on ids and versions, but it
+    doesn't associate those. That makes it hard to query with tools like
+    Elasticsearch. We thus rewrite that field so ids and versions are in a
+    single string, separated by a colon.
+
+    In case the original list of add-ons has an odd number of elements, we
+    keep the last element as is.
+
+    For example:
+
+    >>> processed_crash['addons'] = [
+    ...     'addon_1', '12.0',
+    ...     'mySuperADDON', '1.143.2',
+    ...     '{xws2r28dsoqq}', '42'
+    ... ]
+    >>> AddonsRewriteRule().act({}, {}, processed_crash, {})
+    >>> print(processed_crash['addons'])
+    ['addon_1:12.0', 'mySuperADDON:1.143.2', '{xws2r28dsoqq}:42']
+
+    """
+
+    def version(self):
+        return '1.0'
+
+    def _predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
+        """Verify that the `addons` field is present in the processed_crash
+        and that it is exploitable. """
+        return (
+            'addons' in processed_crash and
+            isinstance(processed_crash['addons'], (tuple, list)) and
+            len(processed_crash['addons']) > 1
+        )
+
+    def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
+        """Transform the `addons` field into a list of 'id:version' elements.
+        """
+        old_addons = processed_crash['addons']
+        new_addons = []
+
+        parts_nb = len(old_addons)
+
+        for i in range(0, parts_nb, 2):
+            if i + 1 == parts_nb:
+                # This is the last element of a list with an odd number of
+                # elements. We add it solo so as to not lose data.
+                new_addons.append(old_addons[i])
+            else:
+                new_addons.append('%s:%s' % (old_addons[i], old_addons[i + 1]))
+
+        processed_crash['addons'] = new_addons
+        return True
--- a/socorro/unittest/processor/rules/test_addons_rewrite.py
+++ b/socorro/unittest/processor/rules/test_addons_rewrite.py
@@ -1,0 +1,101 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import os
+from nose.tools import eq_, ok_
+
+from socorro.processor.rules.addons_rewrite import (
+    AddonsRewriteRule,
+)
+from socorro.unittest.testbase import TestCase
+from socorro.unittest.processor.test_signature_utilities import (
+    create_basic_fake_processor
+)
+
+
+HERE = os.path.dirname(__file__)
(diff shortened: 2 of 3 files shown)


peterbe:
I think you should ask @willkg to review instead. He's already closer to processors than I am. 

I did look through the code by the way and it looks fine.

socorro/processor/rules/addons_rewrite.py
```diff
@@ -0,0 +1,61 @@
...
+        return '1.0'
+
+    def _predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
+        """Verify that the `addons` field is present in the processed_crash
+        and that it is exploitable. """
+        return (
+            'addons' in processed_crash and
+            isinstance(processed_crash['addons'], (tuple, list)) and
+            len(processed_crash['addons']) > 1
+        )
```
willkg: I'm puzzled by this. Why does it have to be `> 1`? Should that be `> 0` or `>= 1`?

willkg: My bad--I just realized why. Sorry!

willkg: Instead of checking whether it's `> 1`, you might want to check if it's even.

willkg: I should be reviewing my comments before posting them. Sorry about that.

This should be `> 0` and then the "if it's odd, then we're cool" thing in the `_action` will handle the odd case just fine.

willkg: Oh, hrm. Actually, I take that back. It looks weird, but it's fine.

socorro/processor/rules/addons_rewrite.py
```diff
@@ -0,0 +1,61 @@
...
+
+        parts_nb = len(old_addons)
+
+        for i in range(0, parts_nb, 2):
+            if i + 1 == parts_nb:
+                # This is the last element of a list with an odd number of
+                # elements. We add it solo so as to not lose data.
+                new_addons.append(old_addons[i])
+            else:
+                new_addons.append('%s:%s' % (old_addons[i], old_addons[i + 1]))
```
willkg: What happens if `old_addons[i]` has a `:` in it? Should we scrub that?

socorro/processor/rules/addons_rewrite.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
+        """Transform the `addons` field into a list of 'id:version' elements.
+        """
+        old_addons = processed_crash['addons']
+        new_addons = []
+
+        parts_nb = len(old_addons)
+
+        for i in range(0, parts_nb, 2):
```
willkg: In a couple of projects I've worked on, I've written a chunk iterator which returns groups of n elements from an iterable with the last group being whatever's left. It's something I always thought was missing from the Python itertools module.

Erik wrote one for more-itertools:

http://more-itertools.readthedocs.io/en/latest/api.html#more_itertools.chunked

but we're not using that in Socorro, so that's not wildly helpful here.

Anyhow, I wish Python had it.

socorro/unittest/processor/rules/test_addons_rewrite.py
```diff
@@ -0,0 +1,101 @@
...
+        action_result = rule.action({}, {}, processed_crash, {})
+        ok_(action_result)
+        ok_('addons' in processed_crash)
+
+        expected_res = [
+            'abcd:1',
+            'efgh:32',
+            'xyz',
+        ]
+        eq_(processed_crash['addons'], expected_res)
```
willkg: All these tests could be reduced to a couple of parametrized tests.

py.test supports parametrization nicely. nose has test generators, but I think those don't work with `unittest.UnitTest` derived classes.

Maybe we should be using something like https://github.com/wolever/parameterized ?

I'm just thinking--this doesn't need to change for this PR.

willkg:
The only issue I can see is that it's not clear what happens if `addons[i]` has a `:` in it. Everything else seems fine.

socorro/processor/rules/addons_rewrite.py
adngdb: I looked for ``:`` in the ``addons`` field in the last 6 months, and I can only find 69 occurrences: https://crash-stats.mozilla.com/search/?addons=~%3A&date=%3E%3D2016-10-13T14%3A59%3A54.000Z&date=%3C2017-04-13T14%3A59%3A54.000Z&_sort=-date&_facets=signature&_facets=addons&_columns=date&_columns=signature&_columns=product&_columns=version&_columns=addons#crash-reports

That seems negligible enough for me, but I'd like to know what @marco-c thinks about it.

socorro/processor/rules/addons_rewrite.py
adngdb: Well, maybe now is the time to include ``more_itertools`` in the project?

willkg: I'm not sure there are other things we'd use. Writing a `chunk` function is pretty straight-forward:

https://github.com/mozilla/fjord/blob/master/fjord/search/index.py#L249

I wrote that for Mozilla, so we could lift it if we wanted. I don't know whether there are other parts of Socorro that implement similar things or not.

marco-c:
> @marco-c What do you think about that new format? Is that going to work fine in Telemetry?

This is definitely better than the current format! It should be easy to use with Spark or in Redash, as it just requires a substring operation.

Regarding the `:` character, perhaps we can use a delimiter that is actually forbidden in an addon ID? Are there any forbidden characters in addon IDs, @eviljeff?

Last question, what will happen to existing consumer of the SuperSearch API that are relying on the current format?

eviljeff:
> Regarding the : character, perhaps we can use a delimiter that is actually forbidden in an addon ID? Are there any forbidden characters in addon IDs, @eviljeff?

@marco-c addon IDs are either email@domain [format or guid style](https://developer.mozilla.org/en-US/Add-ons/Install_Manifests#id) and Firefox/AMO uses [this regex](https://dxr.mozilla.org/mozilla-central/rev/20dff607fb88ee69135a280bbb7f32df75a86237/toolkit/mozapps/extensions/internal/XPIProvider.jsm#310) to validate, so a `:` should be forbidden.

adngdb:
Thanks @eviljeff, that confirms my theory that the only cases we see with ``:`` in the id or the version are bugs. But that now leads to another question: how do we want to handle those cases? Do we want to "fix" them, as in, remove the invalid ``:`` sign in them? I'm not a fan of changing crash data, but that might be for the best here (I'm afraid leaving those signs will lead to bugs further down the road, e.g. in Telemetry). @marco-c what do you think?

marco-c:
Looking at the SuperSearch query, I think the `:` characters are always in the version string (the `:` character is probably allowed in the version).
If that's the case, querying the data should be easy (split the string on the first occurrence of `:`, the first part is the name, the second part is the version).

marco-c:
@adngdb should we mention this change in an email to the stability mailing list? There are consumers of the SuperSearch API which might be affected.

PR #3728, opened 2017-04-13
Fixes bug 1351706 - Removed GCCrashes public API endpoint.

adngdb:
This is removing the public facing API endpoint. In a few weeks, once we know for sure that removing it didn't break anything, I'll remove the underlying implementation.


adngdb:
@peterbe I don't think that's needed, I have a bunch of bug trackers and todo items set up to not forget about it. :)

PR #3730, opened 2017-04-14
fixes bug 1356571 - ignore version checking on certain products

peterbe:
cc @willkg @m8ttyB 

Another SeaMonkey thing. Thus I'm not going to scrutinize this problem too much. 

I do see SeaMonkey crashes coming in to stage. But none of the recent ones come from any of *active* SeaMonkey versions.

### DIFF
--- a/e2e-tests/tests/test_api.py
+++ b/e2e-tests/tests/test_api.py
@@ -8,9 +8,9 @@ import requests
 import pytest
 
 
-# Because there are so few crashes for these product, instead of
+# Because there are so few crashes for these products, instead of
 # looking for crashes from the *active* versions, we'll for any
-# any recent crashes for these.
+# version of any recent crashes for these.
 ANY_VERSION_PRODUCTS = ('SeaMonkey',)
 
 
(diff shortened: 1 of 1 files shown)


e2e-tests/tests/test_api.py
```diff
@@ -8,6 +8,12 @@
 import pytest
 
 
+# Because there are so few crashes for these product, instead of
+# looking for crashes from the *active* versions, we'll for any
+# any recent crashes for these.
```
willkg: This last clause has two "any" and is missing a word. Maybe it should be "we'll look for any recent crashes"?

e2e-tests/tests/test_api.py
```diff
@@ -26,6 +32,9 @@ def test_public_api_navigation(self, base_url):
         # for a single crash for that product
         product_to_uuid = {}
         for product, versions in product_versions.items():
+            if product == ANY_VERSION_PRODUCTS:
```
willkg: Should this be:

```
if product in ANY_VERSION_PRODUCTS:
```

?

peterbe: Wow. Yeah. Bad refactoring when I cleaned up.

peterbe:
Nits addressed.

PR #3731, opened 2017-04-17
fixes bug 1357159 - Stop using SocorroCommon.fetch for bugzilla API

peterbe:
CC @willkg 
Check out the new `requests_retry_session()` function. I think it can be useful. In particular because bugzilla can be quite flaky. 
If you do it like this, it'll retry on even making a connection which is akin or our web head not having a reliable network. 

Also, I tested this code by running it in a little test file, with decent backoff, whilst my WiFi was switched off. Then, whilst it was sitting there, I turned on the WiFi and, voila, got a 200 OK eventually. 

So, with 3 retries (connection errors) the worst that can happen is that it takes 0.3+0.6+1.2=2.1 seconds. And if it keeps timing out, it'll take a total of 5+0.3+5+0.6+5+1.2=17.1 seconds. But that's very unlikely to happen because timeouts are not systemic, unlike broken DNS or 500 errors.

### DIFF
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -194,7 +194,6 @@ def measure_fetches(method):
         self = args[0]
         url_or_implementation = args[1]
         if isinstance(url_or_implementation, basestring):
-            raise NotImplementedError(url_or_implementation)
             url = url_or_implementation
         else:
             url = url_or_implementation.__class__.__name__
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -42,6 +42,7 @@ from django.utils.encoding import iri_to_uri
 from django.template.defaultfilters import slugify
 
 from crashstats import scrubber
+from crashstats.base.utils import requests_retry_session
 
 
 DEPRECATION_RAMPAGE_WARNING = (
@@ -193,6 +194,7 @@ def measure_fetches(method):
         self = args[0]
         url_or_implementation = args[1]
         if isinstance(url_or_implementation, basestring):
+            raise NotImplementedError(url_or_implementation)
             url = url_or_implementation
         else:
             url = url_or_implementation.__class__.__name__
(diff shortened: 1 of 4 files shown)


webapp-django/crashstats/crashstats/models.py
```diff
@@ -193,6 +194,7 @@ def inner(*args, **kwargs):
         self = args[0]
         url_or_implementation = args[1]
         if isinstance(url_or_implementation, basestring):
+            raise NotImplementedError(url_or_implementation)
             url = url_or_implementation
```
willkg: I'm guessing that the `raise` line is there for testing things out. Is that right? If not, then setting the `url` here is moot.

peterbe: Interesting mistake. I started with https://bugzilla.mozilla.org/show_bug.cgi?id=1357156 but found that the BugzillaBugInfo thing still uses `requests.get()` etc. Thanks to that `raise` this is how I discovered that. I guess I forgot to remove it afterwards. But will.

peterbe:
Note-to-self: We chatted about this and realized that we don't confidently understand how this coincide with timeouts. If a timeout happens, does that trigger a retry?
The way to test it is to use the `requests_retry_session` thing but set the timeout in the `.get()` to something really really short that not even bugzilla has a chance to respond to.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -1085,19 +1086,14 @@ class CrontabberState(SocorroMiddleware):
...
-
-    # This is for the whole model.
-    cache_seconds = 0
+class BugzillaBugInfo(SocorroCommon):
 
     # This is for how long we cache the metadata of each individual bug.
     BUG_CACHE_SECONDS = 60 * 60
 
+    # How patient we are with the Bugzilla REST API
+    BUGZILLA_REST_TIMEOUT = 5  # seconds
```
adngdb: I *think* this is too short a timeout for bugzilla. Maybe bugzilla is not as slow as it used to be nowadays, but I don't think it would hurt to increase this a bit, especially since we query this API via AJAX. It is not a blocking resource.

peterbe: I disagree. 5 seconds is a really long time. Sure, it can happen, but if it does it's probably just a temporary glitch. The new `requests_retry_session` machine takes care of it. Its default is 3 retries. At means, if it, the first time, takes more than 5 seconds (every time), it'll cancel that request and do another one after a 0.3 second sleep. The total time is going to be about 21.8 seconds. 
```
>>> 5 + 0.0 + 5 + 0.3 + 5 + 0.6 + 5 + 0.9
21.799999999999997
```
That means it gets a total of 22 seconds to sort itself out. 
And thanks to the way we do caching, all we're requesting is a few fields for just a few bug IDs. 

If we change 5s to 10s the whole thing will take 42 seconds before it gives up. 

22s should be plenty enough.

peterbe:
I did a bunch more research into this retry code stuff. 
https://www.peterbe.com/plog/best-practice-with-retries-with-requests

In my blog post I also suggested you consider 500, 502 and 504 errors the same as connection errors. With bugzilla I'm not so sure what the right thing is to do. 

What I need to do is implement this in the ftpscraper because there we do get a fair amount of connection errors talking to S3 via https://archive.mozilla.org.

PR #3732, opened 2017-04-18
Inproved PolystorageError exceptions in log

Lehych:


willkg:
@Lehych This looks pretty interesting, but I don't really understand where you're coming from with it. Can you elaborate on the problem you're solving and how this improves it?

Lehych:
We had and issue with ons of storages in collector. But all we see in log was traceback to PolyCrashStorage methods and following exception presentation:
`PolyStorageError: (PolyStorageError(...), '')`
now we got
`PolyStorageError:  UnicodeDecodeError('utf8', '\x90\xff\x01', 0, 1, 'invalid start byte')`

socorro/external/crashstorage_base.py
```diff
@@ -494,6 +494,10 @@ def __setitem__(self, index, value):
         this method is required by the MutableSequence abstract base class"""
         self.exceptions.__setitem__(index, value)
 
+    def __str__(self):
+        return '%s %s' % (self.message,
+                          ','.join(repr(e[1]) for e in self.exceptions))
```
willkg: Is it possible for this to kick up a `UnicodeDecodeError`?

Maybe it should be something like this?:

```
def __str__(self):
    return '%s %s' % (self.message,
                      ','.join(repr(e[1]).encode('ascii', 'ignore') for e in self.exceptions))
```

willkg:
Oh, whoops--I had no idea you added a commit. I'll check it out today.

willkg:
Looks good. Thank you! 👍

PR #3733, opened 2017-04-18
fixes bug 1357536 - UnicodeEncodeErrors on broken raw crashes

peterbe:
Assuming you're using stage data, open https://crash-stats.allizom.org/report/index/c7c20b2e-ad32-487a-9363-fa5080170414 locally and see that it breaks before applying this PR. 

CC @willkg  if you're curious.


peterbe:
> I can't reproduce the error with the crash you linked, but the code looks good.

I forgot to say, you have to be signed in.

PR #3734, opened 2017-04-19
fixes bug 1357156 - Clean up HTTP middleware fetching from webapp

peterbe:
Feels nice and clean. 

Cool thing; I edited `models.py` and then ran the tests. They worked on the first attempt.


peterbe:
I'm not going to land this until AFTER we push to prod today. 
Too many people are off on PTO this week, and next week. So I want to give it some more time to brew on stage. Just in case we missed something.

PR #3736, opened 2017-04-20
Fixes bug 1358107 - Spring cleaning.

adngdb:
cc @lonnen


adngdb:
Sorry, this is a bit big, but I don't think you need to care about the content. As far as I could tell, none of those files were used at all anywhere in the Socorro app.

willkg:
Can any of them get pulled in because of configuration values we use in -stage or -prod?

adngdb:
I went on prod, made a backup dump of our consulate configuration (using ``consulate backup -f file.json``) and then grep'ed that file for the all the names of the files removed here, and didn't get any match. I believe we are good with removing them. 

I don't know if there's a better way to verify this. I'm hopeful that if this would cause a problem, we'd see in stage first (but we all know that stage and prod are sometimes different in surprising ways... )

PR #3738, opened 2017-04-24
Revert "Fixes bug 1250132 - Added a Rule to rewrite addons."

adngdb:
Reverts mozilla/socorro#3727

This entire work was based on an false assumption. :( The ``addons`` field contains an array of arrays, and not an array of strings as I thought. I blame myself for not challenging my knowledge and verifying the data.


adngdb:
@willkg Mind merging this when it's green? I'll work on fixing it tomorrow, for now we should stop producing incorrect data on stage.

willkg:
Reverting as @adngdb requested.

PR #3740, opened 2017-04-25
Fixes bug 1250132 - Added a Rule to rewrite addons. 

adngdb:

### DIFF
--- a/socorro/processor/mozilla_processor_2015.py
+++ b/socorro/processor/mozilla_processor_2015.py
@@ -57,6 +57,7 @@ mozilla_processor_rule_sets = [
         "socorro.processor.mozilla_transform_rules.ThemePrettyNameRule, "
         "socorro.processor.rules.memory_report_extraction"
         ".MemoryReportExtraction, "
+        "socorro.processor.rules.addons_rewrite.AddonsRewriteRule, "
         "socorro.processor.signature_utilities.SignatureGenerationRule,"
         "socorro.processor.signature_utilities.StackwalkerErrorSignatureRule, "
         "socorro.processor.signature_utilities.OOMSignature, "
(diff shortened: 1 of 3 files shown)


adngdb:
@willkg I am very sorry, I rebased my work and the changes from the last version are lost in a single commit. :/ The tests haven't changed much, it's just the Rule itself that is doing things differently.

willkg:
@adngdb No worries! I don't remember much beyond this morning anyhow, so I would have had to go through it again to see what past me did.

socorro/processor/mozilla_processor_2015.py
```diff
@@ -57,6 +57,7 @@
         "socorro.processor.mozilla_transform_rules.ThemePrettyNameRule, "
         "socorro.processor.rules.memory_report_extraction"
         ".MemoryReportExtraction, "
+        "socorro.processor.rules.addons_rewrite.AddonsRewriteRule, "
```
willkg: One of the things the new processor is going to have is better names. "AddonsRewriteRule" is so droll. The new processor is going to have "AddonsRewriteRuleOfWonderAndMystery" or maybe "FrancoisTheAddonsRewriteRule".

socorro/processor/rules/addons_rewrite.py
```diff
@@ -0,0 +1,58 @@
...
+    def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
+        """Transform the `addons` field into a list of 'id:version' elements.
+        """
+        old_addons = processed_crash['addons']
+        new_addons = []
+
+        for parts in old_addons:
+            addon_str = parts
+            if isinstance(parts, (tuple, list)):
+                addon_str = ':'.join(unicode(x) for x in parts)
```
willkg: This converts `x` to unicode. Do we know what encoding `x` is in? Are we guaranteed that it's in `utf-8`? Maybe we should do something defensive like:

```
x.decode('utf-8', 'ignore')
```

That way we don't bump into UnicodeDecodeErrors.

adngdb:
Guess what I just found out? We already have a rule that generates addons. It does not make sense to add a new one on top of it, I should just change the existing one to make behave like we want... Closing this in favor of a new branch, coming up in a few moments. Sorry for the wasted reviewing time! But it made me look at where those addons came from, and that's how I figured this out, so thanks as well. :)

PR #3741, opened 2017-04-25
fixes bug 1359430 - Stop using ReleasesFeatured

peterbe:
Basically, I removed the "Featured Versions" functionality from the admin interface. 
There is no point in using that interface because there's a cron job that populates what versions are featured. And if someone where to "ignore" the cron job and forcibly set a featured version manually, they'd soon have their choice replaced by the cron job (which runs every hour).


willkg:
Did we ever use the admin view to let us see what was in the table? Perhaps for debugging some problem where featured versions weren't set correctly?

peterbe:
> Did we ever use the admin view to let us see what was in the table? 

Not me. If I ever needed to get insight into what product versions we have it I go to [`/api/`](https://crash-stats.mozilla.com/api/#ProductVersions) and click "Run Test Drive!".

peterbe:
@adngdb in other news, `docs/configuring-crash-stats.rst` is a hot mess of ooooold cruft.

adngdb:
I just made this little report to easily check what versions are featured on prod: http://processeer.io/report/52e26d78-8c3b-49b6-b385-3ddc0066e3f0

PR #3742, opened 2017-04-26
Fixes bug 1250132 - Changed the Addons generation rule.

adngdb:


adngdb:
I'll merge after we have pushed to prod, so we can have some time to verify this on stage.

PR #3743, opened 2017-04-26
fixes bug 1359974 - Clean up model fetches analysis since there is no…

peterbe:
… middleware

### DIFF
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -189,30 +189,24 @@ def measure_fetches(method):
             return result
         t1 = time.time()
         self = args[0]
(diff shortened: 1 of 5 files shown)


adngdb:
@peterbe I don't think those numbers make sense. How come the ``Hits`` is 0 for SuperSearchUnredacted but there are 3 ``Misses``?

peterbe:
> @peterbe I don't think those numbers make sense. How come the Hits is 0 for SuperSearchUnredacted but there are 3 Misses?

Three different supersearch queries with different parameters. Never a cache hit. Always (all 3 times) a cache miss.

adngdb:
Aaaah, my bad, I thought the "Hits" were the total number of requests. This is good then!

PR #3744, opened 2017-04-27
Fixes bug 1358457 - Added thread_name to thread stacks in stackwalker.

CervantesYu:
This adds a string property "thread_name" to thread stacks output by
stackwalker. The thread names are parsed from the crash annotation
"ThreadIdNameMapping", which is the concatination of:

thread_id:"thread name",

Example: 6106:"Timer",6108:"Link Monitor",6109:"Socket Thread",


CervantesYu:
Close for fixing build failure.

PR #3745, opened 2017-04-28
Fixes bug 1358457 - Added thread_name to thread stacks in stackwalker.

CervantesYu:
This adds a string property "thread_name" to thread stacks output by
stackwalker. The thread names are parsed from the crash annotation
"ThreadIdNameMapping", which is the concatination of:

thread_id:"thread name",

Example: 6106:"Timer",6108:"Link Monitor",6109:"Socket Thread",

### DIFF
--- a/minidump-stackwalk/stackwalker.cc
+++ b/minidump-stackwalk/stackwalker.cc
@@ -33,6 +33,7 @@
 #include <algorithm>
 #include <fstream>
 #include <iostream>
+#include <map>
 #include <ostream>
 #include <vector>
 #include <set>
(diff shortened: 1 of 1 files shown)


minidump-stackwalk/stackwalker.cc
```diff
@@ -687,6 +735,10 @@ static void ConvertProcessStateToJSON(const ProcessState& process_state,
     }
     thread["frames"] = stack;
     thread["frame_count"] = stack.size();
+    auto thread_name = thread_id_name_map[raw_stack->tid()];
+    if (!thread_name.empty()) {
+      thread["thread_name"] = thread_name;
```
CervantesYu: Thanks for the suggestion. Though we shouldn't have problems with regex.h functions since it conforms to POSIX.1-2001, POSIX.1-2008, the code is harder to read. The PR is updated with the simpler implementation of split() and stripquotes().

CervantesYu:
The corresponding UI change is in https://github.com/mozilla/socorro/pull/3748

PR #3749, opened 2017-05-03
Fixes bug 1361404 - Handle lists and strings in process_type values.

adngdb:


socorro/middleware/search_common.py
```diff
@@ -337,7 +337,12 @@ def fix_process_type_parameter(parameters):
         marked_for_deletion = []
         for index, process_type in enumerate(parameters['process_type']):
             if 'browser' in process_type.value:
-                process_type.value.remove('browser')
+                try:
+                    # If it's a list.
+                    process_type.value.remove('browser')
+                except AttributeError:
+                    # If it's a string.
```
peterbe: Nit: Instead of these two comments (which aren't actual sentences) remind the code-reader:
```python
# process_type.value can be either a string or a list of string.
try:
    process_type.value.remove('browser')
...
```

Or perhaps even more clear is:
```python
if isinstance(process_type.value, list):
    if 'browser' in process_type.value:
        process_type.value.remove('browser')
elif process_type.value == 'browser':
    process_type.value = ''
```

PR #3750, opened 2017-05-03
Redo README and add current status note

willkg:
We no longer support Socorro and we're making big changes. Seems prudent to let
users know the status and what the last good commit was.


willkg:
@lonnen Is the text here clear enough? My concern is that I'm going to go delete all the collector code and given that users continue to use Socorro and ask for support, I don't think we've been super clear about how we're not supporting Socorro anymore and how we're making backwards-incompatible changes.

So this text makes that clearer and gives them a sha for the last good commit (for some definition of "last good commit").

README.rst
```diff
@@ -0,0 +1,73 @@
...
+<http://code.google.com/p/google-breakpad/>`_.
+
+
+Support
+=======
+
+This is a very Mozilla-specific product. We do not currently have the capacity
+to support external users. If you are looking to use Socorro for your product,
+maybe you want to consider alternatives like `electron/mini-breakpad-server
+<https://github.com/electron/mini-breakpad-server>`_.
```
lonnen: let's list https://github.com/wk8/sentry_breakpad as an alternative also

README.rst
```diff
@@ -0,0 +1,73 @@
+=======
+Socorro
+=======
+
+Socorro is a set of services for collecting, processing, and displaying crash
```
lonnen: consider adding a word or two to emphasize Mozilla-specific

README.rst
```diff
@@ -0,0 +1,73 @@
...
+Devs hang out in the Socorro/Breakpad IRC channel:
+`<irc://irc.mozilla.org/breakpad>`_
+
+
+Development
+===========
+
+Current deployment status: https://whatsdeployed.io/s-7M7
+
+Infrastructure (AWS) code: https://github.com/mozilla/socorro-infra
```
lonnen: should we note that this repo is similarly deprecated?

lonnen:
Rather than do a date-base series of updates as we pull components, consider anointing a tag as the last holistic release of the system.

willkg:
I'll switch it to a tag--that's a good idea.

I was thinking "this is the last known good one" after which we have carte blanche to make any changes we want. I wasn't planning on doing further status updates.

README.rst
willkg: I have no idea if it is or not. I thought it was Mozilla-centric and there was no expectation that it was supported in any way.

willkg: I'll add a note about how this is Mozilla-centric and not supported, too.

willkg: https://github.com/mozilla/socorro-infra/pull/306

willkg:
@lonnen ^^^ Does that look good?

willkg:
I think I addressed all of @lonnen's comments and that this is fine. If not, we can do another PR.

lonnen:
Belated r+

PR #3752, opened 2017-05-04
Fixes bug 1362061 - Use actual new addons format in ThemePrettyNameRule.

adngdb:
The change to the addons field format in bug 1250132 broke the ThemePrettyNameRule rule. This commit makes that rule use the current format.

### DIFF
--- a/socorro/processor/mozilla_transform_rules.py
+++ b/socorro/processor/mozilla_transform_rules.py
@@ -1089,11 +1089,11 @@ class ThemePrettyNameRule(Rule):
 
     #--------------------------------------------------------------------------
     def _predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
-        '''addons is a list of tuples containing (extension, version)'''
+        '''addons is a list of string like 'extension:version'. '''
         addons = processed_crash.get('addons', [])
 
         for addon in addons:
-            extension, version = addon.split(':', 1)
+            extension = addon.split(':')[0]
             if extension in self.conversions:
                 return True
         return False
@@ -1103,9 +1103,12 @@ class ThemePrettyNameRule(Rule):
         addons = processed_crash.addons
 
         for index, addon in enumerate(addons):
-            extension, version = addon.split(':', 1)
-            if extension in self.conversions:
-                addons[index] = ':'.join(
-                    (self.conversions[extension], version)
-                )
+            if ':' in addon:
+                extension, version = addon.split(':', 1)
+                if extension in self.conversions:
+                    addons[index] = ':'.join(
+                        (self.conversions[extension], version)
+                    )
+            elif addon in self.conversions:
+                addons[index] = self.conversions[addons]
         return True
--- a/socorro/processor/mozilla_transform_rules.py
+++ b/socorro/processor/mozilla_transform_rules.py
@@ -1113,5 +1113,5 @@ class ThemePrettyNameRule(Rule):
                         (self.conversions[extension], version)
                     )
             elif addon in self.conversions:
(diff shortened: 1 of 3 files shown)


socorro/processor/mozilla_transform_rules.py
```diff
@@ -1092,7 +1092,8 @@ def _predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
         '''addons is a list of tuples containing (extension, version)'''
         addons = processed_crash.get('addons', [])
 
-        for extension, version in addons:
+        for addon in addons:
+            extension, version = addon.split(':', 1)
```
peterbe: This makes me nervous. If there is ever an `addon` that doesn't contain a `:` you get a `ValueError`.
And, it USED to be that `addons` used to iterate over a list of tuples. Will it now ALWAYS be a string? ...for old and new crashes.

peterbe: Actually, if you're not going to use `version` replace it with a `_`. 
An even safer thing would be:
```python
extension = addon.split(':')[0]
```
However, if you do *this*, there's a chance that the `_predicate()` might return True on an `addon` string that doesn't actually have a `:` in it. Thus you can get the same, above mentioned, `ValueError` later in the `_action` method.

peterbe: If I'm "wrong" being nervous, then please add a comment here that explains why you can ALWAYS guarantee that each item in `addons` always contains a `:`.

Perhaps I just don't fully understand the context but the code doesn't look very defensive.

adngdb:
@peterbe this is now robust to addons without a ``:``.

socorro/unittest/processor/test_mozilla_transform_rules.py
```diff
@@ -2218,3 +2218,23 @@ def test_missing_key(self):
...
+
+        processed_crash.addons = [
+            'addon_with_no_version',
+            '{972ce4c6-7e08-4474-a285-3208198ce6fd}:12.0',
+        ]
+        rule.act({}, {}, processed_crash, processor_meta)
+
+        expected_addon_list = [
+            'addon_with_no_version',
+            '{972ce4c6-7e08-4474-a285-3208198ce6fd} (default theme):12.0',
```
peterbe: Isn't 12.0 a version?

socorro/processor/mozilla_transform_rules.py
```diff
@@ -1089,11 +1089,11 @@ def version(self):
 
     #--------------------------------------------------------------------------
     def _predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
-        '''addons is a list of tuples containing (extension, version)'''
+        '''addons is a list of string like 'extension:version'. '''
```
peterbe: There's a clash. (sorry for being difficult)
If this doc string is true, the code doesn't need to do `if ':' in addon:`. 
Perhaps the doc string needs to be `addons is a list of string like 'extension:version' but there can be addons that lack the :version part`

socorro/unittest/processor/test_mozilla_transform_rules.py
adngdb: Yeah, but the addon before that has none. This is just to trigger the ``action``.

adngdb: OK, right, there's actually a bug if the theme doesn't have a version... Good catch!

peterbe: Ah! I see it now.

PR #3755, opened 2017-05-05
fixes bug 1362371 - Don't assume BZAPI always works

peterbe:
I didn't bother writing a test because then we'd have to mock urllib3 from within the requests vendor package. 

I've tried a similar solution on my songsear.ch for many weeks and it's been great.

### DIFF
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -55,6 +55,10 @@ class DeprecatedModelError(DeprecationWarning):
     """Used when a deprecated model is being used in debug mode"""
 
 
+class BugzillaRestHTTPUnexpectedError(Exception):
+    """Happens Bugzilla's REST API doesn't give us a HTTP error we expect"""
+
+
 def config_from_configman():
     definition_source = Namespace()
     definition_source.namespace('logging')
@@ -998,8 +1002,9 @@ class BugzillaBugInfo(SocorroCommon):
                 headers=headers,
                 timeout=self.BUGZILLA_REST_TIMEOUT,
             )
-            # Better fail here than getting a JsonDecodeError on the next line.
-            assert response.status_code == 200, response.status_code
+            if response.status_code != 200:
+                raise BugzillaRestHTTPUnexpectedError(response.status_code)
+
             for each in response.json()['bugs']:
                 cache_key = self.make_cache_key(each['id'])
                 cache.set(cache_key, each, self.BUG_CACHE_SECONDS)
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -987,11 +987,19 @@ class BugzillaBugInfo(SocorroCommon):
             url = settings.BZAPI_BASE_URL + (
                 '/bug?id=%(bugs)s&include_fields=%(fields)s' % params
             )
-            response = requests_retry_session().get(
+            response = requests_retry_session(
+                # BZAPI isn't super reliable, so be extra patient
+                retries=5,
+                # 502 = Bad Gateway
+                # 504 = Gateway Time-out
+                status_forcelist=(500, 502, 504),
+            ).get(
                 url,
                 headers=headers,
                 timeout=self.BUGZILLA_REST_TIMEOUT,
             )
+            # Better fail here than getting a JsonDecodeError on the next line.
+            assert response.status_code == 200, response.status_code
             for each in response.json()['bugs']:
                 cache_key = self.make_cache_key(each['id'])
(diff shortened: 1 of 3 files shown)


webapp-django/crashstats/crashstats/models.py
```diff
@@ -987,11 +987,19 @@ def get(self, bugs):
             url = settings.BZAPI_BASE_URL + (
                 '/bug?id=%(bugs)s&include_fields=%(fields)s' % params
             )
-            response = requests_retry_session().get(
+            response = requests_retry_session(
+                # BZAPI isn't super reliable, so be extra patient
+                retries=5,
+                # 502 = Bad Gateway
+                # 504 = Gateway Time-out
+                status_forcelist=(500, 502, 504),
```
willkg: It might be more flexible to pass in a predicate here. Maybe call it `should_retry`. Then this could be like:

```
response = requests_retry_session(
    retries=5,
    should_retry=lambda resp: resp.status_code != 200,
)
```

or more like yours:

```
response = requests_retry_session(
    retries=5,
    should_retry=lambda resp: resp.status_code in (500, 502, 504)
)
```

I've seen that kind of API with network retry decorators.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -987,11 +987,19 @@ def get(self, bugs):
...
+            ).get(
                 url,
                 headers=headers,
                 timeout=self.BUGZILLA_REST_TIMEOUT,
             )
+            # Better fail here than getting a JsonDecodeError on the next line.
+            assert response.status_code == 200, response.status_code
```
willkg: I'd make this assert message more informative. Maybe something like:

```
assert response.status_code == 200, (
    'Got non-200 status code: %s' % response.status_code
)
```

Otherwise it just has a number and it's much harder to figure out what's going on without tracking down the code that kicked it up.

webapp-django/crashstats/crashstats/models.py
peterbe: This utility function is a thin thing on top of `urllib3.util.retry.Retry` which is vendored into `requests`. 
https://github.com/shazow/urllib3/blob/master/urllib3/util/retry.py#L152

I'm just trying to map parameters to that. I see no `should_retry` thing natively in there.

webapp-django/crashstats/crashstats/models.py
peterbe: If, against all odds, you do get an `AssertionError` you'll see where in the code it was and you'll have to look it up and it'll make sense in my humble opinion. 

If BZPI is really really down and keep responding with 500 errors, the lines above will raise a `Retry` error (I think it's `requests.exceptions.Retry` or something)

I suspect @adngdb is going to say "remove the assert anyway" :)

I just thought it was a slight upgrade from getting a `JsonDecodeError` on the next line.

willkg: `AssertionError: 303` doesn't seem very meaningful to me. Seeing a traceback with that in it doesn't give me any clue as to what's wrong without tracking down the code that kicked it up and reading it.

webapp-django/crashstats/crashstats/models.py
willkg: Oh, I see. That's a bummer. Not particularly flexible.

webapp-django/crashstats/crashstats/models.py
peterbe: I think I'm more in favor of something like
```python
class BZAPIHttpResponseError(Exception):
    """happens when we get a unexpected response from the bugzilla REST API"""

...

if response.status_code != 200:
    raise BZAPIHttpResponseError(response.status_code)
```
We can tune the name but it'd be a heck of a lot more informative as a traceback.

willkg: Either way is fine with me.

webapp-django/crashstats/crashstats/tests/test_models.py
```diff
@@ -112,6 +112,22 @@ def new_mocked_get(**options):
...
+        api = model()
+
+        def mocked_get(url, **options):
+            return Response("I'm a teapot", status_code=418)
+
+        rsession().get.side_effect = mocked_get
+        assert_raises(
+            models.BugzillaRestHTTPUnexpectedError,
+            api.get,
+            '123456789'
```
willkg: What's the 123456789 thing here for?

peterbe: You need a parameter when you call [`.get`](https://github.com/peterbe/socorro/blob/7ea020f29aa0694ff6b8749119b64e6728432b74/webapp-django/crashstats/crashstats/models.py#L969)

PR #3756, opened 2017-05-05
Fix Makefile and yml file editing

willkg:
Quick r?


willkg:
I have no idea how many people are using editorconfig-enabled editors, but I am and editing Makefiles without this fix sux0rz. The yml file fix seemed prudent to add, too.

willkg:
I think I might be the only person. I'm going to self-r+ this.

PR #3757, opened 2017-05-05
fixes bug 1362185 - split addons correct for report index extensions tab

peterbe:


peterbe:
To test this, you can load up the report index for any new crash.
And to test the legacy you have to supersearch far far back in history and click on any report index page.

PR #3759, opened 2017-05-08
1363130 integration test

willkg:


willkg:
If the tests pass, then this is good to go even if it's buggy and I removed too much.

In a later PR, we/I will be removing the collector and all the collector-specific tests.

willkg:
The tests took 8m38s. Looking at the history of tests for pull requests, I don't think it made much of a difference to remove the integration tests.

That's ok. It's still 1,100 lines of code less.

PR #3761, opened 2017-05-09
Change commit sha to a tag

willkg:
We should point at a tag for the last known good release rather than a single
commit sha. This fixes that.

r?


willkg:
I'm going to self-r+ this. It was something @lonnen mentioned that I forgot to do.

PR #3762, opened 2017-05-09
bug 1362434 - send processor exceptions to sentry

peterbe:
So this PR just tackles *one* problem. It injects itself in every `._predidate(...)` and `._action(...)` call. Only. 

In [processor_2015.py it does a loop and calls .act() on every rule in a set](https://github.com/mozilla/socorro/blob/fc6841672e8179cd6721a52cdc1a34d233338bf2/socorro/processor/processor_2015.py#L231-L236).

I *could* inject this sentry code in the `Rule.act` method (which is overwritten in `TransformRule.act` by the way) but I don't think that's safe. 
I actually wanted to be very defensive here. I didn't want to mess too much which how the whole rule act of predicate and action works. What I just try to do is stop swallowing real Python exceptions whenever a `_predicate` or `_action` is called on **any** rule. 

Who knows. There could be some rule where exceptions is "kinda expected" and is abused as a way of saying it shouldn't run. For example:

```python
class MyCoolTransformRule(TransformRule):
    def _predicate(self, *args, **kwargs):
         if datetime.now().strftime("%A") == "Monday":
             raise Exception("Don't run this transform rule on Mondays")
         return True
```
This would be a way of basically saying "Don't run the _action on this transform rule". 
Instead, what the code should be is:

```python
class MyCoolTransformRule(TransformRule):
    def _predicate(self, *args, **kwargs):
         if datetime.now().strftime("%A") == "Monday":
             return False
         return True
```
But how do we hunt these down? Exceptions should NOT be OK unless it's extreme. Almost all cases of letting an exception mean "don't run" can be changed. And that's what this PR sets us in the right direction of.

### DIFF
--- a/socorro/lib/transform_rules.py
+++ b/socorro/lib/transform_rules.py
@@ -54,7 +54,7 @@ class Rule(RequiredConfig):
         self.config = config
         self.quit_check_callback = quit_check_callback
 
-    def _send_to_sentry(self, tag, *args, **kwargs):
+    def _send_to_sentry(self, tag, func, *args, **kwargs):
         """Execute this when an exception has happened only.
         If self.config.sentry.dsn is set up, it will try to send it
         to Sentry. If not configured, nothing happens.
@@ -70,6 +70,7 @@ class Rule(RequiredConfig):
                 'tag': tag,  # this can 'predicate' or 'action'
             }
 
+            args = inspect.getcallargs(func, *args, **kwargs)['args']
             # For every crash acted on, it's always
             # act(raw_crash, ...) etc.
             # But be defensive in case the first argument isn't there,
@@ -115,7 +116,12 @@ class Rule(RequiredConfig):
         try:
             return self._predicate(*args, **kwargs)
         except Exception as exception:
-            if not self._send_to_sentry('predicate', *args, **kwargs):
+            if not self._send_to_sentry(
+                'predicate',
+                self.predicate,
+                *args,
+                **kwargs
+            ):
                 # Only log if it couldn't be sent to Sentry
                 self.config.logger.debug(
                     'Rule %s predicicate failed because of "%s"',
@@ -146,21 +152,18 @@ class Rule(RequiredConfig):
         classification system itself."""
         try:
             return self._action(*args, **kwargs)
-        except KeyError, x:
-            if not self._send_to_sentry('action', *args, **kwargs):
-                # Only log if it couldn't be sent to Sentry
-                self.config.logger.debug(
-                    'Rule %s action failed because of missing key "%s"',
-                    to_str(self.__class__),
-                    x,
-                )
-        except Exception, x:
-            if not self._send_to_sentry('action', *args, **kwargs):
+        except Exception as exception:
+            if not self._send_to_sentry(
+                'action',
+                self.action,
+                *args,
+                **kwargs
+            ):
                 # Only log if it couldn't be sent to Sentry
                 self.config.logger.debug(
                     'Rule %s action failed because of "%s"',
                     to_str(self.__class__),
-                    x,
+                    exception,
                     exc_info=True
                 )
         return False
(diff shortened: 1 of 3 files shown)


peterbe:
Note, there are other places where exceptions are swallowed. 
[Here for example](https://github.com/mozilla/socorro/blob/fc6841672e8179cd6721a52cdc1a34d233338bf2/socorro/processor/processor_2015.py#L243-L252).
[Here also](https://github.com/mozilla/socorro/blob/fc6841672e8179cd6721a52cdc1a34d233338bf2/socorro/processor/processor_app.py#L103-L113).
I just didn't want the PR to get too big all at once. 

Let's close the bug once we're happy with stamping out enough of these dangerous swallowing of `exception Exception`.

peterbe:
Example error https://sentry.prod.mozaws.net/operations/socorro-stage/issues/603897/

socorro/lib/transform_rules.py
```diff
@@ -47,12 +49,61 @@ class Rule(RequiredConfig):
...
+            extra = {
+                'class': self.__class__.__name__,
+                'tag': tag,  # this can 'predicate' or 'action'
+            }
+
+            # For every crash acted on, it's always
+            # act(raw_crash, ...) etc.
+            # But be defensive in case the first argument isn't there,
+            # isn't a dict or doesn't have a 'uuid'.
+            if args and isinstance(args[0], collections.Mapping):
```
willkg: Seems like this code is trying to inspect the args/kwargs and figure out if any of them is a crash id. You have the crash frame, so you can use that and `inspect.getcallargs` and know for sure whether there was a crash id and which one it was:

https://docs.python.org/2/library/inspect.html#inspect.getcallargs

socorro/lib/transform_rules.py
```diff
@@ -94,18 +147,22 @@ def action(self, *args, **kwargs):
         try:
             return self._action(*args, **kwargs)
         except KeyError, x:
-            self.config.logger.debug(
-                'Rule %s action failed because of missing key "%s"',
-                to_str(self.__class__),
-                x,
-            )
+            if not self._send_to_sentry('action', *args, **kwargs):
```
willkg: If we're going to start treating `KeyError` as a "thing that should never happen", then we should squash this block and the next one into a big "catch all exceptions and log to sentry" block.

socorro/lib/transform_rules.py
peterbe: None of the arguments is the `crash_id` or `uuid`. I *think* the first argument to `_predicate` and/or `_action` is always the raw crash which is either a `dict` or a `DotDict` instance.

socorro/lib/transform_rules.py
peterbe: Yeah. Let's be more bold.

socorro/lib/transform_rules.py
willkg: The thing I said still holds. You can still use `inspect.getcallargs` to pull out the specific argument you're looking for.

willkg: I'd make two changes.

First, change `_send_to_sentry` to also take the function that kicked up the error.

```
def _send_to_sentry(self, tag, fun, *args, **kwargs):
```

Then change all the uses of `_send_to_sentry` so that they pass in `self._predicate` or whatever function it is.

Then change the above code along these lines:

```
argspec = inspect.getcallargs(fun, *args, **kwargs)
if 'raw_crash' in argspec:
    crash_id = argspec['raw_crash'].get('uuid')
    ...
```

Why do it this way? It's more specific and resilient to changes. It uses existing information in the function signature and how Python binds call values to function arguments.

peterbe: I'm really not convinced this is any better. 
There is no `argspec['raw_crash']` in the spec. It's just `args` and `kwargs`.
Look at how `_predicate` and `_action` is defined. 

Even the `act` method only has `args` and `kwargs`. 

This new way just made things more complicated and I don't see how it's any more specific or resilient to be honest.

PR #3764, opened 2017-05-10
Fixes bug 1363427 - Removed obsolete cron jobs.

adngdb:
This removes a lot of matview cron jobs that we do not run anymore, as well as a few that we stopped running a while ago but never deleted (correlation ones).


socorro/cron/jobs/reprocessingjobs.py
```diff
@@ -1,54 +0,0 @@
...
-        for crash_id in crash_ids:
-            self.queuing_connection_factory.save_raw_crash(
-                DotDict({'legacy_processing': 0}),
-                [],
-                crash_id
-            )
-            execute_no_results(connection, delete_sql, {
-                'crash_id': crash_id
-            })
-            connection.commit()
```
willkg: I thought in one of the recent conversations we had, that we discovered we still use this. Are we sure we don't?

socorro/cron/jobs/matviews.py
```diff
@@ -68,16 +68,6 @@ class SignaturesCronApp(_MatViewBackfillBase):
     depends_on = ('reports-clean',)
 
 
-class TCBSCronApp(_MatViewBackfillBase):
-    proc_name = 'update_tcbs'
```
willkg: If we're removing all these cron jobs, can we remove the related postgres procedures, too?

willkg: If we're not removing them in this PR, can we write up a follow-up bug to do it?

socorro/cron/jobs/matviews.py
```diff
@@ -90,48 +80,6 @@ class BuildADUCronApp(_MatViewBackfillBase):
     depends_on = ('fetch-adi-from-hive', 'reports-clean')
 
 
-class CrashesByUserCronApp(_MatViewBackfillBase):
-    proc_name = 'update_crashes_by_user'
-    app_name = 'crashes-by-user-matview'
```
peterbe: Are you sure we want to remove this? I'm not. 
Same applies to `CrashesByUserBuildCronApp`.

The reason is that we still haven't decided on a way to archive this. 
A couple of months ago we remove the UI for this and replaced it with the real-time one. 
Some day we need to export and archive up ALL of these historical records and from that day onwards, if someone wants to get historical data they can use the exported dump.

adngdb: We sorted out in the meeting that we can remove these. We only care about the historical data that is in the table.

socorro/cron/jobs/reprocessingjobs.py
adngdb: We were not sure, and I have since confirmed that we don't actually use it. We now put stuff in rabbitmq directly via an API endpoint in the webapp instead.

socorro/cron/jobs/matviews.py
adngdb: That will be part of the postgres simplification, yes. I'm going to write a list of all the database elements we can remove, and will proceed later on. I like to have migrations in their own, separate commits.

PR #3767, opened 2017-05-10
Fixes bug 1363550 - squash fs crashstorage classes

willkg:
This pulls in work I did when I worked on socorro-collector. It squashes the
hierarchy of filesystem crash stores down to two: FSRaditTreeStorage and
FSTemporaryStorage.

It changes all the configuration that required other classes to use one of these
two. It removes some classes that we don't use that was using one of these
classes. It updates the tests accordingly.

r?


willkg:
This doesn't *remove* classes so much as it squashes a bunch of functionality and behavior spread across a family of classes down to two. That's really super duper hard to review well.

I did this work originally for socorro-collector and it took a while. I followed methods via the MRO of the classes I wanted to keep and pulled the correct methods into those classes. After that, I went through and fixed the unittests that were broken.

Since I did that work in socorro-collector, there have been two changes to the filesystem crashstorage classes:

* 9c3ba26 Fixes bug 1316185: fold socorrolib back in (#3584)
* 3454078 fixes bug 1297760: Adjust logging for OSError (#3519)
* 6914c8d Fixes bug 1297760: handle OSError when symlinking (#3515)

The first one is trivial and not interesting. The other two were pretty minor and I folded them into these code changes.

So... let's say I made a horrible mistake here and there are bugs. What's the impact?

I don't think there's any impact. We no longer use the collector, crashmover, or middleware. We don't use the default source or destination crashstorage for the processor and I don't think the processor uses filesystem crashstorage for anything else.

Given that, I think we can rely on the tests and see if anything breaks on -stage which we should know within a few minutes of deploying and if all that's fine, I think we're good to go.

willkg:
I'm tagging this with a DO NOT MERGE because it's prudent to wait until after a -prod deploy today to merge this. Then it gets more time to sit around.

willkg:
We're doing a -prod push now, so this is ready to merge again.

willkg:
Oops--closed by accident.

socorro/collector/crashmover_app.py
```diff
@@ -26,67 +26,6 @@ class CrashMoverApp(FetchTransformSaveApp):
 
 
 #==============================================================================
-class ProcessedCrashCopierApp(FetchTransformSaveWithSeparateNewCrashSourceApp):
```
willkg: This app seems like it's used to move processed crash data from one crash storage to another. It'd be used if we migrated to a new data store or maybe to a new ES cluster or something like that. We haven't done that in a while and I don't think we're likely to use this to do that sort of thing in the future.

There are no tests for this and it defaults to the TarFileCrashStore. Given that, I nixed it.

socorro/external/fs/crashstorage.py
```diff
@@ -28,10 +28,11 @@
...
 
-def dates_to_strings_for_json(obj):
-    if isinstance(obj, datetime.datetime):
-        return obj.isoformat()
-    return json.JSONEncoder.default(self, obj)
+class JSONISOEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, datetime.date):
+            return obj.isoformat()
+        raise NotImplementedError("Don't know about {0!r}".format(obj))
```
willkg: This was taken from the boto crashstorage code.

We should probably centralize this in lib, but I decided it's easier to copy it. It replaces a function that uses "self", but really shouldn't.

willkg:
@peterbe, @adngdb: Do we use filesystem crashstorage anywhere else?

willkg:
I'm going to nix this. It had more value yesterday before I fixed the file system tests in other ways. At this stage, the value isn't worth the additional time reviewing. We can always resurrect this later if we want.

peterbe:
> Do we use filesystem crashstorage anywhere else?

Not that I can think of.

ezolenko:
> I don't think there's any impact. We no longer use the collector, crashmover, or middleware.

Does this imply that socorro-collector and socorro-crashmover services could eventually go away? What are they replaced with for Mozilla install? Should I be doing upgrades little more carefully for my own setup? (so far starting from 253 upgrading was seamless)

willkg:
@ezolenko We're no longer supporting non-Mozilla users for Socorro. I added a notice to the README and the docs along these lines recently, but that's just a formalization of something that changed a while ago:

https://github.com/mozilla/socorro/blob/master/README.rst#may-3rd-2017-update

The difference is that now we're removing large swaths of code including the collector and crashmover.

PR #3768, opened 2017-05-10
Fix elasticsearch tests

willkg:
This fixes two things about the elasticsearch tests:

1. reduces errors due to the cluster not being ready
2. fixes a test that assumed the cluster was at localhost:9200 which it might
   not be


willkg:
@adngdb These are minor changes, but they fix all the tests so that they run both in the Docker environment I'm working on as well as Travis, Vagrant, etc.

Does this look ok to you?

willkg:
Thank you!

PR #3769, opened 2017-05-11
Fix postgresql tests

willkg:
The postgresql server_status test stomped would overwrite two files, run the
test, and then delete the files altogether. This fixes that behavior by mocking
out the "read the file from disk" function so we're still testing the code
itself, but not stomping on the files in question.

There was another postgresql test that had the database host, username, and
password hardcoded. This changes those to use the relevant environment
variables.

### DIFF
--- a/socorro/unittest/external/postgresql/test_server_status.py
+++ b/socorro/unittest/external/postgresql/test_server_status.py
@@ -2,13 +2,31 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import os
-import socorro
+import contextlib
+import tempfile
+
 from nose.tools import eq_
 
 from socorro.external.postgresql import server_status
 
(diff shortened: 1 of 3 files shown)


socorro/unittest/external/postgresql/test_server_status.py
```diff
@@ -2,14 +2,32 @@
...
+import contextlib
+import tempfile
+
 from nose.tools import eq_
 
 from socorro.external.postgresql import server_status
 
 from unittestbase import PostgreSQLTestCase
 
+TEMP_DIR = tempfile.gettempdir()
```
adngdb: What the point of this? I don't see any other occurrences of ``TEMP_DIR`` in this file.

willkg: Oops--that's leftover from a previous attempt fix. I'll nix it.

willkg:
^^^ That removes the unused tempfile stuff. If that passes, I'll land this.

PR #3771, opened 2017-05-11
Fixes bug 1362048 - Redacted away StackTraces from raw_crash in ES.

adngdb:
The raw_crash nowadays contain a key called "StackTraces" that contains a pretty big string. Storing it in Elasticsearch has a pretty high cost in terms of disk space and memory. This thus adds a new redactor, specifically made for the raw_crash, that we want to use every time we send a crash to Elasticsearch.

@willkg this is quite opinionated, and I would like to know what you think about a few things in here. 

1. This means we will deepcopy the raw_crash as well as the processed_crash for every crash we process (only once, though, since the ES crashstorage is the only one that mutates crashes).
2. This means we modify the raw_crash, which is something we want to avoid, but in this case I don't think there's anyway around it (other than asking the Firefox team to not send that data in the first place, but we want to be able to handle this on our own). 
3. I don't like the fact that redacted keys are in config, so I made the RawCrashRedactor simply ignore whatever is in config and use a hard-coded list. I am very, very unsure about that. It is inconsistent with the rest of the app, _but_ it is also a lot easier to understand and maintain. And it is less bug-prone (the problem with the ``update_default`` function or whatever it's called isn't solved as far as I know). 

Also, the second commit just removes a bunch of decoration comments that I don't like and try to remove whenever I touch files that have them.

### DIFF
--- a/socorro/external/crashstorage_base.py
+++ b/socorro/external/crashstorage_base.py
@@ -596,26 +596,26 @@ class PolyCrashStorage(CrashStorageBase):
             try:
                 actual_store = getattr(a_store, 'wrapped_object', a_store)
 
-                if (
-                    hasattr(actual_store, 'is_mutator') and
-                    actual_store.is_mutator()
-                ):
+                if hasattr(actual_store, 'is_mutator') and actual_store.is_mutator():
                     # We do this because `a_store.save_raw_and_processed`
                     # expects the processed crash to be a DotDict but
                     # you can't deepcopy those, so we deepcopy the
                     # pure dict version and then dress it back up as a
                     # DotDict.
-                    processed_crash = SocorroDotDict(
+                    my_processed_crash = SocorroDotDict(
                         copy.deepcopy(processed_crash_as_dict)
                     )
-                    raw_crash = SocorroDotDict(
+                    my_raw_crash = SocorroDotDict(
                         copy.deepcopy(raw_crash_as_dict)
                     )
+                else:
+                    my_processed_crash = processed_crash
+                    my_raw_crash = raw_crash
 
                 a_store.save_raw_and_processed(
-                    raw_crash,
+                    my_raw_crash,
                     dump,
-                    processed_crash,
+                    my_processed_crash,
                     crash_id
                 )
             except Exception:
--- a/socorro/external/crashstorage_base.py
+++ b/socorro/external/crashstorage_base.py
@@ -589,26 +589,33 @@ class PolyCrashStorage(CrashStorageBase):
         # processed crash which is a SocorroDotDict into a pure python
         # dict which we can more easily copy.deepcopy() operate on.
         processed_crash_as_dict = socorrodotdict_to_dict(processed_crash)
+        raw_crash_as_dict = socorrodotdict_to_dict(raw_crash)
 
         for a_store in self.stores.itervalues():
             self.quit_check()
             try:
                 actual_store = getattr(a_store, 'wrapped_object', a_store)
 
-                if hasattr(actual_store, 'is_mutator') and actual_store.is_mutator():
+                if (
+                    hasattr(actual_store, 'is_mutator') and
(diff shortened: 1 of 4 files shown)


socorro/external/crashstorage_base.py
```diff
@@ -620,7 +580,6 @@ def save_processed(self, processed_crash):
         if storage_exception.has_exceptions():
             raise storage_exception
 
-    #--------------------------------------------------------------------------
```
willkg: I think if we don't like these, we should do a one-time pass and remove them all. I'm game for doing that if we want to go that route. Seems like an easy pass to do towards cleaning things up. I'd rather do that (spend an hour to do it once and for all) than have it show up in PRs (have to skim through them every time I review a PR).

socorro/external/crashstorage_base.py
```diff
@@ -630,26 +589,33 @@ def save_raw_and_processed(self, raw_crash, dump, processed_crash,
...
         for a_store in self.stores.itervalues():
             self.quit_check()
             try:
                 actual_store = getattr(a_store, 'wrapped_object', a_store)
 
-                if hasattr(actual_store, 'is_mutator') and actual_store.is_mutator():
+                if (
+                    hasattr(actual_store, 'is_mutator') and
+                    actual_store.is_mutator()
+                ):
```
willkg: I don't understand this change. Did anything change here or is it just getting around an 80-char limit?

If it's the latter, I'd rather we didn't change this since it doesn't make the code easier to read.

socorro/external/crashstorage_base.py
```diff
@@ -630,26 +589,33 @@ def save_raw_and_processed(self, raw_crash, dump, processed_crash,
...
                     # DotDict.
-                    crash = SocorroDotDict(copy.deepcopy(processed_crash_as_dict))
-                else:
-                    crash = processed_crash
+                    processed_crash = SocorroDotDict(
+                        copy.deepcopy(processed_crash_as_dict)
+                    )
+                    raw_crash = SocorroDotDict(
+                        copy.deepcopy(raw_crash_as_dict)
+                    )
```
willkg: `processed_crash` and `raw_crash` are passed in as arguments to this function. So now the values are sometimes the original values and sometimes copies. That's messy and I claim it makes things harder to debug. I think we want to avoid re-using those variable names in this way.

I'd change this to:

```
    my_processed_crash = SocorroDotDict(copy.deepcopy(processed_crash_as_dict))
    my_raw_crash = SocorroDotDict(copy.deepcopy(raw_crash_as_dict))
else:
    my_processed_crash = processed_crash
    my_raw_crash = raw_crash
```

Then use `my_raw_crash` and `my_processed_crash` in the `.save_raw_and_processed()` call below.

socorro/external/es/crashstorage.py
```diff
@@ -17,9 +17,27 @@
...
+    Not that this overwrites the list of forbidden_keys that would be defined
+    through configuration. That list is hard-coded in the __init__ function.
+    """
+    def __init__(self, config):
+        super(RawCrashRedactor, self).__init__(config)
+
+        # Overwrite the list of fields to redact away.
+        self.forbidden_keys = [
+            'StackTraces',
+        ]
```
willkg: I'm pretty sure you could simplify this by just setting it as a class property. E.g.

```
class RawCrashRedactor(Redactor):
    """docstring here"""
    forbidden_keys = ['StackTraces']
```

That'd work fine and then you don't have to have all the boiler plate for overriding `__init__`.

willkg: Oh, scratch that--that won't work. Nevermind.

socorro/external/crashstorage_base.py
adngdb: Yeah, that's to work around the 80-char limit. I'll revert it.

adngdb:
@willkg comments addressed!

PR #3773, opened 2017-05-11
Bug 1364193 - Add MessageChannel::Clear to the skip list

amccreight:
Adding Clear() and the dtor are enough to split the crashes in different protocols into separate signatures.


adngdb:
Thank you!

PR #3775, opened 2017-05-12
bug 1363177 - send code_id and code_file in query string when fetching symbols via HTTP

luser:
This will allow the symbol server to store the info necessary for missing
symbols so we can fetch them from the Microsoft symbol server.

I tested this by generating a minidump from my local browser with the crashme extension, and then running the stackwalker binary with `--symbols-url http://s3-us-west-2.amazonaws.com/org.mozilla.crash-stats.symbols-public/v1/`, which is where we currently point the production stackwalker. The stackwalker log showed:
```
2017-05-12 12:11:44: simple_symbol_supplier.cc:196: INFO: No symbol file at /tmp/symbols/firefox/8EF4BE7ED885345C9942AA5C8DC0C6C90/firefox.sym
2017-05-12 12:11:44: http_symbol_supplier.cc:338: INFO: HTTPSymbolSupplier: fetching http://s3-us-west-2.amazonaws.com/org.mozilla.crash-stats.symbols-public/v1/firefox/8EF4BE7ED885345C9942AA5C8DC0C6C90/firefox.sym?code_file=firefox&code_id=id
2017-05-12 12:11:44: http_symbol_supplier.cc:348: INFO: HTTPSymbolSupplier: querying http://s3-us-west-2.amazonaws.com/org.mozilla.crash-stats.symbols-public/v1/firefox/8EF4BE7ED885345C9942AA5C8DC0C6C90/firefox.sym?code_file=firefox&code_id=id
2017-05-12 12:11:44: http_symbol_supplier.cc:370: INFO: HTTPSymbolSupplier: fetch succeeded, saving to /tmp/symbols/firefox/8EF4BE7ED885345C9942AA5C8DC0C6C90/firefox.sym
```

So it successfully fetched firefox.sym from s3 without issues, and the URL it used did contain the query string. (Files on Mac in Breakpad don't have a useful code_id.)

I also verified that the stackwalker JSON output contained the proper URL in the `modules` list. I decided to continue to store the URL without a query string there, since that field is only used for generating links to symbol files from the Modules tab in the report/index page:

```json
      {
         "base_addr" : "0x1063dc000",
         "code_id" : "id",
         "debug_file" : "firefox",
         "debug_id" : "8EF4BE7ED885345C9942AA5C8DC0C6C90",
         "end_addr" : "0x1063de000",
         "filename" : "firefox",
         "loaded_symbols" : true,
         "symbol_disk_cache_hit" : false,
         "symbol_fetch_time" : 338.2170104980469,
         "symbol_url" : "http://s3-us-west-2.amazonaws.com/org.mozilla.crash-stats.symbols-public/v1/firefox/8EF4BE7ED885345C9942AA5C8DC0C6C90/firefox.sym",
         "version" : ""
      }
```

### DIFF
--- a/minidump-stackwalk/http_symbol_supplier.cc
+++ b/minidump-stackwalk/http_symbol_supplier.cc
@@ -236,6 +236,20 @@ namespace {
   string JoinURL(CURL* curl, const string& url, const string& sub) {
     return url + "/" + URLEncode(curl, sub);
   }
+
+  string GetCodeIDAndFileQueryString(CURL* curl, const CodeModule* module) {
+      string code_file = PathnameStripper::File(module->code_file());
+      if (code_file.empty()) {
+        return "";
+      }
+
+      string code_id = PathnameStripper::File(module->code_identifier());
+      if (code_id.empty()) {
+        return "";
+      }
+
+      return "?code_file=" + URLEncode(curl, code_file) + "&code_id=" + URLEncode(curl, code_id);
+  }
 }
 
 bool
@@ -290,6 +304,8 @@ bool HTTPSymbolSupplier::FetchSymbolFile(const CodeModule* module,
     return false;
   }
 
+  string query = GetCodeIDAndFileQueryString(curl_, module);
+
   string full_path = JoinPath(cache_path_, path);
 
   bool result = false;
@@ -297,8 +313,12 @@ bool HTTPSymbolSupplier::FetchSymbolFile(const CodeModule* module,
        server_url < server_urls_.end();
        ++server_url) {
     string full_url = *server_url + url;
+    string url_with_query = full_url;
+    if (!query.empty()) {
+      url_with_query += query;
+    }
     float fetch_time;
-    if (FetchURLToFile(curl_, full_url, full_path, &fetch_time)) {
+    if (FetchURLToFile(curl_, url_with_query, full_path, &fetch_time)) {
       StoreCacheMiss(module, fetch_time, full_url);
       result = true;
       break;


luser:
@peterbe It's OK if you don't thoroughly review the C++, I'd just like you to verify that the query string this code is generating matches what you're expecting.

peterbe:
@luser Can you do one more test (locally). Run it with `--symbols-url https://symbols.dev.mozaws.net` 
Then, run 
```
curl https://symbols.dev.mozaws.net/missingsymbols.csv\?today\=1
```
...and compare that CSV output with what you'd expect. 

That'll test that it works with redirects too.

peterbe:
Hang on. That example did have to do a lookup for `firefox/8EF4BE7ED885345C9942AA5C8DC0C6C90/firefox.sym` which will 200 OK. Can you force it to look up one that will definitely 404? Otherwise `https://symbols.dev.mozaws.net/missingsymbols.csv?today=1` will (and should) be empty.

minidump-stackwalk/http_symbol_supplier.cc
```diff
@@ -236,6 +236,20 @@ namespace {
...
+
+  string GetCodeIDAndFileQueryString(CURL* curl, const CodeModule* module) {
+      string code_file = PathnameStripper::File(module->code_file());
+      if (code_file.empty()) {
+        return "";
+      }
+
+      string code_id = PathnameStripper::File(module->code_identifier());
+      if (code_id.empty()) {
+        return "";
```
peterbe: I analyzed one of those csv files (`missing_symbols_20170424.csv`).
It had...:
```
2,144,682 rows with code_file && code_id
1,567 rows with code_file && !code_id
0 in all other combinations. 
```

minidump-stackwalk/http_symbol_supplier.cc
```diff
@@ -236,6 +236,20 @@ namespace {
...
+      }
+
+      return "?code_file=" + URLEncode(curl, code_file) + "&code_id=" + URLEncode(curl, code_id);
```
peterbe: Perhaps a useless nit, but should we perhaps omit the `?` there?
Technically a "query string" is the stuff *after* the `?` in a URL. 

Then, we could do something like this:

```c++
// (in HTTPSymbolSupplier::FetchSymbolFile)
...

// pseudo code...
if (!query.empty()) {
    found = url_with_query.find("?");
    if (found > 1) {
         url_with_query += "&";
    } else {
        url_with_query += "?";
    }
    url_with_query += query;
}
```
That would make it possible for us to do something like:
```
./bin/stackwalker --symbols-url=https://symbols.mozilla.org/?use_private=true&other=config
```

luser: Good idea, I'll make that change.

luser: Oh, we'd need to do actual URL parsing for this to work. Currently we just append the relative URL to the end of the symbols URL, so putting a query string on there doesn't work at all.

luser: It doesn't look like libcurl offers any URL parsing APIs, and finding a way to do this in C++ just to handle this scenario feels like too much work for the benefit, so I'm going to leave this as-is instead. Sorry!

peterbe: I'm assuming you're actually referring to line 316'ish where it appends the query string. 
THis function here looks like a plain function that returns a string. 

But isn't `full_url` a regular string?

luser: Yes, but if we have `server_url` set to something like `https://symbols.mozilla.org/?use_private=true&other=config` then we have to parse out the query string so we can append the relative url (like `foo.pdb/abcd/foo.sym`). We don't currently do that, we just blindly append the relative url. Since we don't already have some sort of URL parsing code handy it'd be a lot of work for a feature we don't actually need.

peterbe: Gotcha! Yeah, because we use the URL "as is" when assuming the file URL. 

Tecken does this differently. It deconstructs the URL to extract out the domain, bucket name and prefix. And other config variables are extracted from actually parsing the query string. 
https://github.com/mozilla-services/tecken/blob/master/tecken/base/symboldownloader.py#L82-L85

peterbe:
So the only thing remaining is to do 1) a manual test using https://symbols.dev.mozaws.net and 2) a manual test with that URL but one that causes a 404. 3) Expect to find that in `https://symbols.dev.mozaws.net/missingsymbols.csv?today=1`

I believe https://symbols.dev.mozaws.net is back and healthy again.

luser:
OK, I will re-run the test today when I get a minute.

luser:
I re-ran my test against https://symbols.dev.mozaws.net and it seems to have worked successfully. I got all the symbols loaded that I expected, such as:
```
2017-05-15 14:38:38: http_symbol_supplier.cc:353: INFO: HTTPSymbolSupplier: querying https://symbols.dev.mozaws.net/firefox/8EF4BE7ED885345C9942AA5C8DC0C6C90/firefox.sym?code_file=firefox&code_id=id
2017-05-15 14:38:39: http_symbol_supplier.cc:375: INFO: HTTPSymbolSupplier: fetch succeeded, saving to /tmp/symbols/firefox/8EF4BE7ED885345C9942AA5C8DC0C6C90/firefox.sym
```

And there was at least one missing symbol:
```
2017-05-15 14:43:09: http_symbol_supplier.cc:353: INFO: HTTPSymbolSupplier: querying https://symbols.dev.mozaws.net/CoreFoundation/51D1F34F9FEC3EF283822D13D75F845A0/CoreFoundation.sym?code_file=CoreFoundation&code_id=id
2017-05-15 14:43:12: http_symbol_supplier.cc:372: INFO: HTTPSymbolSupplier: HTTP response code: 404
```

And querying the missingsymbols.csv URL shows that missing symbol:
```
$ curl https://symbols.dev.mozaws.net/missingsymbols.csv?today=1
debug_file,debug_id,code_file,code_id
CoreFoundation,51D1F34F9FEC3EF283822D13D75F845A0,CoreFoundation,id
```

So it looks like everything is working as expected.

PR #3777, opened 2017-05-15
Fixes bug 1364933 - Made some Rules more conservative.

adngdb:
This fixes some exceptions that were very frequently occurring in the processor, and that we now see since we have added error reporting. I am fixing those 3 errors at once because they are very similar.

### DIFF
--- a/socorro/processor/mozilla_transform_rules.py
+++ b/socorro/processor/mozilla_transform_rules.py
@@ -622,12 +622,14 @@ class FlashVersionRule(Rule):
     def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
         processed_crash.flash_version = ''
         flash_version = None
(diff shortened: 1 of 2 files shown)


socorro/processor/mozilla_transform_rules.py
```diff
@@ -676,7 +676,7 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
         processed_crash.flash_version = ''
         flash_version = None
         for index, a_module in enumerate(
-            processed_crash['json_dump']['modules']
+            processed_crash['json_dump'].get('modules', [])
```
willkg: This still assumes that you get back a list. We should pull this out and make sure it's a list and not a string which you can also enumerate over, but would produce junk data later.

PR #3780, opened 2017-05-16
Update flake8 and friends; change to 100 char max line length

willkg:
This updates flake8 and it's dependencies to recent versions. That's mostly cosmetic. The one weird looking thing is that the pep8 library changed to pycodestyle.

The other thing this changes is sets the max line length to 100. Socorro uses a lot of long names and doing 80-character lines with long class/method/function names is super hard and I claim it makes our code far less readable and harder to review.


setup.cfg
```diff
@@ -1,3 +1,4 @@
 [flake8]
 ignore = E265
 exclude=webapp-django/crashstats/*/migrations/*,docs/conf.py,socorro/unittest/testlib/*
+max-line-length = 100
```
peterbe: Weak r-

80 chars is better for pasting around and the github diff view.

peterbe: Plus the times I'd editing in an ssh window in jed and not using my fancy Atom.

But if the team prefers 100 I'm happy to change.

willkg: All of those situations work fine with 100 characters for me, so I'm not sure why you're experiencing something different. Are you using a low resolution monitor?

Also, honestly, the code isn't linted at all currently, so there's a lot of code that already exceeds 100 characters. For example:

https://github.com/mozilla/socorro/blob/master/socorro/database/cachedIdAccess.py

This is mostly a cosmetic change that doesn't affect the status quo at all.

willkg:
Oops--inadvertently closed this.

willkg:
On the one hand, this sort of sets up a convention for 100-character line lengths, but in actuality, it doesn't change the status quo at all since we're not linting the Socorro code and this doesn't really change that.

Given that, I'm just going to land this now because it doesn't actually affect anything except my local setup and we can talk about it in retrospect if need be.

PR #3782, opened 2017-05-18
Fixes bug #1366016 - Fix collectstatic for webapp

willkg:
This fixes collectstatic when building the image as well as when running tests
and running the webapp service.

This also fixes running the tests.


docker/Dockerfile.webapp
```diff
@@ -15,17 +15,21 @@ COPY ./requirements.txt /tmp/requirements.txt
...
 ENV PYTHONUNBUFFERED 1
 ENV PYTHONDONTWRITEBYTECODE 1
 ENV PYTHONPATH /app:/app/webapp-django
 
-# FIXME(willkg): Run collectstatic for the container which puts the files in
-# the default place for static files.
-#RUN cd /app/webapp-django/ && python manage.py collectstatic --noinput
+ENV LESS_BINARY /usr/local/lib/node_modules/socorro-webapp-django/node_modules/.bin/lessc
+ENV UGLIFYJS_BINARY /usr/local/lib/node_modules/socorro-webapp-django/node_modules/.bin/uglifyjs
+ENV CSSMIN_BINARY /usr/local/lib/node_modules/socorro-webapp-django/node_modules/.bin/cssmin
```
willkg: The webapp needs these variables set and pointing to where these binaries get put. Since we're doing `npm install -g`, it "helpfully" puts them in this super easy to find path!

docker/Dockerfile.webapp
```diff
@@ -15,17 +15,21 @@ COPY ./requirements.txt /tmp/requirements.txt
...
+
+COPY . /app/
+
+RUN cd /app/webapp-django/ && npm install -g
+
+# Run collectstatic in container which puts files in the default place for
+# static files.
+RUN cd /app/webapp-django/ && python manage.py collectstatic --noinput
```
willkg: I uncommented this so that it runs when building the image.

docker/run_test.sh
```diff
@@ -75,5 +75,5 @@ $NOSE socorro -s
 # Collect static and then test webapp
 pushd webapp-django
 python manage.py collectstatic --noinput
-./bin/ci.sh
+./bin/ci.sh --docker
```
willkg: I had forgotten which way I did this flag and not having it here was causing the tests to fail because it would look for a `socorro-virtualenv` thing which we don't do in the docker environment.

docker/set_up_ubuntu.sh
```diff
@@ -11,19 +11,11 @@ set -v -e -x
...
-
-# Install git
-apt-get install -y git
+# Install packages for building python packages, postgres, lxml, sasl, and cffi
+apt-get install -y gcc apt-transport-https build-essential python-dev \
+        libpq-dev \
+        libxml2-dev libxslt1-dev \
+        libsasl2-dev \
+        libffi-dev \
+        git
```
willkg: I collapsed these into a single pass which reduces the time it takes to build containers.

willkg:
This looks fine, so I'm going to self-r+ it and merge it.

PR #3784, opened 2017-05-18
Fixes bug 1366046 - speed up "make dockerbuild"

willkg:
This nixes building the test service since it uses the same image as the webapp
service.

This pulls the redundant parts from building the processor and webapp images
into a base image and reorders some things to make that more convenient.

The end result is that it went from taking 15 minutes to 8 minutes to do "make
dockerbuild" on my local machine. I suspect/hope CircleCI gets similar results.

If this isn't sufficient, then I think future gains are going to be harder since
we'll need to reduce individual steps and I'm not sure what's left to cut/slim
down.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,9 +44,9 @@ DC := $(shell which docker-compose)
 	make dockerbuild
 
 dockerbuild:
+	${DC} build base
 	${DC} build processor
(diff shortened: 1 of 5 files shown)


willkg:
@milescrabill Does this look ok?

PR #3786, opened 2017-05-19
Fixes bug 1366051 - fix alembic issues in "make dockertest"

willkg:
This fixes alembic issues in "make dockertest" which came about from a lack of
alembic.ini file.


willkg:
Self-r+. Merging.

PR #3787, opened 2017-05-19
Bug 1361764 - initial switch to pytest

willkg:
This switches the test runner from nose to pytest for socorro/unittest/
tests (i.e. not the webapp tests).

This doesn't get rid of nose completely--we still have to go through and remove
ok_ and eq_ instances. You can continue to use nose if you like.

This fixes tests that were broken in pytest so that they work with both pytest
and nose.


willkg:
To verify correctness, I did a couple of things:

1. ran nose and pytest and made sure they came up with the same number of tests run (750) and skipped (1)
2. ran nose and pytest and compared run times -- both take 1m40s on my local machine

I tried to keep this set of changes as minimal as possible, but enough of a first step that we can make future changes in small, manageable batches (`eq_` and `ok_` to regular asserts, etc).

I'll update the bug with the other things we should do before the bug is completed.

adngdb:
A few things are broken when running outside of docker. https://github.com/willkg/socorro/pull/1/ is my attempt at fixing them. Also, circle is failing (not sure what that means though).

willkg:
Circle doesn't work right, yet, plus it's testing the docker-environment so it's "alpha-quality". @milescrabill and I have been working on that.

willkg:
@adngdb ^^^ That takes your changes and tweaks them slightly so they run in the docker environment again. Does that work in local dev environment still?

willkg:
Bah--I changed things in here. I'll fix it today.

willkg:
^^^ That rebases against master tip to account for some changes I made when getting the tests working in docker in circle-ci.

willkg:
This looks good. Landing it! Welcome to our new pytest overlords!

PR #3788, opened 2017-05-19
WIP: Build docker in circleci

willkg:


.circleci/config.yml
```diff
@@ -0,0 +1,100 @@
...
+          name: Build Docker images
+          working_directory: /socorro
+          command: |
+            docker info
+            make dockerbuild
+
+      # - run:
+      #     name: Setup test environment
+      #     working_directory: /socorro
+      #     command: make dockersetup
```
willkg: dockersetup doesn't set up the test environment--it sets up the environment for running the processor and webapp. We don't need it just to run the tests.

.circleci/config.yml
```diff
@@ -0,0 +1,100 @@
...
+          # Create the superuser account
+          - POSTGRES_USER=postgres
+          - POSTGRES_PASSWORD=aPassword
+          - POSTGRES_DB=breakpad
+      - image: elasticsearch:1.4.5
+      - image: rabbitmq:3.6.9
+        environment:
+          - RABBITMQ_DEFAULT_USER=rabbituser
+          - RABBITMQ_DEFAULT_PASS=rabbitpwd
+          - RABBITMQ_DEFAULT_VHOST=rabbitvhost
```
willkg: This duplicates what's in `docker-compose.yml`. If we're doing `make dockertest`, shouldn't it set up the required services with the correct configuration?

milescrabill: Right, this is the circley way of doing things as opposed to the docker-composey way. I haven't ripped out `make dockertest` but I think it will become just a single `docker` call.

.circleci/config.yml
milescrabill: Cool - if this is only needed for dev workflow I will rip it out of CI.

.circleci/config.yml
willkg: It's kind of irksome to have things in two places, but I'm not sure how to unify these.

PR #3791, opened 2017-05-23
Fixes bug 1367071 - delete build artifacts before building

willkg:
It's possible for a developer to have build artifacts from building stackwalk on
their host machine which won't work when building in the container. This fixes
the script to delete build artifacts before building.


willkg:
I'm 95% sure this will fix it, so I'm going to self-r+ it and land it.

PR #3793, opened 2017-05-24
Fixes bug 1363712 - Use a placeholder in addons without version.

adngdb:
In order to make queries easier, we want to ensure that all addons have the same format of addon_identifier:addon_version. We thus add a placeholder version to addons that miss it, to make things consistent.

### DIFF
--- a/socorro/processor/mozilla_transform_rules.py
+++ b/socorro/processor/mozilla_transform_rules.py
@@ -151,6 +151,16 @@ class AddonsRule(Rule):
     def version(self):
         return '1.0'
 
+    def _get_formatted_addon(self, addon):
+        """Return a properly formatted addon string.
+
+        Format is: addon_identifier:addon_version
+
+        This is used because some addons are missing a version. In order to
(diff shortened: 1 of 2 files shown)


socorro/processor/mozilla_transform_rules.py
```diff
@@ -151,6 +151,16 @@ class AddonsRule(Rule):
...
 
+    def _get_formatted_addon(self, addon):
+        """Return a properly formatted addon string.
+
+        Format is: addon_identifier:addon_version
+
+        This is used because some addons are missing a version. In order to
+        simplify subsequent queries, we make sure the format is consistent.
+        """
+        return ':' in addon and addon or addon + ':NO_VERSION'
```
willkg: Python has a "ternary" format. This would be better written as:

```
return addon if ':' in addon else ':NO_VERSION'
```

It reads a little weird if you're used to ternary operators in other languages, but the semantics are more correct and handle non-string values better.

PR #3794, opened 2017-05-29
Fixes bug 1368495 - Stop reading from bugs table in bugs service.

adngdb:


socorro/external/postgresql/bugs.py
```diff
@@ -40,23 +40,21 @@ def get(self, **kwargs):
...
-                SELECT ba.signature, bugs.id
-                FROM bugs
-                    JOIN bug_associations AS ba ON bugs.id = ba.bug_id
+                SELECT signature, bug_id as id
+                FROM bug_associations
                 WHERE EXISTS(
-                    SELECT 1 FROM bug_associations
-                    WHERE bug_associations.bug_id = bugs.id
+                    SELECT 1 FROM bug_associations as ba
+                    WHERE ba.bug_id = bug_associations.bug_id
```
willkg: I'm not sure I understand this. I think you don't need the join anymore because you're only looking at a single table. Maybe this is better?:

```
SELECT 1 from bug_associations
WHERE signature IN %s
```

Does that sound right?

willkg:
What does the bugs service get used by? If we get this wrong, what're the ramifications to Socorro and Socorro users?

socorro/external/postgresql/bugs.py
adngdb: So that's what I initially thought, but the sub-query actually is needed, because, we want all associations for all bugs that have the given signatures. Say, for example, we have this data: 

bug id | signature
--------|-------------
1 | foo
1 | bar
2 | foo

If you ask for bugs related to ``foo``, the service should actually return all three results, including the second one, ``1 | bar``, even though that one doesn't have ``foo`` as a signature.

adngdb:
The ``bugs`` service is used by the webapp to show the bugs associated to a signature in Top crashers, Signature report and Index report (the individual crash page). We have unit test coverage that passes with these changes, which I think is a good proof that this change is valid (also because I one of my changes broke them during development, and helped me realize that the aforementioned sub-query was actually needed).

socorro/external/postgresql/bugs.py
willkg: Oh, my bad. I see what's going on.

PR #3795, opened 2017-05-30
Comment out failing circle step

willkg:


willkg:
Yay--self-r+ because it's just commenting it out.

PR #3796, opened 2017-05-30
Add .cache to dockerignore

willkg:
This will ignore .cache directory for COPY steps.


willkg:
I think this will fix @adngdb's problem where it's copying over the `.cache` directory, but it has bad permissions.

willkg:
I'm going to self-r+ this. It affects the docker infrastructure only.

PR #3797, opened 2017-05-30
Fixes bug 1368726 - Fix perms for data container

willkg:
This fixes creation of the data container so that the files and directories have
the correct permissions so that the tests can run (and possibly fail) and
everything works right.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,7 @@
 .git
 **/*.swp
 **/*.pyc
+__pycache__
 .vagrant
 Vagrantfile
 .cache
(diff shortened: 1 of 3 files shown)


willkg:
@adngdb See if this works for you.

adngdb:
The permission issue is resolved with this, but I still get a lot of errors like this, preventing the tests from running: https://pastebin.mozilla.org/9023170

I assume it is unrelated, if so, r+ on this.

willkg:
Those errors are coming from running tests on your host computer. They shouldn't have been affected by running the tests in the docker environment with "make dockertest" since that doesn't touch your host machine. I'm not sure where those python cache files are coming from.

Maybe try setting:

```
PYTHONDONTWRITEBYTECODE=1
```

in your environment?

willkg:
I pushed another tweak to norecursedirs, though I'm skeptical it'll do anything.

Regardless, I'll wait until CI passes and then land this.

PR #3801, opened 2017-05-31
circleci conf: require DOCKER_DEPLOY=true to do any dockerhubby things

milescrabill:
@willkg r?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,6 +71,6 @@ jobs:
               bin/ci/deploy-dockerhub.sh latest
               bin/ci/deploy-dockerhub.sh "circle_build_$CIRCLE_BUILD_NUM"
             fi
-            if [ -n "${CIRCLE_TAG}"]; then
+            if [ -n "${CIRCLE_TAG}" ]; then
               bin/ci/deploy-dockerhub.sh "$CIRCLE_TAG"
             fi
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -72,5 +72,5 @@ jobs:
               bin/ci/deploy-dockerhub.sh "circle_build_$CIRCLE_BUILD_NUM"
             fi
             if [ -n "${CIRCLE_TAG}"]; then
```
willkg: I'm pretty sure you need a space between `"` and `]` at the end of this test, too.

willkg: Shame you can't run shellcheck on this circle config file. Maybe this is complex enough it's worth moving to a shell script?

bin/ci/deploy-dockerhub.sh
```diff
@@ -22,17 +22,19 @@ function retry() {
     return 0
 }
 
-# configure docker creds
-retry 3 docker login -e="$DOCKER_EMAIL" -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD"
+if [[ "$DOCKER_DEPLOY" == true ]]; then
```
willkg: The Antenna version doesn't have this. Is it good to keep these files in sync between projects?

milescrabill: Yes - definitely a good idea. I'll put in a PR now.

PR #3802, opened 2017-06-01
Fixes bug 1361832 - remove socorro collector and crashmover

willkg:
This removes the socorro collector and crashmover, some of the things that used
it, tests, scripts, configuration, and various related errata.

This moves the submitter (which we're still using) to it's own Python module.

r?


willkg:
This doesn't remove the documentation. The docs are way out of date and instead of spending time updating them which is non-trivial, we should probably wipe them and start anew or something along those lines. That's a project for another day.

willkg:
If the tests pass, then that's cool. Otherwise to test this, you could start the webapp and processor docker containers and if they start up, then we're probably fine.

We'll know really quickly if we're not fine when this hits -stage.

willkg:
Also, this doesn't remove *everything*. There's probably a bunch of crashmover classes floating around that are now orphaned as well as utility classes and probably some other things.

We can probably remove some of the metrics infrastructure that I wrote specifically for the collector, too. We can do that in future passes, though.

socorro/external/postgresql/crash_migration_app.py
```diff
@@ -42,66 +41,5 @@ def _get_processed_crash_transaction(self, connection, crash_id):
...
-            raw_crash,
-            dumps,
-            processed_crash,
-            crash_id
-        )
-        self.config.logger.info('saved - %s', crash_id)
-
-
 if __name__ == '__main__':
     main(CrashMigrationApp)
```
adngdb: This needs to be removed as well. Actually, this entire file should be removed, since the remaining ``DumbPostgreSQLCrashStorage`` class is only used in here.

willkg: Cool! Thank you!

willkg:
^^^ That rebases against master and also removes the unused file.

adngdb:
r+

willkg:
I think Circle is failing on a legitimate issue that has nothing to do with this PR but rather something in the world changed. I'll deal with that in a separate PR.

Squashing and merging this now.

Thank you!

PR #3803, opened 2017-06-04
Fixes bug 1368698 - Improve tests infrastructure for dev

willkg:
This improves the test infrastructure in docker for development. It adds a
"dockertestshell" make rule which sets up the docker environment with the repo
directory mounted in the container and then starts a shell. With this,
developers can run individual tests and also pass in different arguments to the
test runners.

This also documents all that in the docker README.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ json_enhancements_pg_extension: bootstrap
 
 # Docker related rules
 
-.PHONY: dockerbuild dockersetup dockerclean dockertest dockerrun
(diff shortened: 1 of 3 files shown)


willkg:
@adngdb Does this work for you and cover your testing needs?

willkg:
I'd need to see the errors you're getting to debug `make dockertest` further. So far the problems you've been having are because you're running the tests on the host as well as in the container, so it's probably from state being transferred that shouldn't be. It shouldn't be transferring state, so those are legitimate issues. Probably worth either re-opening the previous bug or starting a new one.

willkg:
I rebased against master and will land after that passes.

PR #3804, opened 2017-06-05
Fixes bug 1368498 - Do not use bugs table in bugzilla cron job.

adngdb:
This makes the bugzilla cron job stop writing any data into the bugs table. It also removes the foreign key from bug_associations to bugs in a migration, which is needed for the cron job to keep working.

### DIFF
--- a/socorro/cron/jobs/bugzilla.py
+++ b/socorro/cron/jobs/bugzilla.py
@@ -13,7 +13,6 @@ from crontabber.mixins import with_postgres_transactions
 
 from socorro.lib.datetimeutil import utc_now
 from socorro.external.postgresql.dbapi2_util import (
-    single_row_sql,
     execute_query_fetchall,
     execute_no_results,
     SQLDidNotReturnSingleRow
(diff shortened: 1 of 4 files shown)


willkg:
@adngdb This needs to be rebased--can you do that before I look through it?

socorro/cron/jobs/bugzilla.py
```diff
@@ -169,74 +134,31 @@ def inner_transaction(
...
-
-        if useful:
-            if insert_made:
-                self.config.logger.info(
-                    'new bug: %s - %s, %s, "%s"',
-                    bug_id,
-                    status,
-                    resolution,
-                    short_desc
+                execute_no_results(
```
willkg: This if block looks different. It used to only insert into `bug_associations` if `_has_signature_report` returned True, but now it does it all the time.

It looks like you took that method out, but it uses the `reports` table and not the `bugs` table, so I'm puzzled.

adngdb: Right, I forgot to mention it but I am also removing the call to ``reports``. We want to get rid of that eventually, and I suspect that's a source of problems anyway. It is possible that some signatures are not making it to our reports table for whatever reason, in which case bug associations might not happen. Considering the amount of data is not humongous and we need to stop using ``reports``, I made the call to remove that now.

adngdb:
I'll merge tomorrow, I want this to run on stage a bit before we push it to prod. Thanks for the review @willkg!

PR #3805, opened 2017-06-06
Fixes bug 1370561 - Ignored IOError while trying to extract memory re…

adngdb:
…port during processing.


socorro/processor/mozilla_transform_rules.py
```diff
@@ -362,14 +366,15 @@ def _extract_memory_info(self, dump_pathname, processor_notes):
                         self.config.max_size_uncompressed,
                     )
                 )
-                processor_notes.append(error_message)
-                return {"ERROR": error_message}
+                return error_out(error_message)
 
             memory_info = json_loads(memory_info_as_string)
+        except IOError, x:
```
willkg: These exception statements should be in this form:

```
except IOError as x:
```

PR #3806, opened 2017-06-07
Fixes bug 1366814 - Accept square brackets in signatures in bugzilla …

adngdb:
…cron job.

This should not go before #3804 as they are going to conflict (and rebasing this one is going to be easier).

### DIFF
--- a/socorro/cron/jobs/bugzilla.py
+++ b/socorro/cron/jobs/bugzilla.py
@@ -34,6 +34,31 @@ _URL = (
 )
 
 
+def find_signatures(content):
+    """Return a list of signatures found inside a string.
+
+    Signatures are found between `[@` and `]`. There can be any number of them
+    in the content.
+
+    Example:
+    >>> "some [@ signature] [@ another_one] and [@ even::this[] ]"
+    set(['signature', 'another_one', 'even::this[]'])
+    """
+    if not content:
+        return set()
+
+    signatures = set()
+    parts = content.split('[@')
+    for part in parts[1:]:
+        try:
+            last_bracket = part.rindex(']')
+            signature = part[:last_bracket].strip()
+            signatures.add(signature)
+        except ValueError:
+            pass
+    return signatures
+
+
 class NothingUsefulHappened(Exception):
     """an exception to be raised when a pass through the inner loop has
     done nothing useful and we wish to induce a transaction rollback"""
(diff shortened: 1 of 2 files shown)


willkg:
@adngdb ^^^ I clicked on the "update branch" button and it did a merge commit from master. I think that makes me unenthused about the "update branch" button.

Do you want to force push that merge commit out?

adngdb:
Done!

socorro/cron/jobs/bugzilla.py
```diff
@@ -36,6 +36,31 @@
...
 
 
+def find_signatures(content):
+    """Return a list of signatures found inside a string.
+
+    Signatures are found between `[@` and `]`. There can be any number of them
+    in the content.
+
+    Example:
+    >>> "some [@ signature] [@ another_one] and [@ even::this[] ]"
```
willkg: This needs to be passed to `find_signatures` like this:

```
>>> find_signatures("some [@ signature] [@ another_one] and [@ even::this[] ]")
set(['signature', 'another_one', 'even::this[]'])
```

Also, does this work only on strings? What happens with unicode? Is it ever possible to get a unicode signature?

socorro/cron/jobs/bugzilla.py
```diff
@@ -36,6 +36,31 @@
...
+    set(['signature', 'another_one', 'even::this[]'])
+    """
+    if not content:
+        return set()
+
+    signatures = set()
+    parts = content.split('[@')
+    for part in parts[1:]:
```
willkg: Why does this skip index 0?

socorro/cron/jobs/bugzilla.py
```diff
@@ -36,6 +36,31 @@
...
+        try:
+            last_bracket = part.rindex(']')
+            signature = part[:last_bracket].strip()
+            signatures.add(signature)
+        except ValueError:
```
willkg: In what circumstances do you get a `ValueError`?

socorro/unittest/cron/jobs/test_bugzilla.py
```diff
@@ -377,3 +378,28 @@ def test_run_job_based_on_last_success(self):
         cursor.execute('select id, short_desc from bugs where id = 8')
         bug = cursor.fetchone()
         eq_(bug[1], 'newlines in sigs')
+
+    def test_find_signatures(self):
```
willkg: This would be easy to rewrite using parametrize:

```
@pytest.mark.parametrize('content, expected', [
    # Simple signature
    ('[@ moz::signature]', set(['moz::signature'])),

    # 2 signatures and some junk
    ('@@3*&^!~[@ moz::signature][@   ns::old     ]', set(['moz::signature', 'ns::old'])),
    ...
])
def test_find_signatures(self, content, expected):
    assert find_signatures(content) == expected
```

adngdb:
@willkg I have addressed your comments, and answered your questions with comments in the code. :)

socorro/cron/jobs/bugzilla.py
```diff
@@ -36,6 +36,35 @@
...
+    >>> find_signatures("some [@ signature] [@ another] and [@ even::this[] ]")
+    set(['signature', 'another', 'even::this[]'])
+    """
+    if not content:
+        return set()
+
+    signatures = set()
+    parts = content.split('[@')
+    # The first item of this list is always not interesting, as it cannot
+    # contain any signatures. We thus skip it.
```
willkg: Ahh... I see. If `[@` is at the beginning of the string, then the first item is an empty string.

PR #3807, opened 2017-06-08
Fix "make dockertest" to sync files better

willkg:
When you run "make dockertest" it copies the contents of the repository
directory into /app in a data container and then runs the tests in that. The
problem is that it doesn't remove files that have been removed.

This fixes the script to remove all the files in /app and then copy the files
from the repository directory into /app. In this way, it better handles files
that have been removed.


willkg:
I'm going to self-r+ this because it's minor, but blocks another PR I have.

PR #3808, opened 2017-06-08
Bug 1361764 - remove skip_if stuff

willkg:
Only one test is skipped. This removes all the skip_if infrastructure and
switches that test to use the pytest skipif decorator.


willkg:
I'm not sure why circle is failing. I pushed another PR that fixes the tests so that they're more careful about copying the repo root into the data container and you see that code kick off in the logs. There shouldn't be any `skip_if` stuff left and I'm not sure where the residual bits could be in the circle environment.

Anyhow, because I did this PR and the fix-dockertest PR at the same time and then rebased this one, I'm inclined to think that circle is just confused by my superfly development and that this isn't a real problem. If we see something like this again in a future PR, we can declare it a real problem and I can look into it further.

peterbe:
So the tests fail on `socorro/unittest/test_skip_if.py` even though the PR is (almost) all about deleting `socorro/unittest/test_skip_if.py`. Could it be that there are `.pyc` files lingering between tests that pytest finds?

willkg:
Circle runs the tests with the repo copied into a volume into a data container. The PR I did earlier deletes everything in the volume before doing a `docker cp`.

Further, we run with `PYTHONDONTWRITEBYTECODE=1` (https://github.com/mozilla-services/socorro/blob/ae914aef1577a754fa654de15919adbdf041ba05/docker/Dockerfile#L16) so there shouldn't be any `.pyc` files.

So, I guess it's possible, but I don't know how they would come into being or how they'd get cached.

I'll toss a `ls -l "/app/*` line in there and see if that sheds any light on things.

willkg:
I figured it out--the fix I had in the previous PR didn't work. ^^^ That should fix it.

willkg:
After that last fix, I can't reproduce the issue, so I'm back to being puzzled. My best guess is a hand-wavey "maybe something something cache something something container in cirlce land".

Regardless, this is a problem with the test infrastructure or Circle and not with the PR, so I'm back to voting we ignore Circle for now and press forward and if this happens with another PR we can look into it further.

willkg:
I think it's a bug in circle. This branch clearly has no `test_skip_if.py` file, but I added a `ls -l socorro/unittest/` to the `docker/run_tests_in_docker.sh` which prints out what circle is testing and it lists this:

```
+ (run_tests_in_docker.sh): ls -l socorro/unittest/
total 52
-rw-r--r-- 1 root root    0 Jun 10 01:49 __init__.py
drwxr-xr-x 2 root root 4096 Jun 10 01:49 app
drwxr-xr-x 3 root root 4096 Jun 10 01:49 cron
drwxr-xr-x 2 root root 4096 Jun 10 01:49 database
drwxr-xr-x 9 root root 4096 Jun 10 01:49 external
drwxr-xr-x 2 root root 4096 Jun 10 01:49 lib
drwxr-xr-x 3 root root 4096 Jun 10 01:49 middleware
drwxr-xr-x 3 root root 4096 Jun 10 01:49 processor
drwxr-xr-x 2 root root 4096 Jun 10 01:49 siglists
-rw-r--r-- 1 root root 1021 Jun 10 01:49 skip_if.py
drwxr-xr-x 2 root root 4096 Jun 10 01:49 submitter
-rw-r--r-- 1 root root  505 Jun 10 01:49 test_skip_if.py
-rw-r--r-- 1 root root  110 Jun 10 01:49 testbase.py
drwxr-xr-x 2 root root 4096 Jun 10 01:49 webapi
```

This isn't a listing of the data in the container--this is a listing of the data in `/socorro` which is supposedly the git checkout that circle does before running the tests. It's clearly wrong--there are files in that directory that are explicitly removed in this branch.

I added a "DEBUG file listing" step before "make dockertest" and printed out the contents of `/socorro/socorro/unittest` and it's totally wrong. You can see that output in this build:

https://circleci.com/gh/mozilla-services/socorro/144?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

It's a problem either with our circle configuration or circle itself.

I tried doing a rebuild with no cache, but that didn't help at all. I can't figure out what's reconstituting those deleted files from the grave.

@milescrabill Any ideas? How can this branch delete files, then circle go and run the tests with those files existing?

peterbe:
In the "Checkout code" section it says:
*"Warning: Either git or ssh (required by git to clone through SSH) is not installed in the image. Falling back to CircleCI's native git client but this is still an experiment feature. We highly recommend using an image that has official git and ssh installed."*

Which is odd because isn't that a core feature of a CI to be able to do a relevant git checkout.

peterbe:
Perhaps I just don't know how to use the CircleCI SSH, but this is odd:
```
root@d6a3721980e7:/socorro# cat .git/HEAD
ref: refs/heads/master
```

willkg:
Typically what CI systems do is merge the PR, then run the tests. That way you know whether the results of the PR merged pass or not.

peterbe:
Ah. I did open `./socorro/unittest/processor/test_symbol_cache_manager.py` and it correctly has the right changes. 

Seems it has an issue with deleted files. 
Strange that it even failed with a "Rebuild without cache" kinda thing.

willkg:
I'm going to land this and if we have issues in the future we can debug circle stuff then.

PR #3809, opened 2017-06-13
Fixes bug 1370484 - Refactored bugzilla query to use REST API and JSON.

adngdb:
This query used to be a very long, hand-written URL, using the old bugzilla query system that returned CSV data. With this, it now uses the REST API instead, manipulating JSON data with the requests library. The query is much simpler to read and has only the few arguments it actually needs. Also, the query is not configurable anymore, because that option was never used and it would have made the code less maintainable.

### DIFF
--- a/socorro/cron/jobs/bugzilla.py
+++ b/socorro/cron/jobs/bugzilla.py
@@ -59,12 +59,6 @@ def find_signatures(content):
     return signatures
 
 
-class BugzillaConnectionError(Exception):
-    """Raised when an HTTP request to bugzilla did not return a 2XX status.
-    """
-    pass
-
-
 class NothingUsefulHappened(Exception):
     """an exception to be raised when a pass through the inner loop has
     done nothing useful and we wish to induce a transaction rollback"""
@@ -185,7 +179,7 @@ class BugzillaCronApp(BaseCronApp):
         payload['chfieldfrom'] = from_date
         r = requests.get(BUGZILLA_BASE_URL, params=payload)
         if r.status_code < 200 or r.status_code >= 300:
-            raise BugzillaConnectionError()
+            r.raise_for_status()
         results = r.json()
         for report in results['bugs']:
             yield (
--- a/socorro/cron/jobs/bugzilla.py
+++ b/socorro/cron/jobs/bugzilla.py
@@ -3,9 +3,8 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import datetime
-import urllib2
-import csv
 
+import requests
 from dateutil import tz
 from configman import Namespace
 from crontabber.base import BaseCronApp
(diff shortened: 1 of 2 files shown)


socorro/cron/jobs/bugzilla.py
```diff
@@ -155,10 +145,14 @@ def inner_transaction(
         if not useful:
             raise NothingUsefulHappened('nothing useful done')
 
-    def _iterator(self, query):
-        for report in csv.DictReader(urllib2.urlopen(query)):
+    def _iterator(self, from_date):
+        payload = BUGZILLA_PARAMS.copy()
+        payload['chfieldfrom'] = from_date
+        r = requests.get(BUGZILLA_BASE_URL, params=payload)
+        results = r.json()
```
peterbe: One thing we've often seen is that bugzilla's rest API simply fails. And when it does, the error message is in HTML. 
It's more of a nit but it might be worth doing an assertion of sorts on the `r.status_code` first. Otherwise you get a nasty JSONDecodeError which might be confusing. 
Either `assert r.status_code == 200, r.status_code` or `if r.status_code != 200: return`

socorro/unittest/cron/jobs/test_bugzilla.py
```diff
@@ -145,7 +158,8 @@ def test_run_job_with_reports_with_existing_bugs_same(self):
         assert ('another::legitimate(sig)',) in associations
         assert ('legitimate(sig)',) in associations
 
-    def test_run_job_based_on_last_success(self):
+    def test_run_job_based_on_last_success(self, mocker):
+        mocker.get(BUGZILLA_BASE_URL, json=SAMPLE_BUGZILLA_RESULTS)
         """specifically setting 0 days back and no prior run
         will pick it up from now's date"""
```
willkg: The docstring should go before the code in this method.

socorro/unittest/cron/jobs/test_bugzilla.py
```diff
@@ -87,7 +98,8 @@ def test_basic_run_job(self):
         assert ('another::legitimate(sig)',) in associations
         assert ('legitimate(sig)',) in associations
 
-    def test_run_job_with_reports_with_existing_bugs_different(self):
+    def test_run_job_with_reports_with_existing_bugs_different(self, mocker):
+        mocker.get(BUGZILLA_BASE_URL, json=SAMPLE_BUGZILLA_RESULTS)
         """Verify that an association to a signature that no longer is part
         of the crash signatures list gets removed.
         """
```
willkg: The docstring should go before the code in this method.

socorro/unittest/cron/jobs/test_bugzilla.py
```diff
@@ -34,28 +58,15 @@ def tearDown(self):
...
         return get_config_manager_for_crontabber(
             jobs='socorro.cron.jobs.bugzilla.BugzillaCronApp|1d',
             overrides={
-                'crontabber.class-BugzillaCronApp.query': query,
                 'crontabber.class-BugzillaCronApp.days_into_past': days_into_past,
             }
         )
 
-    def test_basic_run_job(self):
+    def test_basic_run_job(self, mocker):
```
willkg: I would have named the `mocker` argument `requests_mocker` or something like that to make it clearer what it's mocking. But it's not a big deal because these are short methods.

socorro/cron/jobs/bugzilla.py
willkg: Oo--I missed this because reasons. I agree with this, too. Definitely worth doing something here so that errors are clearer for future debuggers.

adngdb:
@peterbe @willkg What do you think about this last commit?

socorro/cron/jobs/bugzilla.py
```diff
@@ -184,9 +180,15 @@ def inner_transaction(
...
             raise NothingUsefulHappened('nothing useful done')
 
-    def _iterator(self, query):
-        for report in csv.DictReader(urllib2.urlopen(query)):
+    def _iterator(self, from_date):
+        payload = BUGZILLA_PARAMS.copy()
+        payload['chfieldfrom'] = from_date
+        r = requests.get(BUGZILLA_BASE_URL, params=payload)
+        if r.status_code < 200 or r.status_code >= 300:
+            raise BugzillaConnectionError()
```
willkg: This should have some kind of message with the status code any anything else that might be helpful in figuring out why it failed to connect.

peterbe:
@adngdb you might have already noticed but you accidentally committed some other work into this branch.

adngdb:
@peterbe Well, I was wondering where that commit went... Thanks!

willkg:
What's this waiting on? Can we merge it now?

adngdb:
@willkg this was waiting for that last commit. I'm using ``requests.Response.raise_for_status`` for error reporting, is that fine by you?

willkg:
Looks good to me!

PR #3812, opened 2017-06-15
Add rust panic symbols to prefix list

jrmuizel:


socorro/siglists/prefix_signature_re.txt
```diff
@@ -188,6 +189,8 @@ ssl_.*
 SSL_.*
 std::_Allocate.*
 std::list<.*>::.*
+std::panicking::begin_panic<&str>
```
adngdb: I am not sure how this is going to work with the ``<&str>`` bit. Maybe you should rewrite that to ``<.*>`` like we do elsewhere? 

The thing is, we have a normalization process that kicks in before we get to the signature generation part, and that turns stuff like ``<parm1, param2>`` into ``<T>``. I _assume_ that would apply here as well, and so the frame signature would become ``std::panicking::begin_panic<T>``.

jrmuizel:
Updated to fix review comments.

PR #3816, opened 2017-06-20
Fixes bug 1372999 - add document size metrics for elasticsearch crashstorage

willkg:
I had to implement a new metrics infrastructure before I could do anything. But I think that's good in the long run. We can probably retire a bunch of the statsd stuff we've got now and add more granular metrics capture to relevant parts.

The metrics infrastructure degrades to logging metrics data if statsd isn't set up. This re-uses existing statsd configuration keys, so theoretically, it should just work without requiring configuration changes.

r?

### DIFF
--- a/socorro/app/socorro_app.py
+++ b/socorro/app/socorro_app.py
@@ -362,6 +362,61 @@ def setup_logger(config, local_unused, args_unused):
     return wrapped_logger
 
 
+def setup_metrics(config, local_unused, args_unused):
+    """Sets up metrics client which is either a DogStatsd client or a LoggingClient
+
+    NOTE(willkg): This is a little gross, but that's mostly to keep all the code in one place and
+    minimal which makes it easier to deal with for now. If this ever grows, then we'll probably want
+    to split this out into its own module.
+
+    This gets tossed in config (along with everything else), so you can use it like this::
+
+        class SomeComponent(RequiredConfig):
+            def something(self):
+                self.config.metrics.increment('somekey')
+
+    """
+    # We do the import here in case there are parts of Socorro that are running in environments that
+    # don't have this library installed.
+    #
+    # FIXME(willkg): We're using a *super* old version of dogstatsd-python which doesn't have the
+    # DogStatsd class. We should upgrade.
+    try:
+        from datadog.dogstatsd import statsd
+    except ImportError as ioe:
+        statsd is None
+
+    host = config.metricscfg.statsd_host
+    port = config.metricscfg.statsd_port
+
+    if host and statsd:
+        statsd.host = host
+        statsd.port = port
+        return statsd
+
+    class LoggingMetrics(object):
+        """Logging-based metrics class that mimics DogStatsd class"""
+        def __init__(self):
+            self.logger = logging.getLogger('socorro.metrics')
+
+        def _log(self, operation, metric, value, tags):
+            self.logger.info('%s: %s=%s tags=%s', operation, metric, value, tags or [])
+
+        def increment(self, metric, value=1, tags=None):
+            self._log('increment', metric, value, tags)
+
+        def gauge(self, metric, value, tags=None):
+            self._log('gauge', metric, value, tags)
+
+        def timing(self, metric, value, tags=None):
+            self._log('timing', metric, value, tags)
+
+        def histogram(self, metric, value, tags=None):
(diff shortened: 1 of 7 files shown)


socorro/app/socorro_app.py
```diff
@@ -362,6 +362,61 @@ def setup_logger(config, local_unused, args_unused):
...
+        class SomeComponent(RequiredConfig):
+            def something(self):
+                self.config.metrics.increment('somekey')
+
+    """
+    # We do the import here in case there are parts of Socorro that are running in environments that
+    # don't have this library installed.
+    #
+    # FIXME(willkg): We're using a *super* old version of dogstatsd-python which doesn't have the
+    # DogStatsd class. We should upgrade.
```
willkg: Covered in https://bugzilla.mozilla.org/show_bug.cgi?id=1374507 .

socorro/app/socorro_app.py
```diff
@@ -362,6 +362,61 @@ def setup_logger(config, local_unused, args_unused):
...
+    try:
+        from datadog.dogstatsd import statsd
+    except ImportError as ioe:
+        statsd is None
```
willkg: I have no idea if there exist some things that use `SocorroApp` or some subclass that run in an environment that doesn't have the dogstatsd-python library. So I did this goofy thing.

Amongst other things, it makes it easier to deal with in the tests since many tests create configuration environments, so because of how this is structured, all those tests use `LoggingMetrics` without any adjustment. Yay!

socorro/external/es/crashstorage.py
```diff
@@ -119,6 +120,29 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
             'raw_crash': raw_crash
         }
 
+        # NOTE(willkg): this is a hard-coded keyname to match what the statsdbenchmarkingwrapper
+        # produces for processor crashstorage classes. This is only used in that context, so we're
+        # hard-coding this now rather than figuring out a better way to carry that name through.
+        try:
+            self.config.metrics.histogram(
+                'processor.es.raw_crash_size',
+                len(json.dumps(raw_crash, cls=JsonDTEncoder))
```
willkg: I don't know offhand what the performance consequences of serializing these again are. I'm hoping it's not too much of a change compared to the time it normally would take. If it ends up taking more time than we'd like, we can adjust this code to sample metrics gathering and only capture 1/100 or something like that.

socorro/app/socorro_app.py
adngdb: I am puzzled by that syntax. As far as I know, it is not valid, as it will always raise a ``NameError`` (since ``statsd`` is not defined here). What is you goal with that? Did you simply want to write ``statsd = None`` instead? 

Also, you declare ``ioe`` but don't use it, maybe you can remove it?

socorro/external/es/crashstorage.py
```diff
@@ -119,6 +120,29 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
...
+            )
+        except Exception as exc:
+            # NOTE(willkg): An error here shouldn't screw up saving data. Log it so we can fix it
+            # later.
+            self.config.logger.exception('something went wrong when capturing raw_crash_size')
```
adngdb: Don't you want to add the stack trace with this, so it's easier to debug?

socorro/external/es/crashstorage.py
```diff
@@ -119,6 +120,29 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
...
+
+        try:
+            self.config.metrics.histogram(
+                'processor.es.processed_crash_size',
+                len(json.dumps(processed_crash, cls=JsonDTEncoder))
+            )
+        except Exception as exc:
+            # NOTE(willkg): An error here shouldn't screw up saving data. Log it so we can fix it
+            # later.
+            self.config.logger.exception('something went wrong when capturing raw_crash_size')
```
adngdb: Same here.

socorro/app/socorro_app.py
willkg: Oops--the `is` is totally a typo. I'll fix that. And the `ioe` is from debugging. I'll fix that, too.

Good catch!!

willkg:
Thank you!

Also, I'll write up an email to socorro-dev about the new metrics infrastructure.

PR #3818, opened 2017-06-20
Add a proper exception to LoggerWrapper

willkg:
The LoggerWrapper didn't have a .exception() method. It's a handy shortcut
that's used in a bazillion places and if Socorro is going to wrap Logger, it
should mimic it properly.

r?


willkg:
Thank you!

PR #3819, opened 2017-06-20
Fixes bug 1372999 - fix statsd host/port configuration

willkg:
This fixes the configman options for statsd_host and statsd_port so they look in
the right place for values.

r?


willkg:
I ran it locally to verify that it was picking up the value correctly. I'm not sure how else to verify it.

PR #3822, opened 2017-06-21
Bug 1353371 - change middleware service to be a no-op

willkg:
r?

### DIFF
--- a/config/package/usr/lib/systemd/system/socorro-middleware.service
+++ b/config/package/usr/lib/systemd/system/socorro-middleware.service
@@ -3,8 +3,8 @@ Description=Socorro Middleware
 
 [Service]
 Environment=VENV=/data/socorro/socorro-virtualenv
-ExecStart=/bin/bash -c "envconsul -upcase=false -prefix socorro/common -prefix socorro/middleware $VENV/bin/uwsgi -H $VENV -M --need-app -w wsgi.middleware -s /var/run/uwsgi/socorro/socorro-middleware.sock --chmod-socket=664 --uid=socorro --gid=nginx"
-Restart=always
+ExecStart=/bin/bash -c "echo No more middleware"
+Restart=no
 
 [Install]
 WantedBy=multi-user.target


willkg:
I tested this out on a -stage webhead where I:

1. did `systemctl stop socorro-middleware` to run the currently running one
2. replaced `socorro-middleware.service` with this one
3. did `systemctl daemon-reload` to pick up the new `.service` file
4. did `systemctl start socorro-middleware` to kick off the new one which is a non-restarting no-op
5. verified that it echoed "No more middleware" in `/var/log/messages` and that there were no middleware processes running

That worked super.

The one minor thing I'm not sure about is that the middleware process is a uwsgi process. There's an nginx conf file that sets up to proxy incoming requests to that process. If the process is not running, I think nginx will just throw a 500 error and it's not a big deal. Does that sound right?

config/package/usr/lib/systemd/system/socorro-middleware.service
```diff
@@ -3,8 +3,8 @@ Description=Socorro Middleware
 
 [Service]
 Environment=VENV=/data/socorro/socorro-virtualenv
-ExecStart=/bin/bash -c "envconsul -upcase=false -prefix socorro/common -prefix socorro/middleware $VENV/bin/uwsgi -H $VENV -M --need-app -w wsgi.middleware -s /var/run/uwsgi/socorro/socorro-middleware.sock --chmod-socket=664 --uid=socorro --gid=nginx"
-Restart=always
+ExecStart=/bin/bash -c "echo No more middleware"
```
willkg: This will print a line to `/var/log/messages` and then the process ends.

config/package/usr/lib/systemd/system/socorro-middleware.service
```diff
@@ -3,8 +3,8 @@ Description=Socorro Middleware
...
+Restart=no
```
willkg: This makes sure that when the process ends, it doesn't get restarted. Thus systemd will bring this service up, it echos a line, it ends, and then systemd is all like, "well, fine. be that way."

willkg:
^^^ Updated the branch because of this branch-update nonsense.

PR #3823, opened 2017-06-21
Fixes bug 1368498 - Do not error out if crash signature field is missing.

adngdb:


willkg:
I'm going to merge this now so that it's ready when @adngdb is around and he doesn't have to wait.

PR #3826, opened 2017-06-22
Fixes bug 1353371 - Removed all remnants of the deceased middleware.

adngdb:


peterbe:
We have some really odd behaviors in Circle.

willkg:
The circle failure is exactly the problem I was having in another PR where we remove a file in the PR, but the tests end up running in an environment that has that file in it.

I thought maybe I had fixed that, but (hand-waving) things conspired to make that particular PR continue to break. But this clearly proves I didn't fix it and it's still an issue. I wrote up this bug for it: https://bugzilla.mozilla.org/show_bug.cgi?id=1375551

willkg:
I'm going to merge this. It'll be good to let it sit on stage for a bit.

PR #3828, opened 2017-06-22
Fixed missing_symbols partition creation

Lehych:
This should prevent fresh socorro installation from wrong state of
report_partition_info table.

Think this relates Bug 1123742.

### DIFF
--- a/socorro/external/postgresql/staticdata.py
+++ b/socorro/external/postgresql/staticdata.py
@@ -180,7 +180,7 @@ class ReportPartitionInfo(BaseTable):
              '{}',
              'report_date', 'DATE'],
             ['missing_symbols', '14',
-             '{"date_processed,debug_file,debug_id"}',
+             '{}',
              '{}',
              '{}',
              'date_processed', 'DATE']]


willkg:
Is this tied to a bug? If not, can you explain the problem this is solving more?

Lehych:
This solvers issue with unnecessary pkey for missing_symbols table when you run setupdb_app.py.
This pkey leads to errors on MissingSymbolsRule in processor https://github.com/mozilla-services/socorro/blob/master/socorro/processor/mozilla_transform_rules.py#L758 since there is no check for existence.

Here is how this row migration look like in alembic https://github.com/mozilla-services/socorro/blob/master/alembic/versions/556e11f2d00f_bug_948644_record_missing_symbols_from_.py

I suppose it solves a Bug 1123742 I've mentioned in description.

willkg:
Whoops--my bad. We specify the bug in the commit summary and I didn't see it there, so I just missed it.

I'll try to look at this in the next couple of days.

willkg:
This seems fine to me and I think it only affects new Socorro installs which pretty much means local development at this stage. If this creates new issues, we can fix those and move forward.

Given that, I'm r+ on this and will merge it now.

PR #3829, opened 2017-06-22
1375551 fix circle ci 

willkg:


willkg:
@milescrabill and I worked through this. The changes are minimal and if it's still broken, we can continue to fix it because it can't be more broken than it is now. Given that, I'm going to summon the spirits of @milescrabill to r+ it.

milescrabill:
I DON'T KNOW, THESE ARE SOME PRETTY HUGE CHANGES

PR #3830, opened 2017-06-23
Fix instances of "except Exception, e" and some other things

willkg:
This fixes instances of `except Exception, e` (deprecated syntax) to `except Exception as e` (correct syntax and the only one supported by Python 3).

https://docs.python.org/3.0/whatsnew/3.0.html#changed-syntax

https://www.python.org/dev/peps/pep-3110/

There are a couple of other fixes that accidentally pop in, too.

r?


willkg:
This doesn't affect anything behavior-wise. I didn't see any cases where the ambiguous `,` syntax did things we weren't expecting.

willkg:
Thank you!

PR #3831, opened 2017-06-23
bug 1375426 - rewrite graphics report with SuperSearch

peterbe:
Here's how I tested it. 

1. Generate a CSV file (for a given Sunday):
(you'll need to use the master branch and be connected to stage's ES and PG)
```
curl -v -H "auth-token:41711e..." http://socorro.dev/graphics_report/\?date\=2017-06-18 > graphics-sql.csv
```
2. Switch on this branch and run it again:
```
curl -v -H "auth-token:41711e..." http://socorro.dev/graphics_report/\?date\=2017-06-18 > graphics-supersearch.csv
```
3. Now, to compare the two outputs, run this script:
```python
import csv, random

old = {}

with open('graphics-sql.csv') as f:
    reader = csv.DictReader(f, delimiter='\t')
    #reader = csv.reader(f, delimiter='\t')
    header = next(reader)
    for row in reader:
        #print(row)
        old[row['crash_id']] = row
        if len(old) == 10000:
            break

new = {}

with open('graphics-supersearch.csv') as f:
    reader = csv.DictReader(f, delimiter='\t')
    #reader = csv.reader(f, delimiter='\t')
    header = next(reader)
    for row in reader:
        new[row['crash_id']] = row
        if len(new) == 10000:
            break

for crash_id in random.sample(old.keys(), 100):
    try:
        this = old[crash_id]
        that = new[crash_id]
    except KeyError:
        continue
    assert len(this) == 30
    for key in this:
        value = this[key]
        their = that[key]
        if value != their:
            print('\t', key, (value, their))
    assert len(that) == 30
```


webapp-django/crashstats/crashstats/views.py
```diff
@@ -1103,28 +1105,97 @@ def graphics_report(request):
...
-        # exact order as the `header` tuple above.
-        # However, because the csv writer module doesn't "understand"
-        # python's None, we'll replace those with '' to make the
-        # CSV not have the word 'None' where the data is None.
-        writer.writerow([
-            row[x] is not None and row[x] or ''
-            for x in GRAPHICS_REPORT_HEADER
-        ])
+    pages = data['total'] // batch_size
+    pages += data['total'] % batch_size and 1 or 0
```
willkg: Given that I stopped and stared at this for a while to figure out what it's doing, I claim it's not readable. Seems like it's trying to figure out the total number of pages and to do that it's doing an integer division by `batch_size` and then if there are additional results, adding an additional page. Is that right?

There's got to be a more readable way of expressing that. Maybe something like this?:

```
pages = data['total'] // batch_size
if data['total'] % batch_size:
    pages += 1
```

Maybe combine that with `divmod`?:

```
pages, remainder = divmod(data['total'], batch_size)
if remainder:
    pages += 1
```

webapp-django/crashstats/crashstats/views.py
```diff
@@ -1103,28 +1105,97 @@ def graphics_report(request):
...
+        'crash_id': 'uuid',
+        'os_name': 'platform',
+        'os_version': 'platform_version',
+        'date_processed': 'date',
+        'build': 'build_id',
+        'uptime_seconds': 'uptime',
+    }
+    # Make sure that we don't have an alias for a header we don't need
+    alias_excess = set(alias.keys()) - set(GRAPHICS_REPORT_HEADER)
+    assert not alias_excess, alias_excess
```
willkg: This will generate an unhelpful exception with an unhelpful message. Something like ``AssertionError('5')``. That won't help us if that shows up in a Sentry error report. We'd have to drill down to figure out what's going on.

Definitely worth making this more helpful.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -1103,28 +1105,97 @@ def graphics_report(request):
...
+            return ''
+        if key == 'date_processed':
+            value = timezone.make_aware(datetime.datetime.strptime(
+                value.split('.')[0],
+                '%Y-%m-%dT%H:%M:%S'
+            ))
+            value = value.strftime('%Y%m%d%H%M')
+        if key == 'uptime_seconds' and value == 0:
+            value = ''
+        return value
```
willkg: This looks weird and there's no explanation for why it's this way. Why is the order of those `if` statements this way? What does the row/key data look like? What's this trying to do and trying to avoid?

Also, should this be an inner function? Maybe it should be external so it can be tested in isolation to make it easier to update in the future?

webapp-django/crashstats/crashstats/views.py
```diff
@@ -1103,28 +1105,97 @@ def graphics_report(request):
...
+
+    for page in range(pages):
+        if page > 0:
+            params['_results_offset'] = 500 * page
```
willkg: Where's this 500 from? Should this be `batch_size`?

peterbe:
> The way you tested it in your PR description is curious. Why sample the results? Why not verify everything is the same?

Because the CSV file is many many megabytes and I wouldn't be able to easily see it on the command line. 
But still a valid point. It could be done if I don't expect to have to "eye" any outputs. 
Will do it again when I've addressed all the other comments.

webapp-django/crashstats/crashstats/views.py
peterbe: I've added a doc string that explains that this whole function is just there to match what the ancient SQL based report use to be able to do.

peterbe:
Rebased and nits addressed. Also, I re-ran the end-to-end test but comparing every single row. Each CSV file is 14MB.

PR #3834, opened 2017-06-29
Fixes bug 1373997 - Docs overhaul

willkg:
* removes a bunch of obsolete documentation
* removes a bunch of tutorial documentation for postgres, git, and other tools
  for which there's better documentation out there on the internet
* moves some things around into a socorro-developer-centric structure
* adds docker environment documentation and fleshes that out a bit
* adds a bunch of FIXMEs
* fixes sphinx errors
* reformats headers and some other things (cosmetic)
* adds a CONTRIBUTING.rst file and move dev process stuff there

This is a big overhaul. It's probably impossible to review well. It moves the docs from 2 years out-of-date to something better, integrates the Docker environment docs, and removes some obviously obsolete things.

It could use more work, but I think it's probably good to do a first-pass quick review (don't spend a lot of time on it), then we should just land it and fail forward as we find issues.


willkg:
I added a note about adding the json schema readme and the signatures readme to the bug--that's a good idea. We can do it in a future PR.

Going to land this now. woohoo!

PR #3835, opened 2017-06-29
[DO NOT MERGE] fixes bug 1377254 - Run flake8 on all code in CI

peterbe:


setup.cfg
```diff
@@ -5,7 +5,7 @@ max-line-length = 100
 
 [tool:pytest]
 addopts = -rsxX --tb=native
-norecursedirs = .git docs scripts config docker __pycache__
+norecursedirs = __pycache__
```
willkg: I don't understand why this change happened. Why remove all these directories from the `norecursedirs`?

willkg:
We can't land this for a while, so I'm going to close it out. We can reopen when we're ready to do this.

PR #3836, opened 2017-06-29
Bug 1377254 flake8 unittests

peterbe:
baby steps.


willkg:
We don't want to merge anything until we're done with Elasticsearch migration testing. Marking the PR accordingly so we don't accidentally merge it.

PR #3837, opened 2017-06-30
dockerstop

peterbe:


willkg:
We don't want to merge anything until we're done with Elasticsearch migration testing. Marking the PR accordingly so we don't accidentally merge it.

PR #3839, opened 2017-07-03
Bug 1373997 - fixes toctree issue

willkg:
This fixes one of the options in the toctree declaration such that the docs
build a toctree.

It also fixes a syntax error and caps the number of levels in the toctree.


willkg:
I'm going to self-r+ this. It fixes the docs which currently don't have a toc, it's super basic, and I'm not sure anyone else is around this week.

PR #3840, opened 2017-07-04
Fresh socorro installations misses some win-versions

Lehych:
Changes from Bug899641 and Bug1132652 are not presented in staticdata
file that used on fresh socorro installations.
Here we add missed win versions.


peterbe:
@adngdb @willkg Do you use the static data thing? Can you review this?

willkg:
I verified the version numbers with https://msdn.microsoft.com/en-us/library/windows/desktop/ms724833(v=vs.85).aspx

This seems fine to me. As soon as the checks pass, I'll merge it.

PR #3843, opened 2017-07-07
Fixes bug 1379089 - add a couple of rust functions to the skip list

jcristau:


willkg:
This seems fine to me. I'll land it and check it on -stage after updating the branch.

PR #3844, opened 2017-07-07
Fixes bug 1100354 - Read Super Search Fields from a json file

adngdb:


adngdb:
I am going to self-r+ this because I have spent enough time on it, testing and re-reading my code, that I'm confident it's good to go. Also, there is more value in landing it now to progress with stage than in waiting for someone else to review. 

However, @willkg and @peterbe, if one of you wants to retroactively review this, please go ahead!

socorro/external/es/super_search_fields.py
```diff
@@ -2,16 +2,20 @@
...
+from socorro.lib import BadArgumentError, datetimeutil
 from socorro.external.es.base import ElasticsearchBase
-from socorro.lib import datetimeutil, external_common
+
+
+SUPER_SEARCH_FIELDS_JSON_PATH = os.path.join(
+    os.path.dirname(__file__),
+    'data',
+    'super_search_fields.json'
+)
```
willkg: Pretty sure this isn't going to work in -stage/-prod since it gets installed as a package and I don't see anything that adds the `super_search_fields.json` file to the `data_files` section of `setup.py`.

I think this should work like you did siglists:

https://github.com/mozilla-services/socorro/blob/master/socorro/siglists/__init__.py#L41

willkg: (This is a fly-by comment. I haven't tested this theory.)

adngdb: You are absolutely correct: https://sentry.prod.mozaws.net/operations/socorro-stage/issues/621154/

PR #3847, opened 2017-07-11
fixes bug 1379665 - Telemetry tab on report index

peterbe:
CC @adngdb @lonnen 

Looks like this:
screen shot 2017-07-11 at 1 17 31 pm

I'm not convinced this is the best way to display the data. The advantage is that it's simple. You can ⌘-F and find what you're looking for. 

Another alternative is to use the JSON syntax highlighter we use in the SuperSearch documentation:
screen shot 2017-07-11 at 1 20 20 pm
Then you can still use ⌘-F. 

At least it's there. 
But the PR lacks tests (since I'm not sure the work is ideal yet).

Another feature we could consider is one of those jQuery plugins (they all look kinda icky) where you can click little plus and minuses.

### DIFF
--- a/webapp-django/crashstats/crashstats/static/crashstats/css/tree.css
+++ b/webapp-django/crashstats/crashstats/static/crashstats/css/tree.css
@@ -1,0 +1,61 @@
+/* From http://cssdeck.com/labs/pure-css-tree-menu-framework */
+
+.tree, .tree ul {
+    margin: 0 0 0 2em;
+    /* indentation */
+    padding: 0;
+    list-style: none;
+    position: relative;
+}
+
+.tree ul {
+    margin-left: .5em
+}
+
+
+/* (indentation/2) */
+
+.tree:before, .tree ul:before {
+    content: "";
+    display: block;
+    width: 0;
+    position: absolute;
+    top: 0;
+    bottom: 0;
+    left: 0;
+    border-left: 1px solid;
+}
+
+.tree li {
+    margin: 0;
+    padding: 0 1.5em;
+    /* indentation + .5em */
+    line-height: 1.6em;
+    /* default list item's `line-height` */
+    font-weight: bold;
+    position: relative;
+}
+
+.tree li:before {
+    content: "";
+    display: block;
+    width: 10px;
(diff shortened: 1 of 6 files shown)


willkg:
I'm +1 on whatever is the simplest thing we can do that satisfies the requirements of whoever is going to use it. I suspect this iteration is fine. If not, we can iterate further later.

I'll wait to review this until it's ready to go.

peterbe:
@bsmedberg @marco-c Gentlemen, can you take a look at the screenshots + ideas here above.
Point is, I'm not a UX expert and most important thing is "Can you get your work done with this" and that we can improve on it later.

willkg:
This needs to land in the next couple of hours. So, if no one comments, so be it.

@peterbe Can you write a test for this view?

bsmedberg:
Either of these screenshots would be good. The JSON syntax highlighter seems slightly better mainly because you can collapse things you don't care about. But worse is better here!

webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -386,3 +387,41 @@ def show_filesize(bytes, unit='bytes'):
...
+        # then it has to be a JSON string
+        structure = json.loads(structure)
+
+    def tree(container, depth=1):
+        P = ' ' * 4
+        html = []
+
+        # The reason for using `key=lambda s: s.lower()` instead
+        # of just `str.lower` is because it's not a guarantee that
+        # all strings are byte strings. Or unicode for that matter.
```
willkg: I don't understand this comment at all. What're the two things you're comparing here?

webapp-django/crashstats/crashstats/static/crashstats/css/tree.css
```diff
@@ -0,0 +1,61 @@
+/* From http://cssdeck.com/labs/pure-css-tree-menu-framework */
```
willkg: I don't think you can legally lift that. There's no indication that you're allowed to copy it on that page or on that site as far as I can tell. Under what license is that material distributed under?

PR #3849, opened 2017-07-13
[DO NOT MERGE] fixes bug 1379662 - TelemetryEnvironment in super search fields

peterbe:
Yeah, we don't want to cache this at all. Reading from memcache (over the network) is much slower than simply reading it from a file. Even better, I realized that we can just load this in once during start-up as a class attribute. 

This means that `socorro.external.es.super_search_fields.SuperSearchFields.get` gets executed (reading the .json) as soon as Python imports anything in `crashstats.crashstats.models`. 

Seems to work.


peterbe:
By the way, I'm not sure why, but I ran the server with uwsgi (`--workers=4`) just to be absolutely sure it works in multi-processing as well.

peterbe:
UPDATE:

This PR still needs a review. But independent of that it doesn't make sense to land this until AFTER https://github.com/mozilla-services/socorro/pull/3854 has been reverted.

peterbe:
A hot mess. Will make a new one.

PR #3853, opened 2017-07-18
bug 1378836 - Make stackwalker detect likely-wrong crash addresses

luser:
There's also a cleanup commit here that removes the pipe dump output that we haven't used in ages.


peterbe:
@luser Do you have a Crash ID I can test this locally with?

luser:
Here's a Windows crash:
https://crash-stats.mozilla.com/report/index/bafda5b7-429b-48bb-b35a-a73740170718#tab-details

An OS X crash:
https://crash-stats.mozilla.com/report/index/d47985c5-9676-4807-86e3-52feb0170718

And a Linux crash:
https://crash-stats.mozilla.com/report/index/8d156b6f-c98c-4758-94d0-882b10170718#tab-details

One thing I just realized from doing this is that on Linux we're going to have *a lot* of false positives. On Windows we shouldn't, since the crash address we see there isn't actually a valid address. On OS X we shouldn't, since the exception code we get from the OS is very fine-grained, so we can say reliably that we have hit this situation. I wonder if instead of a boolean, I should make this more like an enum, like `"address_reliable": ("yes" | "maybe" | "no")` ?

peterbe:
Pardon my ignorance but those are all crashes where the Crash Address is something that immediately looks bogus. Is this not something we can attack in the webapp?
Or are these just lucky samples that were easy to find. 
(I reread the bug you filed and I don't feel confident I understand it)

Instead of enums, can we not use a numeric number of "confidence"? Just a gut reaction thought.

luser:
> Pardon my ignorance but those are all crashes where the Crash Address is something that immediately looks bogus. Is this not something we can attack in the webapp?

No, the crash address will legitimately be 0x0 for many crashes (null pointer dereferences). We actually might have enough information in the existing output of the stackwalker, I'd have to double-check.

> Or are these just lucky samples that were easy to find.

I did crafted SuperSearches to find the right kinds of crashes.

> (I reread the bug you filed and I don't feel confident I understand it)

Don't worry, it's an extremely low-level detail of one bit of x86-64 exception handling!

> Instead of enums, can we not use a numeric number of "confidence"? Just a gut reaction thought.

There are only really three scenarios here:
1. We know the crash address is wrong.
2. We think the crash address could be wrong.
3. We know the crash address is right.

peterbe:
@luser Had any more time to think about this one?

willkg:
This has been around for a while and needs some unspecified amount of work to land. In IRC, Ted said:

> <ted> willkg: we should just close that one for now, we can figure out a better plan
> <ted> i would like to have that functionality somehow--people get really confused by it :-(

I changed the related bug to be a P2 so it's on a radar somewhere. Someone will get to it.

In the meantime, I'm going to close this out.

PR #3857, opened 2017-07-19
Fixes bug 1381902 - update localstack to 0.4.1

willkg:


willkg:
This is pretty benign, but actually kind of important. 0.4.0 isn't available on dockerhub anymore, so this blocks anyone from creating a new Socorro docker environment. I'm following localstack pretty closely. I don't remember there being any exciting changes between 0.4.0 and 0.4.1. Plus I'm pretty sure tecken is using a more recent version now.

Anyhow, this should be safe and if not, it just affects developer environments.

PR #3858, opened 2017-07-20
fixes bug 1382508 - upgrade to raven 6.1.0

peterbe:
[I've only tested it in the webapp](https://bugzilla.mozilla.org/show_bug.cgi?id=1382508#c1) but it seems to work. I tried to *read* the [changelog](https://github.com/getsentry/raven-python/blob/master/CHANGES) but nothing scary stood out. 

Going to test this in crontabber whilst the PR brews on travis.


peterbe:
Also, seems to work from crontabber https://sentry.prod.mozaws.net/operations/socorro-stage/issues/626810/

PR #3859, opened 2017-07-20
Revert "fixes bug 1379665 - Telemetry tab on report index"

peterbe:
Reverts mozilla-services/socorro#3847


willkg:
Looks like this reverts two PRs. Do you really want to do that?

peterbe:
Github's UI for reverts are weird. There's only one commit it reverts but above it looks like there's too. 
Anyway, going to "fail forward" with some different fixes.

PR #3860, opened 2017-07-20
fixes bug 1379665 - Use JSON widget to display TelemetryEnvironment

peterbe:
CC @adngdb for possible feedback. 

Now it looks like this:
screen shot 2017-07-20 at 12 01 46 pm

And if the TelemetryEnvironment JSON string has been tampered with (I tested by manually messing with it in `views.py`):
screen shot 2017-07-20 at 12 28 02 pm

The PR became bigger that I'd hoped for because I had to rearrange where this jsonview jquery lib is located. Now it's in the "base" app so it can more easily be shared by "crashstats" and "documentation".


webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -78,7 +80,7 @@
                             <span>Extensions</span>
                         </a>
                     </li>
-                    {% if has_telemetry_environment %}
+                    {% if raw.TelemetryEnvironment %}
```
willkg: I don't think I'd do this. Otherwise the tab is sometimes there and sometimes not and it's weird navigationally-speaking. I think I'd always show it.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -799,9 +801,15 @@
                 </div>
                 <!-- /extensions -->
 
-                {% if has_telemetry_environment %}
+                {% if raw.TelemetryEnvironment %}
                 <div id="telemetryenvironment" class="ui-tabs-hide">
-                    {{ dict_as_html_tree(raw.TelemetryEnvironment) }}
+                     {# Remmber, raw.TelemetryEnvironment is a string.
+                        But jQuery till automatically convert it a map upon
+                        accessing it.
```
willkg: This sentence doesn't parse in my head. What do you mean here?

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/report.js
```diff
@@ -2,6 +2,35 @@
...
+                        $('#telemetryenvironment-json').JSONView(jsonData);
+                    } catch (ex) {
+                        console.warn(
+                            "The data in the #telemetryenvironment-json dataset is not valid JSON"
+                        );
+                        container.append(
+                            $('<p>Invalid JSON in TelemetryEnvironment</p>')
+                        );
+                    }
+                }
```
willkg: You could add an `else` here that spits out "No Telemetry Environment data available."

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
peterbe: There's a similar logic with "Correlations" and "Reprocess" already.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
peterbe: Fixed. Meant "will" not "till".

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/report.js
peterbe: Instead, I'm going to remove the if statement. It doesn't make sense to even have it. The test if there is telemetryenvironment data is done in the template already.

peterbe:
Rebased, nits addressed. Assuming you're happy with my nit-addresses, I'll merge this when travis smiles.

willkg:
Looks good to me. 👍

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
adngdb: And I hate both of those, personally. It's not a drama but I would do away with those conditional tabs too. I'd rather show a message that says there's no data than hide the tab. Consistency in navigation is better imho.

PR #3863, opened 2017-07-21
fixes bug 1382505 - remove FennecBetaError20150430 rule

willkg:
This rule kicks up a KeyError if the raw crash is missing the keys it checks. I
went to fix that, but then noticed there are no tests for this rule and it's to
fix a situation that was a problem over 2 years ago. I don't think this is
relevant anymore.

Given that, this removes the rule altogether.


willkg:
Thank you!

PR #3864, opened 2017-07-21
fixes bug 1383269 - UnicodeEncodeError on report index related to bug…

peterbe:
See bug comment for a way to reproduce it.

### DIFF
--- a/webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
+++ b/webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
@@ -13,6 +13,7 @@ from django.core.cache import cache
 from django.core.urlresolvers import reverse
 from django.template import engines
 from django.utils.safestring import mark_safe
+from django.utils.encoding import smart_str
 
 from crashstats import scrubber
 from crashstats.crashstats.utils import parse_isodate
@@ -244,8 +245,8 @@ def bugzilla_submit_url(report, bug_product):
         'product': bug_product,
         'op_sys': op_sys,
         'rep_platform': report['cpu_name'],
-        'cf_crash_signature': u'[@ {}]'.format(report['signature']),
-        'short_desc': u'Crash in {}'.format(report['signature']),
+        'cf_crash_signature': '[@ {}]'.format(smart_str(report['signature'])),
+        'short_desc': 'Crash in {}'.format(smart_str(report['signature'])),
         'comment': (
             'This bug was filed from the Socorro interface and is \n'
             'report bp-{}.\n'
--- a/webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
+++ b/webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
@@ -244,8 +244,8 @@ def bugzilla_submit_url(report, bug_product):
         'product': bug_product,
         'op_sys': op_sys,
         'rep_platform': report['cpu_name'],
-        'cf_crash_signature': '[@ {}]'.format(report['signature']),
-        'short_desc': 'Crash in {}'.format(report['signature']),
+        'cf_crash_signature': u'[@ {}]'.format(report['signature']),
+        'short_desc': u'Crash in {}'.format(report['signature']),
         'comment': (
             'This bug was filed from the Socorro interface and is \n'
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -244,8 +244,8 @@ def bugzilla_submit_url(report, bug_product):
         'product': bug_product,
         'op_sys': op_sys,
         'rep_platform': report['cpu_name'],
-        'cf_crash_signature': '[@ {}]'.format(report['signature']),
-        'short_desc': 'Crash in {}'.format(report['signature']),
+        'cf_crash_signature': u'[@ {}]'.format(report['signature']),
+        'short_desc': u'Crash in {}'.format(report['signature']),
```
willkg: I'd force `report['signature']` to be a string. That allows us to keep the machinery between Socorro and Bugzilla simple (it always uses strings) which is easier to test and less likely to break if anything in Bugzilla changed.

webapp-django/crashstats/crashstats/tests/test_jinja_helpers.py
```diff
@@ -232,6 +232,21 @@ def test_with_os_name_is_null(self):
         qs = self._extract_query_string(url)
         ok_('op_sys' not in qs)
 
+    def test_with_unicode_signatuer(self):
```
willkg: Nit: There's a typo in "signature" here.

webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
peterbe: What did you have in mind? To use `unicode(...)` (or something equivalent from django's utils lib for making unicode strings)?

willkg: I looked through the codebase and I'm kind of surprised we don't have something already.

I'm thinking something like Django's `smart_str` (https://docs.djangoproject.com/en/1.11/ref/utils/#django.utils.encoding.smart_str).

If you want, I can write it and add it to `socorro/lib/` and then you can use that. Would that help?

peterbe: Ah! I think I know what you mean now. Sorry, I think I get it now. 
You mean this scenario:
```python
>>> u'{}'.format('unicode sträng')
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 11: ordinal not in range(128)
```

What my fix did, was to fix this error:
```python
>>> '{}'.format(u'unicode sträng')
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe4' in position 11: ordinal not in range(128)
```

So it's two different problems. Do you mind if I solve both of them using this PR? Or do you prefer a fresh new bug plus PR?

willkg: I'm not sure what you mean by that. I think the resulting value should be a string--not a unicode. So you can solve this problem two ways:

```
'{}'.format(u'unicode sträng')
```

You can either turn the `'{}'` into a unicode so there's no encoding going on or you can convert the `u'unicode sträng'` into a string using something like `smart_str`.

I think we should be doing the second one since then we don't have to deal with changing how we're doing things with Bugzilla which isn't tested and who knows if it could change.

So I think the better change to make is something like this:

```
+        'cf_crash_signature': '[@ {}]'.format(smart_str(report['signature'])),
+        'short_desc': 'Crash in {}'.format(smart_str(report['signature'])),
```

Does that make sense? If you want, I can write a `smart_str` for Socorro and put it in `socorro/lib/`.

peterbe: Hmm.. I wonder why did it work when it was a unicode string?
`urllib.urlencode` doesn't support an encoding. 

I'll just make them all byte strings. I think there's a Django function for it somewhere.

peterbe: See updated commit. I tested it and clicked on the link and how the Summary is filled in correctly:
screen shot 2017-07-24 at 3 19 50 pm

PR #3865, opened 2017-07-24
1382506 json dump

willkg:
Minor changes to the code so that it always uses getitem notation and switched out `eq_` for assert.

After that, I fixed `json_dump` access to use `.get()` and default to `{}`.

That fixes the specific issue in the bug.

### DIFF
--- a/socorro/processor/general_transform_rules.py
+++ b/socorro/processor/general_transform_rules.py
@@ -25,7 +25,7 @@ class CPUInfoRule(Rule):
         cpu_name = ''
         cpu_info = ''
 
-        system_info = processed_crash.json_dump.get('system_info')
+        system_info = processed_crash.get('json_dump', {}).get('system_info')
         if system_info:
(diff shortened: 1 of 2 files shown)


willkg:
It might be easier to review the two commits separately.

socorro/processor/general_transform_rules.py
```diff
@@ -22,22 +22,25 @@ def version(self):
...
-                processed_crash.cpu_info = (
+            cpu_name = system_info.get('cpu_arch', '')
+
+            if 'cpu_info' in system_info and 'cpu_count' in system_info:
+                cpu_info = (
                     '%s | %s' % (
                         system_info['cpu_info'],
                         system_info['cpu_count']
                     )
                 )
```
peterbe: Feel free to switch to `str.format(...)` if you like.

willkg: I don't generally use `str.format`. It's super helpful in some specific cases, but `%s` is generally safer.

PR #3869, opened 2017-07-26
fixes bug 1373930 - Remove "aurora" from the correlations channels

peterbe:


peterbe:
As far as I can see `settings.CHANNELS` is only used by that view function for the signature report. It uses that to populate the `data-channels` attribute on `<div id="mainbody"`.

marco-c:
We can also remove `aurora` from https://github.com/mozilla-services/socorro/blob/38d00d541f800522610d46ccb0ea78a9516d2935/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js#L45 and https://github.com/mozilla-services/socorro/blob/13ad791bc2071359c94e75c2bbcbf096191a5200/webapp-django/crashstats/signature/views.py#L381.

PR #3870, opened 2017-07-28
fixes bug 1373930 - remove all correlations mentions of "aurora"

peterbe:


webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
```diff
@@ -42,10 +42,7 @@ window.correlations = (function () {
                     'date': totals['date'],
                 };
 
-                var channels = ['release', 'beta', 'aurora', 'nightly'];
-                if (product === 'Firefox') {
-                    channels.push('esr');
-                }
+                var channels = $('#mainbody').data('channels');
```
peterbe: So, before it used to *only* add `esr` to the list **if** the product was `Firefox`. Now `esr` is always in that list. But I don't care. It's just too much hassle to complicate things since I don't think we have any correlations in Telemetry for non-Firefox products.

webapp-django/crashstats/signature/views.py
```diff
@@ -378,8 +378,6 @@ def signature_correlations(request, params):
     elif 'version' in params and params['version']:
         if all('b' in version for version in params['version']):
             context['channel'] = 'beta'
-        elif all('a2' in version for version in params['version']):
-            context['channel'] = 'aurora'
         elif all('a1' in version for version in params['version']):
```
peterbe: The only way this could still be needed is if someone pulls up the Correlations tab (under Signature Report), and adds additional filtering on specifically an old "XX.0a2" version. E.g:
screen shot 2017-07-28 at 10 24 02 am
But mind you, any of those XX.0a2 versions have to be fully "manually" typed in since it's not an option from the dropdown suggestions that appear in that input field.

peterbe:
Basically, this PR is a response on @marco-c 's [comment](https://github.com/mozilla-services/socorro/pull/3869#issuecomment-318211634).

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
marco-c: There's Firefox for Android that doesn't have ESR.

peterbe: What does that *mean*?

marco-c: The `FennecAndroid` product has correlations, but doesn't have a `esr` channel.

PR #3871, opened 2017-07-28
bug 1383113 - convert signature rules to getitem notation

willkg:
* reduce some extra lines
* convert the things that were using attribute notation to getitem notation
* clean up data structure set up in the tests
* convert tests to use {} instead of DotDict for all non-config things

Cleaning up this code now will make the PR when I move signature stuff around for a cli a lot easier since there will be fewer things that are changing.

At some point, we might start using a lib for easier deep-item access in trees, but that's still on the drawing board and I want to move forward on cli signature generation now.

r?

### DIFF
--- a/socorro/processor/signature_utilities.py
+++ b/socorro/processor/signature_utilities.py
@@ -26,17 +26,10 @@ class SignatureTool(RequiredConfig):
     def __init__(self, config, quit_check_callback=None):
         self.config = config
         self.max_len = config.setdefault('signature_max_len', 255)
-        self.escape_single_quote = \
-            config.setdefault('signature_escape_single_quote', True)
+        self.escape_single_quote = config.setdefault('signature_escape_single_quote', True)
         self.quit_check_callback = quit_check_callback
 
-    def generate(
-        self,
-        source_list,
-        hang_type=0,
-        crashed_thread=None,
-        delimiter=' | '
-    ):
+    def generate(self, source_list, hang_type=0, crashed_thread=None, delimiter=' | '):
         signature, signature_notes = self._do_generate(
             source_list,
             hang_type,
@@ -51,13 +44,7 @@ class SignatureTool(RequiredConfig):
                                    'length')
         return signature, signature_notes
 
-    def _do_generate(
-        self,
-        source_list,
-        hang_type,
-        crashed_thread,
-        delimiter
-    ):
+    def _do_generate(self, source_list, hang_type, crashed_thread, delimiter):
         raise NotImplementedError
 
 
@@ -91,11 +78,7 @@ class CSignatureToolBase(SignatureTool):
         self.fixup_comma = re.compile(r',(?! )')
 
     @staticmethod
-    def _is_exception(
-        exception_list,
-        remaining_original_line,
-        line_up_to_current_position
-    ):
+    def _is_exception(exception_list, remaining_original_line, line_up_to_current_position):
         for an_exception in exception_list:
             if remaining_original_line.startswith(an_exception):
                 return True
@@ -110,10 +93,14 @@ class CSignatureToolBase(SignatureTool):
         replacement_open_string,
         close_string,
         replacement_close_string,
-        exception_substring_list=(),  # list of exceptions that shouldn't collapse
+        exception_substring_list=[],
     ):
         """this method takes a string representing a C/C++ function signature
-        and replaces anything between to possibly nested delimiters"""
+        and replaces anything between to possibly nested delimiters
+
+        :arg list exception_substring_list: list of exceptions that shouldn't collapse
+
+        """
         target_counter = 0
         collapsed_list = []
         exception_mode = False
@@ -194,7 +181,7 @@ class CSignatureToolBase(SignatureTool):
             # Ensure a space after commas
             function = self.fixup_comma.sub(', ', function)
             return function
-        #if source is not None and source_line is not None:
+        # if source is not None and source_line is not None:
         if file and line:
             filename = file.rstrip('/\\')
             if '\\' in filename:
@@ -208,11 +195,7 @@ class CSignatureToolBase(SignatureTool):
             module = ''  # might have been None
         return '%s@%s' % (module, module_offset)
 
-    def _do_generate(self,
-                     source_list,
-                     hang_type,
-                     crashed_thread,
-                     delimiter=' | '):
+    def _do_generate(self, source_list, hang_type, crashed_thread, delimiter=' | '):
         """
         each element of signatureList names a frame in the crash stack; and is:
           - a prefix of a relevant frame: Append this element to the signature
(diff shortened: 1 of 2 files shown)


socorro/unittest/processor/test_signature_utilities.py
```diff
@@ -1566,116 +1546,108 @@ def test_predicate_no_match(self):
...
         fake_processor = create_basic_fake_processor()
         rule = StackwalkerErrorSignatureRule(fake_processor.config)
-        action_result = rule.action(rc, rd, pc, fake_processor)
+        action_result = rule.action({}, {}, processed_crash, fake_processor)
         ok_(action_result)
-        ok_(
-            pc.signature,
-            "EMPTY: like my soul; catastrophic stackwalker failure"
-        )
+        eq_(processed_crash['signature'], 'EMPTY: like my soul; catastrophic stackwalker failure')
```
willkg: This was a bug. Instead of comparing two things to see if they were equal, it ended up checking to see if the first one was truthy with the second argument being what gets returned if it's not. I switched it from `ok_` to `eq_`.

socorro/processor/signature_utilities.py
```diff
@@ -462,42 +434,42 @@ def _create_frame_list(
...
         try:
             if processed_crash.get('hang_type', None) == 1:
-                # force the signature to come from thread 0
+                # Force the signature to come from thread 0
                 signature_list = self._create_frame_list(
-                    processed_crash.json_dump["threads"][0],
-                    processed_crash.json_dump['system_info']['os'] in
-                    "Windows NT"
+                    processed_crash['json_dump']['threads'][0],
+                    processed_crash['json_dump']['system_info']['os'] == 'Windows NT'
```
willkg: I'm pretty sure this was a bug. Originally, it was checking to see if `processed_crash.json_dump['system_info']['os']` value was in the sequence `Windows NT`. That's probably not what we want. Maybe it was a list of multiple items before? Maybe the author meant `is` (though, that's wrong, too)`?

I changed it to `== 'Windows NT'`.

Anyhow, so before it would always pass in `False` to `make_modules_lower_case`. Now it might pass `True` if it's `Windows NT`. We shouldn't see that often, so I think this change even if it's wrong now won't have much effect on things.

willkg:
GitHub does a good job diffing this. You can see the individual characters I added for most of the lines which should make skimming them easier.

socorro/processor/signature_utilities.py
lonnen: Difficult to tell. This was introduced in a 1k+ LoC patch. I believe this is an improvement.

socorro/processor/signature_utilities.py
```diff
@@ -564,19 +538,17 @@ def version(self):
         return '1.0'
 
     def _predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
-        return 'AbortMessage' in raw_crash and raw_crash.AbortMessage
+        return bool(raw_crash.get('AbortMessage'))
```
lonnen: curious. I rarely see `bool` used, even where we're clearly testing for a boolean. Should it be the convention going forward for `_predicate`s?

lonnen: apparently it's common in this module already. I retract my assertion.

willkg: The file had other instances of wrapping things in `bool` so I just went with that figuring existing convention is good. But you're right--converting to a bool isn't needed.

I don't have a preference. If someone has a preference, I'm happy to do that.

lonnen: I think matching the file context is good enough.

lonnen:
r+ 👍

socorro/processor/signature_utilities.py
```diff
@@ -462,42 +434,42 @@ def _create_frame_list(
         return frame_signatures_list
 
     def _get_crashing_thread(self, processed_crash):
-        return processed_crash.json_dump['crash_info']['crashing_thread']
+        return processed_crash['json_dump']['crash_info']['crashing_thread']
```
willkg: In `treevert`, this would be:

```
tree_get(processed_crash, 'json_dump.crash_info.crashing_thread', default=None)
```

socorro/processor/signature_utilities.py
```diff
@@ -462,42 +434,42 @@ def _create_frame_list(
...
 
     def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
-        if 'JavaStackTrace' in raw_crash and raw_crash.JavaStackTrace:
-            # generate a Java signature
+        # If this is a Java crash, then generate a Java signature
+        if 'JavaStackTrace' in raw_crash and raw_crash['JavaStackTrace']:
```
willkg: We could convert this to:

```
if raw_crash.get('JavaStackTrace'):
```

socorro/processor/signature_utilities.py
willkg: In `treevert`, this would be:

```
tree_get(processed_crash, 'json_dump.system_info.os', default=None)
```

socorro/processor/signature_utilities.py
```diff
@@ -462,42 +434,42 @@ def _create_frame_list(
...
+
         try:
             if processed_crash.get('hang_type', None) == 1:
-                # force the signature to come from thread 0
+                # Force the signature to come from thread 0
                 signature_list = self._create_frame_list(
-                    processed_crash.json_dump["threads"][0],
-                    processed_crash.json_dump['system_info']['os'] in
-                    "Windows NT"
+                    processed_crash['json_dump']['threads'][0],
```
willkg: In `treevert`, this would be:

```
tree_get(processed_crash, 'json_dump.threads.[0]', default=None)
```

willkg:
I tossed some "if we were using `treevert`, then this would be ..." comments in there just to see if that tastes good to anyone. `treevert` is still on the drawing board and we wouldn't make `treevert` changes in this PR (and `treevert` may not even be the final name), but figured there are some good examples of possible usage in here.

willkg:
@lonnen Thank you so much for the expedient review! I really appreciate it.

socorro/processor/signature_utilities.py
peterbe: Why not `raw_crash.get('JavaStackTrace')` here?

peterbe:
Awesomeness! Not that it really matters but I think the use of `bool` feels slightly unpythonic. If there was a log that wrote down the result of all the `_predicate()` calls, *it* could do a conversion to booleans. But even that would best be `'yes' if each_thing._predicate(*args) else 'nope'`.

peterbe:
Perhaps the reason you used `bool()` was because of existing code used it. So you wanted consistency.

willkg:
> Perhaps the reason you used bool() was because of existing code used it. So you wanted consistency.

I was trying to reduce unobvious changes and keeping it to how the surrounding code looked. Whether we call `bool` or not doesn't impact anything. The conversation about that is here:

https://github.com/mozilla-services/socorro/pull/3871#pullrequestreview-53068550

PR #3873, opened 2017-07-31
fixes bug 1385957 - fix to safely get json_dump

willkg:


socorro/processor/general_transform_rules.py
```diff
@@ -53,7 +53,7 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
         processed_crash.os_name = ''
         processed_crash.os_version = ''
 
-        system_info = processed_crash.json_dump.get('system_info')
+        system_info = processed_crash.get('json_dump').get('system_info')
```
peterbe: Missing the `, {}` default. 
If `processed_crash` is `{'foo': 'bar'}` this line would raise an AttributeError since `None` doesn't have a `get` method.

willkg: Oh, whoops--you're right. My bad.

willkg:
Pretty sure the test failure is for realz. Looking into it now.

PR #3875, opened 2017-07-31
fixes bug 1385962 - Handle AsyncShutdownTimeout string conditions

willkg:
If the condition is a dict with a "name" key, then this worked fine. However,
Socorro would occasionally see conditions that were just strings.

This fixes the rule behavior to work with strings as well.

### DIFF
--- a/socorro/processor/signature_utilities.py
+++ b/socorro/processor/signature_utilities.py
@@ -662,12 +662,22 @@ class SignatureShutdownTimeout(Rule):
     def _predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
         return bool(raw_crash.get('AsyncShutdownTimeout'))
 
+    def get_name(self, c):
+        """Return best name for an AsyncShutdownTimeout notation condition"""
+        # The AsyncShutdownTimeout notation condition can either be a string that looks like a
+        # "name" or a dict with a "name" in it.
+        #
+        # This handles both variations.
+        if isinstance(c, dict):
+            return c.get('name')
+        return c
+
     def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
         parts = ['AsyncShutdownTimeout']
         try:
             shutdown_data = ujson.loads(raw_crash['AsyncShutdownTimeout'])
             parts.append(shutdown_data['phase'])
-            conditions = [c['name'] for c in shutdown_data['conditions']]
+            conditions = [self.get_name(c) for c in shutdown_data['conditions']]
             if conditions:
                 conditions.sort()
                 parts.append(','.join(conditions))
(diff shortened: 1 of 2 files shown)


socorro/processor/signature_utilities.py
```diff
@@ -662,12 +662,22 @@ def version(self):
...
 
+    def get_name(self, c):
+        """Return best name for an AsyncShutdownTimeout notation condition"""
+        # The AsyncShutdownTimeout notation condition can either be a string that looks like a
+        # "name" or a dict with a "name" in it.
+        #
+        # This handles both variations.
+        if isinstance(c, dict):
+            return c.get('name')
+        return c
```
peterbe: Nit: This should probably be an inline function or a static method. Especially since the parameter is just called `c`.

socorro/processor/signature_utilities.py
```diff
@@ -662,12 +662,22 @@ def version(self):
     def _predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
         return bool(raw_crash.get('AsyncShutdownTimeout'))
 
+    def get_name(self, c):
+        """Return best name for an AsyncShutdownTimeout notation condition"""
+        # The AsyncShutdownTimeout notation condition can either be a string that looks like a
+        # "name" or a dict with a "name" in it.
+        #
+        # This handles both variations.
+        if isinstance(c, dict):
```
peterbe: Perhaps it's paranoid but instead of dict, we should maybe use `collections.Mapping` so it works for any thing that `__getitem__`.

socorro/processor/signature_utilities.py
```diff
@@ -662,12 +662,22 @@ def version(self):
...
+            return c.get('name')
+        return c
+
     def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
         parts = ['AsyncShutdownTimeout']
         try:
             shutdown_data = ujson.loads(raw_crash['AsyncShutdownTimeout'])
             parts.append(shutdown_data['phase'])
-            conditions = [c['name'] for c in shutdown_data['conditions']]
+            conditions = [self.get_name(c) for c in shutdown_data['conditions']]
```
peterbe: Or much simpler;
```python
conditions = [
    c['name'] if isinstance(c, dict) else c
    for c in shutdown_data['conditions']
]
```

socorro/processor/signature_utilities.py
willkg: c is from `json.loads`, so it's either going to be a dict or not. We don't need to handle other cases.

socorro/processor/signature_utilities.py
willkg: I like this. I'll change it.

willkg:
@peterbe ^^^ I switched to your suggestion. Tests pass. Can you look at this again?

PR #3877, opened 2017-08-01
fixes bug 1386409 - remove extra processors

willkg:
This removes extra processors that we're not using and haven't used in some
time.


willkg:
This is a big PR. I think the way we want to review this is the following:

1. Double-check the new `processor_class` configuration default. We explicitly set that in -stage and -prod, so even if it's not right, that's ok. Still, it'd be nice if it was right.
2. Let Travis and Circle run.
3. Merge and then make sure -stage is ok, but I'm pretty sure it'll be fine.

Generally, these files don't touch much else, so removing them is doesn't affect much.

If you have other ideas of things we can check/verify/whatever, go for it.

willkg:
I thought about also collapsing Processor2015 and MozillaProcessorAlgorithm2015 and possibly changing that `processor_class` option.

I didn't do that for two reasons:

1. That's slightly less trivial of a change because configman does things and we'd probably have to update tests. Though I suppose if I got it wrong, it'd be obvious.
2. Part of me wonders if we'd ever want to switch the `processor_class` for testing things. I suppose that if I haven't done it by now, then I'm probably not going to do it.

That first reason suggests I should do the change in a separate PR that doesn't have 6000 lines of changes in it. I'll land this and think about doing that in a follow-up PR.

willkg:
Thank you!

PR #3879, opened 2017-08-02
fixes bug 1386237 - Display "junk" field keys better in report index

peterbe:
Before:
screen shot 2017-08-02 at 10 26 20 am
After:
screen shot 2017-08-02 at 10 26 29 am

Other keys that contain other crazy unicode characters is probably best to just leave as is. 

Note; this doesn't so much to do with "unicode junk". It just makes it so that any key that becomes visually longer than 250 pixels gets truncated.

### DIFF
--- a/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
+++ b/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
@@ -697,7 +697,13 @@
                         <tbody>
                             {% for key in raw_keys %}
                             <tr title="{{ fields_desc.get(make_raw_crash_key(key), empty_desc) }}">
-                                <th scope="row">{{ key }}</th>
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -697,7 +697,13 @@
                         <tbody>
                             {% for key in raw_keys %}
                             <tr title="{{ fields_desc.get(make_raw_crash_key(key), empty_desc) }}">
-                                <th scope="row">{{ key }}</th>
+                                <th scope="row">
+                                    {% if key %}
+                                        {{ key }}
+                                    {% else %}
+                                        <i>empty key</i>
+                                    {% endif %}
```
willkg: I like this. Good idea!

PR #3881, opened 2017-08-02
fixes bug 1366482 - sort signature comments by default by date

peterbe:
(Not sure who should review this so pinning both of y'all)

This patch is a compromise. 
The signature report has two tabs that uses tables with pagination and stuff. The crazy thing is that if you go to the Reports tab and click to sort by "Build ID" (for example) and then click the "Comments" tab the query string in the URL will still be `&sort=build_id`. So it'd potentially be a clash. 

The "right way" would be to replicate all the sorting functionality on the Reports tab into the Comments tab but perhaps change the parameters from `page` to `comments-page` and from `sort` to `comments-sort`. 

I got started doing exactly that but quickly got stuck in a deep rabbit hole with regards to how this `SignatureReport` JavaScript object works. It just got messier and messier. So I bailed and went for something much simpler. Hardcode the sort to `date` in descending order. The pagination now works but it's still kinda crazy that if you click to go to page 24 on the Reports tab and then click the Comments tab, there might not be a page 24. 

I'd be happy to build something more advanced some day but I'm not sure the return of investment is worth it at this point. The tab works OKish and we don't get a ton of feedback that people require more advanced filtering and sorting on comments. 

Fair enough?


adngdb:
So that means any sort of filtering that users might do on the table will be useless. If you're going this way, which I do not mind, I think you should disable ``tablesort`` on the comments table as well.

peterbe:
>  disable tablesort on the comments table 

I can do that. It's, like I said, a compromise.

peterbe:
Still fishing for a r+ here. @willkg 
I updated it so that the table on the Comments tab is not client-side sortable. I means all comments are sorted server-side by date and the user can't change that. Arguably, if he wants to he can use supersearch. My argument is that it's not worth the effort of making this table fancy sortable.

PR #3882, opened 2017-08-02
fixes bug 1377758 - no thunderbird aurora version featured

peterbe:


willkg:
This seems fine to me. r+

peterbe:
Note-to-self; Going to wait for Wayne to weigh in.

peterbe:
This is going to require an upgrade of the admin nodes once this goes to production. So we should see the effect late next week. I.e. Aug 10.

willkg:
@peterbe Next week is a release week, so we're not doing a -prod deployment on August 10th.

PR #3883, opened 2017-08-02
Clean up create_basic_fake_processor definitions and usage

willkg:
We had this function in three different places and one of those instances was
slightly different than the other two. Seems like it helps to have a centralized
"fake processor" configuration that is used everywhere.

This centralizes that code and updates the other files to use it.

### DIFF
--- a/socorro/unittest/processor/__init__.py
+++ b/socorro/unittest/processor/__init__.py
@@ -1,0 +1,25 @@
+import mock
+
+from socorro.lib.util import DotDict
+from socorro.processor.signature_utilities import CSignatureTool
+
+
+csig_config = DotDict()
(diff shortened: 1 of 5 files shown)


willkg:
I thought about writing up a bug for this, but it's pretty minor cleanup and didn't seem to warrant a bug. I extracted this from my bigger "redo the signature generation stuff" PR figuring it's easier to review on its own. I can create a bug retroactively if needed.

Travis and Circle will adequately test this. It's worth eye-balling.

The code in question was essentially copy-and-pasted across three files. One of the instances was missing this line:

```
csig_config.collapse_arguments = True
```

Seems like it's an inconsequential difference.

willkg:
Thank you!

PR #3888, opened 2017-08-04
Fix bug 1387362 - add 'std::collections::hash::map::.*' to prefix ski…

calixteman:
…plist


willkg:
This looks ok to me. I'm going to merge it and then check it on -stage.

PR #3889, opened 2017-08-04
bug 1387096 - add docs for creating a user and setting up signin

willkg:


docs/components/webapp.rst
```diff
@@ -43,6 +43,53 @@ Then you can start and stop the webapp, adjust files, and debug.
...
+4. Open ``webapp-django/.env`` in an editor and add these lines::
+
+       OAUTH2_CLIENT_ID=<client id>
+       OAUTH2_CLIENT_SECRET=<secret>
+
+   where ``<client id>`` is the oauth client id and ``<secret>`` is the oauth
+   client secret that you got from step 3
+
+
+After that, Google signin should work.
```
willkg: Are these instructions for setting up the oauth stuff right? Are there any other steps that are needed?

docs/gettingstarted.rst
```diff
@@ -49,7 +49,12 @@ Quickstart
       Pull requests welcome!
 
 
-At that point, you should have a functional Socorro development environment.
+At that point, you should have a basic functional Socorro development
+environment.
+
+See :ref:`webapp-chapter` for additional setup and running the webapp.
+
+See :ref:`processor-chapter` for additional setup and running the processor.
```
willkg: If someone is reading Getting Started, they do some setup and then this connects them to additional bits they'll need to do for the processor and/or the webapp depending on what they want to do now.

docs/components/webapp.rst
```diff
@@ -43,6 +43,53 @@ Then you can start and stop the webapp, adjust files, and debug.
...
+  $ docker-compose run webapp python manage.py createsuperuser
+
+
+Then follow the prompts. The username and password don't matter at all, but the
+email address needs to match what you plan to use for Google auth.
+
+You can do this as many times as you like.
+
+
+Setting up the webapp for Google signin
```
peterbe: Nit: It's called Google Sign-In as that's a Google product name. We'd be more consistent if we call it this.

docs/components/webapp.rst
```diff
@@ -43,6 +43,53 @@ Then you can start and stop the webapp, adjust files, and debug.
...
+
+Creating a superuser
+--------------------
+
+If you want to do anything in the webapp admin, you'll need to create a
+superuser.
+
+Run this::
+
+  $ docker-compose run webapp python manage.py createsuperuser
```
peterbe: We have our own command and it's called [`makesuperuser`](https://github.com/mozilla-services/socorro/blob/master/webapp-django/crashstats/authentication/management/commands/makesuperuser.py)
It doesn't bother with username and password and works for both existing and new users.

docs/components/webapp.rst
```diff
@@ -43,6 +43,53 @@ Then you can start and stop the webapp, adjust files, and debug.
...
+Then follow the prompts. The username and password don't matter at all, but the
+email address needs to match what you plan to use for Google auth.
+
+You can do this as many times as you like.
+
+
+Setting up the webapp for Google signin
+---------------------------------------
+
+In order to authenticate in the webapp using the "Google signin" button, you
```
peterbe: ditto on the name

docs/components/webapp.rst
willkg: It doesn't. If the user doesn't exist, it errors out:

https://github.com/mozilla-services/socorro/blob/master/webapp-django/crashstats/authentication/management/commands/makesuperuser.py#L25

We talked about this earlier. You said we should change to something like what Tecken does--that's a good idea. I figured it's easier for now to just do a docs change.

willkg:
@peterbe ^^^ That fixes "signin" to "Sign-In" which addresses the nit you had. The other thing I think we should push off to a future PR.

These are just docs changes, so we don't need to wait for Travis to pass.

Given that the only issues to the process for acquiring oauth credentials were a minor nit which is fixed, I'm going to land this.

docs/components/webapp.rst
peterbe: Ah! I didn't remember that. Good catch. 
However, it wouldn't be hard to fix. 

So, I'll make a PR that changes the docs to use `makesuperuser` and improve `makesuperuser` to create the user if need be.

peterbe: https://bugzilla.mozilla.org/show_bug.cgi?id=1387574

willkg: That's awesome--thank you!

PR #3890, opened 2017-08-04
[DO NOT MERGE] fixes bug 1387574 - Improve makesuperuser to create users if need be

peterbe:
This does NOT work yet. 

The management command works. I've tested it outside of `docker-compose`.
But I get this:

```
▶ docker-compose run webapp python manage.py makesuperuser pbengtsson@mozilla.com
Starting socorro_rabbitmq_1 ... done
Starting socorro_localstack-s3_1 ...

Starting socorro_postgresql_1 ... doneone
python: can't open file 'manage.py': [Errno 2] No such file or directory
```


webapp-django/crashstats/authentication/management/commands/makesuperuser.py
```diff
@@ -1,11 +1,19 @@
-import sys
+from __future__ import print_function
+
+try:
+    _input = raw_input
+except NameError:
+    # You're on Python >=3
+    _input = input
```
willkg: Can we use `six` for this?

willkg:
I did a quick glance and this looks good. I need to finish this other thing, but then I'll double-check this.

webapp-django/crashstats/authentication/management/commands/makesuperuser.py
peterbe: Could. Or leave it as is until we switch everything up to python 3 ;)

willkg: We've got six in the requirements. I'd rather we used that since that's what it's for.

willkg: In six, this is:

```
from six.moves import input
```

docs/components/webapp.rst
```diff
@@ -54,12 +54,9 @@ superuser.
 
 Run this::
 
-  $ docker-compose run webapp python manage.py createsuperuser
+  $ docker-compose run webapp python manage.py makesuperuser email@example.com
```
willkg: This is good, but there's a bug in that when you run the webapp container, the cwd is `/app/` and it needs to be `/app/webapp-django/`. So this line should be:

```
$ docker-compose run webapp python webapp-django/manage.py makesuperuser willkg@mozilla.com
```

Can you fix that in this PR? If you don't want to, that's totally fine--I'll do a follow-up PR.

peterbe: Cool! I was able to use that and it actually worked. 
Also, rebased this up. Will land this PR once travis is happy.

PR #3891, opened 2017-08-07
bug 1361764 - convert signature tests to pytest format

willkg:
This nixes ok_, eq_, and friends for pytest equivalents.

### DIFF
--- a/socorro/unittest/processor/test_signature_utilities.py
+++ b/socorro/unittest/processor/test_signature_utilities.py
@@ -6,7 +6,6 @@
 import copy
 import json
 import mock
-from nose.tools import eq_, ok_
 import re
 
 from configman.dotdict import DotDict as CDotDict
@@ -33,17 +32,7 @@ from socorro.unittest.processor import create_basic_fake_processor
 from socorro.unittest.testbase import TestCase
 
 
-class BaseTestClass(TestCase):
-
-    def assert_equal_with_nicer_output(self, expected, received):
-        eq_(
-            expected,
-            received,
-            'expected:\n%s\nbut got:\n%s' % (expected, received)
-        )
-
-
-class TestCSignatureTool(BaseTestClass):
+class TestCSignatureTool(TestCase):
 
     @staticmethod
     def setup_config_c_sig_tool(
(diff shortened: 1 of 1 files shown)


willkg:
This is pretty much a straight conversion of `ok_` and `eq_` to their assert equivalents. I also removed `assert_equal_with_nicer_output` because pytest gives decent output for failed assertions, so I don't think we need this anymore.

socorro/unittest/processor/test_signature_utilities.py
```diff
@@ -1091,20 +1050,20 @@ def test_action_1(self):
         }
 
         # the call to be tested
-        ok_(sgr._action(raw_crash, {}, processed_crash, processor_meta))
+        assert sgr._action(raw_crash, {}, processed_crash, processor_meta) is True
```
willkg: For predicate and action returns, I think we really want to verify the actual return value and not whether or not it's truthy, so I made it a test against the actual `True` or `False` value.

That will protect us against accidentally not returning something which in Python will return `None` which is falsey, but possibly a bug.

socorro/unittest/processor/test_signature_utilities.py
```diff
@@ -1117,15 +1076,15 @@ def test_action_2(self):
...
 
-        eq_(
-            processed_crash['signature'],
+        assert (
+            processed_crash['signature'] ==
```
lonnen: This seems too subtle. I think it's easy to lose the comparison operator. Above (line `1005` post diff) we avoid direct comparisons to multiline literals. Consider similar here and below.

lonnen:
I did not mean to merge this, sorry. I intended instead to approve the change. I immediately reverted but this PR is now closed.

socorro/unittest/processor/test_signature_utilities.py
willkg: Depending on your editor and how it does syntax highlighting this might be less subtle. I've been doing this for a while so it clicks for me, but using an intermediary variable makes sense and readability and it's stepsisge, maintainability, are important. I'll do that going forward.

PR #3893, opened 2017-08-08
fixes bug 1388333 - UnicodeEncodeError in signature report

peterbe:
Go to any signature report and add `&accessibility=föö` or something. Or just look at the new unit test :)

By the way, that whole `to_python` didn't have any test coverage at all before.

### DIFF
--- a/webapp-django/crashstats/supersearch/form_fields.py
+++ b/webapp-django/crashstats/supersearch/form_fields.py
@@ -4,6 +4,7 @@ import isodate
 
 from django import forms
 from django.utils.timezone import utc
+from django.utils.encoding import smart_str
 
 from crashstats.crashstats.utils import parse_isodate
 
@@ -180,6 +181,7 @@ class BooleanField(forms.CharField):
         """
         if value is None:
             return None
-        if str(value).lower() in ('__true__', 'true', 't', '1', 'y', 'yes'):
+        truthy_strings = ('__true__', 'true', 't', '1', 'y', 'yes')
+        if smart_str(value).lower() in truthy_strings:
             return '__true__'
         return '!__true__'
--- a/webapp-django/crashstats/supersearch/form_fields.py
+++ b/webapp-django/crashstats/supersearch/form_fields.py
@@ -172,6 +172,9 @@ class StringField(MultipleValueField):
 
 
 class BooleanField(forms.CharField):
+
+    truthy_strings = ('__true__', 'true', 't', '1', 'y', 'yes')
+
     def to_python(self, value):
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/supersearch/form_fields.py
```diff
@@ -180,6 +181,7 @@ def to_python(self, value):
         """
         if value is None:
             return None
-        if str(value).lower() in ('__true__', 'true', 't', '1', 'y', 'yes'):
+        truthy_strings = ('__true__', 'true', 't', '1', 'y', 'yes')
```
willkg: We could pull this into a module level "constant".

webapp-django/crashstats/supersearch/tests/test_form_fields.py
```diff
@@ -154,3 +156,32 @@ def test_several_fields_illogically_datetimefield(self):
...
+        cleaned_value = field.clean(None)
+        eq_(cleaned_value, None)
+
+        # The list of known truthy strings
+        for truthy in ('__TRUE__', 'true', 't', '1', 'Y', 'Yes'):
+            cleaned_value = field.clean(truthy)
+            eq_(cleaned_value, '__true__')
+
+        # Not truthy strings
+        for truthy in ('__False__', 'False', 'f', '0', 'n', 'No'):
```
willkg: I don't think I'd call this variable `truthy` but rather `data` or `text` or something. It's a little confusing.

Also, I'd check for empty string and maybe `None` just so we know what happens with those in case this gets used elsewhere or copied or something.

webapp-django/crashstats/supersearch/form_fields.py
peterbe: Yes and no. There's no reuse of it anywhere so the gain of moving it 180ish lines away from where it's being used is only making the code slightly harder to read at the moment. 
Right now, I think the code is pretty easy to read.

webapp-django/crashstats/supersearch/tests/test_form_fields.py
peterbe: `None` is a special case and it's related to the fact that you can make the `BooleanField` *not* required. 
The name "truthy"  is poor. Will change. 

Also, this list of "falsy" strings is silly because it's never actually thing. I'll refactor it.

peterbe:
@willkg I took a second look and made it slightly nicer. Now the list of strings that are considered "truthy" is a class attribute on the `BooleanField`. That way I can pull it into the test.

willkg:
Looks good to me. r+!

PR #3894, opened 2017-08-08
bug 1388130 - copy -prod crontabber jobs to DEFAULT_JOBS

willkg:
This syncs the crontabber.jobs value that we have in -prod with DEFAULT_JOBS
enabling us to run all the jobs we're running in -prod in local dev
environments.

### DIFF
--- a/socorro/cron/crontabber_app.py
+++ b/socorro/cron/crontabber_app.py
@@ -6,24 +6,30 @@
 from crontabber.app import CronTabberBase
 
 
+# NOTE(willkg): This is what we have in -prod
 DEFAULT_JOBS = '''
-  socorro.cron.jobs.weekly_reports_partitions.WeeklyReportsPartitionsCronApp|7d
-  socorro.cron.jobs.matviews.ProductVersionsCronApp|1d|05:00
-  socorro.cron.jobs.matviews.SignaturesCronApp|1d|05:00
-  socorro.cron.jobs.matviews.RawUpdateChannelCronApp|1d|05:00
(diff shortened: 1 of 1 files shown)


socorro/cron/crontabber_app.py
```diff
@@ -6,24 +6,30 @@
...
+socorro.cron.jobs.ftpscraper.FTPScraperCronApp|1h
+socorro.cron.jobs.elasticsearch_cleanup.ElasticsearchCleanupCronApp|7d
+socorro.cron.jobs.drop_old_partitions.DropOldPartitionsCronApp|7d
+socorro.cron.jobs.truncate_partitions.TruncatePartitionsCronApp|7d
+socorro.cron.jobs.clean_raw_adi_logs.CleanRawADILogsCronApp|1d
+socorro.cron.jobs.clean_raw_adi.CleanRawADICronApp|1d
+socorro.cron.jobs.clean_missing_symbols.CleanMissingSymbolsCronApp|1d
+socorro.cron.jobs.missingsymbols.MissingSymbolsCronApp|1d
+socorro.cron.jobs.featured_versions_automatic.FeaturedVersionsAutomaticCronApp|1h
+socorro.cron.jobs.upload_crash_report_json_schema.UploadCrashReportJSONSchemaCronApp|1h
```
willkg: This is a copy of the -prod section from: https://bugzilla.mozilla.org/show_bug.cgi?id=1388130#c1

The only difference is that I removed commas from the end of the lines because it didn't have them before.

willkg:
Why do this now? We need to fix it to get the -dev environment working.

PR #3895, opened 2017-08-08
fixes bug 1387597 - Fix transaction_executor to not log benign errors

peterbe:
I appreciate this feels like a scary mess but the change is quite simple. 
Look at the two uploaded log files about ALL SQL that happens before and after this PR. 
Basically, when we remove this `NothingUsefulHappened` stuff, instead of ROLLBACK after each `SELECT signature FROM bug_associations WHERE bug_id = ...` now it does a  COMMIT instead. That's all. 

The good outcome of this is that we don't trigger [this logging](https://github.com/mozilla-services/socorro/blob/7681a91c1211e294c0d37de2011c4ba8a11a3e0c/socorro/database/transaction_executor.py#L52-L56) anymore.

### DIFF
--- a/socorro/cron/jobs/bugzilla.py
+++ b/socorro/cron/jobs/bugzilla.py
@@ -3,14 +3,17 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import datetime
+import sys
 
 import requests
 from dateutil import tz
 from configman import Namespace
+from configman.converters import class_converter
 from crontabber.base import BaseCronApp
 from crontabber.mixins import with_postgres_transactions
 
 from socorro.lib.datetimeutil import utc_now
+from socorro.app.socorro_app import App, main
 from socorro.external.postgresql.dbapi2_util import (
     execute_query_fetchall,
     execute_no_results,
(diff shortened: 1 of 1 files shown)


socorro/cron/jobs/bugzilla.py
```diff
@@ -186,3 +179,52 @@ def _iterator(self, from_date):
                 int(report['id']),
                 find_signatures(report.get('cf_crash_signature', ''))
             )
+
+
+class BugzillaCronAppDryRunner(App):  # pragma: no cover
```
willkg: It looks like this is doing actual db stuff and isn't a dry run. Is that right? If so, then maybe we should rename it.

willkg: I tested running bugzilla-associations via crontabber and via this class and I don't think I'll ever use this class. It's much easier to run crontabber and specify the job in the dev environment which picks up configuration from the environment than it is to run this class with the additional arguments.

```
$ socorro/cron/crontabber_app.py --job=bugzilla-associations
```

vs.

```
$  python socorro/cron/jobs/bugzilla.py --database.user=postgres --database.password=aPassword --database.dbname=breakpad
```

peterbe: Right! Good point. The reason I started writing this class was to be able to do a bit more detailed "manual" logging of exactly what SQL was being executed. Then I fell into a deep rabbit hole of trying to get mock to work. I eventually failed, took a step back to using a real postgres table. What I didn't realize was this very point you're making. That it's easier to just run it with crontabber_app.py

Will update.

peterbe:
@willkg I've removed that dry run app that wasn't a dry run and added a doc string about how to use `crontabber_app.py` to run it locally. 

If travis is happy I'm going to land it.

willkg:
Sounds perfect!

PR #3896, opened 2017-08-08
bug 1361764 pytest sigtests 2

willkg:
This is identical to the previous PR, except that it has an additional commit that breaks up multi-line asserts as @lonnen suggested.


willkg:
@lonnen approved the original PR with one minor comment. But then he inadvertently merged it and reverted the merge commit breaking several of the beams to the Dark Tower.

Since then, I created a new PR with the original work plus a commit that addresses the comment. In order to save the universes and prevent a universe reset in which all the work we did this quarter is for naught, I'm going to merge this now.

REMEMBER THE FACE OF YOUR FATHER! SPOON!

PR #3899, opened 2017-08-08
fixes bug 1388474 - Make test_blocking_start_with_quit_on_empty not f…

peterbe:
I honestly don't get it but I'm honestly not that bothered either. Just avoiding `Mock` seems to mean something. 

I ran `watch -e -n 4 pytest socorro/unittest/lib/test_threaded_task_manager.py` for about 1-2 minutes till it failed. 
Then, using the code in this PR, I ran it again and I never failed in the 10+ minutes I stepped away from the computer. 

Also, my git hook linter yelled at me for other crap in that file so I took the opportunity to lazily kill two birds with one commit.

### DIFF
--- a/socorro/unittest/lib/test_threaded_task_manager.py
+++ b/socorro/unittest/lib/test_threaded_task_manager.py
@@ -4,7 +4,7 @@
 
 import time
 
-from nose.tools import ok_, eq_
+from nose.tools import ok_
 from mock import Mock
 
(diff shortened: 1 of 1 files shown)


peterbe:
Hopefully once we go full jansky on this code this'll be a thing of the past. Till then, would be nice to not have to press the "Rebuid job" button so often.

PR #3900, opened 2017-08-08
bug 1387100 - improve webapp in dev environment

willkg:
This adds lookup table data population and documented run lines that add release
data to the db. These make it possible to go to
http://localhost:8000/home/product/Firefox and have it mostly work.

Note that this only pulls release data for Firefox. Each additional product
takes another like 10 minutes to run. The thinking is that we're primarily
working on Firefox crashes, so if someone needs release data for something else,
they can adjust the scripts accordingly. If we need to make that more flexible,
we can make that change later.

### DIFF
--- a/docker/run_update_release_data.sh
+++ b/docker/run_update_release_data.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# Updates product release data in the docker environment.
+#
+# Usage: docker/run_update_release_data.sh
+
+# Fetch and update release information for these products (comma-delimited)
+PRODUCTS="firefox"
+
+# Update with this frequency
+FREQUENCY="1d"
+
+# Fetch release data
+python socorro/cron/crontabber_app.py --job=ftpscraper \
+       --crontabber.class-FTPScraperCronApp.frequency=${FREQUENCY} \
+       --crontabber.class-FTPScraperCronApp.products=${PRODUCTS}
+
+# Update featured versions data based on release data
+python socorro/cron/crontabber_app.py --job=featured-versions-automatic \
+       --crontabber.class-FeaturedVersionsAutomaticCronApp.frequency=${FREQUENCY} \
+       --crontabber.class-FeaturedVersionsAutomaticCronApp.products=${PRODUCTS}
--- a/docker/run_update_release_data.sh
+++ b/docker/run_update_release_data.sh
@@ -4,6 +4,8 @@
 #
 # Usage: docker/run_update_release_data.sh
 
+set -eo pipefail
+
 # Fetch and update release information for these products (comma-delimited)
 PRODUCTS="firefox"
 
--- a/docs/gettingstarted.rst
+++ b/docs/gettingstarted.rst
@@ -41,15 +41,18 @@ Quickstart
 
      $ make dockersetup
 
+
+   You can run ``make dockersetup`` any time you want to wipe the database, pick
+   up changes in static data, stored procedures, types, migrations, etc.
+
    .. Warning::
 
-      This is a work in progress, isn't idempotent, and is fussy about the state
-      of things.
+      This is a work in progress and might be fussy about some things.
 
       Pull requests welcome!
 
 
-At that point, you should have a basic functional Socorro development
+At this point, you should have a basic functional Socorro development
 environment.
 
 See :ref:`webapp-chapter` for additional setup and running the webapp.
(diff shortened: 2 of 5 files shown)


willkg:
The goal here is that this gets you to a mostly functional Firefox homepage:

1. `make dockerbuild`
2. `make dockersetup`
3. `docker-compose run processor docker/run_update_release_data.sh`
4. `docker-compose up webapp`

Then you browse to ``http://localhost:8000`` and get a Firefox home page that has active versions, but doesn't have any ADI or crash data.

peterbe:
One that we're going to need to do is to get fancy with the `collectstatic` stuff. It's a bore to have to wait for since when you run with DEBUG=False you don't need static assets collected first. In fact, with DEBUG=True the collect static files aren't being used. 

Devs need to be able to test running the webapp with DEBUG=False and javascript files minified. But generally this is only needed in CI after the tests pass and artifacts are built.

willkg:
I agree--that's definitely worth spinning out into a separate bug because it's tricky.

socorro/external/postgresql/staticdata.py
```diff
@@ -189,8 +212,79 @@ class ReportPartitionInfo(BaseTable):
...
+        ['Firefox', '1', '5.0', 'firefox', '23.0'],
+        ['SeaMonkey', '6', '2.3', 'seamonkey', '999.0'],
+        ['FennecAndroid', '4', '5.0', '**SPECIAL**', '999.0'],
+    ]
+
+
+class ProductBuildTypes(BaseTable):
+    table = 'product_build_types'
+    columns = [
+        'product_name', 'build_type'
```
peterbe: Add `, 'throttle'` here and ...

socorro/external/postgresql/staticdata.py
```diff
@@ -189,8 +212,79 @@ class ReportPartitionInfo(BaseTable):
...
+    ]
+    rows = [
+        ['Firefox', 'release'],
```
peterbe: Make this `['Firefox', 'release', 0.1],` and for all other 2-item lists make it `1.0` instead of `0.1`. E.g. `['Firefox', 'beta', 1.0]`.

socorro/external/postgresql/staticdata.py
```diff
@@ -100,14 +100,37 @@ class OSVersions(BaseTable):
...
+    rows = [
+        ['mozilla-central'],
+        ['mozilla-1.9.2'],
+        ['comm-central'],
+        ['comm-1.9.2'],
+        ['comm-central-trunk'],
+        ['mozilla-central-android'],
+        ['mozilla-release'],
+        ['mozilla-beta'],
+        ['mozilla-aurora'],
```
willkg: I want this to be -prod like. I'm not wildly excited about removing things here that are in -prod. When we remove the aurora stuff in bug #1385285 and assuming that covers references in the db, too, I'd be game for removing it in staticdata. Until then, I want to keep it.

socorro/external/postgresql/staticdata.py
willkg: Huh... I'm not sure how I forgot that. I'll add those now.

docker/run_update_release_data.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
```
willkg: Yes! I'll add that now.

docs/gettingstarted.rst
```diff
@@ -41,24 +41,66 @@ Quickstart
...
+   which can take a long time, maybe hours.
+
+
+General architecture
+====================
+
+.. image:: block-diagram.png
+
+Arrows direction represents the flow of interesting information (crashes,
+authentication assertions, cached values), not trivia like acks.
```
willkg: Yeah. I want to switch this to dot and graphviz, but haven't, yet. I'm using that with Antenna and it's been good:

http://antenna.readthedocs.io/en/latest/architecture.html#high-level-architecture

It has the advantage of being text-based and easier to edit in the docs:

https://github.com/mozilla-services/antenna/blob/master/docs/architecture.rst#high-level-architecture

I think that's outside the scope of this pass, though.

willkg:
> I think the homepage is failing because there is absolutely no crashes in ES. Should we bother updating the JS code to not choke if there is exactly 0 crashes?

It's definitely from missing data. When we talked yesterday, you said you'd work on building a list of all the tables the product home page needs to work completely. I know this PR covers some/most of them. I know it's missing ADI data (I'm working on that in a separate PR). I know it's missing crash data. Beyond that, I have no idea what it needs and the code is hard to follow.

That's why this PR limits the scope to "get the things required so we don't get an HTTP 404 on the product home page anymore".

Towards updating the JS code to not choke, if that's not too hard to do, I think it's worth doing because it improves the ergonomics of the dev environment.

> How can I help write a thing that populates the ADI from prod? Thing is what's in the public ADIs is the materialized view. To populate locally we'd need to first fill in raw_adi then execute the update_adu.sql stored procedure so it combines the raw_adi table with product_versions normalized.

I'm working on a script now that pulls ADI data from the API and populates postgres. I'll let you know if I hit problems.

> I'm getting this traceback ...

I don't get any tracebacks when running the webapp. I'm guessing either you have a `.env` file that's different or maybe you've got additional bits in your `socorro/` directory that aren't part of the repo maybe from running Socorro on the host or something along those lines.

> SuperSearchFields isn't working. And that's a whole lot of other work but I'm not interested in solving this the old way (populating ES with the supersearch fields) until we can start using the super_search_fields.json file that is currently "pending" until we upgrade to ES 5.

I'd like to get supersearch working, but I agree we should wait until `super_search_fields.json` lands. Depending on how long that takes, I might take that work and land it before ES 5. I'll figure that out later, though.

I think that covers all the outstanding things in this issue.

willkg:
Travis is fine, so I'm going to land this.

peterbe:
Yeah. This was the PR I was waiting for to see what the home page needs. 
I think it's simpler and harder than we imagined. There's nothing else needed in lookup tables needed. However, it won't work without something out of supersearch and without any adi it won't error but it'll simply not show the line in the graph. 

Let's worry about the JS error (if the supersearch result is empty) until you've made the other progress.

PR #3902, opened 2017-08-09
bug 1377254 fix lint issues in socorro/lib/

willkg:


socorro/lib/prioritize.py
```diff
@@ -1,158 +0,0 @@
-# This Source Code Form is subject to the terms of the Mozilla Public
```
willkg: This module isn't used anywhere, so I removed the whole thing.

socorro/lib/util.py
```diff
@@ -2,219 +2,147 @@
...
-        else:
-          self.cache.append(x)
+def report_exception_and_continue(logger=FakeLogger(), loggingLevel=logging.ERROR,
+                                  ignoreFunction=None, showTraceback=True):
+    try:
+        exc_type, exc, tb = sys.exc_info()
+        if ignoreFunction and ignoreFunction(exc_type, exc, tb):
+            return
+        if exc_type in (KeyboardInterrupt, SystemExit):
+            raise
```
willkg: This is the only thing I changed that's behaviorally different. This used to be:

```
raise exception
```

The problem with that is that it's raising a *new* exception with a new traceback and I think this should be re-raising the exception that was raised.

I think this was a bug. Not a big one because those two particular exceptions are "the user has killed this process intentionally", but still, not great.

willkg:
I added some comments for things that aren't cosmetic. The rest of this is mostly cosmetic and doesn't change the behavior at all.

willkg:
Oh, one other thing. I don't know what to do with `socorro/lib/ooid.py`. A bunch of those functions were only used by the collector which is gone now, but they cover things like generating and manipulating crash ids.

Should that stay in the tree or get removed? If it stays, should we fix the names for things?

(These aren't "let's think about this for a week" questions. These are "let's think about this for 30 seconds and maybe roll a d6 and move on with our lives" questions.)

socorro/lib/prioritize.py
lonnen: +1

lonnen: Looks like this was missed when we refactored out the other griswolf code some years back. Good catch.

socorro/lib/util.py
lonnen: Thanks for pointing this out. The diff on much of this was a bear. I ended up looking at your post-change files to understand. I think there are secret args to change the diffing strategy to one of the other defaults but I have forgotten what they are, and they may not even be valid any more. OR I'm making it all up.

socorro/lib/ver_tools.py
```diff
@@ -106,12 +113,14 @@ def _convert_to_str(x):
...
+
 class NotAVersionException(Exception):
     """An exception raised by the class VersionComparer if given a string
     outside the standards of a version string as specified by the regular
     expression _version_part_re"""
     pass
 
+
 @memoize(1000)
 def normalize(version_string, max_version_parts=4):
```
lonnen: I believe this is the only method from this module that is still in use.

lonnen: consider pruning this or moving this method and deleting the rest. The SupportClassificationRule class has other normalize methods. Maybe it belongs in there.

willkg:
^^^ That removes unused things in `ver_tools.py`. I thought about moving the functions to the processor, but I'm not sure they're long for this world and it didn't seem worth spending another 5-10 minutes on.

That also cleans up `ooid.py`. I didn't end up removing functions. I'm going to write a thing later that inserts crash data into the processor and figured I'd wait until after that to remove things I didn't end up using.

@lonnen Can you skim those? Also, are there better ways to do linting fix PRs that yield easier to work through diffs? Maybe we should split linting-related PRs from code-removal PRs?

willkg:
I got an r+, then made some minor changes, but I'm going to self-r+ those changes to land this and move forward.

PR #3903, opened 2017-08-09
Add docker-compose.override.yml to .gitignore

willkg:
docker-compose.override.yml allows people to override docker-compose things
without having to create a new file and point to it. These files shouldn't get
added to git, so let's ignore them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,4 @@ e2e-tests/.cache/
 # docker things
 .docker-build
 .cache/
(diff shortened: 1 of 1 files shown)


willkg:
I keep forgetting to add this.

It's talked about here: https://docs.docker.com/compose/extends/#multiple-compose-files

willkg:
Actually, I'm just going to land this. It's benign.

PR #3906, opened 2017-08-10
fixes bug 1388878 - Memoize ftpscraper downloads

peterbe:
I estimate that the whole thing will take about 20-25% shorter time now. 
I guess we should land this, upgrade stage admin and watch the crontabber duration times for `ftpscraper` and watch it decrease over time.


socorro/cron/jobs/ftpscraper.py
```diff
@@ -271,14 +271,23 @@ def __init__(self, *args, **kwargs):
         )
 
     def download(self, url):
+        try:
+            return self._download_cache[url]
```
peterbe: The speed difference between try:except and `in` is negligible when doing only a few thousand of these. I did it like this to keep the cache local and not require changes outside that method. 

Instead, a cleaner solution altogether is to use a memoizer. I'll just switch to that instead.

peterbe:
Code updated.

socorro/cron/jobs/ftpscraper.py
```diff
@@ -25,6 +26,45 @@
...
+        try:
+            cache = obj.__cache
+        except AttributeError:
+            cache = obj.__cache = {}
+        key = (self.func, args[1:], frozenset(kw.items()))
+        if key in cache:
+            res = cache[key]
+        else:
+            res = cache[key] = self.func(*args, **kw)
+        return res
```
willkg: This is overly complex. All you need is something like this:

```
def memoize_download(fun):
    cache = {}
    def _memoize_download(url):
        if url in cache:
            return cache[url]
        ret = fun(url)
        cache[url] = ret
        return ret
```

Or something along those lines.

socorro/cron/jobs/ftpscraper.py
```diff
@@ -25,6 +25,17 @@
 from socorro.lib.datetimeutil import string_to_datetime
 
 
+def memoize_download(fun):
+    cache = {}
+
+    def inner(self, url):
```
willkg: My bad--this should have a `functools.wraps` around it. I forgot that in my earlier comment.

peterbe:
Now with wraps.

PR #3907, opened 2017-08-10
avoid unnecessary 404s when finding candidate builds

peterbe:
Found this low hanging fruit. Today, about 10% of the URLs going through `download` cause a 404.

### DIFF
--- a/socorro/cron/jobs/ftpscraper.py
+++ b/socorro/cron/jobs/ftpscraper.py
@@ -343,7 +343,11 @@ class FTPScraperCronApp(BaseCronApp, ScrapersMixin):
 
     def _insert_build(self, cursor, *args, **kwargs):
         self.config.logger.debug('adding %s', args)
-        if not self.config.dry_run:
+        if self.config.dry_run:
+            print "INSERT BUILD"
+            print args
+            print kwargs
+        else:
             buildutil.insert_build(cursor, *args, **kwargs)
 
     def _is_final_beta(self, version):
--- a/socorro/cron/jobs/ftpscraper.py
+++ b/socorro/cron/jobs/ftpscraper.py
@@ -152,8 +152,21 @@ class ScrapersMixin(object):
         )
 
         for platform in possible_platforms:
-            platform_urls = self.get_links(latest_build, starts_with=platform)
+            platform_urls = self.get_links(
+                latest_build,
+                starts_with=platform
+            )
+
             for platform_url in platform_urls:
+                # We're only interested in going into depper directories.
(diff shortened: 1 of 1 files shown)


socorro/cron/jobs/ftpscraper.py
```diff
@@ -313,11 +329,7 @@ def _scrape_json_releases_and_nightlies(
 
     def _insert_build(self, cursor, *args, **kwargs):
         self.config.logger.debug('adding %s', args)
-        if self.config.dry_run:
-            print "INSERT BUILD"
-            print args
-            print kwargs
-        else:
```
willkg: This is important for dry runs because it specifies what would have been inserted. Please don't take it out.

peterbe: It's output is no different from the line above. The logging. On stdout you get:

```
2017-08-11 11:23:11,393 DEBUG - SocorroAppBaseClass -  - MainThread - adding ('firefox', '49.0.1', 'win', u'20160922113459', 'release', None, u'mozilla-release', 'build3')
INSERT BUILD
('firefox', '49.0.1', 'win', u'20160922113459', 'release', None, u'mozilla-release', 'build3')
{'ignore_duplicates': True}
```
When I ran this a bunch of test to test I noticed the excess and thought it's no longer worth keeping.

willkg: They're actually not the same.

But whatever. When I go work on the buildhub script and need to compare what it's doing with ftpscraper, I'll re-add id locally. :(

willkg:
I need to make sure this is not skipping things it shouldn't be skipping which has some big consequences. I'll try to do it next week.

peterbe:
> I need to make sure this is not skipping things it shouldn't be skipping which has some big consequences. I'll try to do it next week.

I tested it by inserting something like this inside `download()`:
```python
with open('/tmp/urls.log', 'a') as f:
    f.write('{}\t{}\n'.format(url, response.status_code))
```
I concluded that the ones that did 200 OK were the same before and after.

socorro/cron/jobs/ftpscraper.py
peterbe: They both print `args` one with `config.logger` and one with `print`. Feel free to add whatever more extra you want when you really need to do the comparison.

lonnen: I'm with WillKG. Please retain this dry run feature.

willkg:
@lonnen Peter undid the print-removal change. Can you approve this?

PR #3908, opened 2017-08-10
Add script to fetch normalization data

willkg:
This adds a script to fetch normalization data. It has to run after we've run ftpscraper, so I stuck it in the `run_update_release_data.sh`. I'm thinking at some point, we'll want to rename that script, but I don't have better ideas at the moment.

This PR lets us do the following:

1. run `make dockersetup`
2. run `docker-compose run processor docker/run_update_release_data.sh`
3. now the adi url endpoint will work for localhost:8000

### DIFF
--- a/docker/fetch_normalization_data.py
+++ b/docker/fetch_normalization_data.py
@@ -130,10 +130,10 @@ def main(args):
         help='comma separated set of products to fetch ADI for'
     )
     parser.add_argument(
-        '--start_date', default='LASTWEEK', help='start date (mm/dd/yyyy)'
+        '--start_date', default='LASTWEEK', help='start date (yyyy-mm-dd)'
     )
     parser.add_argument(
-        '--end_date', default='TODAY', help='end date (mm/dd/yyyy)'
+        '--end_date', default='TODAY', help='end date (yyyy-mm-dd)'
     )
     parser.add_argument(
         '--endpoint', default='https://crash-stats.mozilla.com/api/ADI/',
@@ -141,16 +141,16 @@ def main(args):
     )
     args = parser.parse_args()
 
-    today = datetime.datetime.now()
-    last_week = datetime.datetime.now() - datetime.timedelta(days=7)
+    today = datetime.datetime.utcnow()
+    last_week = today - datetime.timedelta(days=7)
 
     start_date = args.start_date
     if start_date == 'LASTWEEK':
-        start_date = last_week.strftime('%m/%d/%Y')
+        start_date = last_week.strftime('%Y-%m-%d')
 
     end_date = args.end_date
     if end_date == 'TODAY':
-        end_date = today.strftime('%m/%d/%Y')
+        end_date = today.strftime('%Y-%m-%d')
 
     if start_date > end_date:
         print('Start date (%s) cannot be after end date (%s). Exiting.' % (start_date, end_date))
--- a/docker/fetch_normalization_data.py
+++ b/docker/fetch_normalization_data.py
@@ -22,14 +22,14 @@ import requests
 PLATFORMS = ['Mac OS X', 'Linux', 'Windows', 'Unknown']
 
 
-def fetch_data_from_api(endpoint, product, version, platform, start_date, end_date):
+def fetch_data_from_api(endpoint, product, platform, versions, start_date, end_date):
     """Fetches data from API"""
     req = requests.get(endpoint, params={
         'end_date': end_date,
         'start_date': start_date,
         'product': product,
         'platforms': platform,
-        'versions': version,
+        'versions': versions,
     })
     if req.status_code != 200:
         print('Error retrieving ADI: %s' % req.content)
@@ -39,7 +39,7 @@ def fetch_data_from_api(endpoint, product, version, platform, start_date, end_da
     return data['hits']
 
 
-def fetch_versions(conn, product_name):
+def get_versions(conn, product_name):
     """Fetch the version strings for a product from product_versions table"""
 
     cursor = conn.cursor()
@@ -54,7 +54,7 @@ def fetch_versions(conn, product_name):
     return list(versions)
 
 
-def fetch_product_version_id(conn, product_name, version_string):
+def get_product_version_id(conn, product_name, version_string):
     """Given a product name and a version_string, returns a product_version_id"""
     cursor = conn.cursor()
     cursor.execute("""
@@ -73,7 +73,7 @@ def fetch_product_version_id(conn, product_name, version_string):
 def insert_adu(conn, adu_count, adu_date, product_name, os_name, version_string):
     """Inserts ADU data into product_adu table"""
     # Fetch the product_version_id we need
-    product_version_id = fetch_product_version_id(conn, product_name, version_string)
+    product_version_id = get_product_version_id(conn, product_name, version_string)
 
     if product_version_id is None:
         print('No product_version_id for %s %s' % (product_name, version_string))
@@ -164,31 +164,28 @@ def main(args):
 
     for product in args.products.split(','):
         # Fetch versions from the product_versions table that we should pull data for
-        versions = fetch_versions(conn, product)
+        versions = get_versions(conn, product)
 
-        for version in versions:
-            for platform in PLATFORMS:
-                # For each (product, version, platform) combo, fetch adu data and insert it
-                print('Fetching data for (%s, %s, %s)...' % (product, version, platform))
+        for platform in PLATFORMS:
+            print('Fetching data for (%s, %s, %s)...' % (product, versions, platform))
+            adu_data = fetch_data_from_api(
+                endpoint=args.endpoint,
+                product=product,
+                platform=platform,
+                versions=versions,
+                start_date=start_date,
+                end_date=end_date
+            )
 
-                adu_data = fetch_data_from_api(
-                    endpoint=args.endpoint,
-                    product=product,
-                    version=version,
-                    platform=platform,
-                    start_date=start_date,
-                    end_date=end_date
+            for hit in adu_data:
+                insert_adu(
+                    conn=conn,
+                    adu_count=hit['adi_count'],
+                    adu_date=hit['date'],
+                    product_name=product,
+                    os_name=platform,
+                    version_string=hit['version'],
                 )
-
-                for hit in adu_data:
-                    insert_adu(
-                        conn=conn,
-                        adu_count=hit['adi_count'],
-                        adu_date=hit['date'],
-                        product_name=product,
-                        os_name=platform,
-                        version_string=hit['version'],
-                    )
 
     conn.close()
 
--- a/docker/fetch_normalization_data.py
+++ b/docker/fetch_normalization_data.py
@@ -1,0 +1,192 @@
+#!/usr/bin/env python
+
+"""
+Fetches normalization data.
+
+Usage:
+
+    python docker/fetch_normalization_data.py
+
+"""
+
+import argparse
+import datetime
+import os
+import sys
+
+import psycopg2
+import requests
+from six.moves.urllib.parse import quote_plus
+
+
+# NOTE(willkg): These platforms come from the -prod data.
+PLATFORMS = ['Mac OS X', 'Linux', 'Windows', 'Unknown']
+
+
+def fetch_data_from_api(endpoint, product, version, platform, start_date, end_date):
+    """Fetches data from API"""
+    params = []
+    params.append(('end_date', end_date))
+    params.append(('start_date', start_date))
+    params.append(('product', product))
+    params.append(('platforms', platform))
+    params.append(('versions', version))
+
+    qs = '&'.join(['%s=%s' % (param[0], quote_plus(param[1])) for param in params])
+    url = endpoint + '?' + qs
+    req = requests.get(url)
+    if req.status_code != 200:
+        print('Error retrieving ADI: %s' % req.content)
+        return []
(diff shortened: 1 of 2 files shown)


docker/fetch_normalization_data.py
```diff
@@ -0,0 +1,192 @@
...
+
+    # FIXME(willkg): There's probably a better way to do this, but I can't think of it offhand.
+    cursor = conn.cursor()
+    cursor.execute("""
+        SELECT version_string
+        FROM product_versions
+        WHERE
+            product_name=%s
+    """, (product_name,))
+    versions = set([record[0] for record in cursor if not record[0].endswith('b')])
```
willkg: This is curious. It seems like the API call returns all the data for versions with the specified *prefix*. So if we pass in something like 55.0b, then it returns data for 55.0b1, 55.0b2, ... and that gets messy.

https://crash-stats.mozilla.com/api/ADI/?end_date=2017-08-10&platforms=Mac%20OS%20X&product=Firefox&start_date=2017-08-03&versions=55.0b

So what I do here is skip that version because I don't know what else to do with it and the data from the API doesn't include it anyhow.

Is that the right thing to do? Is there a better way to deal with it?

docker/fetch_normalization_data.py
```diff
@@ -0,0 +1,192 @@
...
+        'database_name': dbname
+    }
+    dsn = (
+        "host=%(database_hostname)s "
+        "dbname=%(database_name)s "
+        "port=%(database_port)s "
+        "user=%(database_username)s "
+        "password=%(database_password)s"
+    ) % local_config
+    return psycopg2.connect(dsn)
```
willkg: I tossed around whether to use App and postgres ConnectionContext and the other things and then didn't. I think we're trying to move away from those things. Having said that, it's nice to have easy-to-reuse db scaffolding that works with existing configuration.

Anyhow, kick me if this is a bad idea and I can spend some time to rewrite it.

docker/fetch_normalization_data.py
```diff
@@ -0,0 +1,192 @@
...
+    params = []
+    params.append(('end_date', end_date))
+    params.append(('start_date', start_date))
+    params.append(('product', product))
+    params.append(('platforms', platform))
+    params.append(('versions', version))
+
+    qs = '&'.join(['%s=%s' % (param[0], quote_plus(param[1])) for param in params])
+    url = endpoint + '?' + qs
+    req = requests.get(url)
```
willkg: Oo--I can redo this params stuff now that I'm not passing in multiples of things. I'll do that now.

docker/fetch_normalization_data.py
```diff
@@ -0,0 +1,197 @@
...
+
+    cursor = conn.cursor()
+    cursor.execute("""
+        SELECT version_string
+        FROM product_versions
+        WHERE
+            product_name=%s
+    """, (product_name,))
+    # NOTE(willkg): We don't want versions that end with "b" because they're weird.
+    versions = set([record[0] for record in cursor if not record[0].endswith('b')])
```
willkg: This is curious. It seems like the API call returns all the data for versions with the specified prefix. So if we pass in something like 55.0b, then it returns data for 55.0b1, 55.0b2, ... and that gets messy.

https://crash-stats.mozilla.com/api/ADI/?end_date=2017-08-10&platforms=Mac%20OS%20X&product=Firefox&start_date=2017-08-03&versions=55.0b

So what I do here is skip that version because I don't know what else to do with it and the data from the API doesn't include it anyhow.

Is that the right thing to do? Is there a better way to deal with it?

docker/fetch_normalization_data.py
```diff
@@ -0,0 +1,192 @@
...
+    today = datetime.datetime.now()
+    last_week = datetime.datetime.now() - datetime.timedelta(days=7)
+
+    start_date = args.start_date
+    if start_date == 'LASTWEEK':
+        start_date = last_week.strftime('%m/%d/%Y')
+
+    end_date = args.end_date
+    if end_date == 'TODAY':
+        end_date = today.strftime('%m/%d/%Y')
```
peterbe: Perhaps our Django form validation thing supports the US format. Either way, I think ISO is best to stick to.

peterbe:
One thing I've come to understand is that the ADI is the same as PROD but the home page works differently because the product versions are different. I wonder why. 
In particular, I think my localhost:8000 lacks versions 55.0, 56.0b1 and 57.0a1.

docker/fetch_normalization_data.py
willkg: The API documentation for this endpoint says to use "mm/dd/yyyy" format.

peterbe:
One thing I've come to understand is that the ADI is the same as PROD but the home page works differently because the product versions are different. I wonder why. 
In particular, I think my localhost:8000 lacks versions 55.0, 56.0b1 and 57.0a1.

docker/fetch_normalization_data.py
```diff
@@ -0,0 +1,197 @@
...
+    parser.add_argument(
+        '--end_date', default='TODAY', help='end date (mm/dd/yyyy)'
+    )
+    parser.add_argument(
+        '--endpoint', default='https://crash-stats.mozilla.com/api/ADI/',
+        help='endpoint for ADI API'
+    )
+    args = parser.parse_args()
+
+    today = datetime.datetime.now()
```
willkg: This is used to build a range of days to use for the API query. There's nothing in the API documentation that suggests what timezone things are in. Given that these are days, I'm not sure it matters a whole lot whether we use utc.

I can change it if you think it's important.

docker/fetch_normalization_data.py
```diff
@@ -0,0 +1,197 @@
...
+
+    if product_version_id is None:
+        print('No product_version_id for %s %s' % (product_name, version_string))
+        return
+
+    # Insert the data
+    cursor = conn.cursor()
+    try:
+        cursor.execute("""
+            INSERT INTO product_adu (adu_count, adu_date, os_name, product_version_id)
```
willkg: The `raw_adi` table requires data that's not available via the API endpoint, so it's not possible to do that.

docker/fetch_normalization_data.py
willkg: I fixed that 18 hours ago.

docker/fetch_normalization_data.py
willkg: One thing @milescrabill and I were talking about yesterday was how much redundancy we have in configuration regarding database configuration. I've got too many things on my plate to tackle that now before this PR lands, but it sounds like a good bug to write up to clean up Socorro.

docker/fetch_normalization_data.py
willkg: Nevermind--I'll just change it now.

docker/fetch_normalization_data.py
```diff
@@ -0,0 +1,197 @@
...
+
+                for hit in adu_data:
+                    insert_adu(
+                        conn=conn,
+                        adu_count=hit['adi_count'],
+                        adu_date=hit['date'],
+                        product_name=product,
+                        os_name=platform,
+                        version_string=hit['version'],
+                    )
```
peterbe: Perhaps it's nitpickery but all of this can be simplified. The `/api/ADI/` uses an `OR` clause on multiple platforms and multiple versions (and multiple products too). 
So to avoid having to do `len(products) x len(product_versions) x len(platforms)` number of network requests you can just do:

```python
for product in args.products.split(','):
    adi_data = fetch_data_from_api(
        product=product,
        versions=fetch_versions(conn, product),
        platforms=PLATFORMS,
        start_date=start_date,
        end_date=end_date
    )
    for hit in adi_data:
         insert_adi(...)
```

peterbe:
General comment; Can you replace (almost) all mentions of the term "adu" and replace it with "adi". The ancient term is "Active Daily Users". The modern term is "Active Daily Installs". (if you can even call that "modern" since kilousagehours is the new hotness). 
The only reason there's "adu" inside the code is because it's too cumbersome to rename SQL stuff. I.e. it wasn't worth making an alembic migration to rename `product_adu` to `product_adi`.

docker/fetch_normalization_data.py
peterbe: Yeah, it's cryptic. In the footer of the webapp it says "All dates are UTC" and we have some code somewhere that makes sure the database's timezone is always set in UTC.

docker/fetch_normalization_data.py
peterbe: Good point. It's a mess. The webapp's got it right. It has two URIs. `DATABASE_URL` and `IMPLEMENTATIONS_DATABASE_URL`. The latter is optional if you want to have two different database connections. One for Django stuff and one for ssh tunnelling. 

But perhaps we can start a separate thread to get the processor to use `DATABASE_URL`. Then we can slowly eradicate all these `database_*` things. 

Places where it uses individual keys are: crontabber and processor. (I think)

docker/fetch_normalization_data.py
peterbe: No, that's probably just your browser that has a date widget. If you click in a date and submit the form, the browser will actually send it in ISO format.

docker/fetch_normalization_data.py
peterbe: The default rate limit is 100/m but more worrying is that if we do 43 requests, one of them could get stuck and fail due to back luck.

docker/fetch_normalization_data.py
willkg: Hot diggity--it is sending yyyy-mm-dd format. I'll fix that.

I really don't like our API docs. I find them misleading in many circumstances and lacking critical details. I have to go hunt down the model and read through the code to figure out what it's doing. It's frustrating.

willkg:
Things to fix:

1. switch to utc dates
2. change date format to yyyy-mm-dd
3. see about changing the `fetch_data_from_api` call to fetch multiple things at the same time -- I started with that and had to switch, but I don't remember why; I'll try again

Any thoughts on either of these two questions?

https://github.com/mozilla-services/socorro/pull/3908#discussion_r132538413

https://github.com/mozilla-services/socorro/pull/3908#discussion_r132551844

docker/fetch_normalization_data.py
peterbe: Yep. It was an old pending comment.

docker/fetch_normalization_data.py
peterbe: It's a hot mess. 
https://github.com/mozilla-services/socorro/blob/d8080a5e45909fbe7d0ed415d49bd23e1794939c/socorro/external/postgresql/adi.py#L47-L58

Actually you don't need to do this at all. 
If you do and `/api/ADI` query for `?versions=55.0b&versions=55.0b3` you get the same output as if you'd done `?versions=55.0b3`. 
I.e. some ADI counts for `55.0b3` but `55.0b` omitted from the result. 

```
breakpad=> select version_string, adu_count, adu_date from product_adu a, product_versions pv where pv.product_version_id=a.product_version_id and version_string like '55.%' and version_string like '%b';
 version_string | adu_count | adu_date
----------------+-----------+----------
(0 rows)
```

So in conclusion, I suggest you simply change it to:
```python
versions = [record[0] for record in cursor]
```

willkg: >  If you do and `/api/ADI` query for `?versions=55.0b&versions=55.0b3` you get the same output as if you'd done `?versions=55.0b3`.

It's not... The latter only gets you ADI for that specific version and the former gets you ADI for all the beta versions including that version:

55.0b and 55.0b3:

https://crash-stats.mozilla.com/api/ADI/?end_date=2017-08-10&platforms=Mac%20OS%20X&product=Firefox&start_date=2017-08-03&versions=55.0b&versions=55.0b3

just 55.0b3:

https://crash-stats.mozilla.com/api/ADI/?end_date=2017-08-10&platforms=Mac%20OS%20X&product=Firefox&start_date=2017-08-03&versions=55.0b3

Thank you for explaining what was going on! I think I'm going to leave it in because I don't want to deal with a prefix match.

willkg:
@peterbe ^^^ That fixes the issues you brought up.

docker/fetch_normalization_data.py
```diff
@@ -0,0 +1,194 @@
...
+
+    print('Fetching normalization data for %s from %s to %s' % (
+        args.products, start_date, end_date)
+    )
+
+    for product in args.products.split(','):
+        # Fetch versions from the product_versions table that we should pull data for
+        versions = get_versions(conn, product)
+
+        for platform in PLATFORMS:
```
peterbe: Nit: If there's a reason for not fetching for all platforms all at once perhaps a comment here. 
After all, it *could* be this:
```python
        print('Fetching data for (%s, %s, %s)...' % (product, versions, PLATFORMS))
        adu_data = fetch_data_from_api(
            endpoint=args.endpoint,
            product=product,
            platform=PLATFORMS,
            versions=versions,
            start_date=start_date,
            end_date=end_date
        )
```
But consider it a nit. It's not painful to have to do 4 API calls instead of 1.

willkg: We can't because the platform doesn't come back in the data and we need it to insert a row.

willkg:
Thank you!

docker/fetch_normalization_data.py
peterbe: Good catch! But that's kinda nuts. Who knows why we did it like that?! 
Let's let that stay that way if it worksish.

PR #3910, opened 2017-08-11
fixes bug 1389129 - make StackwalkerErrorSignatureRule more resilient

willkg:
It's possible for a processed crash to not have an mdsw_status_string, so the
rule should allow for that. This fixes the predicate accordingly and adds tests.


socorro/processor/signature_utilities.py
```diff
@@ -610,7 +610,10 @@ def version(self):
         return '1.0'
 
     def _predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
-        return processed_crash['signature'].startswith('EMPTY')
+        return bool(
+            processed_crash['signature'].startswith('EMPTY') and
+            processed_crash.get('mdsw_status_string')
+        )
```
willkg: This adjusts the predicate so it only returns true iff the signature starts with `EMPTY` and there's an `mdsw_status_string` value. The figuring is that if we don't have a value, then there's nothing to concatenate in the action anyhow.

socorro/unittest/processor/test_signature_utilities.py
```diff
@@ -1547,13 +1547,23 @@ def test_predicate_no_match(self):
         predicate_result = rule.predicate({}, {}, processed_crash, fake_processor)
         assert predicate_result is False
 
-    def test_predicate(self):
+    def test_predicate_no_match_missing_mdsw_status_string(self):
         processed_crash = {
             'signature': 'EMPTY: like my soul'
```
willkg: Some of the test data is very sad.

willkg:
Thank you!

PR #3911, opened 2017-08-11
Fix bug 1389474 - add 'alloc::oom::oom' to prefix skiplist

calixteman:


willkg:
Once this is deployed in -stage, I'll verify it there.

PR #3912, opened 2017-08-11
fixes bug 1382727 - remove bad keys from ES documents

willkg:
This adds a pass to remove bad keys from the raw_crash part of the document
before indexing it into Elasticsearch.

Good keys contain one or more ascii alpha-numeric characters plus underscore and
hyphen. Everything else is invalid and gets removed.

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -37,6 +37,22 @@ class RawCrashRedactor(Redactor):
         ]
 
 
+# Valid Elasticsearch keys contain one or more ascii alphanumeric characters, underscore, and hyphen
+# and that's it.
+VALID_KEY = re.compile(r'^[a-zA-Z0-9_-]+$')
+
+
+def is_valid_key(key):
+    """Validates an Elasticsearch document key
+
+    :arg string key: the key to validate
+
+    :returns: True if it's valid and False if not
+
+    """
+    return bool(VALID_KEY.match(key))
+
+
 class ESCrashStorage(CrashStorageBase):
     """This sends raw and processed crash reports to Elasticsearch."""
 
@@ -168,12 +184,38 @@ class ESCrashStorage(CrashStorageBase):
                 # not there? we don't care
                 pass
 
+    @staticmethod
+    def remove_bad_keys(raw_crash):
+        """Removes keys from the top-level of the dict that are bad
+
+        Good keys satisfy the following properties:
+
+        * have one or more characters
+        * are composed of a-zA-Z0-9_-
+
+        Anything else is a bad key and needs to be removed.
+
+        This modifies the crash in-place.
+
+        :arg dict raw_crash: raw crash data
+
+        """
+        if not raw_crash:
+            return
+
+        for key in list(raw_crash.keys()):
+            if not is_valid_key(key):
+                del raw_crash[key]
+
     def _submit_crash_to_elasticsearch(self, connection, crash_document):
         """Submit a crash report to elasticsearch.
         """
         # Massage the crash such that the date_processed field is formatted
         # in the fashion of our established mapping.
         self.reconstitute_datetimes(crash_document['processed_crash'])
+
+        # Remove bad keys from the raw crash.
+        self.remove_bad_keys(crash_document['raw_crash'])
 
         # Obtain the index name.
         es_index = self.get_index_for_crash(
@@ -190,8 +232,7 @@ class ESCrashStorage(CrashStorageBase):
         # Submit the crash for indexing.
         # Don't retry more than 5 times. That is to avoid infinite loops in
         # case of an unhandled exception.
-        times = range(5)
-        while times.pop(-1):
+        for attempt in range(5):
             try:
                 connection.index(
                     index=es_index,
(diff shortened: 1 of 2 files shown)


socorro/external/es/crashstorage.py
```diff
@@ -37,6 +37,22 @@ def __init__(self, config):
         ]
 
 
+# Valid Elasticsearch keys contain one or more ascii alphanumeric characters, underscore, and hyphen
+# and that's it.
+VALID_KEY = re.compile(r'^[a-zA-Z0-9_-]+$')
```
willkg: This rule is based on the rule in this comment:

https://bugzilla.mozilla.org/show_bug.cgi?id=1382727#c6

The rule in that comment looks for "bad characters". This is the inverse of that where it makes sure the key only contains good characters.

socorro/external/es/crashstorage.py
```diff
@@ -37,6 +37,22 @@ def __init__(self, config):
...
+
+def is_valid_key(key):
+    """Validates an Elasticsearch document key
+
+    :arg string key: the key to validate
+
+    :returns: True if it's valid and False if not
+
+    """
+    return bool(VALID_KEY.match(key))
```
willkg: Doing this as a separate function makes it much easier to test.

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -165,6 +167,29 @@ def test_redact_raw_crash(self):
...
+        'Abc-123-def'
+    ])
+    def test_valid_key(self, key):
+        assert is_valid_key(key) is True
+
+    @pytest.mark.parametrize('key', [
+        '',
+        '.',
+        'abc def',
+        u'na\xefve',
```
willkg: Three of these are cases @milescrabill's script deals with. The one with a space in it I added because it seemed like an interesting case.

@peterbe, @milescrabill: Are there other known-bad cases this is missing that we should add?

socorro/external/es/crashstorage.py
peterbe: 👍 I can't think of any real valid keys with spaces in them.

socorro/external/es/crashstorage.py
```diff
@@ -168,13 +184,39 @@ def reconstitute_datetimes(processed_crash):
...
+
+        This modifies the crash in-place.
+
+        :arg dict raw_crash: raw crash data
+
+        """
+        if not raw_crash:
+            return
+
+        for key in list(raw_crash.keys()):
```
peterbe: If there's a reason to no do `for key in raw_crash:` it should probably be explained with a comment.

socorro/external/es/crashstorage.py
```diff
@@ -168,13 +184,39 @@ def reconstitute_datetimes(processed_crash):
                 # not there? we don't care
                 pass
 
+    @staticmethod
+    def remove_bad_keys(raw_crash):
```
peterbe: I appreciate that we only run this static method for the raw crash (because processed_crash is controlled any won't get crazy) but perhaps not call it `raw_crash` here because this, in fact, works for any mapping.

socorro/unittest/external/es/test_crashstorage.py
peterbe: I can't think of any others.

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -235,6 +260,33 @@ def test_index_crash(self):
...
+        )
+
+        # Ensure that the document was indexed by attempting to retreive it.
+        doc = self.es_client.get(
+            index=self.config.elasticsearch.elasticsearch_index,
+            id=a_processed_crash['uuid']
+        )
+        # Make sure the invalid keys aren't in the crash.
+        raw_crash = doc['_source']['raw_crash']
+        assert raw_crash == {'foo': 'alpha'}
```
peterbe: Nice!

socorro/external/es/crashstorage.py
```diff
@@ -168,13 +184,39 @@ def reconstitute_datetimes(processed_crash):
...
+
+        Anything else is a bad key and needs to be removed.
+
+        This modifies the crash in-place.
+
+        :arg dict raw_crash: raw crash data
+
+        """
+        if not raw_crash:
+            return
```
peterbe: Does this ever happen?

socorro/external/es/crashstorage.py
willkg: It's a good practice in any language to do something like this when you're iterating over a thing and removing things from it. In Python 3, `keys()` returns a *view* of the list of keys, so you want to make a copy of the things you're iterating over before iterating over them and removing things.

I'll add a comment because you asked.

socorro/external/es/crashstorage.py
willkg: The processor is a crazy place and defensive programming seems like a good thing.

willkg:
Things to do here:

* add a comment before the `list(raw_crash.keys())` line explaining why

socorro/external/es/crashstorage.py
peterbe: Ha! This is (negative) effect of code reviewing. I read it line by line and stopped as soon as I saw that line. Had I read 3 more lines down I would have understood the why. 
Just goes to show how it's easy to get caught up in something (that stands out) quickly when reading code. 
Yeah, something like `for key in list(raw_crash.keys()):   # iterate over a *copy* of the keys` would work for me.

willkg:
Thank you!

PR #3913, opened 2017-08-11
fixes bug 1388445 - add crontabber container

willkg:
This adds a Dockerfile, behavioral configuration file, shell script, and other
bits to create a crontabber container.


Makefile
```diff
@@ -67,6 +68,9 @@ dockertestshell:
 dockerdocs:
 	./docker/run_build_docs.sh
 
+dockerupdatedata:
+	./docker/run_update_data.sh
+
```
willkg: I created this make rule in this PR because I tweaked the script to use the crontabber container, but it's not actually related to the crontabber container change.

docker-compose.yml
```diff
@@ -49,6 +49,23 @@ services:
...
+      dockerfile: ./docker/Dockerfile.crontabber
+    image: local/socorro_crontabber
+    env_file:
+      - docker/config/docker_common.env
+      - docker/config/crontabber.env
+    depends_on:
+      - localstack-s3
+      - postgresql
+      - elasticsearch
+      - rabbitmq
```
willkg: I haven't got a clue what services crontabber uses, so I say "use them all!"

milescrabill: @peterbe which does it use? I thought it was just postgres.

docker/run_crontabber.sh
```diff
@@ -0,0 +1,14 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Runs crontabber.
+
+# Run crontabber sleeping 5 minutes between runs
+while true; do
+    python socorro/cron/crontabber_app.py
+    echo "Sleep 5 minutes..."
+    sleep 300
+done
```
willkg: This is a big behavior change from how crontabber runs currently.

Currently, it's kicked off every 5 minutes by cron. So if there's a crontabber running for longer than 5 minutes, a new one will get kicked off and they'll fight to the death and then things keep happening.

With this implementation, we wait for crontabber to finish running, then sleep 300 seconds, then run a new one. So instead of "runs every 5 minutes" it's now more like "runs and then sleeps for 5 minutes".

This is a stupid simple implementation, so that's why I did it this way for now. If the behavioral implications of this aren't ok, then we probably need to write a real scheduler or have it run crontabber in the background (will it capture stdout correctly?) or something like that.

docker/run_crontabber.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
milescrabill: is this a reasonable entrypoint/cmd for a production crontabber container?
I think it should probably be OK, but I'm not sure.

docker/run_crontabber.sh
milescrabill: We will be running the future dockerized production Socorro crontabber with this script, so if there are reasons to write a real scheduler / something else we should consider them.

willkg:
Thank you!

I created https://bugzilla.mozilla.org/show_bug.cgi?id=1389571 to cover figuring out what to do about the behavior change.

docker-compose.yml
peterbe: I think that's fair. It definitely needs S3, elasticsearch and (obviously) postgres. I can't think of any crontabber app that connects to rabbit. 

@milescrabill crontabber itself uses the same postgres connection as we have for stuff like productversions, django data, crashes etc. Basically, it uses postgres as the state store but obviously it imports and runs apps (aka. crontabber jobs) that are written in python and need access to things like elasticsearch (e.g. https://github.com/mozilla-services/socorro/blob/master/socorro/cron/jobs/elasticsearch_cleanup.py)

docker/config/crontabber.env
```diff
@@ -0,0 +1,10 @@
+# Crontabber configuration file covering behavior configuration for
+# crontabber
+
+# NOTE: This file shouldn't contain any environment-specific or secret
+# configuration.
+
+# ----------------------------------------------------------------------------------
+# from socorro/crontabber
+# ----------------------------------------------------------------------------------
+crontabber.class-BugzillaCronApp.days_into_past=0
```
peterbe: Is just just an example or does that need to be set for things to work?
`0` is the default already.

docker/run_crontabber.sh
peterbe: +1 on this
crontabber is smart to know to not run any jobs if it's already in the middle of running something. That's how it determines to exit with 0 or 2 (or 3 in extreme cases). 
So there's no "they'll fight to the death and then things keep happening" :)

I had to check but if any exception happens *within* a crontabber app, it will exit with 0 nevertheless.
See https://github.com/mozilla/crontabber/blob/57c463a521cdd2f9db4ae3c1104559abf88d1607/crontabber/app.py#L840-L852

We should put `set -e` at the top of this file actually. Then if something is wrong with the configuration (e.g. a typo in `crontabber.env` like `jobs=socorro.cron.jobs.bogzilla.BogzillaCronApp`) then it won't just sit there failing start at all over and over.

docker/config/crontabber.env
willkg: I literally pulled all the configuration from -prod and broke it up into behavioral configuration files. I didn't then go through all the keys and see which ones we could remove.

Sounds like a good pass for someone to do.

docker/run_crontabber.sh
willkg: If we add a `set -e`, then the entire crontabber container will die if crontabber stops.

Interestingly, I just tweaked that in this PR:

https://github.com/mozilla-services/socorro/pull/3914/files#diff-9252dc07e44d2505f344c67d658b204eR27

peterbe: It's late Friday but isn't it a good thing that it doesn't continue to run if the configuration is so broken that crontabber can't get up and running. All "runtime exceptions" when executing crontabber apps are all captured and dealt with. 

It's a bit akin to how you start celery. You don't want a celery working to start without saying anything if it can't connect to the service where it gets jobs from. But when it's up and running it doesn't stop running just because a task within raises an exception.

I admit, this is a drive-by comment. I haven't really booted my brain as to how this'll run/work in a local docker dev world.

willkg: I think we should move this conversation to this bug which is related:

https://bugzilla.mozilla.org/show_bug.cgi?id=1389571

PR #3914, opened 2017-08-11
fixes bug 1389218 - redo behavior configuration in docker environment

willkg:
This redoes how behavior configuration works. Previously, docker-compose would
inject the behavior configuration into the docker container environment. It can
do this because it's being run in a working directory that has the socorro code
in it, so the necessary env files exist *outside* the container it's running.

When we go to run these containers in a server environment or any environment
where you don't have the socorro repo as the working directory, then the
behavior configuration needs to be pulled in from *inside* the container.

This fixes the run_ scripts so that if an indicator flag isn't in the
environment, it pulls the appropriate env file in and executes the component in
the new environment with the behavior configuration.

### DIFF
--- a/bin/build_env.py
+++ b/bin/build_env.py
@@ -4,10 +4,14 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-"""Opens the specified env file, pulls in configuration, then executes the specified command in that
+"""Opens specified env files, pulls in configuration, then executes the specified command in the new
 environment.
 
-Usage:
+This can take one or more env files. Env files are processed in the order specified and variables
+from later env files will override ones from earlier env files. Env files must end with ".env" and
+come before the command.
+
+Usage example:
 
     build_env.py somefile.env somecmd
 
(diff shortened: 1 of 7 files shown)


bin/build_env.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python
```
willkg: There are a bunch of tools that we could use, except configman is using identifiers that aren't valid bash identifiers, so none of them work because they're all validating the environment variable keys before throwing them willy-nilly into the environment.

Thus, we had to roll our own. Yay.

docker/run_crontabber.sh
```diff
@@ -6,9 +6,27 @@
...
+
+
 # Run crontabber sleeping 5 minutes between runs
-while true; do
-    python socorro/cron/crontabber_app.py
+while true
+do
+    # FIXME(willkg): We don't want crontabber to exit weird and then have that
+    # kill the container, but this is a lousy thing to do.
+    ${CMDPREFIX} python socorro/cron/crontabber_app.py || true
```
willkg: If we change how this works per bug 1389571, then I think that'll fix this, too.

willkg:
Mmm... We could extend cmdprefix without too much problem. I vote we add that when we need it.

Regarding multiple env files, that's doable, too. We could "cheat" and just slurp up arguments at the beginning that end in `.env`.

willkg:
^^^ That fixes `build_env.py` to support multiple env files. I'll wait for travis and circle to finish, then land it.

willkg:
Thank you!

PR #3915, opened 2017-08-14
bug 1388449 - add add_crashid_to_queue and scaffolding

willkg:
* adds a is_crash_id_valid lib function for verifying crash ids
* adds a scripts/add_crashid_to_queue.py script which is a stub that executes
  functionality in socorro/scripts/add_crashid_to_queue.py
* adds unit tests for new module

### DIFF
--- a/scripts/add_crashid_to_queue.py
+++ b/scripts/add_crashid_to_queue.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
--- a/scripts/add_crashid_to_queue.py
+++ b/scripts/add_crashid_to_queue.py
@@ -1,0 +1,13 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# See socorro/scripts/add_crashid_to_queue.py
+
+import sys
+
+from socorro.scripts import add_crashid_to_queue
+
+
+if __name__ == '__main__':
+    sys.exit(add_crashid_to_queue.main(sys.argv[1:]))
--- a/socorro/lib/ooid.py
+++ b/socorro/lib/ooid.py
@@ -2,8 +2,12 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-# OOID is "Our opaque ID"
+"""
+OOID is "Our opaque ID"
+"""
+
 import datetime as dt
+import re
 import uuid as uu
 
 from socorro.lib.datetimeutil import utc_now, UTC
@@ -95,3 +99,27 @@ def dateFromOoid(ooid):
 
     """
     return dateAndDepthFromOoid(ooid)[0]
+
+
+CRASH_ID_RE = re.compile(r"""
+    ^
+    [a-f0-9]{8}-
+    [a-f0-9]{4}-
+    [a-f0-9]{4}-
+    [a-f0-9]{4}-
+    [a-f0-9]{6}
+    [0-9]{6}      # date in YYMMDD
+    $
+""", re.VERBOSE)
+
+
+def is_crash_id_valid(crash_id):
+    """Returns whether this is a valid crash id
+
+    :arg str crash_id: the crash id in question
+    :arg boolean strict: whether or not to be strict about the throttle character
+
+    :returns: True if it's valid, False if not
+
+    """
+    return bool(CRASH_ID_RE.match(crash_id))
--- a/socorro/scripts/add_crashid_to_queue.py
+++ b/socorro/scripts/add_crashid_to_queue.py
@@ -4,6 +4,7 @@
 
 import argparse
 import os
+import os.path
 
 import pika
 
@@ -13,7 +14,7 @@ from socorro.lib.ooid import is_crash_id_valid
 EPILOG = """
 To use in a docker-based local dev environment:
 
-    python scripts/add_crashid_to_queue.py socorro.normal <CRASHID>
+    scripts/add_crashid_to_queue.py socorro.normal <CRASHID>
 
 To use in -prod:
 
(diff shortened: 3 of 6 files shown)


scripts/add_crashid_to_queue.py
```diff
@@ -0,0 +1,13 @@
...
+
+# See socorro/scripts/add_crashid_to_queue.py
+
+import sys
+
+from socorro.scripts import add_crashid_to_queue
+
+
+if __name__ == '__main__':
+    sys.exit(add_crashid_to_queue.main(sys.argv[1:]))
```
willkg: This is a stub. The reason for making this a stub that calls another module is that this is not in an importable place, so it's much harder to write unittests for functionality in this stub. The easy thing to do is make this a stub and put the functionality in a place that's easier to access for tests.

We want to be testing scripts that are important to our flow so anything that makes that doable is good.

socorro/lib/ooid.py
```diff
@@ -95,3 +99,27 @@ def dateFromOoid(ooid):
...
+def is_crash_id_valid(crash_id):
+    """Returns whether this is a valid crash id
+
+    :arg str crash_id: the crash id in question
+    :arg boolean strict: whether or not to be strict about the throttle character
+
+    :returns: True if it's valid, False if not
+
+    """
+    return bool(CRASH_ID_RE.match(crash_id))
```
willkg: I want to be testing crash ids in other places, too, so I added this lib function.

socorro/scripts/add_crashid_to_queue.py
```diff
@@ -0,0 +1,129 @@
...
+        formatter_class=WrappedTextHelpFormatter,
+        prog='add_crashid_to_queue.py',
+        description='Send crash id to rabbitmq queue for processing',
+        epilog=(
+            'Note: This does not move crash data--crash data must be in the raw_crash location '
+            'on S3 already.\n\n'
+            'Queues:\n\n'
+            '* socorro.normal - normal processing\n'
+            '* socorro.priority - priority processing\n'
+            '* socorro.submitter (-prod only) - sends crash to -stage environment'
```
willkg: This is a repeat of stuff at the top of the file. I'll de-duplicate that.

willkg:
^^^ That fixes the issues I found. This is ready for review.

The stub thing is annoying in the sense that it adds a bunch of extra-ish files. But it makes it convenient to run things if you're expecting them to be in the `scripts/` directory and it makes it possible to write tests. Right now none of the scripts have tests which makes it easy for them to bitrot without anyone noticing. For one-offs, that's fine. For things that are part of our flow, that sets us up for some frustrating mornings.

What do you think? Is it good enough to do for now?

scripts/add_crashid_to_queue.py
peterbe: My only objection is to the folder name "scripts". It should probably be called "bin".
...implying that if you want to put it in `/bin/` instead, I wouldn't object. I think it makes sense to put the "main meat" of the code in "/socorro/scripts/"

scripts/add_crashid_to_queue.py
```diff
@@ -0,0 +1,13 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
peterbe: Why does this not have `#!/usr/bin/env python`?

socorro/scripts/add_crashid_to_queue.py
```diff
@@ -0,0 +1,121 @@
...
+import pika
+
+from socorro.lib.ooid import is_crash_id_valid
+
+
+EPILOG = """
+To use in a docker-based local dev environment:
+
+    python scripts/add_crashid_to_queue.py socorro.normal <CRASHID>
+
```
peterbe: Would be awesome to have a `docker-compose` example here too.

socorro/scripts/add_crashid_to_queue.py
```diff
@@ -0,0 +1,121 @@
...
+        parts = text.split('\n')
+        for i, part in enumerate(parts):
+            parts[i] = super(WrappedTextHelpFormatter, self)._fill_text(part, width, indent)
+        return '\n'.join(parts)
+
+
+def main(argv):
+    parser = argparse.ArgumentParser(
+        formatter_class=WrappedTextHelpFormatter,
+        prog='add_crashid_to_queue.py',
```
peterbe: I'm not familiar with `prog` in argparse but couldn't this be `os.path.basename(__file__)` so we don't repeat the name "twice"?

socorro/unittest/scripts/test_add_crashid_to_queue.py
```diff
@@ -0,0 +1,67 @@
...
+from socorro.scripts.add_crashid_to_queue import (
+    main,
+)
+
+
+def test_missing_args(capsys):
+    # Make sure that running main with no args causes the script to exit with exit_code 2
+    with pytest.raises(SystemExit) as exc_info:
+        main([])
+    assert exc_info.exconly() == 'SystemExit: 2'
```
peterbe: neat! I've never seen this before.

scripts/add_crashid_to_queue.py
willkg: Socorro convention is that scripts are in `scripts/`. I'm inclined to keep that for now. If we want to change it, we can do that in a new bug.

scripts/add_crashid_to_queue.py
willkg: The problem with the shebang thing is that it requires you to be in the right environment such that the correct python is used. The original script this is based on had `#!/data/socorro-virtualenv/bin/python` or some such nonsense which doesn't work in the docker environment.

I left off the shebang in an attempt to make it easier to run these scripts in both the docker and admin node environments.

That problem goes away once we switch to the new docker infrastructure. So this is a "what do we do for now" kind of thing.

I'll re-add the shebang line and make it docker-environment-friendly.

socorro/scripts/add_crashid_to_queue.py
willkg: It's just adding some additional bits to the beginning, but that's true of running anything through docker-compose run. I'm not sure that's prudent here, but might be a good thing for the docs.

socorro/scripts/add_crashid_to_queue.py
willkg: That's a good idea--I'll change that.

socorro/unittest/scripts/test_add_crashid_to_queue.py
willkg: This relies on the repr of the SystemExit exception which shouldn't change ever plus the check seemed pretty readable. Having said that, I'm not wildly excited about comparing against the repr of an exception in general.

willkg:
Thank you!

PR #3916, opened 2017-08-15
Update and clean up some getting started docs

willkg:
I made some changes based on Jonathan's suggestions. The big issue was that the
documentation about running "make dockerupdatedata" was after the link sending
you off to the webapp documentation, so he missed it. This fixes that flow and
adds some more details which are probably helpful to know.

### DIFF
--- a/docs/components/crashstorage.rst
+++ b/docs/components/crashstorage.rst
@@ -11,6 +11,10 @@ Base class implemented in ``socorro/external/crashstorage_base.py``
 
 These are our base classes for all crash storage for Socorro.
 
+.. Warning::
+
+   August 17th, 2017: These docs are outdated.
+
 
 socorro.external.crashstorage_base
 ==================================
--- a/docs/components/crontabber.rst
+++ b/docs/components/crontabber.rst
@@ -22,6 +22,10 @@ The configuration is done using ``configman`` and it looks something like this::
 The default config lives in ``socorro/cron/crontabber_app.py`` as
 ``DEFAULT_JOBS``.
 
+.. Warning::
+
+   August 17th, 2017: Everything below this point is outdated.
+
 
 crontab runs crontabber
 =======================
--- a/docs/components/processor.rst
+++ b/docs/components/processor.rst
@@ -34,6 +34,11 @@ Then you can start and stop the processor and tweak files and all that jazz.
    The stackwalk binaries are in ``/stackwalk`` in the container.
 
 
+.. Warning::
+
+   August 17th, 2017: Everything below this point is outdated.
+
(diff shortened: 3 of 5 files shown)


willkg:
@lonnen I'm tagging you to review this figuring you haven't done much docker stuff with Socorro, so your eyes are newer.

I kind of hate the make rule names. We'll get rid of the "docker" prefix once I ditch vagrant and we can ditch some other things. "build" is a common rule name--and I think that one is fine". "setup" and "updatedata" seem vague and unhelpful.

What do you think about Getting Started? Does it flow well? At the end of it, do you feel like you understand enough of what happened and what you now have to do work?

What do you think about the make rule names?

willkg:
Just to reign this in a bit, the goal for this PR is "make things a bit better" and "identify other things we can do to improve the docs without turning them into an omnibus, but defer the non-trivial ones to a later pass".

lonnen:
Had to install the docker app on OSX and run that. Once I had that down, this was smooth sailing. `make dockerupdatedata` takes a few minutes but that's alright. 

Once I got things running I tried to set up the webapp and had some trouble. It may be beyond the scope here.

1. the compose script outputs `Starting development server at http://0.0.0.0:8000/` but `0.0.0.0` is not an allowed host and you should instead use `localhost`
2. `http://localhost:8000/_debug_login/` is referenced in the docs but in this setup complains that it cannot write or read from memcached.
3. even with updated data tools, the homepage graph does not load and trying to load topcrashers returns a BadArgument 500

Are those out of scope of this change?

willkg:
This is helpful. I'll tweak the docker install section.

Django is saying "starting development server at http://0.0.0.0:8000". That sets it up to listen to all the available ip address. Port 8000 is exported from the container so that's the only one you can access from the host. Thus the message doesn't line up with the address you actually need to connect to. I'm not really sure what we can do about that. It's a problem any project running Django, Jinja2, etc has when used with Docker. Maybe other people have ideas about how to reduce that papercut.

The webapp docs definitely need to be updated. That's a project for a different day. The same can be said for pretty much everything outside of the gettingstarted.rst file.

I'm pretty sure item 3 is an issue because there isn't any crash data. @peterbe was suggesting we might want to fix the homepage such that it works in "no data" scenarios. That's an option. Once I get more processor stuff done, then we can populate the system with crashes and that'd be an additional step. Might be easiest to prime the system with a single "initial crash". I'll toss that in my list of things to think about.

willkg:
I tweaked the getting started tools installation section. I think that's good enough to land this PR.

willkg:
Thank you!

PR #3918, opened 2017-08-15
Add socorro_env script

willkg:
This adds a socorro_env script that launches bash in a modified environment with
the Python virtual environment activated and with all the configuration loaded.

This makes running Socorro scripts much easier.

So instead of:

```
    $ . /data/socorro/socorro-virtualenv/bin/activate
    $ envconsul -prefix some/nonsense mycmd
    $ envconsul -prefix some/nonsense someothercmd
```

You can do:

```
    $ /data/socorro/bin/socorro_env.sh
    (socorro) $ mycmd
    (socorro) $ someothercmd
```

### DIFF
--- a/bin/socorro_env.sh
+++ b/bin/socorro_env.sh
@@ -1,0 +1,30 @@
+#!/bin/sh
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Launches bash in a socorro-ized environment to use on the servers. This
+# includes the Socorro Python virtualenv as well as pulling in consul
+# configuration.
+#
+# This doesn't work in docker and isn't needed.
+#
+# Usage:
+#
+#     /data/socorro/bin/socorro_env.sh
+
+set -e
+
+echo "Activate Python virtualenv..."
+# Active virtualenv in this environment
+. /data/socorro/socorro-virtualenv/bin/activate
+
+echo "Add all the stuff from envconsul..."
+export "$(envconsul -prefix socorro/common -prefix socorro/processor -prefix socorro/webapp -prefix socorro/crontabber env)"
+
+echo "Set prompt..."
+export PS1='(socorro) \u@\h:\w\$ '
+
+echo "Launch bash... (type 'exit' when finished)"
+/bin/bash
--- a/bin/socorro_env.sh
+++ b/bin/socorro_env.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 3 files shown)


willkg:
@peterbe This solves my "we've got two environment to run things in" problem with script documentation. It makes running things on the servers much easier.

What do you think?

willkg:
Is `activate_socorro` a better name?

bin/socorro_env.sh
```diff
@@ -0,0 +1,30 @@
...
+#     /data/socorro/bin/socorro_env.sh
+
+set -e
+
+echo "Activate Python virtualenv..."
+# Active virtualenv in this environment
+. /data/socorro/socorro-virtualenv/bin/activate
+
+echo "Add all the stuff from envconsul..."
+export "$(envconsul -prefix socorro/common -prefix socorro/processor -prefix socorro/webapp -prefix socorro/crontabber env)"
```
willkg: Wha wha what?!?

What this is doing is getting envconsul to pass an environment to env which spits it all out as `key=val` pairs which is then exported as environment variables into the environment this script is running in.

bin/socorro_env.sh
```diff
@@ -0,0 +1,30 @@
...
+#     /data/socorro/bin/socorro_env.sh
+
+set -e
+
+echo "Activate Python virtualenv..."
+# Active virtualenv in this environment
+. /data/socorro/socorro-virtualenv/bin/activate
+
+echo "Add all the stuff from envconsul..."
+export "$(envconsul -prefix socorro/common -prefix socorro/processor -prefix socorro/webapp-django -prefix socorro/crontabber env)"
```
willkg: Wha wha what?!?

What this is doing is getting envconsul to pass an environment to env which spits it all out as `key=val` pairs which is then exported as environment variables into the environment this script is running in.

bin/socorro_env.sh
```diff
@@ -0,0 +1,30 @@
...
+
+echo "Set prompt..."
+export PS1='(socorro) \u@\h:\w\$ '
```
willkg: This sets a helpful `PS1` because that's not inherited from the shell.

bin/socorro_env.sh
peterbe: Neato! 

The description you mentioned in the PR description isn't really fair though. Those are different things. If one of the commands you start with `envconsul` is long-running (or daemon) it'll be restarted if you change any config in any of the connected nodes. This new approach is different in that it won't update once set. So it's only really useful for one-off script executions. 

Perhaps it might be worth, as an optional, nit to remind of that. Either in an `echo` or as a doc string here in this script.

peterbe:
Yes, `activate_socorro` (or `activate_socorro_env`) is a better name. I like that.

bin/socorro_env.sh
willkg: This script is explicitly for making it easier for you and I to run scripts. It's not for running components and I'm not currently planning to change any of the component run scripts. The need for this script goes away when we switch to a Docker-based server infrastructure.

I'll add an echo to make that clear.

peterbe:
Doubleapproved. Go for it.

PR #3920, opened 2017-08-16
fixes bug 1380761 - Deprecate AduBySignature

peterbe:
It occurred to me that I hadn't emailed the Stability list officially (only talked to Calixte and Marco) about this back when the API was marked as deprecated. Or at least I can't find any such email from myself or Adrian. So I sent one today with subject line `Deprecating /api/AduBySignature/`

If it seems strange that I'm deleting a whole file called `crashes.py`, and why was it not called `adu_by_signature.py` or something? It's because that `crashes.py` used to have a bunch of methods related to querying various views and tables from postgres. They were used by the middleware and was one-by-one removed. Now the only thing left was this thing about adu-by-signature.


peterbe:
Note; https://crash-stats.mozilla.com/admin/analyze-model-fetches/ is terrible in that it's not known when the collection started but AduBySignature hasn't been used once. And I suspect this is since we did the last prod upgrade because that flushes memcached.

PR #3922, opened 2017-08-16
Move activate_socorro so it gets picked up in RPM

willkg:


scripts/activate_socorro
```diff
@@ -12,7 +12,7 @@
 #
 # Usage:
 #
-#     /data/socorro/bin/activate_socorro
+#     /data/socorro/scripts/activate_socorro
```
peterbe: Perhaps there's something I just don't get but there's no `/data/socorro/scripts` either. 
```
[centos@ip-172-31-32-25 socorro]$ ls /data/socorro/
application/        contribute.json     socorro-virtualenv/ stackwalk/          webapp-django/
```

willkg: Oops--it's in `/data/socorro/application/`.

PR #3924, opened 2017-08-17
bug 1388449 - add socorro_aws_s3.sh script

willkg:
* add awscli to the processor in a place where it shouldn't affect anything
* create socorro_aws_s3.sh script and friends that wrap aws by pulling
  configuration out of the environment making it easier to use

The end result of this is that we can copy files into and out of the s3
container, see what's in there, manipulate buckets, and so on.

### DIFF
--- a/docker/set_up_awscli.sh
+++ b/docker/set_up_awscli.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Install aws cli as the current user in /tmp so it doesn't interfere with
+# socorro/system requirements
+HOME=/tmp pip install awscli --user
+
+# Fix permissions so anyone can use it
+find /tmp/.local/ -type d -exec '/bin/chmod' '755' '{}' ';'
+find /tmp/.local/ -type f -exec '/bin/chmod' '644' '{}' ';'
+chmod 755 /tmp/.local/bin/*
--- a/scripts/socorro_aws_s3.sh
+++ b/scripts/socorro_aws_s3.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+set -e
+
+# Wrapper for aws script that pulls connection bits from the environment.
+#
+# Usage:
+#
+# make bucket
+#
+#     scripts/socorro_aws_s3.sh mb s3://dev_bucket/
+#
+# list bucket
+#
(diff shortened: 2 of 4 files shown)


docker/set_up_awscli.sh
```diff
@@ -0,0 +1,14 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Install aws cli as the current user in /tmp so it doesn't interfere with
+# socorro/system requirements
+HOME=/tmp pip install awscli --user
+
+# Fix permissions so anyone can use it
+find /tmp/.local/ -type d -exec '/bin/chmod' '755' '{}' ';'
+find /tmp/.local/ -type f -exec '/bin/chmod' '644' '{}' ';'
+chmod 755 /tmp/.local/bin/*
```
willkg: This has a lot of "ugh" in it. The aws cli uses a different version of boto and friends than what we use in Socorro. So I figured I'd "install" it in such a way that it doesn't affect everything else.

The best way I could figure to do that was to install it for a "user", but the app user doesn't have a home directory, so I set `HOME` to `/tmp`. Then it installs stuff in weird places and with permissions that don't let anyone else on the box use it. Thus the permissions fixes.

I'm game for a better way to fix permissions in a directory tree. Particularly the file ones--that's goofypants.

scripts/socorro_aws_s3.sh
```diff
@@ -0,0 +1,65 @@
...
+# that don't
+function getenv {
+    python -c "import os; print os.environ['$1']"
+}
+
+AWS_HOST="$(getenv 'resource.boto.host')"
+AWS_PORT="$(getenv 'resource.boto.port')"
+AWS_ACCESS_KEY_ID="$(getenv 'resource.boto.access_key')"
+AWS_SECRET_ACCESS_KEY="$(getenv 'secrets.boto.secret_access_key')"
+AWS_BUCKET="$(getenv 'resource.boto.bucket_name')"
```
willkg: I figured out a function so this is less gooftastic.

scripts/socorro_aws_s3.sh
```diff
@@ -0,0 +1,65 @@
...
+EOF
+fi
+if [[ ! -f /tmp/.aws/credentials ]]
+then
+    cat > /tmp/.aws/credentials <<EOF
+[default]
+aws_access_key_id = ${AWS_ACCESS_KEY_ID}
+aws_secret_access_key = ${AWS_SECRET_ACCESS_KEY}
+EOF
+fi
```
willkg: Since this is only being used in the local dev environment, I decided to not worry about flexibility of the various ways to get credentials. So I'm adding the required files so it can pull credentials from those.

docker/set_up_awscli.sh
milescrabill: Weeeird. This works, I guess.

scripts/socorro_aws_s3.sh
milescrabill: Nice! And it looks like the docker environment sets all of these things for local dev correctly such that this works.

willkg:
Thank you!

PR #3925, opened 2017-08-17
Bug 1391402 - Add _platform_memmove$VARIANT$Haswell

EricRahm:
`_platform_memmove$VARIANT$Haswell ` is a variant of `memmove` on macOS and should be skipped.


willkg:
Grabbing this to review on Monday. I have to check to see if `$` is ok in that file. It doesn't show up anywhere else and I don't remember offhand how the file is used.

willkg:
I looked into `$` and I think it should be fine. I'm going to merge this and verify it on -stage.

PR #3926, opened 2017-08-18
fixes bug 1391637 - add my.env to override variables in local dev env…

willkg:
…ironment

my.env will work across all components--not just the webapp. Further, variables
defined in it will work for running one-off scripts, too.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,10 +40,17 @@ json_enhancements_pg_extension: bootstrap
 
 DC := $(shell which docker-compose)
 
+my.env:
+	@if [ ! -f my.env ]; \
+	then \
+	echo "Copying my.env.dist to my.env..."; \
+	cp docker/config/my.env.dist my.env; \
+	fi
+
 .docker-build:
 	make dockerbuild
 
-dockerbuild:
+dockerbuild: my.env
 	${DC} build base
 	${DC} build processor
 	${DC} build webapp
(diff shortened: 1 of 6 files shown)


willkg:
@peterbe What do you think? Does this work for you? Is it clear how to use it?

@milescrabill Pretty sure this doesn't affect anything in a server environment, but tagging you for review just in case.

willkg:
Also, some context....

Because we're using python-decouple in the webapp, we can use `webapp-django/.env` already, but that only affects the webapp and doesn't affect the other Socorro components.

We could name it `.env` and put it in the root directory, but docker-compose will look for a `.env` and assume it's for Docker/docker-compose things, so we can't use that.

I figured I'd name it `my.env` since it's clearer that it's specific to individual developers and not general to anyone doing local development.

docker/config/my.env.dist
```diff
@@ -0,0 +1,19 @@
...
+# don't check them in.
+
+# ---------------------------------------------
+# Webapp Google auth settings
+# ---------------------------------------------
+
+# Google auth credentials from setting up a project
+OAUTH2_CLIENT_ID=
+OAUTH2_CLIENT_SECRET=
+
```
peterbe: Why this example? Is it just an example? 
For setting personal secrets and stuff, for the *webapp* you already have the `.env` file for that.

Makefile
```diff
@@ -40,10 +40,17 @@ json_enhancements_pg_extension: bootstrap
 
 DC := $(shell which docker-compose)
 
+my.env:
+	@if [ ! -f my.env ]; \
+	then \
+	echo "Copying my.env.dist to my.env..."; \
+	cp docker/config/my.env.dist my.env; \
+	fi
```
peterbe: Nice!

docs/gettingstarted.rst
```diff
@@ -162,42 +166,82 @@ Run::
...
-``processor.env`` and ``webapp.env``
-    These configuration files hold behavioral configuration for these two things
-    that work across ALL environments--local development and servers.
+``my.env``
+    This file lets you override any environment variables for your local
+    development environment.
 
-    For example, if you want to add a new destination crash store to the system,
-    you'd add it to ``processor.env``.
+    The template for this is in ``docker/config/my.env.dist``.
```
peterbe: I'd love an extra comment about something like "This is your personal file. It's never checked into git.".
Perhaps I'm wrong. Perhaps is "obvious" from the name.

docker/config/my.env.dist
willkg: I want one file that covers all the components. I talk about why I did it this way in the other PR comments already. After reading those, I'm game for discussing it further.

docs/gettingstarted.rst
willkg: I'll improve this. Thank you for pointing it out!

willkg:
I'm going to land this now. If @milescrabill has issues, we can fix those in a future PR. That way I can work on the next step.

PR #3928, opened 2017-08-18
fixes bug 1388449 - implement fetch_crash_data and pull it all together

willkg:
* add scripts/fetch_crash_data.py to fetch data from -prod
* add docker/as_me.sh to automate running one-off scripts in processor container
  with the host's uid/gid so file permissions are ok
* add documentation to processor docs covering how to use the scripts


docker/config/my.env.dist
```diff
@@ -16,4 +16,4 @@ OAUTH2_CLIENT_SECRET=
 
 # API token from crash-stats.mozilla.com that has access to download
 # raw dumps
-API_TOKEN=
+SOCORRO_API_TOKEN=
```
willkg: I renamed this to make it clearer as to what it was an API token for. I'm game for other naming ideas.

docker/run_socorro_aws.sh
```diff
@@ -1,12 +0,0 @@
...
-# This Source Code Form is subject to the terms of the Mozilla Public
-# License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/.
-
-# Convenience wrapper for running socorro_aws.sh script in a container with a
-# uid/gid that will respect file permissions.
-
-HOSTUSER=$(id -u):$(id -g)
-
-docker-compose run -u ${HOSTUSER} processor /app/scripts/socorro_aws_s3.sh $@
```
willkg: We don't need this script anymore since we have `docker/as_me.sh` which is generalized.

socorro/lib/datetimeutil.py
```diff
@@ -143,6 +143,12 @@ def uuid_to_date(uuid, century='20'):
 
 
 class JsonDTEncoder(json.JSONEncoder):
+    """JSON encoder that handles datetimes
+
+    >>> json.dumps(some_data, cls=JsonDTEncoder)
+    ...
+
+    """
```
willkg: Figured I'd add a docstring to this since I use it periodically and always have to look it up.

socorro/scripts/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+    def _fill_text(self, text, width, indent):
+        """Wraps text like HelpFormatter, but doesn't squash lines
+
+        This makes it easier to do lists and paragraphs.
+
+        """
+        parts = text.split('\n')
+        for i, part in enumerate(parts):
+            parts[i] = super(WrappedTextHelpFormatter, self)._fill_text(part, width, indent)
+        return '\n'.join(parts)
```
willkg: I'm using this in two scripts now, so I extracted it out.

willkg:
To review this:

1. read the changes in the documentation first since that explains how all the pieces come together
2. then read through the `fetch_crash_data.py` script -- it should be pretty straight forward
3. then run through it

Here's an example.

First, you need a valid api token that has "View raw dumps" permission. Set that up per the documentation.

After that, do the following from your host computer:

```
# Fetch the data from -prod to your host machine
$ docker/as_me.sh scripts/fetch_crash_data.py ./testdata 5c9cecba-75dc-435f-b9d0-289a50170818

# Create the bucket in your S3 container
$ docker/as_me.sh scripts/socorro_aws_s3.sh mb s3://dev_bucket/

# Sync the contents of ./testdata into the S3 container
$ docker/as_me.sh scripts/socorro_aws_s3.sh sync ./testdata s3://dev_bucket/

# Add the crash id to the processing queue
$ docker/as_me.sh scripts/add_crashid_to_queue.py socorro.normal 5c9cecba-75dc-435f-b9d0-289a50170818

# Run the processor
$ docker-compose up processor
```

That should result in a processed crash! ... Except that there's a bug in the processor or the processor configuration that's preventing it from seeing the S3 container or something along those lines, so it says it can't retrieve the crash and rejects processing it. 

We can verify that the crash contents are in the container:

```
$ docker/as_me.sh scripts/socorro_aws_s3.sh ls s3://dev_bucket/
```

Figure that's outside the scope of this PR, so I'm going to defer figuring that out to a different bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1391771

So... that's everything that this PR facilitates. The command lines are long-ish, but I think they're nicely put together to make it easy to use in scripts covering a variety of use cases.

How do they feel to you? Are there sharp edges that we should fix?

docs/components/processor.rst
```diff
@@ -34,6 +34,114 @@ Then you can start and stop the processor and tweak files and all that jazz.
...
+
+Usage from host::
+
+  $ docker/as_me.sh scripts/fetch_crash_data.py <outputdir> <crashid> [<crashid> ...]
+
+
+For example (assumes this crash exists)::
+
+  $ docker/as_me.sh scripts/fetch_crash_data.py ./testdata 5c9cecba-75dc-435f-b9d0-289a50170818
+
```
peterbe: Why would I want to mess with the directory? Couldn't this be hardcoded to something that's also mentioned in `.gitignore`. I can appreciate that you might want to override it when you're deep in debugging but that could be `fetch_crash_data.py --save-dir=/tmp/other ...`

socorro/scripts/fetch_crash_data.py
```diff
@@ -0,0 +1,137 @@
...
+    fn = os.path.join(outputdir, 'v2', 'raw_crash', crash_id[0:3], '20' + crash_id[-6:], crash_id)
+    create_dir_if_needed(os.path.dirname(fn))
+    with open(fn, 'w') as fp:
+        json.dump(raw_crash, fp, cls=JsonDTEncoder, indent=2, sort_keys=True)
+
+    # Save dump_names to file system
+    fn = os.path.join(outputdir, 'v1', 'dump_names', crash_id)
+    create_dir_if_needed(os.path.dirname(fn))
+    with open(fn, 'w') as fp:
+        json.dump(dumps.keys(), fp)
```
peterbe: Why do we do this even if there are no dump files?
I'm partially curious and partially wondering if it should be explained with a comment. E.g. `# Save the  list of all dump names even if it's empty because ....`

socorro/scripts/fetch_crash_data.py
```diff
@@ -0,0 +1,137 @@
...
+                'format': 'raw',
+                'name': dump_name
+            },
+            headers=headers,
+        )
+
+        if resp.status_code != 200:
+            raise Exception('Something unexpected happened. status_code %s, content %s' % (
+                resp.status_code, resp.content)
+            )
```
willkg: I want the content, too. In the case of errors, we might get an error message that clarifies the issue.

socorro/scripts/fetch_crash_data.py
willkg: I don't know offhand why this is required but this is what the old collector did and antenna does now. I think if we want to change this behavior, we have to change whatever looks at that file.

docs/components/processor.rst
willkg: I think "Why would I want to mess with the directory?" unpacks into two questions:

1. why would I want to have the uid/gid for those files be my user account on my host?
2. why support multiple directories and not just a single one?

Towards the first one, if the uid/gid isn't the user account on your host, then in order to manipulate files in that directory from the host, you have to use sudo and that gets real old real fast. Adding crashes, removing crashes, adjusting raw crash data, and so on.

Towards supporting multiple directories, this makes it much easier to support working on multiple things at the same time. It lets you group crashes for a specific work project. Then you can process them and manipulate them and all that separate from other projects. Also, you can delete them when you're done that project which is pretty important.

I'm not sure I want to have it default to a specific directory because I really want crashes associated with the project/bug that they're being used for so they can be deleted asap. If you want that anyways, I think that's better solved by you having a shell script covering your specific work flow on your machine.

willkg:
> The only thing I don't get is why do I have to run the download, the sync and the processor. ...all manually as separate steps.

The goal of this was to write atoms from which we can build more complex scripts to solve immediate needs. You can use them individually if you like.

When I was sussing out processor configuration issues, I wrote a script that uses them to automate testing:

```
#!/bin/bash

# Make the bucket and sync contents
./scripts/socorro_aws_s3.sh mb s3://dev_bucket/
./scripts/socorro_aws_s3.sh sync ./testdata s3://dev_bucket/
./scripts/socorro_aws_s3.sh ls --recursive s3://dev_bucket/

# Should have in s3 at this point:
# 2017-08-18 20:22:13          0 v1/dump/5c9cecba-75dc-435f-b9d0-289a50170818
# 2017-08-18 20:22:13          8 v1/dump_names/5c9cecba-75dc-435f-b9d0-289a50170818
# 2017-08-18 20:22:13      16353 v2/raw_crash/5c9/170818/5c9cecba-75dc-435f-b9d0-289a50170818

# Add a crash id to the queue
./scripts/add_crashid_to_queue.py socorro.normal 5c9cecba-75dc-435f-b9d0-289a50170818

# Run the processor -- ctrl-c to exit
./docker/run_processor.sh

# Check webapp: http://localhost:8000/report/index/5c9cecba-75dc-435f-b9d0-289a50170818
```

Then I'd iterate with this by tweaking some stuff and re-running the script.

willkg:
Thank you!

PR #3929, opened 2017-08-21
fixes bug 1391771 - fixes processor in local dev environment

willkg:
* fix configuration so processor uses S3 container
* add two more things to run_update_data.sh that maintain the appropriate
  weekly tables
* update localstack to 0.7.4

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -111,10 +111,10 @@ services:
   # https://hub.docker.com/r/localstack/localstack/
   # localstack running a fake S3
   localstack-s3:
-    image: localstack/localstack:0.7.3
+    image: localstack/localstack:0.7.4
     environment:
       - SERVICES=s3:5000
-      - DEFAULT_REGION=us-east-1
+      - DEFAULT_REGION=us-west-2
       - HOSTNAME=localstack-s3
     ports:
       - "5000:5000"
--- a/docker/Dockerfile.processor
+++ b/docker/Dockerfile.processor
@@ -11,6 +11,8 @@ COPY . /app/
 
 USER app
 
+RUN mkdir -p /tmp/symbols/cache
+
 # Install aws cli using --user so it doesn't mess with Socorro requirements
 RUN /app/docker/set_up_awscli.sh
 
--- a/docker/config/docker_common.env
+++ b/docker/config/docker_common.env
@@ -45,19 +45,25 @@ secrets.rabbitmq.rabbitmq_user=rabbituser
 secrets.rabbitmq.rabbitmq_password=rabbitpwd
 
 # elasticsearch things
+
 elasticsearch_url=http://elasticsearch:9200
 resource.elasticsearch.elasticsearch_urls=http://elasticsearch:9200
 resource.elasticsearch.elasticSearchHostname=elasticsearch
 
 # boto things
+
 resource.boto.host=localstack-s3
 resource.boto.port=5000
 resource.boto.secure=false
 resource.boto.access_key=foo
 secrets.boto.secret_access_key=foo
 resource.boto.bucket_name=dev_bucket
+resource.boto.temporary_file_system_storage_path=/tmp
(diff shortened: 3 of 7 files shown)


willkg:
After this lands, the processor should be working, processing crashes, and saving data to the crash storage destinations.

The one thing that I know still isn't working is Elasticsearch crash storage. It "works", but since we don't have any super search fields set up, it's not working in a way that let's you use the webapp to view crashes. There might be some other things that aren't working, but I can't tell without fixing Elasticsearch and then seeing what everything looks like.

I was tempted to figure out how to populate Elasticsearch with the super search fields we use in production, but one of the plans this week is to re-land all the Elasticsearch updates including moving the super search fields from Elasticsearch to a JSON file in the repository. Once we do that, then I think my issue goes away.

Anyhow, if you see other things that are broken that aren't related to missing supersearch fields, let me know.

For instructions on setting up the processor and processing, see pull #3928. It's probably a good idea to land that first, then I can rebase this on top of that, and then review this one since the tools in pull #3928 make this easier to review.

docker-compose.yml
```diff
@@ -111,10 +111,10 @@ services:
   # https://hub.docker.com/r/localstack/localstack/
   # localstack running a fake S3
   localstack-s3:
-    image: localstack/localstack:0.7.3
+    image: localstack/localstack:0.7.4
```
willkg: I'm having difficulties with localstack losing data in the S3 container. Seems like data just disappears even if I'm not shutting down the container. I haven't been able to reproduce it at will, and I have some tools to work around it, so I've been ignoring it.

I thought updating to 0.7.4 might fix it, but it doesn't appear to have. I'll keep an eye on it and if it continues to be problematic, I may switch us to another system like minio or something that persists s3 contents to a volume.

docker/run_update_data.sh
```diff
@@ -8,14 +8,22 @@ set -eo pipefail
...
        --crontabber.class-FeaturedVersionsAutomaticCronApp.products=${PRODUCTS}
 
 # Fetch normalization data for versions we know about
 docker-compose run processor python docker/fetch_normalization_data.py
+
+# Truncate tables
+${CRONTABBER} --job=truncate-partitions
+
+# Create new weekly tables
+${CRONTABBER} --job=weekly-reports-partitions
```
willkg: This one creates `raw_crashes_yyyymmdd` tables which we need for storing crashes. Otherwise the postgres crashstorage fails epically.

willkg:
Oops--this isn't ready, yet. I just saw this in the logs:

```
processor_1      | [2017-08-21 18:57:09,474 pyinotify ERROR] add_watch: cannot watch /mnt/symbols/cache WD=-1, Errno=No such file or directory (ENOENT)
```

I'll fix that and anything else that pops up as I'm verifying the processor work now that `fetch_crash_data.py` PR landed.

docker/Dockerfile.processor
```diff
@@ -11,6 +11,8 @@ COPY . /app/
 
 USER app
 
+RUN mkdir -p /tmp/symbols/cache
```
willkg: This has to exist otherwise pyinotify gets sad.

docker/config/docker_common.env
```diff
@@ -10,9 +10,13 @@
 # specific to the server environment you're setting up.
 
 # alembic
+# -------
```
willkg: I added "header lines" because this file is big and ornery and getting hard to follow. :(

docker/run_update_data.sh
willkg: I'll add a comment, but I'm not sure that we can get rid of this if we ditch Postgres as a crashstorage. We have a bunch of tables that are in the `xyz_yyyymmdd` format and I think crashstorage only covers some of them.

So it's possible, but I think someone should verify that before we go remove things.

willkg:
^^^ That fixes my table problem nicely and turned out to not be that difficult or crappy.

willkg:
Thank you!

PR #3930, opened 2017-08-21
fixes bug 1390168 - add defensive programming to JitCrashCategorizeRule

willkg:
* tweak the code so it's less fragile regarding what's in json_dump
* rewrite the original test_no_crashing_thread test so it's more minimal
* add a bunch more tests covering other data scenarios, too


socorro/processor/breakpad_transform_rules.py
```diff
@@ -666,13 +666,9 @@ def _predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
...
 
-        if (
-            'crashing_thread' in processed_crash.json_dump and
-            processed_crash.json_dump['crashing_thread']['frames'][0].get(
-                'module',
-                False
-            )
-        ):  # there is a module at the top of the stack, we don't want this
+        frames = processed_crash.get('json_dump', {}).get('crashing_thread', {}).get('frames', [])
+        if frames and frames[0].get('module', False):
```
willkg: That was in the original code.

willkg: I can change it--I don't have an opinion either way. This is one of the things that will get cleaned up when I rework this rule with getitem notation and a helper library for that.

peterbe: Best just leave it then.

PR #3931, opened 2017-08-21
fixes bug 1392236, 1389217 - ValueError in enhance_json_dump

peterbe:
This patch does fix https://bugzilla.mozilla.org/show_bug.cgi?id=1392236 *and* https://bugzilla.mozilla.org/show_bug.cgi?id=1389217 in that it also updates the extra vcs mapping. 

Ultimately what the patch does is...

1. Coping with the fact that the frame's `file` might not necessarily reference a repository server that has a `/` in it. 
2. If the frame's `file` contains a reference to file which is a 128 character long sha, it removes that from what is displayed in the report index page. 

The new unit test is based on copying a real value from a real crash and then making sure it passes. 

I'm still on the fence about depending on a new web service but I'll file a dependent bug about that.


willkg:
We talked on IRC and @peterbe is going to adjust this to pull in the source-highlight app as a new webapp view. I'll wait to review to see how that pans out.

peterbe:
I wouldn't mind landing this first and then tackle the new Django pygments view as a separate PR. 
That way we have a landed solution that unbreaks prod's report indexes. If there's delays or problems with landing the new webapp view, at least we'll have solved the more immediate problem.

willkg:
I'm going to write the view and push it to this branch. Pretty sure it's like 15 minutes of work. Then we don't have to keep talking about thinking about doing it. If I hit issues, we can adjust the plan accordingly.

willkg:
@peterbe ^^^ That throws together a highlight view. What do you think? Is the url and view name ok? Do we need to wrap it in any decorators to handle CSP or other concerns?

It's lacking tests. Figured I'd run it by you first.

To test:

1. get the webapp going
2. go to http://localhost:8000/source/generated/?url=https://gecko-generated-sources.s3.amazonaws.com/318c1e8eb64b13914272e71f5c084728437cbe739235e89199ff606667d7f6a80e6e65e1da4a226a24737b4c03cdba307fc820c1a0222e929eb4906aba073e48/dom/bindings/AudioNodeBinding.cpp#l-309

Then you can play with the url and get the 404 and 403 variations.

peterbe:
@willkg updated. Added a light test and moved it into its own app. r?

webapp-django/crashstats/crashstats/tests/test_utils.py
```diff
@@ -146,6 +147,62 @@ def test_enhance_frame(self):
...
+            'offset': '0x108b7b0bd',
+            'short_signature': 'mozilla::layers::PCompositorBridgeChild::OnMessageReceived',
+            'signature': (
+                'mozilla::layers::PCompositorBridgeChild::OnMessageReceived(IP'
+                'C::Message const&)'
+            ),
+            'trust': 'cfi'
+        }
+        # Remember, enhance_frame() mutates the dict.
+        frame = copy.copy(original_frame)
```
willkg: Do you need a deepcopy because you've got nested dicts?

webapp-django/crashstats/crashstats/tests/test_utils.py
```diff
@@ -146,6 +147,62 @@ def test_enhance_frame(self):
...
+        ok_(frame['source_link'])
+        eq_(
+            frame['source_link'],
+            'https://example.com/36d62ce2ec2925f4a13e44fe534b246c23b4b3d540788'
+            '4d3bbfc9b0d9aebe4929985935ae582704c06e994ece0d1e76528ff1edf4543e4'
+            '00d0aaa8f7251b15ca/ipc/ipdl/PCompositorBridgeChild.cpp#l-1495'
+        )
+        # And that links text is the frame's 'file' but without the 128 char
+        # sha.
+        eq_(frame['file'], 'ipc/ipdl/PCompositorBridgeChild.cpp')
```
willkg: This test is fantastic!

willkg:
This doesn't change the CSP settings. Because pygments embeds the style in a `<style>` on the page, don't we have to do something like add `unsafe-inline`?

willkg:
Maybe we can adjust the `HtmlFormatter` so it's not adding style information and we put that in a separate `.css` file?

http://pygments.org/docs/formatters/

peterbe:
unsafe-inline is already in place. I.e. we have a bunch of existing `<style type="text/css>` and that's ok.

peterbe:
See `curl -v https:/crash-stats.m.c ...`

PR #3932, opened 2017-08-22
fixes bug 1392567 - change 1/l to L

willkg:


webapp-django/crashstats/sources/views.py
```diff
@@ -50,5 +50,5 @@ def highlight_url(request):
     else:
         lexer = get_lexer_for_filename(filename)
 
-    formatter = HtmlFormatter(full=True, title=parsed.path, linenos='table', lineanchors='1')
+    formatter = HtmlFormatter(full=True, title=parsed.path, linenos='table', lineanchors='L')
```
willkg: The lineanchor here was set to `1` (one), but should have been `l` (lower-case ell) which is my fault. @luser, after spending some time noticing the problem and then hunting for the cause, suggested that we change it from an `l` (lower-case ell) to an `L` (upper-case ell) which I thought was a pretty swell idea.

willkg:
To review, skim it, wait for Travis to pass, then run the webapp and go to this link:

http://localhost:8000/sources/highlight/?url=https://gecko-generated-sources.s3.amazonaws.com/318c1e8eb64b13914272e71f5c084728437cbe739235e89199ff606667d7f6a80e6e65e1da4a226a24737b4c03cdba307fc820c1a0222e929eb4906aba073e48/dom/bindings/AudioNodeBinding.cpp#L-309

Now the line anchor should bring the browser directly to the line in question.

PR #3933, opened 2017-08-22
fixes bug 1391402 - $ needs to be escaped in prefix list

willkg:


willkg:
This fixes the line in the prefix list so that the `$` are escaped.

You can do this to verify it:

```
(container) $ python
>>> import re
>>> from socorro import siglists
>>> regex = re.compile('|'.join(siglists.PREFIX_SIGNATURE_RE))
>>> regex.match('_platform_memmove$VARIANT$Haswell')
<_sre.SRE_Match object at 0x7f2fe6957108>
```

I wish I had gotten further with my signature generator cli--would have made this way easier to verify.

willkg:
Thank you!

PR #3934, opened 2017-08-22
Fix stackwalker execution

willkg:
This fixes things so that stackwalker runs successfully

### DIFF
--- a/docker/Dockerfile.processor
+++ b/docker/Dockerfile.processor
@@ -12,6 +12,7 @@ COPY . /app/
 USER app
 
 RUN mkdir -p /tmp/symbols/cache
+RUN mkdir -p /tmp/symbols/tmp
 
 # Install aws cli using --user so it doesn't mess with Socorro requirements
 RUN /app/docker/set_up_awscli.sh
--- a/docker/config/docker_common.env
+++ b/docker/config/docker_common.env
@@ -77,8 +77,6 @@ resource.boto.resource_class=socorro.external.boto.connection_context.HostPortS3
 # processor
 # ---------
 
-minidump_stackwalk_pathname=/stackwalk/stackwalker
-
 # store symbols in /tmp/symbols -- a environment server would want to store
 # them in a volume that's shared across processor nodes
 symbol_cache_path=/tmp/symbols/cache
--- a/docker/config/docker_common.env
+++ b/docker/config/docker_common.env
@@ -77,12 +77,12 @@ resource.boto.resource_class=socorro.external.boto.connection_context.HostPortS3
 # processor
 # ---------
 
-# store symbols in /tmp/symbols -- a environment server would want to store
-# them in a volume that's shared across processor nodes
-symbol_cache_path=/tmp/symbols/cache
+# In the docker local dev environment, we store symbol cache and other things in /tmp because
+# there's only one processor node. For server environments, we probably want to store that
+# in a volume. These three vars are all affected.
 companion_process.symbol_cache_path=/tmp/symbols/cache
-
-processor_symbols_pathname_list=/tmp/symbols/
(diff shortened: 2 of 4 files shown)


docker/Dockerfile.processor
```diff
@@ -12,6 +12,7 @@ COPY . /app/
 USER app
 
 RUN mkdir -p /tmp/symbols/cache
+RUN mkdir -p /tmp/symbols/tmp
```
willkg: stackwalker requires these two directories to exist before it runs. Otherwise it says something singularly pithy like "symbols-cache and symbols-tmp must be on the same filesystem".

docker/config/docker_common.env
```diff
@@ -77,8 +77,6 @@ resource.boto.resource_class=socorro.external.boto.connection_context.HostPortS3
 # processor
 # ---------
 
-minidump_stackwalk_pathname=/stackwalk/stackwalker
```
willkg: `BreakpadStackwalkerRule` uses this, but `BreakpadStackwalkerRule2015` does not. When I added this, I was confused by which rule we were using. So we can nix this line.

docker/config/docker_common.env
```diff
@@ -77,8 +77,6 @@ resource.boto.resource_class=socorro.external.boto.connection_context.HostPortS3
...
-
 # store symbols in /tmp/symbols -- a environment server would want to store
 # them in a volume that's shared across processor nodes
 symbol_cache_path=/tmp/symbols/cache
```
willkg: Actually, now that I'm looking at this, this line is also moot. I'll fix that.

docker/Dockerfile.processor
luser: FYI that comes from here: https://github.com/mozilla-services/socorro/blob/1faba7a9c861cc71f4ff79d5909dbf064d156525/minidump-stackwalk/common.h#L69

It's intended to make sure that you don't put them on separate filesystems because the http fetching code fetches symbol files to a temp path then calls `rename` to move them to the intended name, and that doesn't work across filesystem boundaries. Stupidly, that code doesn't separately check that the directories *exist*, so if a call to `stat` fails (as it would if the directory didn't exist) it prints out the "not on the same filesystem" error. We could fix that to either spit out a more useful error in that case, or just create the directories that are passed in.

willkg: That makes a lot of sense. I hadn't looked at the stackwalker code, yet.

We could fix it, but I don't think it's necessary. One of the things that made this hard is that the `BreakpadStackwalkerRule2015` doesn't say anything about `stackwalker` execution. It's completely silent in all the error situations I've seen so far which is unhelpful. If you see something, log something. I'll probably fix that at some point.

Thank you!

willkg:
Yay! I can process crashes locally and get real signatures!

willkg:
To test this out:

1. fetch a crash with id `<CRASHID>` into `<DATADIR>`
2. `docker/as_me.sh ./scripts/socorro_aws_s3.sh sync <DATADIR> s3://dev_bucket/`
3. `docker/as_me.sh ./scripts/add_crashid_to_queue.py socorro.normal <CRASHID>`
4. `docker-compose up processor`

That'll process the crash. Takes a while. After the crash has saved:

1. `docker/as_me.sh ./scripts/socorro_aws_s3.sh sync s3://dev_bucket/ <DATADIR>`
2. `cat < <DATADIR>/v1/processed_crash/<CRASHID> | jq .[signature]`

It should spit out a real signature for you.

willkg:
^^^ That adds `scripts/process_and_quit.sh` which will process whatever crashes are in the queue and then quit. That makes it much easier to write test/verify/automation scripts.

willkg:
Per conversation with @peterbe, I'm going to self-r+ this and land it.

PR #3935, opened 2017-08-22
fixes bug 1342081 - Upgrade elasticsearch to support ES 5.3 (take II)

peterbe:
This redoes https://github.com/mozilla-services/socorro/pull/3676
I couldn't use the revert (on the revert) because of the reason *"Sorry, this pull request couldn’t be reverted automatically. It may have already been reverted, or the content may have changed since it was merged. "* 
So I took the diff, applied it manually. But I had to fiddle a change to `socorro/lib/util.py` because of a pep8 cleanup. Also, I got a pyflake error with this so I fixed that too.


willkg:
I think we want Travis to pass before we land this.

Everything looks ok to me, however, this doesn't update the docker environment at all. I think we should do that in a subsequent PR and not this one.

willkg:
Oops--my bad. This does update the base image in `docker-compose.yml`. That might be fine. I'll check to see if there's anything else that needs updating.

willkg:
I'm +1 on this.

PR #3938, opened 2017-08-23
 fixes bug 1381877 - ignore winsock_lsp gt 30000

willkg:
The winsock_lsp field is being indexed using the keyword analyzer. That's fine,
except this field value is pretty big and exceeds the maximum size.

This changes that to ignore any values larger than 30000 in size.

### DIFF
--- a/socorro/external/es/data/super_search_fields.json
+++ b/socorro/external/es/data/super_search_fields.json
@@ -3042,7 +3042,8 @@
             "type": "text",
             "fields": {
                 "full": {
-                    "type": "keyword"
+                    "type": "keyword",
+                    "ignore_above": 30000
                 }
             }
         },
--- a/socorro/external/es/data/super_search_fields.json
+++ b/socorro/external/es/data/super_search_fields.json
@@ -80,7 +80,8 @@
         "query_type": "enum",
         "permissions_needed": [],
         "storage_mapping": {
-            "type": "keyword"
+            "type": "keyword",
+            "ignore_above": 10922
         },
         "namespace": "raw_crash",
         "has_full_version": false,
@@ -132,7 +133,8 @@
             "type": "text",
             "fields": {
                 "full": {
-                    "type": "keyword"
+                    "type": "keyword",
+                    "ignore_above": 10922
                 }
             }
         },
@@ -156,7 +158,8 @@
             "type": "text",
             "fields": {
                 "full": {
-                    "type": "keyword"
+                    "type": "keyword",
+                    "ignore_above": 10922
                 }
             }
         },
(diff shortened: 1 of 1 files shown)


willkg:
Documentation for this is here:

https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html

Without this, the crash fails to save in ES. With this, the crash saves in ES.

willkg:
@peterbe Link to `ignore_above` is in https://github.com/mozilla-services/socorro/pull/3938#issuecomment-324459917 . I think that covers questions about what it's doing.

Regarding how people use `Winsock_LSP` or why indexing for it is implemented the way it's implemented, I did some poking around but came up with nothing.

I don't see any bugs with "winsock_lsp" in them:

https://bugzilla.mozilla.org/buglist.cgi?quicksearch=product%3ASocorro%20status%3Aany%20winsock_lsp

The `Winsock_LSPRule` processor rule comes from commit 3ed296c76a, but there's nothing in the commit about it and nothing in the rule docstring, either.

Beyond that, I'm not sure how to figure that out. I was kind of hoping you or @adngdb knew.

willkg:
@milescrabill pointed out some issues. I'm going to make some changes to this PR.

willkg:
^^^ That updates the PR to cover all the keywords fields.

Regarding whether we should ditch `Winsock_LSPRule` or change it in some other way, I think that's worth spinning off into a different bug.

The one thing I'm not sure of is whether we should address these field lengths in the mapping using `ignore_above` which drops the value entirely if it's too long, or whether we should truncate the values in Python-land such that we at least get some data in ES.

@peterbe, @milescrabill Thoughts on that?

milescrabill:
The only fields that we saw hitting these limits were mostly garbage, but in general truncation makes sense. If I'm right re: mostly garbage, I'd say just drop them.

peterbe:
1. Regarding the de-duplication in the processed crash, let's continue that in https://bugzilla.mozilla.org/show_bug.cgi?id=1393416
2. Winsock_LSP is rarely useful I don't think. I can't find the comment where that was pointed out but I doubt many people Super Search search for it. It will always be available in the report index page when people do some deep debugging of a crash. ...Meaning, I'm with @milescrabill that if can go ahead with the `ignore_above` for those extreme few crashes where the string is nuts (ie. super long). 

I did some poking around and [the median size of that field is 1,760bytes](https://docs.google.com/a/mozilla.com/spreadsheets/d/13pPj9MUAdGc9xYpacRvdSFkOC0mD1NUl8w6izpbmjpQ/edit?usp=sheets_home&ths=true) and the median is pretty close to the mean. Meaning (no pun intended) that the oddball huge ones are rare.

willkg:
^^^ That's the exact same code, but I spruced up the commit message so as to capture everything we observed and why we think this is ok and what we can do later if we decide it's not.

Travis/Circle should be fine with that. I'll land it after.

willkg:
Thank you!

PR #3940, opened 2017-08-24
fixes bug 1393472 - KeyError on 'caused_by' in RequestError handling

peterbe:
This took forever to get my head around. In hindsight, so obvious!

Some `RequestError` exception objects contain a `exception['error']['caused_by']`. Some don't. I *think* all `RequestError` exceptions contain a `exception['error']['root_cause']` which is a list. Either way, I don't want to mess with that. The original code here is all about trying to find if possible you get a `parse_exception` cause and if so change the whole exception to a `BadArgumentError`.

The crux is that some if you don't raise the **original** exception, and just do a bare `raise` you will actually raise the last exception that happened. Even if it was swallowed, since it's still at the top of the stack. 

What this PR does is that it basically lets all other `RequestError` exceptions **bubble up correctly**.


peterbe:
I don't know how else to reproduce this. I switched back to the SSH tunnel to be able to use the Stage ES5 cluster. Then I hit this URL: `http://localhost:8000/signature/summary/?product=Firefox&version=55.0.2&signature=OOM%20%7C%20small&date=%3E%3D2017-08-17T09%3A29%3A05.000Z&date=%3C2017-08-24T09%3A29%3A05.000Z` and waited. 

The real error is this by the way:
```
  File "/opt/boxen/homebrew/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/wsgiref/handlers.py", line 85, in run
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/signature/views.py", line 65, in inner
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/signature/views.py", line 432, in signature_summary
    search_results = api.get(**params)
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/supersearch/models.py", line 230, in get
    return super(SuperSearch, self).get(**kwargs)
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/crashstats/models.py", line 333, in get
    return self._get(expect_json=expect_json, **kwargs)
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/crashstats/models.py", line 391, in _get
    expect_json=expect_json,
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/crashstats/models.py", line 174, in inner
    result = method(*args, **kwargs)
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/webapp-django/crashstats/crashstats/models.py", line 264, in fetch
    result = implementation_method(**params)
  File "/Users/peterbe/dev/MOZILLA/SOCORRO/socorro/socorro/external/es/supersearch.py", line 518, in get
    raise e
RequestError: TransportError(400, u'illegal_argument_exception', u'ReleasableBytesStreamOutput cannot hold more than 2GB of data')
```

socorro/external/es/supersearch.py
```diff
@@ -511,7 +511,11 @@ def get(self, **kwargs):
                         # The reason cannot be found in the exception info, or
                         # we cannot find anything while parsing the exception.
                         pass
-                raise
+                # If we don't raise the original exception explictly we run we
+                # would actually raise other (even swallowed) exceptions that
+                # might have happened during this block here were we peek
+                # into the exception object.
+                raise e
```
willkg: This hoses the exception context.

A better way to deal with this is at the beginning of this exception handling block, you should capture `sys.exc_info()`. Then raise those things. For example, we do that here:

https://github.com/mozilla-services/socorro/blob/059d82c6f03603ac6ad39d014fbb99b85618820a/socorro/processor/processor_app.py#L144-L185

Pretty sure that handles your needs here while preserving the exception context correctly.

peterbe: Good catch! Thanks!

willkg:
@peterbe says this isn't related to ES 5. i'll look through it on Monday.

PR #3943, opened 2017-08-28
bug 1390847 - remove update-crash-adu-by-build-signature as default job

peterbe:
Gently removing all remnants of this now deprecated stored procedure. 
Earlier today [I removed it from consulate `socorro/crontabber/crontabber.jobs`](https://bugzilla.mozilla.org/show_bug.cgi?id=1390847#c3).

I want to make sure it not running in -prod admin any more doesn't break anything. 
After this I can start to really remove all its code.


willkg:
As a side note, this looks good, but doesn't affect -stage or -prod at all since they both set the jobs in configuration. This only affects the local dev environment.

peterbe:
Yep. The job has (once this lands) been "config'ed away" from default list, from stage consulate and from prod consulate.

PR #3946, opened 2017-08-29
bug 1100354 - fix fields mutations

willkg:
Super search fields come from a Python module now and are handed around and some
of the users mutate the value. Because of that, we want to hand around a copy of
the structure--not the structure itself. This fixes that.

Additionally, this cleans up some comments that were wrong because we're not
storing the fields in JSON anymore.

### DIFF
--- a/webapp-django/crashstats/supersearch/models.py
+++ b/webapp-django/crashstats/supersearch/models.py
@@ -1,3 +1,4 @@
+import copy
 import functools
 
 from socorro.external.es import query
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/supersearch/models.py
```diff
@@ -232,15 +233,15 @@ def get(self, **kwargs):
...
+    # Read it in once as a class attribute since it'll never change unless the
+    # Python code changes and if that happens you will have reloaded the
+    # Python process.
     _fields = SuperSearchFieldsWithoutConfig().get()
 
     API_WHITELIST = None
 
     def get(self):
-        return self._fields
+        return copy.deepcopy(self._fields)
```
willkg: This isn't a huge datastructure, but there's always a penalty for a deepcopy. I don't know how big a difference it'll make. We should notice it in New Relic.

If it turns out to be a big problem, then someone can look at fixing the users of the data to copy before adjusting and we can remove this copy. Or something along those lines.

willkg:
Thank you for catching these!

PR #3947, opened 2017-08-29
fixes bug 1393533 - remove collectstatic before running webapp

willkg:
For local development, the webapp runs with DEBUG=True which sources static
files without needing to run collectstatic. Thus running collecstatic is costly
and not helpful, so we should nix it.


willkg:
Thank you!

PR #3950, opened 2017-08-29
fixes bug 1394800 - Remove gccrashes table

peterbe:
It's another reminder that we don't really have a way of playing back the whole migration history from an initial snapshot (like Django can). 

There are old migrations in there that depend on loading `.sql` files (in the `raw_sql/procs/` directory) which means that if you change any of those `.sql` files you mess up the history playback'ability. 

Also, I couldn't find a single migration where it rolled back a stored procedure **without** a `.sql` file. After all, the `downgrade` isn't really "important". It's there for CI testing. If we decide, against all odds, that we want the gccrashes (tables and stored procedures) back we'd make another PR that reinstates the `.sql` files.


alembic/versions/afc2f95a298b_remove_gccrashes.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def upgrade():
+    op.execute(
+        'DROP FUNCTION update_gccrashes(date, boolean, interval)'
+    )
+    op.execute(
+        'DROP FUNCTION backfill_gccrashes(date, interval)'
+    )
+    op.drop_table('gccrashes')
```
willkg: Like the other PR, I think we should drop the table and the functions with sql and an IF EXISTS part.

peterbe: will do.

PR #3951, opened 2017-08-29
bug 1361764 - update lib/ to pytest [no rush]

willkg:


socorro/unittest/lib/test_datetimeutil.py
```diff
@@ -27,172 +30,155 @@ def test_string_to_datetime():
...
+    ('2012-01-10T12:13:14.98765Z', UTC),
+    ('2012-01-10 12:13:14', UTC),
+    ('2012-01-10 12:13:14.98765', UTC),
+    ('2012-01-10 12:13:14.98765+03:00', PLUS3),
+    ('2012-01-10 12:13:14.98765Z', UTC),
+])
+def test_string_datetime_with_timezone_variations(ts, timezone):
+    res = datetimeutil.string_to_datetime(ts)
+    # NOTE(willkg): isodate.tzinfo.FixedOffset doesn't define __eq__, so we compare the
+    # reprs of them. :(
```
willkg: I don't know why it doesn't define an `__eq__`, but that's mightily annoying.

socorro/unittest/lib/test_external_common.py
```diff
@@ -19,83 +19,83 @@ def test_check_type(self):
...
 
         # Test 3: integer
         param = "12"
         datatype = "int"
         res = external_common.check_type(param, datatype)
 
-        ok_(isinstance(res, int))
-        eq_(res, 12)
+        assert isinstance(res, int)
+        assert res == 12
```
willkg: This is redundant. I can take out a bunch of these type check lines.

willkg:
I did one commit per file figuring that might make it easier to break up the review work.

socorro/unittest/lib/test_datetimeutil.py
```diff
@@ -27,172 +30,155 @@ def test_string_to_datetime():
...
-        (datetime.datetime(2015,  1,  8), '20150105'),
-        (datetime.datetime(2015,  1,  9), '20150105'),
-    )
-    for to_be_tested, expected in proposed_and_expected:
-        result = datetimeutil.datestring_to_weekly_partition(to_be_tested)
-        eq_(result, expected)
+    assert partition == partition_exp
+
+
+@pytest.mark.parametrize('data, expected', [
```
lonnen: huh. I had to look up parametrize and it is absolutely the right thing here

socorro/unittest/lib/test_threaded_task_manager.py
```diff
@@ -132,15 +124,9 @@ def insert_into_list(anItem):
...
-                % len(ttm.thread_list))
-            ok_(len(my_list) == 5,
-                'expected to do 5 inserts, '
-                'but %d were done instead' % len(my_list))
-            ok_(sorted(my_list) == range(5),
-                'expected %s, but got %s' % (range(5),
-                                             sorted(my_list)))
+            assert len(ttm.thread_list) == 2
+            assert len(my_list) == 5
+            assert sorted(my_list) == range(5)
```
lonnen: +1

lonnen: removing the text is fine. With or without it I have to go lookup this part of the test suite when things fail.

socorro/unittest/lib/test_transform_rules.py
```diff
@@ -4,26 +4,15 @@
...
 from socorro.lib import transform_rules
 from socorro.unittest.testbase import TestCase
 
 
-def assert_expected(actual, expected):
-    assert actual == expected, "expected:\n%s\nbut got:\n%s" % (str(expected),
-                                                                str(actual))
-
-
-def assert_expected_same(actual, expected):
```
lonnen: curious. I'm surprised these were used in a limited scope and not removed or spread more widely

socorro/unittest/lib/test_vertools.py
```diff
@@ -2,29 +2,24 @@
...
+    ('3', [3, phs, 0, phs] + pl * 3, '3'),
+    ('3.', [3, phs, 0, phs] + pl * 3, '3'),
+    ('3.0', [3, phs, 0, phs] + pl * 3, '3'),
+    ('3.0.0', [3, phs, 0, phs] + pl * 3, '3'),
+    ('3.5', [3, phs, 0, phs, 5, phs, 0, phs] + pl * 2, '3.5'),
+    ('3.5pre', [3, phs, 0, phs, 5, 'pre', 0, phs] + pl * 2, '3.5pre'),
+    ('3.5b3', [3, phs, 0, phs, 5, 'b', 3, phs] + pl * 2, '3.5b3'),
+    ('3.6.4plugin3', [3, phs, 0, phs, 6, phs, 0, phs, 4, 'plugin', 3, phs] + pl, '3.6.4plugin3'),
+])
+def test_normalize(ver, expected, ver2):
```
lonnen: concise. I dig it.

socorro/unittest/lib/test_threaded_task_manager.py
willkg: Oh--but that's the magic of pytest. pytest will show you the assertion error, then walk the stack and show you the line, the values, where the values came from, and some other information. But it only shows you that information if there's no assertion message.

Check out the top of https://docs.pytest.org/en/latest/ for a basic example.

`eq_` and `ok_`, would generate the assertion message for you, but they didn't walk the stack and show you as much information as pytest does.

socorro/unittest/lib/test_transform_rules.py
willkg: Sometimes I feel like I'm digging through eras of Socorro conventions.

socorro/unittest/lib/test_datetimeutil.py
willkg: parametrize is super awesome because each individual iteration becomes a separate test case which can be executed specifically.

Having said that, it's sometimes tricky to figure out what the id is based on the data.

So I wrote pytest-wholenodeid which helps. https://pypi.python.org/pypi/pytest-wholenodeid

socorro/unittest/lib/test_vertools.py
willkg: I don't really understand this test and certainly couldn't debug it without fleshing out some of the values. Now that I'm looking at it, `ver2` isn't used at all.

I tossed around auditing the tests more thoroughly. Are they readable? Do they test what we think they're testing? But that's a lot of work and I'm not sure we get much value out of it right now.

willkg:
Thank you!

PR #3953, opened 2017-08-30
fixes bug 1395163 - right link to supersearch fields missing in admin home

peterbe:


webapp-django/crashstats/manage/jinja2/manage/home.html
```diff
@@ -56,15 +56,6 @@
...
-
-        <h3>Configuration</h3>
-        <ul>
-            <li><a href="{{ url('manage:supersearch_fields') }}">
-                <span class="fa fa-th-list fa-5x"></span>
-                Super Search Fields
-            </a></li>
-        </ul>
-
         <h3>Tools</h3>
```
peterbe: So, the whole bug was basically that 1) there was two links called "Super Search Fields" on the admin home page and 2) the view `supersearch_fields` still existed.

webapp-django/crashstats/manage/jinja2/manage/supersearch_fields_missing.html
```diff
@@ -47,7 +47,8 @@
                 </td>
             </tr>
         {% endfor %}
+        </table>
```
peterbe: We must have accidentally lost this in the flurry of recent changes. You'd only notice if you open the page. The effect was kinda amusing. The page footer ended up appearing above the table.

PR #3954, opened 2017-08-31
fixes bug 1394909 - Remove all signature_summary_* tables

peterbe:
cc @adngdb 

One concern I have is that this might time out. 
It's going to have to try to delete [64GB](https://bugzilla.mozilla.org/show_bug.cgi?id=1394909#c2) of data in one transaction. I can't think of any other approach other than "Let's just try and if it doesn't work we'll have to think of a way to break this apart".
If it does time out, the migration will never succeed and we can go back and rewrite the migration so that it does multiple things in separate migrations or transactions. 

Also, what's the point of the `downgrade` here? It's only really there for the sake of integration testing the migration. If we go ahead with this and later decide "My bad! We need these tables back" then isn't it kinda too late?


alembic/versions/07c6633fa1b6_delete_signature_summaries.py
```diff
@@ -0,0 +1,82 @@
...
+        op.execute('DROP TABLE IF EXISTS {}'.format(table_name))
+
+    # Now delete all stored procedures
+    op.execute("""
+        DROP FUNCTION IF EXISTS
+        update_signature_summary_architecture(date, boolean)
+    """)
+    op.execute("""
+        DROP FUNCTION IF EXISTS
+        update_signature_summary_device(date, boolean)'
```
willkg: This and the later ones all have a `'` at the end that needs to be removed.

willkg: I'm pretty sure these get tested if you run `make dockertest`.

peterbe: I ran the migration in bash with alembic in docker. All worked. Then I made the PR, realized I forgot the "IF EXISTS" because reasons reasons. Then when I added that I used a macro to edit multiple lines and screwed it up and noticing that. Then I screwed up correcting my typo.

socorro/external/postgresql/raw_sql/procs/backfill_matviews.sql
```diff
@@ -102,10 +87,6 @@ WHILE thisday <= lastday LOOP
     PERFORM backfill_graphics_devices(thisday);
     RAISE INFO 'crash_adu_by_build_signature';
     PERFORM backfill_crash_adu_by_build_signature(thisday);
-    FOR tablename in SELECT name from temp_signature_summaries LOOP
-        RAISE INFO 'signature summary - %', tablename;
-        PERFORM backfill_named_table(tablename, thisday);
-    END LOOP;
```
willkg: You also need to remove the line on 117:

```
DROP TABLE temp_signature_summaries;
```

PR #3955, opened 2017-09-01
fixes bug 1396001 - add mdsw_status_string to ProcessedCrash API

willkg:


willkg:
We can't test this in the local dev environment because the webapp can't pull data from s3--have to wait for it to get merged to -stage.

On there, it's easy:

```
$ curl https://crash-stats.allizom.org/api/ProcessedCrash/?crash_id=1eabe5e2-0dca-47a0-bcba-e924f0170830 | jq .mdsw_status_string
```

If it's `null`, then we blew it. If it's `ERROR_NO_MINIDUMP_HEADER`, then we win!

willkg:
This is a one-line change that was talked about in bug #1396001 and the plan was approved by two people. No one is around right now and this is blocking my progress on the signature generator.

Given that, I'm going to self-r+ this since it's low risk, we've talked about it as a group so it's not a surprise, and we have a chance to look at it before it goes to -prod.

adngdb:
This should actually be done using the Super Search Fields list. That whitelist thing should not even exist anymore. If a field is in the Super Search Fields list with ``is_exposed`` and ``is_returned`` as ``True``, and is in the ``processed_crash`` namespace, then it will be returned through the ``ProcessedCrash`` API. 

This might still work, but you'll want to apply the same change to the list of fields, for the day we remove that whitelist, it might undo your fix.

adngdb:
Just read the bug and saw that Peter already explained it all. It seems you're on top of things! Sorry for the noise. :)

PR #3956, opened 2017-09-01
fixes bug 1395681 - cache .json files on disk for ftpscraper

willkg:
This adds a cachedir configuration property letting you specify a directory on
disk to cache .json files. It defaults to "don't cache on disk", but this
enables the disk cache for the local dev environment.

### DIFF
--- a/docker/run_update_data.sh
+++ b/docker/run_update_data.sh
@@ -6,18 +6,25 @@
 
 set -eo pipefail
 
+HOSTUSER=$(id -u):$(id -g)
+
+# The assumption is that you're running this from /app inside the container
+CACHEDIR=/app/.cache/ftpscraper
+
 # Fetch and update release information for these products (comma-delimited)
 PRODUCTS="firefox"
 CRONTABBER="docker-compose run crontabber python socorro/cron/crontabber_app.py"
 
 
-# Fetch release data
-${CRONTABBER} --job=ftpscraper \
-       --crontabber.class-FTPScraperCronApp.products=${PRODUCTS}
+# Fetch release data -- do it as the user so it can cache things
+docker-compose run -u "${HOSTUSER}" crontabber python socorro/cron/crontabber_app.py \
+               --job=ftpscraper \
+               --crontabber.class-FTPScraperCronApp.cachedir=${CACHEDIR} \
+               --crontabber.class-FTPScraperCronApp.products=${PRODUCTS}
 
 # Update featured versions data based on release data
 ${CRONTABBER} --job=featured-versions-automatic \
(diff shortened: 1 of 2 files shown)


docker/run_update_data.sh
```diff
@@ -6,18 +6,25 @@
...
 
 
-# Fetch release data
-${CRONTABBER} --job=ftpscraper \
-       --crontabber.class-FTPScraperCronApp.products=${PRODUCTS}
+# Fetch release data -- do it as the user so it can cache things
+docker-compose run -u "${HOSTUSER}" crontabber python socorro/cron/crontabber_app.py \
+               --job=ftpscraper \
+               --crontabber.class-FTPScraperCronApp.cachedir=${CACHEDIR} \
+               --crontabber.class-FTPScraperCronApp.products=${PRODUCTS}
```
willkg: This has to run as the host user so that the file permissions are right. It's the same thing as what the `docker/as_me.sh` script does, but that specifically runs the processor container and we need to run the crontabber one.

socorro/cron/jobs/ftpscraper.py
```diff
@@ -296,8 +305,26 @@ def __init__(self, *args, **kwargs):
...
+
+        if url.endswith('.json') and self.config.cachedir:
+            is_caching = True
+            fn = os.path.join(self.config.cachedir, self.url_to_filename(url))
+            if not os.path.isdir(os.path.dirname(fn)):
+                os.makedirs(os.path.dirname(fn))
+            if os.path.exists(fn):
+                with open(fn, 'r') as fp:
+                    print('hit')
+                    return fp.read()
```
willkg: I tinkered with making this a decorator, but it's tricky since we need the `self.config.cachedir` value. I figured this is good enough for now.

socorro/cron/jobs/ftpscraper.py
```diff
@@ -296,8 +305,26 @@ def __init__(self, *args, **kwargs):
...
+        fn = None
+
+        if url.endswith('.json') and self.config.cachedir:
+            is_caching = True
+            fn = os.path.join(self.config.cachedir, self.url_to_filename(url))
+            if not os.path.isdir(os.path.dirname(fn)):
+                os.makedirs(os.path.dirname(fn))
+            if os.path.exists(fn):
+                with open(fn, 'r') as fp:
+                    print('hit')
```
peterbe: deliberately left?

willkg: Yeah... it helps to know whether it's a cache hit or not. Having said that, this particular message is clunky. I'll remove it.

willkg:
Thank you!

PR #3961, opened 2017-09-06
fixes bug 1393286 - fix IntegrationTestIndexCleaner tests

willkg:
If indices existed already, then these tests will fail. That's problematic
because it also means that it's stomping on non-test data. This adds a test_
prefix so that it's only looking at test_ things and not stomping on things it
shouldn't be.

To test:

1. process a crash
2. run `curl localhost:9200/_aliases` from the host--that shows you the indices and aliases that exist in Elasticsearch
3. run `make dockertest` which will run all the tests -- the `test_index_cleaner` tests should pass
4. run `curl localhost:9200/_aliases` and it should be the same indices and aliases as in step 2

### DIFF
--- a/socorro/unittest/external/es/test_index_cleaner.py
+++ b/socorro/unittest/external/es/test_index_cleaner.py
@@ -5,8 +5,6 @@
 import datetime
 import elasticsearch
 
-from nose.tools import ok_
-
 from socorro.lib.datetimeutil import utc_now
 from socorro.external.es.index_cleaner import IndexCleaner
 from socorro.unittest.external.es.base import (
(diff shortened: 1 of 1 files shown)


willkg:
While working on this, I noticed we don't have any idea of an index prefix in the Elasticsearch code. I think we should rework all our Elasticsearch code to support a prefix for indices as well as aliases.

Couple of good things that would come out of that:

1. it makes running tests safer in regards to the local dev environment
2. it makes it possible to run two environments that use the same Elasticsearch cluster

I don't think it's *hard* to do. I think it mostly requires some code re-arranging.

If we think that's a good thing to do, I'll open a new bug and work on that and we can close this PR out.

willkg:
@peterbe Per conversation on IRC, I think we should go forward with this and I'll think about writing up a bug to fix it better later.

willkg:
Thank you!

PR #3965, opened 2017-09-07
Add cache hits summary to ftpscraper

willkg:
This helps us know whether the file-based cache is helping or not without being
overly verbose.

### DIFF
--- a/socorro/cron/jobs/ftpscraper.py
+++ b/socorro/cron/jobs/ftpscraper.py
@@ -306,6 +306,7 @@ class FTPScraperCronApp(BaseCronApp, ScrapersMixin):
         )
 
         self.cache_hits = 0
+        self.cache_misses = 0
 
     def url_to_filename(self, url):
         fn = re.sub('\W', '_', url)
@@ -326,6 +327,7 @@ class FTPScraperCronApp(BaseCronApp, ScrapersMixin):
                 self.cache_hits += 1
                 with open(fn, 'r') as fp:
                     return fp.read()
+            self.cache_misses += 1
 
         response = self.session.get(
             url,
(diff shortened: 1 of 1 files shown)


willkg:
I wanted some kind of indicator that the cache is getting hit and for how much. Originally, I had it printing "hit!" every time, but agreed that's a bit much. This changes it to a final count at the end.

@peterbe Is this ok with you? If not, I can figure out another way to do it.

willkg:
I like that idea. ^^^ improves it by adding cache misses and percentages. It looks like this:

```
2017-09-07 20:12:50,209 DEBUG - MainThread -  - MainThread - Cache: hits: 608 (1.00%) misses: 0 (0.00%)
```

(That's after running it twice in a row.)

socorro/cron/jobs/ftpscraper.py
```diff
@@ -368,7 +370,13 @@ def run(self, date):
...
 
         if self.config.cachedir:
-            self.config.logger.debug('Cache hits: %d' % self.cache_hits)
+            total = float(self.cache_hits + self.cache_misses)
+            self.config.logger.debug('Cache: hits: %d (%2.2f%%) misses: %d (%2.2f%%)' % (
+                self.cache_hits,
+                self.cache_hits / total * 100,
+                self.cache_misses,
+                self.cache_misses / total * 100,
+            ))
```
peterbe: It's not like you didn't know this but one convenient way to do percentages in python 2.7 of ints is to do:
```
p = 100.0 * one_int / another_int
```

peterbe:
Still approve.

willkg:
Thank you!

PR #3966, opened 2017-09-08
fixes bug 1397857 - configure webapp s3 bits

willkg:
The local dev environment requires that we use a different boto connection
context and that requires a bunch of additional configuration. This adjusts the
webapp configuration to carry all that through.

The defaults are taken from the configuration of the appropriate classes.


willkg:
To test:

1. process a crash in your local dev environment
2. then try to go to `http://localhost:8000/report/index/<CRASHID>`

Prior to this commit, that would return a "Crash not found" page. Now it returns the appropriate data.

If there are better ways to thread configman-based configuration through the webapp settings system, I'm all ears.

willkg:
Yeah... localstack-s3 is kind of painful in the persistence department. I haven't written up a bug, yet, but probably will when I have a chance to think about it more.

willkg:
Thank you!

PR #3967, opened 2017-09-08
Add links to siglists in the signature docs and README

luser:


willkg:
I'm totally with you on this. My next step is to reformat the siglists README and then include that in the docs. I think that's a better way to tackle pulling all this stuff together because it has a lower maintenance burden.

I'll do a PR this morning for that.

willkg:
I threw together PR #3968. I think that covers the underlying issue you were trying to fix here.

@luser Does that sound right?

luser:
That's perfect! I was just looking for a quick way to get that relevant info into these places, and having it in the docs is even better. Thanks!

PR #3968, opened 2017-09-08
Reformat siglists README and add to docs

willkg:
This reformats the siglists README from markdown to restructured text and adds an include in the docs.

I was thinking doing includes like this means we can document things once in a helpful place near the code it covers and have it automatically part of the manual. Thus it's findable by users via different paths, but doesn't increase the maintenance burden.


socorro/siglists/README.rst
```diff
@@ -0,0 +1,195 @@
+Signatures Utilities Lists
```
willkg: I ran this through pandoc to do the first pass of reformatting. Then I went through it and updated some stuff.

Making sure it's correct is good, so if you see anything that's not right, we should fix it in this PR.

socorro/siglists/README.rst
```diff
@@ -0,0 +1,195 @@
...
+   element the top of the new sub-stack.
+3. We rewrite every signature missing symbols that matches the `Trim DLL
+   Signatures <#trim-dll-signatures>`_ to be the module only (the part before
+   the first ``@`` sign). We also merge them so only one of those frames makes
+   it to the final signature.
+4. We accumulate signatures that match the `Prefix Signatures
+   <#prefix-signatures>`_, until something doesn't match.
+5. We normalize each signature we accumulated. Signatures that match the
+   `Signatures With Line Numbers <#signatures-with-line-numbers>`_ have their
+   associated code line number added to them, like this: ``signature:42``.
```
willkg: I have no idea if the above links will work on socorro.readthedocs.org... we'll have to find out. I figure we can do a followup PR with any fixes for things we discover.

socorro/siglists/README.rst
```diff
@@ -0,0 +1,195 @@
...
+
+How to review a siglist change
+------------------------------
+
+The first step is to verify that there is no typo in the change (usually, the
+bug contains examples of crash reports that should be impacted, look at their
+frames). Note that we have a unit test that verifies there are no syntax errors
+in those files.
+
+Run the pull request changes through signature generation using the command line
```
luser: This could stand to have a link to the doc section about the command line interface.

socorro/siglists/README.rst
```diff
@@ -0,0 +1,195 @@
...
+2. We walk the stack, ignoring everything that matches the `Irrelevant
+   Signatures <#irrelevant-signatures>`_. We consider the first non-matching
+   element the top of the new sub-stack.
+3. We rewrite every signature missing symbols that matches the `Trim DLL
+   Signatures <#trim-dll-signatures>`_ to be the module only (the part before
+   the first ``@`` sign). We also merge them so only one of those frames makes
+   it to the final signature.
+4. We accumulate signatures that match the `Prefix Signatures
+   <#prefix-signatures>`_, until something doesn't match.
+5. We normalize each signature we accumulated. Signatures that match the
```
luser: As a followup, it would be great to document what "normalize" means here. I was looking at [a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1397926) this morning that's going to want to add a new normalization rule, and I looked at the siglists doc first before realizing that it wasn't documented there. I can probably write the explanation from reading the current code, or you can take a crack at it and I can edit.

socorro/siglists/README.rst
willkg: Oo--good idea!

socorro/siglists/README.rst
willkg: Yeah... this is still kind of crap for docs about how it all works.

Bug 1306643 (https://bugzilla.mozilla.org/show_bug.cgi?id=1306643) covers documenting the rules. I definitely want to do that. I was thinking of fixing the docstrings and then running them through autodoc and exposing the `DEFAULT_PIPELINE` for ordering. Or something along those lines that fleshes out the documentation without creating a maintenance burden.

I think this will improve when someone works on that bug.

willkg:
Thank you!

I'll check readthedocs later and see if the links and such work and do a followup PR if necessary.

PR #3969, opened 2017-09-08
fixes bug 794524 - drop plugins_report table

peterbe:
When I started working on this, I discovered that it appears that we're still actively writing to these tables. I can't find any reference to it ever being queried. Which kinda makes sense since it used to be used by the search we had before we had super search. 

So I'm going to be careful and first make a PR that deprecates the writes. Then we'll let that go into stage and check that all is well with the world.

### DIFF
--- a/alembic/versions/a38b03765a79_drop_plugins_reports_tables.py
+++ b/alembic/versions/a38b03765a79_drop_plugins_reports_tables.py
@@ -1,0 +1,48 @@
+"""drop plugins_reports tables
+
+Revision ID: a38b03765a79
+Revises: ae2c8ff8c073
+Create Date: 2017-09-08 13:23:44.142686
+
+"""
+
+from alembic import op
+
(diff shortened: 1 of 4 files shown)


peterbe:
Here's the PR that this depends on 
https://github.com/mozilla-services/socorro/pull/3970
Technically it doesn't because codewise they're disjoint but it doesn't make sense to drop a table when you still have SQL insert statements in the Python code.

peterbe:
Travis is failing right now. That'll get taken care of once https://github.com/mozilla-services/socorro/pull/3970 lands and we rebase up this one.

willkg:
I vote we wait until Monday to work on this further. That gives -stage some time to be happy/sad about `plugins_report` changes in PR #3970.

Does that sound ok?

peterbe:
> Does that sound ok?

Totally. No rush. Also, it's 3pm on a Friday.

PR #3970, opened 2017-09-08
fixes bug 1398195 - Deprecate storing plugins_reports

peterbe:


socorro/external/postgresql/crashstorage.py
```diff
@@ -238,54 +237,19 @@ def _save_plugins(self, connection, processed_crash, report_id):
...
-                            processed_crash['date_processed'],
-                            plugin_version)
-            execute_no_results(connection,
-                               plugins_reports_insert_sql,
-                               values_tuple)
+                single_value_sql(
+                    connection,
+                    insert_plugsins_sql,
+                    (plugin_filename, plugin_name)
+                )
```
willkg: You could change this to use `execute_no_results` and not return an id.

peterbe:
Both lines you mention in your grep are plugin_reports. The e2e test is the home page and that's super search based. 
The commented out junk in `check_partitions.sql` is a typo if anything. It's not "plugin_reports" it's "plugins_reports*".

I'll make the switch to use `execute_no_results`.

willkg:
Oh, ugh. That's really annoying. Sorry about that!

peterbe:
Updated.

PR #3972, opened 2017-09-08
fixes bug 1397857 - fix calling format

willkg:
For boto connections, we use RegionalS3ConnectionContext and subclasses. This
family of connection contexts default to OrdinaryCallingFormat rather than
SubdomainCallingFormat. This is important because we have periods in our bucket
names and thus SubdomainCallingFormat doesn't work.

However, when I picked the default for resource.boto.calling_format in the
original commit, I looked at S3ConnectionContext rather than
RegionalS3ConnectionContext and thus used the wrong one and now everything is
hosed in -stage and Sentry is giving me dirty looks late on a Friday just before
the weekend.

This fixes all of that.

### DIFF
--- a/webapp-django/crashstats/settings/base.py
+++ b/webapp-django/crashstats/settings/base.py
@@ -659,7 +659,7 @@ SOCORRO_IMPLEMENTATIONS_CONFIG = {
             'port': config('resource.boto.port', None),
             'secure': config('resource.boto.secure', True),
             'calling_format': config(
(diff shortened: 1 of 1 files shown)


willkg:
Travis and friends are useless here because they're not running in a server environment. So I'm going to self-r+ this and merge it now and then see what happens on -stage later tonight. If it's fixed, yay! If it's not fixed, I'll revert the whole thing and wear a cone of shame all next week.

PR #3975, opened 2017-09-11
Remove broken badges

stephendonner:
@chartjes / @willkg r?

This host hasn't existed for a while, and we're in the middle of transition to a new Ops-provisioned host.  If desired, I can do another PR when that's up and running.

(But it'll still be behind VPN, so the badges won't reflect status, they'll just function as links to the builds.)

### DIFF
--- a/e2e-tests/README.md
+++ b/e2e-tests/README.md
@@ -3,10 +3,6 @@ End-to-End Tests for Socorro
 
 Continuous Integration
 ----------------------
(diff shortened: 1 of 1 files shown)


willkg:
@chartjes Is this ok to land?

chartjes:
@willkg 👍

willkg:
Thank you!

PR #3976, opened 2017-09-11
fixes bug 1398116 - fetch lots of test data

willkg:
This adjusts the `scripts/fetch_crash_data.py` script to have a mode where it can
fetch a bunch of data for specified dates.

To test:

```
$ ./docker/as_me.sh bash

# Fetch 10 crashes from yesterday
app@...:/app$ ./scripts/fetch_crash_data.py

# Fetch 5 crashes from September 1st
app@...:/app$ ./scripts/fetch_crash_data.py --num=5 --date=2017-09-01

# Fetch a specific crash
app@...:/app$ ./scripts/fetch_crash_data.py 19a766c5-e388-481c-bcae-6d4bb0170911
```

This defaults to saving in the directory `./crashdata`. That's a change--previously, you'd have to specify the directory. Now you can optionally specify it.

When fetching a bunch of unspecified crashes, it *only* uses publicly available data. I was thinking this is more responsible to do. If you want to fetch 1000 crashes with the private data, you can still do that, but you have to be explicit about what crashes you're fetching. Seems like that's the right thing to do here.

Two things to note:

1. There are no tests for `fetch_crash_data.py`, yet. It's probably worth writing tests, but since it's mostly hitting production, it'd be a lot of mocking and I'm on the fence about how valuable that ends up being.
2. We've got a lot of fake data generating code that should get removed. I wrote up bug https://bugzilla.mozilla.org/show_bug.cgi?id=1398946 to cover that.

So does this feel good enough for now? Should it get split out into two scripts?

### DIFF
--- a/socorro/scripts/fetch_crash_data.py
+++ b/socorro/scripts/fetch_crash_data.py
@@ -5,28 +5,42 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import argparse
+import datetime
 import json
 import os
 import os.path
 
 import requests
 
-from socorro.lib.datetimeutil import JsonDTEncoder
+from socorro.lib.datetimeutil import JsonDTEncoder, utc_now
 from socorro.scripts import WrappedTextHelpFormatter
 
 
+DESCRIPTION = """
+Fetches data for specific crashes or most recent crashes
+"""
+
 EPILOG = """
-Given a crash id, fetches crash data and puts it in specified directory
+This script has two modes: fetching specific crashes and fetching arbitrary crashes.
 
-This requires an auth-token to be in the environment::
+If you specify crash ids, then it fetches data for those specific crashes. This mode requires an
+auth-token to be in the environment. For example:
 
-    SOCORRO_API_TOKEN=xyz
+    $ SOCORRO_API_TOKEN=xyz ./scripts/fetch_crash_data.py CRASHID
 
 To create an API token for Socorro in -prod, visit:
 
     https://crash-stats.mozilla.com/api/tokens/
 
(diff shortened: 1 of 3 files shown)


socorro/scripts/fetch_crash_data.py
```diff
@@ -127,10 +197,17 @@ def main(argv):
         print('%s is not a directory. Please fix. Exiting.' % outputdir)
         return 1
 
-    api_token = os.environ.get('SOCORRO_API_TOKEN')
-    print('Using api token: %s' % api_token)
+    crashes = []
+    if args.crashid:
+        api_token = os.environ.get('SOCORRO_API_TOKEN')
+        print('Using api token: %s' % api_token)
```
peterbe: Can we please truncate this. Just in case I, or someone equally dangerous, copy all the stdout as part of a pastebin in a bug or a irc conversation or something.

socorro/scripts/fetch_crash_data.py
```diff
@@ -127,10 +197,17 @@ def main(argv):
...
+
+        crashes = args.crashid
+    else:
+        api_token = None
```
peterbe: I don't get this. Why can't the API token be used to download 10 crashes?

willkg:
I talked with @peterbe about this and we collectively decided this is a terrible idea. It mixes two different things and it'd be better if we let `fetch_crash_data.py` do its one thing well and instead wrote a script that you can feed some supersearch criteria to and it yields a list of crash ids.

So I'm going to close this out and write that script.

Having said that, there are a couple of issues that @peterbe brought up that should get fixed in `fetch_crash_data.py`. I'll fix those in a separate PR now.

PR #3980, opened 2017-09-12
Make instances restart if they crash in docker

claudijd:
When fuzzing the webapp on the bench, we (@caggle and I) discovered that it would crash unexpectedly and need to be re-deployed.  Hoping this PR might help make it more stable.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,6 +9,7 @@ services:
       context: .
       dockerfile: ./docker/Dockerfile
     image: local/socorro_base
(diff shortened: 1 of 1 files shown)


willkg:
This is actually not what we want to have happen by default, but I do see the problem you're hitting. I suggest you override using `docker-compose.override.yml` instead.

There's some documentation here:

https://docs.docker.com/compose/extends/

claudijd:
@willkg alright, thanks!

PR #3981, opened 2017-09-13
Adjust run_webapp.sh text to clarify url

willkg:
The django runserver tells you to connect at a url formed by taking the host and
port the server is listening to. That's great, however we set it to listen to
0.0.0.0 (all interfaces) and thus the url it says to use is wrong. The code that
generates that message is deep within the bowels of the runserver code, so it's
not really changeable.

This adjusts the message that docker/run_webapp.sh spits out so it's more
prominent and clearer which url to use.


willkg:
@lonnen You mentioned this problem a week or two ago. It's impossible to change the message that Django spits out because it's deep in the bowels of a big function:

https://github.com/django/django/blob/2d4bc5a60aa8a076689667c550ded96b87bc463e/django/core/management/commands/runserver.py#L122

That's Django 1.8. This is master tip:

https://github.com/django/django/blob/master/django/core/management/commands/runserver.py#L126

We could roll our own runserver command, but that's a bunch of work and ongoing maintenance. Though maybe we could do some other things in it that are helpful.

With the change in this PR, it spits this out:

```
$ docker-compose up webapp
socorro_postgresql_1 is up-to-date
socorro_localstack-s3_1 is up-to-date
socorro_rabbitmq_1 is up-to-date
socorro_elasticsearch_1 is up-to-date
Starting socorro_webapp_1 ... 
Starting socorro_webapp_1 ... done
Attaching to socorro_webapp_1
webapp_1         | Already have webapp behavior configuration...
webapp_1         | ******************************************************************
webapp_1         | Running webapp in local dev environment.
webapp_1         | Connect with your browser using: http://localhost:8000/ 
webapp_1         | ******************************************************************
webapp_1         | Performing system checks...
webapp_1         | 
webapp_1         | System check identified no issues (0 silenced).
webapp_1         | September 13, 2017 - 00:46:21
webapp_1         | Django version 1.8.18, using settings 'crashstats.settings'
webapp_1         | Starting development server at http://0.0.0.0:8000/
webapp_1         | Quit the server with CONTROL-C.
```

Do you think this helps enough?

If not, I think we could make it more prominent by wrapping it with cowsay or ascii art. If that's not good enough, then I think we'll have to roll our own Django runserver command.

lonnen:
I do think that helps! thank you

willkg:
Thank you!

PR #3982, opened 2017-09-13
1398116 fetch crashids

willkg:
This adds a `scripts/fetch_crashids.py` script that will generate a list of crash ids based on specified parameters.

In this first pass, we're keeping the parameters pretty limited to date and number and a url. We can enhance that later as needs arise.

To test, first get into a bash session in the processor container:

```
```

Then to get command help:

```
app@...:/app$ ./scripts/fetch_crashids.py --help
```

To pull 100 crashes for Firefox from yesterday:

```
app@...:/app$ ./scripts/fetch_crashids.py
```

To pull 5 crashes for Firefox from September 1st:

```
app@...:/app$ ./scripts/fetch_crashids.py --date=2017-09-01 --num=5
```

To pull 100 crashes from the results of a Super Search query:

```
app@...:/app$ ./scripts/fetch_crashids.py "--url=https://crash-stats.mozilla.com/search/?product=Firefox&date=%3E%3D2017-09-05T15%3A09%3A00.000Z&date=%3C2017-09-12T15%3A09%3A00.000Z&_sort=-date&_facets=signature&_columns=date&_columns=signature&_columns=product&_columns=version&_columns=build_id&_columns=platform#facet-signature"
```

Specifying `--date` will override date params in any specified `--url` query.

If you also add `-v`, then it'll print out the params it's using to do the search with. That's super helpful for debugging and verification.

### DIFF
--- a/docs/components/processor.rst
+++ b/docs/components/processor.rst
@@ -92,7 +92,7 @@ Use with ``scripts/fetch_crashids.py`` to fetch crash data from 100 crashes from
 yesterday for Firefox::
 
   $ docker/as_me.sh bash
-  app@...:/app$ scripts/fetch_crashids.py | xargs scripts/fetch_crash_data.py ./testdata
+  app@...:/app$ scripts/fetch_crashids.py | scripts/fetch_crash_data.py ./testdata
 
 
 You can get command help::
--- a/docs/components/processor.rst
+++ b/docs/components/processor.rst
@@ -47,6 +47,31 @@ crash id to the "socorro.normal" RabbitMQ queue.
 We have helper scripts for these steps.
 
 
+``scripts/fetch_crashids.py``
+-----------------------------
+
+This will generate a list of crash ids from -prod that meet specified criteria.
+Crash ids are printed to stdout, so you can use this in conjunction with other
+scripts or redirect to a file.
+
+This pulls 100 crash ids from yesterday for Firefox product::
+
+  $ docker/as_me.sh scripts/fetch_crashids.py
+
+This pulls 5 crash ids from 2017-09-01::
+
+  $ docker/as_me.sh scripts/fetch_crashids.py --num=5 --date=2017-09-01
+
+This pulls 100 crash ids for criteria specified with a Super Search url that we
+copy and pasted::
+
+  $ docker/as_me.sh scripts/fetch_crashids.py "--url=https://crash-stats.mozilla.com/search/?product=Firefox&date=%3E%3D2017-09-05T15%3A09%3A00.000Z&date=%3C2017-09-12T15%3A09%3A00.000Z&_sort=-date&_facets=signature&_columns=date&_columns=signature&_columns=product&_columns=version&_columns=build_id&_columns=platform#facet-signature"
+
+You can get command help::
+
+  $ docker/as_me.sh scripts/fetch_crash_data.py --help
+
+
 ``scripts/fetch_crash_data.py``
 -------------------------------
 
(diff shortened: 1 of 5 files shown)


socorro/scripts/__init__.py
```diff
@@ -10,7 +10,17 @@ def _fill_text(self, text, width, indent):
...
+                subparts = part.split('\n')
+                for j, subpart in enumerate(subparts):
+                    subparts[j] = super(WrappedTextHelpFormatter, self)._fill_text(
+                        subpart, width, indent
+                    )
+                parts[i] = '\n'.join(subparts)
+            else:
+                parts[i] = super(WrappedTextHelpFormatter, self)._fill_text(part, width, indent)
+
+        return '\n\n'.join(parts)
```
willkg: I pulled this out of a previous PR. It's helpful for printing out command line help that has lists.

docs/components/processor.rst
```diff
@@ -47,6 +47,31 @@ crash id to the "socorro.normal" RabbitMQ queue.
...
+  $ docker/as_me.sh scripts/fetch_crashids.py
+
+This pulls 5 crash ids from 2017-09-01::
+
+  $ docker/as_me.sh scripts/fetch_crashids.py --num=5 --date=2017-09-01
+
+This pulls 100 crash ids for criteria specified with a Super Search url that we
+copy and pasted::
+
+  $ docker/as_me.sh scripts/fetch_crashids.py "--url=https://crash-stats.mozilla.com/search/?product=Firefox&date=%3E%3D2017-09-05T15%3A09%3A00.000Z&date=%3C2017-09-12T15%3A09%3A00.000Z&_sort=-date&_facets=signature&_columns=date&_columns=signature&_columns=product&_columns=version&_columns=build_id&_columns=platform#facet-signature"
```
willkg: This line is probably going to look terrible in the docs, but I wasn't sure what else to do since Super Search urls are long.

docs/components/processor.rst
```diff
@@ -47,6 +47,31 @@ crash id to the "socorro.normal" RabbitMQ queue.
 We have helper scripts for these steps.
 
 
+``scripts/fetch_crashids.py``
+-----------------------------
+
+This will generate a list of crash ids from -prod that meet specified criteria.
```
peterbe: The `-prod` nomenclature is a bit cryptic. I don't really know where it comes from but perhaps it'd be better to simply state `crash-stats.mozilla.com` instead.

willkg:
^^^ That rebases against master, improves things per your suggestions, and also fixes `scripts/fetch_crash_data.py` so it can take crash ids via the command line or via stdin which is nice because then we don't have to use xargs plus it probably works better for large numbers of crash ids.

socorro/scripts/fetch_crash_data.py
```diff
@@ -140,7 +142,14 @@ def main(argv):
     if api_token:
         print('Using api token: %s%s' % (api_token[:4], 'x' * (len(api_token) - 4)))
 
-    for crash_id in args.crashid:
+    if args.crashid:
+        crashids_iterator = iter(args.crashid)
```
peterbe: Is it not already a list?

peterbe:
Still works like a charm! Still approved.

socorro/scripts/fetch_crash_data.py
willkg: I actually have no idea what it is. It comes from args parsing, so I figured I'd wrap it in an iterator figuring can't hurt. But you're right--it's probably fine not to do this.

willkg:
Thank you!

PR #3984, opened 2017-09-13
[DO NOT MERGE] bug 1375434 - populate signatures from supersearch

peterbe:
@adngdb Can you please have a look and check that I've got the basics right. 

@willkg Perhaps stall reviewing. But please have a think about how/when this should be executed in docker. Basically, if you don't run this after crashes have been inserted the Top Crashers report won't work. So it should ideally be executed some time after you have run `fetch_crash_data` and that being synced. 

-----

This PR is not ready yet because it doesn't have any test coverage of the new code. I just want to get it out early. Primarily for Adrian's early feedback. 

The idea is that it's supposed to run every 1h. So 24 times per day. It's backfill based so it won't miss any time windows. 
Each hour it then does a facet search for ALL different signatures. The `_facets_size` is 1,000. 
Then once it has the signatures, it removes the ones already in the `signatures` table. 
Now for each new signature it does 2 SuperSearch queries. One by sorting on `date` ascending and one sorting by `build_id` ascending. 

One tricky thing is that when you try to find the oldest crash (by signature) you have to expand the date interval as much as possible. I picked the same number of days as there are weeks in the `index_cleaner`. Perhaps because I tested this with stage's ES1 cluster, I got lots of errors about indexes not existing when going that far back. That's stage ES for you :)
Also, I didn't want this to break and get stuck on that so I made the `sensitive_to_errors` off by default. 

Testing this is in docker is going to be hard since we have so little historic record there. I used stage's ES cluster like this:
```
python socorro/cron/jobs/first_signatures.py --date="2017-09-13T15:13:14.98765Z" --elasticsearch.elasticsearch_urls=localhost:9222
```
...using the virtualenv. 
This testing "obviously fails". Some of the signatures I found were actually first since in 2016-sometime-sometime but this crontabber app can only go back 26 weeks at best. 
However, I'm not worried because this app will slowly replace the old stored procedure that populates `signatures`. In fact, I think we should install this (when it's ready) and look at the logs to make sure it populates. We might want to randomly deleting some entries in `signatures` on stage's PG and make sure it gets filled in by this new app.

### DIFF
--- a/socorro/cron/jobs/first_signatures.py
+++ b/socorro/cron/jobs/first_signatures.py
@@ -82,7 +82,6 @@ class FirstSignaturesCronApp(BaseCronApp):
         signatures = [
             x['term'] for x in results['facets']['signature']
         ]
-        print("LIST LENGTH", len(signatures), 'SET LENGTH', len(set(signatures)))
         self.config.logger.info(
             'Found %s unique signatures between %s and %s',
             len(signatures),
--- a/socorro/cron/jobs/first_signatures.py
+++ b/socorro/cron/jobs/first_signatures.py
@@ -74,7 +74,7 @@ class FirstSignaturesCronApp(BaseCronApp):
                 '<{}'.format(date.isoformat()),
             ],
             '_facets': 'signature',
-            '_facets_size': 1000,
+            '_facets_size': 10000,
             '_results_number': 0,
             '_fields': all_fields,
         }
--- a/socorro/cron/jobs/first_signatures.py
+++ b/socorro/cron/jobs/first_signatures.py
@@ -1,0 +1,204 @@
+import datetime
+import sys
+
+import mock
+
+from configman import Namespace
+from configman.converters import class_converter
+from crontabber.base import BaseCronApp
+from crontabber.mixins import (
+    as_backfill_cron_app,
+    with_postgres_transactions
+)
+
+from socorro.external.postgresql.signature_first_date import SignatureFirstDate
+from socorro.external.es.base import ElasticsearchConfig
+from socorro.external.es.supersearch import SuperSearch
+from socorro.external.es.super_search_fields import SuperSearchFields
+from socorro.app.socorro_app import App, main
+from socorro.lib.datetimeutil import string_to_datetime
+
+
+class SuperSearchErrors(Exception):
+    """Happens when we make a SuperSearch query and there's something in the
+    results 'errors' dict."""
+
+
+@with_postgres_transactions()
+@as_backfill_cron_app
+class FirstSignaturesCronApp(BaseCronApp):
+    app_name = 'first-signatures'
+    app_description = 'Using SuperSearch to log when signatures first appeared'
+    app_version = '0.1'
+
+    required_config = Namespace()
+    required_config.add_option(
+        'window_seconds',
+        default=60,  # seconds
+        doc='Number of seconds to search for signature facets back'
+    )
+    required_config.add_option(
+        'sensitive_to_errors',
+        # default=True,
+        default=False,
+        doc=(
+            'If there are, for example, missing indexes in ES you can get errors '
+            "in supersearch doing queries across dates we don't have indices for. "
+            'If this is the case, halt this crontabber app.'
+        )
+    )
+    required_config.add_option(
+        'historic_days',
+        # The number 26 comes from the default in index_cleaner which is used
+        # as a retention policy for cleaning elasticsearch.
+        default=26 * 7,  # days
+        doc=(
+            'When looking for the oldest crash possible (by a specific signature) '
+            'go back this many days'
+        )
+    )
+    required_config.add_option(
+        'elasticsearch_class',
+        default=ElasticsearchConfig,
+    )
+
+    def run(self, date):
+        # First get a list of all unique signatures, in the given timespan
+        all_fields = SuperSearchFields(config=self.config).get()
+        api = SuperSearch(config=self.config)
+        assert isinstance(date, datetime.datetime), type(date)
+        start_date = date - datetime.timedelta(seconds=self.config.window_seconds)
+        params = {
+            'date': [
+                '>={}'.format(start_date.isoformat()),
+                '<{}'.format(date.isoformat()),
+            ],
+            '_facets': 'signature',
+            '_facets_size': 1000,
+            '_results_number': 0,
+            '_fields': all_fields,
+        }
+        results = api.get(**params)
+        signatures = [
+            x['term'] for x in results['facets']['signature']
+        ]
+        self.config.logger.info(
+            'Found %s unique signatures between %s and %s',
+            len(signatures),
+            start_date,
+            date,
+        )
+        if not signatures:
+            self.config.logger.info(
+                'Exit early because no signature facets'
+            )
+            return
+
+        # Now to figure out which ones we've never seen before
+        signature_first_date_api = SignatureFirstDate(config=self.config)
+        results = signature_first_date_api.get(signatures=signatures)
+        new_signatures = set(signatures) - set(results['hits'])
+        self.config.logger.info('%s new signatures', len(new_signatures))
+
+        # Now, for each of these new signatures we have to find the oldest
+        # crash that has it and write down its date and buildid.
+        historic_date = date - datetime.timedelta(
+            days=self.config.historic_days
+        )
+        for signature in new_signatures:
+            # First look up by the oldest 'date'.
+            params = {
+                '_results_number': 1,
+                'signature': '={}'.format(signature),
+                '_sort': 'date',
+                '_fields': all_fields,
+                '_columns': ['date'],
+                '_facets_size': 0,
+                'date': [
+                    '>={}'.format(historic_date.isoformat()),
+                    '<{}'.format(date.isoformat()),
+                ]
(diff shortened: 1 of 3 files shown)


socorro/cron/jobs/first_signatures.py
```diff
@@ -0,0 +1,204 @@
...
+@as_backfill_cron_app
+class FirstSignaturesCronApp(BaseCronApp):
+    app_name = 'first-signatures'
+    app_description = 'Using SuperSearch to log when signatures first appeared'
+    app_version = '0.1'
+
+    required_config = Namespace()
+    required_config.add_option(
+        'window_seconds',
+        default=60,  # seconds
```
adngdb: Why is this in seconds and not in minutes? Or at least, why is not equal to an hour?

socorro/cron/jobs/first_signatures.py
```diff
@@ -0,0 +1,204 @@
...
+        api = SuperSearch(config=self.config)
+        assert isinstance(date, datetime.datetime), type(date)
+        start_date = date - datetime.timedelta(seconds=self.config.window_seconds)
+        params = {
+            'date': [
+                '>={}'.format(start_date.isoformat()),
+                '<{}'.format(date.isoformat()),
+            ],
+            '_facets': 'signature',
+            '_facets_size': 1000,
```
adngdb: This is unsatisfying. You risk missing some signatures if there are more new, different signatures in an hour's time. 

One possibility is to add another query before that to estimate the number of distinct signatures. You can do that using ``_facets=_cardinality.signature`` ([doc](https://crash-stats.mozilla.com/documentation/supersearch/#cardinality)). Note that the number it returns is an estimate, it might be safe to add some extra integers to it to make sure you get everything. 

The other possibility is to simply use an arbitrarily large number, because you want to receive all of them signatures anyway.

socorro/cron/jobs/first_signatures.py
```diff
@@ -0,0 +1,204 @@
...
+
+        # Now to figure out which ones we've never seen before
+        signature_first_date_api = SignatureFirstDate(config=self.config)
+        results = signature_first_date_api.get(signatures=signatures)
+        new_signatures = set(signatures) - set(results['hits'])
+        self.config.logger.info('%s new signatures', len(new_signatures))
+
+        # Now, for each of these new signatures we have to find the oldest
+        # crash that has it and write down its date and buildid.
+        historic_date = date - datetime.timedelta(
```
adngdb: You do not need to fetch historical data. You can trust that the table has correct data up until an hour ago, so if something isn't in that table yet, you can be sure that it only appeared in the last hour.

socorro/cron/jobs/first_signatures.py
```diff
@@ -0,0 +1,204 @@
...
+            days=self.config.historic_days
+        )
+        for signature in new_signatures:
```
adngdb: It's sad that you have to iterate over the signatures. Elasticsearch would allow you to do that with just a single query, using a ``min`` aggregation ([doc](https://www.elastic.co/guide/en/elasticsearch/reference/1.4/search-aggregations-metrics-min-aggregation.html)). Unfortunately SuperSearch doesn't support that. Maybe that feature would be worth adding?

socorro/cron/jobs/first_signatures.py
peterbe: Yeah, it should be 60 * 60 by default.

socorro/cron/jobs/first_signatures.py
peterbe: So if I set _facets_size=1000 does that just limit the number of facets based on the number of crashes it matches? So to get more unique signatures, should I use `_facets=_cardinality.signature` instead and set a high `_results_number`?

socorro/cron/jobs/first_signatures.py
peterbe: I don't get it. I really want the oldest crash possible. I sort by oldest first. But that's just oldest first in the time window. So if I don't use historical data, I'd get the oldest of the 1hour most recent crashes. 
No?

socorro/cron/jobs/first_signatures.py
peterbe: There are so few of these. New signatures in 24 hours are rare. If we have, god forbid, 5 new signatures in a day we'd have to do 2*5 supersearch queries (+1 for the cardinality facet search). It's nothing for the crontabber to poke around with. It has time.

socorro/cron/jobs/first_signatures.py
adngdb: If you have a set of 11 crash results, each with a different signature, and you set ``_facet_size=10`` then you are missing out on one of those signatures. The same goes if you have 1001 distinct signatures and set size to 1000. This is the risk I'm seeing here, especially since it seems that over the last hour, we [processed more than 4000 different signatures](https://crash-stats.mozilla.com/api/SuperSearch/?_facets=_cardinality.signature&_results_number=0&date=%3E%3D2017-09-14T20%3A00%3A00&date=%3C%3D2017-09-14T21%3A00%3A00).

The number of crash reports returned by your search is unrelated to the size of the facet. You can set ``_results_number=0`` and ``_facets_size=1000`` and get your 1000 facet results.

socorro/cron/jobs/first_signatures.py
adngdb: Yes, but since that signature was not in the ``signature`` table beforehand, that means its first appearance date has to be in the time window of the currently running job. 

Say you have this in your ``signatures`` table: 

signature | date | build
-|-|-
foo | 2010-01-01 | 20100101
bar | 2016-10-12 | 20161010

Now you are running that job for the hour between ``2017-01-01 00:00:00`` and ``2017-01-01 01:00:00``, and you find a signature that doesn't belong to the table, ``bean``. Because you trust that the data in the ``signatures`` table is correct, you know that that ``bean`` signature never appeared before. Hence its first appearance date, and first build, has to be contained within the date range of the job, between ``2017-01-01 00:00:00`` and ``2017-01-01 01:00:00``. There is no need to look for it in older data, because the ``signatures`` table tells you that it never appeared before.

socorro/cron/jobs/first_signatures.py
adngdb: Yeah, I agree. It's just a shame that it isn't easy to solve this in a more elegant way. :)

socorro/cron/jobs/first_signatures.py
peterbe: Ah! Makes sense now. I get it. Thanks!

socorro/cron/jobs/first_signatures.py
peterbe: Once this is up and running I can't imagine there being that many *new* signatures *per hour*.

socorro/cron/jobs/first_signatures.py
peterbe: Fixed.

socorro/cron/jobs/first_signatures.py
peterbe: I'm losing confidence in this. I just now tried changing it from 1,000 to 10,000 and sure enough it found 5,282 unique signatures. Setting it to 10,000 seems to work. 

However, I don't how to use the cardinality. I tried changing `'_facets': 'signature'` to `'_facets.cardinality': 'signature'` but it was no different. 

So is it sane to use 10,000 for `_facets_size`??
I just makes me nervous when the queries are so big. Like there's something fundamentally wrong. 

Other crazy ideas? What about instead of running this crontabber app every hour, we run it every 30 minutes. Or 10 minutes. With Prod ES and a 30 min window I get 3,101 unique signatures. For 10 minutes, 1,330 unique signatures.

peterbe:
@willkg In a few short hours I'm officially leaving the team and I need to focus on my new tasks when I come back. This PR feels a lot bigger than something I can finish the first Monday I'm back. 

Overall the pressing issue is that discussion there between Adrian and me about if it makes sense to use a MASSIVE `_facets_size` or if there is a better way to confidently get a unique/distinct list of all signatures of crashes in a particular time window. I don't get Adrian's point about using the `_facets.cardinality=signature`. And I'm eager to hear what Adrian thinks about keeping the code as is but using `_facets_size=10000` instead. 

Lastly, the other reason I'm not confident I can pursue this patch is because it's simply not ready. The existing code breaks unit tests for reasons I don't understand. And this patch doesn't even have its own tests of the new code.

willkg:
I don't have time to work on this issue *right now* and so it's going to bit rot for a bit. Additionally, in the bug comments we talked about some other options which I want to look into to get a feel for the problem domain and context.

I'm going to close this out. If after looking at the bug more this is the best option, then I'll reopen it and move forward with it.

socorro/cron/jobs/first_signatures.py
adngdb: I don't have a good answer for you here. Setting an arbitrarily high number for ``_facets_size`` is a good thing to do, to ensure you don't miss anything. Running it more regularly is probably good too, though I suspect, as your numbers show, that we won't get much less results, since many signatures overlap. 

To use ``_cardinality.signature`` you would have to make another query, in order to get only that number. It's probably pointless, since we want to get all the results anyway. Using an arbitrary number is probably better.

PR #3987, opened 2017-09-14
fixes bug 1399952 - switch to memcached

willkg:
This fixes logging in for me.


willkg:
Yeah... I think if we add that comment, we have to maintain it going forward and make sure it continues to work. The .ini-dist files now are peppered with similar things and they just never get updated.

Maybe a better place is to put it in the docs or in a blog post that tackles a topic like "things I do when working on certain things in the webapp"?

willkg:
Thank you!

PR #3989, opened 2017-09-14
fixes bug 1399299 - how to run webapp with DEBUG=False

peterbe:
This works really well actually. We talked about the confusion of why there's a `docker/config/docker_common.env` (the webapp part) and `docker/config/local_development.env` but I hope that doesn't affect this. 

Basically, when I follow the instructions I mention in `webapp.rst` I get something really close to prod. I get DEBUG disabled and I get uwsgi. And since the instructions "force" to seting `DEBUG=False` as part of it I don't need to explain how to temporarily mess with your `my.env` file. 

When I go back to regular `docker-compose up webapp` things work great still. If I edit a file like `documention.less` simply editing that file (and Ctrl-S) doesn't restart the runserver but if I make that change AND edit any .py file it does restart runserver and the changes I made in the .less file are immediately reflected. 

Also, the reason this works with httP://localhost:8000 is thanks to [`docker_common.env`](https://github.com/mozilla-services/socorro/blob/dcd868ce9933644725c99a78ed68d9b582fc0282/docker/config/docker_common.env#L111)

### DIFF
--- a/docs/components/webapp.rst
+++ b/docs/components/webapp.rst
@@ -218,3 +218,42 @@ with stack traces by adding this to
       ('Your Name', 'your_email@domain.com'),
   )
   MANAGERS = ADMINS
+
+
+Running Web App in a Prod-like Way
+==================================
+
+By default, for local development when you run ``docker-compose up webapp`` it
+starts the web app geared towards local development. This means it starts
+Django's ``runserver`` and you most likely have ``DEBUG=True`` in your
+``my.env`` file. That means that static assets are automatically served from
+within the individual Django apps rather than serving the minified and
+concatenated static assets you'd get in a production-like environment.
+
+If you want to, locally on your laptop, run the web app in a more
+"prod-like way" here's how you'd do that. First start a ``bash`` shell with
+service ports::
+
+  $ docker-compose run --service-ports webapp bash
+
+Now, the intention is to run the web app with ``DEBUG=False`` but for
+static assets to work you first need to run ``collectstatic``::
+
+  app@...:/app$ cd webapp-django/
+  app@...:/app/webapp-django$ ./manage.py collectstatic --noinput
+  app@...:/app/webapp-django$ cd ..
+
+Now to get as close as possible to a production-like environment, start
+``uwsgi`` instead of Django's ``runserver`` but do it with ``DEBUG=False``::
+
+  app@...:/app$ DEBUG=False bash docker/run_webapp.sh
+
+You will now be able to open ``http://localhost:8000`` on the host and if you
+view the source you see that the minified and concatenated static assets
+are served instead.
+
+When running the web app with ``DEBUG=False`` without using a web server in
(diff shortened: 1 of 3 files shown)


docs/components/webapp.rst
```diff
@@ -218,3 +218,42 @@ with stack traces by adding this to
...
   MANAGERS = ADMINS
+
+
+Running Web App in a Prod-like Way
+==================================
+
+By default, for local development when you run ``docker-compose up webapp`` it
+starts the web app geared towards local development. This means it starts
+Django's ``runserver`` and you most likely have ``DEBUG=True`` in your
+``my.env`` file. That means that static assets are automatically served from
```
willkg: `DEBUG=True` comes from `docker/config/local_development.env`. So unless you're overriding that with your `my.env` file, then the `my.env` file doesn't really enter into it.

webapp-django/crashstats/settings/base.py
```diff
@@ -102,6 +102,7 @@ def path(*dirs):
 
 
 MIDDLEWARE_CLASSES = (
+    'whitenoise.middleware.WhiteNoiseMiddleware',
```
willkg: Do we know if this change affects our current server environments?

peterbe: Not with confidence other than theory. 
In Nginx we have it so that every /static URL gets taken care of by Nginx as a straight file serving. 
https://github.com/mozilla/socorro-infra/blob/master/puppet/modules/socorro/templates/etc_nginx/conf_d/socorro-webapp.conf.erb#L49-L54
Whitenoise doesn't do it by prefix. It basically looks at the `request.path_info` and tries to match that whole path to a file on disk (based on that STATIC_ROOT)

docs/components/webapp.rst
peterbe: Corrected now.

willkg:
This looks good. However, when I run it, it seems like we're not bundling assets into fewer files. So loading the Firefox home page is like 29 requests and almost 600kb.

Does Socorro create bundles? Did the change to whitenoise change that?

peterbe:
> This looks good. However, when I run it, it seems like we're not bundling assets into fewer files. So loading the Firefox home page is like 29 requests and almost 600kb.

If that happens you're not running with `DEBUG=False` for some reason. That's what makes django-pipeline decide what `<script>` and `<link>` tags to put in the html templates.

peterbe:
Could it possibly be that even though you typed `DEBUG=False bash docker/run_webapp.sh` there's a file based `.env` file or something that overrides it back to DEBUG=True?

willkg:
I have no idea what I did. Maybe I typo'd something. It's working as I would expect now!

peterbe:
> I have no idea what I did. Maybe I typo'd something. It's working as I would expect now! 

Oh no. You sound like me. That's totally something I would have said.

PR #3991, opened 2017-09-14
fixes bug 1399120 - fixes symbol upload in local dev environment

willkg:
The local dev environment requires that we can specify host, port, and whether
to use a secure connection in order to connect to the local S3. Other parts of
Socorro already handle this, but symbol upload didn't.

This fixes that.

### DIFF
--- a/docker/config/docker_common.env
+++ b/docker/config/docker_common.env
@@ -89,6 +89,8 @@ processor.raw_to_processed_transform.BreakpadStackwalkerRule2015.symbol_cache_pa
 # ------
 
 ALLOWED_HOSTS=localhost
+AWS_HOST=localstack-s3
+AWS_PORT=5000
 AWS_ACCESS_KEY=foo
 AWS_SECRET_ACCESS_KEY=foo
 CACHE_LOCATION=memcached:11211
@@ -109,5 +111,5 @@ RABBITMQ_VIRTUAL_HOST=rabbitvhost
 RAVEN_DSN=
 SECRET_KEY=secretkey
 SESSION_COOKIE_SECURE=False
-SYMBOLS_BUCKET_DEFAULT_NAME=
+SYMBOLS_BUCKET_DEFAULT_NAME=symbols
 SYMBOLS_BUCKET_EXCEPTIONS=
--- a/docker/config/local_development.env
+++ b/docker/config/local_development.env
@@ -15,3 +15,6 @@ DEBUG=True
 # static files for the webapp get put in /tmp/crashstats-static/ so we
 # need to set STATIC_ROOT to that.
 STATIC_ROOT=/tmp/crashstats-static/
+
+# For connecting to local S3, we need to connect insecurely.
+AWS_SECURE=False
--- a/webapp-django/bin/ci.sh
+++ b/webapp-django/bin/ci.sh
@@ -18,11 +18,12 @@ echo "Linting..."
 git ls-files crashstats | grep '\.py$' | xargs flake8 | bin/linting.py
 
 echo "Starting tests..."
+# Override configuration with better defaults for the tests
 export SECRET_KEY="doesn't matter, tests"
 export CACHE_BACKEND="django.core.cache.backends.locmem.LocMemCache"
 export CACHE_LOCATION="crashstats"
-export OLD_PYTHONPATH=$PYTHONPATH
-export PYTHONPATH=../:$PYTHONPATH
-FORCE_DB=true python manage.py test --noinput
-export PYTHONPATH=$OLD_PYTHONPATH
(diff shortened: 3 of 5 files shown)


willkg:
I suspect maybe the Circle failure is real, but I haven't been able to reproduce it locally. I'll have to look at it tomorrow.

docker/config/docker_common.env
```diff
@@ -109,5 +111,5 @@ RABBITMQ_VIRTUAL_HOST=rabbitvhost
 RAVEN_DSN=
 SECRET_KEY=secretkey
 SESSION_COOKIE_SECURE=False
-SYMBOLS_BUCKET_DEFAULT_NAME=
+SYMBOLS_BUCKET_DEFAULT_NAME=symbols
 SYMBOLS_BUCKET_EXCEPTIONS=
```
willkg: The changes in this file are specific to our local dev environment--nothing fancy here.

docker/config/local_development.env
```diff
@@ -15,3 +15,6 @@ DEBUG=True
 # static files for the webapp get put in /tmp/crashstats-static/ so we
 # need to set STATIC_ROOT to that.
 STATIC_ROOT=/tmp/crashstats-static/
+
+# For connecting to local S3, we need to connect insecurely.
+AWS_SECURE=False
```
willkg: We require this setting, but since it would be super bad if this ever made it to a server environment by accident, we put it in this file as opposed to `docker/config/docker_common.env`.

webapp-django/bin/ci.sh
```diff
@@ -18,11 +18,12 @@ echo "Linting..."
...
 export CACHE_LOCATION="crashstats"
-export OLD_PYTHONPATH=$PYTHONPATH
-export PYTHONPATH=../:$PYTHONPATH
-FORCE_DB=true python manage.py test --noinput
-export PYTHONPATH=$OLD_PYTHONPATH
+# Change AWS_HOST to empty string so tests use boto.s3.connect_to_region
+export AWS_HOST=
+
+PYTHONPATH=../:$PYTHONPATH FORCE_DB=true python manage.py test --noinput
 echo "Tests finished."
```
willkg: The tests for the symbol upload view mock boto. However, they mock the `boto.s3.connect_to_region` way of getting a connection instance rather than the `boto.connect_s3` way. Rather than adding a lot of complexity to the tests to mock both situations (Travis effectly runs with `AWS_HOST=` and Circle runs with `AWS_HOST=localstack-s3`), we set `AWS_HOST=` so all test systems test the path the server environments will be running.

webapp-django/crashstats/settings/base.py
```diff
@@ -536,6 +536,9 @@ def path(*dirs):
 # Credentials for being able to make an S3 connection
 AWS_ACCESS_KEY = config('AWS_ACCESS_KEY', None)
 AWS_SECRET_ACCESS_KEY = config('AWS_SECRET_ACCESS_KEY', None)
+AWS_HOST = config('AWS_HOST', None)
+AWS_PORT = config('AWS_PORT', 0, cast=int)
+AWS_SECURE = config('AWS_SECURE', True, cast=bool)
```
willkg: These three get added to settings, but the code only uses them iff `AWS_HOST` and `AWS_PORT` have truthy values.

webapp-django/crashstats/symbols/views.py
```diff
@@ -67,12 +67,26 @@ def unpack_and_upload(iterator, symbols_upload, bucket_name, bucket_location):
...
+        # variables should not be set.
+        params.update({
+            'host': settings.AWS_HOST,
+            'port': settings.AWS_PORT,
+            'is_secure': settings.AWS_SECURE,
+        })
+        conn = boto.connect_s3(**params)
+
+    else:
+        conn = boto.s3.connect_to_region(bucket_location, **params)
```
willkg: The comment here should be clear--if not, we should improve it.

This is essentially what we do in the `socorro.external.boto.connection_context` classes, too, though it's hard to see that because we have a non-trivial class hierarchy there.

willkg:
^^^ I added comments to the code changes. I'm feeling pretty comfortable with this change and I think the tests cover the important bits, so if @peterbe doesn't get to reviewing it today, I'll self-r+ it.

willkg:
The server-environment aspect of the changes here are covered by Circle CI and Travis tests. The local dev environment aspect aren't covered at all. I think that's ok for now. The symbol upload code isn't going to be in Socorro for much longer. Writing the necessary test isn't hard, but it involves mocking so it's of limited value. It's super easy to test in a local dev environment. The code in question hasn't changed much in the last 3 years.

For all those reasons, I claim this change is low risk and good enough as is. self-r+.

PR #3993, opened 2017-09-15
bug 1400076 - add Antenna service

willkg:
This completes the local development crash ingestion pipeline by adding the
Antenna service and pointing it to the local S3 `dev_bucket`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -77,6 +77,22 @@ services:
   # External services
   # -----------------------------
 
+  # https://hub.docker.com/r/mozilla/antenna/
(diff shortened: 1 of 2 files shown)


willkg:
This doesn't include the bridge. I was thinking I'd land this now and we could discuss how we want to bridge services and then do that in a separate PR.

willkg:
This doesn't affect anything--it just sets us up to do full-pipeline work which no one is doing right now and it unblocks the penetration testers. If there are issues, we can fix them as they come up.

I'm self-r+ing this.

PR #3994, opened 2017-09-15
fixes bug 1400301 Add rtc::FatalMessage -- webrtc.org's impl for RTC_CHECK/RTC_DCHECK/etc

jesup:
Any assertion within media/webrtc/trunk/webrtc (webrtc imported code) ends up in rtc::FatalMessage::~FatalMessage right now, regardless of which assert it is.

### DIFF
--- a/socorro/siglists/irrelevant_signature_re.txt
+++ b/socorro/siglists/irrelevant_signature_re.txt
@@ -46,6 +46,7 @@ nvmap@0x.*
 org\.mozilla\.f.*-\d\.apk@0x.*
 PR_WaitCondVar
 RaiseException
+rtc::FatalMessage.*
 RtlpAdjustHeapLookasideDepth
 std::_Atomic_fetch_add_4
 std::panicking::.*


willkg:
Can you write up a bug in the Socorro product for this along with links to example crashes?

jesup:
Example: https://bugzilla.mozilla.org/show_bug.cgi?id=1344557

jesup:
Bug 1400301

willkg:
@jesup Is this the kind of signature you were expecting with this change?:

```
app@240f7981f52c:/app$ python -m socorro.signature a76e0171-8333-495f-bd30-507550170915
Crash id: a76e0171-8333-495f-bd30-507550170915
Original: mozalloc_abort | abort | rtc::FatalMessage::~FatalMessage
New:      mozalloc_abort | abort | rtc::PlatformThread::Start
Same?:    False
```

jesup:
yes, that would be fine.  We could leave out the mozalloc_abort | abort | part, I just need it to be unique and not all hidden in FatalMessage

willkg:
Let's go back a bit. Seems like you're trying to get signatures that don't end on `rtc::FatalMessage`, but rather continue and show you the next frame.

Is `rtc::FatalMessage` an important part of the signature? Is this:

```
mozalloc_abort | abort | rtc::FatalMessage::~FatalMessage | rtc::PlatformThread::Start
```
preferable to this:

```
mozalloc_abort | abort | rtc::PlatformThread::Start
```

If so, we should add the change to the prefix list and not the irrelevant list.

Regarding `mozalloc_abort` and `abort`, we don't have the ability to drop signature parts if certain strings show up in the signature without writing additional rules. So unless you're saying `mozalloc_abort` and `abort` shouldn't show up in any signatures, we can't do anything about those.

willkg:
@jesup ^^^ Any thoughts?

jesup:
I don't really care if rtc::FatalMessage shows up in the signature.  Perhaps it adds a little (separating it from other crashes in the same function), but those other crashes wouldn't have 'abort' so there really is no gain from rtc::FatalMessage.
Summary:  mozalloc_abort | abort | rtc::PlatformThread::Start is preferable

willkg:
@jesup Awesome--thank you!

I'll get this merged and pushed today.

PR #4001, opened 2017-09-18
bug 1399979 - move environment vars around

willkg:
This moves another couple of environment variables that should never get set
this way in a server environment into the local_development.env file where they
should be.

### DIFF
--- a/docker/config/docker_common.env
+++ b/docker/config/docker_common.env
@@ -2,8 +2,8 @@
 # Environment-specific configuration
 # -------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


willkg:
This solely affects the local dev environment.

I tested this PR by doing the following:

1. Running `make dockertest`. Works fine. Travis and Circle both pass, too.
2. Running `docker-compose up webapp` and clicking around the webapp. This works fine.
3. Running `docker-compose up processor` and processing crashes and making sure they're saved to local S3.

This is low-risk, so self-r+.

PR #4002, opened 2017-09-18
fixes bug 1399979 - rename base config files

willkg:
This renames the two base config files that we use for local development
environment to make it clearer what they do.


willkg:
This is pretty straight-forward file rename followed by updating a bunch of places where the files are used or referred to. This only affects the local dev environment and Circle CI.

self-r+.

PR #4003, opened 2017-09-18
fixes bug 1385966 abort signature

willkg:
This fixes a signature rule that's regularly been throwing errors for a while now, though not in huge numbers.

It also renames `signature_utilities.py` to `rules.py` which is shorter, less redundant, and a better name.

It also creates a new function `to_printable_string`, though that's a bit of a misnomer since that function also removes some printable characters and drops all unicode data.

Two things:

1. I'm on the fence about whether this is a good idea or not. Dropping unicode data seems like a good idea so as to reduce bucketing problems, but maybe I'm wrong about that.
2. I think we definitely want data sanitizing done to any text we get from who-knows-where that we add to signatures and possibly other parts of the processed crash. I wrote `to_printable_string`, but maybe that should really be `sanitize_string`? Maybe it should be different in some manner?

@lonnen: Thoughts on this?


lonnen:
The code looks good. I think that it's better to be explicit about how we handle unicode, and to make sure that we can handle it.

The current name and implementation are fine for dealing with the common error case. I'd ask two additional questions that may be better for follow on fixes:

1. Are we certain that dropping the unicode characters is the right outcome, versus replacing them with one or more ASCII char?

2. Should we be expanding this into real sanitation and doing it elsewhere in the process or to other fields? For example, we may have to deal with unicode file names at some point.

lonnen:
For 1, is there an advantage to something like https://pypi.python.org/pypi/Unidecode/ instead?

willkg:
So, I'm conflating a few issues here. The bug this fix needs to solve relates to `abort_message` strings like this:

```
u'unable to find a usable font (微软雅黑)'
```

That doesn't jive because it's got unicode in it and the particular way we're composing strings doesn't jive with strings and unicode.

So we need to change the way it's composing strings and/or deal with the unicode.

Other parts of `AbortSignature` remove bits of the `abort_message`. So another way to tackle this specific bug is to remove the parenthesized bit for messages in this shape. Thenhen the `abort_message` would get converted to `u'unable to find a usable font'`. That'd work.

But instead of doing that, I stepped back and wondered about the general problem and what kinds of things we get in `abort_message`.

I think that in the context of generation signatures, any unicode is coming from a localized message for a user. Because we want to bucket around problems, I think it's a mistake to have localized strings in the signatures. Thus I claim we want to categorically remove all unicode characters from a string that will be part of a signature as opposed to converting them to some ascii equivalent.

If that's right, then we don't want to use `Unidecode`. We really want to drop unicode characters altogether.

Towards that, I was thinking it's easier to remove unicode from all `abort_message` strings rather than add another clause for strings that start with `u'unable to find a usable font'`.

Then it occurred to me that we really want to be sanitizing any strings in the signature that come from the `raw_crash`, but maybe we should spin that off into a separate bug rather than do that here.

So, I think I've argued myself into the following:

1. we don't want to use `Unidecode`--we want to drop unicode characters altogether
2. we want to drop unicode characters from any `abort_message` string
3. we should push the other sanitization off to a separate bug

@lonnen Does that sound ok to you? Does it sound like I'm breaking anything with the above theories/assumptions?

willkg:
Closing this out in favor of PR #4015.

PR #4004, opened 2017-09-18
fixes bug 1394834 - fix alembic in local dev environment

willkg:
Prior to this fix, alembic had the wrong sqlalchemy.url, so it was tricky to do
migration testing in the local dev environment.

### DIFF
--- a/docker/config/alembic.ini
+++ b/docker/config/alembic.ini
@@ -11,8 +11,6 @@ file_template = %%(rev)s_%%(slug)s
 # the 'revision' command, regardless of autogenerate
 # revision_environment = false
 
-sqlalchemy.url = postgresql://postgres:aPassword@postgresql:5432/socorro_migration_test
-
 # Logging configuration
 [loggers]
 keys = root,sqlalchemy,alembic
--- a/docker/config/local_dev.env
+++ b/docker/config/local_dev.env
@@ -29,12 +29,7 @@ resource.postgresql.database_port=5432
 secrets.postgresql.database_username=postgres
 secrets.postgresql.database_password=aPassword
 
-# resource.postgresql.database_replication_hostname=
-# secrets.postgresql.database_replication_password=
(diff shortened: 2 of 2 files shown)


docker/config/alembic.ini
```diff
@@ -11,8 +11,6 @@ file_template = %%(rev)s_%%(slug)s
 # the 'revision' command, regardless of autogenerate
 # revision_environment = false
 
-sqlalchemy.url = postgresql://postgres:aPassword@postgresql:5432/socorro_migration_test
```
willkg: We can take this out because our `alembic/env.py` pulls this value from the environment.

docker/config/local_dev.env
```diff
@@ -29,12 +29,7 @@ resource.postgresql.database_port=5432
 secrets.postgresql.database_username=postgres
 secrets.postgresql.database_password=aPassword
 
-# resource.postgresql.database_replication_hostname=
-# secrets.postgresql.database_replication_password=
-# secrets.postgresql.database_replication_username=
```
willkg: I don't know what these were doing, so I removed them.

docker/config/local_dev.env
```diff
@@ -29,12 +29,7 @@ resource.postgresql.database_port=5432
...
-
-# postgres things for alembic
-# FIXME - sqlalchemy.url=postgresql://postgres:aPassword@postgresql:5432/socorro_migration_test
```
willkg: We need to set this here so that migrations work in the local dev environment.

When the tests run, they override this value with the appropriate value for migration testing, so this works fine and doesn't affect anything else.

willkg:
These changes only affect the local dev environment and the tests. I tested the local dev environment:

```
app@...:/app$ alembic --config docker/config/alembic.ini history
<so much history>
```

Circle runs the tests using this environment and that worked fine.

Given that, this is low risk, so I'm going to self-r+ it.

PR #4007, opened 2017-09-20
bug 1401669 - allow filtering for crashing thread name in super search

johnp:
This is my first commit here and I have <1h experience with the code base, but this seemed to be easy enough(?).  

I don't know the exact meaning of most of the members in super_search_fields and therefore set them based on other entries for now.


socorro/schemas/crash_report.json
```diff
@@ -307,6 +307,10 @@
                 "crashing_thread": {
                     "type": "object",
                     "properties": {
+                        "thread_name": {
+                            "type": ["string", "null"],
+                            "description": "Name of the crashing thread."
+                        },
```
willkg: The changes to this file aren't needed if you're just adding a field for querying to Super Search.

willkg:
I need to spend some time looking at the contents of that field to determine whether setting it up this way is ok or not and what changes I'd make.

Is there any urgency for this change? Is this blocking work that you're doing?

johnp:
No, no urgency for this, just something that I felt was missing when trying to find similar crash reports.

willkg:
I will try to get to this soon--probably at some point in the next couple of weeks.

willkg:
I haven't forgotten about this, but haven't had a chance to get to it, yet, either. I'm really sorry about that.

I'll try to get to it soon.

willkg:
As an update, this is waiting on me to figure out how super search fields work so I can effectively review this. I haven't had time to do that, yet. Maybe in the next month.

willkg:
I'm sorry I haven't gotten to this, yet. I'm still in the same boat. We're pushing hard on switching to a new infrastructure and I'll be working on that this month. After that, I hope to come up to speed on the bits that this changes.

willkg:
I was trying to get this to work and then couldn't figure out why it wasn't working. What's going on is that we're using `ESCrashStorageRedactedJsonDump` which dumps the `crashing_thread` part of `json_dump`. As a result of that, the `thread_name` never makes it to the index.

We need a more involved fix than this PR. Probably need to extract the `thread_name` from the `json_dump` and insert it into the processed crash and then index that. Or something along those lines.

I need to spend some more time looking at the code before I know what we should do.

Given that, I'm going to close this out. We still have the bug open. I'll work on it some more next week.

Sorry that it took so long to get to!

PR #4008, opened 2017-09-20
Support stdin with scripts/add_crashid_to_queue.py

willkg:


willkg:
This adds support for passing in crash ids via stdin. This mirrors changes I've done in other scripts. So now you can do this:

```
$ ./scripts/fetch_crashids.py --num=5 | xargs ./scripts/add_crashid_to_queue.py socorro.normal
```

as well as:

```
$ ./scripts/fetch_crashids.py --num=5 | ./scripts/add_crashid_to_queue.py socorro.normal
```

(Though since that's fetching crash ids from prod and not saving the crash data, I'm not sure why you'd want to do this. But it does work.)

willkg:
This only affects the local development environment, so I'm going to self-r+ this and merge it.

PR #4009, opened 2017-09-21
bug 1390530 - Add https scheme to crash-stats CSP srcs

g-k:
fixes: https://bugzilla.mozilla.org/show_bug.cgi?id=1390530

As April pointed out these domains the HSTS preloaded, so this will only help people with older browsers that don't support HSTS.

This also makes things consistent with the img-src.

r? @peterbe or @willkg

### DIFF
--- a/webapp-django/crashstats/settings/base.py
+++ b/webapp-django/crashstats/settings/base.py
@@ -731,8 +731,8 @@ CSP_OBJECT_SRC = (
 )
 CSP_SCRIPT_SRC = (
     "'self'",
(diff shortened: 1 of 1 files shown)


willkg:
The changes look fine to me. I verified that this doesn't break the local development environment. It probably would if protocols were relative, but it looks like they're not.

This won't raise any issues with the tests, so the next place we can verify goodness is after it's deployed to -stage.

I'll make sure to check it there later today.

willkg:
Thank you!

PR #4010, opened 2017-09-21
Rename signature/signature_utilities.py to rules.py

willkg:
I pulled this out of #4003. Figured I'd land this separately.


willkg:
This switches two file names. It's the kind of change where if it was bad, Travis and Circle would be racing to tell us. They didn't.

willkg:
This switches two file names. It's the kind of change where if it was bad, Travis and Circle would be racing to tell us. They didn't.

PR #4011, opened 2017-09-22
fixes bug 1402190 - add --different-only flag to signature generation cli

willkg:
This adds a --different-only flag to the signature generation cli that restricts
output to just signatures that changed.

This also removes caching of the "normalized" form of stack frames because that
prevents them from getting normalized which is annoying when you're changing how
normalization works.


willkg:
This only affects the signature generator cli and it's pretty minimal change, so it's low risk.

PR #4014, opened 2017-09-22
Update signature generation documentation

willkg:
We had a wiki page that sort of walked through the process for requesting
signature generation changes. We changed that wiki page to point to these docs,
but didn't have a process in these docs. This adds that.

This also cleans up the process for reviewing changes to signature generation.


willkg:
These are just docs changes. self-r+.

PR #4015, opened 2017-09-22
fixes bug 13785966 - fix UnicodeEncodeError in AbortSignature

willkg:
This fixes the UnicodeEncodeError in AbortSignature by stripping abort_message
of unicode characters.

This also fixes a bucketing issue for "unable to find a usable font (...)" where
the ... part contains a number of different fonts, thus all these crashes are
falling into separate buckets. This removes that parenthesized bit.

### DIFF
--- a/socorro/signature/rules.py
+++ b/socorro/signature/rules.py
@@ -10,6 +10,7 @@ import ujson
 
 from socorro import siglists
 from socorro.lib.treelib import tree_get
+from socorro.lib.util import drop_unicode
 
 
 SIGNATURE_MAX_LENGTH = 255
@@ -525,13 +526,24 @@ class AbortSignature(Rule):
             # Recent crash reports added some irrelevant information at the
             # beginning of the abort message. We want to remove that and keep
             # just the actual abort message.
-            abort_message = abort_message.split('###!!! ABORT:', 1)[1].strip()
+            abort_message = abort_message.split('###!!! ABORT:', 1)[1]
 
         if ': file ' in abort_message:
             # Abort messages contain a file name and a line number. Since
             # those are very likely to change between builds, we want to
             # remove those parts from the signature.
-            abort_message = abort_message.split(': file ', 1)[0].strip()
+            abort_message = abort_message.split(': file ', 1)[0]
+
+        if 'unable to find a usable font' in abort_message:
+            # "unable to find a usable font" messages include a parenthesized localized message. We
+            # want to remove that. Bug #1385966
+            open_paren = abort_message.find('(')
+            if open_paren != -1:
+                end_paren = abort_message.rfind(')')
+                if end_paren != -1:
+                    abort_message = abort_message[:open_paren] + abort_message[end_paren + 1:]
+
+        abort_message = drop_unicode(abort_message).strip()
 
         if len(abort_message) > 80:
             abort_message = abort_message[:77] + '...'
--- a/socorro/unittest/signature/test_rules.py
+++ b/socorro/unittest/signature/test_rules.py
@@ -1281,58 +1281,73 @@ class TestAbortSignature:
         expected_sig = 'Abort | {}... | hello'.format('a' * 77)
         assert processed_crash['signature'] == expected_sig
 
-    def test_action_success_remove_unwanted_parts(self):
-        rule = AbortSignature()
+    @pytest.mark.parametrize('abort_msg, expected', [
+        # Test with just the "ABOR" thing at the start
+        (
+            '[5392] ###!!! ABORT: foo bar line 42',
+            'Abort | foo bar line 42 | hello'
+        ),
 
-        raw_crash = {}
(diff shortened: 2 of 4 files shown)


willkg:
Output from this from crashes that don't have Unicode (none of them got processed):

```
app@2d67eadfd913:/app$ ./scripts/fetch_crashids.py --signature-contains="unable" | python -m socorro.signature
Crash id: de2c8396-bd87-4d92-83ef-fabc70170921
Original: Abort | unable to find a usable font (sans-serif) | mozalloc_abort | NS_DebugBreak | gfxFontGroup::GetDefaultFont
New:      Abort | unable to find a usable font | mozalloc_abort | NS_DebugBreak | gfxFontGroup::GetDefaultFont
Same?:    False

Crash id: acc79f3b-f35e-4dc2-8b46-ce9b20170921
Original: Abort | unable to find a usable font (serif) | mozalloc_abort | NS_DebugBreak | gfxFontGroup::GetDefaultFont
New:      Abort | unable to find a usable font | mozalloc_abort | NS_DebugBreak | gfxFontGroup::GetDefaultFont
Same?:    False

Crash id: a71a55e1-c441-4562-92c4-759d70170921
Original: Abort | unable to find a usable font (a) | mozalloc_abort | NS_DebugBreak | gfxFontGroup::GetDefaultFont
New:      Abort | unable to find a usable font | mozalloc_abort | NS_DebugBreak | gfxFontGroup::GetDefaultFont
Same?:    False

Crash id: b4b1c344-c6fd-43bf-8c97-529550170921
Original: Abort | unable to find a usable font (Segoe UI) | mozalloc_abort | NS_DebugBreak | gfxFontGroup::GetDefaultFont
New:      Abort | unable to find a usable font | mozalloc_abort | NS_DebugBreak | gfxFontGroup::GetDefaultFont
Same?:    False
```

The ones that do have unicode, will get signature-ized similarly.

@lonnen Does this look ok?

socorro/unittest/signature/test_rules.py
```diff
@@ -1281,58 +1281,73 @@ def test_action_success_long_message(self):
...
-        assert processed_crash['signature'] == 'Abort | foo bar line 42 | hello'
-
-        # Test with a file name and line number.
-        raw_crash['AbortMessage'] = (
-            '[7616] ###!!! ABORT: unsafe destruction: file '
-            'c:/builds/moz2_slave/m-rel-w32-00000000000000000000/build/src/'
-            'dom/plugins/ipc/PluginModuleParent.cpp, line 777'
-        )
-        processed_crash['signature'] = 'hello'
+    @pytest.mark.parametrize('abort_msg, expected', [
```
willkg: I redid this test because it's a bunch of cases with the same structure.

One thing about this test is that the input data in some cases involves very very long strings. I did this to keep the existing string. We could probably shorten the strings to smaller versions, though.

socorro/unittest/signature/test_rules.py
```diff
@@ -1281,58 +1281,73 @@ def test_action_success_long_message(self):
...
+                'SessionStore.jsm, line 1459'
+            ),
+            'Abort | hello'
+        ),
+
+        # Test with "unable to find a usable font" case
+        (
+            u'unable to find a usable font (\u5fae\u8f6f\u96c5\u9ed1)',
+            'Abort | unable to find a usable font | hello'
+        ),
```
willkg: This is the new test case that was added.

lonnen: This seems fine. I was worried earlier about destroying information and if it was appropriate to combine these. Having some time to consider it, I now suspect it is a good move and like the new signature.

socorro/unittest/signature/test_rules.py
lonnen: no preference here

socorro/signature/rules.py
```diff
@@ -525,13 +526,24 @@ def action(self, raw_crash, processed_crash, notes):
...
+        if 'unable to find a usable font' in abort_message:
+            # "unable to find a usable font" messages include a parenthesized localized message. We
+            # want to remove that. Bug #1385966
+            open_paren = abort_message.find('(')
+            if open_paren != -1:
+                end_paren = abort_message.rfind(')')
+                if end_paren != -1:
+                    abort_message = abort_message[:open_paren] + abort_message[end_paren + 1:]
+
+        abort_message = drop_unicode(abort_message).strip()
```
lonnen: +1. I was worried about dropping unicode before, based on an intuition that I don't understand all the possible error messages and where unicode may be important. After some consideration I think this is a good, general way to handle it, the specifics should be preserved in individual files, and we can address this later if it becomes a problem.

willkg:
Bah--forgot to add the tests. Adding now and waiting for travis.

willkg:
Thank you!

willkg:
Bah. I got the bug number wrong, too.

PR #4019, opened 2017-09-26
fixes bug 1280374 - reset additional_minidumps every time

willkg:
If you process a crash, it sets "additional_minidumps". However, if you
re-process a crash, it appends to that field. That's totally not the right thing
to do.

This fixes it so that it always resets the value of that field.


willkg:
The only effect of this is that it initializes the `additional_minidumps` field to `[]` during reprocessing. That's absolutely what we want to do.

PR #4020, opened 2017-09-27
fixes bug 1403246 - implements paging in fetch_crashids.py

willkg:
Super Search lets you fetch at most 1000 results in a query. This fixes
fetch_crashids.py so that you can specify a maximum number of results to get or
"all". In both cases, it'll run as many Super Search queries as it needs to
fetch all the results you want.

This further fixes the code to retry on 429s which is handy.

This also also fixes the code to be more resilient of errors. Instead of losing
everything because of an error and having to do it again, you only lose what
hasn't been requested, yet.


socorro/scripts/fetch_crashids.py
```diff
@@ -29,17 +31,56 @@
...
+
+        # If there are no more crash ids to get, we return
+        total = resp.json()['total']
+        if not hits or crashids_count >= total:
+            return
+
+        # Get the next page
+        params['_results_offset'] += MAX_PAGE
+        if num_results != 'all':
+            params['_results_number'] = min(MAX_PAGE, total - crashids_count)
```
willkg: It's possible there's some off-by-one shenanigans in here. Maybe one of the `>=` should be a `>`. It's hard to tell. I thought about adding tests for this, but they'd have to be heavily mocked and at that point, it didn't seem worth it.

Someone else can write those tests if we change our minds.

socorro/scripts/fetch_crashids.py
```diff
@@ -29,17 +32,77 @@
...
-        crashids = [hit['uuid'] for hit in hits]
-        return crashids
 
-    raise Exception('Bad response: %s %s' % (resp.status_code, resp.content))
+        for hit in hits:
+            crashids_count += 1
+            yield hit['uuid']
+
+            # If we've gotten as many crashids as we need, we return
+            if crashids_count >= num_results:
```
willkg: This works for non-all cases with normal comparison of two integers.

This also works with the `all` case because `num_results` will be `INFINITY` and since nothing is bigger than `INFINITY` (as we've coded it), this will always return `False` and allow the loop to continue.

socorro/scripts/fetch_crashids.py
```diff
@@ -29,17 +32,77 @@
...
-    if resp.status_code == 200:
+    # Set up retrying on 429s
+    retries = Retry(total=32, backoff_factor=1, status_forcelist=[429])
+    session = requests.Session()
+    session.mount('http://', HTTPAdapter(max_retries=retries))
+    session.mount('https://', HTTPAdapter(max_retries=retries))
+
+    # Set up first page
+    params['_results_offset'] = 0
+    params['_results_number'] = min(MAX_PAGE, num_results)
```
willkg: We never want to request more than 1000 results because then Super Search gets cranky. So we want the smaller of 1000 or "the number of results we want to get". This works for non-all cases with normal comparison of integers.

This works for the `all` case because `num_results` will be `INFINITY` and the smaller of 1000 and `INFINITY` will always be 1000. Thus for the first page, we'll request 1000 results and everything is super duper.

socorro/scripts/fetch_crashids.py
```diff
@@ -29,17 +32,95 @@
...
+        params['_results_number'] = min(
+            # MAX_PAGE is the maximum we can request
+            MAX_PAGE,
+
+            # The number of results Super Search can return to us that is hasn't returned so far
+            total - crashids_count,
+
+            # The numver of results we want that we haven't gotten, yet
+            num_results - crashids_count
+        )
```
willkg: This works in the non-all case because it's just numbers.

This works in the `all` case, because `num_results` is `INFINITY` and we've coded `INFINITY` to return `INFINITY` when in the left-hand-side of a subtraction operation so that number will always be bigger than the other two.

willkg:
^^^ That redoes some of the "all" awkwardness by introducing an "INFINITY" that let's us treat it like a "number".

Maybe that's too fancy, but it does make `fetch_crashids` easier to read.

willkg:
This only affects `fetch_crashids.py` and it works for me. I ran the following:

```
app@7ae8f1f1d798:/app$ ./scripts/fetch_crashids.py  | wc -l
100
app@7ae8f1f1d798:/app$ ./scripts/fetch_crashids.py --num=1010  | wc -l
1010
app@7ae8f1f1d798:/app$ ./scripts/fetch_crashids.py --num=1010 --signature-contains=shutdownhang  | wc -l
1010
app@7ae8f1f1d798:/app$ ./scripts/fetch_crashids.py --num=1010 --signature-contains=shutdownhangs | wc -l
0
app@7ae8f1f1d798:/app$ ./scripts/fetch_crashids.py --num=all --signature-contains=shutdownhangs | wc -l
0
app@7ae8f1f1d798:/app$ ./scripts/fetch_crashids.py --num=all --signature-contains=shutdownhang | wc -l
29447
```

There are no signatures with "shutdownhangs" (with an "s"), so we get back 0 results.

At the time I ran this, there were 29447 crashes with "shutdownhang" in the signature yesterday. Obviously that'll change when yesterday is a different day.

The rest of the run lines make sure that we get back as many results as we asked for.

Given that, I'm going to self-r+ this and land it.

PR #4022, opened 2017-09-28
Add FennecAndroid to local development environment

willkg:
It's important that we test with FennecAndroid, too, so this adds that to the
list of things to pull down in a local development environment.


willkg:
This just affects the development environment and it fixes an issue I was having when looking into bug 1403172. Given that, I'm going to self-r+ it.

PR #4025, opened 2017-10-02
Bug 1292594 - Remove test_that_products_are_sorted_correctly

davehunt:
After some discussion it has been decided that this test is not providing enough value, especially as it has been marked as expected to fail for at least six months.


chartjes:
@davehunt 👍

willkg:
Thank you!

PR #4026, opened 2017-10-04
fixes bug 1380013 - switch to print function

willkg:
This goes through the codebase and switches things to use print function.
Technically, in a few areas, I didn't add the "`from __future__...`" header, so
it's still a print statement, but I wanted to minimize the number of "`from
__future__...`" things I added.

Also, this removes `socorro/external/fetch_app.py` which we don't use. There's a
script that does essentially the same thing, but is more flexible.


willkg:
Thank you!

PR #4027, opened 2017-10-06
bug 1405675 - switch to pytest for webapp [no rush]

willkg:
This drops django-nose and adds pytest-django, sets up configuration for using
pytest to run webapp tests, adds some helpful comments, fixes a test, and
updates the relevant shell scripts.


willkg:
I ran everything in the local development environment and it works super there. We'll see if it works in Circle CI and Travis CI and I'll fix anything that doesn't.

willkg:
Well hot diggity--worked in both CI systems first try.

requirements.txt
```diff
@@ -333,3 +330,6 @@ Pygments==2.2.0 \
 whitenoise==3.3.0 \
     --hash=sha256:1d62a003a0ab747de96da45c831cbb512dcb7f69c1ef0bd20b1cd4ae45d8a0c4 \
     --hash=sha256:d098327276de6fd189398a7bcb95789d1bb2d41b3e011eeae4562f6b1a107dd4
+pytest-django==3.1.2 \
+    --hash=sha256:038ccc5a9daa1b1b0eb739ab7dce54e495811eca5ea3af4815a2a3ac45152309 \
+    --hash=sha256:00995c2999b884a38ae9cd30a8c00ed32b3d38c1041250ea84caf18085589662
```
willkg: We removed django-nose which provided nose-specific Django-y things and added pytest-django which provides pytest-specifc Django-y things.

setup.cfg
```diff
@@ -4,9 +4,14 @@ exclude=webapp-django/crashstats/*/migrations/*,docs/conf.py,socorro/unittest/te
 max-line-length = 100
 
 [tool:pytest]
-addopts = -rsxX --tb=native
+# -rsxX        - show skipped, failed, and passed tests
+# --tb=native  - print native traceback
+# -p no:django - disable the pytest-django plugin for Socorro tests
+addopts = -rsxX --tb=native -p no:django
```
willkg: This file (`setup.cfg`) is at the root of the repository. That's the directory you sit in when running pytest on the unit tests located in `socorro/unittest/`.

We've added pytest-django to the Python environment, but really don't want it loading when we run the regular Socorro tests. We only want it loading when running the webapp tests. So adding this `-p` option prevents it from loading.

While doing that, I figured I'd add some descriptive text about the other two options we do.

setup.cfg
```diff
@@ -4,9 +4,14 @@ exclude=webapp-django/crashstats/*/migrations/*,docs/conf.py,socorro/unittest/te
...
 norecursedirs = .git docs scripts config docker __pycache__
 testpaths = socorro/unittest/
+# NOTE(willkg): These values are used by Travis. Once we ditch Travis, we can
+# probably either ditch these values or fix them or both.
```
willkg: Figured I'd add this note so that future me doesn't spend 10 minutes trying to figure out why these values are the way they are like present me just did because past me didn't think to tell the rest of us what's going on.

webapp-django/bin/ci.sh
```diff
@@ -18,12 +18,6 @@ echo "Linting..."
 git ls-files crashstats | grep '\.py$' | xargs flake8 | bin/linting.py
 
 echo "Starting tests..."
-# Override configuration with better defaults for the tests
-export SECRET_KEY="doesn't matter, tests"
-export CACHE_BACKEND="django.core.cache.backends.locmem.LocMemCache"
-export CACHE_LOCATION="crashstats"
-# Change AWS_HOST to empty string so tests use boto.s3.connect_to_region
-export AWS_HOST=
```
willkg: These are all in `webapp-django/crashstats/settings/test.py` with all the other test-specific settings. Figured it's easier to have one place for test-specific settings.

webapp-django/crashstats/crashstats/urls.py
```diff
@@ -105,8 +105,7 @@
     url(
         r'^status/$',
         RedirectView.as_view(
-            pattern_name='monitoring:index',
-            permanent=not settings.DEBUG,
```
willkg: I have no idea why we switch the permanence of this depending on the value of `settings.DEBUG`. That seems unhelpful to me. Also, it was the lone test that didn't work in pytest. I didn't look into why, though. I'm guessing one of a couple of things:

1. pytest bootstrapping is different, so `settings.DEBUG` has a different value at module-load time than what the test is testing for
2. `settings.DEBUG` has a different value when running tests in pytest vs. in nose

Number 1 seems more likely to me.

Anyhow, seemed easier to just ditch this selective permanence.

webapp-django/pytest.ini
```diff
@@ -0,0 +1,6 @@
+# pytest configuration for running webapp tests
+[pytest]
+# -rsxX        - show skipped, failed, and passed tests
+# --tb=native  - print native traceback
+addopts = -rsxX --tb=native
+DJANGO_SETTINGS_MODULE = crashstats.settings.test
```
willkg: We run the tests for the webapp from within this directory, so pytest finds this file first and these become the settings used.

Osmose: We should remove the bit in `settings/__init__.py` that apply test settings since this now handles setting those.

willkg: Good catch!

willkg:
^^^ That fixes that extraneous import. Assuming Travis and Circle are happy with that, this is ready for re-review.

Regarding the r+wc kind of thing, either way is fine. There are pros and cons either way and I don't have a strong preference.

webapp-django/crashstats/crashstats/urls.py
adngdb: The reason for this was to avoid having a permanent redirect be saved into your browser's cache during development of that feature. It made development easier, but is probably very useless now.

PR #4028, opened 2017-10-08
bug 1377254 - flake8 socorro code [no rush]

willkg:
This fixes flake8 issues for a bunch of socorro directories. After this, we have three to do: socorro/database/, socorro/external/, and socorro/unittest/. Those are somewhat involved, so I'll do them in separate PRs.

Why do this? It behooves us to be able to run flake8 as part of CI. Plus it's helpful when the code follows a single set of conventions.

Could we do more to make the code conform to rules and conventions? Sure. We can do that in other passes if people feel strongly.


socorro/app/fetch_transform_save_app.py
```diff
@@ -33,11 +33,7 @@
 from configman.converters import class_converter
 
 from socorro.lib.task_manager import respond_to_SIGTERM
-from socorro.app.generic_app import App, main  # main not used here, but
-                                               # is imported from generic_app
-                                               # into this scope to offer to
-                                               # apps that derive from the
-                                               # class defined here.
+from socorro.app.socorro_app import App
```
willkg: I adjusted code that imported `main` from this module to import it from the source so we don't need to import `main` here.

socorro/app/generic_app.py
```diff
@@ -1,12 +0,0 @@
...
-# file, You can obtain one at http://mozilla.org/MPL/2.0/.
-
-"""This module used to contain the overly complicated definition of the
-Socorro App class definitions.  That system has been moved to socorro_app.py.
-The following two imports into this module are for backwards compatibility.
-"""
-
-from socorro.app.socorro_app import main
-from socorro.app.socorro_app import App
-
```
willkg: I nixed this module and pointed anything that was pointing here to `socorro.app.socorro_app`.

socorro/submitter/submitter_app.py
```diff
@@ -158,15 +152,6 @@ def new_crashes(self):
             yield (((prefix, crash_pathnames), ), {})
 
 
-# this class was relocated to a more appropriate module and given a new name.
-# This import is offered for backwards compatibilty.  Note, that there has also
-# been an internal change to the required config, with the source
-# implementation moved into a namespace
-from socorro.external.postgresql.new_crash_source import (
-    DBCrashStorageWrapperNewCrashSource as DBSamplingCrashSource
-)
```
willkg: Nothing uses this name from this module, so I nixed this.

lonnen: 👍

willkg:
^^^ That renames the `generic_app.rst` file.

willkg:
Thank you!

PR #4029, opened 2017-10-09
bug 1377254 - flake8 socorro/database/ [no rush]

willkg:
This removes two files that aren't used anywhere, then tidies up the third.


socorro/database/database.py
```diff
@@ -1,200 +0,0 @@
-# This Source Code Form is subject to the terms of the Mozilla Public
```
willkg: There's no evidence of `from socorro.database` or `from socorro import database` (code or configuration) except for uses of `transaction_executor` which is used in a lot of places, so I removed this file.

socorro/database/cachedIdAccess.py
```diff
@@ -1,168 +0,0 @@
-# This Source Code Form is subject to the terms of the Mozilla Public
```
willkg: There's no evidence of `cachedIdAccess` in the codebase or configuration at all, so I removed this file.

willkg:
Thank you!

PR #4030, opened 2017-10-09
bug 1377254 - flake8 socorro/external/ [no rush]

willkg:
This PR flake8s `socorro/external/`. In the process of doing that, I nixed some code, and removed some "here for backwards compatibility" kinds of things. This is less straight-forward than other lint-related PRs.


socorro/external/fs/crashstorage.py
```diff
@@ -24,16 +24,10 @@
...
 from socorro.lib.ooid import dateFromOoid, depthFromOoid
-from socorro.lib.datetimeutil import utc_now
+from socorro.lib.datetimeutil import utc_now, JsonDTISOEncoder
 from socorro.lib.util import DotDict
 
 
-def dates_to_strings_for_json(obj):
-    if isinstance(obj, datetime.datetime):
-        return obj.isoformat()
-    return json.JSONEncoder.default(self, obj)
```
willkg: Over the course of my tenure on Socorro, I've removed like 5 of these now. I centralized a bunch of them into `socorro/lib/datetimeutil.py`. However, this one is a little different since it uses `.isoformat()` for the format.

I don't know if that's meaningful or how much it differs from the other `JsonDTEncoder`. I figure that's a project for another day.

socorro/external/fs/crashstorage.py
```diff
@@ -765,12 +755,6 @@ class TarFileSequentialReadingCrashStore(CrashStorageBase):
         from_string_converter=class_converter
     )
 
-    @staticmethod
-    def stringify_datetimes(obj):
-        if isinstance(obj, datetime.date):
-            return obj.iso_format()
-        return json.JSONEncoder.default(self, obj)
-
```
willkg: This wasn't used anywhere, so I nixed it.

socorro/external/fs/filesystem.py
```diff
@@ -119,13 +121,14 @@ def defaultAcceptanceFunction(dummy):
     return True
 
 
+# FIXME(willkg): This is only used in the submtter. We should delete it or rewrite with os.walk.
```
willkg: I'm pretty sure the submitter isn't set up to run in the local development environment. Otherwise I'd do this now.

There's another bug where we want to adjust the submitter so that it submits to multiple environments. I suspect it'll be easier to just rewrite the submitter using what I built for Antenna. If that happens, we can delete this rather than rewrite it.

socorro/external/fs/filesystem.py
```diff
@@ -119,13 +121,14 @@ def defaultAcceptanceFunction(dummy):
...
-  rootDirectory,
-  acceptanceFunction=defaultAcceptanceFunction,
-  directoryAcceptanceFunction=defaultAcceptanceFunction,
-  directorySortFunction=cmp,
-  maxDepth=100000,
-  **kwargs
+    rootDirectory,
+    acceptanceFunction=defaultAcceptanceFunction,
+    directoryAcceptanceFunction=defaultAcceptanceFunction,
+    directorySortFunction=cmp,
```
willkg: These argument names are super long.

socorro/external/fs/filesystem.py
```diff
@@ -136,18 +139,18 @@ def findFileGenerator(
     """
     if 0 >= maxDepth:
         return
+    keep_dir = directoryAcceptanceFunction
```
willkg: I added this to reduce the line length later. It's less work to do it inside the function rather than change the argument and adjust uses of this function elsewhere in the codebase.

socorro/external/postgresql/staticdata.py
```diff
@@ -1,3 +1,5 @@
+# flake8: noqa
```
willkg: This file has a bunch of long lines. It's just data. We could move the data to CSV files, but seemed like a better trade-off to just noqa the whole file.

socorro/external/statsd/statsd_rule_benchmark.py
```diff
@@ -89,13 +88,8 @@ class CountStackWalkerTimeoutKills(CountAnythingRuleBase):
...
     def _predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
-        # override me to check any condition within a raw, processed crash
-        # or even the state of the processor itself from the proc_meta
-        return reduce(
-            lambda x, y: x or "SIGKILL" in y,
-            proc_meta.processor_notes,
-            False
-        )
+        # Only run the action if "SIGKILL" is in the processor notes
+        return any(['SIGKILL' in note for note in proc_meta.processor_notes])
```
willkg: Python 3 doesn't have `reduce`. I figured I'd rewrite this because the way it's written doesn't lend itself to sussing out what it's trying to do. I also ditched the original comment since it seems generic and unhelpful.

socorro/external/statsd/statsd_rule_benchmark.py
```diff
@@ -2,50 +2,49 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from socorro.external.statsd.statsd_base import (
-    StatsdBenchmarkingWrapper as StatsdRuleBenchmarkWrapper,
-)
```
willkg: This was here solely for backwards compatibility of names. I nixed it and updated the rest of the code base accordingly.

I also verified we're not using this name in our configuration.

willkg:
Thank you!

PR #4031, opened 2017-10-10
fixes bug 1403248 - add reprocess.py script

willkg:
This adds a reprocess script that works with the other scripts like
`fetch_crashids.py` and sends specified crash ids to a server environment for
reprocessing. This lets us reprocess individual crashes as well as groups of
crashes that match criteria specified with a Super Search.


socorro/scripts/reprocess.py
```diff
@@ -0,0 +1,85 @@
...
+    url = args.host.rstrip('/') + '/api/Reprocessing/'
+
+    # NOTE(willkg): This will pause until all crash ids are available which won't work for some
+    # situations.
+    crash_ids = list(args.crashid or sys.stdin)
+    crash_ids = [item.strip() for item in crash_ids]
+    print('Processing %s crashes...' % len(crash_ids))
+
+    groups = list(chunkify(crash_ids, CHUNK_SIZE))
+    for i, group in enumerate(groups):
```
willkg: It'd be nice if this had an estimated completion. For reprocessing 10s of things, it's not a big deal. For reprocessing 1000s of things, it'd help a whole lot.

But I think we should defer it until later. Simple is better for now.

willkg:
This is easy to test, but you need an API token that has "reprocess" permission. Once you have that, then you can do:

```
# Get into the processor container

# Show command help
(container) $ python scripts/reprocess.py

# Process a crash id in -stage (you need a valid crash id)
(container) $ python scripts/reprocess.py --host=https://crash-stats.allizom.org CRASHID

# Process a crash id via stdin
(container) $ echo CRASHID | python scripts/reprocess.py --host=https://crash-stats.allizom.org
```

socorro/scripts/reprocess.py
```diff
@@ -0,0 +1,85 @@
...
+        formatter_class=WrappedTextHelpFormatter,
+        prog=os.path.basename(__file__),
+        description=DESCRIPTION.strip(),
+    )
+    parser.add_argument('--host', help='host for system to reprocess in', default=DEFAULT_HOST)
+    parser.add_argument('crashid', nargs='*', help='one or more crash ids to fetch data for')
+
+    args = parser.parse_args(argv)
+
+    # FIXME(willkg): Rethink this?
```
Osmose: Is leaving this in intentional?

socorro/scripts/reprocess.py
```diff
@@ -0,0 +1,85 @@
...
+            data={'crash_ids': group},
+            headers={
+                'Auth-Token': api_token
+            }
+        )
+        if resp.status_code != 200:
+            print('Got back non-200 status code: %s %s' % (resp.status_code, resp.content))
+            continue
+
+        time.sleep(SLEEP)
```
Osmose: Why the sleep?

socorro/scripts/reprocess.py
```diff
@@ -0,0 +1,85 @@
...
+    # NOTE(willkg): This will pause until all crash ids are available which won't work for some
+    # situations.
+    crash_ids = list(args.crashid or sys.stdin)
+    crash_ids = [item.strip() for item in crash_ids]
+    print('Processing %s crashes...' % len(crash_ids))
+
+    groups = list(chunkify(crash_ids, CHUNK_SIZE))
+    for i, group in enumerate(groups):
+        print('Processing group... (%s/%s)' % (i + 1, len(groups)))
+        resp = session_with_retries(url).post(
```
Osmose: Shouldn't we be able to reuse the session for all of our requests?

socorro/scripts/reprocess.py
```diff
@@ -0,0 +1,85 @@
...
+
+    url = args.host.rstrip('/') + '/api/Reprocessing/'
+
+    # NOTE(willkg): This will pause until all crash ids are available which won't work for some
+    # situations.
+    crash_ids = list(args.crashid or sys.stdin)
+    crash_ids = [item.strip() for item in crash_ids]
+    print('Processing %s crashes...' % len(crash_ids))
+
+    groups = list(chunkify(crash_ids, CHUNK_SIZE))
```
Osmose: `chunkify` fails here because it hasn't been imported.

socorro/scripts/reprocess.py
willkg: I left this in intentionally. I do this a lot. There are a bunch of things I want to think about, but I've been pushing to get things done and deferring thinking about them especially with things where we don't need to figure it out now. In some cases, it helps to wait to see how other things nearby gel.

The issue here is that now we have two API tokens with different permissions. Is it better to have a local development environment and require the API token to have *all* the permissions, or better to have a bunch of different environment variables for the various tokens we need.

Having said all that, this particular line isn't wildly helpful for anyone who isn't me. I'll remove it.

socorro/lib/util.py
```diff
@@ -4,12 +4,38 @@
...
 
+from itertools import islice
 import logging
 import sys
 import threading
 import traceback
 
 
+def chunkify(iterable, n):
+    """Return chunks of n length of iterable
```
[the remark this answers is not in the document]
willkg: Sure!

socorro/scripts/reprocess.py
willkg: Rrr... I think I moved it at the last minute after I had used it.

socorro/scripts/reprocess.py
willkg: Yes, I think so. I lifted this from Peter's code and wasn't really thinking about it.

socorro/scripts/reprocess.py
willkg: We have a bunch of rate-limiting in the webapp. The sleep alleviates the likelihood that we hit that. The retryable session code alleviates it further.

I'll add a comment because that's not obvious here.

willkg:
You wrote:

> When I run this locally with no arguments, instead of showing the command description, it stalls on the crash_ids = list(args.crashid or sys.stdin) line.
> I'd like to test it on staging or locally; how would I tell if a crash was successfully reprocessed?
> There's no tests for the script itself, do we normally not write tests for scripts like this?

Mmm... I almost always write `--help` when I want help, so I didn't think about this.

We want to have the full list of crash ids so we can print how far along you are. Crash ids come from either args OR stdin. If there aren't any args, the script assumes they're coming from stdin and will block on consuming that until it gets an EOF. It won't get an EOF if there's nothing populating stdin. I bet there's a way to see if something is going to populate stdin without blocking. I'd have to look into it.

The processor is set up to look at three different queues: `socorro.normal`, `socorro.priority`, and `socorro.reprocessing` (or something like that). If you run the processor in your local development environment and point this script at your local development environment and use an API token from your local development environment, then you can send crash ids to the reprocessing API and it'll add items to the queue and the processor will process them. You should be able to see all that happening.

I didn't write tests for this script. We've had bits of this conversation elsewhere. We should have more tests, but we don't have the scaffolding for that, yet. I wrote up bug 1409547 to write scaffolding to enable testing scripts. After that exists, then we should add tests for scripts and cover at least the "does it run and print help?" case.

willkg:
Oh, actually, it's much easier to just point this at stage and reprocess there. You'll know a crash is reprocessed because it'll say something in the processor notes about it.

For example, if you look at this crash:

https://crash-stats.allizom.org/report/index/9b5f24eb-810e-431e-a2e3-ae72a0171016

the processor notes (it's the bottom of that table) say this:

> processor_stage-processor-i-0ae608c7187677e9d_1420; MozillaProcessorAlgorithm2015; skunk_classifier: reject - not a plugin hang

After reprocessing, it'll say something like "This was originally processed on such-and-such a date." and have two sets of notes.

willkg:
^^^ That rebases against master and then fixes a bunch of things based on your comments.

socorro/scripts/reprocess.py
```diff
@@ -0,0 +1,103 @@
...
+
+    if args.crashid:
+        crash_ids = args.crashid
+    elif not sys.stdin.isatty():
+        # If a script is piping to this script, then isatty() returns False. If there is no script
+        # piping to this script, then isatty() returns True and if we do list(sys.stdin), it'll
+        # block waiting for intput.
+        crash_ids = list(sys.stdin)
+    else:
+        crash_ids = []
```
willkg: I'm glad you poked me into fixing this. It works much nicer.

Osmose:
Code looks good, just waiting on stage to finish reprocessing a crash to confirm it's working before I approve.

willkg:
Thank you! Best Review Ever!

PR #4032, opened 2017-10-11
fixes bug 1389463 - create architecture diagram using dot

willkg:
This redoes the architecture diagram using graphviz and dot notation. It's not
the bestest thing in the world, but it's editable without additional tools and
it seems good for now.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -28,7 +28,9 @@ import sphinx_rtd_theme
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = []
+extensions = [
+    'sphinx.ext.graphviz'
+]
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
--- a/docs/gettingstarted.rst
+++ b/docs/gettingstarted.rst
@@ -259,11 +259,14 @@ General architecture
       subgraph stores {
          rank=same;
 
-         aws [shape=tab, label="S3", style=filled, fillcolor=gray];
          postgres [shape=tab, label="Postgres", style=filled, fillcolor=gray];
          elasticsearch [shape=tab, label="Elasticsearch", style=filled, fillcolor=gray];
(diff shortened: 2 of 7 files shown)


willkg:
The end result isn't the bestest thing in the world in terms of layout, but it's more correct than what's there currently:

http://socorro.readthedocs.io/en/latest/gettingstarted.html#general-architecture

We can improve on this.

Best way to build docs and see the graph is to do this:

1. run `make dockerdocs`
2. point your browser at `docs/_build/html/gettingstarted.html`

It's missing some things like the submitter and that box in scl3. It has crontabber, but I didn't put any arrows on that because I know it uses postgres, but don't know about other things.

Thoughts on improvements?

docs/conf.py
```diff
@@ -28,7 +28,9 @@
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = []
+extensions = [
+    'sphinx.ext.graphviz'
```
Osmose: I get the following error when trying to build:

```
WARNING: dot command 'dot' cannot be run (needed for graphviz output), check the graphviz_dot setting
```

Looks like graphviz isn't installed in the docker container? Or the build isn't happening in a container?

docs/gettingstarted.rst
```diff
@@ -242,13 +242,62 @@ environment, put it in ``my.env``.
...
+      client [shape=box3d, label="firefox"];
+
+      subgraph antpig {
+         rank=same;
+
+         antenna [shape=rect, label="antenna"];
+         pigeon [shape=cds, label="pigeon"];
+      }
+
+      subgraph stores {
```
Osmose: Grouping all the stores together makes the lines to a from the processor a little difficult to read, could we split this into multiple groups?

willkg:
^^^ That redoes the infrastructure for building docs so that people other than @willkg can do it.

It also adjusts grouping of services such that the arrows are less incomprehensible. I think that works better. I also added edges for crontabber.

willkg:
Thank you!

PR #4033, opened 2017-10-11
bug 1405679 - Remove unused SubmitterFileSystemWalkerSource

willkg:
This removes a class we don't use (and probably haven't in a long time). I was going to tinker with things and figured I'd remove this now.

If Travs/Circle are happy, this is fine. It's not used in configuration at all.


willkg:
Yeah... Socorro does runtime dependency injection, so it's not enough to grep the codebase to determine whether things are in use or not--you also need to look at the configuration for the -prod and -stage environments. So "not used in config" really means "I checked -prod and -stage configuration and didn't see this used".

willkg:
Thank you!

PR #4034, opened 2017-10-12
bug 1405679 - add submitter as a docker service

willkg:
This adds the submitter as a docker service with configuration and a run script.
This is different than other services because the configuration conflicts with
configuration for other services.

For now, I'm just slapping this together for the purposes of being able to test
changes to the submitter in the local development environment. If we decide to
use this for other purposes, we can redo it then.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -82,6 +82,25 @@ services:
     volumes:
       - .:/app
 
+  # Submitter app
+  #
+  # NOTE(willkg): This app runs only in -prod and lets us slurp crashes from
+  # -prod to -stage. This service is here so we can test changes to the
+  # submitter, but you should probably never need to run this in your local
+  # development environment. It does configuration differently than all the
+  # other services because otherwise we end up with conflicts with things
+  # we care about.
+  submitter:
+    image: local/socorro_processor
+    env_file:
+      - docker/config/local_dev_submitter.env
+      - my.env
+    depends_on:
+      - rabbitmq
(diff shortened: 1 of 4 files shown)


willkg:
To test this:

```
$ docker-compose up submitter
```

It'll print a bunch of stuff and then say "Sleep 1 minute...". If that works, then this is good to go.

willkg:
You're absolutely right about the testing thing.

The submitter is a "middle" piece in that it expects certain things on either side of it to be working and set up. That's not trivial and we don't have scripts for that. This would be a good candidate for a system test we could run in the local development environment. Maybe something like:

```
$ docker-compose up submitter
```

Then in another terminal:

```
$ docker-compose up antenna
```

Then in a third terminal:

```
$ ./docker/as_me.sh bash
(processor) $ ./scripts/systemtest_submitter.sh
Fetch crash id...
Crash id is: xxxxx
Fetch crash data...
Sync into S3...
Put crash id in socorro.submitter queue...

Please verify the submitter picked up crash: xxxxx
Please verify antenna collected crash: xxxxx
```

Then you could watch the submitter terminal, verify it picked up the crash from the queue and submitted it. Then you watch the antenna terminal and verify that it got a crash. Then verify all the crashids are the same.

That's pretty manual still. I'm not sure how to make it better. Maybe we make the S3 buckets different? That might be more involved, though.

willkg:
Tossed the last comment into https://bugzilla.mozilla.org/show_bug.cgi?id=1409223 .

willkg:
Thank you!

PR #4035, opened 2017-10-12
Add example of using shell scripts for processing crashes

willkg:
I didn't actually test this, but pretty sure this is correct.

Can you test this out? Is it clear what's going on and how the pieces work with each other?

### DIFF
--- a/docs/components/processor.rst
+++ b/docs/components/processor.rst
@@ -179,6 +179,41 @@ For example::
    Processing will fail unless the crash data is in the S3 container first!
 
 
+Example using all the scripts
+-----------------------------
+
+Let's process crashes for Firefox from yesterday. We'd do this::
+
+  # Start bash in the processor container as me
+  $ docker/as_me.sh bash
+
+  # Generate a file of crashids--one per line
+  (container) $ scripts/fetch_crashids.py > crashids.txt
+
+  # Pull raw crash data from -prod for each crash id and put it in the
+  # "crashdata" directory on the host
+  (container) $ cat crashids.txt | scripts/fetch_crash_data.py ./crashdata
+
+  # Create a dev_bucket in localstack-s3
+  (container) $ scripts/socorro_aws_s3.sh mb s3://dev_bucket/
+
+  # Copy that data from the host into the localstack-s3 container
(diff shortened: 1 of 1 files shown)


Osmose:
I ran the steps locally and got this output when I finally ran the processor: https://gist.github.com/Osmose/1aa91bf76ee9e6fda0e8f8958ffa352c

I'm not entirely sure how to parse that. After running it, my webapp now hangs indefinitely when I try to load a page.

Osmose:
I ran this locally again on different Wifi, and I think it worked. I see some entries in my top crashers list, and have some individual crash data I can look at. However, the front page of my webapp shows an error message still:

screen shot 2017-10-13 at 12 12 30

Is this something to worry about? If not, then I'll approve this PR.

willkg:
Yeah... So I talked with @peterbe about this a bit and I forget the specifics, but that warning shows up if there's not enough data or some kind of data is missing or something like that.

It'd be good to figure out "what do I need to do to get the front page to work". Maybe we should codify that in this example. It might be something like "1000 crashes" or "100+ crashes for the last few days" or something like that.

We could experiment with that as part of this PR, but I'm inclined to land this as is and write up a bug for figuring that out. It'd definitely be helpful to know when doing webapp work.

peterbe:
For what it's worth, I don't think it's a matter of if you need 100 or 1,000 datapoints. I think you basically need 1.

But isn't that error because there's no data in the weekly indexes? I.e. the socorroYYYYWW doesn't exist.

willkg:
@peterbe Indexes get created automatically when crashes are processed, so if we're pulling 100 firefox crashes from yesterday, than the week should be created.

Is there a crontabber job that needs to run?

peterbe:
@willkg Sorry for sounding lazy but I'm not sure what condition causes that red alert to appear. Best to track it down in the code with some print statements. 

PS. If it's not a missing index, it could also be that ES isn't successful on every shard.

willkg:
@Osmose What you're seeing is not unexpected. I created bug 1408536 to figure out the product home page. We can do that work there. I think everything else here is good to go.

willkg:
Thank you!

PR #4042, opened 2017-10-16
Fix bug 1388593: Use NPM to manage frontend dependencies.

Osmose:
This is split into two commits to make reviewing a little easier; the first commit contains the interesting changes, while the second commit simply deletes a bunch of unused vendored dependencies.

This specifically avoids updating the dependencies; we can do that in a later PR.

I don't entirely know the best way to test this. I ran the site locally in development and prod mode and clicked around a bunch of pages, and things seemed to work. I dunno if there's a better way to test that I didn't break the frontend somehow.

This probably has implications for the non-docker deployment process, mainly:

- Installing NPM 5 or higher for `package-lock.json` support.
- Installing the dependencies somewhere and configuring `NPM_ROOT_PATH` to point to where they're installed.

I don't know how to update that. If anyone can point me to how to update that I'd appreciate it.


docker/Dockerfile.webapp
```diff
@@ -1,15 +1,20 @@
 FROM local/socorro_base
 
-# Install nodejs and npm
-RUN DEBIAN_FRONTEND=noninteractive apt-get install -y npm nodejs nodejs-legacy
+# Install nodejs and npm from Nodesource's 8.x branch
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y curl
+RUN DEBIAN_FRONTEND=noninteractive curl -sL https://deb.nodesource.com/setup_8.x | bash -
```
Osmose: I couldn't find a better way to get a recent node installed on Debian. Should we maybe copy this script into the repo instead of downloading and executing it?

docker/Dockerfile.webapp
```diff
@@ -1,15 +1,20 @@
...
+COPY ./webapp-django/package-lock.json /webapp-frontend-deps/package-lock.json
+RUN cd /webapp-frontend-deps/ && npm install
 
 COPY . /app/
 
-RUN cd /app/webapp-django/ && npm install -g
+ENV LESS_BINARY /webapp-frontend-deps/node_modules/.bin/lessc
+ENV UGLIFYJS_BINARY /webapp-frontend-deps/node_modules/.bin/uglifyjs
+ENV CSSMIN_BINARY /webapp-frontend-deps/node_modules/.bin/cssmin
+ENV NPM_ROOT_PATH /webapp-frontend-deps/
```
Osmose: I moved these mostly for better cache usage, since we don't really need to redownload the frontend or app directory files if they change.

webapp-django/package.json
```diff
@@ -8,10 +8,28 @@
...
+    "Select2": "github:select2/select2#3.5.4",
+    "ace-builds": "1.2.8",
+    "d3": "3.5.17",
+    "d3-sankey": "0.7.1",
+    "filesize": "2.0.3",
+    "flatpickr": "4.0.3",
+    "font-awesome": "4.6.1",
+    "jquery": "2.1.0",
+    "jquery-jsonview": "1.2.3",
+    "jquery-ui": "github:jquery/jquery-ui#1.11.1",
```
Osmose: jQuery didn't start uploading to npm until 1.12, I believe. We'll update this in a future PR.

webapp-django/package.json
```diff
@@ -8,10 +8,28 @@
...
   "license": "MPL-2.0",
-  "dependencies": {
+  "devDependencies": {
     "cssmin": "0.4.3",
     "graceful-fs": "4.1.4",
     "less": "2.7.1",
     "uglify-js": "2.6.1"
+  },
+  "dependencies": {
+    "Select2": "github:select2/select2#3.5.4",
```
Osmose: Same as jQuery, this can be switched to pulling from NPM once we update.

docker/Dockerfile.webapp
willkg: We should download it and put it in the repo so we know what we're executing.

Normandy is doing this:

https://github.com/mozilla/normandy/blob/master/recipe-server/Dockerfile#L9

That project is using python3.6-slim and we're using python2.7.13-slim (we should upgrade to python2.7.14-slim). So theoretically it should be in the same boat debian-wise. Can we reuse that?

willkg:
This has a conflict on `requirements.txt`. We should change how we're installing node. Otherwise, this looks ok eyeball-wise.

After those two things are fixed, I'll run through the site.

The next thing we need to figure out is how this works with our current infrastructure. I think installing npm and node is baked into the AMI. Running collectstatic happens in this script:

https://github.com/mozilla-services/socorro/blob/master/webapp-django/bin/bootstrap.sh

and I think that gets called when building the socorro RPM.

@milescrabill will know more about this.

Osmose:
Rebased, and I replaced the curl|bash step with what Normandy does.

willkg:
I pulled down the changes, ran `make dockerbuild`, verified that it's using npm to fetch dependencies, and made sure everything worked in local development environment.

So this looks good to me as far as I can tell.

Have you talked with @milescrabill to work out how this is going to affect deployments?

Osmose:
> Have you talked with @milescrabill to work out how this is going to affect deployments?

Yep. The bug is assigned to him now while he worked on updating deployments. Until then we should wait on merging this PR.

willkg:
I tried to do all the things that I could so you can land this when you're ready.

PR #4044, opened 2017-10-17
Fix bug 1400952: Update Breakpad to a more recent version.

Osmose:
Tested locally by rebuilding my processor container with the new breakpad version and successfully processing the crash from [bug 1400952](https://bugzilla.mozilla.org/show_bug.cgi?id=1400952). I was able to get proper symbols for the crashing thread, and a proper signature.

### DIFF
--- a/minidump-stackwalk/get-minidump-instructions.cc
+++ b/minidump-stackwalk/get-minidump-instructions.cc
@@ -415,6 +415,7 @@ int main(int argc, char** argv)
       if (p && (p - line == 8 || p - line == 16) && !strstr(line, "<.data>:")) {
         frame.instruction = strtoull(line, nullptr, 16);
         symbolizer.FillSourceLineInfo(module_list,
+                                      NULL, // unloaded_modules
                                       &system_info,
                                       &frame);
         print_frame(frame, last_frame);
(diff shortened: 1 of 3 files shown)


Osmose:
Automation is failing because docker is pulling the old Breakpad revision from TaskCluster. I'll update it and re-run automation after this passes code review, since building on TaskCluster will also alter the version actually being shipped.

Osmose:
@luser r? 

Also, it's unclear to me if we want to merge and then trigger the TC build, or the other way around. And if one or the other requires retriggering other jobs?

luser:
> Also, it's unclear to me if we want to merge and then trigger the TC build, or the other way around. And if one or the other requires retriggering other jobs?

It's up to you, but if you trigger the TC build first, then you can retrigger CI on this PR and verify that it's green and then merge, and then CI on master should be green as well.

We should definitely file a followup on adding the Breakpad changeset to the Taskcluster index so that we can pin the revision of the snapshot in use to make this better in the future.

Osmose:
The TaskCluster job for building this pulls from the master branch, so we had to land this first anyway.

The job I triggered is up at https://tools.taskcluster.net/groups/MfdazrdLSEeEvuNr2Bb0gg/tasks/MfdazrdLSEeEvuNr2Bb0gg/details.

PR #4046, opened 2017-10-18
fixes bug 1409745 - update to python 2.7.14-slim

willkg:
This is pretty straight-forward. To test, I did this:

```
$ make dockerbuild
$ make dockertest
```

Circle CI will cover that, so if it's fine, then we're fine.


willkg:
Thank you!

PR #4047, opened 2017-10-18
fixes bug 1408054 - fixes bash prompt in containers

willkg:
When running a command in a container, this will now show the username and
container name in teh bash prompt. For example:

```
app@processor
```

When using docker/as_me.sh which runs the container using a uid and gid of a
user on the host, this will now show "you" as the username. For example:

```
you@processor
```

This will make it much easier to use containers for development.


willkg:
To test this, I'd do the following:

```
$ docker-compose run webapp bash
app@webapp:/app$ exit

$ ./docker/as_me.sh bash
you@processor:/app$ exit

app@processor:/app$ exit
```

as well as any other variations you can think of.

docker/set_up_ubuntu.sh
```diff
@@ -19,3 +19,20 @@ apt-get install -y gcc apt-transport-https build-essential python-dev \
...
+
+# Stomp on the bash prompt with something more useful for development.
+cat > /etc/bash.bashrc <<EOF
+# Get the name for the current uid and drop error messages.
+MYUSER="\$(id -u -n 2> /dev/null)"
+
+# If id -u -n 2 didn't return a 0 exit code, then it's because there is
+# no name for that uid in the container. So set it to "you".
+if [[ "\$?" != "0" ]]; then
+    MYUSER="you"
```
willkg: Technically we don't know if you're running with the uid/gid of a user on the host and we can't tell what the account name for that user is from inside the container. So I'm using "you" here, but maybe we should use a different indicator?

The other possible options are "root" and "app"--both of those are users that exist inside the container.

Osmose: "you" seems fine to me.

willkg:
There are three wins here:

1. It's less likely I'll forget which terminal I'm doing what in because it tells me who I'm logged in as and which container it is.
2. We can fix up the documentation that involves command lines and it'll be a lot clearer.
3. Any time we do command lines, it'll look clearer and less weird. For example, when I copy and paste signature generation output.

I'm excited to land this.

willkg:
Thank you!

PR #4049, opened 2017-10-18
bug 1405679 - submitter urls

willkg:
This adds support for posting to multiple urls. It's a little tricky to test--we want to test both the single value and the multiple value situations.

For single value do this:

1. in terminal 1, run `docker-compose run submitter bash` and then `./docker/run_submitter.sh`
2. in terminal 2, run `docker-compose up antenna`
3. in terminal 3, run `./docker/as_me.sh bash` and then the following:
3.1. `./scripts/fetch_crashids.txt --num=1 > tosubmit.txt`
3.2. `cat tosubmit.txt | ./scripts/fetch_crash_data.py ./crashdata`
3.3. `./scripts/socorro_aws.sh sync ./crashdata s3://dev_bucket/`
3.4. `cat tosubmit.txt | ./scripts/add_crashid_to_queue.py socorro.stagesubmitter`

That downloads a crash, puts it in s3, and tosses the crashid in the `socorro.stagesubmitter` queue.

The submitter will wake up every minute. It'll wake up, see the crash id, and submit it to the url. You should see this in terminal 2.

In terminal 3, you should see Antenna accept the crash and save it.

For the multi-value, in terminal 1, do this:

```
env destination.urls=http://antenna:8000/submit,http://antenna:8000/submit ./docker/run_submitter.sh
```

That'll cause it to submit to Antenna twice. That's goofy, but it'll work fine here. You should see it submitting twice in the submitter terminal and getting received and saved twice in the Antenna terminal.

### DIFF
--- a/docker/config/local_dev_submitter.env
+++ b/docker/config/local_dev_submitter.env
@@ -9,7 +9,8 @@
 # ----------------------------------------------------------------------------------
 
 destination.crashstorage_class=socorro.submitter.breakpad_submitter_utilities.BreakpadPOSTDestination
-destination.url=http://antenna:8888/
+destination.urls=http://antenna:8000/submit
+
 
 source.crashstorage_class=socorro.external.boto.crashstorage.BotoS3CrashStorage
 source.temporary_file_system_storage_path=/tmp
--- a/docker/config/local_dev_submitter.env
+++ b/docker/config/local_dev_submitter.env
@@ -11,9 +11,9 @@
 destination.crashstorage_class=socorro.submitter.breakpad_submitter_utilities.BreakpadPOSTDestination
 destination.urls=http://antenna:8000/submit
 
-
 source.crashstorage_class=socorro.external.boto.crashstorage.BotoS3CrashStorage
 source.temporary_file_system_storage_path=/tmp
+resource.boto.keybuilder_class=socorro.external.boto.connection_context.DatePrefixKeyBuilder
 
 new_crash_source.crashstorage_class=socorro.external.rabbitmq.crashstorage.RabbitMQCrashStorage
 new_crash_source.new_crash_source_class=socorro.external.rabbitmq.rmq_new_crash_source.RMQNewCrashSource
@@ -34,3 +34,14 @@ resource.rabbitmq.host=rabbitmq
 resource.rabbitmq.virtual_host=rabbitvhost
 secrets.rabbitmq.rabbitmq_user=rabbituser
 secrets.rabbitmq.rabbitmq_password=rabbitpwd
+resource.boto.host=localstack-s3
+resource.boto.port=5000
+resource.boto.access_key=foo
+secrets.boto.secret_access_key=foo
+resource.boto.bucket_name=dev_bucket
+resource.boto.temporary_file_system_storage_path=/tmp
+resource.boto.calling_format=boto.s3.connection.OrdinaryCallingFormat
+resource.boto.resource_class=socorro.external.boto.connection_context.HostPortS3ConnectionContext
+
+# Never on a server!
+resource.boto.secure=False
--- a/socorro/scripts/add_crashid_to_queue.py
+++ b/socorro/scripts/add_crashid_to_queue.py
@@ -29,7 +29,7 @@ Queues:
 
 * socorro.normal - normal processing
 * socorro.priority - priority processing
-* socorro.submitter (-prod only) - sends crash to -stage environment
+* socorro.stagesubmitter (-prod only) - sends crash to -stage environment
 
 """
 
--- a/socorro/submitter/breakpad_submitter_utilities.py
+++ b/socorro/submitter/breakpad_submitter_utilities.py
@@ -1,68 +1,58 @@
-#! /usr/bin/env python
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/
-from configman import Namespace
-from socorro.external.crashstorage_base import CrashStorageBase
 
 import os
 import urllib2
+
+from configman import Namespace
 import poster
+
+from socorro.external.crashstorage_base import CrashStorageBase
+
+
 poster.streaminghttp.register_openers()
 
 
 class BreakpadPOSTDestination(CrashStorageBase):
-    """This a crashstorage derivative that pushes a crash out to a
-    Socorro collector waiting at a url.
+    """This a crashstorage derivative that pushes a crash out to a Socorro collector waiting at a url.
 
     """
     required_config = Namespace()
     required_config.add_option(
-        'url',
-        short_form='u',
-        doc="The url of the Socorro collector to submit to",
-        default="http://127.0.0.1:8882/submit"
+        'urls',
+        doc='One or more urls to submit to separated by commas',
+        default="http://127.0.0.1:8888/submit"
     )
-    required_config.add_option(
-        'echo_response',
-        short_form='e',
-        doc="echo the submission response to stdout",
-        default=False
-    )
-
-    def __init__(self, config, quit_check_callback=None):
-        super(BreakpadPOSTDestination, self).__init__(
-            config,
-            quit_check_callback
-        )
-        self.hang_id_cache = dict()
 
     def save_raw_crash_with_file_dumps(self, raw_crash, dumps, crash_id):
+        """Saves a raw crash by wrapping it up in a multipart/form-data payload and submitting it as an HTTP
+        POST to a collector
+
+        """
+        urls = self.config.urls.split(',')
+
         try:
+            # Create raw crash with file pointers to minidumps for poster to pull from
             for dump_name, dump_pathname in dumps.iteritems():
                 if not dump_name:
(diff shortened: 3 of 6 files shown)


docker/config/local_dev_submitter.env
```diff
@@ -9,10 +9,11 @@
 # ----------------------------------------------------------------------------------
 
 destination.crashstorage_class=socorro.submitter.breakpad_submitter_utilities.BreakpadPOSTDestination
-destination.url=http://antenna:8888/
```
willkg: This value was wrong. So this change both changes the configuration key and also sets it to the right value. It's port 8888 if you're connecting from the host, but port 8000 if one container is connecting to another.

docker/config/local_dev_submitter.env
```diff
@@ -9,10 +9,11 @@
...
+destination.urls=http://antenna:8000/submit
 
 source.crashstorage_class=socorro.external.boto.crashstorage.BotoS3CrashStorage
 source.temporary_file_system_storage_path=/tmp
```
willkg: We can remove this line. I'll do that.

socorro/scripts/add_crashid_to_queue.py
```diff
@@ -29,7 +29,7 @@
 
 * socorro.normal - normal processing
 * socorro.priority - priority processing
-* socorro.submitter (-prod only) - sends crash to -stage environment
+* socorro.stagesubmitter (-prod only) - sends crash to -stage environment
```
willkg: This fixes the documentation for this script--I had the queue name wrong.

socorro/submitter/breakpad_submitter_utilities.py
```diff
@@ -1,68 +1,58 @@
...
+
         try:
+            # Create raw crash with file pointers to minidumps for poster to pull from
             for dump_name, dump_pathname in dumps.iteritems():
                 if not dump_name:
                     dump_name = self.config.source.dump_field
                 raw_crash[dump_name] = open(dump_pathname, 'rb')
+
+            # Get uuid for logging purposes
+            uuid = raw_crash.get('uuid', 'NO UUID')
```
willkg: I have no idea if a crash could ever get here and not have a uuid. I decided to add some defensive-fu here because the world is crazy.

socorro/submitter/breakpad_submitter_utilities.py
```diff
@@ -1,68 +1,58 @@
...
+        'urls',
+        doc='One or more urls to submit to separated by commas',
+        default="http://127.0.0.1:8888/submit"
     )
-    required_config.add_option(
-        'echo_response',
-        short_form='e',
-        doc="echo the submission response to stdout",
-        default=False
-    )
```
willkg: I dropped this configuration item because the submitter already logs the response to stdout.

socorro/submitter/breakpad_submitter_utilities.py
```diff
@@ -1,68 +1,58 @@
...
-
-    def __init__(self, config, quit_check_callback=None):
-        super(BreakpadPOSTDestination, self).__init__(
-            config,
-            quit_check_callback
-        )
-        self.hang_id_cache = dict()
```
willkg: The sole purpose of overriding `__init__` is to initialize `self.hang_id_cache`, but that never gets used, so I nixed this whole thing.

socorro/submitter/breakpad_submitter_utilities.py
```diff
@@ -1,68 +1,58 @@
...
-                    'submitted %s (original crash_id)',
-                    raw_crash['uuid']
+                    'submitted %s to %s; response %s', uuid, url, submission_response
                 )
-            except KeyError:
-                pass
-            self.config.logger.debug(
-                'submission response: %s',
-                submission_response
-            )
```
willkg: I collapsed all this logging into a single line. Since we're using `uuid` which handles the "no uuid" case, we don't need to catch `KeyError`.

socorro/submitter/breakpad_submitter_utilities.py
```diff
@@ -1,68 +1,58 @@
...
-                datagen,
-                headers
-            )
-            submission_response = urllib2.urlopen(request).read().strip()
-            try:
+
+            # Submit payload to all urls
+            for url in urls:
+                request = urllib2.Request(url, datagen, headers)
+                submission_response = urllib2.urlopen(request).read().strip()
```
Osmose: Should we maybe try-catch here in case submission fails? That way if it does, we can continue submitting to the other URLs if it's a URL-specific thing.

willkg: I don't know... This is a transactiony kind of mess. We have that situation whether we add exception handling or not. We can't enforce "submit it to all urls or none of the urls". So if there are two urls, is it better to have it submit to one, but not another?

I think we can be hand-wavey since we don't need to make sure *all* possible things going to *all* -stage environments make it there. We just need to make sure that some things get to each -stage environment so that we know the environment is ok.

Given that, I don't think I want to add exception handling in this case.

willkg:
Regarding automated tests, what do you think is missing?

Osmose:
> Regarding automated tests, what do you think is missing?

If creating an instance of `BreakpadPOSTDestination` isn't difficult (which I do not know if it isn't), than I think the `save_raw_crash_with_file_dumps` method seems like it would be pretty easy to test with pytest. At a minimum, verifying that POST requests are sent to the given URLs with the right data would be a good test of the main functionality. Tests of the parameters as well would be nice to have.

willkg:
Oh, my bad. I knew submitter had tests, but I didn't notice that there were no tests for `BreakpadPOSTDestination`.

I'll write at least a basic one up and attach it to this PR.

willkg:
^^^ That rebases against master, tweaks the `BreakpadPOSTDestination` code to make it easier to test, and then adds a couple of basic minimal tests. It's not comprehensive, but it feels sufficient for now.

@Osmose Does this look ok?

socorro/submitter/breakpad_submitter_utilities.py
```diff
@@ -45,12 +58,20 @@ def save_raw_crash_with_file_dumps(self, raw_crash, dumps, crash_id):
             # Build the payload
             datagen, headers = poster.encode.multipart_encode(raw_crash)
 
+            # requests.post expects a string or buffer, so we run out the datagen
+            # part generator
+            payload = ''.join(datagen)
```
Osmose: I think we can actually remove the `poster` dependency here by using the `files` parameter in requests:

```python
# This variable name is kinda meh
raw_crash_post_files = {key: (None, value) for key, value in raw_crash.iteritems()}
for url in urls:
    resp = requests.post(url, files=raw_crash_post_files)
```

`files` takes a dictionary of data to encode in multipart format. If the values are strings, it assumes the strings are file contents and the key is the name of the file; passing a tuple indicates an explicit filename different from the key, and passing `None` as the filename supresses the output of a filename. I tested this in my terminal:

```python
Python 2.7.11 (default, Feb 25 2016, 12:11:05)
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)] on darwin
>>> import poster
>>> result = poster.encode.multipart_encode({'foo': 'bar', 'baz': 'biff'})
>>> ''.join(result[0])
'--7ef02afc4e6f4d22815feeb33c455eb5\r\nContent-Disposition: form-data; name="foo"\r\nContent-Type: text/plain; charset=utf-8\r\n\r\nbar\r\n--7ef02afc4e6f4d22815feeb33c455eb5\r\nContent-Disposition: form-data; name="baz"\r\nContent-Type: text/plain; charset=utf-8\r\n\r\nbiff\r\n--7ef02afc4e6f4d22815feeb33c455eb5--\r\n'
>>> import requests
>>> r = requests.post('http://example.com', files={'foo': (None, 'bar'), 'baz': (None, 'biff')})
>>> r.request.body
'--4da50a73f09f47029f71b0ad7ef3bba0\r\nContent-Disposition: form-data; name="foo"\r\n\r\nbar\r\n--4da50a73f09f47029f71b0ad7ef3bba0\r\nContent-Disposition: form-data; name="baz"\r\n\r\nbiff\r\n--4da50a73f09f47029f71b0ad7ef3bba0--\r\n'
```

By default it seems to be missing the `Content-Type` headers for each field that `poster` has. If that's important, you can pass a third parameter in the tuple:

```python
>>> r = requests.post('http://example.com', files={'foo': (None, 'bar', 'text/plain; charset=utf8')})
>>> r.request.body
'--af3641ebfd874dd2b4b78ec1d8feb413\r\nContent-Disposition: form-data; name="foo"\r\nContent-Type: text/plain; charset=utf8\r\n\r\nbar\r\n--af3641ebfd874dd2b4b78ec1d8feb413--\r\n'
```

willkg: I'm somewhat loathe to remove poster as part of this PR because while I trust you, I have vague memories of having to write a mini_poster in Antenna after hitting issues with poster (didn't support Python 3) and requests `files` argument (I forget what issues I hit--it's entirely possible I just did it wrong).

I think we should push that bit off. I wrote up bug 1411399 (https://bugzilla.mozilla.org/show_bug.cgi?id=1411399) to cover it.

willkg:
Thank you!

PR #4050, opened 2017-10-19
bug 1409648 - GC rule sets in source code

lonnen:
This is a partial garbage collect of extraneous rule sets as proposed in bug 1409648. I wanted to break up the change for easier review and early feedback.

Each commit is a discrete generation of the 'garbage collect' Some of the removed methods and classpaths will still show up in a grep because they are referenced in alembic migrations from a time when we stored rulesets in the database. We no longer interact with that table but they remain in the historic files. I'm not sure what, if anything, we should do there.

The first few commits are all totally dead code. The final commit removes skunk and support classifiers entirely. These are currently applied, but they apply to a very small fraction of crashes and GA has no record of anyone querying for them in supersearch in the last year.

This patch leaves skunk and support fields in supersearch fields and the crash report schema we send to telemetry. I'm not sure when, but cleaning those up should happen after this ships.


socorro/unittest/processor/test_processor_2015.py
```diff
@@ -63,64 +39,30 @@ def test_rule_sets_from_string_1(self):
         assert rc.ruleset01.rule_system_class.default == expected
         assert 'apply_all_rules' == rc.ruleset01.action.default
         expected = (
-            'socorro.processor.support_classifiers.BitguardClassifier, '
-            'socorro.processor.support_classifiers.OutOfDateClassifier'
+            'socorro.processor.general_transform_rules.CPUInfoRule, '
+            'socorro.processor.general_transform_rules.OSInfoRule '
```
lonnen: These classifiers are pulled in for the test but not applied. Near as I can tell they are used to verify that the config loads the classes.

socorro/unittest/processor/test_processor_2015.py
```diff
@@ -63,64 +39,30 @@ def test_rule_sets_from_string_1(self):
...
         )
         assert rc.ruleset01.rules_list.default == expected
 
-    def test_rule_sets_from_string_2(self):
```
lonnen: I didn't replace this. It seemed pretty low value, and the follow on patches are likely to move back from rule sets to a rule list, which will necessitate redoing this set of processor_2015 stuff shortly

socorro/unittest/lib/test_transform_rules.py
```diff
@@ -355,28 +355,6 @@ def increment_1(s, d):
...
-        assert not (
-            transform_rules.is_not_null_predicate(
-                {'alpha': ''}, None, None, None, 'alpha'
-            )
-        )
-        assert not (
-            transform_rules.is_not_null_predicate(
-                {'alpha': None}, None, None, None, 'alpha'
-            )
-        )
```
willkg: Funny how only this one had tests.

socorro/unittest/processor/test_processor_2015.py
willkg: I concur.

socorro/unittest/processor/test_processor_2015.py
willkg: 👍

lonnen:
thanks!

PR #4051, opened 2017-10-20
fixes bug 1408448 - stomp on signature when parentBuildID != childBuildID

willkg:
If `MozCrashReason` is `"MOZ_RELEASE_ASSERT(parentBuildID == childBuildID)"`, then
stomp on the signature with `"parentBuildID != childBuildID"`. This allows all
these crashes to end up in the same bucket where the original signature is
likely to be bogus since there are two different builds involved.

This also adds `raw_crash['MozCrashReason']` to the list of things that we use to
generate signatures.

I tested this out in this bug comment with the specific crashes and Andrew approved the output.

I think to test, we should let Travis/Circle do their thing. You can also rerun what I did in this comment and verify the output is the same:

https://bugzilla.mozilla.org/show_bug.cgi?id=1408448#c6

### DIFF
--- a/socorro/signature/rules.py
+++ b/socorro/signature/rules.py
@@ -710,3 +710,43 @@ class SignatureIPCMessageName(Rule):
             raw_crash['IPCMessageName']
         )
         return True
+
+
+class SignatureIPCMessageName(Rule):
+    """augments the signature if there is a IPC message name in the crash"""
+
+    def predicate(self, raw_crash, processed_crash):
+        return bool(raw_crash.get('IPCMessageName'))
+
+    def action(self, raw_crash, processed_crash, notes):
+        processed_crash['signature'] = '{} | IPC_Message_Name={}'.format(
+            processed_crash['signature'],
+            raw_crash['IPCMessageName']
+        )
+        return True
+
+
+class SignatureParentIDNotEqualsChildID(Rule):
+    """Stomp on the signature if MozCrashReason is parentBuildID != childBuildID
+
+    In the case where the assertion fails, then the parent buildid and the child buildid are
+    different. This causes a lot of strangeness particularly in symbolification, so the signatures
+    end up as junk. Instead, we want to bucket all these together so we replace the signature.
+
+    """
+
+    def predicate(self, raw_crash, processed_crash):
+        value = 'MOZ_RELEASE_ASSERT(parentBuildID == childBuildID)'
+        return bool(raw_crash.get('MozCrashReason', '') == value)
+
+    def action(self, raw_crash, processed_crash, notes):
+        notes.append(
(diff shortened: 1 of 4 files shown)


socorro/signature/rules.py
```diff
@@ -710,3 +710,43 @@ def action(self, raw_crash, processed_crash, notes):
             raw_crash['IPCMessageName']
         )
         return True
+
+
+class SignatureIPCMessageName(Rule):
```
Osmose: This is a copy-paste of the rule above, maybe you hit paste twice :P

socorro/signature/rules.py
```diff
@@ -710,3 +710,43 @@ def action(self, raw_crash, processed_crash, notes):
...
+        return bool(raw_crash.get('MozCrashReason', '') == value)
+
+    def action(self, raw_crash, processed_crash, notes):
+        notes.append(
+            'Signature replaced with MozCrashAssert, was: "%s"'
+            % processed_crash['signature']
+        )
+
+        # This is a failed assertion and the crash reason is the assertion that failed, so we put
+        # "!=" in the signature because that's what caused the assertion failure.
```
Osmose: I think it's clear what the line does without the comment. We can leave it in if you like anyway.

socorro/signature/rules.py
```diff
@@ -710,3 +710,43 @@ def action(self, raw_crash, processed_crash, notes):
...
+
+    In the case where the assertion fails, then the parent buildid and the child buildid are
+    different. This causes a lot of strangeness particularly in symbolification, so the signatures
+    end up as junk. Instead, we want to bucket all these together so we replace the signature.
+
+    """
+
+    def predicate(self, raw_crash, processed_crash):
+        value = 'MOZ_RELEASE_ASSERT(parentBuildID == childBuildID)'
+        return bool(raw_crash.get('MozCrashReason', '') == value)
```
Osmose: Is this bool cast necessary?

socorro/signature/rules.py
willkg: I implemented this wrong the first time. I thought it looked sufficiently curious and wrong that it was prudent to add a verbose comment.

socorro/signature/rules.py
willkg: You're right--it's already a bool. Oops.

socorro/signature/rules.py
willkg: I'm having a fucking hell of a time with my keyboard right now. It keeps double-doing things. But I should have caught this before committing.

I'll finish up the work for bug 1377254 (flake8) which would have caught this in CI and local testing.

willkg:
^^^ That rebases against master and then fixes the issues you brought up. This is ready for review again.

willkg:
Thank you!

PR #4052, opened 2017-10-20
fixes bug 1409032 - add FNODOBFM to irrelevant list

willkg:
At present, there are two forms:
```
* ?? ?? ::FNODOBFM::`string''
* ?? ::FNODOBFM::`string''
```
We want to ignore both of them and anything that has `FNODOBFM` in the frame
string. This is internal compiler stuff and isn't good for bucketing or pointing
at causes.

I tested it with example crashes. Output is here:

https://bugzilla.mozilla.org/show_bug.cgi?id=1409032#c5


socorro/siglists/irrelevant_signature_re.txt
```diff
@@ -17,6 +17,7 @@ dalvik-jit-code-cache
 dalvik-LinearAlloc
 dalvik-mark-stack
 data@app@org\.mozilla\.f.*-\d\.apk@classes\.dex@0x.*
+.*FNODOBFM.*
```
Osmose: I believe the `.*` at the end is unnecessary since `re.match` matches from the start of the line. But other lines in this file use it, maybe we should just be consistent. This is a nit anyway.

willkg: Oh, funky--I didn't notice that. I'll nix the `.*` at the end.

willkg:
Thank you!

PR #4056, opened 2017-10-22
fixes bug 1410580 - improve "mach_msg_trap" signatures

willkg:
See analysis in https://bugzilla.mozilla.org/show_bug.cgi?id=1410580#c6 .

Most of the things I added don't show up much in the last 6 months. However, it's entirely possible they would show up more if we made a few of the changes in this PR. Given that, it's hard to know whether looking at history helps much.

There's a `.csv` attached to bug 1410580 that shows how these changes affect `mach_msg_trap` signatures over the last few days.


willkg:
@luser Can you sanity check this? I don't know exactly what some of these things are--I'm doing a "well, it looks like a sphere, so I'll put it with the other spheres" kind of thing. But for all I know, one of those spheres could be a Death Star and shouldn't be in the irrelevant list at all...

socorro/siglists/signature_sentinels.txt
```diff
@@ -5,5 +5,6 @@
 # Note: no regular expressions here.
 
 _purecall
+_sigtramp
```
willkg: This will cause signature generation to start the signature at this frame if it exists. In cases where this shows up, everything above it is crash reporter frames.

This is a good example of that: https://crash-stats.mozilla.com/report/index/2e177d22-8236-40dc-88f7-2e5b10171018

socorro/siglists/irrelevant_signature_re.txt
```diff
@@ -20,7 +21,12 @@ data@app@org\.mozilla\.f.*-\d\.apk@classes\.dex@0x.*
 .*FNODOBFM
 framework\.odex@0x.*
 gfxPlatform::GetPlatform
-google_breakpad::ExceptionHandler::HandleInvalidParameter.*
+glFlush_ExecThread
+glDrawArrays.*
+google_breakpad::CrashGenerationClient::RequestDumpForException
+google_breakpad::ExceptionHandler::SignalHandler
+google_breakpad::ExceptionHandler::WriteMinidumpWithException
+google_breakpad::ReceivePort::WaitForMessage
```
willkg: These `google_breakpad*` frames are all the crash reporter. We can skip these with reckless abandon.

socorro/siglists/irrelevant_signature_re.txt
```diff
@@ -8,6 +8,7 @@
 __aeabi_fcmpgt.*
 ashmem
 app_process@0x.*
+AppleIntelHD3000GraphicsGLDriver@.*
```
willkg: These frames are super long and were causing some of the crashes I was looking at to get truncated, so I added them to this list to get ignored.

willkg:
@Osmose There's a lot of analysis and thoughts in bug 1410580. It's worth verifying the example crashes have changed signatures. That verifies I didn't forget to commit something.

Osmose:
Unassigning myself since the discussion for this is still ongiong in the bug, re-add me when it's ready. :D

willkg:
@Osmose This is ready for review now. I think all the parties are in agreement. If not, we can tweak things later.

socorro/siglists/irrelevant_signature_re.txt
```diff
@@ -21,7 +23,12 @@ data@app@org\.mozilla\.f.*-\d\.apk@classes\.dex@0x.*
 framework\.odex@0x.*
 GetLCIDFromLangListNodeWithLICCheck
 gfxPlatform::GetPlatform
-google_breakpad::ExceptionHandler::HandleInvalidParameter.*
+glFlush_ExecThread
+glDrawArrays.*
```
Osmose: My reading of @luser's [last comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1410580#c19) is that we want both the `gl` lines above in signatures, wouldn't including them in the irrelevant list remove them?

socorro/siglists/signature_sentinels.txt
Osmose: To confirm my understanding, signature generation will start here but will not include this frame in the final signature, since this is also in the irrelevant list?

willkg: Yes.

socorro/siglists/irrelevant_signature_re.txt
willkg: Yeah. So, some of the comments in that bug suggest this (and the other `gl*` ones) should be in. And some suggest they should be out. And honesty, I feel like we're playing the hokey-pokey at this point.

I'm not qualified to state with confidence what should be in/out. However, seems better to hedge towards not including things in the irrelevant list that we're not confident should definitely be there.

I pinged @luser to sanity-check this one more time. I'm loathe to do more iterations on change, then comment in the bug, then get a consensus since this spohl indicated this was critical and should get fixed asap. I'm thinking we'll want to reprocess the last week's worth of crashes, too.

That's about where I'm at.

luser:
I don't have a very strong opinion about the remaining issue. If you've already got them in the patch then leaving them is fine.

willkg:
I nixed the `gl*` lines. We can see how this performs and adjust it going forward.

Thank you all!

PR #4059, opened 2017-10-24
fixes bug 1392587 - remove BreakpadStackwalkerRule

willkg:
We don't use this anymore--we use BreakpadStackwalkerRule2015. This removes the
rule and its tests.


socorro/processor/breakpad_transform_rules.py
```diff
@@ -596,7 +374,7 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
 
                 if self.config.chatty:
                     self.config.logger.debug(
-                        "BreakpadStackwalkerRule: %s, %s",
+                        "BreakpadStackwalkerRule2015: %s, %s",
```
willkg: Seems like this was confusing before.

socorro/unittest/processor/test_breakpad_transform_rules.py
```diff
@@ -210,20 +203,14 @@ def get_basic_config(self):
...
-            '/mnt/socorro/symbols/symbols_sbrd,'
-            '/mnt/socorro/symbols/symbols_os'
+        config.command_line = (
+            BreakpadStackwalkerRule2015.required_config .command_line.default
         )
+        config.command_pathname = '/bin/stackwalker'
+        config.public_symbols_url = 'https://localhost'
+        config.private_symbols_url = 'https://localhost'
         config.symbol_cache_path = '/mnt/socorro/symbols'
+        config.temporary_file_system_storage_path = '/tmp'
```
willkg: I pulled this configuration from the `BreakpadStackwalkerRule2015` test.

socorro/unittest/processor/test_breakpad_transform_rules.py
```diff
@@ -310,41 +297,34 @@ def test_stackwalker_fails_2(self, mocked_subprocess_module):
...
         expected = [
-            "MDSW output failed in json: Expected String or Unicode",
-            "MDSW failed on 'upload_file_minidump': unknown error"
+            '/bin/stackwalker output failed in json: Expected String or Unicode',
+            (
+                'MDSW failed on \'timeout -s KILL 30 /bin/stackwalker --raw-json '
+                '00000000-0000-0000-0000-000002140504.json --symbols-url https://localhost '
+                '--symbols-url https://localhost --symbols-cache /mnt/socorro/symbols '
+                'a_fake_dump.dump 2>/dev/null\': unknown error'
+            )
```
willkg: `BreakpadStackwalkerRule2015` produces slightly different output. So this just updates it to the new output.

socorro/unittest/processor/test_breakpad_transform_rules.py
```diff
@@ -310,41 +297,34 @@ def test_stackwalker_fails_2(self, mocked_subprocess_module):
...
 
-        rule = BreakpadStackwalkerRule(config)
-        with rule._temp_file_context('foo.TEMPORARY.txt'):
-            pass
-        mocked_unlink.assert_called_once_with('foo.TEMPORARY.txt')
-        mocked_unlink.reset_mock()
-
-        with rule._temp_file_context('foo.txt'):
+        rule = BreakpadStackwalkerRule2015(config)
+        with rule._temp_raw_crash_json_file('foo.json', example_uuid):
```
willkg: `BreakpadStackwalkerRule` has a `_temp_file_context` context manager which works differently than the `BreakpadStackwalkerRule2015` in a couple of respects:

1. it doesn't check if there's `TEMPORARY` in the filename--it does the same thing for everything
2. it uses a different filename which has the thread name and the crashid in it to reduce thread contention

willkg:
Thank you!

PR #4060, opened 2017-10-26
fixes bug 1409745 - update python image in test container

willkg:
I missed this when I updated the image in the `Dockerfile`.

This affects Circle CI and `make dockertest` and that's it. If Circle is happy, we're fine.


willkg:
Thank you!

PR #4062, opened 2017-10-26
Fix bug 1407449: Use a constraints file for our Python requirements.

Osmose:
This splits the requirements.txt file into two:

- requirements/default.txt, which contains direct dependencies that are used by
  Socorro.
- requirements/constraints.txt, which contains dependencies of the direct
  dependencies, and the hashes that we accept for them.

bootstrap.sh was updated to use this new format for the existing infrastructure,
and Dockerfile was updated for the new infrastructure.

The following dependencies were removed because they were either not used, or
were dependencies of another dependency that was not used:

- argparse
- BeautifulSoup
- certifi
- cffi
- cryptograph
- futures
- google-api-python-client
- gunicorn
- honcho
  - A Procfile existed in the repo root to be used by honcho, but it is
    out-of-date (it refers to the old middleware) and was also removed.
- idna
- ipaddress
- mercurial
- ndg
- Paste
- path
- pgxnclient
  - This was only used by scripts/json-enhancements.sh, which is not used
    anymore and was also removed, along with its entry in Makefile.
- pycparser
- pyOpenSSL
- sqlalchemy
- toposort
- uritemplate
- wsgiref

The following dependencies were moved to requirements/constraints.txt as they
are not direct dependencies:

- configobj
- configparser
- cssselect
- decorator
- django-jsonfield
- enum34
- httplib2
- Mako
- MarkupSafe
- mccabe
- py
- pyasn1
- pyasn1-modules
- pycodestyle
- pyflakes
- python-editor
- rsa
- sasl
- simplejson
- six
- thrift
- urllib3

---

I tested this locally by rebuilding my docker setup and running `make dockertest`. I also pulled down and locally processed a crash from prod and compared the report page to ensure they were the same. 

While I was testing this, I ran into issues running out of disk space in my containers. I suspect that was an unrelated issue. I destroyed my entire Docker for Mac container/image storage and remade everything, and then ran into memory issues with elasticsearch and the processes while running with 2 gigs of memory for the containers. Once I bumped this up to 6 gigs, things started working again. I am not sure if the memory issues are related to these dependency changes; we should keep an eye out for memory issues as we test and deploy this change.


willkg:
I :heart: the level of details here. I'm working through this slowly and double-checking everything figuring that's the appropriate thing to do to review this. I'm pretty sure I'll be done Monday.

I'm psyched about the whole reqs/constraints thing.

willkg:
I think working through the requirements changes is easier done in a google doc where we can edit and comment in-place and eventually end up with a complete record of things. So I'm moving that discussion there: https://docs.google.com/document/d/1zGGGVSxsCzzDB9IZPAQvb64zLQkMLLcItE-SvQN2yqI/edit#

The outcome of that might result in a +1 or required changes.

Osmose:
I've updated this PR based on our discussions:

- Re-added `futures` as a constraint for `django-pipeline`.
- Updated `requests` to the latest version, which doesn't vendor its requirements anymore and adds a few constraints.
  - The new `urllib3` version we're including no longer depends on `ndg-httpsclient`, and only requires `pyOpenSSL` if you're on a Python version older than 2.7.9 for SNI support. This allows us to keep those requirements removed.
- Updated the two imports in the webapp to `six` to use `django.utils.six` in their place.

CONTRIBUTING.rst
```diff
@@ -103,12 +103,19 @@ Code reviews should review the changes in the context of the rest of the system.
...
+2. ``requirements/constraints.txt``, containing dependencies required by the
+   dependencies in ``default.txt`` that Socorro does not use directly.
+
+Dependencies in both files must be pinned and hashed. Use
+`hashin <https://pypi.python.org/pypi/hashin>`_.
 
 For example, to add ``foobar`` version 5::
 
-  hashin -r requirements.txt foobar==5
+  hashin -r requirements/default.txt foobar==5
```
willkg: This adds a first-order dependency for `foobar==5`, but what should we do with anything that that requires?

Osmose: I filed a followup PR that clarifies this: https://github.com/mozilla-services/socorro/pull/4065

PR #4066, opened 2017-10-31
Fix bug 1406121: Display all threads when no crashing thread is available.

Osmose:
That test is more or less a copy-paste of the test above it with minor changes. Ideally I'd also like to test something about the "hidden" class being present but it seemed kind of awkward to do and falls into the "frontend is getting replaced soon anyway" area. But I can totally add it if we think we need it.

I also tested this manually against three crashes from the [query spohl provided](https://bugzilla.mozilla.org/show_bug.cgi?id=1406121#c4).

### DIFF
--- a/webapp-django/crashstats/crashstats/views.py
+++ b/webapp-django/crashstats/crashstats/views.py
@@ -871,15 +871,7 @@ def report_index(request, crash_id, default_context=None):
         context['raw_stackwalker_output'] = 'No dump available'
         parsed_dump = {}
 
-    # If the parsed_dump lacks a `parsed_dump.crash_info.crashing_thread`
-    # we can't loop over the frames :(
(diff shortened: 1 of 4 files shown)


webapp-django/crashstats/crashstats/views.py
```diff
@@ -871,15 +871,7 @@ def report_index(request, crash_id, default_context=None):
         context['raw_stackwalker_output'] = 'No dump available'
         parsed_dump = {}
 
-    # If the parsed_dump lacks a `parsed_dump.crash_info.crashing_thread`
```
Osmose: I traced this back to https://github.com/mozilla-services/socorro/pull/2237. It seems like it was added to avoid an UndefinedError on an older version of the report index, but the current version (with the modifications made in this PR) handle a missing `crashing_thread` value without a problem.

willkg: 👍

PR #4067, opened 2017-10-31
fixes bug 1413210 - Clean up logging configuration

willkg:
This cleans up the logging configuration and removes vestiges of CEF and playdoh
which we don't use anymore.

This also adds py.warnings because we really want to see those so we can fix
them which will reduce future headaches.

### DIFF
--- a/webapp-django/crashstats/settings/base.py
+++ b/webapp-django/crashstats/settings/base.py
@@ -3,7 +3,6 @@
 # variables.
 
 import os
-import logging
 from pkg_resources import resource_string
 
 import dj_database_url
@@ -177,18 +176,33 @@ TEMPLATES = [
 # Always generate a CSRF token for anonymous users.
 ANON_ALWAYS = True
 
-LOG_LEVEL = logging.INFO
+LOGGING_LEVEL = config('LOGGING_LEVEL', 'INFO')
 
-HAS_SYSLOG = True  # XXX needed??
-
-LOGGING_CONFIG = None
-
-# This disables all mail_admins on all django.request errors.
-# We can do this because we use Sentry now instead
 LOGGING = {
+    'version': 1,
+    'disable_existing_loggers': False,
+    'handlers': {
+        'console': {
+            'level': LOGGING_LEVEL,
+            'class': 'logging.StreamHandler',
+        },
+    },
     'loggers': {
+        'django': {
+            'handlers': ['console'],
+            'level': LOGGING_LEVEL,
+        },
+        # This disables all mail_admins on all django.request errors because we
+        # use Sentry for that
         'django.request': {
-            'handlers': []
+            'handlers': ['console'],
+        },
+        'py.warnings': {
+            'handlers': ['console'],
+        },
+        'crashstats': {
+            'handlers': ['console'],
+            'level': LOGGING_LEVEL,
         }
     }
 }
@@ -359,9 +373,6 @@ ADMINS = (
 )
 MANAGERS = ADMINS
 
-# import logging
-# LOGGING = dict(loggers=dict(playdoh={'level': logging.DEBUG}))
-
 # When you don't have permission to upload Symbols you might be confused
 # what to do next. On the page that explains that you don't have permission
 # there's a chance to put a link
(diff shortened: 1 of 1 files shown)


webapp-django/crashstats/settings/base.py
```diff
@@ -177,18 +176,33 @@ def path(*dirs):
 # Always generate a CSRF token for anonymous users.
 ANON_ALWAYS = True
 
-LOG_LEVEL = logging.INFO
+LOGGING_LEVEL = config('LOGGING_LEVEL', 'INFO')
```
willkg: This is nice because then it's easier to change in `my.env` when you're doing things.

webapp-django/crashstats/settings/base.py
```diff
@@ -177,18 +176,33 @@ def path(*dirs):
...
 
-HAS_SYSLOG = True  # XXX needed??
-
-LOGGING_CONFIG = None
```
willkg: I'm 95% sure that having this line caused the later `LOGGING` to get ignored because it's telling Django "Yo--we'll do all our own logging configuration".

https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-LOGGING_CONFIG

If that's correct, then we don't want this line.

webapp-django/crashstats/settings/base.py
```diff
@@ -177,18 +176,33 @@ def path(*dirs):
...
+            'level': LOGGING_LEVEL,
+        },
+        # This disables all mail_admins on all django.request errors because we
+        # use Sentry for that
         'django.request': {
-            'handlers': []
+            'handlers': ['console'],
+        },
+        'py.warnings': {
+            'handlers': ['console'],
```
willkg: I claim we want to see these and not ignore them. It prints things that we should deal with under the auspices of a future bug. For example:

```
/usr/local/lib/python2.7/site-packages/ratelimit/middleware.py:2: RemovedInDjango19Warning: django.utils.importlib will be removed in Django 1.9.
  from django.utils.importlib import import_module

/app/webapp-django/crashstats/crashstats/urls.py:108: RemovedInDjango19Warning: Default value of 'RedirectView.permanent' will change from True to False in Django 1.9. Set an explicit value to silence this warning.
  pattern_name='monitoring:index'
```

Should we keep this section in?

webapp-django/crashstats/settings/base.py
```diff
@@ -177,18 +176,33 @@ def path(*dirs):
...
+        'django': {
+            'handlers': ['console'],
+            'level': LOGGING_LEVEL,
+        },
+        # This disables all mail_admins on all django.request errors because we
+        # use Sentry for that
         'django.request': {
-            'handlers': []
+            'handlers': ['console'],
+        },
```
willkg: This was in there before, but assuming I'm right about `LOGGING_CONFIG`, it was ignored and had no effect.

I've never had to disable the `mail_admins` thing in a Django project before. I'm not sure how to verify whether we need this or not in a local development environment. Do you think we want this or can I nix it?

webapp-django/crashstats/settings/base.py
```diff
@@ -359,9 +373,6 @@ def path(*dirs):
 )
 MANAGERS = ADMINS
 
-# import logging
-# LOGGING = dict(loggers=dict(playdoh={'level': logging.DEBUG}))
-
```
willkg: This was commented out, playdoh-related, and we don't need it anymore. So nixing it.

webapp-django/crashstats/settings/base.py
```diff
@@ -480,15 +491,6 @@ def path(*dirs):
...
 SECRET_KEY = config('SECRET_KEY', 'this must be changed!!')
 
-# Log settings
-
-# Make this unique to your project.
-SYSLOG_TAG = config('SYSLOG_TAG', 'http_app_playdoh')
-
-# Common Event Format logging parameters
-CEF_PRODUCT = config('CEF_PRODUCT', 'Playdoh')
-CEF_VENDOR = config('CEF_VENDOR', 'Mozilla')
```
willkg: We don't need any of this anymore.

willkg:
To test this, I did the following:

1. ran `make dockertest` (which circle and travis can handle nicely)
2. ran `docker-compose run --service-ports webapp bash` and then in that terminal, did hit the home page with each of these run lines:
   1. dev mode: `./docker/run_webapp.sh --dev`
   2. uwsgi: `./docker/run_webapp.sh`
   3. debug=false and uwsgi: `DEBUG=false ./docker/run_webapp.sh`

      This one shows a lot of 404 errors because of static assets.

webapp-django/crashstats/settings/base.py
willkg: Awesome! Good to know. I'll nix the comment.

willkg:
Thank you! This is the beginning of a new era for webapp logging. I'm excited to be a part of it!

PR #4069, opened 2017-11-01
Fix bug 1413164: Downgrade to d3-sankey 0.4.2.

Osmose:
This should be the version of d3-sankey we were using before, instead of the
latest version.

I had trouble testing this locally, because there is [a bug in d3-sankey that triggers when links in the chart all have a value of 0](https://github.com/d3/d3-sankey/pull/39). In our case, we hit this bug when none of the crontabber tasks have any dependencies. When I run crontabber locally, it populates the database with data about some crontabber tasks, but none of them have dependencies: https://gist.github.com/Osmose/3b39699117677ebc2991ea7ec0d02168

It looks like this is normally not an issue since crontabber on the servers and other dev environments has at least one job with a dependency. Fixing that issue requires a more extensive rewrite of the crontabber page that I don't think we should bother with for this bug. Advice on how to either get my crontabber data fixed locally, or on alternative methods for testing, would be appreciated.

### DIFF
--- a/webapp-django/package-lock.json
+++ b/webapp-django/package-lock.json
@@ -84,27 +84,27 @@
       "resolved": "https://registry.npmjs.org/d3-collection/-/d3-collection-1.0.4.tgz",
       "integrity": "sha1-NC39EoN8kJdPM/HMCnha6lcNzcI="
     },
-    "d3-path": {
-      "version": "1.0.5",
-      "resolved": "https://registry.npmjs.org/d3-path/-/d3-path-1.0.5.tgz",
-      "integrity": "sha1-JB6xhJvZ6egCHA0KeZ+KDo5EF2Q="
+    "d3-color": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/d3-color/-/d3-color-1.0.3.tgz",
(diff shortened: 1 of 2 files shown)


willkg:
The lack of `depends_on` information is puzzling. That should come from the jobs themselves, but maybe if the job hasn't run fully, then it doesn't populate the db or something. I wrote up bug 1413386 (https://bugzilla.mozilla.org/show_bug.cgi?id=1413386) to cover that.

Having said that, the page loads fine for me. I did the following:

1. `make dockerbuild` -- this picks up the js lib changes
2. `make dockersetup` and `make dockerupdatedata` to prep the db
3. `docker-compose up crontabber` and wait a bit--this runs all the crontabber jobs
4. `docker-compose up webapp`
5. make my browser go to `http://localhost:8000/crontabber-state/`

I see a box with a lot of jobs listed, but only one dependency between `duplicates` and `reports-clean`. So it looks like one item is showing the dependency.

And sure enough, now that I've written that and then puzzled about it and then re-checked `http://localhost/api/CrontabberState/`, I see that a couple of the jobs do have `depends_on` data.

I think this fix works.

PR #4070, opened 2017-11-01
fixes bug 1413703 - Fix statsd container so it works

willkg:
This fixes the statsd container in the local development environment which didn't work at all. Now it uses graphite which isn't exciting, but does work and lets you see statsd metrics over time.

To test:

1. do `docker-compose up processor`
2. process some crashes
3. go to `http://localhost:8080` (eighty eighty) to view graphite and view the metrics generated


willkg:
Thank you!

PR #4071, opened 2017-11-01
Fix bug 1109069: Include top 10 stack frames in Bugzilla bug summaries.

Osmose:
Here's a sample of the comment that this produces:

```
This bug was filed from the Socorro interface and is
report bp-f0544ebe-211f-4b37-8628-007a40171026.
=============================================================

Top 10 frames of crashing thread:

0 mozglue.dll pages_commit memory/build/mozjemalloc.cpp:1417
1 mozglue.dll arena_t::SplitRun memory/build/mozjemalloc.cpp:2504
2 mozglue.dll arena_t::AllocRun memory/build/mozjemalloc.cpp:2663
3 mozglue.dll je_malloc memory/build/malloc_decls.h:48
4 xul.dll xul.dll@0x7bad14 
5 xul.dll xul.dll@0xb6940b 
6 xul.dll xul.dll@0x3b527a7 
7 xul.dll xul.dll@0x1214aba 
8 xul.dll xul.dll@0xb6916d 
9 xul.dll xul.dll@0x15fc690 

=============================================================

```


webapp-django/crashstats/crashstats/jinja2/crashstats/bugzilla_comment.txt
```diff
@@ -0,0 +1,18 @@
...
+This bug was filed from the Socorro interface and is
+report bp-{{ uuid }}.
+=============================================================
+
+{% if parsed_dump.threads and crashing_thread is not none -%}
+Top 10 frames of crashing thread:
+
+{% for frame in parsed_dump.threads[crashing_thread].frames[:10] -%}
+{{ frame.frame }} {{ frame.module }} {{ frame.signature|remove_signature_arguments }} {{ frame_source(frame) }}
+{% endfor %}
```
willkg: Have you tested this with one of the crashes that doesn't have a crashing thread?

willkg: Scratch that--it looks ok. Not sure why I was confused.

webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -239,6 +244,12 @@ def bugzilla_submit_url(report, bug_product):
     elif op_sys in ('Windows Unknown', 'Windows 2000'):
         op_sys = 'Windows'
 
+    comment = render_to_string('crashstats/bugzilla_comment.txt', {
+        'uuid': report['uuid'],
+        'parsed_dump': parsed_dump,
+        'crashing_thread': crashing_thread,
+    })
```
willkg: I think I would calculate the stack portion in Python-land rather than in Jinja2 land and then pass it in as a rendering param. I'd much rather people be writing PRs to tweak the stack rendering code in python-land than in Jinja2 land.

webapp-django/crashstats/crashstats/jinja2/crashstats/bugzilla_comment.txt
```diff
@@ -0,0 +1,18 @@
...
+{%- macro frame_source(frame) -%}
+  {{ frame.file if frame.file }}{{ ':%s'|format(frame.line) if frame.line }}
+{%- endmacro -%}
+
+This bug was filed from the Socorro interface and is
+report bp-{{ uuid }}.
+=============================================================
+
+{% if parsed_dump.threads and crashing_thread is not none -%}
+Top 10 frames of crashing thread:
```
willkg: If the stack has fewer than 10 frames, would someone write up a bug for Socorro because this number doesn't match the number of frames we generated?

webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -247,15 +258,7 @@ def bugzilla_submit_url(report, bug_product):
...
         'short_desc': 'Crash in {}'.format(smart_str(report['signature'])),
-        'comment': (
-            'This bug was filed from the Socorro interface and is \n'
-            'report bp-{}.\n'
-            '{}'
-            '\n'
-        ).format(
-            report['uuid'],
-            '=' * 61
-        ),
```
willkg: +1 for making this a template.

webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -222,8 +222,13 @@ def read_crash_column(crash, column_key):
     return crash.get(column_key, '')
 
 
+@library.filter
+def remove_signature_arguments(signature):
+    return re.sub(r'\(.*\)', '', signature)
```
willkg: I have vague memories for `.` being greedy, but I don't remember if that's with Python regexes or some other regex system I used over the years. Because of that scar tissue, I would have doing `.*` as `[^\)]+` since we want to match on a bunch of non-right-parens. Your tests pass, so this is fine as is.

Osmose: Wouldn't that fail for nested parentheses? Like `foo:bar(char* baz, int (this ain't valid but whateber))` would end up being `foo:bar)`, since it would consume all the open parenthesis but only one of the closed.

Osmose: Ah, but maybe the current version fails for multiple groups of parentheses? It is a naive way to remove arguments.

willkg: I don't think I've seen nested parens, but you're right--doing `[^\)]+` doesn't work with nested parens. +1 to keeping what you have now.

willkg: I did some testing:

```
Python 2.7.13 (default, Jan 19 2017, 14:48:08) 
[GCC 6.3.0 20170118] on linux2
>>> import re
>>> p = re.compile(r'\(.*\)')
>>> p.sub('abc', '')
''
>>> p.sub('', 'abc')
'abc'
>>> p.sub('', 'abc(123)')
'abc'
>>> p.sub('', 'abc(123(456))')
'abc'
>>> p.sub('', 'abc(123(456), 333)')
'abc'
```

That seems fine to me for now. If we need something fancier, we can write a parser.

Osmose:
Updated with feedback changes. They're significant enough that I think this could use another review pass if you're willing. :D

willkg:
It looks good to me. When Circle/Travis are happy, merge it!

PR #4072, opened 2017-11-02
bug 1413925 - Update jinja_helpers.py

philipp-sumo:
reports from the buggy amd cpus have a different cpu_info when they come from win64 firefox builds.
sample: https://crash-stats.mozilla.com/report/index/fa0c831d-cff4-4337-8dc6-53ac30171102

(also see https://bugzilla.mozilla.org/show_bug.cgi?id=1269894)

### DIFF
--- a/webapp-django/crashstats/base/templatetags/jinja_helpers.py
+++ b/webapp-django/crashstats/base/templatetags/jinja_helpers.py
@@ -107,7 +107,9 @@ def is_dangerous_cpu(cpu_info):
     # for ease of find by users.
     return (
         cpu_info.startswith('AuthenticAMD family 20 model 1') or
-        cpu_info.startswith('AuthenticAMD family 20 model 2')
+        cpu_info.startswith('AuthenticAMD family 20 model 2') or
+        cpu_info.startswith('family 20 model 1') or
+        cpu_info.startswith('family 20 model 2')
     )
 
 


willkg:
@philipp-sumo Can you write up a bug for this? That way we have a bug covering what the problem is and can discuss it there.

philipp-sumo:
@willkg sure, i've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1413925

webapp-django/crashstats/base/templatetags/jinja_helpers.py
```diff
@@ -107,7 +107,9 @@ def is_dangerous_cpu(cpu_info):
     # for ease of find by users.
     return (
         cpu_info.startswith('AuthenticAMD family 20 model 1') or
-        cpu_info.startswith('AuthenticAMD family 20 model 2')
+        cpu_info.startswith('AuthenticAMD family 20 model 2') or
+        cpu_info.startswith('family 20 model 1') or
+        cpu_info.startswith('family 20 model 2')
```
willkg: Since the new strings are pretty vague, I'm kind of loathe to do this. I think we should additionally constrain them by `cpu_arch == 'amd64'`.

willkg:
I think this should have a test, too, but I'm not seeing any existing tests for `is_dangerous_cpu`. I'll have to think about that. We can manually test it for now.

webapp-django/crashstats/base/templatetags/jinja_helpers.py
philipp-sumo: currently there are no apparent collisions at least: https://crash-stats.mozilla.com/search/?cpu_info=~family%2020%20model%201&cpu_info=~family%2020%20model%202&date=%3E%3D2017-05-02T18%3A19%3A18.000Z&_facets=cpu_info#facet-cpu_info

willkg: It's nice that there are no collisions, but things change over time and we can't guarantee there won't be collisions and this is the kind of change where if there were, it'd probably go unnoticed for some time. Meanwhile, we'd be ignoring those crash reports because we incorrectly think they have buggy CPUs.

I think it's prudent to tie this explicitly to the `cpu_arch`.

philipp-sumo:
yes, i'd be glad if you or someone from the socorro team could take over.

willkg:
I added a comment to the bug with the relevant bits. I'm going to close this out. Someone will get to it when we get a chance.

PR #4073, opened 2017-11-02
fixes bug 1410284 - prune migrations

willkg:
This prunes the alembic migrations down to the last few that we've done and
deletes all the other ones. As the tables and code has changed over time, many
of the migrations can't run with the current code base. So keeping them around
isn't helping.

If we ever need to look at migrations for code archaeology, we can look at them
in git history.

I'm pretty sure this should be fine. If you check out the master branch and do:

```
$ docker-compose run processor alembic -c docker/config/alembic.ini history
```

it shows a ton of migrations. If you then look at this branch and do the same, it shows this:

```
app@processor:/app$ alembic -c docker/config/alembic.ini history
a38b03765a79 -> 77856f165be7 (head), bug 1403172 fix fennecandroid api 16
ae2c8ff8c073 -> a38b03765a79, drop plugins_reports tables
07c6633fa1b6 -> ae2c8ff8c073, drop crash_adu_by_build_signature
<base> -> 07c6633fa1b6, delete signature summaries
```

I think that suggests that the pruning is fine and didn't put alembic in some weird state.

The changes here are:

1. remove a lot of files except for the most recent 4
2. change 07c6633fa1b6 `down_revision` to `None` as suggested in the alembic docs

Since we have full control over all our Socorro environments, I think this is actually a no-op and doesn't really change anything except ditches a bunch of files that aren't used anymore.


willkg:
Thank you!

PR #4074, opened 2017-11-03
fixes bug 1413674 - Update datadog python lib

willkg:
We use datadog in a few places, but all of them do:

    from datadog.dogstatsd import statsd

and the mocks all mock that, too.

Updating the datadog library maintains that statsd instance, so we should be
fine.


willkg:
I tested this thoroughly to make sure it didn't somehow change any of our statsd keys and it seems fine. I did this:

1. switch to master branch
   1. run `make dockerbuild` to make sure I have the latest requirements for this branch
   2. delete any statsd container state with `docker-compose stop statsd` and then `docker-compose rm -fsv statsd`
   3. start up the processor and process a few crashes
   4. visit `http://localhost:8080` and see what keys exist in the graphite hierarchy
2. switch to this PR branch and follow the same steps

Verify that graphite shows the same key hierarchy which looked like this for me:

master_branch

That's sort of involved, but that's what I did. Figured it'd be good to know this now for when I switch everything to use Markus later.

willkg:
I'm game for waiting until after the change freeze for this, but I think it's low risk and it's trivial to verify that it's ok on -stage. We can revert it if anything looks funny in the datadog dashboards.

willkg:
I'll watch datadog and make sure things are ok after the next -stage deploy.

willkg:
Thank you!

PR #4075, opened 2017-11-03
Bug 1411036: Add pyup configuration.

Osmose:
I'm following the steps from the checklist on https://github.com/mozilla-services/foxsec/, which says to add this config, then notify secops that we want to enable pyup.

### DIFF
--- a/.pyup
+++ b/.pyup
@@ -1,0 +1,9 @@
+schedule: "every month"
+search: False
+requirements:
+  - requirements/default.txt:
+      update: all
+      pin: true
+  - docs/requirements.txt
+      update: all
+      pin: true


.pyup
```diff
@@ -0,0 +1,9 @@
+schedule: "every month"
+search: False
+requirements:
+  - requirements/default.txt:
```
willkg: If I understand this correctly, the PRs that pyup will do won't be looking at constraints, so it's possible that we'll have to do follow-up commits to get things working. Does that sound right?

Osmose: That's what I intend. I'm not entirely sure whether we should or shouldn't be monitoring the constraints too, given how we're using it. I figure the security monitoring will include the constraints, at least.

PR #4076, opened 2017-11-03
Update to jQuery 3.2.1 and jQuery-UI 1.12.1.

Osmose:
I tested this manually by searching for places where we use the jQuery-UI bundle, which found:

- Crontabber state
- Crashes per day
- Crash report detail page
- Signature report page
- Supersearch

I confirmed on each of these pages that all the UI controls I could find worked properly. I also skimmed the other pages I know about to make sure they work as well.

I attempted to run the e2e tests against my local instance, but they failed because of an assumption about the data my local copy has (something about having a "signature first date" for each signature on a page). Running them against stage would probably be prudent after we land this.

This was easier than I expected! :D


willkg:
That's awesome that we can know what pages use these!

How risky is this PR? Especially given that our webapp ui testing is manual for the most part. Should we move forward with this now or wait until the change freeze is over?

Osmose:
> How risky is this PR? Especially given that our webapp ui testing is manual for the most part. Should we move forward with this now or wait until the change freeze is over?

We should wait on merging this until after the change freeze.

PR #4078, opened 2017-11-06
Pyup initial update

mozsvcpyup:
This is an automated update from pyup.io


requirements/default.txt
```diff
@@ -1,154 +1,205 @@
...
+    --hash=sha256:697ff63bc5451e0b0db48ad205151123d25683b3754198be7ab5fcb44334e519 \
+    --hash=sha256:fc993c9331d91766d54757bbc70231e29d5ceb2d1ac08b1570feaa0c38ab9582 \
+    --hash=sha256:9d64fed2681552ed642e9c0cc831a9e95ab91de72b47d0cb68b5bf506ba88647 \
+    --hash=sha256:5c3213be557d0468f9df8fe2487eaf2990d9799202c5ff5cb8d394d09fad9b2a
+elasticsearch==5.4.0 \
+    --hash=sha256:8f02d626cff98befd298ffb1d19d8e519643d86d3aa27f6d7388566ec4ba48ee \
+    --hash=sha256:e754c688e20fe73160fb6f7f5b63f2a71c78788dc9e6908950681d3a39b56e85
+elasticsearch-dsl==5.3.0 \
+    --hash=sha256:de5fd68901c5b94e51e2ae4e8ad6f618b22665d89794d96f9b6b763986665527 \
+    --hash=sha256:de81b2322cde9bf1171c8a41d80596cdadd433b88b93daa09060af3a512523a7
```
willkg: We talked about this yesterday, but figured I'd leave a comment here so we have it written down.

We can't update elasticsearch or elasticsearch-dsl--those need to get updated at the same time we update our Elasticsearch cluster.

requirements/default.txt
```diff
@@ -1,154 +1,205 @@
...
-python-decouple==2.3 \
-    --hash=sha256:b9fbf747fa8e711d330f2f436b9b2ecf5eed9eb67a637ca81abcd11c2abedec8
-dj-database-url==0.3.0 \
-    --hash=sha256:f2e273ed34acbb560962d5cf12917936d8df02297df09bd3089b8546d4584138 \
-    --hash=sha256:ca01768fdecde134301f3170743226f60edff5c3935f12437378ebd911506353
+Django==1.11.7 \
+    --hash=sha256:75ce405d60f092f6adf904058d023eeea0e6d380f8d9c36134bac73da736023d \
+    --hash=sha256:8918e392530d8fc6965a56af6504229e7924c27265893f3949aa0529cd1d4b99
+eventlog==0.11.0 \
+    --hash=sha256:2a942547f6bb352b3c131b9a0b3972c13f16f764bf144c08e5e610ad52231671
```
willkg: Ahh... this is what you were having issues with. I think eventlog is small enough code-wise and problem-domain-wise that we could vendor it and even roll our own version of it if either of those are easier.

willkg:
I went through all the requirements looking for ones that we can't update unless we update something else like a service. I didn't see any other than the Elasticsearch ones.

One thing that popped in my head is that the local development environment uses Ubuntu, but the server environments are still using Centos 6. It's not the case that if binary dependencies install fine in the local development environment that we're guaranteed they'll install fine in the deploy to -stage and -prod. It's entirely possible we might need to update system libraries, too. That might cover SSL and postgres related libraries. I don't know how we could figure that out short of deploying to -stage and seeing what happens. Maybe this is a good reason to break this PR up into smaller chunks?

vladikoff:
Quick FYI: This thing is a big crazy with queuing up Circle Ci and blocking other builds:

Osmose:
> Quick FYI: This thing is a big crazy with queuing up Circle Ci and blocking other builds:

I knew I forgot something in the cleanup. Sorry for all the spam!

Pyup is _supposed_ to only be creating 1 PR, and only once a month. I'm going to be looking into why it's creating all these PRs tonight/tomorrow.

vladikoff:
No worries :)

Osmose:
@g-k has disabled Pyup for the time being. I'm going to do some testing on my personal account to get the config right, and then we'll re-enable to see if it happens again.

Osmose:
The test suite is now passing for me locally. Next up is actually testing the site manually.

Osmose:
Rebased and added more fixes. I've run through the site, tried out most of the features I know about, processed a few batches of crashes, and am no longer finding any obvious errors.

One important note is that, because we're switching from `eventlog` to `pinax-eventlog`, we need to fake the first migration of the new library in existing environments since it has already run but is using a different app name:

1. Run `docker/as_me.sh bash`.
2. `cd webapp-django`
3. `./manage.py migrate eventlog 0001 --fake`
4. Also, run the new migrations: `./manage.py migrate`

I still need to go over possible documentation changes with this, and I think we'll need to configure pyup to ignore the following libraries, which we're intentionally keeping on old versions:

- Elasticsearch
- Elasticsearch-py
- simplejson

I also have not yet tested if our updated compiled dependencies will work on the environments in staging and prod. I'll be looking at that soon as well.

I think this is ready for a review / more thorough testing, though. Thanks!

Osmose:
Also, I highly recommend reviewing this per-commit past the pyup commits, as they form a narrative structure of the changes that makes way more sense.

willkg:
I suggested we use our Vagrant environment to test compiling since it's got the same stuff as our deploy environment. Then we don't have to "see what happens on -stage" which is exciting, but it's nicer to catch things earlier.

However ... the Vagrant environment isn't working. It needs updating and there are some things that are missing. I'm going to fiddle with it for 30 minutes and if I can get it working enough to test installing requirements, I'll do that. Otherwise we'll have to test compiling on -stage.

willkg:
:muscle:

I fought with my Vagrant environment and then got it working. In that environment, I had it install all the requirements and everything installed fine. While this isn't exactly the same thing as doing a deploy on -stage, it's close enough that I'm not worried.

willkg:
I will review this, but need to get some other things done first including updating the test checklist so I have something helpful to test with (maybe I'll do both at the same time).

Maybe end of this week or early next week.

willkg:
Based on what we talked about in the meeting just now, I'm going to finish up my crontabber PR and then start reviewing this and testing it all out. That makes it possible to push out Monday even if we have to do a few iterations of fixes.

webapp-django/crashstats/crashstats/static/crashstats/css/screen.less
```diff
@@ -830,7 +830,7 @@ div.code {
     float: right;
     .current {
         float: right;
-        background-image: url(../../../img/3rdparty/fatcow/stop16x16.png);
+        background-image: url(../../img/3rdparty/fatcow/stop16x16.png);
```
willkg: I don't understand this change. Was the original url wrong and the fixed url right? What was the error this was kicking up?

socorro/lib/ooid.py
```diff
@@ -62,19 +62,19 @@ def dateAndDepthFromOoid(ooid):
     year = month = day = None
     try:
         day = int(ooid[-2:])
-    except:
+    except ValueError:
```
willkg: For all of these, since we're not doing any checking on the `ooid` argument, it could be an empty string or `None` or something bogus. Prior to this change, those cases would get handled. After this change, they don't.

I'd add `TypeError`. We don't need `IndexError` since it's a substring. Are there other things we should handle?

socorro/lib/ooid.py
```diff
@@ -62,19 +62,19 @@ def dateAndDepthFromOoid(ooid):
...
         return None, None
     try:
         year = 2000 + int(ooid[-6:-4])
         depth = int(ooid[-7])
         if not depth:
             depth = oldHardDepth
         return (dt.datetime(year, month, day, tzinfo=UTC), depth)
-    except:
+    except ValueError:
```
willkg: This one needs to handle `IndexError` because of line 73.

webapp-django/crashstats/base/tests/testbase.py
```diff
@@ -21,7 +21,7 @@
                 classes_with_implementation.append(klass)
 
 
-class TestCase(django.utils.unittest.TestCase):
+class TestCase(unittest.TestCase):
 
     def shortDescription(self):
         return None
```
willkg: We don't use `nose` anymore, so we can drop the `shortDescription`. If you wanted to toss that in this PR, that's cool. Otherwise I'll try to remember to do it later.

docker/config/test.env
```diff
@@ -17,3 +17,6 @@ database_hostname=postgresql
 database_port=5432
 database_username=postgres
 database_password=aPassword
+
+# Django tests are meant to run with debug mode disabled.
+DEBUG=False
```
willkg: Interestingly, we have two sets of test variables. Some defined here and some defined in `setup.cfg` and `webapp-django/pytest.ini`. This file predates switching to pytest. I think we should probably move these vars to the pytest place and nix this file.

I wrote up https://bugzilla.mozilla.org/show_bug.cgi?id=1417913 to think about that.

webapp-django/crashstats/crashstats/urls.py
```diff
@@ -104,7 +104,8 @@
     url(
         r'^status/$',
         RedirectView.as_view(
-            pattern_name='monitoring:index'
+            pattern_name='monitoring:index',
+            permanent=perm_legacy_redirect
         ),
         name='status_redirect',
     ),
```
willkg: I'm not even sure we need this endpoint at all. It's probably for something that changed ages ago. We can probably just drop it altogether.

willkg:
Test checklist doc: https://docs.google.com/document/d/1gN8GFeN01wkcTyYyAVMmLdDTZJquDpYn0Zj4-L7KJIw/edit#

webapp-django/crashstats/crashstats/static/crashstats/css/screen.less
Osmose: The error happened during asset compilation; it couldn't resolve the path to an existing collected static file. It's unclear to me why this worked before and didn't work after, though. I couldn't find any Django or django-pipeline changes that seemed to be related.

I don't actually see any instances of "oopp-hang" in the codebase. Maybe this is just unused and that's why we didn't notice it, and something changed to make the failure no longer silent?

socorro/lib/ooid.py
Osmose: Ah, I didn't know int can throw a TypeError. I can't find a list of possible errors, but those two seem like a prudent pair.

Osmose:
Rebased and updated with review feedback, sans the bits that have bugs filed for handling them later.

PR #4138, opened 2017-11-08
fixes bug 1411991 - symbol upload metrics

willkg:
This implements symbol upload metrics generation.

To test:

1. Run `docker-compose up webapp`
2. Log into the webapp
3. Click on "Symbols" in the bottom right hand corner
4. Upload a symbols file via the web and via the api

You can get symbols files from Tecken because they're awesome. I use this one:

https://github.com/mozilla-services/tecken/blob/master/tests/sample.zip

You can verify metrics were generated by going to `http://localhost:8080` and looking at `statsd.symbols.upload`. Since that's statsd, it won't show tags.

You can also verify metrics were generated by looking at the webapp terminal. You'll see lines like this:

```
webapp_1         | METRICS|2017-11-08 01:50:43|incr|symbols.upload.web_upload|1|#email:wkahngreenemozilla.com
```

We'll have to test the datadog configuration on -stage. It should be very easy to verify yes/no plus verify it's generating tag data we can use in Datadog graphs.

I think this is pretty low risk, but I could be convinced otherwise. The only reason to land it now rather than in a couple of weeks is that we'll be able to accrue symbol upload data and have a better baseline when we migrate.


webapp-django/crashstats/symbols/views.py
```diff
@@ -327,6 +354,7 @@ def upload(request):
         bucket_location
     )
 
+    metrics.incr('api_upload', tags=['email:%s' % _sanitize_email(request.user.email)])
```
willkg: Oops--I forgot to add a test for this. I'll do that tomorrow.

webapp-django/crashstats/symbols/views.py
```diff
@@ -256,6 +282,7 @@ def web_upload(request):
                     symbols_upload.filename
                 )
             )
+            metrics.incr('web_upload', tags=['email:%s' % _sanitize_email(request.user.email)])
```
willkg: Oops--I forgot to add a test for this. I'll do that tomorrow.

webapp-django/crashstats/symbols/views.py
```diff
@@ -29,6 +31,30 @@
...
+
+    :returns: sanitized email value
+
+    """
+    # First convert from unicode to str dropping any non-ascii characters
+    if isinstance(email, unicode):
+        email = email.encode('ascii', 'ignore')
+
+    # Second, drop any bad characters
+    return BAD_CHAR_REGEXP.sub('', email)
```
willkg: This is super gross sanitizing code. We really just need to be able to distinguish email addresses from on another. I thought the datadog python client clean up tag keys and values based on what I read, but the code doesn't suggest that at all so I figured we should do it. The `@` gets used to denote tags in the statsd datagrams.

willkg: This could also use a test. I'll do that tomorrow, too.

webapp-django/crashstats/settings/base.py
```diff
@@ -34,6 +34,8 @@ def path(*dirs):
     cast=bool
 )
 
+# Whether or not we're running in the local development environment
+LOCAL_DEV_ENV = config('LOCAL_DEV_ENV', False, cast=bool)
```
willkg: We don't want to (ab)use `DEBUG` to denote whether we're in a local development environment because we want to be able to flip that on and off to test other things. We really need a separate variable.

I'm game for changing the name. I'm not hung up on this one.

requirements/constraints.txt
```diff
@@ -94,3 +94,10 @@ chardet==3.0.4 \
 futures==3.1.1 \
     --hash=sha256:c4884a65654a7c45435063e14ae85280eb1f111d94e542396717ba9828c4337f \
     --hash=sha256:51ecb45f0add83c806c68e4b06106f90db260585b25ef2abfcda0bd95c0132fd
+statistics==3.4.0b3 \
+    --hash=sha256:535ae04f76a6dae57c41e88de69992ece0cf66eb3d4772c41ab94f064e742ff6 \
+    --hash=sha256:d23485d9666804800b79439776c16f5b18b96edf56182ee8b5f2cb5c7adad0c5
+docutils==0.14 \
+    --hash=sha256:7a4bd47eaf6596e1295ecb11361139febe29b084a87bf005bf899f9a42edc3c6 \
+    --hash=sha256:02aec4bd92ab067f6ff27a38a38a41173bf01bed8f89157768c1573f53e474a6 \
+    --hash=sha256:51e64ef2ebfb29cae1faa133b3710143496eca21c530f3f71424d77687764274
```
willkg: For some bizarre reason, the statistics module depends on docutils.

docker/config/never_on_a_server.env
```diff
@@ -4,6 +4,10 @@
 
 # NOTE: These shouldn't be used on -dev, -stage, or -prod environments.
 
+# Flag indicating we're running things in the local development environment
+LOCAL_DEV_ENV=True
```
willkg: We put this in this file to make sure @milescrabill doesn't accidentally copypasta it to our new server environment. He and I agreed that this was the way to reduce that risk.

docker/config/never_on_a_server.env
```diff
@@ -4,6 +4,10 @@
...
+
+# Django DEBUG mode which shows configuration data on errors
```
willkg: I added this comment because it was weird we didn't have one.

docker/config/local_dev.env
```diff
@@ -88,5 +88,6 @@ RABBITMQ_PASSWORD=rabbitpwd
 RABBITMQ_VIRTUAL_HOST=rabbitvhost
 RAVEN_DSN=
 SECRET_KEY=secretkey
+STATSD_HOST=statsd
```
willkg: `statsd` is the name of the statsd container and thus the hostname we need to use.

willkg:
This needs tests, but otherwise is ready to go. I'll do tests tomorrow.

willkg:
@Osmose This has the tests I wanted to write now. This is ready for review.

webapp-django/crashstats/base/apps.py
```diff
@@ -0,0 +1,42 @@
...
+                {
+                    'class': 'markus.backends.logging.LoggingMetrics',
+                },
+                {
+                    'class': 'markus.backends.statsd.StatsdMetrics',
+                    'options': {
+                        'statsd_host': settings.STATSD_HOST,
+                        'statsd_port': settings.STATSD_PORT,
+                        'statsd_prefix': settings.STATSD_PREFIX,
+                    }
```
willkg: Datadog extends statsd adding tags and some other things. That doesn't work with statsd at all. Markus supports both, though, so we use the statsd backend in the local dev environment and then you can see some things in Graphite in addition to the logs.

webapp-django/crashstats/symbols/views.py
willkg: There's a test for this now.

webapp-django/crashstats/symbols/views.py
willkg: There's a test for this now.

webapp-django/crashstats/symbols/views.py
```diff
@@ -29,6 +31,30 @@
 from . import utils
 
 
+metrics = markus.get_metrics('symbols.upload')
+
+
+BAD_CHAR_REGEXP = re.compile(r'[^0-9a-zA-Z\._-]')
+
+
+def _sanitize_email(email):
```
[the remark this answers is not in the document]
Osmose: Having tag support in markus for backends that support tagging seems prudent to me.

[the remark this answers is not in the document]
willkg: I wrote up this issue for it:

https://github.com/willkg/markus/issues/25

I'll figure it out on Monday.

willkg:
^^^ That rebases against master and switches to using `markus.utils.generate_tag`.

willkg:
@Osmose Does that last commit look ok?

willkg:
Thank you!

PR #4206, opened 2017-11-14
fixes bug 1416905 - redact crontabber error details

willkg:
There are three different urls that yield crontabber state information:

- /api/CrontabberState/
- /monitoring/crontabber/
- /crontabber-state/

All three of them use the CrontabberState model.

This PR adjusts that model to redact some information from the response thus
fixing all three of those endpoints.

### DIFF
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -939,6 +939,23 @@ class CrontabberState(SocorroMiddleware):
     # will never contain PII
     API_WHITELIST = None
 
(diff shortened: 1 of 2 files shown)


willkg:
To test, you have to have at least one failed crontabber app. Then going to the following should all show "See error logging system." for the traceback and value.

* http://localhost:8000/api/CrontabberState/
* http://localhost:8000/monitoring/crontabber/
* http://localhost:8000/crontabber-state/

willkg:
Woo hoo! Thank you!

PR #4207, opened 2017-11-15
fixes bug 1388130 - redo crontabber jobs configuration

willkg:
This redoes crontabber jobs configuration such that it can take a Python dotted path to a string with the configuration.

In order to do that, we also had to redo things so that CronTabberApp was a valid socorro app. This means it now works with the socorro script.

This updates the related docker scripts to use the new way to run it.

In doing all that, I discovered the tests weren't consistent in how they set up a config manager. This fixes tests, too.

This supports the current configuration model as well as the new one (Python dotted paths), so we can make configuration changes after this lands and deploys all the way to -prod.

To test:

1. run `make dockerupdatedata` -- that runs crontabber with a single job
2. run `docker-compose up crontabber` -- that runs the crontabber container and all jobs
3. make sure Travis and Circle are happy

We need to be able to specify different crontabber jobs for our server environments, so this sort of blocks the new docker infrastructure work. Hence it's sort of important to at least get this to stage.

### DIFF
--- a/docker/run_crontabber.sh
+++ b/docker/run_crontabber.sh
@@ -26,7 +26,7 @@ while true
 do
     # FIXME(willkg): We don't want crontabber to exit weird and then have that
     # kill the container, but this is a lousy thing to do.
-    ${CMDPREFIX} python socorro/cron/crontabber_app.py || true
+    ${CMDPREFIX} ./scripts/socorro crontabber || true
     echo "Sleep 5 minutes..."
     sleep 300
 done
--- a/docker/run_update_data.sh
+++ b/docker/run_update_data.sh
@@ -13,17 +13,17 @@ CACHEDIR=/app/.cache/ftpscraper
 
 # Fetch and update release information for these products (comma-delimited)
 PRODUCTS="firefox,mobile"
-CRONTABBER="docker-compose run crontabber python socorro/cron/crontabber_app.py"
+CRONTABBERCMD="./scripts/socorro crontabber"
 
 
 # Fetch release data -- do it as the user so it can cache things
-docker-compose run -u "${HOSTUSER}" crontabber python socorro/cron/crontabber_app.py \
+docker-compose run -u "${HOSTUSER}" crontabber ${CRONTABBERCMD} \
                --job=ftpscraper \
                --crontabber.class-FTPScraperCronApp.cachedir=${CACHEDIR} \
                --crontabber.class-FTPScraperCronApp.products=${PRODUCTS}
 
 # Update featured versions data based on release data
-${CRONTABBER} --job=featured-versions-automatic \
+docker-compose run crontabber ${CRONTABBERCMD} --job=featured-versions-automatic \
               --crontabber.class-FeaturedVersionsAutomaticCronApp.products=${PRODUCTS}
 
 # Fetch normalization data for versions we know about
--- a/socorro/app/for_application_defaults.py
+++ b/socorro/app/for_application_defaults.py
@@ -43,8 +43,7 @@ class ApplicationDefaultsProxy(object):
         return {
             'setupdb': 'socorro.external.postgresql.setupdb_app.SocorroDBApp',
             'submitter': 'socorro.submitter.submitter_app.SubmitterApp',
-            # crontabber not yet supported in this environment
-            #'crontabber': 'socorro.cron.crontabber_app.CronTabberApp',
+            'crontabber': 'socorro.cron.crontabber_app.CronTabberApp',
             'processor': 'socorro.processor.processor_app.ProcessorApp',
             'reprocess_crashlist':
                 'socorro.external.rabbitmq.reprocess_crashlist.ReprocessCrashlistApp',
--- a/socorro/cron/crontabber_app.py
+++ b/socorro/cron/crontabber_app.py
@@ -3,67 +3,109 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from crontabber.app import CronTabberBase, CronTabber
+import sys
 
-# this class is for eventual support of CronTabber with the universal
-# socorro app.
+from configman import Namespace, class_converter
+from crontabber.app import (
+    CronTabberBase,
+    classes_in_namespaces_converter_with_compression,
+    get_extra_as_options,
+    line_splitter,
+    main,
+    pipe_splitter,
+)
+
 from socorro.app.socorro_app import App
 
 
 # NOTE(willkg): This is what we have in -prod
-DEFAULT_JOBS = '''
-socorro.cron.jobs.weekly_reports_partitions.WeeklyReportsPartitionsCronApp|7d
-socorro.cron.jobs.matviews.ProductVersionsCronApp|1d|05:00
-socorro.cron.jobs.matviews.SignaturesCronApp|1d|05:00
-socorro.cron.jobs.matviews.RawUpdateChannelCronApp|1d|05:00
-socorro.cron.jobs.matviews.ADUCronApp|1d|08:30
-socorro.cron.jobs.matviews.DuplicatesCronApp|1h
-socorro.cron.jobs.matviews.ReportsCleanCronApp|1h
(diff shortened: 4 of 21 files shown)


docker/run_crontabber.sh
```diff
@@ -26,7 +26,7 @@ while true
 do
     # FIXME(willkg): We don't want crontabber to exit weird and then have that
     # kill the container, but this is a lousy thing to do.
-    ${CMDPREFIX} python socorro/cron/crontabber_app.py || true
+    ${CMDPREFIX} ./scripts/socorro crontabber || true
```
willkg: crontabber runs as a socorro app so we can use the `socorro` script now.

docker/run_update_data.sh
```diff
@@ -13,17 +13,17 @@ CACHEDIR=/app/.cache/ftpscraper
...
 # Fetch release data -- do it as the user so it can cache things
-docker-compose run -u "${HOSTUSER}" crontabber python socorro/cron/crontabber_app.py \
+docker-compose run -u "${HOSTUSER}" crontabber ${CRONTABBERCMD} \
                --job=ftpscraper \
                --crontabber.class-FTPScraperCronApp.cachedir=${CACHEDIR} \
                --crontabber.class-FTPScraperCronApp.products=${PRODUCTS}
 
 # Update featured versions data based on release data
-${CRONTABBER} --job=featured-versions-automatic \
+docker-compose run crontabber ${CRONTABBERCMD} --job=featured-versions-automatic \
```
willkg: The change to `CRONTABBERCMD` make the run lines in this file more uniform which I kind of like.

socorro/app/for_application_defaults.py
```diff
@@ -43,8 +43,7 @@ def find_all_the_apps():
         return {
             'setupdb': 'socorro.external.postgresql.setupdb_app.SocorroDBApp',
             'submitter': 'socorro.submitter.submitter_app.SubmitterApp',
-            # crontabber not yet supported in this environment
-            #'crontabber': 'socorro.cron.crontabber_app.CronTabberApp',
+            'crontabber': 'socorro.cron.crontabber_app.CronTabberApp',
```
willkg: This appears to be a project that was started a few years ago:

https://github.com/mozilla-services/socorro/commit/3f469e41d600e50f706e62a027e235bd16420eca

but it wasn't finished for crontabber. This PR fixes that for crontabber.

socorro/cron/crontabber_app.py
```diff
@@ -3,67 +3,109 @@
...
+    'socorro.cron.jobs.elasticsearch_cleanup.ElasticsearchCleanupCronApp|7d',
+    'socorro.cron.jobs.drop_old_partitions.DropOldPartitionsCronApp|7d',
+    'socorro.cron.jobs.truncate_partitions.TruncatePartitionsCronApp|7d',
+    'socorro.cron.jobs.clean_raw_adi_logs.CleanRawADILogsCronApp|1d',
+    'socorro.cron.jobs.clean_raw_adi.CleanRawADICronApp|1d',
+    'socorro.cron.jobs.clean_missing_symbols.CleanMissingSymbolsCronApp|1d',
+    'socorro.cron.jobs.missingsymbols.MissingSymbolsCronApp|1d',
+    'socorro.cron.jobs.featured_versions_automatic.FeaturedVersionsAutomaticCronApp|1h',
+    'socorro.cron.jobs.upload_crash_report_json_schema.UploadCrashReportJSONSchemaCronApp|1h',
+]
```
willkg: I just converted this from one big `\n` delimited string into a Python list of strings. That's easier to manipulate so we can do things like "all the default jobs plus this one".

socorro/cron/crontabber_app.py
```diff
@@ -3,67 +3,109 @@
...
+        input_str = class_converter(path_or_jobs)
+    else:
+        input_str = path_or_jobs
+
+    from_string_converter = classes_in_namespaces_converter_with_compression(
+        reference_namespace=Namespace(),
+        list_splitter_fn=line_splitter,
+        class_extractor=pipe_splitter,
+        extra_extractor=get_extra_as_options
+    )
```
willkg: This `from_string_converter` is what the `jobs` configuration variable already used:

https://github.com/mozilla/crontabber/blob/57c463a521cdd2f9db4ae3c1104559abf88d1607/crontabber/app.py#L679

socorro/cron/crontabber_app.py
```diff
@@ -3,67 +3,109 @@
...
+
+# NOTE(willkg): We need to "fix" the defaults here rather than use App's
+# get_application_defaults() because that doesn't support nested configuration
+# defaults
+CronTabberApp.required_config.crontabber.database_class.default = (
+    'socorro.external.postgresql.connection_context.ConnectionContext'
+)
+CronTabberApp.required_config.crontabber.job_state_db_class.default.required_config.database_class.default = (  # noqa
     'socorro.external.postgresql.connection_context.ConnectionContext'
 )
```
willkg: This is so messy, but I couldn't figure out how to fix `get_application_defaults` in a way that didn't affect a ton of other things.

socorro/cron/crontabber_app.py
```diff
@@ -3,67 +3,109 @@
...
-            'socorro.external.postgresql.connection_context.ConnectionContext',
-
-        }
-
-
-# These settings should ideally be done in config, but because, at
-# the moment, it's easier for us to maintain python we're doing it here.
-CronTabber.required_config.crontabber.jobs.default = DEFAULT_JOBS
-CronTabber.required_config.crontabber.database_class.default = (
+    required_config = CronTabberBase.required_config.safe_copy()
```
willkg: Prior to this PR, the `CronTabberApp` was dead code and never used.

socorro/unittest/cron/crontabber_tests_base.py
```diff
@@ -46,7 +48,7 @@ def _setup_config_manager(self, jobs_string, extra_value_source=None):
 
         """
         mock_logging = mock.Mock()
-        required_config = app.CronTabber.get_required_config()
+        required_config = CronTabberApp.get_required_config()
```
willkg: I went through all the tests and made sure they were testing with Socorro's `CronTabberApp`. So there's a bunch of string replacement along those lines.

socorro/unittest/cron/jobs/test_clean_missing_symbols.py
```diff
@@ -1,26 +1,17 @@
...
 from socorro.unittest.cron.jobs.base import IntegrationTestBase
-from socorro.unittest.cron.setup_configman import (
-    get_config_manager_for_crontabber,
-)
 
 
 class TestCleanMissingSymbolsCronApp(IntegrationTestBase):
 
     def _setup_config_manager(self, days_to_keep=None):
-        super(TestCleanMissingSymbolsCronApp, self)._setup_config_manager
```
willkg: I'm puzzled as to what the intention of this line was. It evaluates to a function, but doesn't call the function. This is repeated in a bunch of the tests.

I suspect someone started to do something, left it in the commit, it landed, then it was copy/pasted around without anyone looking at it.

willkg:
Thank you!

socorro/cron/crontabber_app.py
lonnen: `        reference_value_from='resource.postgresql',` ?

willkg: The defaults here are coming from `CronTabber` which is from the crontabber library. We could change the `default` or `reference_value_from` for `database_class` by stomping on it in `CronTabberApp`, but in order to change the default for the `database_class` for `job_state_db_class`, we'd have to subclass `job_state_db_class` and stomp on the option there and change the default or add a `reference_value_from` and then use our subclass. That's doable, but adds a bunch more code.

In the case here, we're only using crontabber one way so (ab)using the classes seems like a better choice.

I'm pretty sure `get_application_defaults` is supposed to fix this issue by letting you adjust defaults to a more application-appropriate value.

https://github.com/mozilla-services/socorro/blob/master/socorro/app/socorro_app.py#L147

But it doesn't seem to work with nested items and I didn't want to add fiddling with that to this PR.

PR #4209, opened 2017-11-17
fixes bug 1410167 - fixes TelemetryS3CrashStorage bucket_name

willkg:
This allows us to set the `bucket_name` value without having to specify it in the
`destination4` configuration key like we do now.

That makes it easier for us to move crashstorage classes around without bumping
into an environment configuration variable which will be defined in a different
place in the new infrastructure.

This will help us fix TelemetryS3 issues with the docker-based infrastructure.

### DIFF
--- a/docker/config/local_dev.env
+++ b/docker/config/local_dev.env
@@ -54,6 +54,7 @@ resource.boto.port=5000
 resource.boto.access_key=foo
 secrets.boto.secret_access_key=foo
 resource.boto.bucket_name=dev_bucket
+resource.boto.telemetry_bucket_name=telemetry_bucket
 resource.boto.temporary_file_system_storage_path=/tmp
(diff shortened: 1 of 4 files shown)


willkg:
To test this out:

1. run the processor:
   
   ```
   docker-compose up processor
   ```

2. the logs should have this line:
   
   ```
   processor_1      | 2017-11-17 14:52:14,548 INFO - processor -  - MainThread - destination.storage4.telemetry_bucket_name: telemetry_bucket
   ```
   
   which indicates it picked up the configuration value from `docker/config/local_dev.env`.
3. process a crash
4. run:
   
   ```
   ./docker/as_me.sh ./scripts/socorro_aws.sh ls s3://
   ```

   That will tell you what buckets exist. You should see a `telemetry_bucket` bucket which means the configuration worked right.

willkg:
Thank you!

PR #4210, opened 2017-11-19
fixes bug 1346996 - adds new product_id_map api endpoint

lonnen:
this exposes a new endpoint, product_id_map, which could called by the processor in lieu of directly access the database

In local testing, I wasn't getting any result from the API call. FakeData.py would populate the table with waterwolf and b2g. I probably need to update _something_ so that this will return a useful value from the test install.


willkg:
We don't use `fakedata.py` and that's one of the things I'm probably going to remove when we ditch the Vagrant infrastructure. We use `staticdata.py` and I don't think it generates any of these things.

I'll take a look at what we have in this table in production and add it to `staticdata.py`. I'll do that tonight or tomorrow morning. That'll make this easier to verify.

Bookkeepingly, it's better for all our bot dohickeys if you change the commit message to be like this:

```
fixes bug 1346996 - adds new product_id_map api endpoint
```

willkg:
I'm super psyched you did this. Thank you!

lonnen:
this intermittent `sh: 1: sw_vers: not found` issue is getting annoying

socorro/external/postgresql/product_id_map.py
```diff
@@ -0,0 +1,29 @@
...
+
+import logging
+
+from socorro.external.postgresql.base import PostgreSQLBase
+
+
+logger = logging.getLogger("webapi")
+
+
+class ProductIDMap(PostgreSQLBase):
```
willkg: Is it likely someone looking for a class for the `product_productid_map` table wouldn't find this one because this is `ProductIDMap` instead of `ProductProductIDMap`? Do other `PostgreSQLBase` class names mirror the tables they derive from?

webapp-django/crashstats/crashstats/models.py
```diff
@@ -1130,6 +1131,16 @@ class ProductBuildTypes(SocorroMiddleware):
     )
 
 
+class ProductIDMap(SocorroMiddleware):
```
willkg: If I were writing this PR, I'd have made this `ProductProductIDMap`. I see how it looks silly, though.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -1130,6 +1131,16 @@ class ProductBuildTypes(SocorroMiddleware):
...
+    cache_seconds = 1
```
willkg: As far as I can tell, we only change that table via migrations which require a deploy. I think we could increase the cache to something higher.

lonnen:
Updated this. Tried to rerun locally and http://localhost:8000/ redirects to http://localhost:8000/home/product/Firefox and which 404s, product not found. Blew up the env, rebuilt it, updated data, and wound history back a little bit... it's still happening. I am no longer certain I busted every environment, but definitely certain my local should be melted to slag. I'll just start over with pen and paper, hand computing that x86.

willkg:
That sounds frustrating. I think the 404 on `/home/product/Firefox` comes from not having a product, but `make dockerupdatedata` should create it. I think either something failed in the copious output produced or you missed a step somehow.

Anyhow, I wiped everything in my local dev environment and this branch works for me, so I can vouch for it working.

lonnen:
Thanks for double checking. I’ve worn all my pencils to stubs trying to
hand-calculate bootstrapping the VM so I’ll probably fix my environment
soonish.

PR #4212, opened 2017-11-21
fixes bug 1418037 - add AuroraVersionFixitRule

willkg:
The AuroraVersionFixitRule checks the crash and stomps on the version if
the crash is from Firefox with specific build ids. The build ids are
for builds that are on the aurora channel for devedition product, but have
a .0b1 beta-style version.

We're fixing the crashes in the processor because the alternative was to
work through the ball of spaghetti that is ftpscraper, the stored procedures,
and related database table data to figure out why it's doing everything WRONG!
The Socorro team therapist suggested we do a processor rule now and rewrite
ftpscraper later and/or switch to buildhub.

### DIFF
--- a/socorro/processor/mozilla_processor_2015.py
+++ b/socorro/processor/mozilla_processor_2015.py
@@ -46,6 +46,7 @@ mozilla_processor_rule_sets = [
         "socorro.processor.general_transform_rules.CPUInfoRule, "
         "socorro.processor.general_transform_rules.OSInfoRule, "
         "socorro.processor.mozilla_transform_rules.BetaVersionRule, "
+        "socorro.processor.mozilla_transform_rules.AuroraVersionFixitRule, "
         "socorro.processor.mozilla_transform_rules.ExploitablityRule, "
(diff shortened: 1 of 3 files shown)


willkg:
The rule is really basic and the test covers it pretty well. You can test this out on a crash like this one:

https://crash-stats.mozilla.com/report/index/d3fb14c9-e1b5-41ce-a3b9-e2d3d0171120

That was processed in -prod. The raw crash data has the version as "58.0" because apparently that's how it works in beta channel. The processor in -prod sees that and changes it to "58.0b0" because it thinks it's a beta-channel crash, but is puzzled that it doesn't have a beta-style version.

This new rule sees that crash, checks the BuildID, and then stomps on the version with the correct one which is "58.0b1".

This rule is a temporary fix that has maintenance issues. I don't think there's any way around that at the moment without investing a lot more time. I keep track of release weeks, so for now, I'll add something to my calendar to update this rule for 59.0b1. Maybe by then we'll have tackled an ftpscraper rewrite or switched to buildhub. If not by then, maybe soon after.

willkg:
Thank you!

PR #4213, opened 2017-11-21
Bug 1411036: Add stage-only crontabber job for monitoring dependencies.

Osmose:
This leverages crontabber + sentry to run and report security checks against our dependencies using [safety](https://github.com/pyupio/safety) and [nsp](https://github.com/nodesecurity/nsp/). The job is set to only run on staging, once per day.

The new job relies on two configuration values:

- `crontabber.class-DependencySecurityCheckCronApp.node_modules`, which should point to a `node_modules` folder on the local filesystem where the job is being run that the dependencies from `webapp-django/package.json` have been installed.

  `scripts/bootstrap.sh` creates this at `webapp-django/node_modules`, but I'm not sure if the crontabber nodes run that script. If not, then we'll need to figure out another way of getting crontabber to install them.
- `secrets.sentry.dsn`, which can either be blank, or set to a Sentry DSN url. If set, the job will log vulnerabilities to Sentry, otherwise it logs them to the console. I'm guessing this is already set.

You can test this locally:

1. Add known-insecure dependencies:
   - In `requirements/default.txt`:
     
     ```
     insecure-package==0.1.0 \
         --hash=sha256:4da7b05a052f19841c1efbd4cc8edea9a78aaaa295360923cb2b052b547b7877 \
         --hash=sha256:4d9485a53c970c892f887e10fd99e97b82848ac9010e79bc715585e1a6a74422
     ```
   - In `webapp-django/package.json`, under the `dependencies` key:

     ```
     "wenluhong1": "1.0.0"
     ```
2. `make dockerbuild`
3. Add the following line to `my.env` to ensure crontabber is using the list of staging jobs:

   ```
   crontabber.jobs=socorro.cron.crontabber_app.STAGE_JOBS
   ```
4. Reset and run the crontabber job manually:

   ```sh
   docker-compose run crontabber ./scripts/socorro crontabber --reset-job=socorro.cron.jobs.monitoring.DependencySecurityCheckCronApp
   docker-compose run crontabber ./scripts/socorro crontabber --job=socorro.cron.jobs.monitoring.DependencySecurityCheckCronApp --crontabber.class-DependencySecurityCheckCronApp.node_modules=/webapp-frontend-deps/node_modules
   ```
5. Confirm that the output of the second command includes logging stating that both libraries have known vulnerabilities.
6. Visit http://localhost:8000/crontabber-state/ and confirm that `dependency-security-check` shows up in the UI as a successfully-run job.
7. Optionally, if you have a Sentry DSN handy, add `secrets.sentry.dsn` to `my.env` with the DSN, run the command, and confirm that your Sentry instance received a new error with information about the vulnerabilities.


Osmose:
Rebased and fixed the tests.

docker/Dockerfile.crontabber
```diff
@@ -1,5 +1,18 @@
...
+    echo 'deb https://deb.nodesource.com/node_8.x jessie main' > /etc/apt/sources.list.d/nodesource.list && \
+    echo 'deb-src https://deb.nodesource.com/node_8.x jessie main' >> /etc/apt/sources.list.d/nodesource.list
+RUN DEBIAN_FRONTEND=noninteractive apt-get update
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y nodejs
+
+# Install developer dependencies from the webapp
+COPY ./webapp-django/package.json /webapp-frontend-deps/package.json
+COPY ./webapp-django/package-lock.json /webapp-frontend-deps/package-lock.json
+RUN cd /webapp-frontend-deps/ && npm install --only=dev
+
```
willkg: This is all repeated from `Dockerfile.webapp`. Maybe it's a better idea to have `Dockerfile.crontabber` derive from `local/socorro_webapp` instead of `local/socorro_base`? Either that or move most of these lines (everything except the `COPY` ones) into a `setup_node_deps.sh` script which gets run in both `Dockerfile.webapp` and `Dockerfile.crontabber`?

socorro/cron/jobs/monitoring.py
```diff
@@ -0,0 +1,171 @@
...
+        process = Popen(
+            [
+                os.path.join(self.config.node_modules, '.bin', 'nsp'),
+                'check',
+                '--reporter=json',
+            ],
+            stdin=PIPE,
+            stdout=PIPE,
+            stderr=PIPE,
+            cwd=os.path.join(REPO_ROOT, 'webapp-django'),
```
willkg: If `node_modules` is always a subdirectory of `webapp-django`, wouldn't it be better to calculate the cwd with something like:

```
os.path.dirname(self.config.node_modules)
```

?

Then they're tied together config-wise. The -stage and -prod servers use the Socorro RPM and not a git checkout.

socorro/cron/jobs/monitoring.py
```diff
@@ -0,0 +1,171 @@
...
+    required_config = Namespace()
+    required_config.add_option(
+        'node_modules',
+        doc=(
+            'Path to node_modules directory where the webapp\'s npm dependencies have been '
+            'installed.'
+        ),
+    )
+
+    def run(self):
```
willkg: What happens if `node_modules` isn't configured? I have vague memories of bumping into that and having to write an explicit check to make sure it has a value.

socorro/cron/jobs/monitoring.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def get_javascript_vulnerabilities(self):
+        """Check JavaScript dependencies via the nsp command.
+
+        :returns list(Vulnerability):
+        :raises DependencySecurityCheckFailed:
+        """
+        process = Popen(
+            [
+                os.path.join(self.config.node_modules, '.bin', 'nsp'),
```
willkg: We should validate this first. Maybe something like this:

```
nsp_path = os.path.join(self.config.node_modules, '.bin', 'nsp')
if not (os.path.exists(nsp_path) and os.path.isfile(nsp_path)):
    raise ValueError('node_mdules is set wrogn!')
```

socorro/cron/jobs/monitoring.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def get_python_vulnerabilities(self):
+        """Check Python dependencies via Pyup's safety command.
+
+        :returns list(Vulnerability):
+        :raises DependencySecurityCheckFailed:
+        """
+        # Safety checks what's installed in the current virtualenv, so no need
+        # for any paths.
+        process = Popen(['safety', 'check', '--json'], stdin=PIPE, stdout=PIPE, stderr=PIPE)
```
willkg: We should use an absolute path for `safety` so we know exactly which one is being executed.

Bandit says this:

```
--------------------------------------------------
>> Issue: [B607:start_process_with_partial_path] Starting a process with a partial executable path
   Severity: Low   Confidence: High
   Location: socorro/cron/jobs/monitoring.py:111
110	        # for any paths.
111	        process = Popen(['safety', 'check', '--json'], stdin=PIPE, stdout=PIPE, stderr=PIPE)
112	        output, error_output = process.communicate()
```

socorro/cron/jobs/monitoring.py
```diff
@@ -0,0 +1,171 @@
...
+    def key(self):
+        return '[%s] %s' % (self.type, self.dependency)
+
+    @property
+    def summary(self):
+        return 'Installed: %s; Affected: %s; %s' % (
+            self.installed_version,
+            self.affected_versions,
+            self.description,
+        )
```
willkg: Periodically, I think about using `attrs` for things like this. Glyph wrote a blog post about it which I look back on periodically:

https://glyph.twistedmatrix.com/2016/08/attrs.html

If we find ourselves doing more of these, maybe we should look into switching to `attrs`.

socorro/unittest/cron/jobs/test_monitoring.py
```diff
@@ -0,0 +1,256 @@
...
+    popen_patch = mock.patch('socorro.cron.jobs.monitoring.Popen')
+    popen = popen_patch.start()
+
+    process = popen.return_value
+    process.returncode = return_code
+    process.communicate.return_value = (output, error_output)
+
+    yield popen
+
+    popen_patch.stop()
```
willkg: When I was working on Antenna, I made a bunch of things like this as pytest fixtures. This is fine, but having it as a pytest fixture might be slightly more ergonomic.

https://github.com/mozilla-services/antenna/blob/1020bb2883ae9ab8d654a8a4e2905b3da68e6a45/tests/unittest/conftest.py#L117

willkg: Though, maybe switching to a pytest fixture doesn't help a whole lot here. I wonder if we mock popen elsewhere in Socorro.

Osmose: The problem is that pytest fixtures don't work right with `unittest.TestCase` subclasses. You can't do the "include an argument named after the fixture" magic, and instead have to mark the test functions as manually using the fixture. And I couldn't even get that to work, actually; the docs only mention using class-scoped fixtures in this way.

However, the only reason I was using `IntegrationTestBase` was for `_setup_config_manager`, which I'm not using anymore. I'll refactor the tests to make this a proper fixture,

willkg: Oh right. I totally forgot about that. I have vague memories of that issue being alleviated with pytest 4. I should look that up, but for now I'm just going to hope it's true and feel good about the world.

socorro/cron/jobs/monitoring.py
Osmose: It is not always a subdirectory of `webapp-django`. In the docker containers it is under `/weapp-frontend-deps/node_modules`.

I can change this to `PROJECT_ROOT` to reflect the lack of a repo in the RPM, though.

socorro/cron/jobs/monitoring.py
Osmose: Should we just switch to having `nsp_path` and `safety_path` options?

willkg: That's a good way to solve it! +1

Osmose:
Rebased and added review fixes. I also added a make command to make testing this locally easier, just run `make dockerdependencycheck` to run the cronjob locally.

docker/config/local_dev.env
```diff
@@ -92,3 +92,11 @@ SECRET_KEY=secretkey
 STATSD_HOST=statsd
 SYMBOLS_BUCKET_DEFAULT_NAME=symbols
 SYMBOLS_BUCKET_EXCEPTIONS=
+
+# crontabber
+# ----------
+# In local development, it's useful to run the staging job set so we can run security package
+# checks.
+crontabber.jobs=socorro.cron.crontabber_app.STAGE_JOBS
```
Osmose: As of right now this is a superset of the non-stage jobs and should be fine, but maybe we don't intend that going forward?

docker/run_dependency_checks.sh
```diff
@@ -0,0 +1,12 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Runs dependency checks. Must be run from within the crontabber docker container.
+
+set -e
+
+./scripts/socorro crontabber --reset-job=socorro.cron.jobs.monitoring.DependencySecurityCheckCronApp
+./scripts/socorro crontabber --job=socorro.cron.jobs.monitoring.DependencySecurityCheckCronApp
```
Osmose: These commands output a ton of logging with environment variable values, and I couldn't figure out a way to suppress them. It'd be really nice if we could make the output less noisy.

willkg:
> Rebased and added review fixes. I also added a make command to make testing this locally easier, just run make dockerdependencycheck to run the cronjob locally.

That's a brilliant idea!

docker/config/local_dev.env
willkg: We need all the jobs in the jobs list available to run/test/develop-on in the local dev environment. When I do the `raw_adi` mover for -stage-new, we might want to change that, but it'll be tricky because I think we'll have two jobs with the same name.

I think what you did is fine. I'm not sure how to do this better in the confines of the current system.

docker/run_dependency_checks.sh
willkg: OMG. SOOOO MUCH LOGGING!

I wrote up a bug to redo how logging works and either we can tackle this directly in that bug or at least pave the way for a better world.

Makefile
```diff
@@ -77,3 +77,6 @@ dockerrun: my.env
 
 dockerstop: my.env
 	${DC} stop
+
+dockerdependencycheck: my.env
+	${DC} run crontabber ./docker/run_dependency_checks.sh
```
willkg: I've been mulling over how to do a better job of local dev environment commands. Makefiles are ok especially when rules require other rules, but they don't easily take arguments and don't yield help and are harder to document.

At some point, I think we want to have a "standard Makefile" and put everything else into some kind of command structure that runs things, is easier to document (or self-documenting, is extensible, and flexible. Just thinking out loud.

socorro/unittest/cron/jobs/test_monitoring.py
```diff
@@ -0,0 +1,289 @@
...
+
+    def _mock_popen(return_code, output='', error_output=''):
+        process = popen.return_value
+        process.returncode = return_code
+        process.communicate.return_value = (output, error_output)
+        return popen
+
+    yield _mock_popen
+
+    popen_patch.stop()
```
willkg: I <3 this about pytest fixtures.

Osmose:
Yay, build is fixed!

PR #4214, opened 2017-11-21
fixes bug 1418488 - drop EmailCampaign, EmailContact, and EmailCampaignsContact

willkg:
Socorro no longer uses these tables, so time to drop them.


willkg:
Tacking on a DO NOT MERGE to this because we don't want to merge it until after we've deployed the requirements updates.

willkg:
Thank you!

After this is deployed to -stage, I'll run the migrations.

PR #4217, opened 2017-11-21
circle config: add a workflows section so we build tags

milescrabill:
this is an intermediary step to revamping our circleci config. my team is still figuring out best practices for circleci 2.0. this should, at the very least, stop pushing tons of tags to dockerhub (triggering lots of builds in new-stage) and ensure that we build properly tagged docker containers on socorro tags.


willkg:
Awesome! Thank you!

PR #4219, opened 2017-11-22
fixes bug 1419856 - implements /__version__ endpoint

willkg:
Rather than fix the existing `/status/revision/` endpoint to work with the docker-based infrastructure, I implemented the `/__version__` endpoint we want to switch to anyhow. We can leave `/status/revision/` as is for now.

To test, switch to this branch and then do `docker-compose up webapp` and go to http://localhost:8000/__version__ . Because this is the local dev environment, you'll seee `{}`.


willkg:
^^^ That fixed the Circle failure. Tests pass now.

willkg:
Thank you!

PR #4221, opened 2017-11-27
Add e2e-tests and system checklist to docs

willkg:
The e2e-tests README is written in Markdown, so this includes all the things
we need to use Markdown in the docs.

This adds the socorro system test checklist thing we had in my unofficial
socorro-zero repository. We've used variations of it several times now. I'm
sure there are things that are missing, but it's designed to be an evolving
document.


docs/tests/system_checklist.rst
```diff
@@ -0,0 +1,212 @@
...
+Last updated: November 15th, 2017
+
+
+How to use
+==========
+
+"Significant change" can mean any number of things, so this is just a template.
+You should do the following:
+
+1. Copy and paste the contents of this into a Google doc.
```
Osmose: Is a google doc still the preferred method for this vs adding it as a comment to a PR or something?

willkg: I don't think working through a system checklist is viable in Github PRs or Bugzilla. I think we're looking at Google Docs or Etherpad as better systems for working through the checklist and the non-trivial status bookkeeping.

Having said that, I'll change this line. I wouldn't use GitHub PRs or Bugzilla comments, but if someone else wants to do that, then power to them.

willkg:
^^^ That fixes the Google-Doc centric text.

willkg:
Thank you!

PR #4223, opened 2017-11-29
WIP: fixes bug 1410167 - fix UploadCrashReportJSONSchemaCronApp bucket name

willkg:
This fixes the `UploadCrashReportJSONSchemaCronApp` cron job to use similar
logic to determine the bucket name to use to save the schema file as the
`TelemetryBotoS3CrashStorage` class does.


willkg:
I need to verify this works in the cron job still and write instructions up on how to test.

willkg:
Also, this is actually the wrong bug number. This doesn't fix that issue, it fixes a different related issue.

willkg:
Ugh. I want to change the branch name. Killing this and starting over.

PR #4228, opened 2017-11-30
Do not show other threads link in case we only have a crashing thread

Tayamarn:
Minor improvement: remove "Show other threads" link in case we don't have any other threads to show.

### DIFF
--- a/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
+++ b/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
@@ -625,9 +625,15 @@
                                 {% endfor %}
                             </tbody>
                         </table>
-                        <p>
-                            <a href="#allthreads" data-show="Show other threads" data-hide="Hide other threads">Show other threads</a>
-                        </p>
+                        {% if parsed_dump.threads|length == 1 %}
+                            <p>
+                                No more threads to show.
+                            </p>
+                        {% else %}
+                            <p>
+                                <a href="#allthreads" data-show="Show other threads" data-hide="Hide other threads">Show other threads</a>
+                            </p>
+                        {% endif %}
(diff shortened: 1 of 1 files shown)


willkg:
Is there a crash you had in mind to test this with?

Tayamarn:
Unfortunately, I don't have any such crashes in open access, we have some in our closed instance. =(

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -625,9 +625,15 @@ <h2>Crashing Thread ({{ crashing_thread }}){% if parsed_dump.threads[crashing_th
...
-                        </p>
+                        {% if parsed_dump.threads|length == 1 %}
+                            <p>
+                                No more threads to show.
+                            </p>
+                        {% else %}
+                            <p>
+                                <a href="#allthreads" data-show="Show other threads" data-hide="Hide other threads">Show other threads</a>
+                            </p>
+                        {% endif %}
```
willkg: I don't know offhand if `.threads` could ever be `None` or some other non-int value or 0 or something like that. This doesn't take that into account. In those weird edge cases, it'll continue to show the "Show/Hide other threads" text.

Is there javascript that's looking for this `a` tag? Does it handle when it's not there at all?

Tayamarn: It works in my environment, but we have quite a lot of different modifications to original code and now I'm not sure it's a good idea to merge this without proper testing. I'll probably close it and reopen in case I have some crash to test this code. 
Sorry for wasting your time.

willkg:
I don't think this was a waste of time at all. I appreciate your efforts to upstream this. Thank you!

PR #4230, opened 2017-11-30
Bug 1411036: Add package_json_path config to dependency check cronjob.

Osmose:
Turns out the directory layout in the RPM build is not the same as in the repo. Thus, running the dependency check on staging failed because it determines the path to `package.json` relative to it's location in the repo.

This fixes the issue by making the path to the package.json file configurable.

You can test this locally by running `make dockerdependencycheck`.

### DIFF
--- a/socorro/unittest/cron/jobs/test_monitoring.py
+++ b/socorro/unittest/cron/jobs/test_monitoring.py
@@ -1,4 +1,5 @@
 import json
+import os
 from contextlib import nested
 
 import pytest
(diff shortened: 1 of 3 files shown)


socorro/unittest/cron/jobs/test_monitoring.py
```diff
@@ -15,14 +16,13 @@
...
-    safety_path.write('fake binary', ensure=True)
-    return {
-        'nsp_path': nsp_path.strpath,
-        'safety_path': safety_path.strpath,
-    }
+    config = {}
+    for key in ('nsp_path', 'safety_path', 'package_json_path'):
+        path = tmpdir.join(key)
+        path.write('fake file', ensure=True)
+        config[key] = path.strpath
```
willkg: Nice!

socorro/unittest/cron/jobs/test_monitoring.py
```diff
@@ -214,36 +215,33 @@ def test_get_javascript_vulnerabilities_cannot_parse_output(self, mock_popen, ap
             app.get_javascript_vulnerabilities()
         assert err.value.args[0] == 'Could not parse nsp output'
 
-    def test_run_option_validation(self, tmpdir, app_config):
-        tmpdir.join('directory').mkdir()
-        tmpdir.join('binary').write('fake binary')
+    @pytest.mark.parametrize('option', ('nsp_path', 'safety_path', 'package_json_path'))
```
willkg: I like how you parametrized this.

PR #4237, opened 2017-12-03
fixes bug 1422551 - move rewrite rules to in-memory in processor

lonnen:
I looked at the contents of the Product ID Map table and realized that the data has not changed in years. see: https://crash-stats.mozilla.com/api/ProductIDMap/

This reverts the patch from 1346996 that exposes this info as an API endpoint, and adds an additional patch to materialize that table into an in-mem dictionary. It also amends tests as needed and simplifies the rule logic to avoid some problematic `try: except:`s that obscured rather than revealed failures when authoring the patch.


socorro/processor/mozilla_transform_rules.py
```diff
@@ -391,89 +391,23 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
         super(ProductRewrite, self).__init__(config)
-        self.product_id_map = setup_product_id_map(
-            config,
-            config,
-            None
-        )
+        self.product_map = {
+            "{aa3c5121-dab2-40e2-81ca-7ea25febc110}": "FennecAndroid",
+            "{99bceaaa-e3c6-48c1-b981-ef9b46b67d60}": "MetroFirefox",
+            "{webapprt@mozilla.org}": "WebappRuntime",
```
willkg: Huh. I thought I pulled the values from production when I populated the static data for the local dev environment:

https://github.com/mozilla-services/socorro/blob/master/socorro/external/postgresql/staticdata.py#L233

Figured I'd mention it now even though I haven't double-checked it. I can double-check it tomorrow morning.

lonnen: I was hasty in writing this patch.

The `product_productid_map` table is used in a bunch of ADI-related tasks. The 'rewrite' field is used in this processor rule. Like the API endpoint, this is only considering rows where rewrite was true. The rest of the table should persist as long as ADI does, and it doesn't seem worth migrating out the rewrite column.

The only still-active product in that list is FennecAndroid, which is also the only `rewrite = t` field in the static version of the table. I think that this patch has unnecessary additional rows in the map, and your staticdata covers all of our actual use cases.

I'll amend this momentarily.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -391,89 +391,23 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
-    )
-
     def __init__(self, config):
         super(ProductRewrite, self).__init__(config)
-        self.product_id_map = setup_product_id_map(
-            config,
-            config,
-            None
-        )
+        self.product_map = {
```
lonnen: should this be a:
* `CONSTANT`
* hoist it out to a class variable like `required_confid`

willkg: I was thinking about that. I want to think about this some more and see if there are other tables in similar situations.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -391,89 +391,21 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
-            self.product_id_map[product_id]['product_name']
-        )
-        raw_crash['ProductName'] = new_product_name
-        self.config.logger.debug(
-            'product name changed from %s to %s based '
-            'on productID %s',
-            old_product_name,
-            new_product_name,
-            product_id
-        )
```
willkg: Looking at this rule, I wonder if it'd be nice to have it add a processor note when it kicks off. Logging isn't helpful, but processor notes are super helpful after the fact to debug what happened.

I'm just thinking out loud--not totally sold on the idea and I've never had to debug this specific issue before.

socorro/processor/mozilla_transform_rules.py
willkg: I would make it a "constant" in this module for now. That makes it easier to re-use elsewhere.

I can't think of any good reasons to make it a configuration option at this time. If something ever popped up, we can change it then.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -391,89 +391,21 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
-    product_id_map = {}
-    for product_name, productid, rewrite in product_mappings:
-        product_id_map[productid] = {
-            'product_name': product_name,
-            'rewrite': rewrite
-        }
-    return product_id_map
-
-
 class ProductRewrite(Rule):
```
willkg: If I had my druthers, the processor rules would have documentation about wtf they're doing. For example, this rule stomps on the ProductName but I have no idea why.

We don't have to do that in this PR, but if it's the case that this sort of thing is in your head, might as well throw that into a docstring now.

willkg:
I still need to read the surrounding code and context for this. I'll try to do that tomorrow.

willkg:
I spent some time (finally--really sorry about that) reading the context.

I think this change makes sense. On the surface, one might say that it seems like we now have two `product_productid_map` sources, but I agree with what you've said that we're using the `product_productid_map` table in a bunch of different ways and this change extracts one of those ways without affecting the others. That seems fine and I don't think we become compelled to make other changes now.

So, I think this is good and the only thing we could do is make it a constant. However, I didn't do that with the rule I just made and I didn't because I thought the data should stay with the rule. I did make it a class attribute:

https://github.com/mozilla-services/socorro/blob/b02bc6f389d68e78348f4bf0dc142316e2937063/socorro/processor/mozilla_transform_rules.py#L912

There are similar things in other rules in that file. I think we should follow suit with this change.

socorro/processor/mozilla_transform_rules.py
willkg: I have another comment elsewhere in this PR, but to make it easier, I'm going back on my "make it a module-level constant" and changing my mind to "make it a class attribute since that's what we do with other rules".

socorro/processor/mozilla_transform_rules.py
```diff
@@ -391,89 +391,21 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
-            config,
-            None
-        )
+        self.product_map = {
+            "{aa3c5121-dab2-40e2-81ca-7ea25febc110}": "FennecAndroid",
+        }
 
     def version(self):
-        return '1.0'
+        return '1.1'
```
willkg: This will be the first non-1.0 rule. I'm pretty sure the versions are ignored. I'm hard-pressed to see how they get used.

lonnen: There are a bunch of 2.0 rules: ESRVersionRewrite, PluginContentURL, PluginUserComment, etc. I didn't want to bump to 2 since I had changed the inner workings but not the expected output or input. I'll bump it right to 2.0

socorro/processor/mozilla_transform_rules.py
lonnen: It was more interesting in the abstract, but with a single product rewrite I think it's less interesting to log.

lonnen:
updating branch. will land this PM

PR #4238, opened 2017-12-04
Fix bug 1421988: Update Breakpad.

Osmose:
Built this locally with no issues, and processed 100 or so crashes with no issues locally as well.

If you want to try this locally, edit `set_up_stackwalk.sh` and comment out the if statement around the "Build breakpad" line so that the docker build will manually build breakpad. Build the docker containers, process some crashes, and check that there were no processing errors.


Osmose:
Automation may fail (but I don't think it will). That's okay, it's downloading the pre-built breakpad on the old revision.

Once this lands, I'll run the Taskcluster task to re-build the pre-built breakpad.

PR #4244, opened 2017-12-07
fixes bug 1407655 - implement RawADIMoverCronApp

willkg:
This implements a third fetch-adi-from-hive cron app. This variation has
a source db and a dest db and copies the data in the raw_adi table over.

This has no unit tests. There is a single integration test that's run
as a shell script by hand and it requires manual verification.

### DIFF
--- a/scripts/rawadimover_integration_test.sh
+++ b/scripts/rawadimover_integration_test.sh
@@ -7,7 +7,11 @@
 # This is an integration test for the RawADIMoverCronApp. This test
 # must be run by hand in a local development environment.
 #
-# To run it, do:
+# To run it, first add this to your ``my.env`` file:
+#
+# crontabber.jobs=socorro.cron.crontabber_app.STAGE_NEW_JOBS
+#
+# Then do:
 #
 # ./scripts/rawadimover_integration_test.sh
 #
--- a/scripts/rawadimover_integration_test.sh
+++ b/scripts/rawadimover_integration_test.sh
@@ -1,0 +1,148 @@
+#!/bin/bash
+
+# This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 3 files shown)


willkg:
This cron job is intended to run only on the -stage-new environment and only for a very small amount of time. It's a temporary stepping stone upon which we will trod with muddy boots, then after we will rip it up and throw it out and never think about it again. For that and a lot of technical reasons I'd be happy to bore you with in Austin, there are no unit tests.

I did, however, include an integration test shell script that's hopefully sufficiently documented that any non-me person can run it and feel confident that the code might work as intended. It works like this:

```
$ ./scripts/rawadimover_integration_test.sh
```

That sort of runs it through its paces.

This is low risk. If it's busted, it only affects -stage-new and we'll know because I'll see friendly messages from Sentry.

scripts/rawadimover_integration_test.sh
```diff
@@ -0,0 +1,148 @@
+#!/bin/bash
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This is an integration test for the RawADIMoverCronApp. This test
+# must be run by hand in a local development environment.
+#
+# To run it, do:
```
Osmose: You also need to add `crontabber.jobs=socorro.cron.crontabber_app.STAGE_NEW_JOBS` to `my.env` so that crontabber can find the job.

willkg: Oh noes! I forgot this step. Sorry about that! I'll fix this now.

willkg:
Thank you!

PR #4247, opened 2017-12-07
fixes bug 1398946 - remove fakedata code

willkg:
This removes a bunch of code we use to generate fakedata. We no longer need
this, so this nixes it.


willkg:
This removes a bunch of dead code that we don't use. To test that everything is still ok:

1. wait for Travis and Circle to be happy (Travis is the only one that might have problems, but I'm pretty sure it won't)
2. run:
   ```
   $ make dockerbuild
   $ make dockersetup
   $ make dockerupdatedata
   ```
   This will make sure the local development environment is fine. It should be since I built it in a way that it didn't use the fakedata code.

Even if we messed up, this is low risk and can only affect setting up a fresh db which happens in local development environments and CI.

willkg:
Goodbye fakedata,
we're deleting you today
goodbye
goodbye
goodbye

Goodbye fakedata,
There's nothing you can say
To make me change my mind
Goodbye.

PR #4248, opened 2017-12-07
bug 1424027 - remove unused stored procedures

willkg:


willkg:
To test this:

1. check out master and build a db with `make dockersetup`
2. then check out this branch and run the migration:
   ```
   app@processor:/app$ alembic -c docker/config/alembic.ini upgrade heads
   ```
   It should run without errors.

If you want, go through and verify that these stored procedures aren't used anywhere.

I didn't write a `downgrade`. Doing that *correctly* for stored procedures is a huge pain in the ass for various reasons. If we need to downgrade, it's easier to write up a new migration and roll forward.

willkg:
Thank you!

I'll keep an eye on -stage after it does the next deploy and over the weekend. I'm pretty sure I'm right, but if I'm wrong, we'll see errors in Sentry from crontabber or something else. If there isn't anything by Monday we should be fine.

PR #4250, opened 2017-12-08
fixes bug 1375255 nix vagrant

willkg:
This removes the Vagrant environment and does a first pass at cleaning up scripts.

I don't think we need the `config/*-dist` files anymore, but I'm not 100% positive they're not used in weird ways during RPM building, deploys, or Travis CI, so I didn't remove them in this pass. We can do a separate pass later.


willkg:
To test this:

1. Wait for Travis and Circle to do their thing.
2. Merge this and make sure -stage deploys ok. If that's fine, then we're fine.

Makefile
```diff
@@ -2,35 +2,14 @@
...
-
-test: bootstrap
-	./scripts/test.sh
-
-dev:
-	SOCORRO_DEVELOPMENT_ENV=1 ./scripts/bootstrap.sh
-
-bootstrap:
+# Used by infra to build rpm
+package:
```
willkg: I collapsed the rules this required into this rule in the proper order. It's used by the socorro-infra for deploys and that's it.

scripts/clean.sh
```diff
@@ -2,9 +2,11 @@
 
 echo "this is clean.sh"
 
+# Remove .pyc files
 find ./ -type f -name "*.pyc" -exec rm {} \;
-rm -rf ./google-breakpad/ ./build/ ./breakpad/ ./stackwalk breakpad.tar.gz
 
+# Remove any mdsw build artifacts
+rm -rf ./google-breakpad/ ./build/ ./breakpad/ ./stackwalk breakpad.tar.gz
```
willkg: I don't know why github thinks this line changed--it didn't. I just added a couple of comments to this file.

scripts/defaults
```diff
@@ -2,7 +2,9 @@ echo "this is the file defaults"
 
 export PYTHON=${PYTHON:-"python"}
 export VIRTUAL_ENV=${VIRTUAL_ENV:-"$PWD/socorro-virtualenv"}
-export BUILD_TYPE=${BUILD_TYPE:-"tar"}
+# Valid values for BUILD_TYPE are "rpm" (which we use in socorro-infra) and
+# "ci" # which we use in Travis CI.
+export BUILD_TYPE=${BUILD_TYPE:-"ci"}
```
willkg: When Travis runs, the value is "tar". When our deploys run, the value is "rpm". We don't do tar or deb packaging anymore. So I changed use of `BUILD_TYPE` in the scripts to do things based on two values: "rpm" and "ci".

scripts/install.sh
```diff
@@ -66,11 +66,6 @@ cp socorro_revision.txt $BUILD_DIR/application/socorro
 cp breakpad_revision.txt $BUILD_DIR/application/socorro
 
 
-if [ "$BUILD_TYPE" == "tar" ]; then
-    pushd $BUILD_DIR/application/scripts/config
-    echo "py config files no longer exist, ignoring"
-    #for file in *.py.dist; do cp $file `basename $file .dist`; done
-    popd
```
willkg: All this did was pushd into a directory and then popd out of it. That's essentially a no-op, so I removed this bit and inverted the `if` block.

scripts/package.sh
```diff
@@ -20,7 +20,7 @@ if [ ! -d "$BUILD_DIR" ]; then
 fi
 
 echo "> Building Socorro $BUILD_VERSION ..."
-if [ "$BUILD_TYPE" != "tar" ]; then
+if [ "$BUILD_TYPE" == "rpm" ]; then
```
willkg: We changed "tar" to "ci". If it's not "ci" then it is "rpm".

This block would build an rpm or a deb depending on the value of `BUILD_TYPE`. We don't do debs anymore, so that lets me convert this `if` test to "is it rpm?".

scripts/test.sh
```diff
@@ -110,8 +110,3 @@ $ENV $FS_RESOURCES $PG_RESOURCES $RMQ_RESOURCES $ES_RESOURCES PYTHONPATH=$PYTHON
 pushd webapp-django
 PYTHONPATH=.. $PYTEST
 popd
-
-# lint puppet manifests; bug 976639
-pushd puppet
-find . -name '*.pp' -exec puppet parser validate {} \; -exec puppet-lint $puppet_lint_args {} \;
-popd
```
willkg: We don't have puppet files in the socorro repo anymore. No need to lint them.

willkg:
Circle CI failed, but it was an ephemeral failure with their system. I did a rebuild in Circle CI, but for some reason that rebuild is different than the original build so it keeps the original build bookkeeping even thought he rebuild passed. Maybe I did it wrong somehow? Maybe a bug in Circle CI? Maybe we set up Circle CI wrong in GitHub? Maybe a bug in GitHub? Who knows? Not sure how to fix it.

So, I'm not going to. If it happens again, we can figure it out then.

PR #4251, opened 2017-12-08
Clean up after removal of crash-analysis

willkg:
* removes scripts only used by crash-analysis server
* remove submitter script--we use a different one now
* add comment to crontabber.sh so it has some vague semblance of
  documentation so I don't have to wonder about it again next time


willkg:
I tagged @lonnen to look at this one since he's more familiar with the crash-analysis thing.

This should be super benign since none of these things are used anymore.

willkg:
Thank you!

PR #4252, opened 2017-12-08
fixes bug 1424027 - remove more stored procedures

willkg:
This removes a batch of stored procedures that fall into a few groups:

1. utility functions for DBAs--we don't do this sort of thing anymore
2. orphaned stored procedures used by things we don't have anymore
3. stored procedures that are mysterious and want their origins and purpose
   to remain in the dark shadows


willkg:
To test this:

1. run `make dockersetup` to make sure there aren't angry stored procedures that require the ones that were deleted
2. run `make dockertest` which sort of tests out the migration

willkg:
Thank you!

PR #4253, opened 2017-12-10
[ready] 1409648 gc rule sets part 2

lonnen:
This is ready for consideration.

notable changes:
* this alters processor config keys
* rules are no longer configurable, but like the signature utils can be supplied at run time to make testing easier

I believe this is up-to-date with processor changes as they happened, but it's good to double check. This was in flight for a while.


socorro/processor/processor_2015.py
```diff
@@ -40,26 +37,41 @@
...
-    ],
-    [   # post processing of the processed crash
-        "processed_transform",
-        "processer.processed",
-        "socorro.lib.transform_rules.TransformRuleSystem",
-        "apply_all_rules",
-        ""
-    ],
+        "transform_rules",
+        "socorro.processor.mozilla_transform_rules.ProductRewrite,"
```
willkg: I'd be super if we could get rid of this "list of two strings, but the second string is really a comma-separated list of strings as a string" thing, too. Is it easy to convert it to a list of dicts, maybe? Something like this?:

```
default_rule_set = [
    {
        'name': 'transform_rules',
        'rules': [
            'socorro.processor.mozilla_transform_rules.RollIt',
            'socorro.processor.mozilla_transform_rules.PatIt',
            'socorro.processor.mozilla_transform_rules.MarkItWithABee',
            ...
        ]
    }
]
```

socorro/processor/processor_2015.py
```diff
@@ -85,37 +97,16 @@ class ProcessorRuleSets(RequiredConfig):
...
-            )
+
             required_config[name].add_option(
                 name='rules_list',
                 doc='a list of fully qualified class names for the rules',
                 default=default_rules_str,
-                from_string_converter=str_to_classes_in_namespaces_converter(
-                    name_of_class_option='rule_class'
-                ),
+                from_string_converter=str_to_classes_in_namespaces_converter(),
```
willkg: After seeing the default rules str, I'm wondering how does this work?

Also, I have vague memories of you and I talking about taking rule sets out of configuration altogether, uplifting the configuration bits the rule classes needed to one of the processor classes, and passing that in when instantiating rule classes. Did that not work?

socorro/processor/processor_2015.py
```diff
@@ -158,19 +149,29 @@ def __init__(self, config, quit_check_callback=None):
...
-            )
-            self.rule_system[a_rule_set_name] = (
-                config[a_rule_set_name].rule_system_class(
-                    config[a_rule_set_name],
-                    self.quit_check
+        self.rules = []
+
+        # here we instantiate the rule sets and their rules.
+        if 'transform_rules' not in config:
+            return  # no rules
```
willkg: Given that we've still got `rule_sets` configurable, it's possible that the rule does not have a `transform_rules` name/namespace. I think I'd go back to iterating over the lists in `config.rule_sets`.

willkg: Actually, I take that back--we should nix the configurability of `rule_sets`. There should be one rule set and it should be defined in code and not available in configuration.

socorro/processor/processor_2015.py
```diff
@@ -279,11 +276,72 @@ def _log_job_end(self, success, crash_id):
...
+                close_method = rule.close
             except AttributeError:
-                # guess we don't need to close that rule
-                pass
-        self.config.logger.debug('done closing rules')
+                self.config.logger.debug('%s has no close',
+                                         rule.__class__)
+                # no close method mean no need to close
+                continue
+            close_method()
```
willkg: I think instead of this, I'd add a vestigal `.close()` to `Rule` that's a no-op.

willkg:
I skimmed this. I like where this is going, but I think if it's possible, we should nix the configurability of `rule_set`. I think that simplifies a bunch of code in a convenient way.

socorro/processor/breakpad_transform_rules.py
```diff
@@ -119,11 +119,6 @@ def _interpret_external_command_output(self, fp, processor_meta):
             return {}
 
     def _execute_external_process(self, command_line, processor_meta):
-        if self.config.get('chatty', False):
-            self.config.logger.debug(
-                "External Command: %s",
-                command_line
-            )
```
willkg: I know I've used this a few times to debug configuration with mdsw in the current and new infrastructures. I think I'm ok with ditching it now and if we need something like it again in the future, we can implement something different.

socorro/processor/processor_2015.py
```diff
@@ -7,140 +7,158 @@
...
+        # we switch to Docker, we should change this.
+        '/data/socorro/stackwalk/bin/stackwalker',
+    )
+    required_config.add_option(
+        'result_key',
+        doc='the key where the external process result should be stored '
+            'in the processed crash',
+        default='%s_result' %
+            required_config.command_pathname.default.split('/')[-1]
+            .replace('-', ''),
```
willkg: This is pretty intense. I think it's pulling out the binary name from `command_pathname` config default value. Given that it's the default and always the default, I think I'd prefer using `stackwalker_result` here rather than calculating it.

Same with the default for `return_code_key` below.

socorro/processor/processor_2015.py
```diff
@@ -7,140 +7,158 @@
...
+            'in the processed crash',
+        default='%s_return_code' %
+            required_config.command_pathname.default.split('/')[-1],
+    )
+    required_config.add_option(
+        name='symbols_urls',
+        doc='comma delimited ordered list of urls for symbol lookup',
+        default='https://localhost',
+        from_string_converter=str_to_list,
+        likely_to_be_changed=True
```
willkg: I don't think I ever figured out what `likely_to_be_changed` does. What does it do?

socorro/processor/processor_2015.py
```diff
@@ -158,19 +176,46 @@ def __init__(self, config, quit_check_callback=None):
...
-                'setting up rule set: %s',
-                a_rule_set_name
-            )
-            self.rule_system[a_rule_set_name] = (
-                config[a_rule_set_name].rule_system_class(
-                    config[a_rule_set_name],
-                    self.quit_check
-                )
-            )
+        rule_set = rules or [
```
willkg: I'd prefer if this default set of rules was at the top of the module with a name like `DEFAULT_RULES` because it feels like a "constant", but buried in the code.

socorro/processor/processor_2015.py
```diff
@@ -207,27 +252,22 @@ def process_crash(self, raw_crash, raw_dumps, processed_crash):
...
-            # apply transformations
-            #    step through each of the rule sets to apply the rules.
-            for a_rule_set_name, a_rule_set in self.rule_system.iteritems():
-                # for each rule set, invoke the 'act' method - this method
-                # will be the method specified in fourth element of the
-                # rule set configuration list.
-                a_rule_set.act(
+            # apply_all_rules
+            for rule in self.rules:
+                predicate_result, action_result = rule.act(
```
willkg: This line captures results in variables, but I don't see it doing anything with them. Did you intend to do something with them? If not, I think it's better not to capture them. Plus I think it'll make the linter sad.

socorro/processor/processor_2015.py
lonnen: `likely_to_be_changed` is a flag indicating that this value should be written out when using configman to write out the configuration. With the flag present, configman writes the value as live configuration. Without it, the value is printed with a comment prefix.

willkg: Ahhh... fascinating!

socorro/processor/breakpad_transform_rules.py
lonnen: open to adding it forever also

socorro/processor/processor_2015.py
lonnen: good catch. these were used in a previous context. not sure why the linter didn't catch it? fixed in the next commit.

socorro/processor/processor_2015.py
lonnen: moved it out to a module level constant in the latest patch, and then made it the default arg. This isn't the same pattern as the one used in the `SignatureGenerator` class. Not sure if one or the other is preferable.

socorro/processor/breakpad_transform_rules.py
willkg: The cases where I needed it are all in the past, so I think let's ditch it for now.

socorro/processor/processor_2015.py
willkg: The SignatureGenerator is different in that it has no configman anything in it. Is that what you mean by "it's not the same pattern"? If not, can you explain a bit more?

socorro/processor/processor_2015.py
```diff
@@ -257,33 +294,17 @@ def process_crash(self, raw_crash, raw_dumps, processed_crash):
         # for backwards compatibility:
         processed_crash.completeddatetime = completed_datetime
 
-        self._log_job_end(
-            processed_crash.success,
+        self.config.logger.info(
+            "finishing %s job: %s",
+            'successful' if processed_crash.success else 'failed',
```
willkg: As a side note, when I'm reading processor logs, I get tripped up by this message. It's not uncommon for me to be processing a crash, see this message, assume it means that the crash is completely through the processor including crash storage, kill the processor, and then discover later that the crash hadn't been stored, yet. So then I have to start up the processor and process it again and wait some more.

It'd help me to clarify this logging statement and the other one as being specific to the transform step and not all of processing.

Does that sound prudent? Maybe "starting transform for crash: %s" and "finished %s transform for crash: %s"?

willkg: We can kick this discussion to another PR. It shouldn't hold up this one. And if it doesn't affect anyone else, that's cool--we can leave it out altogether.

socorro/processor/processor_app.py
```diff
@@ -14,10 +14,14 @@
...
 from socorro.app.socorro_app import main
+
 from socorro.external.crashstorage_base import CrashIDNotFound
+from socorro.external.fs.crashstorage import FSDatedPermanentStorage
+
 from socorro.lib.util import DotDict
 from socorro.lib import raven_client
-from socorro.external.fs.crashstorage import FSDatedPermanentStorage
+
+#from socorro.processor.processor_2015 import Processor2015
```
willkg: We can nix this line if you don't need it anymore.

socorro/unittest/processor/test_processor_2015.py
```diff
@@ -4,123 +4,153 @@
...
         cm = ConfigurationManager(
             definition_source=Processor2015.get_required_config(),
-            values_source_list=[{'rule_sets': rule_set_01_str}],
+            values_source_list=[],
         )
         config = cm.get_config()
         config.logger = Mock()
 
-        p = Processor2015(config)
+        Processor2015(config)
```
willkg: I'm not sure I understand what this is testing. Is it testing that we can initialize a `Processor2015` in general? If so, we can probably just nix it since we do that inherently in other tests.

socorro/unittest/processor/test_processor_2015.py
```diff
@@ -4,123 +4,153 @@
...
+
         config.processor_name = 'dwight'
 
-        p = Processor2015(config)
+        p = Processor2015(config, rules=[
+            CPUInfoRule,
+            OSInfoRule,
+        ])
         raw_crash = DotDict()
+        raw_crash.uuid = "1"
```
willkg: configman's `DotDict` lets you create things by passing in a dict. I tend to do this because I find it easier to read and uses less typing:

```
raw_crash = DotDict({
    'uuid': '1'
})
```

In this case it's not much easier to read, but I think it'll have a bigger effect in other cases.

willkg:
I've gone through and read the code, but want to test it out and poke at rule sets a bit.

I think this is a good cleanup step and landing this will make future processor work easier. Yay!

willkg:
@Osmose just created PR #4307 which completely changes the game for behavioral configuration for Socorro apps including the processor. I want to spend time on that and see if we can land that asap since I think it solves a lot of problems and will make working on the new infrastructure significantly easier.

I suspect the changes in that PR will affect this PR. Among other things, it might alleviate the impetus behind some of the changes in this PR. Maybe after that lands, it'd make sense to dial back some of the changes here? I can think more about that after PR #4307 lands.

It's possible I'm reading the tea leaves wrong and we should prioritize landing this PR before that one. If you think that's the case, kick me.

lonnen:
updated with latest master

socorro/unittest/processor/test_processor_2015.py
lonnen: I was matching the lines below. I rewrote them too. Update incoming.

socorro/unittest/processor/test_processor_2015.py
lonnen: Yep, as the name implies it's only testing that the processor initializes correctly. This was useful when moving configuration around. Rules missing config would break loudly here. Could take it or leave it now that initialization is simplified and fully redundant with the other rules.

socorro/processor/processor_2015.py
lonnen: I'm on board. Both places are modified to use the 'starting transform' logic

socorro/processor/processor_2015.py
lonnen: SignatureGenerator uses something like:

```
def __init__(self, rules=None ...):
    self.rules = rules or list(DEFAULT_RULES)
```

where instead I used something like:

```
def __init__(self, rules=DEFAULT_RULES ...):
    self.rules = rules
```

willkg: Ohhh! Ok. So, I think the difference between those two is a style thing. In the SignatureGenerator case (which I wrote, so I'm using my style), I did what I did because of an annoying thing about Python where argument defaults get created at module load time and persist, so it's good to avoid using mutable values as defaults. Additionally, if Sphinx autodoc runs over code like this:

```
DEFAULT_RULES = [some long list of things]

def __init__(self, rules=DEFAULT_RULES):
    ...
```

it generates a (in my opinion) unreadable signature. For example:

https://bleach.readthedocs.io/en/latest/clean.html#bleach.clean

I think what you're doing and what I did in SignatureGenerator are behaviorally equivalent. I'm game for whichever appeals to you.

socorro/processor/processor_2015.py
willkg: I appreciate it!

socorro/processor/processor_2015.py
lonnen: cool. In the latest patch I've updated it to match SignatureGenerator

lonnen:
re-updated

lonnen:
Many conflicts to resolve. I'm going to cut a new branch off master, and then cherry pick each of these commits resolving conflicts as I go. Then I'll force push over this.

willkg:
@lonnen Much of the bitrot here is my fault. I'm happy to do the branch-and-cherry-pick work, if that helps. I can work through it on Tuesday.

lonnen:
Already done

PR #4255, opened 2017-12-11
fixes bug 1424744 - remove testcrash/ directory

willkg:
This removes a directory that was used by the integration tests which were
removed in 6ffefdc.


willkg:
I'm in an airport and can't run the tests for some reason. So this is a blind run. I'll follow up when I land.

willkg:
Thank you!

PR #4256, opened 2017-12-11
Fix test bugzilla

willkg:
This fixes `test_bugzilla.py` which is hanging for me right now and thus I can't run tests locally.

This also fixes the logger-related warning in that file for bug 1419585.

### DIFF
--- a/socorro/unittest/cron/jobs/test_bugzilla.py
+++ b/socorro/unittest/cron/jobs/test_bugzilla.py
@@ -8,9 +8,6 @@ import requests_mock
 from socorro.cron.crontabber_app import CronTabberApp
 from socorro.cron.jobs.bugzilla import find_signatures
 from socorro.cron.jobs.bugzilla import BUGZILLA_BASE_URL
-from socorro.unittest.cron.setup_configman import (
-    get_config_manager_for_crontabber,
-)
 from socorro.unittest.cron.jobs.base import IntegrationTestBase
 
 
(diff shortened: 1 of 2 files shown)


socorro/unittest/cron/jobs/test_bugzilla.py
```diff
@@ -66,9 +63,9 @@ def tearDown(self):
         super(IntegrationTestBugzilla, self).tearDown()
 
     def _setup_config_manager(self, days_into_past):
-        return get_config_manager_for_crontabber(
-            jobs='socorro.cron.jobs.bugzilla.BugzillaCronApp|1d',
-            overrides={
+        return super(IntegrationTestBugzilla, self)._setup_config_manager(
+            jobs_string='socorro.cron.jobs.bugzilla.BugzillaCronApp|1d',
+            extra_value_source={
                 'crontabber.class-BugzillaCronApp.days_into_past': days_into_past,
```
willkg: This test was doing setup differently than all the other crontabber tests. I don't know why. Though this code looks familiar, so maybe it's something I broke/fixed recently.

socorro/unittest/cron/setup_configman.py
```diff
@@ -79,9 +79,7 @@ def get_config_manager_for_crontabber(
     else:
         more_definitions = [CronTabberApp.get_required_config()]
 
-    local_overrides = {
-        'logger': Mock(),
-    }
```
willkg: This line has been in the codebase since this file was written. However, I think it's wrong. The configuration doesn't have a `logger` object--it has a bunch of `logging-blahblahblah` things. Other test setup functions toss in a logger object after the configuration is configurized.

Fixing this makes the warning go away because then the expected keys matches the given keys.

willkg:
Thank you!

PR #4257, opened 2017-12-11
Remove dead code

willkg:
There's so much stuff in these commits that it's not going to display well, so I'll go through what's here and we can rely on Travis/Circle for "yes--it's fine".

1. *Removed the vestiges of the middleware app.* A while ago, we had a webapp and a middleware app and they were two separate web services. Then we folded all the stuff the middleware app did into the webapp so we no longer have a separate middleware app. This removes the last bits (wsgi files, configuration file, etc) for the middleware app that no longer exists.
2. *Removed tools/ things.* There were a few things in the `tools/` folder that we have better versions of now or never used. They have no tests. There aren't docs. Figured I'd just remove them.
3. *Removed example_app.* There was an example app to show how to write configman-style Socorro apps. We don't really do that sort of thing and I claim this file isn't showing how to do anything that isn't shown elsewhere.
4. *Removed product_details files.* A long long time ago, Socorro was a playdoh app and as such used the mozilla-product-details-long-filename Python package which would download JSON files and stick them on the filesystem. It looks like someone checked those files in. Socorro doesn't have that library and doesn't do anything with those files, so we can remove them.

That's it!


willkg:
Thank you!

PR #4258, opened 2017-12-14
Remove MissingSymbolsCronApp from -stage-new list

willkg:
We're not going to do missing symbols with the new infra, so this removes that crontabber app from running in -stage-new.

This is termporary to quell errors we're seeing. We'll be able to remove the crontabber app from all the environments soon.


willkg:
Also of note, this has no effect on -stage, -prod, or local dev environments--just -stage-new.

PR #4259, opened 2017-12-15
fixes bug 1227900 - add metrics gathering for api

willkg:
This uses markus to get frequency and usage data for API endpoints. It's not
granular to the point where we can see querystring data, but I think that's
ok for answering questions like "Is someone using this API endpoint?" which
I really want to answer now.


willkg:
This nixes the google analytics code that we could never get working and implements a vastly reduced version using markus.

My goal is answering "is this API endpoint used? if so, what's the usage frequency like? is it regular (indicating bots) or is it sporadic (indicating human-driven)?"

The goal for the google analytics code was that plus "exactly what filters were people using?" Answering that might be nice, but I don't need to know that now to determine which API endpoints we can remove because they're not used.

The unit test covers testing the code. I don't think it needs manual testing.

willkg:
I'd love to land this and do a -prod deploy next week so we can gather API usage information for a few weeks and then remove API endpoints in January.

willkg:
Thank you! This harkens a new era for API-deprecation-industries!

PR #4260, opened 2017-12-15
bug 1419585 - all warnings are errors for socorro tests

willkg:
This fixes the socorro tests such that all warnings are now considered
errors. This also fixes the handful of warnings that were getting kicked
up.

This doesn't fix the webapp tests--we'll do that in another PR.

### DIFF
--- a/socorro/external/crashstorage_base.py
+++ b/socorro/external/crashstorage_base.py
@@ -420,8 +420,8 @@ class PolyStorageError(Exception, collections.MutableSequence):
     parameters:
         message - an optional over all error message
     """
-    def __init__(self, message=''):
-        super(PolyStorageError, self).__init__(message)
+    def __init__(self, *args):
+        super(PolyStorageError, self).__init__(*args)
         self.exceptions = []  # the collection
 
     def gather_current_exception(self):
@@ -459,7 +459,7 @@ class PolyStorageError(Exception, collections.MutableSequence):
         self.exceptions.__setitem__(index, value)
 
     def __str__(self):
-        return '%s %s' % (self.message,
+        return '%s %s' % (self.args[0],
                           ','.join(repr(e[1]).encode('ascii', 'ignore')
                                    for e in self.exceptions))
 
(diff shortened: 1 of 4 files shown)


socorro/external/crashstorage_base.py
```diff
@@ -420,8 +420,8 @@ class PolyStorageError(Exception, collections.MutableSequence):
     parameters:
         message - an optional over all error message
     """
-    def __init__(self, message=''):
-        super(PolyStorageError, self).__init__(message)
+    def __init__(self, *args):
```
willkg: Python 2.6 deprecated "message" as the first argument for exceptions. That covers most of the changes in this PR.

socorro/external/crashstorage_base.py
```diff
@@ -459,7 +459,7 @@ def __setitem__(self, index, value):
         self.exceptions.__setitem__(index, value)
 
     def __str__(self):
-        return '%s %s' % (self.message,
+        return '%s %s' % (self.args[0],
```
willkg: Because the "message" argument as an attribute is deprecated, the suggested way to use the message is `args[0]`, so I switched to that here and elsewhere.

socorro/unittest/cron/jobs/test_monitoring.py
```diff
@@ -253,15 +252,11 @@ def test_run_log(self, app_config):
...
-            mock.patch.object(app, 'get_javascript_vulnerabilities', return_value=[]),
-            mock.patch.object(app, 'alert_log'),
-        ]
-
-        with nested(*mocks):
-            app.run()
-            app.alert_log.assert_called_with([vuln])
+        with mock.patch.object(app, 'get_python_vulnerabilities', return_value=[vuln]):
+            with mock.patch.object(app, 'get_javascript_vulnerabilities', return_value=[]):
+                with mock.patch.object(app, 'alert_log'):
```
willkg: `contextlib.nested` was deprecated because it has weird corner cases (I don't know what they are offhand). So, this just manually nests them.

There were a couple of instances of this, but that's it. I figured I'd just do this for now and if we have a better way of handling this, we can change it later.

willkg:
Thank you!

PR #4261, opened 2017-12-15
fixes bug 1424370 - remove all the postgres views

willkg:
None of these views are used, so I removed all the views and the
infrastructure for loading views.


willkg:
I went through each view `.sql` file and checked the entire repository to see if that view is used. There's one view that uses two other views, but otherwise, none of these are used anywhere as near as I could tell.

To test, I did:

1. ran `make dockersetup` which creates a db and runs through all the stuff in `raw_sql`
2. ran the tests which run the migration and then run all the tests--if there was anything using the view, it should raise errors here

willkg:
Thank you!

PR #4262, opened 2017-12-18
fixes bug 1419585 - all the webapp warnings

willkg:


requirements/default.txt
```diff
@@ -207,9 +207,9 @@ django-npm==1.0.0 \
 markus==1.1 \
     --hash=sha256:813ac8cd0eb68bcf7ee618ac2544ce20de5eb0ba27b12f50973c29742800b058 \
     --hash=sha256:821377379e0d225a870996e4dcd1aa5305e4030a8b9b5d8b38dc843cbc30dfa8
-pinax-eventlog==1.1.2 \
-    --hash=sha256:28467f74a9c7ff1b838a43075f46f875daa8a7b37d7dabf097a15b162ef35411 \
-    --hash=sha256:e9c2947fed5773de7cbc610501729d16262a0a913026fad762e4ba940a071258
+pinax-eventlog==2.0.0 \
+    --hash=sha256:ab77944098e6acb0c935a368333f86bf4732106a76a284cb8bc13d3986002245 \
+    --hash=sha256:3261b495ab25842c22e1c38ba3796271f0b7f44db7050d1484eda0148d68fa5d
```
willkg: pinax-eventlog 2.0.0 fixes the migrations so we don't get "deprecated in django 2.0" warnings from them.

webapp-django/crashstats/authentication/context_processors.py
```diff
@@ -8,7 +8,7 @@ def oauth2(request):
     # By default we don't want to encourage the client-side code to
     # sign out, so set this to a falsy value.
     signout = False
-    if getattr(request, 'user', None) and request.user.is_authenticated():
+    if getattr(request, 'user', None) and request.user.is_authenticated:
```
willkg: `is_authenticated` as a function is deprecated in Django 2.0.

webapp-django/crashstats/base/views.py
```diff
@@ -16,7 +16,7 @@ def handler500(request):
     return render(request, '500.html', context, status=500)
 
 
-def handler404(request):
+def handler404(request, exception, template_name='404.html'):
```
willkg: Django changed the signatures for error handlers, so we needed to update ours.

webapp-django/crashstats/crashstats/middleware.py
```diff
@@ -14,11 +14,18 @@ class SetRemoteAddrFromForwardedFor(object):
     anybody can "fake" their IP address. Only use this when you can absolutely
     trust the value of HTTP_X_FORWARDED_FOR.
     """
+    def __init__(self, get_response=None):
+        self.get_response = get_response
```
willkg: This redoes our middlewares to support the new middleware API. The new API has some differences in structure and also supports a shortcut property. I thought about using the middleware mixin, but fixing our middlewares seemed pretty straight-forward, so I did that instead.

https://docs.djangoproject.com/en/2.0/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware

webapp-django/crashstats/crashstats/middleware.py
```diff
@@ -14,11 +14,18 @@ class SetRemoteAddrFromForwardedFor(object):
     anybody can "fake" their IP address. Only use this when you can absolutely
     trust the value of HTTP_X_FORWARDED_FOR.
     """
+    def __init__(self, get_response=None):
```
willkg: We have some tests that create a middleware and in the test we don't have a `get_response` value that's helpful, so I did what the Django middlewares and tests do and have it default to `None`. That totally breaks `__call__` since it expects `get_response` to be something. It feels a little weird and skimming the code suggests it's an error.

Maybe we could fix all the tests to pass in a `get_response` that returns a basic Django Response object. Then we wouldn't need the default. Would that be better?

webapp-django/crashstats/settings/base.py
```diff
@@ -757,7 +757,7 @@ def path(*dirs):
     'https://www.google-analytics.com',
     'data:',  # what jquery.tablesorter.js's CSS uses
 )
-CSP_CHILD_SRC = (
+CSP_FRAME_SRC = (
```
willkg: `child-src` is deprecated and being replaced with `frame-src` and `worker-src`. @peterbe had a conversation with someone at some point about this and we left it as `child-src`. I couldn't find that conversation, though. Anyhow, so I'm switching it to `frame-src` figuring that conversation happened a while ago and it's probably fine to switch now.

webapp-django/crashstats/symbols/migrations/0001_initial.py
```diff
@@ -22,7 +22,7 @@ class Migration(migrations.Migration):
                 ('size', models.IntegerField()),
                 ('created', models.DateTimeField(default=django.utils.timezone.now)),
                 ('content_type', models.TextField(null=True)),
-                ('user', models.ForeignKey(to=settings.AUTH_USER_MODEL)),
+                ('user', models.ForeignKey(to=settings.AUTH_USER_MODEL, on_delete=models.CASCADE)),
```
willkg: Django requires an `on_delete` argument. It used to default to `models.CASCADE`, so this doens't change the behavior at all.

webapp-django/crashstats/settings/base.py
peterbe: I can't remember which is which but the predicament was that if we upgraded django-csp to the latest, it would complain about one of them not being set and that you should set the other (e.g. use child-src, not frame-src (or the other way around)). I checked with Greg and it was correct. 
The problem is that even with Nightly (at the time) you would get a CSP warning every single time in the web console because the new CSP spec hasn't made it to all browsers yet. 
So I decided to hold off a wee bit until Nightly would get that fixed so you wouldn't have to have a warning in the web console every single time.

willkg: Ahhh... ok. We updated django-csp, so we don't have that problem anymore. And according to MDN, frame-src is supported by most browsers.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-src#Browser_compatibility

So I think we're good now. Thank you, @peterbe!

webapp-django/crashstats/crashstats/middleware.py
Osmose: I think in the end we'll want to fix the tests and the middleware to be shaped like newer middleware anyway, but what's in this PR now is fine as an in-between if we don't want to do all that rewriting right now.

willkg:
Thank you!

webapp-django/crashstats/crashstats/middleware.py
willkg: That make sense. +1

PR #4263, opened 2017-12-18
bug 1425245 - log boto fetch errors

willkg:
This adds logging for boto fetch errors. Previously, error messages
were getting thrown away.

This also fixes logging for crashstats models. Previously the logging
configuration didn't cover them, so they were silently ignored.

### DIFF
--- a/socorro/external/boto/crash_data.py
+++ b/socorro/external/boto/crash_data.py
@@ -72,7 +72,8 @@ class SimplifiedCrashData(BotoS3CrashStorage):
                 return get(params.uuid, name=params.name)
             else:
                 return get(params.uuid)
-        except CrashIDNotFound:
+        except CrashIDNotFound as cidnf:
+            self.config.logger.error('%s not found: %s' % (params.datatype, cidnf))
             # The CrashIDNotFound exception that happens inside the
             # crashstorage is too revealing as exception message
             # contains information about buckets and prefix keys.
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/crashstats/models.py
```diff
@@ -40,7 +40,7 @@
 from crashstats.base.utils import requests_retry_session
 
 
-logger = logging.getLogger('crashstats_models')
+logger = logging.getLogger('crashstats.models')
```
willkg: Changing it from `crashstats_models` to `crashstats.models` means that it get covered by the existing configuration. Nothing else looks at `crashstats_models` and it's a webapp-specific bit of code, so this is no-risk.

socorro/external/boto/crash_data.py
```diff
@@ -72,7 +72,8 @@ def get(self, **kwargs):
                 return get(params.uuid, name=params.name)
             else:
                 return get(params.uuid)
-        except CrashIDNotFound:
+        except CrashIDNotFound as cidnf:
+            self.config.logger.error('%s not found: %s' % (params.datatype, cidnf))
```
willkg: Since this block throws away the original exception and creates a new one with no interesting information makes debugging really difficult. This adds a logging message so we can see what's going on in the logs.

willkg:
I messed up the branch name. Closing this and will open the right one.

PR #4264, opened 2017-12-18
bug 1425492 - log boto fetch errors

willkg:
This adds logging for boto fetch errors. Previously, error messages
were getting thrown away.

This also fixes logging for crashstats models. Previously the logging
configuration didn't cover them, so they were silently ignored.

### DIFF
--- a/socorro/external/boto/crash_data.py
+++ b/socorro/external/boto/crash_data.py
@@ -72,7 +72,8 @@ class SimplifiedCrashData(BotoS3CrashStorage):
                 return get(params.uuid, name=params.name)
             else:
                 return get(params.uuid)
-        except CrashIDNotFound:
+        except CrashIDNotFound as cidnf:
+            self.config.logger.error('%s not found: %s' % (params.datatype, cidnf))
             # The CrashIDNotFound exception that happens inside the
             # crashstorage is too revealing as exception message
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/crashstats/models.py
```diff
@@ -40,7 +40,7 @@
 from crashstats.base.utils import requests_retry_session
 
 
-logger = logging.getLogger('crashstats_models')
+logger = logging.getLogger('crashstats.models')
```
willkg: This fixes the logger name so it's covered by our existing configuration.

socorro/external/boto/crash_data.py
```diff
@@ -72,7 +72,8 @@ def get(self, **kwargs):
                 return get(params.uuid, name=params.name)
             else:
                 return get(params.uuid)
-        except CrashIDNotFound:
+        except CrashIDNotFound as cidnf:
+            self.config.logger.error('%s not found: %s' % (params.datatype, cidnf))
```
willkg: Since this block throws away the exception and all interesting information about it, this logs at least the message so we can see issues in the logs.

willkg:
I'm self-r+ing this. It's benign and it'll probably help us debug the -stage-new environment.

PR #4265, opened 2017-12-18
bug 1425492 - set the region for webapp s3 access

willkg:


webapp-django/crashstats/settings/base.py
```diff
@@ -654,8 +654,8 @@ def path(*dirs):
         },
         'boto': {
             'access_key': config('resource.boto.access_key', None),
-            'bucket_name': config(
-                'resource.boto.bucket_name', 'crashstats'),
+            'bucket_name': config('resource.boto.bucket_name', 'crashstats'),
+            'region': config('resource.boto.region', 'us-west-2'),
```
willkg: This moves `bucket_name` to one line which is cosmetic. This also adds `region` using the default from the connection class. Otherwise we have no way of changing the region via env vars in the webapp.

willkg:
This is a no-op for the -stage and -prod environments since it just exposes an existing configuration to be configurable, but we use the default value for -stage and -prod so it doesn't change anything. This does fix our current problem with -stage-new.

So I'm self-r+ing this.

PR #4266, opened 2017-12-19
fixes bug 1426144 - fix job spec for fetch-adi-from-hive jobs

willkg:
The fake one and the rawadimover one were set to run at the wrong time
in relation to the adu jobs that need that data. This fixes that.


willkg:
^^^ That should fix the race condition issue that @milescrabill pointed out in IRC.

PR #4267, opened 2017-12-19
bug 1426225 - capture fetch raw crash errors

willkg:
The part of the processor that fetches raw crash metadata and dumps and
then stores bits on disk could fail and when it did, it'd log something, but
unless someone was watching the logs at the time, no one would ever know.

This fixes that in the same way we capture other processing errors--it
sends any errors to sentry assuming sentry is configured.

### DIFF
--- a/socorro/processor/processor_app.py
+++ b/socorro/processor/processor_app.py
@@ -77,6 +77,43 @@ class ProcessorApp(FetchTransformSaveWithSeparateNewCrashSourceApp):
             "destination.crashstorage_class": FSDatedPermanentStorage,
         }
 
+    def _capture_error(self, crash_id, exc_type, exc_value, exc_tb):
+        """Capture an error in sentry if able
+
+        The `exc_*` arguments come from calling `sys.exc_info`.
+
+        :arg crash_id: a crash id
+        :arg exc_type: the exc class
+        :arg exc_value: the exception
+        :arg exc_tb: the traceback for the exception
+
+        """
+
+        if self.config.sentry and self.config.sentry.dsn:
+            try:
+                if isinstance(exc_value, collections.Sequence):
+                    # Then it's already an iterable!
+                    exceptions = exc_value
+                else:
+                    exceptions = [exc_value]
+                client = raven_client.get_client(self.config.sentry.dsn)
+                client.context.activate()
+                client.context.merge({'extra': {
+                    'crash_id': crash_id,
+                }})
+                try:
+                    for exception in exceptions:
+                        identifier = client.captureException(exception)
+                        self.config.logger.info(
+                            'Error captured in Sentry! Reference: {}'.format(identifier)
+                        )
+                finally:
+                    client.context.clear()
+            except Exception:
+                self.config.logger.error('Unable to report error with Raven', exc_info=True)
+        else:
+            self.config.logger.warning('Sentry DSN is not configured and an exception happened')
+
     def _transform(self, crash_id):
         """this implementation is the framework on how a raw crash is
         converted into a processed crash.  The 'crash_id' passed in is used as
@@ -87,12 +124,18 @@ class ProcessorApp(FetchTransformSaveWithSeparateNewCrashSourceApp):
             raw_crash = self.source.get_raw_crash(crash_id)
             dumps = self.source.get_raw_dumps_as_files(crash_id)
         except CrashIDNotFound:
+            # If the crash isn't found, we just reject it--no need to capture
+            # errors here
             self.processor.reject_raw_crash(
                 crash_id,
                 'this crash cannot be found in raw crash storage'
             )
             return
         except Exception as x:
+            # We don't know what this error is, so we should capture it
+            exc_type, exc_value, exc_tb = sys.exc_info()
+            self._capture_error(crash_id, exc_type, exc_value, exc_tb)
+
             self.config.logger.warning(
                 'error loading crash %s',
                 crash_id,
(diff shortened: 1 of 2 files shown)


willkg:
Right now -stage-new is throwing "disk is full" errors and no one would have noticed until someone looked at the logs. This fix surfaces those issues to Sentry where we'll get notified of what's going on.

socorro/processor/processor_app.py
```diff
@@ -77,6 +77,43 @@ def get_application_defaults():
...
+                        identifier = client.captureException(exception)
+                        self.config.logger.info(
+                            'Error captured in Sentry! Reference: {}'.format(identifier)
+                        )
+                finally:
+                    client.context.clear()
+            except Exception:
+                self.config.logger.error('Unable to report error with Raven', exc_info=True)
+        else:
+            self.config.logger.warning('Sentry DSN is not configured and an exception happened')
```
willkg: This is almost exactly the same code as before, except I extracted it into a separate method and cosmetically tweaked some lines.

socorro/unittest/processor/test_processor_app.py
```diff
@@ -277,6 +277,48 @@ def mocked_save_raw_and_processed(*_):
         assert captured_exception.__class__ == ValueError
         assert captured_exception.args[0] == 'Someone is wrong on the Internet'
 
+    @mock.patch('socorro.lib.raven_client.raven')
+    def test_transform_get_error_with_raven_configured_successful(self, mock_raven):
```
willkg: This test mirrors the other one, except that it covers fetching crashes and the other one covers saving crashes.

socorro/processor/processor_app.py
```diff
@@ -77,6 +77,43 @@ def get_application_defaults():
...
+
     def _transform(self, crash_id):
```
willkg: At some point, we might want to refactor `_transform` so that it's more coherent about error handling through all stages of transforming without nearly as many try/except and other blocks. I bet we could unify it and make it much easier to follow.

socorro/unittest/processor/test_processor_app.py
```diff
@@ -277,6 +277,48 @@ def mocked_save_raw_and_processed(*_):
...
+        # First mock everything so we're not sending real things anywhere
+        captured_exceptions = []  # a global
+
+        def mock_capture_exception(exc_info=None):
+            captured_exceptions.append(exc_info)
+            return 'someidentifier'
+
+        raven_mock_client = mock.MagicMock()
+        raven_mock_client.captureException.side_effect = mock_capture_exception
```
Osmose: Instead of using the global list, could we just assert that it was called with the proper argument?

```python
mock_raven_client = mock_raven.Client.return_value
mock_raven_client.captureException.assert_called_once_with(expected)
```

That'd save a lot of the boilerplate in this test.

socorro/unittest/processor/test_processor_app.py
```diff
@@ -277,6 +277,48 @@ def mocked_save_raw_and_processed(*_):
...
+        config = self.get_standard_config(
+            sentry_dsn='https://abc123@example.com/project'
+        )
+        pa = ProcessorApp(config)
+        pa._setup_source_and_destination()
+
+        def mocked_get_raw_crash(*_):
+            raise ValueError('simulated error')
+
+        pa.source.get_raw_crash.side_effect = (
```
Osmose: IIRC if we just set `side_effect` to the exception itself, it will throw it when called automatically.

socorro/unittest/processor/test_processor_app.py
```diff
@@ -277,6 +277,48 @@ def mocked_save_raw_and_processed(*_):
...
+            mocked_get_raw_crash
+        )
+
+        # Note that the processor silently ignores .get_raw_crash() and
+        # .get_raw_dumps_as_files() errors and only logs stuff with the
+        # assumption that they're ephemeral
```
Osmose: Did you mean to say that it ignores errors when the crash ID isn't found here? Otherwise isn't this false?

socorro/unittest/processor/test_processor_app.py
willkg: We look at the exception message to make sure it's the exception we threw here rather than some other exception thrown elsewhere. I think we need a function for that. If not, can you show an example of what you're thinking about?

willkg:
^^^ That cleans up all the sentry-related tests so there's less boilerplate and clunkery.

socorro/unittest/processor/test_processor_app.py
Osmose: Just do an equality check against the exception you assign to `side_effect`:

```python
>>> import mock
>>> mock_thing = mock.Mock()
>>> error = ValueError('foo')
>>> mock_thing.foo.side_effect = error
>>> try:
...     mock_thing.foo()
...     assert false
... except ValueError as err:
...     assert err == error
```

Or, in this specific case, assign the same exception to `side_effect` that you pass to `assert_called_once_with`.

socorro/unittest/processor/test_processor_app.py
```diff
@@ -185,97 +185,120 @@ def mocked_save_raw_and_processed(*_):
...
+        # Assert that the logger logged the appropriate thing
+        config.logger.info.assert_called_with(
+            'Error captured in Sentry! Reference: someidentifier'
+        )
+
+    @mock.patch('socorro.lib.raven_client.raven')
+    def test_transform_get_error_with_raven_configured_successful(self, mock_raven):
+        # Mock everything
+        def mock_capture_exception(exc_info=None):
+            return 'someidentifier'
```
Osmose: Since this is now just returning a string, we can just use the `return_value` attribute on the mock and avoid defining an extra function at all.

socorro/unittest/processor/test_processor_app.py
Osmose: The code that calls `get_raw_crash` and `get_raw_dumps_as_files` catches exceptions from them and rejects the raw crash if they're not an error about the crash ID not being found, but this comment implies that it should instead silently ignore those errors. That seems to be a contradiction.

willkg:
It's cool. I'm not very familiar with the mock library, so you're pointing out helpful things.

socorro/unittest/processor/test_processor_app.py
willkg: I don't see how this is a contradiction and I'm not following what you're saying. I feel stupid and don't want to belabor this. I don't want to have obtuse comments--that doesn't help anyone. I'll try rewording it and maybe that'll be better.

willkg:
^^^ That cleans up the things you pointed out in the last round. It's a lot cleaner now. I think I addressed everything. Sorry in advance if I didn't!

willkg:
Also, now this PR nets -2 LOC. Yay!

willkg:
Thank you! This was super helpful!

PR #4269, opened 2017-12-20
Tweak post-deploy steps in deploy_bug script

willkg:
This is a minor tweak to the post-deploy steps. I was updating the docs and it occurred to me we were missing a few that we sometimes do.


willkg:
Thank you!

PR #4271, opened 2017-12-29
Ack RMQ message on exactly same channel

Lehych:
RMQ message should be acknowlodged on same channel and connection it was
recived. Lets store channel near with the message.
Here also removed rety on basic_ack since if ack failed connection or
channel error it means we could not ackowledge task anymore.

What lead us to this issue: if one of RMQ HA Cluster nodes fails we got lost of error "unknown delivery tag" errors. And due to using transaction for  both basic_get and basic_ack this error could lead to very long performance degradation on processors. This is because every time "unknown delivery tag"  happen we close connection and all crashes we are processing now will lead to same error.


Lehych:
@willkg what do you think about this?

willkg:
Sorry--I haven't gotten to look at this, yet. I'll try to do it in the next couple of weeks.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -151,7 +151,7 @@ def _basic_get_transaction(self, conn, queue):
         """reorganize the the call to rabbitmq basic_get so that it can be
         used by the transaction retry wrapper."""
         things = conn.channel.basic_get(queue=queue)
-        return things
+        return conn.channel, things
```
willkg: I think it'd be better to unwrap `things` and return a non-nested datastructure. That'd improve readability of this code. I think I would put it all in a dict like this:

```
things = conn.channel.basic_get(queue=queue)
return {
    'channel': conn.channel,
    'method_frame': things[0],
    'header_frame': things[1],
    'body': things[2]
}
```

That affects the code below, too.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -208,11 +210,20 @@ def _suppress_duplicate_jobs(self, crash_id, acknowledgement_token):
...
+                    crash_id,
+                    acknowledgement_token
+                )
+            except Exception:
+                self.config.logger.error(
+                    'RabbitMQCrashStorage tried to acknowledge duplicate crash %s'
+                    ', but failed.',
+                    crash_id,
+                    exc_info=True
+                )
```
willkg: I don't understand the why you took this out of the transaction. What're you trying to fix here?

Lehych:
Sory, I was not aware of new github review rules so my comment was not shown to anyone: I've not pressed Finish review button =(

PR #4276, opened 2018-01-05
Downgrade Commonmark to 0.5.4

willkg:
This fixes building docs on readthedocs.


willkg:
This fixes rtd and doesn't affect anything else, so I'm just going to land it.

PR #4277, opened 2018-01-08
fixes bug 1375183 - middleware removal

willkg:
This removes the middleware service. We no longer use it and @milescrabill has removed the infrastructure bits so we can remove the rest of the bits from this repo.

Note that we also have Django middleware, but this is unrelated to the middleware service.


willkg:
Thank you!

PR #4278, opened 2018-01-08
Expose postgres on a port to make connecting via a client easier.

Osmose:
This makes it easier to use programs like Postico to inspect the database during development.

### DIFF
--- a/docs/howto.rst
+++ b/docs/howto.rst
@@ -12,6 +12,18 @@ You can run the crontabber job that checks for security vulnerabilities locally:
    make dockerdependencycheck
 
 
+Connect to PostgreSQL Database
+==============================
+
+The local development environment's PostgreSQL database exposes itself on a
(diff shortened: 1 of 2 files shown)


docs/howto.rst
```diff
@@ -12,6 +12,18 @@ You can run the crontabber job that checks for security vulnerabilities locally:
    make dockerdependencycheck
 
 
+Connect to PostgreSQL Database
+==============================
+
+The docker container running the PostgreSQL database exposes itself on a
```
willkg: Can you clarify that this is explicitly talking about the local development environment and not some docker container running elsewhere? In the "run security checks for dependencies", we use the word "locally". Something like that here would be super.

Osmose: Updated this to "The local development environment's PostgreSQL database exposes itself on a". Is that better?

PR #4281, opened 2018-01-09
fixes bug 1338190 - add CrashReporter::TerminateHandler to prefix list

willkg:


willkg:
Here's signature generation output:

```
app@processor:/app$ python -m socorro.signature 65f06168-fb40-4ed9-9346-ae7720180109 441d474a-7cfb-4f30-856c-65a9f0180108
Crash id: 65f06168-fb40-4ed9-9346-ae7720180109
Original: CrashReporter::TerminateHandler
New:      CrashReporter::TerminateHandler | std::__terminate
Same?:    False

Crash id: 441d474a-7cfb-4f30-856c-65a9f0180108
Original: CrashReporter::TerminateHandler
New:      CrashReporter::TerminateHandler | libc++abi.dylib@0x23d68
Same?:    False
```

Osmose:
To clarify, I meant that the bug seems like it's suggesting this is a signature to ignore instead of include in the signature. I noun'd wrong.

willkg:
I generally add to the prefix list unless someone explicitly says "this has no value and should never show up in a signature". The prefix list is fine except in cases where the signature is so long it gets truncated. None of the signatures I looked at for this bug were in danger of truncation.

We can always move it to the irrelevant list later if there's a compelling reason.

willkg:
Thank you!

PR #4283, opened 2018-01-09
fixes bug 1155165 - truncate in one place

willkg:
This fixes signature generation rules so that we have just one rule that does signature truncation and it runs at the end of the signature generation pipeline. Yay!


socorro/unittest/signature/test_rules.py
```diff
@@ -278,26 +279,20 @@ def test_generate_signature_3(self):
     def test_generate_signature_4(self):
         j = JavaSignatureTool()
         java_stack_trace = (
-            '   SomeJavaException: %s  \n'
-            'at org.mozilla.lars.myInvention(larsFile.java)' % ('t' * 1000)
+            '   SomeJavaException: t  \nat org.mozilla.lars.myInvention(larsFile.java)'
```
willkg: I messed up--I need to re-fix this because it should be testing that the description gets dropped if it's too long.

socorro/unittest/signature/test_rules.py
```diff
@@ -278,26 +279,20 @@ def test_generate_signature_3(self):
...
-        assert notes == e
+        expected = 'SomeJavaException: t at org.mozilla.lars.myInvention(larsFile.java)'
+        assert sig == expected
 
     def test_generate_signature_4_2(self):
         j = JavaSignatureTool()
         java_stack_trace = (
-            '   SomeJavaException: %s  \n'
-            'at org.mozilla.lars.myInvention(larsFile.java:1234)' % ('t' * 1000)
+            '   SomeJavaException: t  \nat org.mozilla.lars.myInvention(larsFile.java:1234)'
```
willkg: I messed up--I need to re-fix this because it should be testing that the description gets dropped if it's too long.

socorro/unittest/signature/test_rules.py
```diff
@@ -972,9 +965,7 @@ def test_action_1(self):
 
         raw_crash = {
             'JavaStackTrace': (
-                '   SomeJavaException: %s  \n'
-                'at org.mozilla.lars.myInvention('
-                'larsFile.java)' % ('t' * 1000)
+                '   SomeJavaException: t  \nat org.mozilla.lars.myInvention(larsFile.java)'
```
willkg: I messed up--I need to re-fix this because it should be testing that the description gets dropped if it's too long.

willkg:
This isn't ready for review, yet--I need to fix tests I inadvertently de-fanged.

willkg:
^^^ That fixes the thing I borked. All better now.

socorro/signature/rules.py
```diff
@@ -569,13 +574,16 @@ def action(self, raw_crash, processed_crash, notes):
 
 
 class SigTrunc(Rule):
-    """ensure that the signature is never longer than 255 characters"""
+    """Truncates signatures down to SIGNATURE_MAX_LENGTH characters
```
Osmose: The original bug claims that signature length is and should be configurable, but it's not; it's an in-code constant set to 255. Is this intentional?

Also this claims to use `SIGNATURE_MAX_LENGTH` but uses a hard-coded length instead.

willkg: It is intentional--that's something that changed when I redid signature generation over the summer and took out all the configuration. We've always had it as 255 and never changed it. We'd want any changes to be deployable and backed-outable. For something like this, seems better to set a constant in code.

However, it should use `SIGNATURE_MAX_LENGTH`--that's a mistake. I'll fix that.

willkg:
^^^ That fixes `SigTrunc` to use `SIGNATURE_MAX_LENGTH` and rebases against master.

willkg:
Thank you!

PR #4285, opened 2018-01-10
bug 1429437 - first pass at login troubleshooting docs

willkg:
This is a first pass at login troubleshooting docs. I figure any information is better than no information, so any steps forward--even if they're not complete and comprehensive and perfect--are better than where we are now.

This creates a section that has a url that we can point people to, so that satisfies my need for a one-liner response for login problems to catch common scenarios.

What can we improve in the actual steps? Are there other good troubleshooting steps that users can try?

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -41,10 +41,11 @@ Contents
 
    gettingstarted
    contributing
-   symbols
+   troubleshooting
    architecture/*
    components/*
    services/*
    tests/*
+   symbols
    deploy
    howto
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -1,0 +1,26 @@
+.. index:: troubleshooting
+
+===============
+Troubleshooting
+===============
+
+Login problems
+==============
+
+Socorro uses Google to authenticate users.
+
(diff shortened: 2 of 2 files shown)


willkg:
@Osmose @lonnen @peterbe @adngdb: Thoughts on this?

adngdb:
@willkg how is that different / better than https://wiki.mozilla.org/Socorro/GoogleAuth ?

willkg:
@adngdb It's not! I had no idea that page existed and when I was asking for help with login troubleshooting a few months back, I was told there was no troubleshooting documentation for login problems.

I'll nix this in favor of that. Thanks!

peterbe:
Didn't you write that Wiki page originally? :)

willkg:
@peterbe Well, past me should have put it in my bookmarks. Damn you, past me!

peterbe:
It's not my call but now that socorro is all about how mozilla is using it, perhaps it's best to keep it with the rest of the docs. I.e. socorro.readthedocs.org

willkg:
That makes sense. I added a comment to bug 1373997 about it.

PR #4286, opened 2018-01-10
fixes bug 1426186 - add reprocessing howto

willkg:
This adds some crude documentation for reprocessing crashes including some
hand-wavey warnings about being cautious about adding crash ids to the queue
faster than they can be processed.

### DIFF
--- a/docs/howto.rst
+++ b/docs/howto.rst
@@ -24,6 +24,53 @@ the client of your choice using the following connection settings:
 * Port: ``8574``
 
 
+Reprocess crashes
+=================
+
+Reprocessing individual crashes
+-------------------------------
+
+If you have appropriate permissions, you can reprocess an individual crash by
+viewing the crash report on the Crash Stats site, clicking on the "Reprocess"
+tab, and clicking on the "Reprocess this crash" button.
+
+
+Reprocessing lots of crashes if you're not an admin
+---------------------------------------------------
+
+If you need to reprocess a lot of crashes, please `write up a bug
+<https://bugzilla.mozilla.org/enter_bug.cgi?format=__standard__&product=Socorro>`_.
+In the bug description, include a Super Search url with the crashes you want
+reprocessed.
+
+
+Reprocessing crashes if you're an admin
+---------------------------------------
+
+If you're an admin, you can create an API token with the "Reprocess Crashes"
+permission. You can use this token in conjunction with the
+``scripts/reprocess.py`` script to set crashes up for reprocessing.
+
+For example, this reprocesses a single crash::
+
+    $ docker-compose run processor bash
+    app@processor:app$ ./scripts/reprocess.py c2815fd1-e87b-45e9-9630-765060180110
+
+This reprocesses crashes all crashes with a specified signature::
+
+    $ docker-compose run processor bash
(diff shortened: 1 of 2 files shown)


docs/howto.rst
```diff
@@ -12,6 +12,53 @@ You can run the crontabber job that checks for security vulnerabilities locally:
...
+This reprocesses crashes all crashes with a specified signature::
+
+    $ docker-compose run processor bash
+    app@processor:app$ ./scripts/fetch_crashids.py --signature="some | signature" | ./scripts/reprocess.py
+
+
+.. Warning::
+
+   If you're reprocessing more than 10,000 crashes, make sure to specify the
+   ``--sleep`` argument and use a value like 10 seconds. This will slow down
```
Osmose: It'd be helpful if this had ``--sleep 10`` and was reworded to make it clear what units the parameter is in. Otherwise I'd just be looking that up anyway since I don't know how to use that argument.

socorro/scripts/reprocess.py
```diff
@@ -21,7 +21,13 @@
...
 Sends specified crashes for reprocessing
 
-This requires SOCORRO_REPROCESS_API_TOKEN to be set in the environment to a valid API token.
+This requires SOCORRO_REPROCESS_API_TOKEN to be set in the environment to a
+valid API token.
+
+Note: If you're reprocessing more than 10,000 crashes, you should use a sleep
+value of 10 seconds. This slows down adding items to the reprocessing queue so
+they're being added roughly at the same rate as they're being processed. This
+way, the queue doesn't back up and the system doesn't start paging people.
```
Osmose: Maybe as a nice-to-have in a future PR we can avoid having to warn people about this and just do it automatically if `--sleep` isn't manually specified and there's more than 10,000 crashes?

willkg: Yeah... so "10" is actually a fudge-factor and it's probably good enough for some/most cases, so I documented that. However, the last time I reprocessed 100,000 things, I had Miles throw in some more processors, so then the rate they're processing increased significantly and I think I was using 6 seconds as my sleep value.

I didn't want to write an essay about the calculus of appropriate sleep times, but did want to at least raise the issue and provide a "this is probably good enough for some/most cases" value.

The new infrastructure has processors that autoscale with load, so we'll want to remove this note in the future.

docs/howto.rst
willkg: That makes a lot of sense. I'll fix that.

socorro/scripts/reprocess.py
willkg: Also, this is a note for you and me right now and if other developers come along, them. What language would best help us? Is something like this better?:

```
Note: If you're processing more than 10,000 crashes, you should use a sleep value
that balances the rate of crash ids being added to the queue and the rate of
crash ids being processed. Also, if you're processing a lot of crashes, you should
tell ops to increase the number of processor nodes.
```

Osmose: I like that new version except that it removed the example value. I think you're right in that adding a "good enough" value is useful, and helps orient someone unfamiliar to what scale they should use (like, don't use increments of 100+ seconds). 

Adding back the example of 10 seconds as good enough for most use cases would make this perfect.

willkg:
^^^ That updates the note about crash id production consumption.

willkg:
Thank you! This will help a lot!

PR #4290, opened 2018-01-11
Fix bug 1399305: Script deletion and creation of S3 bucket and ES indices in development environment.

Osmose:
This does 3 things:

1. Adds `docker/run_recreate_s3_buckets.sh`, which runs during `make dockersetup` and deletes and re-creates the `dev_bucket` S3 bucket used during development.
2. Adds a new Socorro app called `clear_es_indices`, run during `make dockersetup`, that clears all existing Elasticsearch indexes.
3. Adds a new Socorro app called `create_recent_es_indices`, run during `make dockerupdatedata`, that creates Elasticsearch indices for the next two weeks. This fixes the issues with the homepage and other views that pop up during local development when they're missing.

We don't have great support for testing changes to Elasticsearch, so I filed [bug 1429267](https://bugzilla.mozilla.org/show_bug.cgi?id=1429267) as a followup to add better scaffolding such that we could add automated tests for scripts like these.

To test this:
1. Start with a Socorro installation that has processed some crashes.
2. Run `make dockersetup`. You should see logging output that looks something like this:

   ```
   Dropping and recreating S3 bucket...
   S3 container bucket is dev_bucket
   delete: s3://dev_bucket/v1/dump/01557534-e7c7-4936-970d-86f550180107
   (...snip...)
   ```

   You should also see a log line like this, indicating that the existing ES indices have been deleted:

   ```
   2018-01-10 01:22:23,085 DEBUG - clear_es_indices -  - MainThread - Deleted elasticsearch indices.
   ```
3. Run `make dockerupdatedata`. You should see logging output indicating that new ES indices have been created:

   ```
   2018-01-11 21:13:01,754 INFO - create_recent_es_indices -  - MainThread - Created new elasticsearch index: socorro201801
   2018-01-11 21:13:01,812 INFO - create_recent_es_indices -  - MainThread - Created new elasticsearch index: socorro201802
   2018-01-11 21:13:01,893 INFO - create_recent_es_indices -  - MainThread - Created new elasticsearch index: socorro201803
   ```
4. Run `docker-compose up webapp` and visit http://localhost:8000/home/product/Firefox. You should see a graph on the front page, although it may contain no data (you should not see an error message, which indicates the indices were not found).


socorro/external/es/index_cleaner.py
```diff
@@ -37,11 +37,17 @@ def __init__(self, config):
         super(IndexCleaner, self).__init__()
         self.config = config
 
-    def delete_old_indices(self):
```
Osmose: This PR ended up taking longer than I expected because the original plan of using this method with a `retention_policy` of 0 meant that indices for future weeks (such as those created by the scripts in this PR) would not actually be deleted. I figured refactoring this class while retaining it's existing API was the cleanest fix.

Makefile
```diff
@@ -36,6 +36,8 @@ dockerbuild: my.env
 # postgres going forward and has the needed environment variables.
 dockersetup: my.env .docker-build
 	${DC} run webapp /app/docker/run_setup_postgres.sh
+	${DC} run processor /app/docker/run_recreate_s3_buckets.sh
+	${DC} run processor /app/scripts/socorro clear_es_indices
```
willkg: The reason that the first item runs in the webapp container is that we were thinking that the webapp would ultimately own all the data stores. That's a vision change related to the processor rewrite.

Figured I'd mention it in case that's interesting to know.

The other thought I had here is that it's "faster" if all three of these scripts are executed in a single `docker-compose run` rather than three separate ones. It's probably a second or two in savings, so it's probably not a big deal. If there are other niceties to running all three of these scripts as a single script, then maybe it's worth changing.

socorro/external/es/index_cleaner.py
```diff
@@ -68,13 +74,23 @@ def delete_old_indices(self):
...
+        policy_delay = datetime.timedelta(weeks=self.config.retention_policy)
+        time_limit = (now - policy_delay).replace(tzinfo=None)
+
+        # This won't take the week part of our indices into account...
+        index_date = datetime.datetime.strptime(
+            index,
+            self.config.elasticsearch.elasticsearch_index
+        )
+        # So we need to get that differently, and then add it to the date.
+        index_date += datetime.timedelta(weeks=int(index[-2:]))
```
willkg: It took me a while to figure out what this is about until I tried it all by hand. Ugh. I wonder why it doesn't do the week part.

The last part of this line assumes that the last two characters are the week, but that's configured by `elasticsearch_index`. I'm pretty sure we use some variation of the default everywhere, but I'm not positive. I'm not sure how to do it better. I think I'd give up and add a note along these lines:

```
NOTE(osmose): This assumes that the last two characters of the index are the week.
```

or something like that.

socorro/unittest/external/es/test_index_cleaner.py
```diff
@@ -97,17 +93,43 @@ def test_delete_old_indices(self):
...
-
-        assert self.index_client.exists('socorro_test_temp')
 
         api = IndexCleaner(self.config)
         api.delete_old_indices()
 
         # Verify the email index is still there.
         assert self.index_client.exists('socorro_test_temp')
+
+    @minimum_es_version('1.0')
```
willkg: We never run with ES < 1.0. We can nix these decorators and any others like this.

socorro/external/es/index_cleaner.py
```diff
@@ -37,11 +37,17 @@ def __init__(self, config):
...
-        time_limit = (now - policy_delay).replace(tzinfo=None)
+    def delete_indices(self, predicate=None):
+        """Delete crash indices that match the given predicate.
 
+        :param predicate:
+            A callable of the form ``predicate(index)``, where ``index``
+            is a string containing the name of the index. If the
+            callable returns true, the index will be deleted.
+
+            The default is None, which deletes all crash indices.
```
willkg: We've been using `:arg ARG: some docs here` format and recently, I've been doing `:arg TYPE ARG: some docs here` in cases where the type information is helpful.

* https://github.com/mozilla-services/socorro/blob/master/socorro/signature/generator.py#L85
* https://github.com/mozilla-services/socorro/blob/master/socorro/lib/treelib.py#L163

It's not a big deal, but if you're a stickler for style uniformity, then that's what we've been doing.

socorro/external/es/index_cleaner.py
```diff
@@ -68,13 +74,23 @@ def delete_old_indices(self):
...
 
-            # This won't take the week part of our indices into account...
-            index_date = datetime.datetime.strptime(
-                index,
-                self.config.elasticsearch.elasticsearch_index
-            )
-            # So we need to get that differently, and then add it to the date.
-            index_date += datetime.timedelta(weeks=int(index[-2:]))
+            if predicate is None or predicate(index):
+                index_client.delete(index)
```
willkg: In the case where we've run this from the command line (as opposed to crontabber), I'd really like it to tell me what it deleted. One way to do that is pass in a `verbose` flag and then have it spit it out to the logger. Another way is have this method return the list of indexes it deleted and then the caller can print them out.

socorro/external/es/create_recent_indices_app.py
```diff
@@ -0,0 +1,49 @@
...
+        index_creator = self.config.index_creator_class(self.config)
+        index_name_template = self.config.elasticsearch.elasticsearch_index
+
+        today = date.today()
+        current_monday = today - timedelta(days=today.weekday())
+
+        for week_diff in range(-1, self.config.elasticsearch_weeks_to_create):
+            week_monday = current_monday + timedelta(weeks=week_diff)
+            index = week_monday.strftime(index_name_template)
+            index_creator.create_socorro_index(index)
```
willkg: I'd like it if this told me what it created. Either using `print()` or the logger is fine.

willkg:
I added some comments, but haven't tried it out, yet. I'll go through the rest of this tomorrow.

I'm psyched to land this! It solves a bunch of problems I have with my local dev environment.

docker/run_update_data.sh
```diff
@@ -35,3 +35,6 @@ docker-compose run processor python docker/fetch_normalization_data.py --product
 
 # Create weekly reports
 docker-compose run processor python docker/create_weekly_tables.py
+
+# Create ES indexes for the next 2 weeks
+docker-compose run processor scripts/socorro create_recent_es_indices
```
willkg: I ran this today (Friday, January 12th, week 2), and it created the following indexes: socorro201801, socorro201802, and socorro201803. That seems right, but I think it doesn't jive with the comment here.

I think we could go vague (it's explained elsewhere) and say "Create ES indexes" or go specific and say something like "Create last week's index, this week's index, and next week's index". I think I'd go vague because then we have fewer places to update if we ever change the amount.

socorro/app/for_application_defaults.py
```diff
@@ -48,7 +48,10 @@ def find_all_the_apps():
             'reprocess_crashlist':
                 'socorro.external.rabbitmq.reprocess_crashlist.ReprocessCrashlistApp',
             'purge_rmq':
-            'socorro.external.rabbitmq.purge_queue_app.PurgeRabbitMQQueueApp',
+                'socorro.external.rabbitmq.purge_queue_app.PurgeRabbitMQQueueApp',
+            'clear_es_indices': 'socorro.external.es.clear_indices_app.ClearESIndicesApp',
+            'create_recent_es_indices':
+                'socorro.external.es.create_recent_indices_app.CreateRecentESIndicesApp',
```
willkg: At some point, I think it probably makes more sense to build an Elasticsearch equivalent to the Postgres `setupdb_app` which does all the manipulation rather than have it spread across a few different apps. I wouldn't change anything in this PR, though.

socorro/external/es/create_recent_indices_app.py
willkg: Scratch that--something is printing out what indexes are getting created.

socorro/unittest/external/es/test_index_cleaner.py
willkg: We can nix the decorators in another PR--doesn't have to be this one. One of us is probably going to be doing some ES work in the near future, so we can change things then.

socorro/external/es/index_cleaner.py
Osmose: Stackoverflow pointed me to the [docs](https://docs.python.org/2/library/datetime.html):

> When used with the strptime() method, %U and %W are only used in calculations when the day of the week and the year are specified.

willkg: Ahh... ok. That makes sense.

socorro/unittest/external/es/test_index_cleaner.py
Osmose: I can at least remove them from this test file.

socorro/external/es/index_cleaner.py
```diff
@@ -85,12 +90,11 @@ def is_index_old(self, index):
...
         time_limit = (now - policy_delay).replace(tzinfo=None)
 
-        # This won't take the week part of our indices into account...
+        # strptime ignores week numbers if a day isn't specified, so we append
+        # '-1' and '-%w' to specify Monday as the day.
         index_date = datetime.datetime.strptime(
-            index,
-            self.config.elasticsearch.elasticsearch_index
+            index + '-1',
+            self.config.elasticsearch.elasticsearch_index + '-%w'
```
willkg: Ooo--clever!

PR #4292, opened 2018-01-12
fixes bug 1424659 - nix consecutive spaces in signatures

willkg:
Prior to this, it was possible (though rare) to have two spaces in a row
in signatures. That's tough because the webui shows that as a single
space and if you search for the single-space version, you won't get any
results.

This fixes that by establishing that signatures can't have more than 1
consecutive whitespace character. Further, this makes sure that all
non-space whitespace characters are converted to space and this
happens all before we truncate the signature.

I also changed SigTrunc to SigTruncate because "trunc" is an unnecessary
abbreviation.

### DIFF
--- a/socorro/unittest/signature/test_rules.py
+++ b/socorro/unittest/signature/test_rules.py
@@ -21,8 +21,8 @@ from socorro.signature.rules import (
     SignatureGenerationRule,
     SignatureJitCategory,
     SignatureRunWatchDog,
-    SigTrim,
-    SigTrunc,
+    SigFixWhitespace,
+    SigTruncate,
     SignatureShutdownTimeout,
     SignatureIPCMessageName,
     SignatureParentIDNotEqualsChildID,
@@ -1277,10 +1277,10 @@ class TestAbortSignature:
         assert processed_crash['signature'] == 'Abort | unknown | hello'
 
 
-class TestSigTrim:
+class TestSigFixWhitespace:
 
     def test_predicate_no_match(self):
-        rule = SigTrim()
+        rule = SigFixWhitespace()
 
         processed_crash = {}
         predicate_result = rule.predicate({}, processed_crash)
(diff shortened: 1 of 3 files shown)


willkg:
This fixes some signatures that have two spaces in them. There aren't many--it's pretty rare. But it does happen.

Example of signature generation output:

```
you@processor:/app$ ./scripts/fetch_crashids.py --signature="  " | python -m socorro.signature
Crash id: bf83de41-13e0-4b04-bc17-c989d0180111
Original: mozilla::ipc::MessageChannel::Send  | mozilla::net::PNeckoChild::SendRemoveRequestContext
New:      mozilla::ipc::MessageChannel::Send | mozilla::net::PNeckoChild::SendRemoveRequestContext
Same?:    False

Crash id: e3eada6a-117e-476f-aa4e-3125e0180111
Original: mozilla::detail::MutexImpl::lock  | mozilla::dom::EventSourceImpl::ReadyState
New:      mozilla::detail::MutexImpl::lock | mozilla::dom::EventSourceImpl::ReadyState
Same?:    False
```

socorro/unittest/signature/test_rules.py
```diff
@@ -1291,20 +1291,46 @@ def test_predicate_no_match(self):
...
+        }
+        action_result = rule.action({}, processed_crash, [])
+        assert action_result is True
+        assert processed_crash['signature'] == expected
+
+    @pytest.mark.parametrize('signature, expected', [
+        ('all   good', 'all good'),
+        ('all  |  good', 'all | good'),
+    ])
+    def test_consecutive_spaces(self, signature, expected):
```
Osmose: Is there any reason not to collapse all these test cases into a single one with all the parameterizations? You could use comments and whitespace to annotate what the cases are covering.

willkg: That makes a lot of sense. I'll fix that now.

willkg:
Thank you!

PR #4293, opened 2018-01-12
[wip] fixes bug 1425245 - adds crash_verify/ url endpoint

willkg:
This url endpoint takes a crash id and tells you whether the data for that
crash exists in all the appropriate crash stores.

We can use this for end-to-end integration testing as well as debugging
system oddities.

### DIFF
--- a/socorro/external/boto/crash_data.py
+++ b/socorro/external/boto/crash_data.py
@@ -6,6 +6,7 @@ from socorro.lib import external_common, MissingArgumentError, BadArgumentError
 from socorro.external.boto.crashstorage import (
     BotoS3CrashStorage,
     CrashIDNotFound,
+    TelemetryBotoS3CrashStorage,
 )
 from socorro.external.crash_data_base import CrashDataBase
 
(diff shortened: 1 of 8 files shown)


willkg:
This is a rough version of the "crash_verify" view. I had to write a lot of middley stuff to connect the webapp view with everything else. You go to a url like `http://localhost:8000/crash_verify/SOMECRASHID` and it produces a JSON blob like this:

```
{
    's3_raw_crash': true,
    's3_processed_crash': true,
    's3_telemetry_crash': true,
    'elasticsearch_crash': true,
    'postgres_raw_crash': true,
    'postgres_processed_crash': true
}
```

One thing I don't like about this is how it parallels the processor destinations, but doesn't use the processor destination configuration to figure out which crash storage systems to check. I couldn't figure out how to do that. Plus the processor destination configuration is in the processor environment and not in the webapp environment.

One thing this is missing is tests, but "it works for me!" I figure before I slave over tests, I'd love it if someone (@lonnen, @Osmose) could skim it. Is this a sane approach? Do we want something different?

Should we abandon this and instead use the existing API endpoints to determine if things are where they need to be? We can ignore the Postgres crash store because we want it to go away soon. We would need to add an API endpoint for pulling the data we're saving to the telemetry S3 bucket.

What do you think? Especially in regards to the "less is more--don't write more code" angle.

willkg:
I talked with @Osmose about this a bit. There are some fragile bits in this PR that are unenthusing. It would benefit greatly from the webapp knowing which crash storage systems were valid destinations--something that's specified in processor configuration.

We wrote up bug 1430860 (https://bugzilla.mozilla.org/show_bug.cgi?id=1430860) to cover reworking PolyCrashStorage such that we're codifying destinations in code--not in configuration. That'll make this easier.

The other thing that would make this easier is sorting out crashstorage vs. crash_data. Bug 1299456 (https://bugzilla.mozilla.org/show_bug.cgi?id=1299456) covers deleting crash data classes. Maybe working on that will cause us to merge crash data classes with crashstorage classes and thus further simplify this bug.

Anyhow, in the interim, I'm going to close this out.

PR #4294, opened 2018-01-16
 Bug 1430181: Add an API endpoint (CrashStopData) to Socorro. 

calixteman:
The purpose of this endpoint is to have in one shot all the data needed by the webextension CrashStop which displays crash data in Bugzilla.


willkg:
We talked briefly about this on IRC. I do recognize the importance of supporting the crashstop extension, but this isn't going to fly in its current state.

I'd be interested in sussing out what kinds of data crashstop needs and then figure out what generalized API endpoints we'd need to build to support that. That seems like a more viable direction than this one. That is best discussed in the related bug.

In the meantime, I'm going to close this out. Sorry about that!

PR #4295, opened 2018-01-16
Make the string representation of PolyStorageError more robust.

Osmose:
While working on [bug 687630](https://bugzilla.mozilla.org/show_bug.cgi?id=687630) I ran into an issue while processing crashes, and the error couldn't be output properly because `PolyStorageError` had been called with no arguments, which `PolyStorageError.__str__` could not properly handle.

This makes the method more robust to weird arguments so that it will print out the errors that occurred even if there are no arguments given. It also slightly changes the format of what was output, but none of the tests were relying on this anyway.

### DIFF
--- a/socorro/external/crashstorage_base.py
+++ b/socorro/external/crashstorage_base.py
@@ -459,9 +459,12 @@ class PolyStorageError(Exception, collections.MutableSequence):
         self.exceptions.__setitem__(index, value)
 
     def __str__(self):
-        return '%s %s' % (self.args[0],
-                          ','.join(repr(e[1]).encode('ascii', 'ignore')
-                                   for e in self.exceptions))
+        output = []
+        if self.args:
+            output.append(self.args[0])
+        for e in self.exceptions:
+            output.append(repr(e[1]).encode('ascii', 'ignore'))
+        return ','.join(output)
 
 
 class PolyCrashStorage(CrashStorageBase):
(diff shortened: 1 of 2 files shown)


socorro/external/crashstorage_base.py
```diff
@@ -459,9 +459,12 @@ def __setitem__(self, index, value):
...
     def __str__(self):
-        return '%s %s' % (self.args[0],
-                          ','.join(repr(e[1]).encode('ascii', 'ignore')
-                                   for e in self.exceptions))
+        output = []
+        if self.args:
+            output.append(self.args[0])
+        for e in self.exceptions:
+            output.append(repr(e[1]).encode('ascii', 'ignore'))
+        return ','.join(output)
```
willkg: This doesn't produce the exact same output as before. Before it'd do the first arg, then a space, then a comma-separated string of exception reprs.

I don't know if that matters at all. In the debugging you were doing that led to bumping into this issue, does it help to differentiate between the arg and the wrapped exceptions?

Osmose: Not really. But I was debugging an issue where no args were being passed so it's kind of irrelevant anyway. 

The main reason I changed it was to simplify the implementation. Given what PolyStorageError is used for, I think this doesn't affect the helpfulness one way or another; what would really be helpful would be labeling which storage each error was raised from, but that'd be a bit more of a refactoring.

PR #4297, opened 2018-01-18
Fix bug 1431162: Fix errors in crontabber_status.js due to D3/sankey updates.

Osmose:
- Fix some inconsistent indenting that was failing eslint.
- The `d.x`, `d.y`, and `d.dy` attributes on sankey datums appears to have been
  replaced with `d.x0`, `d.x1`, `d.y0`, and `d.y1`. So we replace all old
  attributes with the new ones (dy is computed from y0 and y1).
- Pass nodes and links via the sankey constructor.
- Replace `sankey.nodeWidth` call with a property computed from the datum's `x0`
  and `x1` values.
- Replace `sankey.update` with `sankey.relayout`.
- Replace `sankey.link` with `d3.sankeyLinkHorizontal`.

Most of these changes were found in the release notes on the sankey Github:
https://github.com/d3/d3-sankey/releases.

### DIFF
--- a/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/crontabber_state.js
+++ b/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/crontabber_state.js
@@ -45,19 +45,22 @@ var format = d3.format(',.0f'),
         return name + ' is working normally.';
     };
 
-var svg = d3.select('#crontabber-chart').append('svg')
+var svg = d3.select('#crontabber-chart')
+    .append('svg')
     .attr('width', width + margin.left + margin.right)
     .attr('height', height + margin.top + margin.bottom)
-  .append('g')
-    .attr('transform',
-          'translate(' + margin.left + ',' + margin.top + ')');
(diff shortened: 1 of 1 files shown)


Osmose:
I'm only kind of confident in my local testing for this, given that my crontabber state is different from stage's. 

If I'm honest, I'd question how useful the graph is vs just the table at the bottom of the page. The graph code seems kind of brittle, although it is going through a major version change for both d3 and d3-sankey (and d3 4.0 was a significant API redesign). Maybe something to consider for the redesign.

willkg:
Regarding "why do we even have this graph?", I think it's handy to see why certain dependency chains are busted. I've definitely used that in the past. While the table has all the information in it, it surfaces that poorly.

There's been a lot of work over the last two years to remove crontabber apps and reduce dependency chains. I think the only important one we have now is the ADI one. Once that goes away, then I'd be +1 on nixing the graph. We'd probably want to run that past @lonnen in case there are reasons to keep it that I'm not thinking of.

PR #4298, opened 2018-01-18
bug 1431442 - add "mark-success" to crontabber app

willkg:
This makes it easier to set up initial crontabber bookkeeping state in
the local development environment without having to run all the crontabber
jobs.

Mike: Do you think this helps? Is it good enough for a first pass? We can also do one for marking jobs as failed. I think that covers basic cases and after that it's probably easier to drop into psql and edit the db manually.

### DIFF
--- a/docker/run_setup.sh
+++ b/docker/run_setup.sh
@@ -9,6 +9,15 @@ set -e
 # Run scripts for setting up data sources for the local development environment.
 # This assumes that it is running in the webapp docker container.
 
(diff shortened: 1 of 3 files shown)


willkg:
Oo--that's a good idea. I'll do that and then make sure it doesn't mess up the first pass of ftpscraper and friends in `dockerupdatedata`.

willkg:
Thank you!

PR #4299, opened 2018-01-18
fixes bug 1416340 - use secure args for ES connections

willkg:
If the `elasticsearch_urls` are all https, then we should be passing in
`verify_certs=True` and `use_ssl=True` args into the Elasticsearch
connection.

### DIFF
--- a/socorro/external/es/connection_context.py
+++ b/socorro/external/es/connection_context.py
@@ -10,11 +10,12 @@ from configman.converters import list_converter
 
 
 class Connection(object):
-    """A facade in front of the ES class that standardises certain gross
-    elements of its API with those of other database connection types.
-    The Elasticsearch interface is a simple HTTP API, and as such, does not
-    maintain open connections (i.e. no persistence). Accordingly, the commit,
-    rollback, and close mechanisms cannot have any useful meaning.
+    """A facade in front of the ES class that standardises certain gross elements
+    of its API with those of other database connection types. The Elasticsearch
+    interface is a simple HTTP API, and as such, does not maintain open
+    connections (i.e. no persistence). Accordingly, the commit, rollback, and
+    close mechanisms cannot have any useful meaning.
+
     """
 
     def __init__(self, config, connection):
@@ -40,7 +41,7 @@ class ConnectionContext(RequiredConfig):
     required_config = Namespace()
     required_config.add_option(
         'elasticsearch_urls',
-        default=['localhost:9200'],
+        default=['http://localhost:9200'],
         doc='the urls to the elasticsearch instances',
         from_string_converter=list_converter,
         reference_value_from='resource.elasticsearch',
@@ -93,7 +94,9 @@ class ConnectionContext(RequiredConfig):
     def connection(self, name=None, timeout=None):
         """Returns an instance of elasticsearch-py's Elasticsearch class as
         encapsulated by the Connection class above.
+
         Documentation: http://elasticsearch-py.readthedocs.org
+
         """
         if timeout is None:
             timeout = self.config.elasticsearch_timeout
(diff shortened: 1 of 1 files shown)


willkg:
Our current -stage and -prod environments don't use https, but the new infra does.

It's possible for `elasticsearch_urls` to be a comma separated list of connection urls. However, the way we use it in all our environments is as a single url. I figured it's not super hard to handle the multiple urls case now. That's why it looks more complex than it otherwise might have.

It's keying off of "http" and "https" in the url. Previously, we allowed for protocol-less hosts. This changes that--now we require a protocol.

This is missing a test that makes sure the additional arguments are used if the protocol is https as well as a test for the case where an http url is mixed with an https url. I'll do those next.

willkg:
^^^ That adds the tests.

I don't have a good way to verify this outside of the unit tests. When this lands, I'll have to verify it's working on -stage and -stage-new and doesn't kick up errors in -stage-new anymore in crontabber.

socorro/external/es/connection_context.py
```diff
@@ -93,24 +98,49 @@ def __init__(self, config):
...
         Documentation: http://elasticsearch-py.readthedocs.org
+
         """
         if timeout is None:
             timeout = self.config.elasticsearch_timeout
 
+        # Separate the HTTP from HTTPS urls so we know if they're all one or
+        # the other because we don't want a mix
+        http = [url for url in self.config.elasticsearch_urls if not url.startswith('https')]
+        https = [url for url in self.config.elasticsearch_urls if url.startswith('https')]
```
Osmose: The `httpspf700://` protocol passes this test. I'm not sure if that matters, but it smells a bit like the kind of things that combine to form a security vulnerability.

Also, the check above includes non-http protocols in `http`.

We should probably be parsing out the protocol here instead of relying on `startswith`.

socorro/external/es/connection_context.py
```diff
@@ -93,24 +98,49 @@ def __init__(self, config):
...
+        else:
+            additional_kwargs = {}
+
         return Connection(
             self.config,
             elasticsearch.Elasticsearch(
                 hosts=self.config.elasticsearch_urls,
                 timeout=timeout,
-                connection_class=elasticsearch.connection.RequestsHttpConnection
+                connection_class=elasticsearch.connection.RequestsHttpConnection,
```
Osmose: Does anything in elasticsearch itself verify that the URLs have protocols and they're all either http or https?

socorro/external/es/connection_context.py
willkg: A while back, you could also use thrift, but I'm pretty sure Elasticsearch doesn't support that anymore. I think the elasticsearch-py library only supports http/https through requests and urllib3.

However, I'll switch it to parse the url instead--that's a good idea.

socorro/external/es/connection_context.py
willkg: The library has this code:

https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/connection/http_requests.py#L45

So you pass in a host and a port and it uses `use_ssl` to determine whether to use http or https in the url.

The socorro elasticsearch connection context pass in the value of elasticsearch_urls in to create the Elasticsearch client instance. In the `__init__` for that, it normalizes the hosts:

https://github.com/elastic/elasticsearch-py/blob/cdb04dfa011600facdfdd8d2eacd7dded4404413/elasticsearch/client/__init__.py

That normalization process assumes everything is http unless it has the https protocol in which case it adds `use_ssl=True`. So... now that I'm looking at that, we can probably drop our code for adding `use_ssl`. It's even possible that we could set `verify_certs=True` regardless of http or https.

I'll check that out tomorrow. If that's true, then this PR gets much much shorter.

willkg:
^^^ Reworks the PR so it only adds `verify_certs=True`. It's a lot simpler now. Thank you!

socorro/external/es/connection_context.py
```diff
@@ -103,14 +106,17 @@ def connection(self, name=None, timeout=None):
...
+                verify_certs=True
```
Osmose: A comment noting that this implies `use_ssl=True` would be useful for future readers.

willkg: Mmm... Adding `verify_certs=True` doesn't imply anything. It just means that in the case of https, the elasticsearch-py library will set up requests with `verify_certs`.

Maybe I misunderstand your suggestion?

Osmose: Ah, no, it looks like I misunderstood your explanation. :+1:

willkg:
Given that the final change was much smaller, I rebased against master and squashed the commits into one. I figure that's easier-to-understand git history.

willkg:
Thank you!

PR #4302, opened 2018-01-19
fixes bug 1431763 - add buildid for 59.0b1 to AuroraVersionFixitRule

willkg:


willkg:
There really isn't much that's testable here. It's worth verifying the buildid is correct. It's a build that's in the aurora channel and marked 59.0b1 in buildhub:

https://mozilla-services.github.io/buildhub/?q=59.0b1

ftpscraper sees that build and since the channel and version don't agree (one is aurora and the other is a beta version), it "fixes" the version to 59.0b0. That's what we're seeing for those crashes on Crash Stats.

After this lands and gets pushed, we'll re-process those crashes.

Also, judging by the buildid, it was out on 20180113, so this is 6 days late. :( I'll try to be more on top of it next time around. I'll probably set up a monitor or something.

willkg:
Thank you!

PR #4303, opened 2018-01-22
fixes bug 1407625 - truncate values in Elasticsearch

willkg:
This does three things:

1. clean up a bit of Elasticsearch code in a fundamental way
2. update the documentation because that helped figure out how I needed to do things
3. implement truncating values for fields analyzed with the "keyword" analyzer

I did some analysis on our data before implementing value truncating. That's covered in comment 4 and comment 5 in bug 1407625 (https://bugzilla.mozilla.org/show_bug.cgi?id=1407625). I can expand on that if that's helpful.

The unit tests cover testing the scenarios pretty well. I don't think I'd bother with manual testing.

The one thing the changes don't do is emit any kind of signal that a value was truncated, so we'll never really know. There's no way to add post-processing notes to crash data like we do in the processor rules. We can log something, but logs aren't a great way for normal people to answer questions like, "I'm looking at crash XYZ--were values truncated?". The person would have to ask one of us and we'd have to log into the logs host and grep the logs for the text to figure it out. That's ... rough. However, I don't think we have better options for signals in the Elasticsearch crashstorage.

We could do it as a processor rule, but I'm really hesitant to do that. This relies on `super_search_fields.py` to tell us what's up and that's Elasticsearch centric. If we convert that to a "processed crash schema" (which is probably a good idea to do some day), then we could use that and make a processor rule. Maybe some day in the future.

My data diving suggested we don't hit this situation very often--I wasn't able to find a problematic crash for the days I looked at. We do know it's happened in the past as evidenced by our problems migrating data over the summer. Maybe it happens less often now?

That's everything I can think of here.

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -12,10 +12,9 @@ import elasticsearch
 from configman import Namespace
 from configman.converters import class_converter, list_converter
 
-from socorro.external.crashstorage_base import CrashStorageBase
+from socorro.external.crashstorage_base import CrashStorageBase, Redactor
 from socorro.lib.converters import change_default
 from socorro.lib.datetimeutil import JsonDTEncoder, string_to_datetime
-from socorro.external.crashstorage_base import Redactor
 
 
 class RawCrashRedactor(Redactor):
@@ -51,6 +50,31 @@ def is_valid_key(key):
 
     """
     return bool(VALID_KEY.match(key))
+
+
+def remove_bad_keys(data):
+    """Removes keys from the top-level of the dict that are bad
+
+    Good keys satisfy the following properties:
+
+    * have one or more characters
+    * are composed of a-zA-Z0-9_-
+
+    Anything else is a bad key and needs to be removed.
+
+    Note: This modifies the data dict in-place and only looks at the top level.
+
+    :arg dict data: the data to remove bad keys from
+
+    """
+    if not data:
+        return
+
+    # Copy the list of things we're iterating over because we're mutating
+    # the dict in place.
+    for key in list(data.keys()):
+        if not is_valid_key(key):
+            del data[key]
 
 
 class ESCrashStorage(CrashStorageBase):
@@ -124,11 +148,25 @@ class ESCrashStorage(CrashStorageBase):
 
         return index
 
-    def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
-                               crash_id):
-        """This is the only write mechanism that is actually employed in normal
-        usage.
+    def save_raw_and_processed(self, raw_crash, dumps, processed_crash, crash_id):
+        """Save raw and processed crash data to Elasticsearch
+
+        Note: This is the only write mechanism that is actually employed in
+        normal usage.
+
         """
+
+        # Massage the crash such that the date_processed field is formatted
+        # in the fashion of our established mapping.
+        self.reconstitute_datetimes(processed_crash)
+
+        # Remove bad keys from the raw crash--these keys are essentially
+        # user-provided and can contain junk data
+        remove_bad_keys(raw_crash)
+
+        # Capture crash data size metrics--do this only after we've cleaned up
+        # the crash data
+        self._capture_crash_metrics(raw_crash, processed_crash)
 
         crash_document = {
             'crash_id': crash_id,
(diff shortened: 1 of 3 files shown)


socorro/external/es/crashstorage.py
```diff
@@ -53,6 +57,61 @@ def is_valid_key(key):
...
+
+    """
+    if not data:
+        return
+
+    # Copy the list of things we're iterating over because we're mutating
+    # the dict in place.
+    for key in list(data.keys()):
+        if not is_valid_key(key):
+            del data[key]
```
willkg: This function didn't need to be part of the crash storage class, so I moved it to make it easier to test in the future if we ever change it.

socorro/external/es/crashstorage.py
```diff
@@ -124,18 +183,62 @@ def get_index_for_crash(self, crash_date):
...
+        # user-provided and can contain junk data
+        remove_bad_keys(raw_crash)
+
+        # Truncate values that are too long
+        truncate_keyword_field_values(FIELDS, raw_crash)
+        truncate_keyword_field_values(FIELDS, processed_crash)
+
+        # Capture crash data size metrics--do this only after we've cleaned up
+        # the crash data
+        self._capture_crash_metrics(raw_crash, processed_crash)
```
willkg: I moved when these methods/functions get called such that they happen outside the transaction. The transaction shouldn't be affected by any of these failing--it should only wrap issues with indexing in Elasticsearch.

socorro/external/es/crashstorage.py
```diff
@@ -124,18 +183,62 @@ def get_index_for_crash(self, crash_date):
...
+            'started_datetime',
+            'startedDateTime',
+            'completed_datetime',
+            'completeddatetime',
+        ]
+        for a_key in datetime_fields:
+            if a_key not in processed_crash:
+                continue
+
+            processed_crash[a_key] = string_to_datetime(processed_crash[a_key])
```
willkg: The side-by-side diff in github is sucky. I didn't change this method at all other than to add a FIXME about how these fields should be specified in `super_search_fields.py` and not hard-coded here.

socorro/external/es/crashstorage.py
```diff
@@ -157,69 +260,10 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash,
...
-        """
-        if not data:
-            return
-
-        # Copy the list of things we're iterating over because we're mutating
-        # the dict in place.
-        for key in list(data.keys()):
-            if not is_valid_key(key):
-                del data[key]
+            self.config.logger.exception('something went wrong when capturing processed_crash_size')
```
willkg: This fixes `raw` to `processed`. The original logging message was a copypasta fail.

willkg:
I want to wait until this lands before I work on but 1429598 (fix pseudo-booleans) because the two changes will overlap.

socorro/external/es/crashstorage.py
```diff
@@ -53,6 +57,61 @@ def is_valid_key(key):
...
+
+    # Copy the list of things we're iterating over because we're mutating
+    # the dict in place.
+    for key in list(data.keys()):
+        if not is_valid_key(key):
+            del data[key]
+
+
+def truncate_keyword_field_values(fields, data):
+    """Truncates values of keyword fields greater than 10,000 characters
```
Osmose: Maybe just reference the constant name here so we don't have to change it in two places in the future?

socorro/external/es/crashstorage.py
```diff
@@ -53,6 +57,61 @@ def is_valid_key(key):
...
+    Note: This modifies the data dict in-place and only looks at the top level.
+
+    :arg dict fields: the super search fields schema
+    :arg dict data: the data to look through
+
+    """
+    # Go through all the fields marked as keyword and truncate values
+    # if they're too large
+
+    for name, properties in fields.items():
```
Osmose: We can use `fields.values()` here since we're not using the name at all.

socorro/external/es/crashstorage.py
```diff
@@ -124,18 +183,62 @@ def get_index_for_crash(self, crash_date):
...
+            'completed_datetime',
+            'completeddatetime',
+        ]
+        for a_key in datetime_fields:
+            if a_key not in processed_crash:
+                continue
+
+            processed_crash[a_key] = string_to_datetime(processed_crash[a_key])
+
+    def _capture_crash_metrics(self, raw_crash, processed_crash):
```
Osmose: Do we need an underscore here? I can't really see a pattern for their use in this class.

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -11,13 +11,15 @@
...
-    is_valid_key,
     ESCrashStorage,
     ESCrashStorageRedactedSave,
     ESCrashStorageRedactedJsonDump,
     ESBulkCrashStorage,
+    is_valid_key,
     RawCrashRedactor,
+    truncate_keyword_field_values,
 )
+# from socorro.external.es.super_search_fields import FIELDS
```
Osmose: Leftover comment?

socorro/external/es/crashstorage.py
willkg: Good idea!

socorro/external/es/crashstorage.py
willkg: I put an `_` there because other things in the class that are internal to the class and aren't used in the tests have one. I don't think it matters a whole lot here and it's probably not a helpful signal. I'll nix it.

socorro/unittest/external/es/test_crashstorage.py
willkg: Yes--oops!

willkg:
Thank you!

PR #4306, opened 2018-01-23
fixes bug 1429863 - populate local dev env with data for graphs

willkg:
This adds two scripts which automate populating the local dev environment with enough information to show graphs on the product home page and other reports.

To use:

```
$ make dockersetup
$ make dockerupdatedata
$ ./docker/as_me.sh bash
you@processr:/app$ ./scripts/populate_data.sh
```

Some thoughts:

1. It's possible we could reduce the number of days and crashes-per-day numbers. I didn't want to keep fiddling with it.
2. I thought about making these two values changeable via command-line arguments, but didn't have a compelling reason to spend time on that (and add the additional script complexity). I vote we create a new bug for that if a compelling reason pops up.
3. I didn't write tests for either of these. We don't have "does this script run at all?" type tests in Socorro. `scripts/find_crashids.py` is really simple and if there's an issue it only affects local dev environment and it should be pretty obvious. `scripts/populate_data.sh` is a little more complicated, but mostly just calls existing scripts that have tests. If you think it's important to have tests, I'm game for ideas on how to write them such that the tests themselves aren't hard to maintain.
4. I didn't have a good idea where to document these scripts. Should it go in Getting Started, maybe? We've got a lot of scripts these days. They're not immensely discoverable.


Osmose:
Random thoughts while looking over this PR:

- I wonder if we're straying close enough to scripting the crash-pull-and-process dance that we just want to land a generic script for doing that, to cover both the "grab 100 crashes to populate my setup" and "grab 15 crashes per day to fix the front page" use cases. 
- I feel like any script that is required for part of the site to work properly should automatically be run by either `dockersetup` or `dockerupdatedata`, depending on whether they're safe to run repeatedly without wiping out data.
- Script discoverability is a problem. It might behoove us to consider things other projects do for stuff like this; Firefox has mach as the catch-all script runner, for example, and script discoverability is built-in. Maybe something like that?

scripts/populate_data.sh
```diff
@@ -0,0 +1,65 @@
...
+#
+# After running this script, run the processor:
+#
+#    $ docker-compose up processor
+#
+# Wait for crashes to process. Then you're good to go!
+
+NUM_DAYS=8
+NUM_CRASHES=15
+DATADIR=./crashdata_populate_tmp
```
Osmose: Do we want to use a temporary directory instead of hardcoding this?

willkg: I hard-coded all the things above figuring it's a good start for now and we can hone it as issues come up.

If you want to write to write a function to generate a temporary directory, I'm game for using that.

willkg:
I see your point about having all the important things in the setup scripts, but I definitely don't want to run this as part of that. It takes a while to run, it requires real data, and I'm not usually working on webapp graphs, so most of the time, I don't care whether they show data or not and wouldn't want to run this.

Regarding, "should this be generalized to more scenarios", maybe, but I can't think of a compelling reason to do that. The tricky thing that `populate_data.sh` deals with is "get x crashes per day for the last y days" where the hard part is the "per day for the last y days" part. That's hard to do by hand. I based this script on a script I've been using for a while.

Generally, I want to keep the number of scripts we've got in the repo to the bare minimum. They should meet real existing needs. Devs can write their own scripts to meet their own needs because these scripts are good building blocks with helpful interfaces. You and I have had a "need to populate the data to get the graphs to work" need for a while now. I think it's come up enough that there's a compelling reason to commit this script to the repo.

I'll toss it in the Getting Started docs for now. I don't want to manually maintain a bunch of script documentation. If I had my druthers, we'd have an autodoc kind of thing that we could point at the scripts and it would slurp out usage information. I've been trying to figure out how to do that, but it's kind of a back-burner thing.

willkg:
Actually, let's back up a step--is the premise even true? Is it compelling to have a "get x crashes per day for y days so we can see graphs" script? If not, then I'll nix this PR. I'm fine with having these scripts local and sticking them in a gist. Less to maintain.

Osmose:
> Actually, let's back up a step--is the premise even true? Is it compelling to have a "get x crashes per day for y days so we can see graphs" script? If not, then I'll nix this PR. I'm fine with having these scripts local and sticking them in a gist. Less to maintain.

FWIW I just use my "download 100ish crashes and process them" script whenever I want the graph to look right. It results in a much more ugly graph but it generally works. I'd use this script if it existed, but I dunno if I really need it.

I think what I feel strongly about is that the graphs should not be/look broken on a fresh setup of the development environment. Whether we make the graphs robust to missing data, fake the data source they're pulling from, or populate an initial install with crash data, I'm not too picky about.

willkg:
That makes sense. I was secretly hoping the graphs would go away and then the need for data goes away, too.

Given that, I'm going to close this out and just keep these local.

Thanks!

PR #4307, opened 2018-01-23
Fix bug 1430860: Move destination config to python-based source.

Osmose:
This is a WIP I wanna put up for feedback before starting to polish it up. It does two things.

1. Changes the `destination.storage_classes` config value to `destination.storage_namespaces`. The old value was a list of classpaths that corresponded to numerically-based config namespaces (i.e. the first class in the list was passed config values from `destination.storage1`, the second got `destination.storage2`, etc.).

   The new config now takes a list of namespace names, e.g. `postgres,s3,telemetry`. Each name is a crash storage location, and the name itself is used to lookup the config, e.g. `destination.postgres`, `destination.s3`, etc. This helps make editing, renaming, and reordering storages way less error-prone.
2. Adds a new method for providing default config values. Any `SocorroApp` subclass can now define a class property `config_module`. It should be set to a string containing a python module (or class) path.

   The module at the path specified will be imported and it's members will be used as configuration defaults. You can use DotDicts (and maybe any mapping? I haven't tested it yet.) to define nested config values. The end result is that any of the Socorro apps can define their own config defaults to replace `get_application_defaults` and `docker/*.env` files. 

This PR focuses on the destination crash storage config mostly because it had nice properties for testing if this method worked. After landing this, we can gradually move config over to python modules, and eventually deprecate the other config methods.


socorro/app/socorro_app.py
```diff
@@ -202,13 +203,18 @@ def _do_run(klass, config_path=None, values_source_list=None):
                 # pull in the application defaults from the 'application'
                 # configman option, once it has been defined
                 application_defaults_proxy,
+                # Get values from the app's config module
```
Osmose: Adding a common config for all apps would be as simple as adding a string to this list like `"socorro.config"` and putting the common config there.

socorro/app/socorro_app.py
```diff
@@ -202,13 +203,18 @@ def _do_run(klass, config_path=None, values_source_list=None):
...
+                klass.config_module,
                 # pull in any configuration file
```
Osmose: Adding a common config for all apps would be as simple as adding a string to this list like `"socorro.config"` and putting the common config there.

socorro/external/crashstorage_base.py
```diff
@@ -467,6 +467,47 @@ def __str__(self):
...
 
+class StorageNamespaceList(collections.Sequence):
+    """A sequence of configuration namespaces for crash stores.
+
+    This acts mostly as a list of strings that correspond to
+    destination.* config keys containing crash store configs. But it
+    also includes a required_config that tells configman how to find
+    the class storage classes, allowing them to define their own
+    required_configs.
+    """
```
Osmose: I'm not really happy with how the magic is explained here. The gist is that we need the value to have a `required_config` that includes the crash storage classes so configman resolves them and sees their `required_config`s during the option overlay stuff.

Osmose: Also, I couldn't figure out a clean way to not need this separate config value with names. It'd be nice to not have to specify it and instead just let Socorro detect `destination.storages.postgres` as a crash storage or something.

Osmose:
@willkg Lemme know your thoughts on this. Once it's in a place we're happy with I'll do up tests and docstrings and docs and the like.

socorro/processor/config.py
```diff
@@ -0,0 +1,69 @@
...
+        'statsd_prefix': 'processor',
+    },
+    'telemetry': {
+        'active_list': 'save_raw_and_processed',
+        'bucket_name': 'org-mozilla-telemetry-crashes',
+        'crashstorage_class': 'socorro.external.statsd.statsd_base.StatsdBenchmarkingWrapper',
+        'statsd_prefix': 'processor.telemetry',
+        'wrapped_object_class': 'socorro.external.boto.crashstorage.TelemetryBotoS3CrashStorage',
+    },
+})
```
willkg: SO MUCH YAY!

socorro/external/crashstorage_base.py
willkg: Pretty sure we need to specify the order of crash storage destinations. That's easily done with a list. There are probably other ways we can specify order, too.

I say "pretty sure" because the way PolyCrashStorage works isn't transactional and we need the crash data to exist in certain places before it exists in others. For example, it needs to exist in S3 before Elasticsearch. It's entirely possible that the code in PolyCrashStorage and the crash storage classes themselves is such that Socorro doesn't support this requirement. I don't think I've looked closely at that, yet.

willkg:
I think this is really great!

List of things we've mentioned earlier and/or just occurred to me as I read this:

1. We should create config modules for anything listed in `for_application_defaults` and move the configuration specified in the class into those files. Then we can ditch `for_application_defaults` related things.
2. Pretty sure we can ditch all of the app-specific `.env` files in `docker/config/`.
3. If that's true, then we can ditch the squirrely `.env` loading stuff like this: https://github.com/mozilla-services/socorro/blob/master/docker/run_crontabber.sh#L11
4. I wonder if this makes it hard to override configuration via `my.env` or whether we should set up a different way of overriding configuration for local development environment based development. My off-the-cuff suspicion is that we can keep `my.env` because that environment file will take precedence to Socorro app `config.py` files.

I think that's it. I might have more thoughts as this develops more.

I think this is super duper helpful. I'm glad you're pursuing this! I'd love to land it asap, so anything I can do to help this along--just let me know.

socorro/external/crashstorage_base.py
Osmose: I believe PolyCrashStorage supports writing to storages in order, but does not stop writing if a storage fails during the write, so we could still end up in a situation where ES has the crash but S3 doesn't.

Osmose:
This should be ready for a proper review. It's still very relevant to [bug 1432569](https://bugzilla.mozilla.org/show_bug.cgi?id=1432569), but doesn't fix it completely, so I focused on the bug it does resolve ([bug 1430860](https://bugzilla.mozilla.org/show_bug.cgi?id=1430860)).

I'll work on followup PRs for the changes you mentioned that are more relevant to the former bug.

socorro/app/socorro_app.py
```diff
@@ -202,13 +207,18 @@ def _do_run(klass, config_path=None, values_source_list=None):
...
                 ConfigFileFutureProxy,
                 # get values from the environment
                 environment,
                 # use the command line to get the final overriding values
                 command_line
             ]
+
+            # Remove potential None values
+            values_source_list = [x for x in values_source_list if x is not None]
```
Osmose: I would've liked to add tests for the fallback behavior of config settings here, but could not figure out a good way to simulate values from the different sources that didn't involve messy mock setups or providing a custom `values_source_list`, which doesn't test this anyway.

Osmose:
FYI I've found that the method this PR uses to pull in config files does not work with apps run via the `SocorroWelcomeApp`. I'm looking into that now, not sure if I'll land a fix in this PR or leave it for the followup where I found the issue.

socorro/app/socorro_app.py
```diff
@@ -202,13 +207,18 @@ def _do_run(klass, config_path=None, values_source_list=None):
                 # pull in the application defaults from the 'application'
                 # configman option, once it has been defined
                 application_defaults_proxy,
+                # Get values from the app's config module
+                klass.config_module,
```
Osmose: So here's the problem line. Since the processor is run directly by docker, `klass` is set to `ProcessorApp` and this finds the correct value. When an app is run via `SocorroWelcomeApp`, that's the value of `klass` here, because the welcome app has no idea what application it is running until after the config is generated.

This is why `for_application_defaults.py` has that proxy stuff, it uses `from_str_converter` to be notified when the application was selected, and then subs in values properly after that point. I did not fully understand why it was doing that until now.

Honestly, I think our best bet is ditching `SocorroWelcomeApp` completely and running all apps directly in the same way the processor is. It doens't seem like a useful level of indirection at this point, but I may be missing something.

The second best solution is probably to store the config module path in the `SOCORRO_APPS` constant so we know it ahead of time. The problem with this is that it makes it a bit awkward to do config inheritance, which mainly affects subclasses of `FetchTransformSaveApp`, which has some default config values it declares. However, I'm not sure if anyone is even using those default values, so this might actually work just fine too.

@willkg Thoughts? Can we kill `SocorroWelcomeApp`?

willkg: `SocorroWelcomeApp` is how `./scripts/socorro <appname>` works. In a different conversation we tossed around a way of coordinating scripts where they're subcommands of a single command. I like that model because I think it'll make discovery and documentation easier.

Would it be possible to rewrite `SocorroWelcomeApp` so it doesn't do any configman things, but instead just plucks off the subcommand, matches that to the app via a lookup somewhere, and then kicks off that app? I'd be game for taking a stab at that in a separate PR after this one lands.

Osmose: It's possible. I was thinking of using a library like [click](http://click.pocoo.org/) for the central script runner, but I dunno if that would conflict with the commandline features of configman. I also dunno if we want this command to be both a tool you'd use in the local development environment, and in the server environments to start off various apps or not.

I'll remove the welcome app as part of this PR and convert all the uses in-repo of it to just call the app directly (I believe they're almost all for crontabber), and then we can figure out what to do about a single command after.

willkg: I don't think we need to add click to do subcommands. I think we can change `scripts/socorro` to something like this:

```
import sys

# app_map would be a dict mapping app name to python dotted path
from socorro.app.app_registry import app_map
from socorro.app.app_registry import str_to_application_class

USAGE = 'socorro <subcommand> [options]'


def print_usage(msg=''):
    print(USAGE)
    print('Available subcommands:')
    for cmd in sorted(app_map.keys()):
        print('   ' + cmd)


def main(args):
    if not args or args[1] == '--help':
        print_usage()
        return 0

    if args[0] not in app_map:
        print('%s is not a valid subcommand. See "socorro --help".' % args[0])
        return 0

    # Knock the subcommand off the front of the args list
    sys.argv = sys.argv[1:]

    # Create the App class and then run it
    app_class = str_to_application_class(app_map[args[0]])
    return app_class.run(config_path=None, values_sources_list=None)


if __name__ == '__main__':
    sys.exit(main(sys.argv[1:]))
```

Then `app_registry.py` has a map of app names to dotted Python paths just like `for_application_defaults` has (sort of).

Or something really close to that.

Then the following would all work and do helpful things:

```
# Prints list of commands
$ ./scripts/socorro
$ ./scripts/socorro --help

# Runs command help
$ ./scripts/socorro crontabber --help

# Runs a command
$ ./scripts/socorro crontabber ...
```

click is pretty super, but I don't think we need it for this case.

I think your plan sounds good and we can redo the socorro script or whatever we decide in a follow-up PR.

socorro/app/socorro_app.py
```diff
@@ -136,6 +136,11 @@ class SocorroApp(RequiredConfig):
     app_version = "1.0"
     app_description = 'base class for app system'
 
+    #: String containing a module import path. The module is used as a
+    #: source for default configuration values. If None, this makes no
+    #: changes to the configuration defaults.
```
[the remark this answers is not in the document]
Osmose: I figured I might as well add the colons, I'd want that comment there even without them.

socorro/processor/config.py
Osmose: I checked prod and the config there matches this one. You should double check too though.

As far as I can tell, `wrapped_object_class` is used by `StatsdBenchmarkingWrapper`, while `wrapped_crashstore` is used by `BenchmarkingCrashStorage`, which is used on the crash source but not on any destinations.

Osmose:
I removed the unused config keys and added a commit to remove `SocorroWelcomeApp` and `scripts/socorro`.

socorro/app/socorro_app.py
willkg: At the top of this file and also on line 231, the comments talk about `SocorroWelcomeApp` which no longer exists.

I think we should fix the comments at the top of this file.

I think we should change the comment on line 231 to something like:

```
# FIXME(mkelly): We used to have a SocorroWelcomeApp that defined an "application" option. We no
# longer have that. This section should get reworked possibly as part of getting rid of application
# defaults.
```
```

Comment (willkg) [inline socorro/app/socorro_app.py]: At the top of this file and also on line 231, the comments talk about `SocorroWelcomeApp` which no longer exists.

I think we should fix the comments at the top of this file.

I think we should change the comment on line 231 to something like:

```
# FIXME(mkelly): We used to have a SocorroWelcomeApp that defined an "application" option. We no
# longer have that. This section should get reworked possibly as part of getting rid of application
# defaults.
```

**Comment by willkg** [comment]
After this lands, I think the following are all true:

1. we no longer have a `scripts/socorro` command and probably want to build one to make discovery easier -- this will be a new bug
2. we have some technical debt around `for_application_defaults`, `ApplicationDefaultsProxy` and `get_application_defaults` that we should clean up -- this should be a new bug
3. we'll have `.env` files in `docker/config/` to clean up -- this should be a new bug
4. we'll want to delete the processor destination configuration from consul in -stage and -prod -- we'll do this when deploying this code

Does that sound right?

**Comment by willkg** [comment]
Tagging this with a do-not-merge because we don't want to merge it until we've done a -prod deploy first.

**Comment by Osmose** [comment]
> Does that sound right?

Yep! I have a branch for (2) already done, actually.

**Comment by Osmose** [comment]
I've filed the followup bugs:

- https://bugzilla.mozilla.org/show_bug.cgi?id=1434132
- https://bugzilla.mozilla.org/show_bug.cgi?id=1434133
- https://bugzilla.mozilla.org/show_bug.cgi?id=1434124

**Comment by willkg** [comment]
This is good to merge now.

**Comment by Osmose** [comment]
I added the commit that removes get_application_defaults and friends, and will soon also be adding a commit that replaces some of the `.env` files with config files. Then this will be ready for a re-review.

**Comment by Osmose** [comment]
Updated, this is ready for review I think. I tested this locally by running through the system test checklist, as it seemed pretty far-reaching.

**Comment by willkg** [comment]
I'll re-review this on Monday.

File: socorro/cron/config.py [hunk 8]
```diff
@@ -0,0 +1,17 @@
...
+crontabber = DotDict({
+    'database_class': 'socorro.external.postgresql.connection_context.ConnectionContext',
+})
+
+
+resource = DotDict({
+    'postgresql': {
+        'database_class': 'socorro.external.postgresql.connection_context.ConnectionContext',
+    }
+})
```
willkg: This seems confusing to me because configuration keys end up defined using a mix of module-level variable names and DotDict keys.

If we do this instead:

```
klass_config = 'socorro.cron.config.CONFIG'
```

in the class and then have this file be:

```
CONFIG = DotDict({
    'crontabber': {
        'database_class': 'socorro.external.postgresql.connection_context.ConnectionContext',
    },
    'resource': {
        'postgresql': {
            'database_class': 'socorro.external.postgresql.connection_context.ConnectionContext',
        }
    }
})
```

Then all the keys for a configuration are keys rather than a mix of module-level names and DotDict keys.

One nice thing about that is that we could have multiple configurations defined in the same Python file, though we don't need that now.

socorro/app/fetch_transform_save_config.py
```diff
@@ -0,0 +1,18 @@
...
+# Crash storage source
+source = DotDict({
+    'crashstorage_class': 'socorro.external.fs.crashstorage.FSPermanentStorage',
+})
+
+
+# Crash storage destinations
+destination = DotDict({
+    'crashstorage_class': 'socorro.external.fs.crashstorage.FSPermanentStorage',
+})
```
willkg: Because the `klass_config` doesn't walk the MRE, this is all moot and won't come into play. We should nix this and make the values the default values for the config options.

socorro/cron/jobs/ftpscraper.py
```diff
@@ -537,6 +538,12 @@ class FTPScraperCronAppDryRunner(App):  # pragma: no cover
         $ python socorro/cron/jobs/ftpscraper.py --product=mobile,thunderbird
 
     """
+    config_module = 'socorro.cron.jobs.ftpscraper.FTPScraperCronAppDryRunner.application_defaults'
+
+    class application_defaults:
+        database = DotDict({
+            'database_class': mock.MagicMock(),
+        })
```
willkg: I don't think I ever use this class. There was a point where I was using it to debug some ftpscraper things, but then I discovered that ftpscraper calls a stored procedure and then a bunch of other things happen none of which happen in the dry-run effectively making dry-run unhelpful.

This class was much more important when there wasn't a local dev environment and you were running against -stage. But we don't do that anymore.

Anyhow, we can nix this. If you wanted to do that in this PR, that's cool. Otherwise I can write up a bug for it.

docker/run_crontabber.sh
```diff
@@ -8,19 +8,6 @@
...
-if [[ -z "${CRONTABBER_BEHAVIOR}" ]];
-then
-    echo "Pulling in crontabber behavior configuration..."
-    CMDPREFIX="/app/bin/build_env.py /app/docker/config/crontabber.env"
-else
-    echo "Already have crontabber behavior configuration..."
-    CMDPREFIX=
-fi
-
-
```
willkg: This was such a gross hack. I'm so glad it's going away.

socorro/cron/config.py
```diff
@@ -7,6 +7,8 @@
 
 crontabber = DotDict({
     'database_class': 'socorro.external.postgresql.connection_context.ConnectionContext',
+
+    'class-BugzillaCronApp.days_into_past': 0,
```
willkg: Mmm... Should this be:
```
'class-BugzillaCronApp': {
    'days_into_past': 0
}
```
?

socorro/processor/config.py
```diff
@@ -76,3 +76,53 @@
...
+
+
+new_crash_source = DotDict({
+    'crashstorage_class': 'socorro.external.rabbitmq.crashstorage.RabbitMQCrashStorage',
+    'new_crash_source_class': 'socorro.external.rabbitmq.rmq_new_crash_source.RMQNewCrashSource',
+})
+
+
+processor = DotDict({
+    'jit_classifers.JitCrashCategorizeRule.chatty': True,
```
willkg: I think this should be:

```
'jit_classifiers': {
    'JitCrashCategorizeRule': {
        'chatty': True
    }
}
```

However, @lonnen is ditching all the chatty things about transform rules, so we could just nix this line--either would be fine.

lonnen: may as well

Osmose:
Updated with changes from feedback. We decided on IRC that it would be even easier to define the in-code defaults as a class attribute on the `SocorroApp` subclasses, which also helps avoid the weird naming issues with the modules.

socorro/processor/processor_app.py
```diff
@@ -18,12 +18,147 @@
 from socorro.lib import raven_client
 
 
+# Defined separately for readability
```
willkg: HUGE +1.

willkg:
I skimmed the changes and they look ok. I want to do the following:

1. double-check the configuration movements
2. run through the system checklist locally

I think I can do that today. If not today, then definitely Monday.

socorro/cron/crontabber_app.py
```diff
@@ -99,6 +98,25 @@ def jobs_converter(path_or_jobs):
...
 class CronTabberApp(CronTabberBase, App):
+    config_defaults = {
+        'always_ignore_mismatches': True,
+
+        'crontabber': {
+            'database_class': 'socorro.external.postgresql.connection_context.ConnectionContext',
+
+            'class-BugzillaCronApp': {
+                'days_into_past': 0,
+            },
```
willkg: I think this is actually in the wrong namespace and has no effect on anything (this is true in consul for prod as well as the `crontabber.env` file).

Also, this is the default value:

https://github.com/mozilla-services/socorro/blob/d442a4df44a2be121a0ad35a7c3151afc885263a/socorro/cron/jobs/bugzilla.py#L95

We can nix the `class-BugzillaCronApp` clause.

socorro/cron/crontabber_app.py
```diff
@@ -99,6 +98,25 @@ def jobs_converter(path_or_jobs):
 
 
 class CronTabberApp(CronTabberBase, App):
+    config_defaults = {
+        'always_ignore_mismatches': True,
+
+        'crontabber': {
+            'database_class': 'socorro.external.postgresql.connection_context.ConnectionContext',
+
```
willkg: I think this one got lost:

```
'job_state_db_class': {
    'database_class': 'socorro.external.postgresql.connection_context.ConnectionContext'
}
```

Either that or you changed the way it gets set by setting `resource.postgresql.database_class` below.

If it's the latter, then `resource.postgresql.database_class` value will cover both `crontabber.database_class` as well as `crontabber.job_state_db_class.database_class` and thus we don't need to set `database_class` above.

Osmose:
Updated for feedback.

Osmose:
I'm still not entirely sure what the root cause of the intermittent failure is (why is one hit getting cached but not the other?), but I've added a reasonable fix. See the commit message for more details.

willkg:
I did some testing. crontabber seems fine. processor seems fine. When I go to `http://localhost:8000` in the webapp, it redirects me to `http://localhost:8000/home/product/WaterWolf`. I did `./manage.py shell` and it says the `settings.DEFAULT_PRODUCT` is WaterWolf. Maybe an errant setting somewhere? I can't tell.

I've got home issues, so I haven't looked into it, yet. But I figured I'd give an update on where I was at in case it rang any bells.

PR #4308, opened 2018-01-29
fixes bug 1433988 - update AuroraVersionFixitRule re: 59.0b2

willkg:
This adds buildids for all the .0b2 versions which are all also aurora channel.


willkg:
I pulled the buildids from buildhub when doing searches for the "devedition" product and the specified version.

willkg:
Thank you!

PR #4314, opened 2018-01-31
fixes bug 1132858 - remove the extensions table

willkg:
The extensions table was a partitioned table, so this creates a migration
to remove the table, all the partitions, and the `report_partition_info` record.
It also removes the model in the code.

### DIFF
--- a/alembic/versions/0193b4725f32_bug_1132858_remove_extensions.py
+++ b/alembic/versions/0193b4725f32_bug_1132858_remove_extensions.py
@@ -1,0 +1,51 @@
+"""bug 1132858 remove extensions table
+
+Revision ID: 0193b4725f32
+Revises: bb8cdbb8a6bd
+Create Date: 2018-01-31 14:21:41.032179
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '0193b4725f32'
+down_revision = 'bb8cdbb8a6bd'
+
+from alembic import op
+from socorro.lib import citexttype, jsontype, buildtype
+from socorro.lib.migrations import fix_permissions, load_stored_proc
+
+import sqlalchemy as sa
+from sqlalchemy import types
+from sqlalchemy.dialects import postgresql
+from sqlalchemy.sql import table, column
+
+
+def upgrade():
+    # Get a list of ALL tables that start with 'extensions_*' and drop them
+    connection = op.get_bind()
+    cursor = connection.connection.cursor()
+    cursor.execute("""
+        SELECT c.relname FROM pg_catalog.pg_class c
+        WHERE c.relkind = 'r' AND c.relname LIKE 'extensions_%'
+    """)
+    all_table_names = []
+    for records in cursor.fetchall():
+        table_name, = records
+        all_table_names.append(table_name)
+
+    all_table_names.sort(reverse=True)
+    for table_name in all_table_names:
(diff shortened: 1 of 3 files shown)


willkg:

1. check out the master branch and run `make dockersetup`
2. reset weekly-report-partitions crontabber job and then run it:
   ```
   app@processor:/app$ ./scripts/socorro crontabber --reset=weekly-reports-partitions
   ...
   app@processor:/app$ ./scripts/socorro crontabber --job=weekly-reports-partitions
   ```
3. verify the `extensions` and some `extensions_YYYYMMDD` tables exist in the db
4. switch to this branch
5. run the alembic migrations
   ```
   ```
6. verify there are no `extensions` tables in the db anymore
7. verify there's no entry in `report_partition_info` with `table_name='extensions'`

All done!

willkg:
As a side note, these tables are all empty in -prod. They haven't been used in a while.

alembic/versions/0193b4725f32_bug_1132858_remove_extensions.py
```diff
@@ -0,0 +1,51 @@
...
+from sqlalchemy.sql import table, column
+
+
+def upgrade():
+    # Get a list of ALL tables that start with 'extensions_*' and drop them
+    connection = op.get_bind()
+    cursor = connection.connection.cursor()
+    cursor.execute("""
+        SELECT c.relname FROM pg_catalog.pg_class c
+        WHERE c.relkind = 'r' AND c.relname LIKE 'extensions_%'
```
Osmose: Is this the same as searching through `information_schema.tables`? My Postgres-fu is very weak.

alembic/versions/0193b4725f32_bug_1132858_remove_extensions.py
```diff
@@ -0,0 +1,51 @@
...
+    """)
+    all_table_names = []
+    for records in cursor.fetchall():
+        table_name, = records
+        all_table_names.append(table_name)
+
+    all_table_names.sort(reverse=True)
```
Osmose: Why the sort?

alembic/versions/0193b4725f32_bug_1132858_remove_extensions.py
willkg: My postgres-fu is also weak.

This isn't my style of SQL--I don't tend to do single-letter names. I pulled this code from a previous migration and kind of left it as is because it worked before and it didn't seem super helpful to hone it into something better.

I can hone if you think it helps, though. It probably wouldn't take long and it's easily tested.

alembic/versions/0193b4725f32_bug_1132858_remove_extensions.py
willkg: So... it looks like the migration I copied this code from picked up `plugin_reports` in addition to the partitioned tables--`plugin_reports` had to be deleted last.

In this case, since my LIKE clause doesn't pick up `extensions`, we don't have the same issue. So we can nix this.

willkg:
I'll clean this up a bit with another commit with the thinking that we're copying this code around and might as well fix it up a bit.

willkg:
^^^ That rebases against master and fixes the migration code so it's cleaner.

willkg:
After this lands, I need to wait for the -stage deploy, then update the -stage admin node and run the migration.

willkg:
Thank you!

PR #4316, opened 2018-01-31
fixes bug 1431108 - removes symbols admin view

willkg:
This removes the symbols admin view and related bits from Socorro. Symbols
is now fully handled by Tecken, so we don't need this anymore.


willkg:
I did the following:

1. grepped around
2. verified we could still build static assets
3. ran the webapp and made sure the admin pages continued to work
4. ran the tests and a linter

Pretty sure I caught everything.

Thank you for reviewing!

PR #4317, opened 2018-01-31
bug 1431108 - remove manage/js/symbols-uploads.js file

willkg:
This file was only used by the admin view and that's gone now.


willkg:
Bah--I missed removing this `.js` file. It's only used by that view, so I orphaned it.

willkg:
Ha! This was my fault. I didn't check to see if the files used in the bundle were used by anything else. Oh well.

Thank you!

PR #4318, opened 2018-01-31
fixes bug 1434676 - remove crashes per 100 adi graphs

willkg:
The product home page had a "crashes per 100 adi" graph which is unhelpful
and misleading.

This removes it altogether.

Also, this fixes bug 1376338 by adding links to the other crash types
on the top crashers page.


willkg:
Two nice things about this:

1. we no longer have to deal with local development environments showing errors on the product home page (yay!)
2. the product home page comes up much faster now

webapp-django/crashstats/home/jinja2/home/home.html
```diff
@@ -21,46 +14,13 @@
...
-            <div id="homepage-graph-legend" class="homepage-graph-legend"></div>
-        </div>
-    </div>
-
-
-    <div class="panel">
+    <div class="panel"
+        id="homepage-data-panel"
+        data-versions="{{ versions | to_json }}"
+        data-product="{{ product }}"
```
willkg: These are both used by the remainder of the javascript, so I relocated them to this div.

webapp-django/crashstats/home/jinja2/home/home.html
```diff
@@ -86,10 +46,30 @@ <h4>{{product}} {{version}}</h4>
                                 version=version,
                                 process_type='plugin',
                                 days=days,
-                            ) }}">
+                                ) }}">
```
willkg: This is a mistake--I'll fix it.

webapp-django/crashstats/home/jinja2/home/home.html
```diff
@@ -86,10 +46,30 @@ <h4>{{product}} {{version}}</h4>
...
+                        <li>
+                            <a href="{{ url('topcrashers:topcrashers') }}?{{ make_query_string(
+                                product=product,
+                                version=version,
+                                process_type='gpu',
+                                days=days,
+                                ) }}">
+                                Top GPU Crashers
+                            </a>
+                        </li>
```
willkg: I added these two figuring I'd fix bug 1376338 while I was working on things here.

webapp-django/crashstats/home/jinja2/home/home.html
```diff
@@ -86,10 +46,30 @@ <h4>{{product}} {{version}}</h4>
                                 version=version,
                                 process_type='plugin',
                                 days=days,
```
willkg: This shouldn't be passing "days" as an argument. Ditto for the other links. I'll fix that.

webapp-django/crashstats/home/static/home/js/home.js
```diff
@@ -31,18 +31,10 @@ $(function () {
...
 
-    var product = container.data('product');
-    var versions = container.data('versions');
-    var platforms = container.data('platforms');
-    var duration = container.data('duration');
-    var esShardsPerIndex = container.data('es-shards-per-index');
-
-    var pageTitle = $('title').text();
+    var product = datapanel.data('product');
+    var versions = datapanel.data('versions');
```
Osmose: `product` is unused in the rest of the remaining file, and is in a closure, so other scripts can't use it either.

`versions` is only used to generate `betaVersions`, which is in a similar state to `product`.

Looking through the whole file, I think the only things that are still used in the entire file are the bits that set the colors for the version headers, and you could get rid of that completely with CSS or the [cycle](https://docs.djangoproject.com/en/2.0/ref/templates/builtins/#cycle) builtin. Seems like it's worth doing now that this page is so simple.

willkg: So... it sounds like you're saying "this works, but it could be cleaner code". Is that right?

Assuming that's right, I'm not sure I understand exactly what you're asking for, so it'd take me some modicum of time to figure it out and do it. I think I'd rather land this now because it's blocking other things I need to get done and the JS/CSS things will get cleaned up during Photonic or some followup PR.

willkg: Scratch that. I feel lousy after writing that comment. I'll try to fix it tonight and we can push forward tomorrow.

willkg: I'm not sure whether the colors are important or not. Previously it was used to make sure the header color was the same as the color used in the graph line. We don't have graphs now. Maybe the header colors don't matter. I think I'll just nix them and we can re-add them later if need be.

willkg:
^^^ That drops `home.js` altogether.

willkg:
Thank you!

PR #4323, opened 2018-02-01
 bug 1422923 - removes symbols upload ui and code

willkg:
This removes all of symbols upload except the db model. We're going to
keep that and remove it in a separate commit later.

This creates a redirect from `/symbols/*` to `https://symbols.mozilla.org/`
in case people try to go here for symbol uploads.


willkg:
The tests cover most of the issues we might have: references in templates to urls that don't exist anymore, references in Python modules to code that doesn't exist anymore, references in bundles to JS/CSS files that don't exist anymore, etc.

Beyond that, you could run the webapp and verify that there aren't symbols upload related links on the bottom. You could also try to go to http://localhost:8000/symbols/ and get happily redirected to Tecken.

This leaves the db model and some empty files around because this is still a Django app. We need to keep the model there and keep it as a Django app to make it possible to do a migration later to wipe everything out. I had left the empty files there because I can't remember exactly what files need to exist for Django to think it's an app. All that will go away once we do a "kill the table" migration. I'm waiting on dumping the db to do that. Figured I'd do all the code first.

willkg:
I dumped the db. I'm going to add a "delete the table" migration and wipe out the app altogether now and add that to this PR.

willkg:
^^^ That adds a migration to delete the table and a README to remind us when to delete the minimal bits we've still got. I'll probably write up a bug for that.

When this goes to -stage and -prod, we'll need to run the Django migrations. I already dumped the data in the table in -prod, so we're good to run this whenever we can.

This is ready to review.

willkg:
Oo--I forgot about the checklists. I'll do that now.

willkg:
Thank you!

PR #4329, opened 2018-02-08
fixes bug 1435068 - remove tcbs and tcbs_build tables

willkg:
These tables have been sitting around unused since May 2017. Time for them to go.

### DIFF
--- a/socorro/external/postgresql/models.py
+++ b/socorro/external/postgresql/models.py
@@ -217,39 +217,6 @@ ischema_names['build_type_enum'] = build_type_enum
 ###############################
 
 
-class Tcbs(DeclarativeBase):
-    __tablename__ = 'tcbs'
-
-    # column definitions
-    signature_id = Column(u'signature_id', INTEGER(), ForeignKey(
-        'signatures.signature_id'), primary_key=True, nullable=False, index=True)
-    report_date = Column(u'report_date', DATE(),
-                         primary_key=True, nullable=False, index=True)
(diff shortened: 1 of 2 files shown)


socorro/external/postgresql/models.py
```diff
@@ -944,12 +911,6 @@ class ReleaseChannel(DeclarativeBase):
         secondary='ProductReleaseChannel',
         secondaryjoin='ProductReleaseChannel.product_name==Product.product_name'
     )
-    signatures = relationship(
-        'Signature',
-        primaryjoin='ReleaseChannel.release_channel==Tcbs.release_channel',
-        secondary='Tcbs',
-        secondaryjoin='Tcbs.signature_id==Signature.signature_id'
-    )
```
willkg: These models are set up as if we used sqlalchemy as an ORM in the backend, but we don't. Everything (and I mean *everything*) is done with raw sql. Further, you can't even use the ORM if you tried because of a bunch of secondary joins that aren't valid.

Given that, I'm just taking the license to remove this because it can't be used in the current condition of things.

Same for the other relationship that was removed in this PR.

willkg:

1. check out the master branch and run make dockersetup
2. verify the `tcbs` and `tcbs_build` tables exist in the db
3. switch to this branch
4. run the alembic migrations
   ```
   ```
5. verify the `tcbs` and `tcbs_build` tables are gone

All done!

willkg:
Thank you!

PR #4331, opened 2018-02-08
fix bug 1299465 - nix CrashData classes we're not using

willkg:


willkg:
Actually, I'm going to make this one big PR. Closing this for now.

PR #4332, opened 2018-02-08
fix bug 1299465, 1435551 - remove CrashData classes and stop saving to processed_crashes_* tables

willkg:


socorro/external/postgresql/crashstorage.py
```diff
@@ -25,20 +17,16 @@
...
 )
 
 
-class PostgreSQLBasicCrashStorage(CrashStorageBase):
-    """this implementation of crashstorage saves processed crashes to
-    an instance of Postgresql.  It only saves certain key values to the
-    partitioned reports table, therefore it is not a source for fetching
-    complete processed reports and doesn't not implement any of the 'get'
-    methods."""
+class PostgreSQLCrashStorage(CrashStorageBase):
```
willkg: I figured it was easier to squash these two classes than keep them separate. We have no need of having two separate classes and it doesn't help us at all--it's just more maintenance.

Squashing them was pretty straight-forward: couple of name changes, move a few functions, remove a bunch of tests, and done.

socorro/external/postgresql/crashstorage.py
```diff
@@ -292,32 +279,6 @@ def _get_raw_crash_transaction(self, connection, crash_id):
...
-                    processed_crash_table_name
-        try:
-            return single_value_sql(connection, fetch_sql, (crash_id,))
-        except ProgrammingError as e:
-            err = 'relation "%s" does not exist' % processed_crash_table_name
-            if err in str(e):
-                raise CrashIDNotFound(crash_id)
-            raise
-        except SQLDidNotReturnSingleValue:
-            raise CrashIDNotFound(crash_id)
```
willkg: Since we're no longer storing in the `processed_crashes_*` tables, we needed to ditch these two methods which fetch from them, too.

These were only used by the `CrashData` class which goes away in this PR, too.

socorro/external/postgresql/crashstorage.py
```diff
@@ -363,49 +324,3 @@ def _save_raw_crash_transaction(self, connection, raw_crash, crash_id):
...
-        UNION ALL
-        SELECT * from insert_processed_crash
-        """ % {'table': processed_crashes_table_name, 'uuid': crash_id}
-
-        values = {
-            'processed_json': json.dumps(processed_crash, cls=JsonDTEncoder),
-            'date_processed': processed_crash["date_processed"],
-            'uuid': crash_id
-        }
-        execute_no_results(connection, upsert_sql, values)
```
willkg: This removes saving to the `processed_crashes_*` tables.

socorro/external/postgresql/raw_sql/procs/truncate_partitions.sql
```diff
@@ -22,10 +22,10 @@ safety_weeks := weeks_to_keep::text || ' weeks';
 
 SELECT into cutoffdate (now() - safety_weeks::interval)::date;
 
--- Currently only truncating raw_crashes and processed_crashes
+-- Currently only truncating raw_crashes
 basenames := $q$SELECT table_name FROM report_partition_info
         WHERE partition_column = 'date_processed'
-        AND table_name IN ('raw_crashes', 'processed_crashes')
+        AND table_name = 'raw_crashes'
```
willkg: Since we're not saving to the `processed_crashes_*` tables, there's nothing to truncate.

socorro/external/postgresql/staticdata.py
```diff
@@ -172,15 +172,6 @@ class ReportPartitionInfo(BaseTable):
...
         ),
-        (
-            'processed_crashes',
-            '12',
-            '{uuid}',
-            '{date_processed}',
-            '{}',
-            'date_processed',
-            'TIMESTAMPTZ'
-        ),
```
willkg: I'm removing the section from the `report_partition_info` section in static data which only affects the local dev environment. This PR doesn't have a migration to remove the tables or the `report_partition_info` row from -stage and -prod.

I'm on the fence about whether to include that in this PR or whether to push it off to a future PR.

Thoughts?

socorro/unittest/external/postgresql/test_crash_data.py
```diff
@@ -1,216 +0,0 @@
...
-        connection.commit()
-
-        cursor.execute("SELECT weekly_report_partitions(2, '2012-03-14');")
-        cursor.execute("SELECT weekly_report_partitions(2, '2012-08-20');")
-        connection.commit()
-
-        # A complete crash report (raw, dump and processed)
-        fake_raw_dump_1 = 'peter is a swede'
-        fake_raw_dump_2 = 'lars is a norseman'
-        fake_raw_dump_3 = 'adrian is a frenchman'
```
willkg: I like reducing the amount of code we have to maintain, but I will miss some of it.

socorro/unittest/external/postgresql/test_crashstorage.py
```diff
@@ -297,13 +294,13 @@ def test_basic_postgres_save_processed_success_3_truncations(self):
         crashstorage.save_processed(a_processed_crash_with_everything_too_long)
 
         assert mocked_database_connection_source.call_count == 1
-        assert mocked_cursor.execute.call_count == 3
+        assert mocked_cursor.execute.call_count == 2
```
willkg: Calling `save_processed` used to save to reports tables, then plugins tables, then the processed crashes tables. Now it just does the first two. Thus this goes from 3 to 2.

socorro/unittest/external/postgresql/test_crashstorage.py
```diff
@@ -266,7 +263,7 @@ def fetch_all_func(*args):
         crashstorage.save_processed(a_processed_crash)
 
         assert mocked_database_connection_source.call_count == 1
-        assert mocked_cursor.execute.call_count == 4
+        assert mocked_cursor.execute.call_count == 3
```
willkg: We're not longer saving to the `processed_crashes_*` tables, so we can remove one call.

socorro/unittest/external/postgresql/test_crashstorage.py
```diff
@@ -230,7 +227,7 @@ def test_basic_postgres_save_processed_success(self):
         crashstorage.save_processed(a_processed_crash)
 
         assert mocked_database_connection_source.call_count == 1
-        assert mocked_cursor.execute.call_count == 3
+        assert mocked_cursor.execute.call_count == 2
```
willkg: We're not longer saving to the `processed_crashes_*` tables, so we can remove one call.

socorro/unittest/external/postgresql/test_crash_data.py
peterbe: :)

socorro/external/postgresql/staticdata.py
Osmose: IMO including it would be fine.

willkg:
I'm going to wait until pr #4329 lands to finish this up because that PR has also got a migration. Alembic allows for complex graphs of migrations, but it's easier if we don't have to exercise that feature.

willkg:
^^^ That rebase against master and adds the migration.

To test this:

1. check out master
2. run `make dockersetup` and `make dockerupdatedata`
3. process some crashes
4. verify there are `processed_crashes_*` tables in the db
5. switch to this branch
6. run the alembic migrations:
   ```
   $ docker-compose run processor alembic -c docker/config/alembic.ini upgrade heads
   ```
7. verify there are no more `processed_crashes*` tables in the db
8. run the `truncate-partitions` crontabber job to make sure that still works:
   ```
   $ docker-compose run crontabber bash
   app@crontabber:/app $ ./scripts/socorro crontabber --reset=truncate-partitions
   [SO MUCH LOGGING]
   app@crontabber:/app $ ./scripts/socorro crontabber --job=truncate-partitions
   [SO MUCH LOGGING]
   ```
9. process some crashes

That verifies the migration, the stored procedure still works, and crash processing works. That pretty much covers it!

lonnen:
whoa. 🥇

willkg:
Thank you!

FYI, this will drop the db size by about 200GB which is roughly by half.

PR #4333, opened 2018-02-12
Use Pipfile to manage dependencies

stephendonner:
@chartjes @davedash sorry for the multiple iterations - TBH, I'm not 100% sure this will have @pyup-bot auto-manage dependencies, as it's already in use, and here I'm specifying another directive/location).

Passing ad-hoc here: https://qa-preprod-master.fxtest.jenkins.stage.mozaws.net/job/socorro.adhoc/124/console


chartjes:
@stephendonner 👍

stephendonner:
I've just discovered an issue, so I've emailed support@pyup.io, asking how to rectify adding and using a Pipfile in /e2e-tests/ while at the same time honoring the other dependencies, which are specified in https://github.com/mozilla-services/socorro/blob/d442a4df44a2be121a0ad35a7c3151afc885263a/.pyup.yml

1) I don't know whether it's possible to have the bot update from two separate .pyup.yml files (as this patch introduces the 2nd), *and*...
2) It'd obviously be preferred to maintain the single file we have now, and just tweak the location/filename for the new Pipfile, but the reference in https://github.com/pyupio/pyup/issues/197#issuecomment-360391226 doesn't spell out how to do custom locations for Pipfiles, in a shared, single .pyup.yml.

Will report back/update the PR as I find out (sorry!)

willkg:
Why're we switching to Pipfile for the e2e-tests? What's the compelling reason to introduce another way of managing dependencies that has different tooling in this repository?

chartjes:
@willkg I think we have two options:

1. Stay with pyup+pip for all the Python dependencies for both testing and the project itself
2. Use pyup+pipenv for the testing-dependencies (I use pipenv locally) and stick with pyup+pip for the app-related dependencies

Since this is your project and not mine, I am happy to go in whatever direction you decide.

willkg:
There's no bug associated with this change and nothing in the PR comments to suggest *why* it's happening, so I have no idea. Why're we making this change?

chartjes:
@willkg On the services QA side we are starting to using pipenv to manage Python dependencies for our projects, so it seemed logical to have the e2e testing dependencies use it as well.

Like I said a few comments back, I'm not gonna put up a fight over sticking with pip + pyup for this repo.

willkg:
Does QA have bugs or documentation somewhere covering how they do things and conventions?

chartjes:
@willkg We do not.

I am recommending this PR be closed.

peterbe:
[*"The bot won't update the hash in Pipfile.lock."*](https://github.com/pyupio/pyup/issues/197#issuecomment-360391226)

This was why we switched from GreenKeeper to RenovateApp on a `yarn` based project. With GreenKeeper they'd make a nice PR to update `package.json` but it didn't update `yarn.lock` after merging one of those, if a developer then did `git pull origin master && yarn install` it would change `yarn.lock`. 

The equivalent is that pyup would change `Pipefile` but not upgrade `Pipefile.lock` which is useless. The two files would basically contradict each other if you accept the PRs from Pyup. 

Pipenv might be awesome for local dev, but until [pyup doesn't support it](https://github.com/pyupio/pyup/issues/197) it's not a good idea. Perhaps there are better alternatives to Pyup that I don't know about yet.

PR #4335, opened 2018-02-12
fixes bug 1437414 - clarify crash data situation in getting started

willkg:
This clarifies that you have a local development environment with no crash
data in it and that you need to process some crashes to add some. It also
moves the link to the processor chapter up a bit so it's clearer where those
docs are.

@lonnen Does this help at all with bug 1437414?


lonnen:
r+, that would have been enough to get unstuck

willkg:
Thank you!

PR #4339, opened 2018-02-14
fixes bug 1434930 - remove missing symbols tracking

willkg:
This removes all the missing symbols tracking. That's done by Tecken now, so we don't need to be doing that anymore.

This removes a processor rule, so it's worth processing a crash.

This also removes a crontabber job, but that job only runs in -prod, so it doesn't affect the local dev environment and we won't notice anything until we push it to -prod. 😢 

1. check out the master branch and run make dockersetup
2. verify there are `missing_symbols*` tables in the db
3. switch to this branch
4. run the alembic migrations
   ```
   ```
5. verify the `missing_symbols*` tables are gone and there's no entry for it in `report_partition_info`

That's it!


willkg:
Thank you!

PR #4340, opened 2018-02-15
fixes bug 1434124 - implement socorro-cmd

willkg:
I threw together a rough `socorro-cmd` script. Here's some sample output from it showing help text:

https://pastebin.mozilla.org/9078017

It can run Socorro Apps as well as the Python-based command line scripts we wrote for the local development environment. It lists all the scripts. It breaks up all the things into "logical groups" around usage scenarios. It points to further documentation. It requires me to tweak the Python-based command line scripts a little, but it's pretty minor. I tweaked `fetch_crashids` as an example. This command will let us get rid of the stubs in `scripts/` which is good.

For processing crashes with this script, we'd do something like this:

```
you@processor:/app$ ./scripts/socorro-cmd fetch_crashids --num=5 | ./scripts/socorro-cmd fetch_crash_data ./testdata
```

I'm up for making "socorro-cmd" shorter, but I didn't have a good idea on how other than "socorro" which I really don't want to do since it's also a Python module. Maybe "sc"? I'm game for putting it in the root of the repo, too.

If we did that and fixed the containers to include `.` in the PATH, then we'd have something like this:

```
you@processor:/app$ sc fetch_crashids --num=5 | sc fetch_crash_data ./testdata
```

What do you think? How does this feel?


Osmose:
This looks rad, thanks for this! Random thoughts as I play with it:

- I like the categorization of commands, especially how it reads out in the docs.
- Having to start a docker container with bash in it to run these commands is kinda annoying. I don't generally have a container running with a prompt in it all the time, so my use would almost always be prefixed by starting a container. Ideally each command would know what container it needs to run in.
- This is out of scope, but I still wonder if having a command for "process X crashes" would be useful to have instead of only having the primitives (download crashes, upload to S3, add to queue) and relying on scripts. I've been running my particular "Process 100 crashes" script all the time and haven't run the individual commands from it in a while.
- I don't think it's a big deal to name the command `socorro`, and I'd also recommend putting it at the root of the repo to help shorten the invocation a bit. I don't think a short name like `sc` is very helpful as a change.

willkg:
Just to clarify, you can run these in any of these ways:

```
$ ./docker/as_me.sh bash
you@processor:/app$ socorro-cmd
```

```
$ ./docker/as_me.sh ./socorro-cmd
```

```
$ docker-compose run processor ./socorro-cmd
```

The `as_me.sh` script gives you a convenient way to run things in containers without dealing with privilege issues. I think that works fine here, too. I highly encourage people to write scripts and aliases to fit their specific workflows. Maintain a gist. Share it with others. Build support. The really important use cases will rise to the top.

Which brings me to the "I wish there was a script for processing 100 crashes". I think @lonnen would like that, too. I think it'd further alleviate his "a local dev environment with no crash data is useless" concern. It's worth writing up a bug for that.

Usage of the script now:

```
$ ./docker/as_me.sh ./socorro-cmd --help
```

I updated the processor docs which include script usage scenarios, so you can see how it's used more there.

I also fixed the existing scripts so you can run them in a bunch of ways and they still produce correct usage help.

The one thing this script doesn't do that I'm still thinking about is running shell scripts. We have `scripts/socorro_aws_s3.sh` which is annoying to type out. This doesn't cover that. I have some thoughts about how to do that, but I haven't done it, yet.

Another thing I'm on the fence about is whether to print the runner in the usage. Is this more helpful?

```
$ ./docker/as_me.sh ./socorro-cmd --help
usage: socorro-cmd [-h] [cmd] [arg [arg ...]]

positional arguments:
  cmd         command to run
  arg         command arguments

optional arguments:
  -h, --help  show this help message and exit

Available commands:

Crash processing utilities:
  add_crashid_to_queue (socorro.scripts.add_crashid_to_queue.main)
  fetch_crash_data (socorro.scripts.fetch_crash_data.main)
  fetch_crashids (socorro.scripts.fetch_crashids.main)
  reprocess (socorro.scripts.reprocess.main)

Local development environment maintenance:
  clear_indices (socorro.external.es.clear_indices_app.ClearESIndicesApp)
  create_recent_indices (socorro.external.es.create_recent_indices_app.CreateRecentESIndicesApp)
  setupdb (socorro.external.postgresql.setupdb_app.SocorroDBApp)

Services (run these in their respective containers):
  crontabber (socorro.cron.crontabber_app.CronTabberApp)
  processor (socorro.processor.processor_app.ProcessorApp)
  submitter (socorro.submitter.submitter_app.SubmitterApp)

Type "socorro-cmd CMD --help" for help on any command.

All commands need to be run in a Docker container. Service commands are run in
the related container. For example, the "processor" command should run in the
"processor" container. All other commands should be run in the processor
container.

For more documentation, see <https://socorro.readthedocs.io/>.
```

I'm on the fence because it makes it a lot busier, however as a dev, it's nice to know *where* that code is. Though we could throw in another command for showing that. Maybe `socorro-cmd showcommands`?

Osmose:
> I'm on the fence because it makes it a lot busier, however as a dev, it's nice to know where that code is. Though we could throw in another command for showing that. Maybe socorro-cmd showcommands?

Having a subcommand for showing that info is a good compromise.

willkg:
^^^ That adds a `showcommands` subcommand plus the ability to have function runners. I doubt we'll use it much, but it was easy to do and isn't very invasive.

willkg:
This is ready for review.

socorro-cmd
```diff
@@ -0,0 +1,199 @@
...
+        parser.print_help()
+        return 0
+
+    cmd = sys.argv[1]
+    runner = get_runner(cmd)
+    if runner is None:
+        parser.error('"%s" is not a valid command; see --help for command list' % cmd)
+
+    # Rewrite sys.argv so anything that parses sys.argv has the right one
+    sys.argv = [sys.argv[0] + ' ' + cmd] + sys.argv[2:]
```
Osmose: Maybe instead of modifying `argv` we can just pass around a truncated copy as arguments to the runners? Then we could avoid modifying global state. I'm not sure if it matters all that much though.

willkg: This does two things:

First, argparse (which everything we've got uses) treats the first argument of sys.argv as the script and that's what it uses for `%PROG`. So the help text shows the right thing regardless of how you executed the script.

Second, the way `SocorroApp` is constructed, it lets configman do its thing in regards to command line arguments. We'd need to make non-trivial changes to `SocorroApp` and how it gets created. We could rework things, but given the first item, I decided not to at this juncture.

willkg: For the record, I think that behavior of argparse is weird. It's overridable, but it's not conveniently so.

willkg: Also, I'm not super wild about the way all this works. I think we'll hone it over time.

willkg:
Thank you! `socorro-cmd woo-hoo`

PR #4343, opened 2018-02-15
Fix bug 1438612: Drop Travis CI config and related code.

Osmose:
- AFAICT, `clean.sh` and `test.sh` are only used in Travis. The other scripts are used by `make package`, which comments claim is how the current Socorro infra builds the RPM for deployment, so I left them in.
- The `BUILD_TYPE` environment variable (used in `install.sh`, `package.sh`, defined in `scripts/defaults`) has a `ci` value that I don't believe Circle currently uses, but I think they are used when building stackwalker locally, so I left it in, although leaving it named `ci` is a little misleading.


scripts/defaults
```diff
@@ -3,7 +3,7 @@ echo "this is the file defaults"
 export PYTHON=${PYTHON:-"python"}
 export VIRTUAL_ENV=${VIRTUAL_ENV:-"$PWD/socorro-virtualenv"}
 # Valid values for BUILD_TYPE are "rpm" (which we use in socorro-infra) and
-# "ci" # which we use in Travis CI.
+# "ci" # which we use in CI.
```
willkg: I wrote this after spending *quality time* with these scripts. I'm pretty sure you're right in that this leaves us with only one build type in use. We could remove it, but it'd be worth verifying the changes.

As I recall, I ended up setting up a Vagrant environment (which I don't think I can reproduce now without effort) to test things. It wasn't fun.

All these scripts go away after we switch to the new infra, so I'd just leave it as is rather than try to figure out how to test everything.

requirements/default.txt
```diff
@@ -150,8 +150,7 @@ django-pipeline==1.6.14 \
 django-session-csrf==0.7.1 \
     --hash=sha256:e17177e6e2e6518ec7ce6693ad10a5c747f8571d09f4cfa9082599334421605d \
     --hash=sha256:ff8c10e30d312c77fc6a6db7710e22b9383e28c03b7fe958876ca96f39aa6cf2
-# the 'linux2' is for TravisCI
-pyinotify==0.9.6 ; sys_platform == 'linux' or sys_platform == 'linux2' \
+pyinotify==0.9.6 ; sys_platform == 'linux' \
```
willkg: Circle failed--we need to re-add "linux2" here. I don't know why the comment was like that, but probably my fault for not updating it for Circle.

PR #4344, opened 2018-02-16
fixes bug 1438969 - fix processor service

willkg:
The processor service was still using the "socorro" script, but we nixed that.
This updates the service file to run the processor app directly.


willkg:
This change isn't covered in CI at all, so I'm going to merge it before CI is done.

PR #4348, opened 2018-02-20
fix bug 1431546 - raise errors on bad API token

willkg:
If `fetch_crash_data` is using a bad api token (expired, non-existent, etc)
it should raise that as an error and not continue rather than save the
error it's getting back from Socorro as the raw crash. This makes that so.


Osmose:
Failing test is the intermittent, if you rebase on master it should go away.

willkg:
Thank you!

PR #4349, opened 2018-02-20
fix bug 1438921 - remove hacky telemetry bucket_name configuration

willkg:
After the settings overhaul where we now store behavior configuration in
Python files and the `PolyCrashStorage` doesn't namespace using indexes, we
no longer need the hack to set the telemetry `bucket_name`.

This removes that code.

### DIFF
--- a/docker/config/local_dev.env
+++ b/docker/config/local_dev.env
@@ -54,7 +54,6 @@ resource.boto.port=5000
 resource.boto.access_key=foo
 secrets.boto.secret_access_key=foo
 resource.boto.bucket_name=dev_bucket
-resource.boto.telemetry_bucket_name=telemetry_bucket
 resource.boto.temporary_file_system_storage_path=/tmp
 resource.boto.calling_format=boto.s3.connection.OrdinaryCallingFormat
 resource.boto.resource_class=socorro.external.boto.connection_context.HostPortS3ConnectionContext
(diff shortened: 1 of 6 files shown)


willkg:
To test this, set everything up with:

```
$ make dockersetup
$ make dockerupdatedata
```

Then run this to verify that the `telemetry_bucket` is empty:

...
$ docker/as_me.sh bash
you@processor:/app$ ./scripts/socorro_aws_s3.sh ls --recursive s3://telemetry_bucket/
```

That should show this output:

```
S3 container bucket is dev_bucket
```

That "S3 container bucket is dev_bucket" is helpful in some contexts, but confusing in this one--you can ignore that. You shouldn't see any other output because the bucket is empty.

Process a crash. Then run crontabber with `docker-compose up crontabber` and let it go through all the jobs.

List the contents of the `telemetry_bucket` and it should show something like this (your crash id and date will probably be different):

```
you@processor:/app$ ./scripts/socorro_aws_s3.sh ls --recursive s3://telemetry_bucket/
S3 container bucket is dev_bucket
2018-02-20 14:54:01      33618 crash_report.json
2018-02-20 14:56:05     253894 v1/crash_report/20180219/17b723c2-49bc-48cd-8ae0-519120180219
```

That's it!

willkg:
Thank you!

PR #4350, opened 2018-02-21
bug 1439928 - fix processor

willkg:
This fixes a bunch of issues related to PRs that landed recently and bad review on my part.

When this lands, we'll need to make some config changes in -stage and -prod as well as the new infrastructure. I'll detail those changes in a followup comment and in the bug.

### DIFF
--- a/socorro/processor/breakpad_transform_rules.py
+++ b/socorro/processor/breakpad_transform_rules.py
@@ -61,7 +61,9 @@ class CrashingThreadRule(Rule):
 
 
 class ExternalProcessRule(Rule):
-
+    # FIXME(willkg): command_line and command_pathname are referenced in the
+    # uplifted versions in Processor2015. The rest of these config values have
+    # no effect on anything and are just here.
     required_config = Namespace()
     required_config.add_option(
         'dump_field',
@@ -189,6 +191,9 @@ class ExternalProcessRule(Rule):
 
 class BreakpadStackwalkerRule2015(ExternalProcessRule):
     """Executes the minidump stackwalker external process and puts output in processed crash"""
+    # FIXME(willkg): command_line and command_pathname are referenced in the
+    # uplifted versions in Processor2015. The rest of these config values have
+    # no effect on anything and are just here.
     required_config = Namespace()
     required_config.add_option(
         name='symbols_urls',
@@ -367,7 +372,9 @@ class BreakpadStackwalkerRule2015(ExternalProcessRule):
 
 
 class JitCrashCategorizeRule(ExternalProcessRule):
-
+    # FIXME(willkg): command_line and command_pathname are referenced in the
+    # uplifted versions in Processor2015. The rest of these config values have
+    # no effect on anything and are just here.
     required_config = Namespace()
     required_config.command_line = change_default(
         ExternalProcessRule,
--- a/socorro/processor/mozilla_transform_rules.py
+++ b/socorro/processor/mozilla_transform_rules.py
@@ -9,8 +9,6 @@ from sys import maxint
 import time
 from urllib import unquote_plus
 
-from configman import Namespace
-from configman.converters import str_to_python_object
 import ujson
 
 from socorro.external.postgresql.dbapi2_util import execute_query_fetchall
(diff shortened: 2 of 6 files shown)


socorro/processor/mozilla_transform_rules.py
```diff
@@ -600,26 +582,9 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
-    )
-    required_config.add_option(
-        'transaction_executor_class',
-        default="socorro.database.transaction_executor."
-                "TransactionExecutorWithInfiniteBackoff",
-        doc='a class that will manage transactions',
-        from_string_converter=str_to_python_object,
-        reference_value_from='resource.postgresql',
-    )
-
```
willkg: These two values come from Processor2015 config, so this is dead code. It's very confusing since it seems like it comes from here, but it doesn't anymore.

socorro/processor/processor_app.py
```diff
@@ -353,7 +351,7 @@ def _setup_source_and_destination(self):
 
         self.processor = self.config.processor.processor_class(
             self.config.processor,
-            self.quit_check
+            quit_check_callback=self.quit_check
```
willkg: Without this fix, the `self.quit_check` becomes the value of `rule_set` which prevents the processor from starting.

willkg:
After these changes, the processor works locally. We'll need to make config changes to -stage before we deploy this. Then we'll need to see if there are additional issues in -stage that we don't have in local dev.

socorro/processor/breakpad_transform_rules.py
```diff
@@ -61,7 +61,9 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
 
 
 class ExternalProcessRule(Rule):
-
+    # FIXME(willkg): command_line and command_pathname are referenced in the
+    # uplifted versions in Processor2015. The rest of these config values have
+    # no effect on anything and are just here.
```
Osmose: I don't understand this comment. I read it to mean "These config values are not used anywhere", but I see `dump_field` being used in both `ExternalProcessRule` and `BreakpadStackwalkerRule2015`. Am I misunderstanding?

socorro/processor/mozilla_transform_rules.py
```diff
@@ -446,70 +436,65 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
 
     def version(self):
         return '1.0'
 
     def _get_flash_version(self, **kwargs):
-        """If (we recognize this module as Flash and figure out a version):
-        Returns version; else (None or '')"""
+        """If we recognize this module as Flash and figure out a version
+
+        :returns: version; else (None or '')
```
Osmose: The returns was part of an oddly-structured sentence, not a line on it's own. Making it `:returns:` makes the first line make no sense.

socorro/processor/breakpad_transform_rules.py
willkg: It's so confusing. So `self.config.dump_field` in this class and subclasses actually refer to the value from `Processor2015`:

https://github.com/mozilla-services/socorro/pull/4350/files#diff-d49f4d6345a81fd5686209f8e3b86554L161

It doesn't refer to the `dump_field` option in this class or subclasses. This `dump_field` and the one in the subclasses don't have any effect on anything.

That's true because when these rules get instantiated here:

https://github.com/mozilla-services/socorro/pull/4350/files#diff-d49f4d6345a81fd5686209f8e3b86554L252

they get `config` for `Processor2015` passed to them which only contains keys/values defined in `Processor2015`.

I wanted to fix this but decided to toss in a FIXME for now and figure it out later.

socorro/processor/mozilla_transform_rules.py
willkg: Oops--you're right. I'll fix this and the commit summaries (they need the bug number) now.

willkg:
Thank you!

I'm doing the config discussion in the bug starting with https://bugzilla.mozilla.org/show_bug.cgi?id=1439928#c3 .

PR #4351, opened 2018-02-21
bug 1439928 - fix mdsw config values

willkg:
The default should be what we use in server environments--not what we use
in local dev environments. This fixes the various values scattered around
so there are fewer of them and they're correct in defaulty ways.

Also, we don't need `resource.signature.*` anything anymore.

### DIFF
--- a/docker/config/local_dev.env
+++ b/docker/config/local_dev.env
@@ -68,6 +68,16 @@ companion_process.symbol_cache_path=/tmp/symbols/cache
 processor.symbol_cache_path=/tmp/symbols/cache
 processor.symbol_tmp_path=/tmp/symbols/tmp
 
+# Drop kill_timeout to 30 because this is a dev environment and 5 minutes is
+# a long time
+processor.kill_timeout=30
+
+# Set symbols_urls to something helpful for local dev
+processor.symbols_urls=https://s3-us-west-2.amazonaws.com/org.mozilla.crash-stats.symbols-public/v1
+
+# Stackwalker is in a different place in the new infra and local dev
+processor.command_pathname=/stackwalk/stackwalker
+
 # Set the telemetry bucket name explicitly
 destination.telemetry.bucket_name=telemetry_bucket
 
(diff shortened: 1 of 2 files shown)


docker/config/local_dev.env
```diff
@@ -68,6 +68,16 @@ companion_process.symbol_cache_path=/tmp/symbols/cache
...
 
+# Drop kill_timeout to 30 because this is a dev environment and 5 minutes is
+# a long time
+processor.kill_timeout=30
+
+# Set symbols_urls to something helpful for local dev
+processor.symbols_urls=https://s3-us-west-2.amazonaws.com/org.mozilla.crash-stats.symbols-public/v1
+
+# Stackwalker is in a different place in the new infra and local dev
+processor.command_pathname=/stackwalk/stackwalker
```
willkg: Local dev overrides defaults which are what we use in -stage and -prod.

socorro/processor/processor_app.py
```diff
@@ -109,17 +109,6 @@
...
-        'processor_class': 'socorro.processor.processor_2015.Processor2015',
-
-        # These are for the minidump-stackwalker command line
-        'command_pathname': '/stackwalk/stackwalker',
-        'kill_timeout': 30,
-        'symbols_urls': ','.join([
-            'https://s3-us-west-2.amazonaws.com/org.mozilla.crash-stats.symbols-public/v1',
-        ]),
-    },
-
```
willkg: These behavior defaults overrode the existing defaults for local dev, but that's not what we want, yet. We want local dev values to be in `local_dev.env` and not scattered around. Everything else should default to -stage and -prod values.

socorro/processor/processor_app.py
```diff
@@ -144,10 +133,6 @@
             'filter_on_legacy_processing': True,
             'routing_key': 'socorro.normal',
         },
-
-        'signature': {
-            'collapse_arguments': True,
-        },
```
willkg: I changed signature generation to not use config at all, so this has been orphaned for months and can be removed.

willkg:
Thank you!

PR #4352, opened 2018-02-23
bug 1429563, 1440474, 1440471, 1440496 - clean up crontabber jobs

willkg:
This cleans up some of the crontabber app code and jobs that it runs.

The goal of this PR is to make it easier to deploy crontabber. Deploys involve killing off the existing crontabber node and then creating a new one.

Towards that goal, this does three things:

1. remove as many crontabber apps as possible
2. for crontabber apps that run daily (1d) or weekly (7d), set a specific time of day for them to run so they're unlikely to be running during possibly deploys
3. for the rest of the crontabber apps, make sure they can be interrupted and restarted ok

I'm pretty sure this PR covers all those three things as best as we can.

The one thing we don't fix here is that if crontabber is in the middle of a job and is killed, then there's a db lock that sticks around for 12 hours. I'm pretty sure we want to change that. Maybe 1 hour?


willkg:
Here's the config for the lock timeout:

https://github.com/mozilla/crontabber/blob/master/crontabber/app.py#L697

socorro/cron/jobs/clean_raw_adi.py
```diff
@@ -6,29 +10,44 @@
...
             """.format(self.config.days_to_keep)
         )
+
+        # Clean raw_adi_logs
+        cursor.execute(
+            """
+            DELETE FROM raw_adi_logs
+            WHERE report_date < NOW() - INTERVAL '{} days'
+            """.format(self.config.days_to_keep)
+        )
```
willkg: I merged this cron job and `CleanRawAdiLogsCronApp` or whatever it was called. They both run fast. They're effectively related. Might as well have one job rather than two.

socorro/unittest/cron/jobs/test_matviews.py
```diff
@@ -154,35 +153,6 @@ def mock_utc_now():
...
-        )
-
-        with config_manager.context() as config:
-            tab = CronTabberApp(config)
-            tab.run_all()
-
-            information = self._load_structure()
-            assert information['reports-clean']
-            assert not information['reports-clean']['last_error']
-            assert information['reports-clean']['last_success']
```
willkg: I'm not sure what this was testing, but if I remove `DuplicatesCronApp` from the list, then it fails. I figured I'd just nix it. It kind of seems like it's testing something in the crontabber library. If that's true, then we don't need to test that here--the library tests can deal with that.

socorro/cron/crontabber_app.py
```diff
@@ -19,26 +19,31 @@
...
 DEFAULT_JOBS_BASE = [
-    'socorro.cron.jobs.weekly_reports_partitions.WeeklyReportsPartitionsCronApp|7d',
-    'socorro.cron.jobs.matviews.ProductVersionsCronApp|1d|05:00',
-    'socorro.cron.jobs.matviews.SignaturesCronApp|1d|05:00',
-    'socorro.cron.jobs.matviews.RawUpdateChannelCronApp|1d|05:00',
+    # DB partition table and ES maintenance
+    'socorro.cron.jobs.weekly_reports_partitions.WeeklyReportsPartitionsCronApp|7d|06:00',
+    'socorro.cron.jobs.drop_old_partitions.DropOldPartitionsCronApp|7d|06:00',
+    'socorro.cron.jobs.truncate_partitions.TruncatePartitionsCronApp|7d|06:00',
+    'socorro.cron.jobs.elasticsearch_cleanup.ElasticsearchCleanupCronApp|7d|06:00',
```
willkg: Partitions go away once we stop saving crash data in postgres.

socorro/cron/crontabber_app.py
```diff
@@ -19,26 +19,31 @@
...
+
+    # ADI maintenance
+    'socorro.cron.jobs.clean_raw_adi.CleanRawADICronApp|1d|06:00',
     'socorro.cron.jobs.matviews.ADUCronApp|1d|08:30',
-    'socorro.cron.jobs.matviews.DuplicatesCronApp|1h',
-    'socorro.cron.jobs.matviews.ReportsCleanCronApp|1h',
-    'socorro.cron.jobs.bugzilla.BugzillaCronApp|1h',
     'socorro.cron.jobs.matviews.BuildADUCronApp|1d|08:30',
```
willkg: These go away when we stop doing ADI.

socorro/cron/crontabber_app.py
```diff
@@ -19,26 +19,31 @@
...
+
+    # Product/version maintenance
     'socorro.cron.jobs.ftpscraper.FTPScraperCronApp|1h',
-    'socorro.cron.jobs.elasticsearch_cleanup.ElasticsearchCleanupCronApp|7d',
-    'socorro.cron.jobs.drop_old_partitions.DropOldPartitionsCronApp|7d',
-    'socorro.cron.jobs.truncate_partitions.TruncatePartitionsCronApp|7d',
-    'socorro.cron.jobs.clean_raw_adi_logs.CleanRawADILogsCronApp|1d',
-    'socorro.cron.jobs.clean_raw_adi.CleanRawADICronApp|1d',
     'socorro.cron.jobs.featured_versions_automatic.FeaturedVersionsAutomaticCronApp|1h',
+    'socorro.cron.jobs.matviews.ProductVersionsCronApp|1d|05:00',
```
willkg: `FTPScraperCronApp` and `ProductVersionsCronApp` go away when we rewrite using buildhub.

socorro/cron/crontabber_app.py
```diff
@@ -19,26 +19,31 @@
...
+
+    # Crash data analysis
+    'socorro.cron.jobs.matviews.ReportsCleanCronApp|1h',
+    'socorro.cron.jobs.bugzilla.BugzillaCronApp|1h',
+    'socorro.cron.jobs.matviews.SignaturesCronApp|1d|05:00',
+    'socorro.cron.jobs.matviews.GraphicsDeviceCronApp|1d|05:00',
```
willkg: The matviews go away when we stop saving crash data.

willkg:
Bah--I messed up and forgot a commit. That last commit cf2b44d should go with this PR.

@Osmose Is that last commit ok?

willkg:
@Osmose said it's ok on IRC.

willkg:
Thank you!

PR #4354, opened 2018-02-26
bug 1440745 - move alt-adi apps to separate module

willkg:
Maintenance for `FetchADIFromHiveCronApp` is a nightmare right now. It
behooves us to keep it completely isolated from the rest of the codebase.
This moves things around to isolate it.


willkg:
It's just to move code around so I can copy-and-paste from scl3 into that `fetch_adi_from_hive.py` file without affecting anything that runs in -stage, -prod, or their new infra equivalents. This should be a cosmetic change.

willkg:
Thank you!

PR #4355, opened 2018-02-26
Downgrade CommonMark

willkg:
This needs to be version 0.5.4. This adds a pyup:ignore in the hope that that
prevents it from updating again.


willkg:
Self-r+!

PR #4356, opened 2018-02-26
bug 1440745 - add DryRunFetchADIFromHiveCronApp

willkg:
This PR adds a dry-run version of the `FetchADIFromHiveCronApp` as well as makes the changes for bug 1440745.

This is a work-in-progress.

### DIFF
--- a/socorro/cron/jobs/fetch_adi_from_hive.py
+++ b/socorro/cron/jobs/fetch_adi_from_hive.py
@@ -64,7 +64,10 @@ _QUERY = """
             domain='addons.mozilla.org' OR
             domain='blocklists.settings.services.mozilla.com'
         )
-        and http_status_code = '200'
+        and (
+            http_status_code='200' OR
+            http_status_code='304'
+        )
         and request_url like '/v1/blocklist/3/%%'
         and ds='%s'
     GROUP BY
--- a/socorro/cron/jobs/fetch_adi_from_hive.py
+++ b/socorro/cron/jobs/fetch_adi_from_hive.py
@@ -37,6 +37,7 @@ import pyhs2
 
 from configman import Namespace, class_converter
 from crontabber.base import BaseCronApp
+from crontabber.datetimeutil import utc_now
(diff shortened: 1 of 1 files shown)


socorro/cron/jobs/fetch_adi_from_hive.py
```diff
@@ -345,13 +352,37 @@ def run(self, date):
...
+
+class DryRunFetchADIFromHiveCronApp(FetchADIFromHiveCronApp):
+    """Dry run version of FetchADIFromHiveCronApp that doesn't persist any data
+
+    """
+    app_name = 'dry-run-fetch-adi-from-hive'
+    app_description = 'Dry Run Fetch ADI From Hive App'
+    app_version = '0.1'
+
+    required_config = Namespace()
```
willkg: I'm not positive we need this line, but I didn't want to spend 20 minutes verifying one way or the other.

socorro/cron/jobs/fetch_adi_from_hive.py
```diff
@@ -345,13 +352,37 @@ def run(self, date):
...
+
+    # Override the as_backfill_cron_app() decorator
+    _is_backfill_app = False
+
+    def main(self, function=None, once=True):
+        return BaseCronApp.main(self, function, once)
```
willkg: This explicitly calls the super-super class. The decorator is curious and stomps on the class method and I didn't want to deal with trying to figure out a valid `super` invocation. Instead, I went with this which definitely does what I want it to.

willkg:
The code changes in here were done in a way to be as minimal a change as possible and as easy to verify as possible. There are no tests that cover the original or new behavior, so the only way to test things is by doing it live in -prod.

That's a sucky dev situation, but we're getting rid of ADI soon, so I was loathe to spend a lot of time addressing the dev environment.

If we end up making more changes in the future, we can determine whether it makes sense to step back and improve the dev environment then.

willkg:
Thank you!

PR #4357, opened 2018-02-27
Fix bug 1441400: Replace custom exception sanitization processor.

Osmose:
Raven added a configurable processor for sanitizing error data in 6.4.0, which
removes the need for our own custom subclass. In addition, the internal API for
SanitizePasswordsProcessor changed such that our error reporting was broken.

I also added a few extra fields with potentially-sensitive data that we don't really need for error debugging. I'm not actually sure how much we care about CSRF tokens, but I figure it's better to be safe than sorry.

To test this:

1. Acquire a valid Sentry DSN. I used the one from staging. Add it to `my.env` as `RAVEN_DSN`.
2. Checkout the current master branch.
3. Run `docker-compose run webapp ./webapp-django/manage.py raven test`. Your output should look something like this:

   ```
   Starting socorro_memcached_1 ... done
   Starting socorro_elasticsearch_1 ...
   Starting socorro_rabbitmq_1 ...
   Starting socorro_postgresql_1 ...
   Starting socorro_postgresql_1 ... done
   Starting socorro_statsd_1 ... done
   StatsdMetrics configured: statsd:8125 None
   Client configuration:
     base_url       : https://sentry.prod.mozaws.net
     project        : 63
     public_key     : **********************
     secret_key     : **********************

   Sending a test message... Error in sys.excepthook:
     File "/usr/local/lib/python2.7/site-packages/raven/base.py", line 274, in handle_exception
       self.captureException(exc_info=exc_info, level='fatal')

   <snip>

     File "/app/webapp-django/crashstats/tokens/utils.py", line 9, in RavenSanitizeAuthTokenProcessor
       tuple(SanitizePasswordsProcessor.FIELDS) + ('auth-token',)
   TypeError: 'property' object is not iterable
   ```

4. Check out this branch. Run `docker-compose run webapp ./webapp-django/manage.py raven test` again. This time, the output should show a successful test:

   ```
   Starting socorro_memcached_1 ... done
   Starting socorro_rabbitmq_1 ...
   Starting socorro_postgresql_1 ... done
   Starting socorro_localstack-s3_1 ... done
   StatsdMetrics configured: statsd:8125 None
   Client configuration:
     base_url       : https://sentry.prod.mozaws.net
     project        : 63
     public_key     : ************************
     secret_key     : ************************
   
   Sending a test message... Event ID was 'ea2284f280674fbebac17b5889d7777b'
   Sentry is attempting to send 1 pending error messages
   Waiting up to 10 seconds
   Press Ctrl-C to quit
   ```
5. Check the Sentry instance you sent errors to. You should see an event marked as "Info" called "This is a test message generated using ``raven test``". You may also want to double check the date it was sent in case someone else is also doing some testing.


webapp-django/crashstats/settings/base.py
```diff
@@ -506,10 +506,16 @@ def path(*dirs):
     RAVEN_CONFIG = {
         'dsn': raven_dsn,
         'release': SOCORRO_REVISION,
+        # Defines keys to be sanitized by SanitizeKeysProcessor
+        'sanitize_keys': [
+            'sessionid',
+            'csrftoken',
+            'anoncsrf',
+            'sc',
```
willkg: What's "sc"?

willkg: @peterbe I don't know much about the webapp. Can you think of other keys that should be sanitized?

Osmose: I'm not entirely sure. In the context of Google Analytics it means "sessionControl", but here I just saw it set to some hashed or base64'd value and figured that we have no idea what it is and aren't using it currently, so we might as well assume it's sensitive and not collect it until we understand what it is.

peterbe: Yeah, what if you use an API token as a header. Will that be sanitized? E.g. 
`curl -H "Auth-Token: 18af2acef8a54dbca9580e2bb8ba9c9x" https://crash-stats.mozilla.com/api/RawCrash/?uuid=fb27d0f3-db8e-4835-9311-288bb0170829`

screen shot 2018-02-28 at 8 43 41 am

Perhaps temporarily but in a `raise Exception("Thanks obama!")` in one of the private API endpoints and see if Raven picks up the 'Auth-Token' header. 

Ideally you'd want to programmatically massage it. E.g.
```python
for key in request.headers:
  if key.lower() == 'auth-token':
    request.headers[key] = request.headers[key][:5] + '.' * 10
```
Then you'd be able to see "Yes, this user used an auth-token" and perhaps using your superuser access somehow figure out who that user belongs to `select * from auth_token where token like 'ad4510cd%';`

willkg: @peterbe I tossed that into my queue of things to look into. I'll write up a bug depending on what I find.

Osmose: Oh, I forgot to mention, I tested this. Auth-Token headers show up under the header list but are marked as `[Filtered]` and you can't see the actual token. Adding it to the list of filtered keys didn't change this behavior so I didn't include it.

willkg: @Osmose You rock! Thank you!

PR #4358, opened 2018-02-27
bug 1373997 - cleanup docs

willkg:
This is an overhaul of the docs I hadn't overhauled, yet.

Major changes:

1. I reduced the amount of stuff on the index page.
2. I added an Overview that I think summarizes Socorro good enough and provides a high-level overview of some basic stuff to make the rest easier to get through. I moved the diagram here and removed some bits from it to simplify it. This section could probably use some more work from someone who is not me.
3. I slimmed down the Getting Started chapter and moved some important bits to a "See also" section which should make them more prominent.
4. I removed tons and tons of obsolete stuff. I updated command lines. I removed a lot of "theoretical" things that aren't how we're running Socorro.
5. I added some basic stubs so that all our services and crash storage destinations are covered.
6. I improved some of the trouble-shooting sections. These sections could probably use some more work from someone who is not me.

I don't think we should spend a ton of time on this PR. It's way better than it was. I think we should push work to future PRs rather than hold this one up except in cases of typos, things that are definitely wrong or misleading, and FIXMEs that should get added.


lonnen:
r+

willkg:
Thank you!

PR #4359, opened 2018-02-28
Fix bug 1438708: Pull GA id from setings at runtime instead of build time.

Osmose:
Also removes some outdated GA settings that were once used for server-side
analytics, but are now unused.


webapp-django/crashstats/crashstats/static/crashstats/js/socorro/google_analytics.js
```diff
@@ -1,3 +1,3 @@
 window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
-ga('create', 'UA-XXXXX-X', 'auto');
+ga('create', document.documentElement.dataset.googleAnalyticsId, 'auto');
```
willkg: Out of curiosity, what happens if this node doesn't exist? Does the js error out, but otherwise the page works?

webapp-django/crashstats/settings/base.py
```diff
@@ -628,21 +623,6 @@ def path(*dirs):
...
-)
-
-# If calls to the Google Analytics API are done asynchronously, this
-# value can be quite high (5-10 seconds).
-GOOGLE_ANALYTICS_API_TIMEOUT = config(
-    'GOOGLE_ANALYTICS_API_TIMEOUT',
-    5,  # seconds
-    cast=int
-)
-
```
willkg: I think this is my fault. I ripped out a bunch of server-side GA code and I think I left this in by accident.

willkg:
I'm pretty sure this will work fine and doesn't require config changes in -stage/-prod. Does that sound correct?

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/google_analytics.js
Osmose: Yeah, I think so. I also don't think it's possible to have no `documentElement` in an HTML page (otherwise it'd mean JS was being executed before the HTML was being parsed).

Osmose:
> I'm pretty sure this will work fine and doesn't require config changes in -stage/-prod. Does that sound correct?

Correct, @milescrabill already confirmed that the `GOOGLE_ANALYTICS_ID` setting is set correctly for all our environments already, and this does not require any config changes to that setting.

PR #4362, opened 2018-03-02
fix bug 1440496 - remove ReportsCleanCronApp dependency on duplicates

willkg:


willkg:
Thank you for finding this!

PR #4368, opened 2018-03-05
Fix bug 1434495 by surfacing module cert info from crash reports

dblohm7:
We have added information about the certificate used to sign modules to Windows crash reports. We would like to surface that information in a report's module list displayed on crash-stats.

This PR adds the required processing to the stackwalker and modifies the webapp to add that information as a new column in the module list.

Regression tests have been added for the webapp parts.

These cert annotations have been available on Nightly 60 since March 2, 2018, so there should already be some reports submitted that contain the relevant annotations.

### DIFF
--- a/minidump-stackwalk/stackwalker.cc
+++ b/minidump-stackwalk/stackwalker.cc
@@ -129,6 +129,53 @@ static string ToInt(uint64_t value) {
   return buffer;
 }
 
+static vector<string> split(const string& s, char delimiter) {
+  vector<string> bits;
+  size_t prev = 0;
+  size_t pos = s.find(delimiter);
+  while (true) {
+    size_t count = pos == string::npos ? string::npos : pos - prev;
+    bits.push_back(s.substr(prev, count));
+    if (pos == string::npos) {
+      break;
+    }
+    prev = pos + 1;
+    pos = s.find(delimiter, prev);
+  }
+  return bits;
+}
+
+static bool startswith(const string& s, const string& f) {
+  return s.find(f) == 0;
+}
+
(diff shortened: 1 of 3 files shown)


willkg:
@luser Can you review the mdsw changes? I can look at the rest.

minidump-stackwalk/stackwalker.cc
```diff
@@ -575,7 +668,8 @@ int ConvertModulesToJSON(const ProcessState& process_state,
       main_module_index = module_sequence;
 
     Json::Value module_data;
-    module_data["filename"] = PathnameStripper::File(module->code_file());
+    string filename_leaf = PathnameStripper::File(module->code_file());
```
luser: It seems a little unfortunate to be keying this by just the module filename, but I would guess that in practice having multiple DLLs with the same filename loaded probably doesn't work anyway, or if it does is extremely unlikely to happen.

willkg:
I'll review the rest of this on Monday.

willkg:
Looks like this change adjusts stackwalker to pull out module cert info from the minidump and then displays that on the report index page.

Is the client side of this in the current nightly? If I process a crash from a recent nightly, will I be able to see these changes?

dblohm7:
Yes, these annotations have been available in crash reports for Windows Nightly since March 2.

minidump-stackwalk/stackwalker.cc
dblohm7: Yeah it doesn't work.

dblohm7:
Rebased.

willkg:
Thank you!

PR #4371, opened 2018-03-06
fix bug 1440712 - add "upload_telemetry_schema" app

willkg:
This app can be run using the `socorro-cmd` in an appropriate environment for
uploading the Telemetry schema to the Telemetry S3 bucket.

### DIFF
--- a/socorro-cmd
+++ b/socorro-cmd
@@ -45,7 +45,7 @@ def showcommands_cmd(argv):
         for cmd, runner in group:
             if not isinstance(runner, basestring):
                 runner = '%s in %s' % (runner.__name__, inspect.getfile(runner))
-            print('  %-22s %s' % (cmd, runner))
+            print('  %-24s %s' % (cmd, runner))
         print('')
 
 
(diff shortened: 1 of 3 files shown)


willkg:
To test this:

1. Get into a bash shell in the docker environment
2. Run:
   ```
   you@processor:/app$ env telemetry.bucket_name=telemetry_bucket socorro-cmd upload_telemetry_schema
   ```
3. Then check the telemetry bucket for the file:
   ```
   you@processor:/app$ ./scripts/socorro_aws_s3.sh ls --recursive s3://telemetry_bucket/
   ```

That's it!

Two things:

1. I wasn't sure how to write a test for this with our current testing infrastructure for s3 things. We could do something like with the existing cron app, but it's pretty mocked out and I'm not sure how useful that is. It would tell us whether the file is valid Python and whether it does something, though. Any better ideas?
2. This requires setting the `telemetry.bucket_name` configuration variable.

willkg:
You need to create the bucket first. One way to do that is to run `make dockersetup` which will create buckets that you could use.

I'm rewriting the boto tests to use moto and then I can do a real test for this.

Osmose:
`dockersetup` fixed things, thanks! I can verify the manual test here now.

willkg:
^^^ That implements tests (yay moto!) and also niceifies the "bucket not found" situation. This should be all set to go now.

willkg:
Thank you!

PR #4374, opened 2018-03-07
fix bug 1441318 - fix crontabber status error

willkg:
If one of the jobs has a `last_run` of `null` (it was running for the first time), then `/monitoring/crontabber` would error out with a `TypeError`. This fixes that by changing that code to only look at non-None `last_run`.


willkg:
Thank you!

PR #4378, opened 2018-03-09
bug 1443981 - rewrite boto tests to use moto

willkg:
This rewrites most of the boto tests to use moto. It doesn't rewrite the tests for:

* `UploadCrashReportJSONSchemaCronApp` -- that's going away
* connection context tests -- these test connection parameters and some other things that moto doesn't make easier

This also centralizes some code that we can probably hoist higher in the test system, but it's ok where it is for now.

This gets rid of `SupportReasonAPI*` stuff--we don't use that.

This gets rid of the tests for S3 crash storage with `TransactionExecutorWithLimitedBackoff` because those tests are really testing `TransactionExecutorWithLimitedBackoff` and that's tested elsewhere.

This adds tests for all the `get_*` methods for the case when the requested files aren't there.

Why do this now? Using `@mock_s3_deprecated` will make it much easier to update from boto to boto3. The existing code mocks out boto API internals. Updating to boto3 would require us to rewrite all that. It'd be hard. We want to switch to boto3 soon and this is the first step towards that.


requirements/constraints.txt
```diff
@@ -176,3 +176,107 @@ testfixtures==5.3.1 \
...
+    --hash=sha256:a453dc4dfa6e0db3d8fd7738a308a88effe6240c59f3226eb93e8f020c216149
+pycparser==2.18 \
+    --hash=sha256:99a8ca03e29851d96616ad0404b4aad7d9ee16f25c9f9708a11faf2810f7b226
+wrapt==1.10.11 \
+    --hash=sha256:d4d560d479f2c21e1b5443bbd15fe7ec4b37fe7e53d335d3b9b0a7b1226fe3c6
+backports.ssl_match_hostname==3.5.0.1 \
+    --hash=sha256:502ad98707319f4a51fa2ca1c677bd659008d27ded9f6380c79e8932e38dcdf2
+backports.weakref==1.0.post1 \
+    --hash=sha256:81bc9b51c0abc58edc76aefbbc68c62a787918ffe943a37947e162c3f8e19e82 \
+    --hash=sha256:bc4170a29915f8b22c9e7c4939701859650f2eb84184aee80da329ac0b9825c2
```
willkg: This is the worst part of moto--it's got a lot of dependencies. Some of these will go away when we upgrade to Python 3. Still, ick.

socorro/external/boto/crashstorage.py
```diff
@@ -391,46 +391,3 @@ def _do_save_processed(boto_connection, processed_crash):
...
-            'support_reason',
-            json.dumps(content)
-        )
-
-    def save_raw_crash(self, raw_crash, dumps, crash_id):
-        """There are no scenarios in which the raw crash could possibly be of
-        interest to the Support Reason API, therefore this function does
-        nothing; however it is necessary for compatibility purposes.
-        """
-        pass
```
willkg: We don't use this anywhere, so we should nix it.

socorro/unittest/external/boto/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+
+    cm = ConfigurationManager(
+        [conf],
+        app_name='testapp',
+        app_version='1.0',
+        app_description='TelemetryBotoS3CrashStorage test',
+        values_source_list=[values_source],
+        argv_source=[],
+    )
+    return cm.get_config()
```
willkg: We had a bunch of code that does something along these lines, but did it worse. In most cases, they built the config object manually which requires you to specify everything instead of building it from the `required_config` of the class which will use the defaults. I'd much rather do the latter since as we change defaults, we can see how it affects our tests. Further, this reduces the amount of configuration code we've got floating around.

It's possible something like this exists in the tests already. If not, we probably want to hoist it upwards and use it everywhere.

socorro/unittest/external/boto/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+
+    conf = cls.get_required_config()
+    conf.add_option('logger', default=mock.Mock())
+    conf.add_option('metrics', default=mock.Mock())
+
+    cm = ConfigurationManager(
+        [conf],
+        app_name='testapp',
+        app_version='1.0',
+        app_description='TelemetryBotoS3CrashStorage test',
```
willkg: Oops--need to fix this.

socorro/unittest/external/boto/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+            bucket = conn.get_bucket(bucket_name)
+        return bucket
+
+    def put_object(self, bucket_name, key, value):
+        """Puts an object into the specified bucket"""
+        bucket = self.get_or_create_bucket(bucket_name)
+        key_object = bucket.new_key(key)
+        key_object.set_contents_from_string(value)
+
+    def fetch_object(self, bucket_name, key):
```
willkg: Oops--I meant to change this method name.

socorro/unittest/external/boto/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+        """Gets or creates the crashstats bucket"""
+        conn = boto.connect_s3()
+        try:
+            bucket = conn.get_bucket(bucket_name)
+        except StorageResponseError:
+            conn.create_bucket(bucket_name)
+            bucket = conn.get_bucket(bucket_name)
+        return bucket
+
+    def put_object(self, bucket_name, key, value):
```
willkg: Oops--I meant to change this method name.

socorro/unittest/external/boto/test_connection_context.py
```diff
@@ -53,40 +48,31 @@ class ConditionallyABadDeal(Exception):
...
-            'logger': mock.Mock(),
-            'host': host,
-            'port': port,
-            'access_key': 'this is the access key',
-            'secret_access_key': 'secrets',
-            'bucket_name': 'silliness',
-            'keybuilder_class': KeyBuilderBase,
-            'prefix': 'dev',
-            'calling_format': mock.Mock()
-        })
```
willkg: Using `get_config` lets us not have to specify values for all the defaults. This makes our tests much better.

socorro/unittest/external/boto/test_connection_context.py
```diff
@@ -53,40 +48,31 @@ class ConditionallyABadDeal(Exception):
...
+    s3_conn._mocked_connection = s3_conn._connect_to_endpoint.return_value
+    s3_conn._calling_format.return_value = mock.Mock()
+    s3_conn._CreateError = mock.Mock()
+    s3_conn.ResponseError = mock.Mock()
+    s3_conn._open = mock.MagicMock()
+
+    return s3_conn
 
+
+class TestConnectionContext:
```
willkg: The rest of the changes in this file are mostly cosmetic:

* connection_source -> conn
* reducing some vertical stuff into single lines
* changing classes so they don't inherit from `unittest.TestCase`

socorro/unittest/external/boto/test_crashstorage.py
```diff
@@ -59,86 +40,45 @@ class ConditionallyABadDeal(Exception):
...
-
-    @classmethod
-    def setUpClass(cls):
-        super(BaseTestCase, cls).setUpClass()
-        cls.TEMPDIR = tempfile.mkdtemp()
-
-    @classmethod
-    def tearDownClass(cls):
-        super(BaseTestCase, cls).tearDownClass()
-        shutil.rmtree(cls.TEMPDIR)
```
willkg: I converted the classes so they aren't subclasses of `unittest.TestCase` which means we can use the pytest `tmpdir` fixture which was nice.

socorro/unittest/external/boto/test_crashstorage.py
```diff
@@ -59,86 +40,45 @@ class ConditionallyABadDeal(Exception):
...
+        "aws_secret_access_key": boto_s3_store.config.secret_access_key,
+        "is_secure": True,
+        "calling_format": (
+            boto_s3_store.connection_source._calling_format.return_value
+        ),
+    }
+    _connect_to_endpoint = (
+        boto_s3_store.connection_source._connect_to_endpoint
+    )
+    _connect_to_endpoint.assert_called_with(**kwargs)
```
willkg: Oops--we can nix this, too.

socorro/unittest/external/boto/test_crashstorage.py
```diff
@@ -193,126 +133,82 @@ def _fake_unredacted_processed_crash_as_string(self):
         s = json.dumps(d)
         return s
 
-    def test_save_raw_crash_1(self):
-        boto_s3_store = self.setup_mocked_s3_storage()
+    @mock_s3_deprecated
```
willkg: This file might be easier to read through by just looking at the file and not what changed.

socorro/unittest/external/boto/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def list(self, bucket_name):
+        """Lists contents of bucket and returns keys as strings"""
+        bucket = self.get_or_create_bucket(bucket_name)
+        return [key.key for key in bucket.list()]
+
+
+@pytest.fixture
+def boto_helper():
+    return BotoHelper()
```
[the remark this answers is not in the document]
willkg: That makes sense. I tend to make things fixtures that are part of the test infrastructure because it makes it easier to re-use elsewhere and it's easier to discover in large test suites (`pytest --fixtures`). Oops--I should add a docstring.

willkg:
Thank you!

PR #4383, opened 2018-03-14
fixes bug 1425245 - adds crash_verify/ url endpoint

willkg:
This url endpoint takes a crash id and tells you whether the data for that
crash exists in all the appropriate crash stores.

We can use this for end-to-end integration testing as well as debugging
system oddities.


willkg:
Two things:

1. This was a total slog. In order to implement this, I practically had to rewrite how we test implementations. Tons of fixes in order to get it working. I'll annotate those later.
2. This is easy to test!

To test:

1. Run unit tests--they covers most of the situations.
2. Run the webapp.
3. Go to `http://localhost:8000/api/CrashVerify/` and you'll get a bad crash id error.
4. Go to `http://localhost:8000/api/CrashVerify/?crash_id=foo` and you'll get a bad crash id error.
5. Go to `http://locahost:8000/api/CrashVerify/?crash_id=83a263f7-0d20-4b4a-9fbc-0f6840180314`. Since that crash doesn't exist in your local crash stores, it'll say "False" for all the things.
6. Process a crash and then look that crash id up. It should exist in all the places.

This only looks at S3 buckets and Elasticsearch. It doesn't look at postgres. That crash store is not long for this world and implementing that was tough, so I didn't bother doing it.

socorro/external/boto/crashstorage.py
```diff
@@ -391,3 +390,17 @@ def _do_save_processed(boto_connection, processed_crash):
             "crash_report",
             processed_crash_as_string
         )
+
+    @staticmethod
+    def _do_get_unredacted_processed(boto_connection, crash_id, json_object_hook):
+        """Overriding this to change "name of thing" to crash_report"""
+        try:
+            processed_crash_as_string = boto_connection.fetch(crash_id, 'crash_report')
```
Osmose: I wonder if the fact that we're doing this twice means we should just make it a class attribute that can be overridden?

willkg: This class creates a new data type to store, but the way it does it is this override nonsense. I spent some time thinking about it while working on this PR.

There's another bug for upgrading to boto3. boto3 has a different API than boto, so I was thinking when we did that, it'd be a good opportunity to collapse all the connection contexts into one, nix the keybuilder code, and fix crash data types which should fix the Telemetry crash storage and make it less hacky.

I think that rewrite is in the near future.

docker/config/local_dev.env
```diff
@@ -101,6 +101,7 @@ RABBITMQ_VIRTUAL_HOST=rabbitvhost
 RAVEN_DSN=
 SECRET_KEY=secretkey
 STATSD_HOST=statsd
+TELEMETRY_BUCKET_NAME=telemetry_bucket
```
willkg: Previously, the webapp blithely ignored the telemetry bucket. Now it needs to know, so I added a variable. We'll need to add this to the -stage, -prod, -new-stage, -new-prod, and other environments as well.

docker/config/test.env
```diff
@@ -20,3 +20,7 @@ database_password=aPassword
 
 # Django tests are meant to run with debug mode disabled.
 DEBUG=False
+
+# Use test buckets
+resource.boto.bucket_name=crashstats-test
+TELEMETRY_BUCKET_NAME=telemetry-test
```
willkg: OMG. Test settings in the webapp are complicated. There are some settings here and some in `webapp-django/crashstats/settings/test.py`. I don't think these two have any effect on things, but I can't be sure. I'd much rather have one set of settings, but the ones in the `test.py` file override a bunch of other stuff that we can't set via the environment.

Anyhow, so I added these here, but they might not affect anything and at some point, we should revisit webapp test configuration.

socorro/external/boto/crash_data.py
```diff
@@ -79,3 +83,34 @@ def get(self, **kwargs):
...
+
+        try:
+            return self.get_unredacted_processed(params.uuid)
+        except CrashIDNotFound as cidnf:
+            self.config.logger.error('telemetry crash not found: %s' % cidnf)
+            # The CrashIDNotFound exception that happens inside the
+            # crashstorage is too revealing as exception message contains
+            # information about buckets and prefix keys. Re-wrap it here so the
+            # message is just the crash ID.
+            raise CrashIDNotFound(params.uuid)
```
willkg: This pretty much mirrors the `SimplifiedCrashData` one. We don't really need the data from the telemetry bucket right now. Mostly we just need to know if the crash is in there. So it's possible this is overkill.

socorro/external/boto/crashstorage.py
willkg: Having said that, if you think switching it to a class attribute is easy to do, I'd be game for doing it now. Less hacky is less hacky.

socorro/external/es/base.py
```diff
@@ -4,12 +4,10 @@
...
 import datetime
 
-from configman import Namespace, class_converter
+from configman import Namespace, class_converter, RequiredConfig
 
-from socorro.app.socorro_app import App
 
-
-class ElasticsearchConfig(App):
+class ElasticsearchConfig(RequiredConfig):
```
willkg: Yeah... so this used to subclass `App`. The problem with that is if you don't mock out the Elasticsearch implementation for the crashstats models, then it tries to configure this and this picks up the configuration from `App` which includes a logger set up to talk to the syslog service and there isn't one when the tests are running and the resulting traceback reads like a chapter from a Chuck Palhuniak book that was deleted because it was just too cray-cray.

I don't think anything should be subclassing `App` unless it's an app, so I switched it to subclass `RequiredConfig` instead. That made some of my problems go away.

socorro/external/es/super_search_fields.py
```diff
@@ -4308,7 +4308,7 @@ def get(self):
         'form_field_type': None,
         'has_full_version': False,
         'in_database_name': 'uuid',
-        'is_exposed': False,
+        'is_exposed': True,
```
willkg: This allows one to do a super search and filter on uuid. I talked to @adngdb and he said it should be fine. It wasn't exposed because it's pretty silly, but it made writing the CrashVerify endpoint a lot easier, so I exposed it.

webapp-django/crashstats/base/tests/testbase.py
```diff
@@ -35,15 +35,22 @@ def setUp(self):
...
             # The class name is used by the internal caching that guards
             # for repeated calls with the same parameter.
             # If we don't do this the cache key will always have the
             # string 'MagicMock' in it no matter which class it came from.
             klass.implementation().__class__.__name__ = klass.__name__
 
+    def tearDown(self):
+        for klass in classes_with_implementation:
+            klass.implementation = self._mockeries[klass]
+
```
willkg: Previously, the `DjangoTestCase` would mock everything in `setUp` and then ... well, that was it. So the problem is that all those things stay mocked and if you run tests in a different order and one of them didn't expect them to be mocked, then it might fail or it might pass. I had one test that when run on its own worked fine, but when run with other tests, failed miserably with weird errors.

I fixed this by adding a `tearDown` that unmocks things. I suspect there are easier ways to do this, but I was trying hard to touch as little as I could in order to get my stuff working.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -79,11 +79,17 @@ def config_from_configman():
...
+    definition_source.namespace('crashdata')
+    definition_source.crashdata.add_option(
         'crash_data_class',
         default=socorro.external.boto.crash_data.SimplifiedCrashData,
     )
+    definition_source.namespace('telemetrydata')
+    definition_source.telemetrydata.add_option(
+        'telemetry_data_class',
+        default=socorro.external.boto.crash_data.TelemetryCrashData,
+    )
```
willkg: This adds the configuration for the telemetry crash data class.

webapp-django/crashstats/crashstats/utils.py
```diff
@@ -347,7 +347,7 @@ def ratelimit_rate(group, request):
     Otherwise return a number according to
     https://django-ratelimit.readthedocs.org/en/latest/rates.html#rates-chapter
     """
-    if group == 'crashstats.api.views.model_wrapper':
+    if group in ('crashstats.api.views.model_wrapper', 'crashstats.api.views.crash_verify'):
```
willkg: I needed to do this otherwise `crash_verify` wouldn't work with rate limiting.

Having said that, I'm not positive I understand what this code is supposed to do. It's curious.

webapp-django/crashstats/supersearch/tests/common.py
```diff
@@ -3,253 +3,6 @@
 from crashstats.crashstats.tests.test_models import Response
 
 
-SUPERSEARCH_FIELDS_MOCKED_RESULTS = {
```
willkg: A while back, @peterbe and @adngdb did some code changes to move super search fields from being stored in Elasticsearch to being stored in a giant JSON file in the repository. We had to back that out because it was tied to upgrading Elasticsearch. Then either @peterbe or I re-implemented it as Python code.

Anyhow, long story short, we needed to mock it out before because it was stored in Elasticsearch and each environment had their own copy and we needed something for the tests. But when we moved it to Python-land, we should have nixed all that code because mocking it wasn't necessary anymore.

Anyhow, this doesn't include the `uuid` field I just exposed, so I either had to add that field to the mockery or remove the mockery. I chose the latter because that seemed like a better end result.

willkg:
Annotated!

socorro/external/boto/crashstorage.py
Osmose: Nah, it doesn't pass the "will be rewritten shortly" test.

webapp-django/crashstats/base/tests/testbase.py
willkg: I have a hell of a time using pytest fixtures with `unittest.TestCase` derived classes. I have no idea how it'd work with subclassing and method overriding. Otherwise, I like idea a lot!

willkg:
Thank you!

PR #4385, opened 2018-03-19
#1436070, #1436069: Remove inlined CSS, fix semantic markup

andymikulski:
- Fixes [#1436070 (non-semantic markup)](https://bugzilla.mozilla.org/show_bug.cgi?id=1436070)
- Fixes [#1436069 (inlined CSS)](https://bugzilla.mozilla.org/show_bug.cgi?id=1436069)

**Notes:**
- Instead of replacing all of the bold markup with something like `<span class="bold">`, I decided to simply unify all of the markup to use the same element - `b`.
  - It turns out that some of the javascript looks for `b` tags when operating. The compromise was to simply unify the markup that's used for bold text so the styles at least stem from the same definition.
- A couple of files missed the CSS->LESS conversion, those are included here ase they were affected by the `strong/b/.bold` change
- Inlined styles are replaced with classes. The `debug_login.less` file was created as a result, which is included.

**To review:**
- Build the site, navigate the site, double check the LESS. Should be pretty straight-forward.


webapp-django/crashstats/api/static/api/css/documentation.less
```diff
@@ -1,24 +1,36 @@
-p.url {
-    margin: 15px;
-    font-size: 1.6em;
-}
-p.url code {
-    padding-left: 25px;
+@import "@{root-path}/base/variables.less";
+
+p {
+    &.url {
```
Osmose: This could also just be `p.url` since it's the only thing in the block. Is this an explicit style choice?

I'm not against it or anything, I just noticed it and was curious.

webapp-django/crashstats/base/jinja2/crashstats_base.html
```diff
@@ -176,8 +176,7 @@ <h1>Product Navigation</h1>
                     {% csrf_token %}
                     <div
                         id="signin2"
-                        class="google-signin"
-                        style="float:right"
+                        class="google-signin float_right"
```
Osmose: Could we maybe just apply the float to the `google-signin` class, or add a different class? `float_right` isn't a terribly semantic classname.

webapp-django/crashstats/profile/jinja2/profile/profile.html
```diff
@@ -34,7 +34,7 @@ <h2>Your Profile</h2>
             </ul>
 
             <!-- clear after two floats -->
-            <div style="clear: both;"></div>
+            <div class="clear"></div>
```
Osmose: Could we use something like [this clearfix](https://css-tricks.com/snippets/css/clear-fix/) to have the container clear the links and remove the need for this element altogether?

webapp-django/crashstats/api/static/api/css/documentation.less
andymikulski: I think I just converted the code in the file from top-to-bottom and simply didn't notice that this selector didn't contain much. I've collapsed it down to `p.url`.

andymikulski:
>I can't figure out why that test error is resulting from your changes, ...

The issue was a static file (`analyze-model-fetches.css`) was being referenced, however I changed the extension to `.less`. I've changed the entire `analyze-model-fetches.html` file to use bundles, which should resolve the issue. Notably, this is also the last of the JS/LESS that was _not_ already in bundles!

I've corrected the code review items and pushed the new bundle, so this should now properly pass CI.

andymikulski:
@Osmose This is all set, though I am unable to merge!

screen shot 2018-04-02 at 4 33 25 pm

Osmose:
Huh, I thought I granted you access. Weird. I'll see if I can fix that.

Osmose:
Yeah, you totally have write access:

screen shot 2018-04-02 at 6 20 02 pm

wtf?

PR #4387, opened 2018-03-20
fix bug 1447416 - JSON responses should have application/json mimetype

willkg:
This fixes the `/__version__` and new `/api/CrashVerify/` endpoints to return `application/json` as the mimetype.


willkg:
Nixing this for the other one.

PR #4388, opened 2018-03-20
fix bug 1373997 - more docs fixes

willkg:
This fixes some outstanding docs issues:

* fixes e2e-tests readme in the sphinx docs: switches the way we include markdown docs because the previous way wouldn't work with readthedocs
* converts schema docs from Markdown to restructued text and then adds those to the Sphinx docs

After this, we're done with the docs overhaul. Hooray!


willkg:
Whoops! I'm always doing that.

^^^ Added the missing file and rebased against MASTER.

willkg:
Thank you!

PR #4389, opened 2018-03-26
Reduce amount of staticfiles being collected.

Osmose:
From 1500 or so files being collected, copied, and post-processed, down to 248. I left some projects with wildcards since they had so few files as to not be an issue.


Osmose:
> This seems fine to me, but I'm not sure I would notice if files were missing.

Yeah, my original idea of this reducing build time was flawed, the copied files aren't being compiled like I thought they were.

PR #4390, opened 2018-03-30
fix bug 1146030 - improve missing product page

willkg:
If you go to `/home/product/<PRODUCTNAME>` and that product doesn't exist, it
would kick up an unhelpful 404. Now it kicks up a 404, but with a bit more
text.

### DIFF
--- a/webapp-django/crashstats/crashstats/tests/test_views.py
+++ b/webapp-django/crashstats/crashstats/tests/test_views.py
@@ -510,8 +510,7 @@ class TestViews(BaseTestViews):
         url = reverse('home:home', args=('Unknown',))
         response = self.client.get(url)
         assert response.status_code == 404
-        assert 'Page Not Found' in response.content
-        assert 'id="products_select"' not in response.content
+        assert 'Missing product' in response.content
 
     def test_handler404_json(self):
         # Just need any view that has the json_view decorator on it.
--- a/webapp-django/crashstats/crashstats/tests/test_views.py
+++ b/webapp-django/crashstats/crashstats/tests/test_views.py
@@ -507,10 +507,9 @@ class TestViews(BaseTestViews):
             assert result['query_string'] is None
 
     def test_handler404(self):
-        url = reverse('home:home', args=('Unknown',))
-        response = self.client.get(url)
+        response = self.client.get('/fillbert/mcpicklepants')
         assert response.status_code == 404
-        assert 'Missing product' in response.content
+        assert 'The requested page could not be found.' in response.content
 
     def test_handler404_json(self):
         # Just need any view that has the json_view decorator on it.
--- a/webapp-django/crashstats/home/jinja2/home/missing_product.html
+++ b/webapp-django/crashstats/home/jinja2/home/missing_product.html
@@ -1,0 +1,24 @@
+{% extends "crashstats_base.html" %}
+{% stylesheet 'home' %}
+
+{% block page_title %}
+Missing product: {{ product }}
+{% endblock %}
+
+{% block content %}
+<div id="mainbody">
+    <div class="page-heading">
+        <h2 id="homepage-heading">Missing product: {{ product }}</h2>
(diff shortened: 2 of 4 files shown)


willkg:
To test:

1. `make dockersetup`
2. `make dockerupdatedata`
3. go to `http://localhost:8000/home/product/Foo` and it'll show a slightly more helpful page

I wanted this page because we see it if we try to go to `/home/product/Firefox` without product version data in the db. Thus I hit it periodically. I assume that new contributors/devs will also hit this. I figure it helps to be a little clearer.

The thing I don't know is whether there's other information we could provide here that's helpful, but still minimal.

Thoughts?

willkg:
Also, if it's the case that this isn't helpful at all, I'm game for nixing it and WONTFIXing bug 1146030, too. That's also a fine outcome here.

webapp-django/crashstats/home/views.py
```diff
@@ -1,10 +1,33 @@
...
 from django.conf import settings
+from django.http import Http404
 from django.shortcuts import render
 from django.views.generic.base import RedirectView
 
 from crashstats.crashstats.decorators import pass_default_context
+from crashstats.crashstats.utils import build_default_context
+
+
+def handle_missing_product(view):
```
Osmose: Why only fix this for the homepage (or, rather, only views using this decorator), instead of for any view using `pass_default_context`?

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -510,8 +510,7 @@ def test_handler404(self):
         url = reverse('home:home', args=('Unknown',))
         response = self.client.get(url)
         assert response.status_code == 404
-        assert 'Page Not Found' in response.content
-        assert 'id="products_select"' not in response.content
+        assert 'Missing product' in response.content
```
Osmose: I think this test is meant to test the generic 404 handler and just happened to be using the homepage. We should probably change the URL to something more reliably returning a 404, and add a separate test somewhere for the new 404 page.

webapp-django/crashstats/home/jinja2/home/missing_product.html
```diff
@@ -0,0 +1,24 @@
...
+{% stylesheet 'home' %}
+
+{% block page_title %}
+Missing product: {{ product }}
+{% endblock %}
+
+{% block content %}
+<div id="mainbody">
+    <div class="page-heading">
+        <h2 id="homepage-heading">Missing product: {{ product }}</h2>
```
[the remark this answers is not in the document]
willkg: I left it in because I thought it might be used in the e2e-tests, but it's not. I'll nix it.

webapp-django/crashstats/home/views.py
willkg: The bug (and the problem I wanted to tackle) is specifically about a fresh Socorro instance missing a product. I think you're likely to hit that when going to `/home/product/XXX` and not other places.

`pass_default_context` adds information to the context to handle the nav bar more than anything else. I think I'd rather those pages continue to do what they're doing now.

So for those reasons, I hedged towards just fixing the home view and not affecting anything else.

Relatedly, we've talked about redoing Socorro navigation and creating a real page that sits at `/` and lists the products available rather than automatically redirecting to `/home/product/XXX`. I think that's a more involved fix and something that's better to push off until after the photon overhaul. When we do that, we can undo the fix I did in this PR because it won't be helpful anymore.

webapp-django/crashstats/crashstats/tests/test_views.py
willkg: Oh, that's a good point. I definitely broke this. I'll fix it.

willkg:
@Osmose ^^^ That should fix the issues you raised.

willkg:
@Osmose ^^^ That adds the missing test.

willkg:
Thank you!

PR #4391, opened 2018-03-30
Add two new histograms for elasticsearch

sciurus:
This adds two new histograms, which track the count and timing of
successful and failed saves of crashes to elasticsearch.

The existing histogram for
processor.es.ESCrashStorageRedactedJsonDump.save_raw_and_processed
encompasses a chunk of python code and potentially multiple calls to
ES. This new metrics will get us just the time spent talking to ES and
let us distinguish between successful and failed calls.

This should help us understand the performance of the new
infrastructure.

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -7,6 +7,7 @@ import re
 from threading import Thread
 from Queue import Queue
 from contextlib import contextmanager
+from datetime import datetime
 
 import elasticsearch
 from configman import Namespace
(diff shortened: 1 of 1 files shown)


willkg:
Mmm... The whole `self.config.metrics` thing is a hack and I'm not sure it's available in all the places it needs to be to do this nor am I sure that the tests will tell us whether something is missing.

This is the bug I was talking about to redo metrics in the processor:

https://bugzilla.mozilla.org/show_bug.cgi?id=1412590

How urgent is it that we look into the ES thing? We don't have these metrics in -prod, so we have nothing to compare the numbers to. Are we thinking that the ES numbers are indicative that something is wrong? If so, we should open up a bug and then tie this change to that bug.

sciurus:
I think ES performance level is fine, and if we want to put off looking at it we can. I just thought I'd take a crack at this since I was already reading the code. When we do want to look at it, I think we should start with this data.

willkg:
That makes sense! We can figure this out next week.

willkg:
I'm going to close this out. I mentioned to @milescrabill that we should talk about ES and what the plan is a the All Hands next week. That conversation should include "do we want to be measuring anything else about ES and if so, what?" Then we can work through that.

It should be a lot easier to do that now that we have a better metrics infrastructure.

Sorry for leaving this around for so long!

PR #4392, opened 2018-03-31
bug 1442148 - tweak crontabber state page

willkg:
This fixes display of next run and last success so that it shows actual dates
and times rather than than a delta between then and now encoded in vague
English prose.

This also tweaks the sankey links so the lines have a fixed width making the
graph legible (at least, on my local machine). I think to fix this further
requires someone to really get to know sankey (which has no docs) or switch
libs or roll our own.


willkg:
Thank you!

PR #4399, opened 2018-04-09
Bug 1452667 - change number of threads to 4

willkg:
We're running processor nodes with 4 virtual CPUs. Since worker threads
are in Python, they run on the same CPU. However, they run stackwalker
which is a separate process, so the number of threads dictates the
number of simultaneous stackwalker processes. Stackwalker is CPU-heavy.

This sets the number of threads equal to the number of CPUs by default.

### DIFF
--- a/socorro/processor/processor_app.py
+++ b/socorro/processor/processor_app.py
@@ -111,7 +111,7 @@ CONFIG_DEFAULTS = {
 
     'producer_consumer': {
         'maximum_queue_size': 32,
-        'number_of_threads': 16,
+        'number_of_threads': 4,
     },
 
     'resource': {


sciurus:
I didn't verify in the code that the threads are where stackwalker is run, but it does seem likely based on the number we've seen running.

Should the maximum_queue_size also be adjusted down? The suggestion below is for twice the number of threads.

https://github.com/mozilla-services/socorro/blob/778dbb949049f7e7c18446aa55711377b6b17304/socorro/lib/threaded_task_manager.py#L29-L45

willkg:
That advice is general and not about using RabbitMQ as the source. If the processor takes a crash id from the queue and then the processor dies (say we do a deploy), that crash id isn't lost forever since it was never acknowledged.

Having said that, I don't have any better ideas on setting `maximum_queue_size`, so I'm game for setting it to 8.

Good catch!

PR #4400, opened 2018-04-10
Add data-no-csrf to API testdrive GET forms to fix ZAP baseline

g-k:
data-no-csrf tells the ZAP baseline scan to ignore GET-only search endpoints 

I haven't filed a bug for this.

Functional Test:

- [x] start up the webapp
- [x] go to /api
- [x] check that all GET-only testdrive forms have the data-no-csrf attribute:

screen shot 2018-04-10 at 1 46 42 pm


willkg:
I'll get to this tomorrow.

willkg:
As a side note to me in case I ever look at this again, it looks like all of the API endpoints are GET-only. If that's always going to be true, it'd be a good idea to remove any complexity around supporting other HTTP methods.

g-k:
Cool, I was thinking having the symbol uploads permission might have enabled a POST endpoint, but antenna should handle that now.

stephendonner:
This is k-rad 👍

g-k:
Ah, there's a POST endpoint for reprocess crashes and the test for it [here](https://github.com/mozilla-services/socorro/blob/34176bd348a0371191e059ba5f1387201b4e41b0/webapp-django/crashstats/api/tests/test_views.py#L870) fails on https://github.com/mozilla-services/socorro/compare/master...g-k:all-api-endpoints-must-be-get?expand=1

Do people still use that endpoint?

willkg:
Yes--that's a pretty key endpoint for development/support. I use it pretty regularly.

willkg:
I forgot that a bunch of API endpoints aren't advertised on that page--they're "hidden".

g-k:
No worries, I'll leave the api docs template as is then.

PR #4404, opened 2018-04-15
bug 1417989 - fix method signature of StackFrameSymbolizerForward::FillSourceLineInfo

luser:
When Breakpad was updated to the most recent snapshot the signature of this
method in the base class changed, but we didn't update our implementation
to match, so our overload was not being called.


willkg:
The tests failed for unrelated reasons. I fixed that and updated this PR branch.

willkg:
@luser The crash in the bug report isn't around anymore, so I looked for another crash that has `xul.dll` showing up in the stack.

I look at this crash:

https://crash-stats.mozilla.com/report/index/8a46ab4b-67a0-4bc8-a065-2adf40180416

It's got source urls for `xul.dll`.

This one has source urls for almost all of the `xul.dll` frames:

https://crash-stats.mozilla.com/report/index/ef84253e-0ae0-4639-b74a-6f5f20180415

Is that the issue here? That sometimes frames don't get a source url?

If not, is there a recent example crash that I can test against?

luser:
@willkg The feature that broke here is that entries in the `modules` list should have a `symbol_url` key when they have symbols loaded. The report/index page has code to generate a link to that URL when present, so you can easily get a copy of the symbol file used for a module in a specific crash report.

luser:
There aren't any examples of the previous behavior left anymore, unfortunately.

luser:
Thanks! This should be easy enough to verify once this is live in stage.

PR #4406, opened 2018-04-16
fix bug 1448957 - ignore specifiers and return types in signature generation

willkg:
This fixes signature generation to ignore specifiers and return types when matching against the signature list.

### DIFF
--- a/socorro/signature/__main__.py
+++ b/socorro/signature/__main__.py
@@ -31,7 +31,7 @@ logger = logging.getLogger('socorro.signature')
 
 
 # FIXME(willkg): This hits production. We might want it configurable.
-API_URL = 'https://crash-stats.mozilla.com/api/'
+API_URL = 'https://crash-stats.mozilla.com/api'
 
 
 def setup_logging(logging_level):
--- a/socorro/signature/rules.py
+++ b/socorro/signature/rules.py
@@ -76,9 +76,11 @@ class SignatureTool(object):
 
 
 class CSignatureTool(SignatureTool):
-    """This is the class for signature generation tools that work on
-    breakpad C/C++ stacks.  It provides a method to normalize signatures
-    and then defines its own '_do_generate' method."""
+    """This is the class for signature generation tools that work on breakpad C/C++
+    stacks. It provides a method to normalize signatures and then defines its
+    own '_do_generate' method.
+
+    """
 
     hang_prefixes = {
         -1: "hang",
@@ -219,6 +221,7 @@ class CSignatureTool(SignatureTool):
             # Ensure a space after commas
             function = self.fixup_comma.sub(', ', function)
             return function
+
         # if source is not None and source_line is not None:
         if file and line:
             filename = file.rstrip('/\\')
@@ -260,8 +263,13 @@ class CSignatureTool(SignatureTool):
         # Get all the relevant frame signatures.
         new_signature_list = []
         for a_signature in source_list:
-            # If the signature matches the irrelevant signatures regex, skip to the next frame.
-            if self.irrelevant_signature_re.match(a_signature):
+            # We want to match against the function signature or any of the parts of the
+            # signature in the case where there are specifiers and return types.
+            a_signature_parts = a_signature.split() + [a_signature]
+
+            # If one of the parts of the function signature matches the
+            # irrelevant signatures regex, skip to the next frame.
+            if any(self.irrelevant_signature_re.match(part) for part in a_signature_parts):
                 continue
 
             # If the signature matches the trim dll signatures regex, rewrite it to remove all but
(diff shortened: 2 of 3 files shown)


socorro/signature/__main__.py
```diff
@@ -31,7 +31,7 @@
 
 
 # FIXME(willkg): This hits production. We might want it configurable.
-API_URL = 'https://crash-stats.mozilla.com/api/'
+API_URL = 'https://crash-stats.mozilla.com/api'
```
willkg: The `/` at the end of this caused the urls to be like `https://crash-stats.mozilla.com/api//RawCrash/` which raises a 404 now. I don't know why it didn't do that before or when it changed. Anyhow, yay.

socorro/signature/rules.py
```diff
@@ -76,9 +76,11 @@ def _do_generate(self, source_list, hang_type, crashed_thread, delimiter):
...
 
 class CSignatureTool(SignatureTool):
-    """This is the class for signature generation tools that work on
-    breakpad C/C++ stacks.  It provides a method to normalize signatures
-    and then defines its own '_do_generate' method."""
+    """This is the class for signature generation tools that work on breakpad C/C++
+    stacks. It provides a method to normalize signatures and then defines its
+    own '_do_generate' method.
+
+    """
```
willkg: I just fixed the formatting here. Cosmetic.

socorro/signature/rules.py
```diff
@@ -260,8 +263,13 @@ def _do_generate(self, source_list, hang_type, crashed_thread, delimiter=' | '):
         # Get all the relevant frame signatures.
         new_signature_list = []
         for a_signature in source_list:
-            # If the signature matches the irrelevant signatures regex, skip to the next frame.
-            if self.irrelevant_signature_re.match(a_signature):
+            # We want to match against the function signature or any of the parts of the
+            # signature in the case where there are specifiers and return types.
+            a_signature_parts = a_signature.split() + [a_signature]
```
willkg: So this is the meat of the fix. We used to get frames like this:

```
mozilla::net::nsHttpConnectionMgr::Shutdown
```

and now we're getting frames like this:

```
static void mozilla::net::nsHttpConnectionMgr::Shutdown
```

The "static" part is a specifier. The "void" part is the return type. I think we could build a list of all the specifiers we're seeing and that list would be stable over time and we wouldn't have to futz with it much. The return types are a different story. There are some common ones like "int" and "bool" and "void", but then there are a lot of other ones that are defined in the Firefox code like "JSObject*".

So instead of stripping specifiers and return types, we split the signature into parts and make sure to add the whole thing so then the irrelevant and prefix things try to match any of the strings in this list:

```
['static', 'void', 'mozilla::net::nsHttpConnectionMgr::Shutdown', 'static void mozilla::net::nsHttpConnectionMgr::Shutdown']
```

Adding the whole signature at the end allows for the `<name omitted>` case in the prefix list which splits poorly.

willkg:
Output from this:

```
you@processor:/app$ socorro-cmd fetch_crashids --num=1000 | python -m socorro.signature --different-only

you@processor:/app$ socorro-cmd fetch_crashids --signature-contains=static | python -m socorro.signature --different-only
Crash id: 8ddf57e0-664d-4b58-91a7-9f5590180415
Original: arena_t::DallocSmall | static void arena_dalloc
New:      arena_t::DallocSmall | static void arena_dalloc | mozilla::dom::TimeRanges::cycleCollection::DeleteCycleCollectable
Same?:    False

Crash id: 0b42693a-82c9-41bf-896f-ac6f90180415
Original: static void std::panicking::rust_panic_with_hook
New:      static union style::properties::animated_properties::AnimationValue geckoservo::glue::compose_animation_segment
Same?:    False

Crash id: 7aed80a8-68a1-4399-a22c-a673a0180415
Original: RtlpWaitOnCriticalSection | EtwEventEnabled | static void arena_dalloc
New:      RtlpWaitOnCriticalSection | EtwEventEnabled | static void arena_dalloc | HeapFree | static void core::ptr::drop_in_place<T>
Same?:    False

Crash id: 75951b04-53bd-4f7d-92a5-737980180415
Original: RtlpWaitOnCriticalSection | EtwEventEnabled | static void arena_dalloc
New:      RtlpWaitOnCriticalSection | EtwEventEnabled | static void arena_dalloc | js::SharedScriptData::decRefCount
Same?:    False

Crash id: 338d25b8-24c9-4e67-81a9-2f2740180415
Original: static void std::panicking::rust_panic_with_hook
New:      static union style::properties::animated_properties::AnimationValue geckoservo::glue::compose_animation_segment
Same?:    False

Crash id: ce973169-dd4a-48fb-a308-163420180415
Original: shutdownhang | static bool mozilla::SpinEventLoopUntil<T>
New:      shutdownhang | static bool mozilla::SpinEventLoopUntil<T> | mozilla::net::nsHttpConnectionMgr::Shutdown
Same?:    False
```

socorro/unittest/signature/test_rules.py
```diff
@@ -221,6 +221,22 @@ def test_generate_4(self):
...
+        generator = self.setup_config_c_sig_tool(
+            # irrelevant
+            ig=['ignored'],
+            # prefix
+            pr=['pre1', 'pre2'],
+        )
+        source_list = (
+            'static void ignored',
+            'pre1',
+            'static bool pre2',
```
Osmose: Should we add another source in the list after `pre2`? That'd help check that `pre2` is actually being treated as a prefix despite having the `static bool` in front. Right now it's not clear from the assertion that `pre2` matched the prefix list vs just being the first unmatched source.

willkg: This is why I :heart: your reviews.

willkg:
Thank you!

PR #4408, opened 2018-04-18
fix bug 1453476 - add "today" and "now" to fetch_crashids

willkg:
* add "today" as a valid value for `--date`
* add "now" as a valid value for `--date` that is like "today", but
  sets sort order to `-date` and adds cache-busting bits

### DIFF
--- a/socorro/scripts/fetch_crashids.py
+++ b/socorro/scripts/fetch_crashids.py
@@ -4,9 +4,12 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+from __future__ import print_function
+
 import argparse
 import datetime
 from functools import total_ordering
+import random
 import sys
 from urlparse import urlparse, parse_qs
 
@@ -141,12 +144,15 @@ def main(argv=None):
         description=DESCRIPTION.strip(),
     )
     parser.add_argument(
-        '--date', default='yesterday',
-        help='Date to pull crash ids from (YYYY-MM-DD)'
+        '--date', default='',
+        help=(
+            'date to pull crash ids from as YYYY-MM-DD, "yesterday", "today", or "now"; '
+            'defaults to "yesterday"'
+        )
     )
     parser.add_argument(
         '--signature-contains', default='', dest='signature',
-        help='Signature contains this string'
+        help='signature contains this string'
     )
     parser.add_argument(
         '--url', default='',
@@ -154,11 +160,11 @@ def main(argv=None):
     )
     parser.add_argument(
         '--num', default=100,
-        help='The number of crash ids you want or "all" for all of them'
+        help='number of crash ids you want or "all" for all of them'
     )
     parser.add_argument(
         '-v', '--verbose', action='store_true',
-        help='Increase verbosity of output'
+        help='increase verbosity of output'
     )
 
     if argv is None:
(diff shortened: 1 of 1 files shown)


socorro/scripts/fetch_crashids.py
```diff
@@ -141,24 +144,27 @@ def main(argv=None):
         description=DESCRIPTION.strip(),
     )
     parser.add_argument(
-        '--date', default='yesterday',
-        help='Date to pull crash ids from (YYYY-MM-DD)'
+        '--date', default='',
+        help=(
+            'date to pull crash ids from as YYYY-MM-DD, "yesterday", "today", or "now"; '
+            'defaults to "yesterday"'
+        )
```
willkg: The previous help text was missing stuff. Plus the default doesn't show up when you do `--help`. I changed the default to empty string here because it makes figuring out what's going on easier later on.

socorro/scripts/fetch_crashids.py
```diff
@@ -177,17 +183,31 @@ def main(argv=None):
     params['_columns'] = 'uuid'
 
     # Override with date if specified
-    datestamp = args.date
-    if 'date' not in params or datestamp != 'yesterday':
-        if datestamp == 'yesterday':
+    if 'date' not in params or args.date:
+        datestamp = args.date or 'yesterday'
```
willkg: This if check didn't work if you provided a `--url` value which had a `date` param and wanted to override the date with `yesterday`. I fixed it so now it should work right.

willkg:
This "now" is the magic sauce that lets us do a faux stage submitter. Landing this makes my life easier since "now" will be in the master branch.

willkg:
Best way to test this is to run the command with "now", "today", and "yesterday" and also pass the `--verbose` option which shows the Super Search params:

```
you@processor:/app$ socorro-cmd fetch_crashids --verbose --num=1
Params: {'date': ['>=2018-04-18', '<2018-04-19'], 'product': 'Firefox', '_columns': 'uuid'}
044b488b-b0ed-44de-bb9e-d692b0180418
you@processor:/app$ socorro-cmd fetch_crashids --verbose --num=1 --date=yesterday
Params: {'date': ['>=2018-04-18', '<2018-04-19'], 'product': 'Firefox', '_columns': 'uuid'}
044b488b-b0ed-44de-bb9e-d692b0180418
you@processor:/app$ socorro-cmd fetch_crashids --verbose --num=1 --date=today
Params: {'date': ['>=2018-04-19', '<2018-04-20'], 'product': 'Firefox', '_columns': 'uuid'}
2fbeb4b0-e71f-4bbb-867e-cf30f0180419
you@processor:/app$ socorro-cmd fetch_crashids --verbose --num=1 --date=now
Params: {'date': ['>=2018-04-19T00:00:04.000Z', '<2018-04-20T00:00:00.000Z'], 'product': 'Firefox', '_columns': 'uuid', '_sort': '-date'}
beb3dccf-72a3-4bfa-b73c-2623b0180419
you@processor:/app$ 
```

socorro/scripts/fetch_crashids.py
```diff
@@ -177,17 +183,31 @@ def main(argv=None):
...
             startdate = datetime.datetime.strptime(datestamp, '%Y-%m-%d')
 
         enddate = startdate + datetime.timedelta(days=1)
+
+        startdate = startdate.strftime('%Y-%m-%d')
+        enddate = enddate.strftime('%Y-%m-%d')
+
+        if datestamp == 'now':
+            # If we want crashes for now, add some cache-busting bits to the
+            # url and also sort in reverse by date
```
Osmose: Why sort in reverse by date?

socorro/scripts/fetch_crashids.py
```diff
@@ -177,17 +183,31 @@ def main(argv=None):
...
+            startdate = '%sT00:00:%02d.000Z' % (startdate, random.randint(0, 59))
```
Osmose: Instead of randomizing the seconds when the datestamp is now, could we change the pattern passed to `strftime` to include hours, minutes, and seconds? Then we avoid the (admittedly low) risk of hitting a cached value by accident.

socorro/scripts/fetch_crashids.py
willkg: If you don't sort reverse by date, then you don't get the most recent crashes.

socorro/scripts/fetch_crashids.py
willkg: I'm trying to make sure it works even if the client and server computers are out of sync time-wise. That's the big reason this code is wonky. But I think I can figure out a better way to do it without the chance of hitting cache that it does now.

willkg:
This is probably good enough. If we need to adjust it, we can do that later.

PR #4411, opened 2018-04-20
fix bug 1455295 - implement basic status page

willkg:
This implements a basic status page that shows version, migration, and
crontabber state. It'll make it easier to surveil the status of the site.

### DIFF
--- a/webapp-django/crashstats/manage/jinja2/manage/site_overview.html
+++ b/webapp-django/crashstats/manage/jinja2/manage/site_overview.html
@@ -1,0 +1,69 @@
+{% extends "manage/base.html" %}
+
+{% block page_title %}{{ super() }} - Site Overview{% endblock %}
+
+{% block admin_title %}{{ super() }} - Site Overview{% endblock %}
+
+{% block mainbody %}
+<div class="panel">
+    <div class="body notitle">
+        <h3>Version info</h3>
+        {% for url, info in version_info.items() %}
+            <p>{{ url }}: {{ info }}</p>
+        {% endfor %}
+
+        <h3>Alembic status</h3>
+        <p>
+            {% if alembic_error %}
+                {{ alembic_error }}
+            {% else %}
+                HEAD: {{ alembic_version }}
+            {% endif %}
+        </p>
+
+        <h3>Django db status</h3>
+        {% if django_error %}
+            <p>{{ django_error }}</p>
+        {% else %}
+            <table class="data-table">
+                <tr>
+                    <th>id</th>
+                    <th>app</th>
+                    <th>name</th>
+                    <th>applied</th>
+                </tr>
+                {% for row in django_db_data %}
+                    <tr>
+                        {% for item in row %}
+                            <td>{{ item }}</td>
+                        {% endfor %}
+                    </tr>
+                {% endfor %}
+            </table>
+        {% endif %}
+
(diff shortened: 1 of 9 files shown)


willkg:
This will need us to add `OVERVIEW_VERSION_URLS` to the cloudops socorro env vars template. It's neat because we'll be able to see antenna, pigeon, and socorro version status from one page. However, it doesn't show much in a local dev environment.

I thought about adding GitHub and ES-derived information, but didn't. I figure we can do that in a separate pass under a new bug.

This accesses the db directly. That's nice because it doesn't require much wiring, but not nice because if those bits change, then maybe the SQL will break. However, it's migration bookkeeping tables, so I don't think they'll change any time soon.

The resulting overview page is not pretty. If someone wants to spruce it up, that's cool. I don't really want to spend more time on it.

The big use case I need this to solve is giving me a single place to go to see deployed shas, migration status, and crontabber error information for deploy monitoring. This will help a ton.

willkg:
The minor problem with this one is that it requires you to be an admin. Is there a need for a site status page for non-admin? I don't know.

There are a lot of monitoring/status pages now. Maybe we should nix them all except the Dockerflow ones and this one?

Here are more monitoring pages/links:

* https://crash-stats.mozilla.com/monitoring/
* https://crash-stats.mozilla.com/monitoring/crontabber/
* https://crash-stats.mozilla.com/crontabber-status/
* https://crash-stats.mozilla.com/monitoring/healthcheck/
* https://crash-stats.mozilla.com/api/Status/ (errors out)
* https://crash-stats.mozilla.com/api/CrontabberState/

webapp-django/crashstats/manage/jinja2/manage/site_overview.html
```diff
@@ -0,0 +1,69 @@
...
+                <tr>
+                    <th>id</th>
+                    <th>app</th>
+                    <th>name</th>
+                    <th>applied</th>
+                </tr>
+                {% for row in django_db_data %}
+                    <tr>
+                        {% for item in row %}
+                            <td>{{ item }}</td>
```
Osmose: Could we use attribute names instead of iterating through the rows? Right now the rows match up only because the HTML and view code are in sync, but defining that in one spot instead of two would avoid them being out of sync.

willkg:
Mmm... I thought about making this predominantly client-side, but we have no testing of JavaScript other than manual testing. While the unit test I wrote isn't wildly exciting, that's more than we have with JS code on the site. That may not always be true, but it's definitely true now.

Adding a couple more API endpoints is doable and probably not hard, but they'd have to be public or require auth tokens. That seems like a lot of work for something that in this PR is 22 lines of code that's pretty straight-forward. We could minimize that work by writing a single Status API endpoint that returns all the data, but at that point it's pretty equivalent to what's here in this PR.

I'm not thrilled about this page, but I do need something now to make deploys easier and figured I could address a few things in a single page.

I'll fix the names and row access code. I think we should defer rethinking the architecture as a client-side webapp until later. Maybe other use cases and issues will come up.

willkg:
@Osmose ^^^ That renames the page and related things to "site status". I think I picked overview earlier because there was already a status thing which was for maintaining the status messages. That commit also fixes row iteration.

willkg:
STATUS PAGE FOR THE WIN!

Thank you!

PR #4417, opened 2018-04-26
fix bug 1380145 - improve super search fields docs and validation test

willkg:
This improves the super search fields docs and adds a validation test.

The validation test isn't amazing--it's pretty rough. At some point, we were thinking about adding more schema stuff. We could switch this to a schema. I think that'd probably be a great next step, but I didn't want to try to tackle that now.


docs/crashstorage/elasticsearch.rst
```diff
@@ -127,41 +127,86 @@ contains data about each field from Elasticsearch that can be manipulated.
...
-+----------------------+---------------------------------------------------------+
-| has_full_version     | Does this field have a full version in Elasticsearch?   |
-|                      | Enable only if you use a multitype field in the         |
-|                      | storage mapping.                                        |
-+----------------------+---------------------------------------------------------+
-| storage_mapping      | Mapping that is used in Elasticsearch for this field.   |
-|                      | See Elasticsearch documentation for more info.          |
-+----------------------+---------------------------------------------------------+
+**name**
+    The name of the field.
```
willkg: I wanted to add links and examples and other bits to this data, so I switched it from a table to a definition list.

docs/crashstorage/elasticsearch.rst
```diff
@@ -127,41 +127,86 @@ contains data about each field from Elasticsearch that can be manipulated.
...
+
+**default_value**
+    The default value for a Super Search filter.
+
+    Always set to ``None``.
+
+**is_mandatory**
+    Is this field required to have a value in the Super Search form?
+
+    Always set to ``False``.
```
willkg: We should write up a bug to ditch these. They're "used" in one place, but they're really just fluff.

socorro/external/es/super_search_fields.py
```diff
@@ -539,8 +554,8 @@ def get(self):
     'SecondsSinceLastCrash': {
         'data_validation_type': 'enum',
         'default_value': None,
+        'description': '',
         'form_field_choices': None,
-        'form_field_type': None,
```
willkg: I nixed these. I don't know whether they were used for something in the old days, but they're not used anymore and some of the fields don't have them.

socorro/unittest/external/es/test_supersearch.py
```diff
@@ -109,28 +109,6 @@ def test_get(self):
...
-        res = self.api.get(_columns=[
-            'date', 'signature', 'removed_fields'
-        ], _facets=['removed_fields'])
-
-        assert 'removed_fields' in res['hits'][0]
-        assert res['hits'][0]['removed_fields'] == 'foo bar'
-
-        assert 'removed_fields' in res['facets']
-        assert len(res['facets']['removed_fields']) == 2
-
```
willkg: We don't have anything in the "root document", so I nixed the test for this feature which we don't use.

socorro/unittest/external/es/test_supersearch.py
```diff
@@ -1743,21 +1722,21 @@ def test_get_with_columns(self):
...
         kwargs = {
-            '_columns': ['product', 'product_2']
+            '_columns': ['cpu_name', 'cpu_arch']
         }
         res = self.api.get(**kwargs)
 
-        assert 'product_2' in res['hits'][0]
-        assert 'product' in res['hits'][0]
+        assert 'cpu_name' in res['hits'][0]
+        assert 'cpu_arch' in res['hits'][0]
```
willkg: `cpu_name` is deprecated and `cpu_arch` is the synonym.

docs/crashstorage/elasticsearch.rst
```diff
@@ -127,41 +127,86 @@ contains data about each field from Elasticsearch that can be manipulated.
...
+**has_full_version**
+    Does this field have a full version in Elasticsearch? Enable only if you use
+    a multitype field in the storage mapping.
+
+**storage_mapping**
+    Mapping that is used in Elasticsearch for this field. See below for more
+    information.
+
+
+We also have these properties:
```
Osmose: So we have to include these but always set them to the same value? It might be good to explicitly say that as part of this demarcation rather than in the descriptions below.

willkg: That makes sense to me!

willkg:
Thank you!

docs/crashstorage/elasticsearch.rst
adngdb: Yeah, that was me being too generic and thinking we would use the same code for other API endpoints. Turns out we never did that, so those never got to be useful.

socorro/external/es/super_search_fields.py
adngdb: I think at some point I inferred the type of form field to use from other values, but never removed this. :+1:

docs/crashstorage/elasticsearch.rst
willkg: That makes a ton of sense! I was thinking they were added for a reason that never came to pass. I shouldn't have used the word "fluff"--sorry about that!

socorro/external/es/super_search_fields.py
willkg: That makes a ton of sense, too! Thank you!

PR #4418, opened 2018-04-26
fix bug 1430178: add "make lint"

willkg:


willkg:
This PR introduces the first of a new generation of make rules that don't need to self-identify as docker-environment-related. This PR represents freedom from the docker-marking tyranny that has plagued so many make rules. This PR marks the end of a sad era for Socorro development and the beginning of a happier one.

willkg:
`make thankyou`

PR #4419, opened 2018-04-27
fix bug 1457349: remove submitter

willkg:
This removes all the submitter code and tests and related bits and bobs.
This also removes the poster requirement.


willkg:
This is just code removal. To test, maybe do something like:

1. `make dockerbuild`
2. `make dockertest`

Pretty sure that covers it.

willkg:
Take that, Socorro codebase!

Thank you!

PR #4420, opened 2018-04-27
Bug 1457664 - [e2e-tests] Use generated Pipfile.lock; don't ignore Pipfile* in git, and remove constraints.txt

stephendonner:
@Osmose @davehunt r?

I'm not 100% sure about --ignore-pipfile

1 pre-existing failure aside, passing ad-hoc against production, here: https://qa-preprod-master.fxtest.jenkins.stage.mozaws.net/job/socorro.adhoc/222/console

Paste: https://gist.github.com/stephendonner/362b697cea0f659106cec75e79f2dd41


e2e-tests/Dockerfile
```diff
@@ -36,8 +36,8 @@ RUN curl -fsSLo /tmp/geckodriver.tar.gz https://github.com/mozilla/geckodriver/r
   && ln -fs /opt/geckodriver-$GECKODRIVER_VERSION /usr/bin/geckodriver
 
 WORKDIR /src
-COPY Pipfile /src/
-RUN pipenv install --system --skip-lock
+COPY Pipfile* /src/
+RUN pipenv install --system --deploy --ignore-pipfile
```
davehunt: I don't think we need `--ignore-pipfile` as the Pipfile.lock will be used by default, and with `--deploy` we'll abort if the lock file is out of date.

Osmose:
> We could remove the version identifiers from Pipfile and add Pipfile.lock to pyup.io. The only issue here is that if a package adds new dependencies, pyup.io will not know to add them to Pipfile.lock`.

I thought pyup.io didn't update the hashes in the lockfile? If the bot submitted a PR that updated the version in the lockfile without updating the hash, wouldn't that also break?

It honestly seems simpler to to me to just skip out on Pyup until they have full, proper Pipfile support if we're going to stick with Pipfiles. Either that or stick with requirements files until then.

We have a job running security checks on Socorro dependencies (using the Pyup command-line client), we could look into adding something similar for these requirements to cover the security side of Pyup. Github's also adding Python support to their security alerts sometime this year, which we might utilize.

davehunt:
Pyup does have support for updating the hashes in lock files.

Osmose:
> Pyup does have support for updating the hashes in lock files.

Huh. Why does Pyup ignore the lockfile when the Pipfile has versions, but update it when the Pipfile doesn't have versions?

I guess if we're using `--deploy` in the container, then pinning versions in the Pipfile doesn't matter since they're pinned in the lockfile.

e2e-tests/Dockerfile
stephendonner: Thanks; fixed.

stephendonner:
For reference, re: pyup and the meta hash updating in Pipfile.lock, etc., see https://github.com/pyupio/pyup/issues/197

stephendonner:
@davehunt r? Should be ready for another pass, and I updated and re-ran the adhoc job: https://qa-preprod-master.fxtest.jenkins.stage.mozaws.net/job/socorro.adhoc/224/console

PR #4426, opened 2018-05-01
fix bug 1440471, 1434425 - remove raw_update_channels and graphics_device related things

willkg:
This removes the `raw_update_channels` table and friends. This also removes the friends of `graphics_device`--but not the `graphics_device` table. That will get moved to Django-land in bug 1457747.

To test:

1. check out master and run `make dockersetup`
2. check out this branch and do:
   ```
   $ docker/as_me.sh bash
   you@processor:/app$ alembic -c docker/config/alembic.ini upgrade heads
   ```
3. check to make sure everything got removed using psql and `\d` and `\df`


alembic/versions/3474e98b321f_bug_1440471_1434425_more_tables.py
```diff
@@ -0,0 +1,44 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '3474e98b321f'
+down_revision = '1e188109fc6b'
+
+
+def upgrade():
+    # Update changed stored procedures
+    load_stored_proc(op, ['backfill_matviews.sql'])
+    load_stored_proc(op, ['update_product_versions.sql'])
```
willkg: It occurred to me that I should be reloading stored procedures before doing anything else because they often refer to things that are about to get removed. So I'm going to start doing that.

alembic/versions/3474e98b321f_bug_1440471_1434425_more_tables.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # Bug 1440471
+    op.execute('DROP TABLE IF EXISTS raw_update_channels')
+    op.execute('DROP TABLE IF EXISTS update_channel_map')
```
willkg: `update_channel_map` is related to `raw_update_channels` so if we remove one, we can remove the other. They're both something something for B2G.

willkg:
Thank you!

PR #4432, opened 2018-05-02
fix bug 1458641: fix reports-clean crontabber app

willkg:
The `reports-clean` crontabber app relies on a stored procedure that should have
been updated when we removed the `update_channels_map` table. This adds a migration
to do that.


willkg:
Thank you, @peterbe!

PR #4434, opened 2018-05-04
fix bug 1375434: redo first signature maintenance

willkg:
This replaces the `SignaturesCronApp` with `UpdateSignaturesCronApp`. There are a myriad of differences:

**`SignaturesCronApp` runs once a day, whereas `UpdateSignaturesCronApp` runs every hour.**

The new one runs once an hour so that the Elasticsearch query is manageable. In my testing, it'll hit Elasticsearch roughly 10 times pulling down about 2mb of data in total. That should be totally doable. If we expand the period from 1 hour to 1 day, it'd be pulling 24 times that which seems a bit much plus since Elasticsearch fails periodically, I'd rather a single run did less than more and thus reduce the risk it'd fail and have to run again later.

**`SignaturesCronApp` runs a matview that looks at the `reports` table fed by the postgres crashstorage classes, whereas `UpdateSignaturesCronApp` looks at crash data in Elasticsearch.** 

We're moving towards a world where the processor doesn't talk to postgres at all. Changing `signatures` table maintenance to be Elasticsearch-based rather than `reports`-table based moves us towards that goal.

**`SignaturesCronApp` updates the `signatures`, `signature_products`, and `signature_products_rollup` tables, whereas `UpdateSignaturesCronApp` only updates the `signatures` table.**

Socorro uses the `signatures` table, but not the other two. I wrote up a bug about how I think the `SignatureFirstDate` API should be looking at `signature_products`. Until we figure that out, I figured it wasn't worth adding maintenance for them. Also, I didn't include removal of those two tables in this PR. I pushed that off to another bug, too.

That's the big differences.

The one annoying thing is that `UpdateSignaturesCronApp` truncates minutes and seconds when determining the window of crashes to look at. That's annoying because it makes it harder to test. You have to process crashes, then wait for the next hour before they get absorbed by the cron job. The thinking for this is that crontabber kicks things off irregularly because of the way it runs in the docker container, so I figured I'd truncate minutes and seconds and make the windows it looks at more predictable regardless of when it actually runs.

As a side note, this uses @peterbe's model for updating the table. I used a different Elasticsearch query, but it's effectively the same idea. When @peterbe originally wrote his PR, I wondered if there were other mechanisms than a crontabber app that we could use. There are... but everything I thought of was more involved and added more moving parts. Right now I'm hedging towards "less code and fewer moving parts to make things work, then we can revisit everything later when the world is simpler".

And that's my brain dump on that!


willkg:
To test this, you want to process some crashes, then wait until the next hour mark, then run the crontabber job.

Here's a script for processing crashes in case you don't have one (I think you do):

```
#!/bin/bash

set -e

# Pulls down crash data for a lot of crashes and sets it up for processing.
#
# Note: This assumes you've run "make dockersetup" and "make dockerupdatedata"
# already.
#
# Usage:
#
#    $ ./docker/as_me.sh bash
#    you@processor:/app$ ./process_many.sh
#
# After running this script, run the processor:
#
#    $ docker-compose up processor
#
# Wait for crashes to process. Then you're good to go!

NUM_CRASHES=100
DATADIR=./crashdata_populate_tmp

function cleanup {
    # Cleans up files generated by the script
    rm -rf "${DATADIR}"
}

# Set up cleanup function to run on script exit
trap cleanup EXIT

mkdir ${DATADIR}

socorro-cmd fetch_crashids --num=${NUM_CRASHES} --date=now | socorro-cmd fetch_crash_data ${DATADIR}

echo "Sync contents to S3 bucket..."
(./scripts/socorro_aws_s3.sh mb s3://dev_bucket/ || true) 2> /dev/null
./scripts/socorro_aws_s3.sh cp --recursive ${DATADIR} s3://dev_bucket/

echo "Add crashids to socorro.normal processing queue..."
find ${DATADIR}/v2/raw_crash/ -type f | sed 's!.*/!!' | socorro-cmd add_crashid_to_queue socorro.normal
```

To run the crontabber app, run this script in the crontabber Docker container:

```
#!/bin/bash

# make dockersetup

export CRONTABBERCMD="socorro-cmd crontabber"

${CRONTABBERCMD} --reset-job=signatures-matview
time ${CRONTABBERCMD} \
     --force \
     --job=signatures-matview
```

socorro/cron/crontabber_app.py
```diff
@@ -42,7 +42,7 @@
     # Crash data analysis
     'socorro.cron.jobs.matviews.ReportsCleanCronApp|1h',
     'socorro.cron.jobs.bugzilla.BugzillaCronApp|1h',
-    'socorro.cron.jobs.matviews.SignaturesCronApp|1d|05:00',
+    'socorro.cron.jobs.update_signatures.UpdateSignaturesCronApp|1h|05:00',
```
willkg: Oops--need to fix that. It shouldn't specify a time.

socorro/external/postgresql/raw_sql/procs/update_signatures_hourly.sql
```diff
@@ -1,127 +0,0 @@
...
-
-DELETE FROM signature_products_rollup;
-
-insert into signature_products_rollup ( signature_id, product_name, ver_count, version_list )
-select
-    signature_id, product_name, count(*) as ver_count,
-        array_accum(version_string ORDER BY product_versions.version_sort)
-from signature_products JOIN product_versions
-    USING (product_version_id)
-group by signature_id, product_name;
```
willkg: Nothing uses the rollup table, so I didn't add support for maintaining it.

socorro/external/postgresql/raw_sql/procs/update_signatures_hourly.sql
```diff
@@ -1,127 +0,0 @@
...
-        min(new_signatures.first_report)
-from new_signatures JOIN signatures
-    ON new_signatures.signature = signatures.signature
-    left outer join signature_products
-        on signatures.signature_id = signature_products.signature_id
-        and new_signatures.product_version_id = signature_products.product_version_id
-where new_signatures.product_version_id is not null
-    and signature_products.signature_id is null
-group by signatures.signature_id,
-        new_signatures.product_version_id;
```
willkg: Nothing uses the `signature_products` table, so I didn't add support for maintaining it. I wrote up a bug that sort of covers this issue and we can figure things out there.

socorro/external/postgresql/signature_first_date.py
```diff
@@ -35,3 +37,51 @@ def get(self, **kwargs):
...
+                WHERE
+                    signature = %(signature)s
+                LIMIT 1
+            )
+            RETURNING 2
+        )
+        SELECT * FROM update_signature
+        UNION ALL
+        SELECT * FROM insert_signature
+        """
```
Osmose: I've never actually seen the SQL for a manual upsert before, this is trippy.

socorro/unittest/cron/jobs/test_update_signatures.py
```diff
@@ -0,0 +1,257 @@
...
+from socorro.cron.crontabber_app import CronTabberApp
+from socorro.unittest.cron.jobs.base import IntegrationTestBase
+
+
+class FakeModel(object):
+    def __init__(self):
+        self._get_steps = []
+
+    def add_get_step(self, assertions, response):
+        assertions = assertions or []
```
Osmose: These assertions aren't used anywhere in the tests, should we drop them?

Actually, without the support for assertions, we can achieve what this is doing by assigning an iterable to the mock SuperSearch. Each time it's called, it will return the next item in the iterable, and raise an error if it is called more times than there are values:

```python
@mock.patch('socorro.cron.jobs.update_signatures.SuperSearch')
def test_no_crashes_to_process(self, mock_supersearch):
    mock_supsersearch.side_effect = [
        {
            'errors': [],
            'hits': [],
            'total': 0,
            'facets': {}
        },
    ]
```

socorro/unittest/cron/jobs/test_update_signatures.py
```diff
@@ -0,0 +1,257 @@
...
+            assertions=None,
+            response={
+                'errors': [],
+                'hits': [],
+                'total': 0,
+                'facets': {}
+            }
+        )
+
+        # Run crontabber
```
Osmose: There's a lot of repeated steps in these tests, could this be parameterized into a single test?

socorro/unittest/cron/jobs/test_update_signatures.py
```diff
@@ -0,0 +1,257 @@
...
+        crontabber_info = self._load_structure()
+        assert crontabber_info['update-signatures']['last_error'] == {}
+        assert crontabber_info['update-signatures']['last_success']
+
+        # Assert that nothing got inserted
+        data = self.fetch_signatures_data()
+        assert len(data) == 0
+
+    @mock.patch('socorro.cron.jobs.update_signatures.SuperSearch')
+    def test_signature_insert_and_update(self, mock_supersearch):
```
Osmose: This doesn't look parameterizable since it needs to insert and then update, but the "run job" and "assert no failures" bits could be factored out into methods to make the test a bit more concise.

socorro/cron/jobs/update_signatures.py
```diff
@@ -0,0 +1,131 @@
...
+
+            # Set up first page
+            '_results_offset': 0,
+            '_results_number': MAX_PAGE,
+        }
+
+        results = {}
+        crashids_count = 0
+
+        while True:
```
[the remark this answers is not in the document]
willkg: One of them is using the Super Search HTTP API and the other is using `SuperSearch` internal model API. I'm hesitant to cross those two streams right now. But I definitely hear you.

socorro/external/postgresql/signature_first_date.py
willkg: This is similar to other upsert code in Socorro. I don't understand exactly why the `UNION ALL` thing works the way it does where it tries update and if that doesn't do anything, then does the insert. The postgres docs didn't enlighten me at all on that account.

socorro/unittest/cron/jobs/test_update_signatures.py
willkg: Yeah! Except that this is a `FakeModel` and I'm thinking of uplifting it and using it for other things, too, since it's generalized to the middleware models. The assertions bit lets us verify the parameters passed in.

So, we could go one of two ways:

1. fuck the future and reduce it now
2. write up a bug to uplift it and use it for mocking other models

socorro/unittest/cron/jobs/test_update_signatures.py
willkg: I see where you're coming from, but it's two tests and they're not that long. The bulk of the code in both tests is Python data structures that would be there in either case.

I think I'd rather leave them as they are now unless you feel very strongly about this. If we add more tests in the future, we can spend time reducing repeated bits because we'll have more repetition then.

willkg:
^^^ That nixes the assertions check and reduces some of the repetition.

socorro/external/postgresql/signature_first_date.py
Osmose: I think the `UNION ALL` does nothing except execute both select statements, which evaluates the queries that populate those two tables. It's the `WHERE NOT EXISTS` bit that only runs the insert for signatures that don't exist yet, and the `WHERE` in the `UPDATE` bit that avoids updating signatures that don't exist. The magic of the `UNION ALL` is that it allows both statements to execute in one query, making it a real upsert instead of faking it with two separate queries.

socorro/unittest/cron/jobs/test_update_signatures.py
willkg: Ahh... ok. I forgot about the `WHERE NOT EXISTS` part. I couldn't figure out why it seemed like only one executed based on the final return codes. That makes soooo much sense now.

socorro/external/postgresql/signature_first_date.py
willkg: Ahh... ok. I forgot about the `WHERE NOT EXISTS` part. I couldn't figure out why it seemed like only one executed based on the final return codes. That makes soooo much sense now.

willkg:
Thank you!

PR #4435, opened 2018-05-04
fix bug 1445685: remove UploadCrashReportJSONSchemaCronApp

willkg:
We do this as a post-deploy step by running:

```
socorro-cmd upload_telemetry_schema
```

We don't need to run this as a crontabber job anymore.


willkg:
Thank you!

PR #4436, opened 2018-05-07
fix bug 1450286: remove old infra-related things

willkg:
This removes a swath of stuff related to the old infrastructure that we no
longer use.

This PR is almost entirely dead-script/config removal. I ran all the remaining rules in the `Makefile` and they're all find (nb the `dockerdependencycheck` one fails, but that's expected since it's not in the cron jobs list). We'll see if CI is fine. We need to land this to see if it affects deploys. It shouldn't, but one never knows.


willkg:
I restarted the build twice and it keeps failing at the same point and it looks like something that should be a transient problem. I can't reproduce it locally. I'll merge some other things and then rebase this and see if that "fixes it".

willkg:
Hooray--CI passed!

Let's land it and see how it deploys!

willkg:
Thank you!

PR #4437, opened 2018-05-07
Fix bug 1450286: re-add deploy-dockerhub.sh script

willkg:
This re-adds the `deploy-dockerhub.sh` script which Circle needs to push the images it built to Dockerhub.

I looked to see if we were using it, but because the thing that mentions it is in `.circleci/config.yml`, and that starts with a `.`, it got skipped by ripgrep. Oops.


willkg:
Thank you!

PR #4438, opened 2018-05-08
fix bug 1375434: fix build_id storage

willkg:
`build_id` is stored as a numeric in the db, so we need to convert it
before saving it.

I'm puzzled as to why it worked locally and doesn't work on -stage. Maybe they're different versions of postgres? Maybe they're configured differently in some way I'm not aware of?

Anyhow, pretty sure this should fix it. It's trivial to verify on -stage.


willkg:
Thank you!

PR #4439, opened 2018-05-08
fix bug 1375434: handle crashes with no build id

willkg:
Some crashes don't have a build id which is why they don't "intify" well.
This skips those because we don't consider them when figuring out
first build and first date.

### DIFF
--- a/socorro/unittest/cron/jobs/test_update_signatures.py
+++ b/socorro/unittest/cron/jobs/test_update_signatures.py
@@ -218,3 +218,30 @@ class UpdateSignaturesCronAppTestCase(IntegrationTestBase):
                 }
             ]
         )
+
+    @mock.patch('socorro.cron.jobs.update_signatures.SuperSearch')
+    def test_crash_with_no_buildid(self, mock_supersearch):
+        """Test crashes with no build id are ignored"""
+        supersearch = FakeModel()
+        mock_supersearch.return_value = supersearch
+
+        # Mock SuperSearch to return 4 crashes covering two signatures
+        supersearch.add_get_step({
+            'errors': [],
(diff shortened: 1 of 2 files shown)


socorro/unittest/cron/jobs/test_update_signatures.py
```diff
@@ -218,3 +218,30 @@ def test_multiple_crash_processing(self, mock_supersearch):
...
             ]
         )
+
+    @mock.patch('socorro.cron.jobs.update_signatures.SuperSearch')
+    def test_crash_with_no_buildid(self, mock_supersearch):
+        """Test crashes with no build id are ignored"""
+        supersearch = FakeModel()
+        mock_supersearch.return_value = supersearch
+
+        # Mock SuperSearch to return 4 crashes covering two signatures
```
Osmose: Leftover from copypaste?

socorro/unittest/cron/jobs/test_update_signatures.py
```diff
@@ -218,3 +218,30 @@ def test_multiple_crash_processing(self, mock_supersearch):
...
+                },
+            ],
+            'total': 1,
+            'facets': {}
+        })
+
+        # Run crontabber
+        self.run_job_and_assert_success()
+
+        # Two signatures got inserted
```
Osmose: Leftover?

socorro/unittest/cron/jobs/test_update_signatures.py
willkg: Blech. Yes.

willkg:
Thank you! Hopefully I got it right this time!

PR #4442, opened 2018-05-09
fix bug 1460125: remove postgres crashstorage in processor

willkg:
This takes the postgres crashstorage out of the processor crash destinations.
It also removes the `ReportsCleanCronApp` since we no longer need that anymore.

This moves some code around to fill the void that removing
`PostgreSQLCrashStorage` left behind and tidies some things up.

This does not remove related tables and stored procedures which are no
longer needed. Those will get removed in a future PR.


socorro/external/postgresql/new_crash_source.py
```diff
@@ -1,146 +0,0 @@
-# This Source Code Form is subject to the terms of the Mozilla Public
```
willkg: The crash source is what tells the processor which crashes need to be processed. We're using the RabbitMQ crash source. We can nix all the other ones including this one.

Figured I'd do that while I was removing things.

socorro/unittest/cron/jobs/test_matviews.py
```diff
@@ -139,15 +130,16 @@ def mock_utc_now():
             tab = CronTabberApp(config)
             tab.run_all()
 
-            information = self._load_structure()
+        information = self._load_structure()
 
-            for app_name in ('product-versions-matview',
-                             'adu-matview',
-                             'build-adu-matview'):
+        print(information)
```
willkg: Oops--I'll remove this.

socorro/unittest/external/postgresql/unittestbase.py
```diff
@@ -50,16 +49,20 @@ class PostgreSQLTestCase(TestCase):
...
+            },
+            {
+                "id": "mac",
+                "name": "Mac OS X"
+            },
+            {
+                "id": "linux",
+                "name": "Linux"
+            }
+        ],
```
willkg: This is cosmetic.

socorro/unittest/external/postgresql/unittestbase.py
```diff
@@ -79,10 +82,8 @@ class PostgreSQLTestCase(TestCase):
 
     @classmethod
     def get_standard_config(cls):
-
         config_manager = ConfigurationManager(
-            [cls.required_config,
-             ],
+            [cls.required_config],
```
willkg: This is cosmetic.

willkg:
When double-checking things, I went through the gamut:

1. `make dockerbuild`
2. `make dockersetup`
3. `make dockerupdatedata`
4. `make dockertest`
5. `make dockerrun`, process a crash, and click around the webapp
6. `docker-compose up crontabber` and let it run for an hour

I spent quality time making sure everything ran in my local dev environment. I think for this review, it'd help to go through and see if I missed anything, and then we'll push it and more formally test everything on stage where we're likely to spend far less time to find issues. This doesn't have any DB-related changes, so fixing issues in stage isn't hard.

willkg:
Thank you! Looking forward to seeing this on stage. Woo hoo!

PR #4444, opened 2018-05-10
fix bug 1460035: implement python 3 test scaffolding

willkg:
This adds a couple of make rules and shell scripts to give us scaffolding
with which we can iterate over fixing code to make it work in both
Python 2 and Python 3.

In order to get the tests to run at all, I had to fix a bunch of Python
3 issues. I also hoisted six to first-class dependency status.

This doesn't connect that with CI because linting fails epically. We'll do
that later.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,10 +62,10 @@ dockerbuild: my.env
 dockersetup: my.env .docker-build
 	${DC} run webapp /app/docker/run_setup.sh
 
-dockertest: my.env
+dockertest: my.env .docker-build
 	./docker/run_tests_in_docker.sh ${ARGS}
 
-dockertestshell: my.env
+dockertestshell: my.env .docker-build
 	./docker/run_tests_in_docker.sh --shell
 
 dockerupdatedata: my.env
@@ -79,3 +79,20 @@ dockerstop: my.env
 
 dockerdependencycheck: my.env
 	${DC} run crontabber ./docker/run_dependency_checks.sh
+
+# Python 3 transition related things
+
+.PHONY: dockerbuild3 dockertest3 dockertestshell3
+
+.docker-build3:
+	make dockerbuild3
+
+dockerbuild3: my.env
+	${DC} build testpython3
+	touch .docker-build3
+
+dockertest3: my.env .docker-build3
+	USEPYTHON=3 ./docker/run_tests_in_docker.sh ${ARGS}
+
+dockertestshell3: my.env .docker-build3
+	USEPYTHON=3 ./docker/run_tests_in_docker.sh --shell
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -79,6 +79,26 @@ services:
     volumes:
       - .:/app
 
+  # For running tests in as we transition to Python 3
+  testpython3:
+    build:
+      context: .
+      dockerfile: ./docker/Dockerfile.python3
+    image: local/socorro_python3
+    env_file:
+      - docker/config/local_dev.env
+      - docker/config/never_on_a_server.env
+      - my.env
+    depends_on:
+      - statsd
+      - localstack-s3
+      - postgresql
+      - elasticsearch
+      - rabbitmq
+      - memcached
+    volumes:
+      - .:/app
+
   # -----------------------------
   # External services
   # -----------------------------
--- a/docker/Dockerfile.python3
+++ b/docker/Dockerfile.python3
@@ -1,0 +1,40 @@
+# From Dockerfile
+FROM python:3.6.5-slim
+
+WORKDIR /app/
+RUN groupadd --gid 10001 app && useradd -g app --uid 10001 --shell /usr/sbin/nologin app
+
+# Install OS-level things
+COPY ./docker/set_up_ubuntu.sh /tmp/
+RUN DEBIAN_FRONTEND=noninteractive /tmp/set_up_ubuntu.sh
+
+# Install Socorro Python requirements
+COPY ./requirements /app/requirements
+RUN pip install -U 'pip>=8' && \
+    pip install --no-cache-dir -r requirements/default.txt -c requirements/constraints.txt
+
+# From webapp
+ENV PYTHONUNBUFFERED 1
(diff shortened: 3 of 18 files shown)


Makefile
```diff
@@ -37,10 +37,10 @@ dockersetup: my.env .docker-build
 lint: my.env
 	${DC} run processor flake8
 
-dockertest: my.env
+dockertest: my.env .docker-build
 	./docker/run_tests_in_docker.sh ${ARGS}
 
-dockertestshell: my.env
+dockertestshell: my.env .docker-build
```
willkg: These dependencies should have been there before. It's hard to run the tests without things being built.

docker-compose.yml
```diff
@@ -79,6 +79,26 @@ services:
...
+      - my.env
+    depends_on:
+      - statsd
+      - localstack-s3
+      - postgresql
+      - elasticsearch
+      - rabbitmq
+      - memcached
+    volumes:
+      - .:/app
```
willkg: This is modeled after the `webapp` container because we use that for the tests. I don't know offhand what the tests needed or didn't need, so I just copied the webapp stuff. I figure this is temporary scaffolding.

docker/Dockerfile.python3
```diff
@@ -0,0 +1,40 @@
...
+RUN cd /webapp-frontend-deps/ && npm install
+
+COPY . /app/
+
+ENV LESS_BINARY /webapp-frontend-deps/node_modules/.bin/lessc
+ENV UGLIFYJS_BINARY /webapp-frontend-deps/node_modules/.bin/uglifyjs
+ENV CSSMIN_BINARY /webapp-frontend-deps/node_modules/.bin/cssmin
+ENV NPM_ROOT_PATH /webapp-frontend-deps/
+
+USER app
```
willkg: This merges the important parts of `Dockerfile` and `Dockerfile.webapp`.

docker/config/test.env
```diff
@@ -6,8 +6,8 @@ alembic_config=/app/docker/config/alembic.ini
 # put filesystem crashstorage stuff in /tmp/test_crashes
 resource.fs.fs_root=/tmp/test_crashes
 
-# use socorro_integration_test for postgres and alembic
-database_url=postgres://postgres:aPassword@postgresql:5432/socorro_integration_test
+# use socorro_test for postgres and alembic
+database_url=postgres://postgres:aPassword@postgresql:5432/socorro_test
```
willkg: While working on things, I cleaned up db creation, so now we only create 2 databases instead of 3 in the tests. In this way, this PR is both 2-to-3 and also 3-to-2!!!!!

docker/run_tests.sh
```diff
@@ -19,65 +19,45 @@ set -v -e -x
...
+
+    # Collect static and then run py.test in the webapp
+    pushd webapp-django
+    python manage.py collectstatic --noinput
+    "${PYTEST}"
+    popd
+else
+    # Run the tests we know work in Python 3
+    ./docker/run_tests_python3.sh
+fi
```
willkg: The python-3-ification is only going to cover the Socorro tests. Once everything is working there, we can add in the webapp tests. I figured it's easier to keep the scaffolding simpler for now.

docker/run_tests_in_docker.sh
```diff
@@ -45,12 +56,13 @@ if [ "$1" == "--shell" ]; then
            --link socorro_elasticsearch_1 \
            --link socorro_postgresql_1 \
            --link socorro_rabbitmq_1 \
+           --link socorro_statsd_1 \
```
willkg: This should have been there before.

docker/run_tests_python3.sh
```diff
@@ -0,0 +1,17 @@
...
+#
+# Usage: ./docker/run_tests_python3.sh
+
+# This is the list of known working tests by directory/filename. When you
+# have tests in a directory/file working, add it to this list as a new line.
+WORKING_TESTS=(
+    socorro/unittest/lib/test_ooid.py
+)
+
+pytest ${WORKING_TESTS[@]}
```
willkg: I think this is pretty slick. It only tests what we tell it to test explicitly--no test discovery happens. That'll let us build things up over time.

It can take files or directories or even specific tests/classes (as long as the module loads).

requirements/constraints.txt
```diff
@@ -8,9 +8,6 @@ configobj==5.0.6 \
     --hash=sha256:a2f5650770e1c87fb335af19a9b7eb73fc05ccf22144eb68db7d00cd2bcb0902
 simplejson==3.1.3 \
     --hash=sha256:941a25bbb00b021dbe026b94278d9fab9f4655e11ab76eb3f022a9ee097f0478 # pyup:ignore
-six==1.11.0 \
-    --hash=sha256:832dc0e10feb1aa2c68dcc57dbb658f1c7e65b9b61af69048abc87a2db00a0eb \
-    --hash=sha256:70e8a77beed4562e7f14fe23a786b54f6296e34344c23bc42f07b15018ff98e9
```
willkg: I moved this to `default.txt` because we're using it.

requirements/default.txt
```diff
@@ -155,7 +155,7 @@ humanfriendly==4.12 \
 jsonschema==2.6.0 \
     --hash=sha256:000e68abd33c972a5248544925a0cae7d1125f9bf6c58280d37546b946769a08 \
     --hash=sha256:6ff5f3180870836cae40f06fa10419f557208175f13ad7bc26caa77beb1f6e02
-functools32==3.2.3-2 \
+functools32==3.2.3-2 ; python_version < '3.2' \
```
willkg: This is a backported module. We'll need to do this sort of thing if we use any other backported modules. Having said that, I didn't hit other similar issues.

willkg:
To test:

1. `make dockerbuild3`
2. `make dockertest3`

Watch the cacophony of linting errors!

willkg:
Thank you for reviewing! This is a small step on a medium-sized journey to a much better place. Hooray!

PR #4445, opened 2018-05-11
fix bug 1454167: improve overview docs

willkg:
I figured I'd do a pass on improving the "Socorro overview" docs to help contributors and interns in the near future. This also brings together links to a lot of relevant things.

I also moved some things around to places where I thought they made more sense.


docs/overview.rst
```diff
@@ -29,50 +32,55 @@ The crash ingestion pipeline that we have at Mozilla looks like this:
       subgraph stores {
          rank=same;
 
-         aws [shape=tab, label="S3", style=filled, fillcolor=gray];
+         s3raw [shape=tab, label="S3 (Raw)", style=filled, fillcolor=gray];
+         pigeon [shape=cds, label="pigeon"];
+         rabbitmq [shape=tab, label="RMQ", style=filled, fillcolor=gray];
```
Osmose: Can we label this "RabbitMQ"? A google search for "RMQ" doesn't bring up RabbitMQ, so it's kinda hard to tell what this is if you're not familiar with RabbitMQ.

docs/overview.rst
```diff
@@ -29,50 +32,55 @@ The crash ingestion pipeline that we have at Mozilla looks like this:
...
       }
 
       subgraph proc {
```
Osmose: It wasn't changed in this PR, but making the arrows to Postgres bidirectional to represent data being written to them would be helpful, as it would be more consistent with arrows representing data flowing from one place to another.

docs/overview.rst
```diff
@@ -86,70 +94,255 @@ Important services in the diagram:
...
-The processor stores crash data in several crash storage destinations:
+* **Crontabber:** Runs periodic housekeeping tasks.
+
+
+Let's take a tour through the crash ingestion pipeline!
+
+
+A tour through the crash ingestion pipeline
+===========================================
+
```
Osmose: If you like being funny and popular you could put another line here: "This is the tour through the crash ingestion pipeline"

docs/overview.rst
```diff
@@ -86,70 +94,255 @@ Important services in the diagram:
...
+     https://chromium.googlesource.com/breakpad/breakpad/+/master/docs/getting_started_with_breakpad.md
+
+   **Firefox-specific breakpad documentation**
+     https://antenna.readthedocs.io/en/latest/breakpad_reporting.html
+
+
+Collected by the Collector (Python, Falcon)
+-------------------------------------------
+
+The collector is the beginning of the crash ingestion pipeline. It accepts the
```
Osmose: Could we mention here that "Antenna" is the codename for the current collector?

docs/overview.rst
```diff
@@ -86,70 +94,255 @@ Important services in the diagram:
...
+It passes all that information through the processing pipeline which consists of
+a series of rules that transform the crash into a processed crash.
+
+One of the rules runs the minidump-stackwalker over the minidump to extract
+information about the process and symbolicates the symbols on the stack. It also
+determines some other things about the state of the process when Firefox
+crashed.
+
+Another rule generates a crash signature from the stack of the crashind thread.
+We use crash signatures to bucket crashes so we can look at them as groups.
```
Osmose: Bucket is a little vague here. Maybe something like: "We use crash signatures to group together crashes that appear similar and were potentially caused by the same bug."

willkg:
^^^ Addresses comments.

PR #4446, opened 2018-05-15
fix bug 1459272, 1459276, 1460990: remove postgres crashstorage db bits

willkg:
This removes all the postgres crashstorage related db bits: tables, stored procedures, models, etc. Since we removed the last partitioned table, it removes the partition table scaffolding, too.

To review:

1. check out master and run `make dockersetup`
2. check out this branch and run `./docker/as_me.sh alembic -c docker/config/alembic.ini upgrade heads`
3. skim the four migrations and check that tables and functions were removed using `\d` and `\df`
4. run `make dockersetup` to make sure that doesn't fail
5. run `make dockerupdatedata` to make sure that doesn't fail

Since we're not doing crash storage in the processor any more, the only thing that could be affected by these changes is crontabber jobs. If those fail, that's not a big deal--we can just fix them as we discover issues and it'll self-heal. This is all pretty low risk stuff now.


willkg:
Thank you thank you thank you!

PR #4448, opened 2018-05-16
bug 1458886: move the permission info next to the api name

is2ei:
- Fixed the position of the permission info on the API document.
- Please run the webapp and see  /api/


willkg:
This will auto-deploy to our stage environment. I can verify it there and make sure it gets deployed to prod.

Thank you so much for this!

PR #4449, opened 2018-05-17
bug 1459276: fix migration

willkg:
This fixes the migration to account for `reports_clean` being a partitioned
table.

stage can't get past this migration, so I think the only way to move forward is to fix the migration in place.


willkg:
This fixes the failure here:

https://app.datadoghq.com/event/event?id=1601902804776233349

willkg:
I'm a little bummed I missed this in the previous pass. `reports_clean` wasn't in staticdata as a partitioned table. So this is a difference between stage/prod and the local dev environment that I missed when I did the db audit back in ... March or whenever it was. :(

willkg:
Hopefully that fixes it. If not, ... well, we'll just keep on tryin!

PR #4451, opened 2018-05-20
bug 1458886: Show UTC on clicking the time shortcut

is2ei:
1. super search time "shortcut" From and To are incorrectly adjusted by timezone, cutting off most recent results
2. Display UTC after clicking the time shortcut (24hours, 3days, 7days,,,)
3. Click the time shortcut on "Super Search page".
4. I didn't add or update any test since it's UI issue.

### DIFF
--- a/webapp-django/crashstats/supersearch/static/supersearch/js/socorro/date_filters.js
+++ b/webapp-django/crashstats/supersearch/static/supersearch/js/socorro/date_filters.js
@@ -26,6 +26,11 @@ $(function () {
             }),
         };
 
+	$("#search-button").on('click', function() {
+	    filters.from.input.value = filters.from.selectedDates[0].toUTCString();
+	    filters.to.input.value = filters.to.selectedDates[0].toUTCString();
+	});
+
         function setDate(key, value) {
             filters[key].setDate(value, true);
         }


webapp-django/crashstats/supersearch/static/supersearch/js/socorro/date_filters.js
```diff
@@ -51,10 +51,15 @@ $(function () {
             var value = parseInt(range.slice(0, -1));
             var unit = range.slice(-1);
             var toDate = moment().utc().toDate();
+            var toDateString = moment(toDate).utc().format("MMM DD, YYYY - HH:mm");
             var fromDate = moment().utc().subtract(value, unit).toDate();
+            var fromDateString = moment(fromDate).utc().format("MMM DD, YYYY - HH:mm");
 
-            setDate('to', toDate);
-            setDate('from', fromDate);
+            filters.to.set('minDate', fromDate);
```
Osmose: Removing `setDate` means that the current value of the flatpickr input isn't being set to a different value. If you click the search button after clicking a shortcut link with this code, you'll notice that the date in the URL doesn't match the displayed value because of this.

Setting the date values will also cause the `onChange` handlers above to run, which will remove the need to manually set the `minDate` and `maxDate` here, as well as removing the need to call `removeSelectedShortcut`.

webapp-django/crashstats/supersearch/static/supersearch/js/socorro/date_filters.js
```diff
@@ -51,10 +51,15 @@ $(function () {
...
+            filters.from.set('maxDate', toDate);
+            removeSelectedShortcut();
+            filters.to.altInput.value = toDateString;
```
Osmose: This works, but isn't documented (vs the config values for `altInput`, `altFormat`, etc.). We probably should avoid using undocumented properties whenever possible so that they don't change out from underneath us when we upgrade the library.

is2ei: > We probably should avoid using undocumented properties

You're right. Let me find another way.

webapp-django/crashstats/supersearch/static/supersearch/js/socorro/date_filters.js
is2ei: I see. Thanks! I will fix it.

is2ei:
Thanks for reviewing!

> I think we want the input to accept local time, to submit UTC to the server, and for the server to return times in local time (and labelled as such).

I see. I will consider the behaviour and may ask you some questions.

willkg:
Mmm... The specific problem we're trying to solve is that when you choose From and To datetimes, the results aren't in that range and that's confusing and problematic (it keeps biting me, for example). I'm pretty sure what's going on is that the From and To fields are local time, but get a "Z" tacked on as if they were UTC when they get sent to the server and the server does a query against a UTC data set and you end up with results not in the period you asked for. This is increasingly worse the further from UTC you are.

For example, I searched with a To field of 2018-05-21 20:06 just now because the To field defaults to "now" in my timezone. That's the time in my timezone right now. That gets sent to the server as "2018-05-21T20:06:09.000Z" which is a lie. It should get converted to UTC and then sent to the server. 

There are a bunch of options for what we could do here, however there are messages on the site about how everything is in UTC, the related Super Search API accepts UTC, and we've done messaging over the years about how everything is in UTC. I think I want to keep this straight-forward and just keep everything in UTC.

Given that, I think there are two things to fix here:

1. make it clear the To and From boxes are UTC date times
2. (optional) make it so they default to "now" in UTC rather than "now" in local time

is2ei:
Hi @Osmose @willkg ,

> For example, I searched with a To field of 2018-05-21 20:06 just now because the To field defaults to "now" in my timezone. That's the time in my timezone right now. That gets sent to the server as "2018-05-21T20:06:09.000Z" which is a lie. It should get converted to UTC and then sent to the server.

I implemented this.
Before the input value is sent to the server, it will be converted to UTC.
You can confirm it with the following the procedure.
1. Open Super Search page.
2. Select any date.
3. Click "Search" button
4. Check the query string in the URL. It will be converted to UTC.

Please let me know if you have any concern! Thanks.

Osmose:
Thank you for the update! Just wanted to note that Monday is a holiday in the US, so I intend to review this on Tuesday.

is2ei:
Thanks for your comment! @Osmose 

I read the flatpickr's document. It seems the UTC related functions are removed from flatpickr.

https://github.com/flatpickr/flatpickr/issues/975

is2ei:
Thanks for your comment! @Osmose 

> I'm sorry I can't offer better advice. Maybe we should close this out and focus on other issues until we can address moving off of Flatpickr?

I see. Let me close this PR and wait for Flatpickr to support UTC. It seems the Flatpickr considers to support UTC again.

PR #4453, opened 2018-05-22
bug 1462749 - add django admin

willkg:
This sets up the django admin at `siteadmin/` parallel to the admin that we hand crafted from bars of hypo-allergenic bars of soap lightly scented with lavender.

I added Django admin model things for all the Django-model-backed things:

1. Users
2. Groups
3. API Tokens
4. Status Messages

In order to do that, I had to fix a few other things first:

1. the `makesuperuser` command now marks the user as `is_staff` so he/she can access the admin
2. running `make dockersetup` will now create the permissions and groups we need

This is the first step. The next step is to create Django admin versions of all the non-Django-model backed pages:

1. reprocessing
2. supersearch fields
3. site status
4. crash me now
5. analyze model fetches

After that, when we move tables from alembic/sqlalchemy land to Django land, we'll add Django admin views.

To test, do the following:

1. `make dockerbuild`, `make dockersetup`, and then `make dockerupdatedata`
2. make a staff user with `docker-compose run webapp webapp-django/manage.py makesuperuser EMAIL`
3. log into the site and check out `http://localhost:8000/admin/` vs. `http://localhost:8000/siteadmin/`


webapp-django/crashstats/settings/base.py
```diff
@@ -98,6 +98,8 @@ def path(*dirs):
     'waffle',
     'pinax.eventlog',
     'django_jinja',
+
+    'django.contrib.admin',
```
willkg: I put this last so that when we override the index template in `crashstats`, we'll get our version.

willkg: I'll add a comment about that in the next PR when I add the overridden index template.

webapp-django/crashstats/status/migrations/0002_auto_20180522_1545.py
```diff
@@ -0,0 +1,20 @@
...
+        ('status', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='statusmessage',
+            name='severity',
+            field=models.CharField(choices=[(b'info', b'Info'), (b'warning', b'Warning'), (b'critical', b'Critical')], max_length=20),
+        ),
+    ]
```
willkg: I moved the choices from the status message form to the model. That way they're available everywhere. That requires a migration.

webapp-django/crashstats/urls.py
```diff
@@ -28,22 +29,10 @@
...
-        namespace='sources'
-    )),
-    url(r'^home/', include(
-        'crashstats.home.urls',
-        namespace='home'
-    )),
+    url(r'^signature/', include('crashstats.signature.urls', namespace='signature')),
+    url(r'^topcrashers/', include('crashstats.topcrashers.urls', namespace='topcrashers')),
+    url(r'^sources/', include('crashstats.sources.urls', namespace='sources')),
+    url(r'^home/', include('crashstats.home.urls', namespace='home')),
```
willkg: This is cosmetic.

webapp-django/crashstats/urls.py
```diff
@@ -54,14 +43,10 @@
...
-    url(r'^profile/', include(
-        'crashstats.profile.urls',
-        namespace='profile'
-    )),
-    url(r'^documentation/', include(
-        'crashstats.documentation.urls',
-        namespace='documentation'
-    )),
+    url(r'^profile/', include('crashstats.profile.urls', namespace='profile')),
+    url(r'^documentation/', include('crashstats.documentation.urls', namespace='documentation')),
```
willkg: This is cosmetic.

willkg:
As a side note, I don't really use the admin much at all currently. All I use it for is giving people PII access--and that works with these changes.

Bugs here are low risk and we can just fix them as we encounter them. The one thing we need to make sure is 100% correct is whether the Django admin can be accessed by users who aren't tagged with `is_staff`. Everything else can be good enough.

Also, this doesn't remove anything from the old admin. I figure once we have the Django admin done, we can nix the old admin altogether.

webapp-django/crashstats/crashstats/admin.py
```diff
@@ -0,0 +1,58 @@
...
+    def user_email(self, obj):
+        return obj.user.email
+
+    def action_name(self, obj):
+        return ACTION_TO_NAME[obj.action_flag]
+
+    def has_add_permission(self, request):
+        return False
+
+    def has_change_permission(self, request, obj=None):
```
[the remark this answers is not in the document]
willkg: Ooo--I'll try that out!

webapp-django/crashstats/settings/base.py
willkg: I originally had that, but had problems with getting it to work right. I'll try again--maybe I  missed something.

willkg: Where "I originally had that" is more like "I tried that when I was doing this with Standups, but hit problems". I'll try the `AdminSite` thing in the next PR, but will leave this as is for now.

willkg:
^^^ Rebased against master and addressed all the comments.

webapp-django/crashstats/crashstats/admin.py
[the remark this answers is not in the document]
Osmose: [Oh, it's a Django 2.1 thing, whoops](https://docs.djangoproject.com/en/2.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.has_view_permission). I forgot we were still on 1.11.

I'm fine with this until the Python 3 work shakes out and we can move to Django 2.0, then.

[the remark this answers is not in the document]
willkg: Oh, neato! I'll add a FIXME comment so we "remember" later.

willkg:
Thank you!

PR #4455, opened 2018-05-22
Removed the "add release" feature from the admin panel.

AlexisDeschamps:
Fixes issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1375422

### Changes
For the removal of the `Add New Release` feature:
* Removed the template mentions (`Release` tab in the admin interface and button in `Home`)
* Removed the url dispatch
* Removed the form
* Removed the tests

### Why
 * the feature is never used for crash-stats, and was added for the sake of external users (we don't support external installs anymore)
 * it depends on a function in Postgres called `add_new_release` which we want to be able to remove

### Steps to reproduce
* Sign in with an admin account and view `/admin/`. The `Releases` tab should be gone.

### Screenshots
Removed this tab:
screen shot 2018-05-22 at 2 29 15 pm

And the button in Home:
screen shot 2018-05-22 at 2 29 31 pm


docs/howto.rst
```diff
@@ -125,18 +125,14 @@ Django needs to write its ORM tables:
    ./manage.py migrate
 
 
-Adding new products and releases
+Adding new products
```
willkg: This is going to interfere with the other PR removing products. If you weren't fond of dealing with git merges like I am, you could remove this whole section in either this PR or the other one so they don't intersect.

webapp-django/crashstats/api/views.py
```diff
@@ -120,8 +120,6 @@ def clean(self):
 
 # Names of models we don't want to serve at all
 BLACKLIST = (
-    # only used for doing posts
-    'Releases',
```
willkg: We want to keep this. It's related to the releases API. We only want to remove the "add release" form from the admin panel.

willkg:
@AlexisDeschamps This has a bunch of conflicts that need to get resolved. Can you fix those and then rebase this against the master branch? Then I can review it again.

AlexisDeschamps:
@willkg Should be rebased. Thanks.

PR #4457, opened 2018-05-23
Tweak contributor docs

willkg:
I need a place I can point new contributors to that makes sure they're
ready to contribute. This adds that.


willkg:
@lonnen Is this something you can get to? The reason I want to do this is so I have something I can easily point new contributors I don't know to make sure they have a minimal setup before I assign bugs to them.

willkg:
Thank you!

PR #4459, opened 2018-05-23
fix bug 997272: added JavaScript linting

AlexisDeschamps:
Fixes issue: https://bugzilla.mozilla.org/show_bug.cgi?id=997272
Should be good for review. It might be better to review each of the four commits individually because of noise from the autofix.

### Changes
- commit 1: added eslint integration with prettier
- commit 2: applied automatic eslint/prettier fix
- commit 3: added manual eslint:recommended fixes
  - I added projects globals for repeated globals
  - I added file globals for globals that only appeared on
  - I disabled eslint rule `no-unused-vars` for a few files that seemed to be making them on purpose... Not sure if I should try to remove them anyway
- commit 4: added eslint to CircleCI

### Why
* jshint has been used in the past but not mandated
* this commit makes the code eslint and Prettier compliant and adds the two to CircleCI

### Notes
The Prettier options were chosen according to the following:
- print width of 120: seems better according to conversations with Osmose
- single quote: the project was already using single quotes
- trailing comma (ES5): seems better according to conversations with Osmose
- the rest is default


Osmose:
@AlexisDeschamps Just an FYI, whenever this is ready for a review, assign me as a reviewer and I'll take a look.

.eslintrc
```diff
@@ -1,22 +1,29 @@
...
+  ga: false
+  location: true
+  MG: false
+  moment: false
+  PaginationUtils: true
+  Promise: false
+  Qs: false
+  SignatureReport: false
+  socorro: false
+  window: false
```
Osmose: Setting the [environment](https://eslint.org/docs/user-guide/configuring#specifying-environments) to include `browser` should negate the need for this and a few other globals in this list.

.eslintrc
```diff
@@ -1,22 +1,29 @@
...
     - 1
 env:
   browser: true
   jquery: true
-extends: 'eslint:recommended'
+globals:
+  $: false
+  alert: true
+  Analytics: false
+  BugLinks: true
```
Osmose: Ideally, I'd prefer these globals to be specified on a per-file basis. Reason being, these aren't automatically global across the whole site, they're global on pages that include the correct bundles that define them. I dunno how many files we'd have to annotate, though; if it's a ton, it might not be worth changing.

Some of these, like the `$` for jQuery, are included in the `crashstats_base` bundle ([defined here](https://github.com/mozilla-services/socorro/blob/master/webapp-django/crashstats/settings/bundles.py#L360), and included in [the base template here](https://github.com/mozilla-services/socorro/blob/master/webapp-django/crashstats/base/jinja2/crashstats_base.html#L194)) and can be safely included here since they're on _every_ page.

webapp-django/package.json
```diff
@@ -10,9 +10,13 @@
   "license": "MPL-2.0",
   "devDependencies": {
     "cssmin": "0.4.3",
+    "eslint": "^4.19.1",
```
Osmose: We prefer to pin our dependencies to a specific version. We should remove the [caret range operator `^`](https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004) which allows minor and patch version updates without our explicit approval.

webapp-django/crashstats/crashstats/static/crashstats/js/lib/accordions.js
```diff
@@ -1,4 +1,4 @@
-(function(window, undefined) {
```
Osmose: huh, wonder what that was even for

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/admin.js
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable no-unused-vars */
```
Osmose: Instead of disabling this rule, we can use the [exported](https://eslint.org/docs/rules/no-unused-vars#exported) comment to explicitly mark the functions as being exported. That lets us continue to look for unused variables while ignoring stuff we mean to export.

Sames goes for the other spots that you're disabling `no-unused-vars`.

Osmose:
@willkg Can you review ` docker/run_lints_in_docker.sh`? It's based of `run_tests_in_docker` but I don't have enough context as to why it's written the way it is to really know what's required and what's not.

docker/run_lints_in_docker.sh
```diff
@@ -0,0 +1,70 @@
...
+docker run \
+        --rm \
+        --user "${APP_UID}" \
+        --volumes-from socorro-repo \
+        --workdir /app/webapp-django \
+        --tty \
+        --interactive \
+        "${LINTIMAGE}" /webapp-frontend-deps/node_modules/.bin/eslint .
+
+echo "Done!"
```
willkg: I can see why you thought this was the right way to go, because we run flake8 in the same step we run pytest. That script is set up the way it is because pytest needs to run with other services running and we can't do that in Circle and mount volumes. So we have to create a data container and blah blah blah. The fact that flake8 runs via `run_tests.sh` isn't necessary. It'd be easier if we extricated it from `run_tests.sh`.

I think we have two possibilities here:

1. run eslint in `run_tests.sh` with everything else continuing what we've got now, or
2. run it as a separate line in `.circleci/config.yml`:
   ```
   docker run local/socorro_webapp /webapp-frontend/deps/node_modules/.bin/eslint
   ```
   We do this in Antenna currently:
   
   https://github.com/mozilla-services/antenna/blob/a700c68f288b85bb112068309c3b791f879629fd/.circleci/config.yml#L40
   
   We could even make a new make rule for this like we have for flake8 and then run that.

It's possible the first option isn't possible. I didn't look very closely.

Sorry for the confusion!

AlexisDeschamps: I transferred the JS linting to `run_tests_in_docker.sh`, since I am not sure on how to access `/webapp-frontend/deps/node_modules/.bin/eslint` in `run_tests.sh` to run `eslint`.

I'm not sure only putting it in `circleci/config.yml` would be super convenient for development, since then you would not be able to run the linting locally.

Is having the JS linting in `run_tests_in_docker.sh` fine?

willkg: I think my attempt to explain what we've currently got and why it's so weird overcomplicated things and made it harder to communicate what I'm thinking well. Sorry about that!

Running eslint should be really straight-forward because it's just running a command in a container and doesn't have the complications that running pytest has.

If you look at the "lint" rule in the `Makefile`, that's essentially all we have to do to run eslint. The difference is that we run `flake8` in the processor container and we need to run eslint in the `webapp` container.

When we run commands in a container, the command is run in the file-system context of the container--not your host machine. In the container, we have a `/webapp-frontend/deps/node_modules/.bin/eslint` binary. You can verify this:

```
$ docker-compose run webapp ls /webapp-frontend/deps/node_modules/.bin/
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                            this is run inside the container
```

You can run it like this:

```
$ docker-compose run webapp /webapp-frontend/deps/node_modules/.bin/eslint
```

I think adding an "eslint" rule to the `Makefile` and then calling that from Circle's config file seems like a good way to go here.

Does that make more sense? If not, we should switch to a video conference or something.

AlexisDeschamps:
@willkg Thank you for the explanation.

So, `${DC} run webapp /webapp-frontend-deps/node_modules/.bin/eslint . ` seems to hang forever unfortunately.

`${DC} run --workdir="/app/webapp-django" webapp /webapp-frontend-deps/node_modules/.bin/eslint . ` seems to fix the problem locally, but in CircleCI, this error appears:

`Cannot start service webapp: oci runtime error: container_linux.go:265: starting container process caused "chdir to cwd (\"/app/webapp-django\") set in config.json failed: no such file or directory"`

I tried adding `-v /app`, but that did not seem to work. Any ideas on how to fix this?

willkg:
So, I think there are two things here.

First, I'm not having any problems running eslint in my local machine like this:

```
$ docker-compose run webapp /webapp-frontend-deps/node_modules/.bin/eslint /app/webapp-django
```

That takes about 5 seconds and then writes some things to stdout and it's done. It's hard to tell for sure, but looks like you've got a `.` at the end of your first command line. `.` in the context of the webapp container will be `/app`, so it's probably running across the entire repository which probably takes a while.

I don't think you need to pass `--workdir`.

The second issue seems to be problems with running this in Circle. Circle can't load volumes with docker-compose. That's why we do that ridiculous thing with pytest. In the circle config file, I'd run it like this:

```
docker run local/socorro_webapp /webapp-frontend/deps/node_modules/.bin/eslint /app/webapp-django
```

I think that'll work. That's what I had in one of my comments, but when I wrote my second one, I forgot. Sorry about that!

This is similar to what we do with Antenna:

https://github.com/mozilla-services/antenna/blob/a700c68f288b85bb112068309c3b791f879629fd/.circleci/config.yml#L40

AlexisDeschamps:
@willkg Thanks a lot for the help. This makes more sense to me now.

Osmose helped me figure out the issue:

We ran eslint with `--debug` and it turns out that the problem was that eslint was linting `node_modules`, which I had installed locally. Hence, the process hung on my machine and CircleCI.

Eslint only ignores `/node_modules` by default, and our `node_modules` were not at the root, so they did not get ignored. Eslint documentation: https://eslint.org/docs/user-guide/configuring#ignoring-files-and-directories

So, setting the workdir to `/app/webapp-django` allows eslint to correctly ignore `node_modules` and your two commands seem to work with this addition.

We also added `static/*` to `.eslintignore` to ignore the generated static files.

willkg:
Ok--looks good to me!

PR #4460, opened 2018-05-24
fix bug 1464107: add script to run migrations

willkg:
This is pretty basic. To test:

```
$ docker-compose run crontabber ./scripts/run_migrations.sh
```

It runs in the crontabber container because that's how we do post-deploy steps right now.

After this lands, we need to update the cloudops repo. I'll take that on.


willkg:
@sciurus mentioned adding sentry support to this. I want to do that. I'm going to kill this PR and create a new one when I've got that working.

PR #4461, opened 2018-05-25
fix bug 1464107: add script to run migrations

willkg:
This codifies all the migration stuff in a single script which we'll run
as a post-deploy script.

This adds curl and gawk to the Ubuntu setup. We need curl for installing
sentry-cli and we need gawk because it's the version of awk that sentry-cli
uses.

This also adds the sentry-cli to the crontabber container. This enables
us to send errors in bash scripts to Sentry.


willkg:
To test:

1. rebuild base, webapp, and crontabber containers
   ```
   $ docker-compose build --no-cache base
   $ docker-compose build webapp
   $ docker-compose build crontabber
   ```
2. run the migration script from the local dev environment
   ```
   $ docker-compose run crontabber scripts/run_migrations.sh
   ```
3. edit `scripts/run_migrations.sh` creating an error like removing `d` from `date` and run the migration script with `RAVEN_DSN` defined (though it's a junk value)
   ```
   $ docker-compose run -e RAVEN_DSN="http://me@localhost:8000/foo" crontabber scripts/run_migrations.sh
   ```
   This is a junk value, so it will kick up errors trying to send to sentry, but it does show the error handling working.

That's pretty much all we can do to test. We'll run it on -stage after it deploys and test it further there with a valid `RAVEN_DSN` value.

docker/Dockerfile.webapp
```diff
@@ -1,7 +1,6 @@
 FROM local/socorro_base
 
 # Install nodejs and npm from Nodesource's 8.x branch
-RUN DEBIAN_FRONTEND=noninteractive apt-get install -y curl
```
willkg: I took it out here and moved it to the base container.

docker/set_up_sentry_cli.sh
```diff
@@ -0,0 +1,18 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Installs sentry-cli.
+
+set -v -e -x
+
+# Comes from https://sentry.io/get-cli/
+SENTRY_URL=https://github.com/getsentry/sentry-cli/releases/download/1.32.0/sentry-cli-Linux-x86_64
```
willkg: If they ever do an update, we'll continue to get this version. We don't have a good way of knowing whether a new version has come out. I'm not sure that's a big deal, but it might be at some point in time.

Maybe a "better" way is to get the script at https://sentry.io/get-cli/ and then parse the download line and then curl it?

sciurus:
When this script runs, we've loaded the socorro config into the environment. That means there is sensitive information like database credentials in environment variables. [This blog post](https://blog.sentry.io/2017/11/28/sentry-bash) says that wehn using sentry-cli "Event environment variables and a bunch of information about your machine will be sent along". In the [documentation](https://docs.sentry.io/learn/cli/send-event/) I don't see a way to disable this. Thus I don't think this is safe to use as-is.

In the python side we use [these settings](https://github.com/mozilla-services/socorro/blob/68d3a585bfd5574725eaf58123052d39a60e4153/webapp-django/crashstats/settings/base.py#L514-L523) to sanitize this data before sending the environment to sentry.

sciurus:
In testing, looks like all the sensitive values are filtered out server-side. See [this event](https://sentry.prod.mozaws.net/operations/socorro-new-stage/issues/4353410/) where I sent a version of the socorro environment with fake values.

docker/set_up_sentry_cli.sh
Osmose: Apparently we can get it from node as well: https://github.com/getsentry/sentry-cli#node. I'm not sure if that form is compatible with our auto-updates, hashing, or vulnerability-checking, though. If not, it's not terribly different from using this URL.

willkg: I thought about adding it to `webapp-django/package.json`. That'd use our existing infrastructure for notifying us there are new versions. However, I'm concerned about how we install a bunch of JS libs for different purposes that run in different contexts all at once. I didn't want to add the sentry cli to the webapp JS lib mix since it's only used in the crontabber container.

Looking at this more, I wonder if it's possible for us to have a `crontabber/package.json` file. Then it's separate from the webapp bits and we get our version update notification. Will that work with whatever is telling us our versions are out of date?

willkg:
^^^ @Osmose That redoes how sentry-cli is installed. It uses npm and `package.json` and all that now.

Pretty sure I got that right, but it's definitely worth double-checking the npm bits since I don't have much experience with npm.

willkg:
Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you! Thank you!

PR #4463, opened 2018-05-31
fix bug 1460707: fixed python 3 lint errors

ceddy-cedd:
Fixed linting errors via flake8. Modified code so that test pass in both
python 2 & 3. Most modifications for flake errors were done using find
and replace. New imports were added according to pep8 import
conventions. After linting errors were fixed another crontabber test
failed because explicit type checking wanted 'unicode' type. Unicode
type is only in python 2. A string was passed instead.Checked 2/3 cheat
sheet. Cheat sheet suggested to import from builtins library and use
str. But this failed because the str function doesnt provide the casting
behavior needed. So instead I coerced the string to unicode using its
own decode method. Lastly there was an error of types being used.
Buildin library types vs actual native language types. I used the six
library to import string_types which is a compatibility package for both
python 2 & 3.

### DIFF
--- a/scripts/hive_adi_test.py
+++ b/scripts/hive_adi_test.py
@@ -8,6 +8,8 @@ import pyhs2
 import tempfile
 import unicodedata
 import urllib2
+from past.builtins import basestring
+from builtins import str
 
 # Example command-line usage:
 # $ . /etc/socorro/socorrorc
@@ -77,7 +79,7 @@ def main():
 
     def remove_control_characters(s):
         if isinstance(s, str):
-            s = unicode(s, 'utf-8', errors='replace')
+            s = str(s, 'utf-8', errors='replace')
         return ''.join(c for c in s if unicodedata.category(c)[0] != "C")
 
     with codecs.open(options.output_filename, 'w', 'utf-8') as f:
--- a/socorro/cron/jobs/fetch_adi_from_hive.py
+++ b/socorro/cron/jobs/fetch_adi_from_hive.py
@@ -32,6 +32,8 @@ import os
 import tempfile
 import unicodedata
 import urllib2
+from past.builtins import basestring
+from past.builtins import str
 
 import pyhs2
 
@@ -238,7 +240,7 @@ class FetchADIFromHiveCronApp(BaseCronApp):
     @staticmethod
     def remove_control_characters(s):
         if isinstance(s, str):
-            s = unicode(s, 'utf-8', errors='replace')
+            s = str(s, 'utf-8', errors='replace')
         return ''.join(c for c in s if unicodedata.category(c)[0] != "C")
 
     def _database_transaction(
--- a/socorro/cron/jobs/fetch_adi_from_hive.py
+++ b/socorro/cron/jobs/fetch_adi_from_hive.py
@@ -36,6 +36,7 @@ from past.builtins import basestring
 from past.builtins import str
 
 import pyhs2
+from six import text_type
 
 from configman import Namespace, class_converter
 from crontabber.base import BaseCronApp
@@ -240,7 +241,7 @@ class FetchADIFromHiveCronApp(BaseCronApp):
     @staticmethod
     def remove_control_characters(s):
         if isinstance(s, str):
-            s = str(s, 'utf-8', errors='replace')
+            s = text_type(s, 'utf-8', errors='replace')
         return ''.join(c for c in s if unicodedata.category(c)[0] != "C")
 
     def _database_transaction(
--- a/socorro/external/postgresql/setupdb_app.py
+++ b/socorro/external/postgresql/setupdb_app.py
@@ -200,7 +200,7 @@ class SocorroDBApp(App):
                 autocommit=False
         ) as db:
             if 'test' not in database_name and not self.config.force:
-                confirm = raw_input(
+                confirm = input(
                     'drop database %s [y/N]: ' % database_name
                 )
                 if not confirm == "y":
--- a/socorro/external/postgresql/setupdb_app.py
+++ b/socorro/external/postgresql/setupdb_app.py
@@ -12,12 +12,13 @@ import os
 import re
 import sys
 
+import six
 from alembic import command
 from alembic.config import Config
 from configman import Namespace, class_converter
+from six.moves import cStringIO
 from sqlalchemy.ext.compiler import compiles
 from sqlalchemy.schema import CreateTable
-from six.moves import cStringIO
 
 from socorro.app.socorro_app import App
 from socorro.external.postgresql import staticdata
@@ -200,7 +201,7 @@ class SocorroDBApp(App):
                 autocommit=False
         ) as db:
             if 'test' not in database_name and not self.config.force:
-                confirm = input(
+                confirm = six.moves.input(
                     'drop database %s [y/N]: ' % database_name
                 )
                 if not confirm == "y":
--- a/webapp-django/crashstats/crashstats/tests/test_utils.py
+++ b/webapp-django/crashstats/crashstats/tests/test_utils.py
@@ -1,13 +1,12 @@
 import copy
 import datetime
-from builtins import str
 from cStringIO import StringIO
 from unittest import TestCase
 import json
 
 from django.http import HttpResponse
 from django.test.client import RequestFactory
-from six import string_types
+from six import string_types, text_type
 
 from crashstats.crashstats import utils
 
@@ -526,7 +525,7 @@ class TestUtils(TestCase):
 
         result = out.getvalue()
         assert isinstance(result, string_types)
-        u_result = str(result, 'utf-8')
+        u_result = text_type(result, 'utf-8')
         assert 'abc,' in u_result
         assert u'\xe4\xc3,' in u_result
         assert '123,' in u_result
--- a/webapp-django/crashstats/crashstats/tests/test_utils.py
+++ b/webapp-django/crashstats/crashstats/tests/test_utils.py
@@ -6,7 +6,7 @@ import json
 
 from django.http import HttpResponse
 from django.test.client import RequestFactory
-from six import string_types, text_type
+from six import text_type, binary_type
 
 from crashstats.crashstats import utils
 
@@ -522,9 +522,11 @@ class TestUtils(TestCase):
             123,
             1.23,
         ])
-
+        #NOTE: This probably needs to be a StringIO in Python 2 and a BytesIO in Python 3
+        # assuming UnicodeWriter even works in Python 3. I made change to binary_type for
+        # a simple fix until we run python 3 harness test in future.
         result = out.getvalue()
-        assert isinstance(result, string_types)
+        assert isinstance(result, binary_type)
         u_result = text_type(result, 'utf-8')
         assert 'abc,' in u_result
         assert u'\xe4\xc3,' in u_result
--- a/webapp-django/crashstats/crashstats/tests/test_utils.py
+++ b/webapp-django/crashstats/crashstats/tests/test_utils.py
@@ -1,11 +1,13 @@
 import copy
 import datetime
+from builtins import str
 from cStringIO import StringIO
 from unittest import TestCase
 import json
 
 from django.http import HttpResponse
 from django.test.client import RequestFactory
+from six import string_types
 
 from crashstats.crashstats import utils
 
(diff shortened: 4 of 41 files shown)


scripts/hive_adi_test.py
```diff
@@ -77,7 +79,7 @@ def main():
 
     def remove_control_characters(s):
         if isinstance(s, str):
-            s = unicode(s, 'utf-8', errors='replace')
+            s = str(s, 'utf-8', errors='replace')
```
lonnen: It looks like some of your changes didn't make it into commit. I recall this was:
```python
s = s.decode('utf-8', errors='replace')
```

socorro/cron/jobs/fetch_adi_from_hive.py
```diff
@@ -238,7 +240,7 @@ class FetchADIFromHiveCronApp(BaseCronApp):
     @staticmethod
     def remove_control_characters(s):
         if isinstance(s, str):
-            s = unicode(s, 'utf-8', errors='replace')
+            s = str(s, 'utf-8', errors='replace')
```
willkg: I'd much rather we were using `six`. Otherwise this looks totally wrong in Python 2 which is our primary context.

In six, this would be:

```
s = six.text_type(s, 'utf-8', errors='replace')
```

socorro/external/postgresql/setupdb_app.py
```diff
@@ -200,7 +200,7 @@ def handle_dropdb(self, superuser_pg_url, database_name):
                 autocommit=False
         ) as db:
             if 'test' not in database_name and not self.config.force:
-                confirm = raw_input(
+                confirm = input(
```
willkg: This changes the behavior in Python 2 which has both an `input` and a `raw_input`. We should use `six` here:

```
confirm = six.moves.input(
```

When run in Python 2, that'll use `raw_input` and when run in Python 3, that'll use `input`. That's what we want here.

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -7,6 +7,7 @@
 import json
 import random
 import urlparse
+from builtins import int
```
willkg: `int` will always be in the namespace--you don't need to import it.

```
$ python
[GCC 7.3.0] on linux2
>>> int
<type 'int'>
>>> 
```

```
$ python3
Python 3.6.5 (default, Apr  1 2018, 05:46:30) 
[GCC 7.3.0] on linux
>>> int
<class 'int'>
>>> 
```

As a side note, Python has a read-eval-print-loop (often called "repl") that lets you play with the language and code iteratively. I often use this to test things out. You can usually run it by typing "python" in a terminal/shell. Some systems have Python 2 and Python 3 installed and "python" will be Python 2  and "python3" will be Python 3. Python will tell you which version it is at startup.

I don't know if that's helpful to know or not.

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -770,7 +771,7 @@ def mocked_adi_get(**options):
                 for version in ['20.0', '19.0']:
                     for build_type in ['beta', 'release']:
                         response['hits'].append({
-                            'adi_count': long(random.randint(100, 1000)),
+                            'adi_count': int(random.randint(100, 1000)),
```
willkg: `random.randint` will always return an `int`, so you don't need to convert it.

webapp-django/crashstats/crashstats/utils.py
```diff
@@ -318,7 +320,7 @@ def __init__(self, f, dialect=csv.excel, encoding="utf-8", **kwds):
         self.encoder = codecs.getincrementalencoder(encoding)()
 
     def writerow(self, row):
-        self.writer.writerow([unicode(s).encode("utf-8") for s in row])
+        self.writer.writerow([str(s).encode("utf-8") for s in row])
```
willkg: This doesn't work in Python 2. It needs to be a unicode--not a str. You can use `six.text_type` here and that'll work.

willkg: This is true of most of the changes from `unicode` to `str` in this PR. I'll leave off commenting on the others.

willkg:
This is great work! Porting an old codebase (which this is) from Python 2 to 3 is really tricky. It'll take a bunch of iterations as we figure out what the old code was doing.

webapp-django/crashstats/crashstats/tests/test_utils.py
```diff
@@ -523,8 +524,8 @@ def test_unicode_writer(self):
         ])
 
         result = out.getvalue()
-        assert isinstance(result, str)
-        u_result = unicode(result, 'utf-8')
+        assert isinstance(result, string_types)
```
willkg: This looked wrong to me because the original code is testing for `str`--not `basestring` which is `str` + `unicode`. I think this should be something like:

```
assert isinstance(result, six.binary_type)
```

After writing that, I got to wondering what the rest of the test is doing such that it's asserting something is a bytes here because that's odd. I read through the test and looked up some things like what does `StringIO.getvalue()` return in Python 2 and Python 3. I think we need to make other changes to this test.

The problem is that `StringIO` is being used to hold a stream of bytes--not characters. In Python 2, that's a `str` and `StringIO`. In Python 3, that's `bytes` and `BytesIO`. So we need to use one set of things in Python 2 and the other set in Python 3.

So I think this probably needs to be something like this:

```
def test_unicode_writer(self):
    if six.PY2:
        out = StringIO()
    else:
        out = BytesIO()
    writer = utils.UnicodeWriter(out)
    writer.writerow([
        'abc',
        u'\xe4\xc3',
        123,
        1.23,
    ])
    result = out.getvalue()
    assert isinstance(result, six.binary_type)
    u_result = result.decode('utf-8')
    # and so on
```

I'm pretty sure this is true, but it's hard to know for sure because we don't have this test running in the Python 3 harness, yet. Thus we can't make some changes and see whether they work in Python 2 and Python 3. In order to do that, we need to make a bunch of other changes first which we can't do until we've fixed linting issues. Also (and I bring this up purely as a "haha--wouldn't it be funny if ..." exercise) it's entirely possible `UnicodeWriter` doesn't work at all in Python 3. So much fun!

Given that we can't run this in the Python 3 test harness and know whether it's correct or not, I vote we slink back to a happier place and make the smallest change to move forward that we know works in Python 2 because that's what's running in production. I think the simplest fix is a minor tweak on what you have:

```
assert isinstance(result, six.bytes_type)
```

Then we leave the rest of this "hot mess" to future us. Because we like us and don't want future us disparaging current us, we should leave ourselves a note along the lines of:

```
# NOTE: This probably needs to be a StringIO in Python 2 and a BytesIO in Python 3
# assuming UnicodeWriter even works in Python 3.
```

I think that's right.

webapp-django/crashstats/supersearch/form_fields.py
```diff
@@ -46,7 +46,7 @@ class PrefixedField(object):
     prefixed_value = None
 
     def to_python(self, value):
-        if isinstance(value, basestring):
+        if isinstance(value, str):
```
willkg: If the original is `basestring`, then we want to use `six.string_types`.

willkg:
Looking good! Almost there!

Also, if it's the case that I'm butting into this PR and shouldn't be, please let me know! I find Python 2to3 porting work fascinating, plus this is a super important project. It's like having a chocolate cake in the refrigerator: it's so easy to convince myself that having one more byte benefits everyone and it's the right thing to do. But maybe it's someone else's cake? Maybe they were saving it for a party? Maybe I've ruined everything? Oh noes!

ceddy-cedd:
@willkg Waiting for review, I think all comments are addressed.

willkg:
@ceddy-cedd Some of the code has changed since you started your branch. Can you rebase your branch off of the current master tip and then force-push it to this PR branch?

I don't know offhand if you've rebased before. On the off chance you haven't, I tend to do something like this:

```
$ git checkout master
$ git pull
$ git checkout 1460707-fix-python-3-linting-errors
$ git rebase master
```

I'm pretty sure there will be merge conflicts at this point. Things changed in the master branch that overlap with changes you've made. You'll need to sort that out with some additional code changes and `git rebase` interaction.

After that's done, you'll have a rebased branch and you can force push to the PR branch:

```
$ git push --force REMOTE 1460707-fix-python-3-linting-errors
```

Replace `REMOTE` with whatever you named your remote.

Once you've rebased the branch, let me know and I can review it.

ceddy-cedd:
@willkg Hey Will Lonnen and I fixed the branch conflict, I think its ready for review now.

scripts/hive_adi_test.py
willkg: Previously, this if-block would check to see if the value was a `str` and convert it to a `unicode`. Now it's kind of silly in that it checks to see if the value is a `str` and if so, converts it to a `str`.

So, what is the big picture here? What is this code doing? What's the context? Can `s` have a binary non-character value? If we ran this code in Python 3, could it be `bytes`? Or is it the case that in Python 2, it was sometimes a string and sometimes a unicode--but always character data?

This function is removing control characters. Why would the value have control characters in it? How curious. The next line converts the result back to a `str`. Even more curious.

I think what we want here is to convert it to a unicode if and only if we're in Python 2 and the value is a str. If we're in Python 3 or if the value is unicode, then we can leave it as is--we don't need to convert it.

So maybe something like this works?:

```
if six.PY2 and isinstance(s, str):
    s = six.text_type(s, 'utf-8', errors='replace')
```

Taking a bigger step back, we don't actually use this file at all and there are no tests for it--it's effectively half-dead code that's seen better days and is waiting for @lonnen to give us the thumbs-up to remove all the ADI stuff.

Thus, nothing we do to this file has any impact on the world so we can leave it as is or make the change you suggested or do a different change--it's all fine.

willkg: Oops--we can't leave it as is since we need to get linting to pass, so the options are:

1. keep the change you have
2. make a different change

socorro/cron/jobs/fetch_adi_from_hive.py
```diff
@@ -238,7 +241,7 @@ class FetchADIFromHiveCronApp(BaseCronApp):
     @staticmethod
     def remove_control_characters(s):
         if isinstance(s, str):
-            s = unicode(s, 'utf-8', errors='replace')
+            s = text_type(s, 'utf-8', errors='replace')
```
willkg: Similar to the change in `scripts/hive_adi_test.py` this code is mostly dead. Technically, code like it runs on a server in a data center somewhere, but that code is maintained by someone logging into the server and "fixing" it live--it's not maintained by making changes in this repository. I copied the most recent version of the file here just so I had a copy of it somewhere if that server died.

Fun fact--that server is running Python 2.6--it's not even running Python 2.7. It will never run Python 3. So given all that, any change we make here is probably fine, too. If I have to go fix the code running on that server, I'll take the obligation to make changes that work with linting here as well as on the server there.

webapp-django/crashstats/supersearch/form_fields.py
```diff
@@ -64,7 +65,7 @@ def value_to_string(self, value):
         """Return the value as a string. """
         if value is None:
             return None
-        return unicode(value)
+        return str(value)
```
willkg: This should probably be `six.text_type`. It needs to be `unicode` in Python 2, but `str` in Python 3.

willkg:
I had one issue that should get fixed and then a lot of pontificating on existential nature of fixing ADI-related code that never runs.

Looking good!

ceddy-cedd:
@willkg  Alright final comments addressed. I know youve heard it before lol Hopefully this is the one!

willkg:
I've approved this. It's on me to get this merged. I have to figure out how to do that with our restriction on signed commits. I'll do that tomorrow.

lonnen:
@willkg I can help @ceddy-cedd set up GPG and wire it into his git config as the default in the AM. Then we'll amend (and squash, if you want?) these commits to have a signature.

willkg:
@lonnen If you want to do that, that'd be cool. This is a more general workflow problem with all contributors that @Osmose and I were talking about, so I'm thinking about it holistically and not just about this specific PR.

lonnen:
https://wiki.mozilla.org/GitHub/Repository_Security/Problems_and_Options
has an answer if you wanted to try it in this branch

willkg:
Thank you! Good job!

PR #4464, opened 2018-05-31
bug 1462749: remove unnneeeded admin code

willkg:
This removes unneeded admin code. Most of this is now done with the Django admin. I also nixed the reprocessing view since we can reprocess via other methods--no sense in having a third way. I also removed two forms that were orphaned in previous code removal passes.

I was going to reimplement the pages that aren't backed by a Django model in the Django admin, but this PR is like 2800 lines now and it seemed like a good idea to do one PR that's entirely code removal before doing another one that's more complicated.


webapp-django/crashstats/manage/views.py
```diff
@@ -1,78 +1,33 @@
...
-    raise NotImplementedError(before.__class__.__name__)
+def log_action(user_id, action_flag, change_message):
+    LogEntry.objects.create(
+        user_id=user_id,
+        content_type_id=None,
+        object_id=None,
+        object_repr='',
+        action_flag=action_flag,
+        change_message=change_message
+    )
```
willkg: Previously, we logged actions with the pinax eventlog module. This changes things so we're logging actions using the Django admin LogEntry thing. It's definitely not as ergonomic and there's like no documentation for it. The `LogEntryManager` has a helper method on it, but requires you to pass in an `object_id` and `content_type_id`, so it doesn't work for us.

So... that's why this is here.

Pretty sure the only thing we need it for is the graphics devices stuff. I want to move that table over to the Django admin. Once I do that, then we can manage it with the admin and won't need this weirdo code. I've never added graphics devices things. So maybe we should just nix this now and not worry about it?

willkg: Hrm. I think I just convinced myself it's a good idea to just nix it--less to deal with, less to test.

Honk your horn twice and retweet if you think that's a bad idea.

willkg:
^^^ I dropped the auditing code for the graphics devices stuff. So, that's that.

Osmose:
> I'm running into an error running this locally.

This might be an issue with my local setup, going to burn down my image cache and try again.

willkg:
Thank you!

PR #4466, opened 2018-06-01
fix bug 1440233 - ignore rust hashes in signature generation

staktrace:
The rust compiler generates a hash/fingerprint at the end of the
function names in order to distinguish between different versions of a
crate built together into the final product. This hash also changes
frequently between builds of Firefox, so it's better to strip it out
when generating signatures.

### DIFF
--- a/socorro/signature/rules.py
+++ b/socorro/signature/rules.py
@@ -108,6 +108,7 @@ class CSignatureTool(SignatureTool):
 
         self.fixup_space = re.compile(r' (?=[\*&,])')
         self.fixup_comma = re.compile(r',(?! )')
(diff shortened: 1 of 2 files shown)


willkg:
We now require all commits to be signed. That's a new thing as of late last week and I haven't figured out how that works in a world where we want to have a low bar for contributing, so there's nothing in the docs about it, yet.

I wanted to take this opportunity to test out a workflow, so I cherry-picked your commit, signed it, and created a new PR. Closing this one in favor of pr #4467 which has your commit in it.

willkg:
I can't get it to merge without signed commits. Closing this out again in favor of PR #4467.

PR #4467, opened 2018-06-04
fix bug 1440233 - ignore rust hashes in signature generation

willkg:
This includes the commits from PR #4466, but signed by me.

We just enabled "require signed commits" on the repo, so we have to figure out the workflow for working around that for contributors. I think this might be the best one that doesn't require us to change how people normally contribute.

To "test" this, you can either take my word for it (I used `git cherry-pick`) or verify it's the same changes that are in PR #4466.


willkg:
Thank you!

PR #4468, opened 2018-06-04
fix bug 1465908: add wcsrtombs to prefix list

willkg:
This adds wcsrtombs to the prefix list so signature generation continues.

Example output:

```
you@processor:/app$ python -m socorro.signature 129cbcd4-248a-45c9-9b89-3a4460180604
Crash id: 129cbcd4-248a-45c9-9b89-3a4460180604
Original: wcsrtombs
New:      wcsrtombs | wcsrtombs | wcsrtombs | js::CloneFunctionReuseScript
Same?:    False
```


willkg:
Thank you!

PR #4470, opened 2018-06-06
bug 1403124: When correlations are missing for a channel, suggest looking at a different one

is2ei:
1. This PR improves the message.
2. Added message.
3. Run webapp and access to http://localhost:8000/signature/?signature=test#correlations

### DIFF
--- a/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
+++ b/webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
@@ -190,15 +190,16 @@ window.correlations = (function() {
         var signatureData = data[product][channel].signatures[signature];
 
         if (!signatureData || !signatureData.results) {
-          return (
+          return [
             'No correlation data was generated for the signature "' +
-            signature +
-            '" on the "' +
-            channel +
-            '" channel, for the "' +
-            product +
-            '" product.'
-          );
+              signature +
+              '" on the "' +
+              channel +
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/crashstats/static/crashstats/js/socorro/correlation.js
```diff
@@ -197,7 +197,7 @@ window.correlations = (function() {
             channel +
             '" channel, for the "' +
             product +
-            '" product.'
+            '" product. The number of crashes was probably too low in this channel. Please try to select another channel.'
```
Osmose: If the signature is long, the text overflows outside the container:

screen shot 2018-06-06 at 2 30 03 pm

This isn't a problem with this PR per se, it's just made more apparently now that the text is longer.

The root fix for this would be to fix the CSS/HTML around that tag and allow it to wrap. I dunno if we don't want it to be a `<pre>` tag though, as that might break the normal correlation output. An acceptable fix for the purposes of this PR would be to add a newline or two in front of the new string in order to break it out into its own line.

is2ei: Thanks! I will fix it.

is2ei:
Hi @Osmose 

I fixed it. There are 2 changes.

1. Changed the message according to your comment on Bugzilla.
2. Added new line for the message.

Please let me know if you have any concern! Thanks.

PR #4471, opened 2018-06-06
fix bug 1325166: switched from Google Sign-In to OpenID Connect

AlexisDeschamps:
fixes issues: https://bugzilla.mozilla.org/show_bug.cgi?id=1325166

## what
Migrated from Google Sign-In to OpenID Connect:
- added mozilla-django-oidc (https://github.com/mozilla/mozilla-django-oidc) library
    - it allows for integration with OpenID Connect enabled authentication services
- added `oidcprovider` docker container
    - it provides a docker setup for Open ID Connect for local development
    - it is based off the `mozillaparsys/oidc_testprovider` docker image from https://github.com/mozilla-parsys/docker-test-mozilla-django-oidc
- removed now unused Google Sign-In code

## why
- makes codebase more maintable by removing a lot of Google Sign-In code
- makes it easier to give access to non mozilla.com people without them needing Google account

## screenshots
**login button**
before:
screen shot 2018-06-06 at 3 42 28 pm

after:
screen shot 2018-06-06 at 3 44 41 pm


docker-compose.yml
```diff
@@ -132,7 +144,7 @@ services:
   statsd:
     image: hopsoft/graphite-statsd
     ports:
-      - "8080:80"
+      - "8081:80"
```
AlexisDeschamps: I would change this container since `oidcprovider` would be taking port `8080`

webapp-django/crashstats/base/jinja2/crashstats_base.html
```diff
@@ -171,18 +166,13 @@ <h1>Product Navigation</h1>
...
-                    {% endif %}
+                        <input class="text-submit-input" type="submit" value="Sign out"/>
+                    </form>
                 {% else %}
                     {% csrf_token %}
-                    <div
-                        id="signin2"
-                        class="google-signin"
-                        data-signin-url="{{ url('auth:oauth2_signin') }}"></div>
+                    <a href="{{ url('oidc_authentication_init') }}">Sign in</a>
```
Osmose: Logging in with this link redirects me back to the homepage. mozilla-django-oidc looks for a [query parameter (`next` by default) that specifies what URL to direct the user to after logging in](https://mozilla-django-oidc.readthedocs.io/en/stable/settings.html#OIDC_REDIRECT_FIELD_NAME).

For constructing a properly escaped querystring, we have a helper function called [`make_query_string`](https://github.com/mozilla-services/socorro/blob/master/webapp-django/crashstats/base/templatetags/jinja_helpers.py#L101):

```jinja
{{ url('foobar') }}?{{ make_query_string(param='value') }}
```

For getting the current URL as a value, the `request` variable is available in templates, and the [`build_absolute_uri`](https://docs.djangoproject.com/en/2.0/ref/request-response/#django.http.HttpRequest.build_absolute_uri) function is available on Django requests to get the current URL, including the querystring.

Osmose:
As for the test failures, my first guess would be that the affected tests are being tested using a fake login, and the `SessionRefresh` middleware you added is attempting to re-authenticate whenever you make a request, probably due to the [token expiration](https://github.com/mozilla/mozilla-django-oidc/blob/master/mozilla_django_oidc/middleware.py#L88) being missing from the session.

One potential solution could be to write your own middleware that subclasses `SessionRefresh`, and overrides `is_refreshable_url` to return `False` whenever the tests are running so that we don't bother re-authing users during tests.

Alternatively, you could modify `webapp-django/crashstats/settings/test.py` (which are the settings used during a test run) to remove that specific middleware during the tests only.

Let me know if that helps with the tests or if you're still hitting issues.

docker/config/oidcprovider-fixtures.json
```diff
@@ -0,0 +1 @@
+[{"model": "oidc_provider.client", "pk": 1, "fields": {"name": "testrpHS256", "client_type": "confidential", "client_id": "1", "client_secret": "bd01adf93cfb", "response_type": "code", "jwt_alg": "HS256", "date_created": "2017-11-10", "website_url": "", "terms_url": "", "contact_email": "", "logo": "", "reuse_consent": true, "require_consent": true, "_redirect_uris": "http://localhost:8000/oidc/callback/", "_post_logout_redirect_uris": ""}}, {"model": "oidc_provider.client", "pk": 2, "fields": {"name": "testrpRS256", "client_type": "confidential", "client_id": "2", "client_secret": "a6b4dad2f215", "response_type": "code", "jwt_alg": "RS256", "date_created": "2017-11-10", "website_url": "", "terms_url": "", "contact_email": "", "logo": "", "reuse_consent": true, "require_consent": true, "_redirect_uris": "http://localhost:8000/oidc/callback/", "_post_logout_redirect_uris": ""}}]
```
willkg: JSON doesn't care about whitespace, but people do. This would be a lot easier to read if it was structured and indented.

```
[
    {
        "model": "oidc_provider.client",
        "pk": 1,
        "fields": {
            "name": "testrpHS256",
            "client_type": "confidential",
            "client_id": "1",
            "client_secret": "bd01adf93cfb",
            "response_type": "code",
            "jwt_alg": "HS256",
            "date_created": "2017-11-10",
            "website_url": "",
            "terms_url": "",
            "contact_email": "",
            "logo": "",
            "reuse_consent": true,
            "require_consent": true,
            "_redirect_uris": "http://localhost:8000/oidc/callback/",
            "_post_logout_redirect_uris": ""
        }
    },
    {
        "model": "oidc_provider.client",
        "pk": 2,
        "fields": {
            "name": "testrpRS256",
            "client_type": "confidential",
            "client_id": "2",
            "client_secret": "a6b4dad2f215",
            "response_type": "code",
            "jwt_alg": "RS256",
            "date_created": "2017-11-10",
            "website_url": "",
            "terms_url": "",
            "contact_email": "",
            "logo": "",
            "reuse_consent": true,
            "require_consent": true,
            "_redirect_uris": "http://localhost:8000/oidc/callback/",
            "_post_logout_redirect_uris": ""
        }
    }
]
```

webapp-django/crashstats/base/jinja2/crashstats_base.html
```diff
@@ -171,18 +166,13 @@ <h1>Product Navigation</h1>
...
                     {% endif %}
                     |
-                    {% if request.session.get('signin_method') == 'google' %}
+                    <form action="/oidc/logout/" style="display: inline" method="POST">
                         {% csrf_token %}
-                        <a href="{{ url('auth:oauth2_signout') }}" class="google-signout">
-                            <span>Sign out</span>
-                        </a>
-                    {% endif %}
+                        <input class="text-submit-input" type="submit" value="Sign out"/>
```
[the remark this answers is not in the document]
AlexisDeschamps: Ah, I was going for `text-submit-input` meaning a class that makes "an `input` tag of type `submit` that is in `text`. And have that be general and reusable.

webapp-django/crashstats/base/jinja2/footer_nav.html
```diff
@@ -9,11 +9,9 @@
...
         <li><a href="https://www.mozilla.org/privacy/websites/">Privacy Policy</a></li>
         <li>
-            <a
-                href="{{ GOOGLE_AUTH_HELP_URL }}"
-                class="google-signin-help">
-                Google Sign-In Help
-            </a>
+            Socorro is transitioning from Google Auth to Mozilla SSO on 2018-06-19.
+            We don't think this will affect you, but if you have problems, please write up a
+            <a href="https://bugzilla.mozilla.org/enter_bug.cgi?product=Socorro">bug</a>.
```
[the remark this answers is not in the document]
AlexisDeschamps: Oh, haha.

docker-compose.yml
AlexisDeschamps: I see, thanks for the explanation. I feel like this might be simpler so I am opting to leaving it at is. I am up for changing if it becomes a problem though.

AlexisDeschamps:
Thanks a ton @Osmose  and Will. I added changes and replied in a few places..

I fixed the tests by adding a `REMOVED_MIDDLEWARE` environment variable.

`REMOVED_MIDDLEWARE = config('REMOVED_MIDDLEWARE', [], cast=Csv())` doesn't seem to work for some reason. Seems that it tries to run `Csv()` and fails if the value is empty. So, I added a `if os.environ.get('REMOVED_MIDDLEWARE') is not None:` check.

webapp-django/crashstats/base/jinja2/crashstats_base.html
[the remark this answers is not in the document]
Osmose: Yeah, that makes sense, but in practice those types of classes often end up being difficult to maintain. It's a balance to choose between "mixin classes" like that or semantic classes, but I tend to lean more towards the semantic side.

webapp-django/crashstats/base/jinja2/crashstats_base.html
Osmose: My expectation (that I haven't confirmed) would be that the `action` URL in the form for logging out also takes a `next` parameter.

webapp-django/crashstats/settings/base.py
```diff
@@ -125,13 +127,20 @@ def path(*dirs):
     'crashstats.tokens.middleware.APIAuthenticationMiddleware',
     'crashstats.crashstats.middleware.Pretty400Errors',
 )
+MIDDLEWARE = list(MIDDLEWARE)
```
Osmose: We can just declare `MIDDLEWARE` as a list from the start instead of converting it.

webapp-django/crashstats/settings/base.py
```diff
@@ -125,13 +127,20 @@ def path(*dirs):
...
+
+if os.environ.get('REMOVED_MIDDLEWARE') is not None:
+    REMOVED_MIDDLEWARE = config('REMOVED_MIDDLEWARE', cast=Csv())
+    for middleware_to_remove in REMOVED_MIDDLEWARE:
+        MIDDLEWARE.remove(middleware_to_remove)
```
Osmose: I don't think we want to have an environment variable that can remove certain middleware; I could see a bad config value removing a middleware that's critical to the site functioning, and we only need to remove one middleware, and only during tests. We should be a bit more conservative in what we allow to be changed.

`webapp-django/crashstats/settings/test.py` contains some extra settings that get applied on top of `base.py` during test runs (this works because we tell Django that the settings are at `crashstats.settings`, which imports `webapp-django/crashstats/settings/__init__.py`, where the magic is). In `test.py` we could modify `MIDDLEWARE` to remove `SessionRefresh` manually. That way there's no config at all that can modify the middleware list.

docs/service/webapp.rst
```diff
@@ -58,40 +58,18 @@ Run this::
...
 
-In order to authenticate in the webapp using the "Google signin" button, you
-need to get a set of oauth credentials.
+In order to authenticate in the webapp using the "Login" button, you
+need to add the following entry to your ``/etc/hosts`` (or equivalent) file:
 
-1. Go to https://console.developers.google.com/apis/credentials
+```
+127.0.0.1 oidcprovider
+```
```
Osmose: This is the Markdown syntax for a literal block, but this file (and the rest of the docs) are in [reStructuredText](http://docutils.sourceforge.net/docs/user/rst/quickstart.html) format. We'll want to use the syntax for a preformatted block for rst:

```rst
In order to authenticate in the webapp using the "Login" button, you
need to add the following entry to your ``/etc/hosts`` (or equivalent) file::
  
   127.0.0.1 oidcprovider
```

Note the double colon at the end of the preceding paragraph. That plus the indent is the syntax in rst for a preformatted block of text.

webapp-django/crashstats/base/jinja2/crashstats_base.html
```diff
@@ -171,18 +166,12 @@ <h1>Product Navigation</h1>
                     <a href="{{ url('manage:home') }}">Admin</a>
                     {% endif %}
                     |
-                    {% if request.session.get('signin_method') == 'google' %}
+                    <form class="logout-form" action="/oidc/logout/" method="POST">
```
Osmose: I missed mentioning this earlier, but we'll want to use the `url` helper for generating the URL here as well.

AlexisDeschamps:
@Osmose I added some updates. Thanks again!

webapp-django/crashstats/settings/base.py
```diff
@@ -648,7 +642,7 @@ def path(*dirs):
 OIDC_OP_AUTHORIZATION_ENDPOINT = config('OIDC_OP_AUTHORIZATION_ENDPOINT', '')
 OIDC_OP_TOKEN_ENDPOINT = config('OIDC_OP_TOKEN_ENDPOINT', '')
 OIDC_OP_USER_ENDPOINT = config('OIDC_OP_USER_ENDPOINT', '')
-LOGOUT_REDIRECT_URL = '/'
+OIDC_OP_LOGOUT_URL_METHOD = 'crashstats.authentication.utils.redirect_after_logout'
```
Osmose: So I was totally wrong about this; we have to be careful about sanitizing the input coming in from the next parameter. Given that, it'd be easier to just make users go back to the homepage after logging out. My bad.

webapp-django/crashstats/settings/test.py
```diff
@@ -3,6 +3,7 @@
 """
 
 from crashstats.settings.base import *  # noqa
+from crashstats.settings.base import MIDDLEWARE
```
Osmose: The line above this already imports `MIDDLEWARE`, we can remove this import thanks to that.

AlexisDeschamps:
@Osmose Updated added.

webapp-django/crashstats/base/jinja2/crashstats_base.html
Osmose: The URL helper disappeared, but we still want to use it here.

willkg:
Woo hoo!

PR #4475, opened 2018-06-09
Bug 1463814: crontabber should spit out version.json on startup

is2ei:
This issue is to make crontabber to spit out the content of `version.json`.
This PR changes crontabber to spit out the content of `version.json` if the file exists.

You can check it by:
1. Provide version.json 
2. Run crontabber.


socorro/cron/crontabber_app.py
```diff
@@ -177,6 +184,11 @@ def mark_success(self, job_list_string):
             self._remember_success(job_class, success_date=now, duration=0)
 
     def main(self):
+        version_json_path = os.path.join(ROOT, "version.json")
+        if os.path.exists(version_json_path):
+            with open(version_json_path, 'r') as fp:
+                self.config.logger.info('version.json: ' + fp.read().strip())
+
```
willkg: I fiddled with the code a bit and I think my idea of putting this in `SocorroApp` is a better solution. It covers both crontabber and processor apps so we don't need to do this twice. It also prints out the version information before all the configuration information. That's pretty handy.

Can you move it there, please?

is2ei: @willkg Thanks for reviewing!
I moved the function. Could you check it, please?

socorro/app/socorro_app.py
```diff
@@ -202,6 +215,7 @@ def fix_exit_code(code):
                     respond_to_SIGHUP,
                     logger=config.logger
                 )
+                spit_out_version_json(config.logger)
```
willkg: It helps me if the version information is logged before all the config information. I'd move this line above the `try` block. Can you do that, please?

is2ei: @willkg Sure! I fixed it. Could you check it, please?

PR #4476, opened 2018-06-09
Bug 1091729: Use json dumps in test fixtures instead of pipe dumps

is2ei:
This PR is to use json dumps in test fixtures instead of pipe dumps.
To test it, please run tests.


webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -1387,8 +1395,6 @@ def mocked_processed_crash_get(**params):
         assert 'peterbe@example.com' in response.content
         assert _SAMPLE_META['Email'] in response.content
         assert _SAMPLE_META['URL'] in response.content
-        assert '&#34;sensitive&#34;' in response.content
-        assert '&#34;exploitability&#34;' in response.content
```
willkg: I'm not sure I understand why these lines got removed after switching the format we're specifying the dump in. Any ideas?

is2ei: @willkg Thanks for reviewing!

I found I should not remove those lines and need to modify the test data.
I updated it. Could you check it, please?

BTW, I found the test fails due to docker issue.
https://circleci.com/gh/mozilla-services/socorro/2851?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #4477, opened 2018-06-13
Bug 1468435: circle config: switch base image to docker:stable, install compose via pip

milescrabill:
`docker:stable` is based on alpine, hence changing package manager invocations.

`curl` is no longer needed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,15 +2,23 @@ version: 2
 jobs:
   build:
     docker:
-      # base image
-      - image: ubuntu:16.04
+      - image: docker:stable
     working_directory: /
     steps:
       - run:
           name: Install essential packages
           command: |
-            apt-get update
-            apt-get install -y ca-certificates curl build-essential make git
+            apk update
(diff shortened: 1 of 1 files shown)


sciurus:
Why do we want to subject ourselves to the potential pain of an alternative libc? 

According to our [dockerflow guidelines](https://github.com/mozilla-services/Dockerflow/blob/master/docs/building-container.md) we should use the [official python image](https://hub.docker.com/_/python/) unless we have a good reason not ot.

milescrabill:
We build and test everything in Docker anyway, so it shouldn't make a difference. Debian Dockerfiles don't come with Docker preinstalled. The point of using `docker:stable` is to have a base image that comes with Docker.

sciurus:
Miles eliminated my confusion in person. This change affects the environment our CI runs in when it builds the socorro docker image, it does not affect the actual base image used for the outputs of our CI process. That remains [python](https://github.com/mozilla-services/socorro/blob/master/docker/Dockerfile#L1).

PR #4478, opened 2018-06-13
circle config: delete deploy-dockerhub.sh, replicate functionality inline in circle config

milescrabill:
this should address the issues we saw in https://circleci.com/gh/mozilla-services/socorro/2856 as well as making the circle config a little closer to my pseudo-standard (ex https://github.com/mozilla-services/antenna/blob/master/.circleci/config.yml).


willkg:
I'm going to land this now because -stage deploys aren't working and this fixes it. We can iterate further in further PRs.

willkg:
For some reason, I did a "rebase and merge". I think that was an error--we shouldn't do that again. Sorry! :(

PR #4479, opened 2018-06-14
bug 1468763: remove "docker" prefixes from make rules

willkg:
Dropping the "docker" in the "dockerbuild" rule.


AlexisDeschamps:
Wouldn't you want to change the other commands starting with `docker` as well to match?

willkg:
Yes! You're right on--there's a bunch of other things to do. I'll do those now.

willkg:
@AlexisDeschamps Can you review this?

AlexisDeschamps:
Seems good. Tried locally and seems to work great.
Question: why not change the `Python 3 transition related things` commands as well?
Suggestion: change the PR title to be more general.

willkg:
I was thinking the Python3 stuff is temporary and it's easier to just leave it as is and then it goes away in a few months.

Changing the PR title is a good idea. I'll do that now.

PR #4480, opened 2018-06-15
bug 1463756: minor cleanup to signature tests

willkg:
This nixes the use of TestCaseBase which makes the siglists tests more
portable.

This also fixes the signature tests which were instantiating SignatureTool
classes. Those classes don't take a config anymore, so these tests shouldn't
be passing in config anymore.


willkg:
This is some minor cleanup of issues I noticed when extracting the signature generation code into a separate project. Figured I'd fix these here, too.

willkg:
Thank you!

PR #4482, opened 2018-06-15
fix bug 1467280: started returning 404 on RawCrash API requests with malformed crash IDs

AlexisDeschamps:
fixes bug https://bugzilla.mozilla.org/show_bug.cgi?id=1467280

## why 
- RawCrash API does not handle malformed crash IDs well. It does extra work and returns `500`

## what
- started returning 404 on RawCrash API requests with malformed crash IDs

### DIFF
--- a/socorro/external/boto/crash_data.py
+++ b/socorro/external/boto/crash_data.py
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from socorro.lib import external_common, MissingArgumentError, BadArgumentError
+from socorro.lib import external_common, MissingArgumentError, BadArgumentError, ooid
 from socorro.external.boto.crashstorage import (
     BotoS3CrashStorage,
     CrashIDNotFound,
@@ -58,6 +58,9 @@ class SimplifiedCrashData(BotoS3CrashStorage):
 
         if not params.uuid:
             raise MissingArgumentError('uuid')
+
+        if not ooid.is_crash_id_valid(params.uuid):
+            raise BadArgumentError('uuid')
 
         if not params.datatype:
             raise MissingArgumentError('datatype')
--- a/webapp-django/crashstats/crashstats/tests/test_models.py
+++ b/webapp-django/crashstats/crashstats/tests/test_models.py
@@ -14,6 +14,7 @@ from django.utils import timezone
 
 from crashstats.base.tests.testbase import DjangoTestCase
 from crashstats.crashstats import models
+from socorro.lib import BadArgumentError
 
 
 class Response(object):
@@ -334,6 +335,16 @@ class TestModels(DjangoTestCase):
         r = api.get(crash_id='some-crash-id')
         assert r['Vendor'] == 'Mozilla'
         assert 'Email' in r  # no filtering at this level
+
+    def test_raw_crash_invalid_id(self):
+        # NOTE(alexisdeschamps): this undoes the mocking of the implementation so we can test
+        # the implementation code.
+        models.RawCrash.implementation = self._mockeries[models.RawCrash]
+        model = models.RawCrash
+        api = model()
+
+        with pytest.raises(BadArgumentError):
+            api.get(crash_id='821fcd0c-d925-4900-85b6-687250180607docker/as_me.sh')
 
     def test_raw_crash_raw_data(self):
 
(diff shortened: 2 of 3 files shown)


webapp-django/crashstats/crashstats/models.py
```diff
@@ -248,6 +249,10 @@ def fetch(
         retries=None,
         retry_sleeptime=None
     ):
+        valid_crash_id = find_crash_id(params['uuid'])
+        if not valid_crash_id:
+            return HttpResponseBadRequest('Invalid crash ID')
```
willkg: This isn't going to work. This method is general across *all* the API endpoints many of which don't take a uuid parameter. We only want to validate the uuid parameter for the RawCrash API endpoint here:

https://github.com/mozilla-services/socorro/pull/4482/files#diff-0aa95732d6b7aebcee83eabf22da5dc7L696

Maybe adding validation to the `.get()` method is good? Maybe instead of putting it in the API code, put it in the model that backs it which is here:

https://github.com/mozilla-services/socorro/blob/19c0e1774ce9477cfc1c87f50926e95a340cd7ff/socorro/external/boto/crash_data.py#L50

webapp-django/crashstats/crashstats/utils.py
```diff
@@ -14,7 +14,7 @@
 from django import http
 from django.conf import settings
 
-from . import models
+import models
```
willkg: This isn't valid Python 3. It needs to either be:

```
from crashstats.crashstats import models
```

or
 
```
from . import models
````

Out of curiosity, what prompted you to change it?

File: socorro/external/boto/crash_data.py [hunk 3]
```diff
@@ -59,6 +59,9 @@ def get(self, **kwargs):
         if not params.uuid:
             raise MissingArgumentError('uuid')
 
+        if not ooid.is_crash_id_valid(params.uuid):
+            raise BadArgumentError('uuid')
+
```
willkg: This looks good. It's definitely worth adding a test.

Looks like the tests for the `RawCrash` api are here:

https://github.com/mozilla-services/socorro/blob/b486727bf42f52327f3140ba4bcd35288df3868a/webapp-django/crashstats/crashstats/tests/test_models.py#L319

That mocks out `.get()` on the `SimplifiedCrashData`. Since you're testing what happens if there are bad crash ids, I think you can skip the mock and verify it raises an exception.

See if that works. If not, something similar should work.

webapp-django/crashstats/crashstats/tests/test_models.py
```diff
@@ -335,6 +336,14 @@ def mocked_get(**params):
         assert r['Vendor'] == 'Mozilla'
         assert 'Email' in r  # no filtering at this level
 
+    def test_raw_crash_invalid_id(self):
+        model = models.RawCrash
+        model.implementation = self._mockeries[models.RawCrash]
```
willkg: Since this line is weird and anyone reading through this code is likely to stop abruptly with a "wait, what?", it's worth adding a comment. Maybe something along the lines of:

```
# NOTE(alexis): this undoes the mocking of the implementation so we can test
# the implementation code.
```

Also, `implementation` is a class property, but what you're doing here is setting an instance property which shadows the class property. I think that's fine in this case, but it's worth knowing that the class property is still a `MagicMock` and if there's code that refers to the property via the class, it won't work. If you create another instance and expect it to have the same value as this one, it won't.

```
$ python
[GCC 7.3.0] on linux2
>>> class Foo(object):
...     someprop = True
... 
>>> foo = Foo()
>>> foo.someprop = False
>>> bar = Foo()
>>> foo.someprop
False
>>> bar.someprop
True
>>> Foo.someprop
True
>>> 
```

Another way to do this that doesn't involve instance properties could be:

```
# NOTE(alexis): this undoes the mocking of the implementation so we can test
# the implementation code.
models.RawCrash.implementation = self._mockeries[models.RawCrash]
model = models.RawCrash
api = model()
```

webapp-django/crashstats/crashstats/tests/test_models.py
```diff
@@ -335,6 +336,14 @@ def mocked_get(**params):
...
+        api = model()
+
+        with pytest.raises(BadArgumentError):
+            api.get(crash_id='821fcd0c-d925-4900-85b6-687250180607docker/as_me.sh')
```
willkg: This test is perfect! We might also want to verify that a request returns an HTTP 400 or something that's not an HTTP 500.

Alternatively (and I should have noticed this the first time I mentioned tests--sorry about that), we could add another test similar to this one that covers the "crash id is there, but not valid" case, but undoes the mocking:

https://github.com/mozilla-services/socorro/blob/b486727bf42f52327f3140ba4bcd35288df3868a/webapp-django/crashstats/api/tests/test_views.py#L376-L383

AlexisDeschamps:
@willkg Thanks for the help. Do my updates fit your suggestions?

willkg:
@AlexisDeschamps Can you rebase this on the current master tip? Then we can merge it in.

PR #4484, opened 2018-06-19
fix bug 1462749: convert static pages to django admin

willkg:
This moves everything else to the Django admin and gets rid of our home grown admin.

The one thing this doesn't do is get rid of CSS defined in the page. The old admin pages did this. The Django admin pages do this, too. We should get rid of that and it's not hard. I figure I'll write up a bug and defer that work.

The other thing I didn't do was be super duper careful about css classes. I think instead of doing what I did and try to reuse the Django admin classes, we should use the base template, then define our own classes and use those. That'll make it easier to update to new Django versions and it's easier to find what's defined where since it's all in our code. I was thinking we could fix that when we fix the CSS defined in the page.

Anyhow, to test this:

1. run the tests
2. log into the site
3. go to the admin and run through the admin pages


socorro/external/es/create_recent_indices_app.py
```diff
@@ -43,7 +41,7 @@ def main(self):
         today = date.today()
         current_monday = today - timedelta(days=today.weekday())
 
-        for week_diff in range(-1, self.config.elasticsearch_weeks_to_create):
+        for week_diff in range(-2, self.config.elasticsearch_weeks_to_create):
```
willkg: The "super search missing fields" page looks at the previous two weeks. That fails with an HTTP 500. I figure it's easier to tweak the local development environment so it exists.

webapp-django/crashstats/crashstats/decorators.py
```diff
@@ -95,9 +95,6 @@ def inner(request, *args, **kwargs):
             # regular users will see a 404.
             if 'version' in str(e):
                 return redirect(reverse('home:home', args=(product,)))
-            elif request.user.is_superuser:
-                url = '%s?product=%s' % (reverse('manage:products'), product)
-                return redirect(url)
```
willkg: Until then, if the product doesn't exist, this raises an HTTP 500 because the view doesn't exist.

We nixed the products management pages. When we move the products table from alembic/sqlalchemy land to Django land, we could re-add this. I think we shouldn't bother since the new 404 page tells the dev what to do.

webapp-django/crashstats/manage/admin.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+@superuser_required
+def crash_me_now(request):
+    # NOTE(willkg): This intentionally throws an error so that we can test
+    # unhandled error handling.
+    1 / 0  # noqa
+
+
+@superuser_required
```
willkg: Most of the rest of this admin view code is just copied over from the old admin views.

webapp-django/crashstats/manage/admin_site.py
```diff
@@ -0,0 +1,17 @@
...
+
+site = SocorroAdminSite()
+
+# Stomp on the myriad of places Django stashes their AdminSite instance so
+# registering works with ours
+admin.site = site
+sites.site = site
+
+# Autodiscover all the admin modules and pull in models and such
+admin.autodiscover_modules('admin', register_to=site)
```
willkg: This is the shenanigans we talked about to get our admin site instance working with other things registering `ModelAdmin` classes.

willkg:
As a side note, I claim these changes are "low risk". If these pages are buggy, it's not a big deal and we can fix them as we hit issues. Some of them we don't really use at all. For example, I've never used the graphics devices page.

willkg:
@AlexisDeschamps I'm switching this from @Osmose reviewing to you. He can look at parts he's concerned with.

If you need help sussing out how bits work, just ask questions in the PR comments.

willkg:
A lot of the template code is copy/paste with minor edits because I converted it from Jinja2 templates to Django templates. The rest is new-ish and/or the changes are more involved.

willkg:
Thank you!

PR #4485, opened 2018-06-19
fix bug 1468763: fix lint and test rules

willkg:
This fixes linting and testing so there's a single make rule for linting and
one for testing. This fixes Circle configuration so it runs both.

To test:

1. run `make lint`
2. run `make test`
3. make sure Circle ran both lint and test


willkg:
Thank you!

PR #4487, opened 2018-06-19
fix bug 1467308 update socorro.lib

ceddy-cedd:
Modified socorro.lib folder to pass python 3 testing. 
Errors fixed: 
* An error occurred when reporting class names and descriptions of expected results in test functions. This changed from python 2 to 3. Python 2 returned the class name and the value of the error. Python 3 returned a qualified class name(absolute reference) and the value of the error. So we modified test expectation to just return its value and removed a redundant verification of the error type. 
* We then resolved implementation difference of range() function between python 2 and 3 by converting the iterator generated in python 3 to a list for value comparison. 
* Fixed import changes in python 2 to 3 using the six lib. 
* Lastly we fixed a function that sanitized strings to return as the default type of string in the testing environment of either python 2 or 3.

### DIFF
--- a/socorro/lib/util.py
+++ b/socorro/lib/util.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 from __future__ import print_function
+import sys
 
 from itertools import islice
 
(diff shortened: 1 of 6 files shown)


socorro/lib/util.py
```diff
@@ -40,12 +41,15 @@ def drop_unicode(text):
     :returns: text with all unicode characters dropped
 
     """
-    if isinstance(text, str):
-        # Convert any str to a unicode so that we can convert it back and drop any non-ascii
-        # characters
-        text = text.decode('unicode_escape')
-
-    return text.encode('ascii', 'ignore')
+    if sys.version_info < (3,):
```
willkg: `six` has a `PY2` and `PY3` which are true and false. I'd use that instead since it's more readable in intent (in my opinion) and it's easy to grep for later. So then this line would be:

```
if six.PY2:
```

ceddy-cedd:
@willkg I will change the line you suggested but after a conceptual conversation with lonnen so I can really understand the deeper nuance of the "why."

willkg:
It's like this:

```
if sys.version < (3,):
```

versus:

```
if six.PY2:
```

The first one reads like this "if the system version is less than a tuple starting with a 3" and the second one reads like this "if python 2". I think the second one is closer to the intention here--we want to check if this code is being run with Python 2 and do something if that's the case.

lonnen:
👍

PR #4489, opened 2018-06-20
fix bug 1469978: nix all the admin links

willkg:
All the pages on crash stats have a link to the admin in the footer. There's
no need to have admin links sprinkled around the site. They're not chocolate
chips.

This removes some from the profile page which were erroring out after I
nixed the "manage:home" view.


willkg:
Thank you!

I'll check this after it deploys.

PR #4490, opened 2018-06-21
fix bug 1470042: add 62.0b1 to AuroraVersionFixitRule

willkg:
62.0b1 was released in devedition yesterday it looks like. This adds an item to the rule to correct the version number in the crash reports.


willkg:
Thank you for merging this! That expedites things greatly!

PR #4491, opened 2018-06-21
Fix deploy_bug.py to show authors

willkg:
The `deploy_bug.py` script spits out all the commits that will get deployed.
This adjusts that output to also specify the authors. The hope is that
this makes it easier to figure out who needs to verify what before a
prod deploy.

Example of old output:

```
323c77c: bug 1462749: move graphics devices to django admin
f24fa3c: fix bug 1462749: remove last vestiges of old admin
d820f11: Fix create_recent_indices_app to create last 2 weeks
```

Example of new output:

```
323c77c: bug 1462749: move graphics devices to django admin (willkg)
f24fa3c: fix bug 1462749: remove last vestiges of old admin (willkg)
d820f11: Fix create_recent_indices_app to create last 2 weeks (willkg)
```

It's showing the github username of the person. For us, I think that will generally work. We could show something else if we wanted.

@Osmose What do you think?

### DIFF
--- a/scripts/deploy_bug.py
+++ b/scripts/deploy_bug.py
@@ -81,11 +81,11 @@ def main(argv):
     next_tag = current_tag + 1
 
     # Figure out the current version
-    sha = fetch_current_revision(HOST)
+    deployed_sha = fetch_current_revision(HOST)
 
     # Get the commits between the currently deployed version and what's in
     # master tip and if there's nothing to deploy, then we're done!
-    resp = fetch_history_from_github('mozilla-services', 'socorro', sha)
+    resp = fetch_history_from_github('mozilla-services', 'socorro', deployed_sha)
     if resp['status'] != 'ahead':
         print('Nothing to deploy!')
         return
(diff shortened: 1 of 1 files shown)


scripts/deploy_bug.py
```diff
@@ -108,9 +108,10 @@ def main(argv):
         if len(commit['parents']) > 1:
             continue
 
-        print('%s: %s' % (
+        print('%s: %s (%s)' % (
             commit['sha'][:7],
-            commit['commit']['message'].splitlines()[0][:80]
+            commit['commit']['message'].splitlines()[0][:80],
+            (commit['author'] or {}).get('login', '?')
```
Osmose: Our of pure curiosity, why this instead of `commit.get('author', {})`?

willkg: I copied this from my `status.py` script which I wrote a while ago. I think the problem is that we need to account for when the value for `commit['author']` is `None` as well as a dict that might not have a `login` key.

So the first part:

```
commit['author'] or {}
```

handles the `None` case because that's falsey and makes sure the result is a dict that has a `.get()` method.

I can make this better via a comment or something if you think that's worthwhile.

willkg: Scratch that--I'll make it better.

willkg:
Thank you! Hopefully this makes deploys a little easier to manage.

PR #4493, opened 2018-06-22
bug 1454482: add timing captures for ES

AlexisDeschamps:
bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1454482
This patch is a work in progress.
This patch is based off #4391.

## what
- added metric for the time to index a crash
  - successes and failures have a different metric

## why
- provide better timing data for document indexing in elasticsearch

## notes
@sciurus I can make other changes within this patch. Does this patch still make sense? What other things would it be helpful to have timings on?

### DIFF
--- a/socorro/external/boto/connection_context.py
+++ b/socorro/external/boto/connection_context.py
@@ -6,6 +6,7 @@ import json
 import socket
 import datetime
 import contextlib
+import time
 from past.builtins import basestring
 
 import boto
@@ -14,6 +15,7 @@ import boto.exception
 
 from configman import Namespace, RequiredConfig, class_converter
 from configman.converters import str_to_boolean
+import markus
 
 from socorro.lib.converters import change_default
 from socorro.lib.ooid import dateFromOoid
@@ -87,7 +89,7 @@ class ConnectionContextBase(RequiredConfig):
             return True
         return False
 
-    def __init__(self, config, quit_check_callback=None):
+    def __init__(self, config, namespace='', quit_check_callback=None):
         self.config = config
 
         self._CreateError = boto.exception.StorageCreateError
@@ -97,6 +99,7 @@ class ConnectionContextBase(RequiredConfig):
         )
 
         self._bucket_cache = {}
+        self.metrics = markus.get_metrics(namespace)
 
     def _connect(self):
         try:
@@ -192,15 +195,28 @@ class ConnectionContextBase(RequiredConfig):
         """
         # can only submit strings to boto
         assert isinstance(thing, basestring), type(thing)
+        try:
+            start_time = time.time()
 
-        conn = self._connect()
-        bucket = self._get_or_create_bucket(conn, self.config.bucket_name)
+            conn = self._connect()
+            bucket = self._get_or_create_bucket(conn, self.config.bucket_name)
 
-        all_keys = self.build_keys(self.config.prefix, name_of_thing, id)
-        # Always submit using the first key
-        key = all_keys[0]
-        key_object = bucket.new_key(key)
-        key_object.set_contents_from_string(thing)
+            all_keys = self.build_keys(self.config.prefix, name_of_thing, id)
+            # Always submit using the first key
+            key = all_keys[0]
+            key_object = bucket.new_key(key)
+            key_object.set_contents_from_string(thing)
+            index_outcome = 'successful'
+        except Exception:
+            index_outcome = 'failed'
+            raise
+        finally:
+            elapsed_time = time.time() - start_time
+            self.metrics.histogram(
+                'submit',
+                value=elapsed_time * 1000.0,
+                tags=['kind:' + name_of_thing, 'outcome:' + index_outcome]
+            )
 
     def fetch(self, id, name_of_thing):
         """retrieve something from boto.
@@ -269,8 +285,8 @@ class S3ConnectionContext(ConnectionContextBase):
         likely_to_be_changed=True,
     )
 
-    def __init__(self, config, quit_check_callback=None):
-        super(S3ConnectionContext, self).__init__(config)
+    def __init__(self, config, namespace='', quit_check_callback=None):
+        super(S3ConnectionContext, self).__init__(config, namespace=namespace)
 
         self._connect_to_endpoint = boto.connect_s3
         self._calling_format = config.calling_format
@@ -302,8 +318,8 @@ class RegionalS3ConnectionContext(S3ConnectionContext):
         'boto.s3.connection.OrdinaryCallingFormat'
     )
 
-    def __init__(self, config, quit_check_callback=None):
-        super(RegionalS3ConnectionContext, self).__init__(config)
+    def __init__(self, config, namespace='', quit_check_callback=None):
+        super(RegionalS3ConnectionContext, self).__init__(config, namespace=namespace)
         self._region = config.region
         self._connect_to_endpoint = boto.s3.connect_to_region
 
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -4,6 +4,7 @@
 
 import json
 import re
+import time
 from past.builtins import basestring
 from threading import Thread
 from Queue import Queue
@@ -319,6 +320,27 @@ class ESCrashStorage(CrashStorageBase):
             # later.
             self.config.logger.exception('something went wrong when capturing processed_crash_size')
 
+    def _index_crash(self, connection, es_index, es_doctype, crash_document, crash_id):
+        try:
+            start_time = time.time()
+            connection.index(
+                index=es_index,
+                doc_type=es_doctype,
+                body=crash_document,
+                id=crash_id
+            )
+            index_outcome = 'successful'
+        except Exception:
+            index_outcome = 'failed'
+            raise
+        finally:
+            elapsed_time = time.time() - start_time
+            self.metrics.histogram(
+                'index',
+                value=elapsed_time * 1000.0,
+                tags=['outcome:' + index_outcome]
+            )
+
     def _submit_crash_to_elasticsearch(self, connection, crash_document):
         """Submit a crash report to elasticsearch"""
         es_index = self.get_index_for_crash(
@@ -337,12 +359,7 @@ class ESCrashStorage(CrashStorageBase):
         # case of an unhandled exception.
         for attempt in range(5):
             try:
-                connection.index(
-                    index=es_index,
-                    doc_type=es_doctype,
-                    body=crash_document,
-                    id=crash_id
-                )
+                self._index_crash(connection, es_index, es_doctype, crash_document, crash_id)
                 break
             except elasticsearch.exceptions.TransportError as e:
                 field_name = None
--- a/socorro/unittest/external/es/test_crashstorage.py
+++ b/socorro/unittest/external/es/test_crashstorage.py
@@ -947,6 +947,36 @@ class TestESCrashStorage(ElasticsearchTestCase):
             assert mm.has_record('histogram', stat='processor.es.raw_crash_size', value=27)
             assert mm.has_record('histogram', stat='processor.es.processed_crash_size', value=1785)
 
+    def test_index_data_capture(self):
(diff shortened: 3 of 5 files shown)


socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -947,6 +947,34 @@ def test_crash_size_capture(self):
...
+                es_storage._index_crash_in_elasticsearch(
+                    connection=mock_connection,
+                    es_index=None,
+                    es_doctype=None,
+                    crash_document=None,
+                    crash_id=None
+                )
+
+            assert mm.has_record('histogram', stat='processor.es.successful_index')
+            assert mm.has_record('histogram', stat='processor.es.failed_index')
```
willkg: This looks cool and is very readable.

I've read the code and it's very unlikely, but if someone rewrote it, it's possible we could end up with code that sends both timings for a success (or a fail) which would be wrong. This test wouldn't catch that. One thing you could do is assert there is one and only one `processor.es.successful_index` timing and one and only one `processor.es.failed_index` timing. You can use `.filter()` for that. Alternatively, you could split this into two separate tests and for each one make sure it has the timing it should have and doesn't have the timing it doesn't have.

socorro/external/es/crashstorage.py
```diff
@@ -319,6 +320,27 @@ def capture_crash_metrics(self, raw_crash, processed_crash):
...
+            )
+            index_outcome = 'successful'
+        except Exception:
+            index_outcome = 'failed'
+            raise
+        finally:
+            elapsed_time = datetime.now() - start_time
+            self.metrics.histogram(
+                '%s_index' % index_outcome,
+                value=elapsed_time.microseconds
```
willkg: There are two problems here. First is that the `.microseconds` property is not the total number of microseconds in the timedelta, but just the microseconds part of the timespan. So if the timedelta was 1.01, then this would just be .01 which isn't what we want.

```
$ python
[GCC 7.3.0] on linux2
>>> from datetime import datetime
>>> start = datetime.now()
>>> delta = datetime.now() - start
>>> delta
datetime.timedelta(0, 6, 479467)
>>> delta.microseconds
479467
```

https://docs.python.org/2/library/datetime.html#datetime.timedelta

Second is that we want timings in milliseconds--not microseconds. That keeps our timings in datadog consistent with the same metric.

When I implemented this in Markus, I did it with `time.time()` which gives you more of what you want without having to do any math or conversions.

This is what Markus does:

https://github.com/willkg/markus/blob/master/markus/main.py#L339-L342

We should change it to something like that.

socorro/external/es/crashstorage.py
```diff
@@ -319,6 +320,27 @@ def capture_crash_metrics(self, raw_crash, processed_crash):
             # later.
             self.config.logger.exception('something went wrong when capturing processed_crash_size')
 
+    def _index_crash_in_elasticsearch(self, connection, es_index, es_doctype, crash_document,
```
willkg: Given that this crashstorage is only about Elasticsearch, I'd probably shorten this to `_index_crash`. You don't have to change it if you don't want to--either is fine.

socorro/external/es/crashstorage.py
willkg: Oh, I see what happened--you copied the code from the previous attempt. That's totally fine. However, what's funny about that is that @sciurus and I puzzled over the stats a couple of months ago when I rewrote the metrics specifically about the milliseconds vs. microseconds and how the crash storage timings prior to the rewrite were wrong (for the last 6+ years). I should have tossed a comment in his PR back then, but missed it. Sorry about that!

willkg:
Thank you for doing this! Getting better ES metrics will be really helpful.

AlexisDeschamps:
@willkg Thanks for the great review. I added some updates.

sciurus:
This looks nice to me!

Regarding other things in the processor that need metrics, another good candidate is the actual uploads to S3. I think those are the five `boto_connection.submit` calls in https://github.com/mozilla-services/socorro/blob/master/socorro/external/boto/crashstorage.py which eventually call https://github.com/mozilla-services/socorro/blob/master/socorro/external/boto/connection_context.py#L203

AlexisDeschamps:
@willkg I added https://github.com/mozilla-services/socorro/pull/4493/commits/abd2d190f6293408f298eda842f3419dff89e69e if you want to take a look. I'm not sure about the namespace part.

socorro/external/boto/connection_context.py
```diff
@@ -202,6 +206,9 @@ def submit(self, id, name_of_thing, thing):
         key_object = bucket.new_key(key)
         key_object.set_contents_from_string(thing)
 
+        elapsed_time = time.time() - start_time
+        self.metrics.histogram(name_of_thing, value=elapsed_time * 1000.0)
```
willkg: I'd add "submit" to this, too. It'll make it easier to distinguish other activities from one another if we add timings for other things later.

So, something like `'submit.' + name_of_thing` if we want the hierarchy to be processor -> boto -> submit -> raw_crash. Or `name_of_thing + '.submit'` if we want the hierarchy to be processor -> boto -> raw_crash -> submit. I'm on the fence about which is better for our purposes. Maybe the latter.

@sciurus Do you have a preference for the hierarchy? Is one more convenient with your Datadog graphing plans?

sciurus: I'd prefer that `name_of_thing` not be in the hierarchy at all, and instead be recorded via a tag. You run into gray areas pretty quickly, but generally I think a metric's name should reflect the service (socorro processor) and the operation (submitting to s3) and tags should be used for any additional dimensions (e.g. what you're submitting).

The key thing with Datadog is that tags, not the hierarchy encoded in the name,  determine how you can aggregate. If we put the type of thing being submitted in the metric's name, it's harder to analyze the timing of all submissions. If we put it in a tag, it's easy to look at all submissions or to break it down to a specific type.

willkg: The "name_of_thing" is a finite and small set of things, so that makes sense to me and would work here nicely.

@AlexisDeschamps Can you change this to use a tag? It'd be something like this:

```
self.metrics.histogram('submit', value=elapsed_time * 1000.0, tags=['type:' + name_of_thing])
```

I used "type" there, but "kind", "genus", or any word like that is fine.

sciurus: `type` in particular isn't a good tag key, because the metric will already have a `Type` tag that is picked up from AWS.

https://mana.mozilla.org/wiki/display/SVCOPS/AWS+EC2+Instance+Tags

sciurus: As currently written, if any of the functions in between start_time and elapsed_time throw an exception, we won't record the histogram.

It would be good to send it in both cases, like we do for elasticsearch.

socorro/external/es/crashstorage.py
```diff
@@ -319,6 +320,26 @@ def capture_crash_metrics(self, raw_crash, processed_crash):
...
+                id=crash_id
+            )
+            index_outcome = 'successful'
+        except Exception:
+            index_outcome = 'failed'
+            raise
+        finally:
+            elapsed_time = time.time() - start_time
+            self.metrics.histogram(
+                '%s_index' % index_outcome,
```
sciurus: I'm second guessing myself from a few months ago now that willkg got me thinking about metric names. Maybe this should just be named `index` and have a tag that indicates success or failure?

Whichever we do here, we should do the same for s3.

willkg: Yeah... +1 for putting the outcome in a tag.

socorro/external/boto/connection_context.py
```diff
@@ -87,7 +89,7 @@ def is_operational_exception(self, x):
             return True
         return False
 
-    def __init__(self, config, quit_check_callback=None):
+    def __init__(self, config, namespace='', quit_check_callback=None):
```
willkg: Oops--I didn't catch that you changed the `__init__` for the base class. You're going to need to update all the subclasses and all the points that one of those classes gets instantiated, too.

AlexisDeschamps: Does https://github.com/mozilla-services/socorro/pull/4493/commits/7d8752ef7a95f3d3faaaad61912d3f3e81413a90 make sense?

socorro/external/boto/connection_context.py
```diff
@@ -269,8 +285,8 @@ class S3ConnectionContext(ConnectionContextBase):
         likely_to_be_changed=True,
     )
 
-    def __init__(self, config, quit_check_callback=None):
-        super(S3ConnectionContext, self).__init__(config)
+    def __init__(self, config, namespace='', quit_check_callback=None):
+        super(S3ConnectionContext, self).__init__(config, namespace)
```
willkg: I'd change this and the other instances like this to:

```
super(S3ConnectionContext, self).__init__(config, namespace=namespace)
```

which makes it a lot easier for us to adjust the signatures later on without arguments ending up in the wrong place.

For example, say we add another argument and for some reason in the API it makes sense to put it after the `config` arg:

```
def __init__(self, config, prefix, namespace='', quit_check_callback=None)
```

Now we'll have to go through and fix all the other situations. If instead we were using keyword-style argument passing, we wouldn't have to do anything further.

Plus it makes the code easier to read and debug. If we're doing keyword-style arguments, then we know the value ends up in the right place. If we're doing index-style arguments, then we have to remember signatures and order and all that.

I don't think we need to pass `config` as a keyword-style argument because we have an unwritten convention in Socorro that `config` is always the first argument in configman-style classes and it's always passed in.

socorro/external/boto/connection_context.py
willkg: This looks good, but it's missing something: we need to pass in the namespace when we create the connection context class:

https://github.com/mozilla-services/socorro/blob/a1100df887a077596539f1fc08c63754705ecd20/socorro/external/boto/crashstorage.py#L91

Because we have metrics set up to log metrics, you can see the keys and values getting spit out when running the processor. Right now it shows this:

```
processor_1      | 2018-06-27 13:33:26,276 INFO - markus - Thread-2 - METRICS|2018-06-27 13:33:26|histogram|submit|70.259809494|#kind:processed_crash,outcome:successful
processor_1      | 2018-06-27 13:33:26,276 INFO - markus - Thread-2 - METRICS|2018-06-27 13:33:26|timing|processor.s3.BotoS3CrashStorage.save_raw_and_processed|77.6069164276|
processor_1      | 2018-06-27 13:33:26,323 INFO - markus - Thread-2 - METRICS|2018-06-27 13:33:26|histogram|processor.elasticsearch.raw_crash_size|14029|
processor_1      | 2018-06-27 13:33:26,324 INFO - markus - Thread-2 - METRICS|2018-06-27 13:33:26|histogram|processor.elasticsearch.processed_crash_size|132336|
processor_1      | 2018-06-27 13:33:26,756 INFO - processor - Thread-2 - Created new elasticsearch index: socorro201826
processor_1      | 2018-06-27 13:33:27,177 INFO - markus - Thread-2 - METRICS|2018-06-27 13:33:27|histogram|processor.elasticsearch.index|420.194149017|#outcome:successful
processor_1      | 2018-06-27 13:33:27,177 INFO - markus - Thread-2 - METRICS|2018-06-27 13:33:27|timing|processor.es.ESCrashStorageRedactedJsonDump.save_raw_and_processed|854.610919952|
processor_1      | 2018-06-27 13:33:27,177 INFO - markus - Thread-2 - METRICS|2018-06-27 13:33:27|incr|processor.save_raw_and_processed|1|
processor_1      | 2018-06-27 13:33:27,280 INFO - markus - Thread-2 - METRICS|2018-06-27 13:33:27|histogram|submit|101.981878281|#kind:crash_report,outcome:successful
processor_1      | 2018-06-27 13:33:27,281 INFO - markus - Thread-2 - METRICS|2018-06-27 13:33:27|timing|processor.telemetry.TelemetryBotoS3CrashStorage.save_raw_and_processed|103.121042252|
```

The `submit` key is missing the namespace.

willkg:
@sciurus What do you think about the above keys and metrics? Does this work for you?

PR #4494, opened 2018-06-24
Bug 1468810: Add scaffolding for development of React-based report index.

Osmose:
This is the first step towards re-implementing the crash report page using a more modern frontend stack.

This adds a bunch of stuff:

- A view at `/report/index/<crash_id>/new` as the placeholder for the new crash report page.
- A webpack config that builds files in `/webapp-django/staticfiles` and stores the bundles in `/webapp-django/webpack_bundles`.
- [`django-webpack-loader`](https://github.com/owais/django-webpack-loader), which reads the `/webapp-django/webpack-stats.json` file output by webpack to determine which bundles were built. It then makes these bundles available for use within templates via the `render_bundle` function.
- A `webpack` container that runs concurrently with the `processor` and `webapp` in `make run` that re-builds the frontend files when they change. __Significantly, this shares the built JS with the webapp container via the local volume mount, which may not work on Linux__ (I only tested using Docker for Mac).
- React
- A basic JS file that users React to render content to the crash report page.
- A bunch of eslint tweaks to support correct linting of the new JS files.

The diff is huge because the `package-lock.json` changes are huge.

To test this:

- Checkout the changes.
- Run `make build`.
- Run `make run`.
- Visit http://localhost:8000/report/index/044b488b-b0ed-44de-bb9e-d692b0180418/new and observe an empty crash-stats page with a header "New crash report page placeholder".
- Open up `webapp-django/staticfiles/index.js` and modify the text in the `h1` element.
- Refresh the crash report page and check that the header contents were updated

Besides actually doing the reimplementation, there's a few other things left to do for scaffolding:

- Add frontend tests for the webpack bundles
- Figure out why the webpack container doesn't output anything when it rebuilds the JS or hits a build error.
- Inspect how much build time has increased due to the added packages. I suspect `make build` takes longer now due to the extra packages this added, but I haven't directly compared it.

I think this could be mergable (except maybe for the increased build time issue) as it is, since it should be able to be run parallel to normal development.


.eslintrc
```diff
@@ -4,9 +4,12 @@ rules:
     - unix
   no-console:
     - 1
+  max-len: "off" # Use prettier default line length
```
Osmose: See https://github.com/prettier/prettier-eslint/issues/55.

webapp-django/webpack.config.js
```diff
@@ -0,0 +1,24 @@
+/* eslint-env node */
+const path = require('path');
+const BundleTracker = require('webpack-bundle-tracker');
+
+module.exports = {
```
Osmose: This file is optimized for development. If we move forward with the reimplmented page, we should revisit this file and make sure the bundles it's building are small and fast.

.eslintrc
AlexisDeschamps: Isn't that fixed though?

Osmose: You'd think, right?

Without this I get lint errors from prettier that, when fixed, result in lines longer than 120 characters. I'm not entirely sure why. It certainly doesn't feel like the worry-free future that prettier promised.

webapp-django/staticfiles/.eslintrc
```diff
@@ -0,0 +1,15 @@
...
+env:
+  browser: true
+  es6: true
+parserOptions:
+  ecmaVersion: 7
+  sourceType: module
+extends:
+  - eslint:recommended
+  - plugin:prettier/recommended
+  - plugin:react/recommended
```
willkg: I don't know offhand what this file means, but we do have a very short list of JavaScript code conventions documented here:

https://socorro.readthedocs.io/en/latest/contributing.html#javascript-code-conventions

Does this file encode those conventions or should we update our conventions?

webapp-django/webpack.config.js
willkg: For things like this since we collectively forget unless there's a string tied around someone's finger, we should either create a Bugzilla bug after this lands or add a FIXME in the comments here or both.

webapp-django/staticfiles/.eslintrc
AlexisDeschamps: From what I remember, we opted for 2-space indentation for https://github.com/mozilla-services/socorro/blob/master/.editorconfig to reduce the amount of changes.

willkg: That file has us at a 4-space indent.

It's entirely possible the conventions didn't match the code. I don't do much in JS-land, so I don't know what the deal is.

Seems like we have mis-matched things. We should update all the things to be either 2-space or 4-space. We could throw that in a bug to do later.

AlexisDeschamps: I think prettier defaults to tab space of 2.

Osmose: 2 spaces is preferred, especially for React code, since with JSX you end up with some pretty deep nesting and the 2-space indents really help.

Given that, I think keeping this separate from the rest of the site's styles is appropriate, at least until we decide if the cupcake is a good idea and should be exported to the rest of the site.

Osmose:
Rebased and added the suggested FIXME commit. Gonna work on the build failure in CI next.

willkg:
I did `make clean` and then `make build` and got this towards the end:

```
Step 16/19 : RUN cd /app/webapp-django/ && $WEBPACK_BINARY --mode=production
 ---> Running in 38b59f6abb3f
✖ ｢config｣: An error occurred while trying to find a config file
              Did you forget to specify a --require?
/webapp-frontend-deps/node_modules/@webpack-contrib/config-loader/lib/load.js:84
    throw new LoadConfigError(e, configPath);
    ^

LoadConfigError: Cannot find module 'webpack-bundle-tracker'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:594:15)
    at Function.Module._load (internal/modules/cjs/loader.js:520:25)
    at Module.require (internal/modules/cjs/loader.js:650:17)
    at require (/webapp-frontend-deps/node_modules/v8-compile-cache/v8-compile-cache.js:159:20)
    at Object.<anonymous> (/app/webapp-django/webpack.config.js:3:23)
    at Module._compile (/webapp-frontend-deps/node_modules/v8-compile-cache/v8-compile-cache.js:178:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:713:10)
    at Module.load (internal/modules/cjs/loader.js:612:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:551:12)
    at Function.Module._load (internal/modules/cjs/loader.js:543:3)
ERROR: Service 'webapp' failed to build: The command '/bin/sh -c cd /app/webapp-django/ && $WEBPACK_BINARY --mode=production' returned a non-zero code: 1
Makefile:55: recipe for target 'build' failed
make: *** [build] Error 1
```

I didn't look into that at all. Is that something you need me to look into?

willkg:
Scratch that... let me rebuild all the images with no cache and see if that fixes anything. I haven't done that in a while, so I'm due anyhow.

willkg:
I rebuilt with no cache and get the same thing. Oh, hot diggity--I'm getting the same thing CI is.

Osmose:
So the latest failure is:

```
+ python manage.py collectstatic --noinput
StatsdMetrics configured: statsd:8125 None
  File "manage.py", line 12, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 364, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 356, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/base.py", line 283, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/base.py", line 330, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python2.7/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 199, in handle
    collected = self.collect()
  File "/usr/local/lib/python2.7/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 115, in collect
    for path, storage in finder.list(self.ignore_patterns):
  File "/usr/local/lib/python2.7/site-packages/django/contrib/staticfiles/finders.py", line 112, in list
    for path in utils.get_files(storage, ignore_patterns):
  File "/usr/local/lib/python2.7/site-packages/django/contrib/staticfiles/utils.py", line 28, in get_files
    directories, files = storage.listdir(location)
  File "/usr/local/lib/python2.7/site-packages/django/core/files/storage.py", line 397, in listdir
    for entry in os.listdir(path):
OSError: [Errno 2] No such file or directory: '/app/webapp-django/webpack_bundles'
```

`/app/webapp-django/webpack_bundles` is created by Webpack when the `webapp` docker image is built. It is also created by the `webpack` container when running local development, but that's not in play here.

[This part of `run_tests_in_docker.sh`](https://github.com/mozilla-services/socorro/blob/master/docker/run_tests_in_docker.sh#L68) creates a data volume, clears it of any data, and copies the contents of `/socorro` in the CircleCI machine (the current working directory when `make test` is run, it's the root of a Socorro checkout) to it. I looked through the commit history but it's been there since the file was created, so I can't figure out why we do this instead of using the Socorro checkout that's already in the container. My best guess is that it's so that local development runs the tests against the modified code instead of against the code as it was when the dev last built the `webapp` container.

What I suspect is happening is that the volume being mounted is overwriting the `/app` that's already in the container, and the volume doesn't have `webpack_bundles` since that directory was only created inside the `webapp` container when it was built.

@willkg Can you confirm why we're creating this volume? I think this volume-copying code should not be used when running the tests in CircleCI. I may try to fix it so that it only runs when running the tests locally, maybe through a flag or another make command or something.

Osmose:
Hooray, the build passes! This should be ready for a proper review now.

willkg:
The whole "create a container with a volume in it" shenangans is explicitly because we can't mount a volume in a container when running in Circle. We should talk via vidyo or something and I can explain it with lots of hand-waving and figure out what's going on.

.dockerignore
```diff
@@ -5,3 +5,5 @@ __pycache__
 .vagrant
 Vagrantfile
 .cache
+webapp-django/node_modules
+webapp-django/static
```
willkg: Why add these to `.dockerignore`? It means they don't get copied into the image with a COPY.

Makefile
```diff
@@ -72,7 +72,7 @@ updatedata: my.env
 	./docker/run_update_data.sh
 
 run: my.env
-	${DC} up webapp processor
+	${DC} up webapp processor webpack
```
willkg: Is it the case that the webpack service produces output that's helpful to see while you're working on webapp stuff?

docker-compose.yml
```diff
@@ -71,6 +71,18 @@ services:
     ports:
       - "8080:8080"
 
+  # Webpack watcher
+  webpack:
```
willkg: I'd treat this like the `oidcprovider` and include it in the `depends_on` section of the `webapp` service. That way doing:

```
docker-compose up webapp
```

brings this up, too. I tend to run the webapp this way so I don't have to deal with the extremely chatty processor.

Does that sound prudent?

willkg:
This doesn't work on linux. When running the webpack service, it dies immediately after this:

```
webpack_1        | 
webpack_1        | ┌───────────────────────────────────────────────────────┐
webpack_1        | │          webpack-command update check failed          │
webpack_1        | │          Try running with sudo or get access          │
webpack_1        | │         to the local update config store via          │
webpack_1        | │ sudo chown -R $USER:$(id -gn $USER) /home/app/.config │
webpack_1        | └───────────────────────────────────────────────────────┘
webpack_1        | Error: EACCES: permission denied, open '/app/webapp-django/webpack-stats.json'
webpack_1        |     at Object.openSync (fs.js:443:3)
webpack_1        |     at Object.writeFileSync (fs.js:1170:35)
webpack_1        |     at Plugin.writeOutput (/webapp-frontend-deps/node_modules/webpack-bundle-tracker/index.js:97:6)
webpack_1        |     at /webapp-frontend-deps/node_modules/webpack-bundle-tracker/index.js:43:12
webpack_1        |     at SyncHook.eval [as call] (eval at create (/webapp-frontend-deps/node_modules/tapable/lib/HookCodeFactory.js:17:12), <anonymous>:7:1)
webpack_1        |     at SyncHook.lazyCompileHook [as _call] (/webapp-frontend-deps/node_modules/tapable/lib/Hook.js:35:21)
webpack_1        |     at hooks.beforeCompile.callAsync.err (/webapp-frontend-deps/node_modules/webpack/lib/Compiler.js:485:23)
webpack_1        |     at AsyncSeriesHook.eval [as callAsync] (eval at create (/webapp-frontend-deps/node_modules/tapable/lib/HookCodeFactory.js:24:12), <anonymous>:6:1)
webpack_1        |     at AsyncSeriesHook.lazyCompileHook [as _callAsync] (/webapp-frontend-deps/node_modules/tapable/lib/Hook.js:35:21)
webpack_1        |     at Compiler.compile (/webapp-frontend-deps/node_modules/webpack/lib/Compiler.js:482:28)
```

After that, the webapp kicks up a 500 when trying to look at the new report page:

```
webapp_1         | Internal Server Error: /report/index/044b488b-b0ed-44de-bb9e-d692b0180418/new
webapp_1         | Traceback (most recent call last):
webapp_1         |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/exception.py", line 41, in inner
webapp_1         |     response = get_response(request)
webapp_1         |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 187, in _get_response
webapp_1         |     response = self.process_exception_by_middleware(e, request)
webapp_1         |   File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 185, in _get_response
webapp_1         |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
webapp_1         |   File "/app/webapp-django/crashstats/crashstats/decorators.py", line 99, in inner
webapp_1         |     return view(request, *args, **kwargs)
webapp_1         |   File "/app/webapp-django/crashstats/crashstats/views.py", line 1238, in new_report_index
webapp_1         |     return render(request, 'crashstats/new_report_index.html', context)
webapp_1         |   File "/usr/local/lib/python2.7/site-packages/django/shortcuts.py", line 30, in render
webapp_1         |     content = loader.render_to_string(template_name, context, request, using=using)
webapp_1         |   File "/usr/local/lib/python2.7/site-packages/django/template/loader.py", line 68, in render_to_string
webapp_1         |     return template.render(context, request)
webapp_1         |   File "/usr/local/lib/python2.7/site-packages/django_jinja/backend.py", line 106, in render
webapp_1         |     return mark_safe(self.template.render(context))
webapp_1         |   File "/usr/local/lib/python2.7/site-packages/jinja2/environment.py", line 1008, in render
webapp_1         |     return self.environment.handle_exception(exc_info, True)
webapp_1         |   File "/usr/local/lib/python2.7/site-packages/jinja2/environment.py", line 780, in handle_exception
webapp_1         |     reraise(exc_type, exc_value, tb)
webapp_1         |   File "/app/webapp-django/crashstats/crashstats/jinja2/crashstats/new_report_index.html", line 1, in top-level template code
webapp_1         |     {% extends "crashstats_base.html" %}
webapp_1         |   File "/app/webapp-django/crashstats/base/jinja2/crashstats_base.html", line 8, in top-level template code
webapp_1         |     {% block site_css %}
webapp_1         |   File "/app/webapp-django/crashstats/base/jinja2/crashstats_base.html", line 9, in block "site_css"
webapp_1         |     {% stylesheet 'crashstats_base' %}
webapp_1         |   File "/usr/local/lib/python2.7/site-packages/pipeline/jinja2/__init__.py", line 37, in package_css
webapp_1         |     return self.render_compressed(package, package_name, 'css')
webapp_1         |   File "/usr/local/lib/python2.7/site-packages/pipeline/templatetags/pipeline.py", line 72, in render_compressed
webapp_1         |     package_type)
webapp_1         |   File "/usr/local/lib/python2.7/site-packages/pipeline/templatetags/pipeline.py", line 98, in render_compressed_sources
webapp_1         |     default_collector.collect(self.request)
webapp_1         |   File "/usr/local/lib/python2.7/site-packages/pipeline/collector.py", line 45, in collect
webapp_1         |     for path, storage in finder.list(['CVS', '.*', '*~']):
webapp_1         |   File "/usr/local/lib/python2.7/site-packages/django/contrib/staticfiles/finders.py", line 112, in list
webapp_1         |     for path in utils.get_files(storage, ignore_patterns):
webapp_1         |   File "/usr/local/lib/python2.7/site-packages/django/contrib/staticfiles/utils.py", line 28, in get_files
webapp_1         |     directories, files = storage.listdir(location)
webapp_1         |   File "/usr/local/lib/python2.7/site-packages/django/core/files/storage.py", line 397, in listdir
webapp_1         |     for entry in os.listdir(path):
webapp_1         | OSError: [Errno 2] No such file or directory: '/app/webapp-django/webpack_bundles'
```

I think that'll fix itself once the webpack service issue is figured out.

Everyone is using a Macbook except for me, so this issue only affects me and possible contributors that are using Linux and want to work on this. I'm game for landing this as is now and writing up a bug report and when I get back from PTO, I can look into. I have a Macbook Pro and a Linux laptop, so I can tinker with it until I get a solution that works on both platforms. Permissions are tricky. We have this ridiculous `./docker/as_me.sh` script to get around permissions problems. It'd be nice to ditch that. Maybe we roll fixing permissions issues into the "let's do one image to rule them all" work.

It's possible I might be able to work on this tomorrow, but I'm running out of time, so I can't make any promises.

willkg:
I couldn't stop thinking about it. Pretty sure it's all fixed in #4505. If we land that first, then I think that will fix my issues here.

.dockerignore
Osmose: We don't want `webapp-django/node_modules` because the node dependencies are currently being installed in `webapp-frontend-deps`. I have a `node_modules` in my webapp directory because that's how Atom finds which eslint to run for linting the project, and having it get copied in masks problems with missing dependencies that come up during automation, when there is no `node_modules` to copy.

`static` is in there because it gets built during image creation and doesn't need to be copied in. `.dockerignore` doesn't affect volume mounts so the static files still get correctly updated during development.

Makefile
Osmose: Yes, it should produce error output when there's a JS error that breaks the webpack build, and output when a re-build runs to help confirm that the latest JS files are being served. Weirdly, that output isn't showing up with the current setup. I'm punting figuring out why that is to another time.

docker-compose.yml
Osmose: The problem is that services in `depends_on` stick around after you close the service you started with `up`, which means webpack sticks around and keeps re-compiling your frontend files, potentially with an out-of-date config (since webpack only updates the config on startup).

If there's a way to say "Start this services the webapp is up, and stop it when it is down" I'd be down for that.

willkg: Mmm... That makes sense. We should update this documentation in some way:

https://socorro.readthedocs.io/en/latest/service/webapp.html#running-the-webapp

PR #4496, opened 2018-06-25
fix bug 1460447: switch to glom

willkg:
This drops our homegrown treelib for glom for manipulating complex Python
data structures.

Mostly this is a lot of removed code. `glom` works similarly to how `tree_get` worked, so much so that the changes are pretty minimal.


willkg:
Glom docs are here: https://glom.readthedocs.io/en/latest/

The treelib docs are in the code. I think the docstring for `tree_get` is pretty decent.

Both of them take a "path spec" as an argument that dictates how to traverse the structure. For `tree_get`, I required `[` and `]` for list indexes thus differentiating them from dict keys. `glom` doesn't do that. Pretty much that's the big difference between the two that affects this code.

willkg:
Thank you!

PR #4497, opened 2018-06-25
fix bug 1470895: move siglists/ under socorro/signature/

willkg:
The siglists are used by the signature generator and only the signature
generator. To make the signature generator a self-contained module, we
need to move the siglists folder under it.

There isn't anything fancy here. It moves a directory under another directory and updates the tests and documentation accordingly.


willkg:
I though about putting the tests in a subfolder, but then didn't figuring they should all be together. The reason we have `siglists` in a subfolder is that it makes it easier for other non-socorro-devs people to edit since there are fewer things in the directory and only the things that they'd want to edit.

We could go a step further and move the code part of siglists into the signature directory. That might be interesting. So then the siglists directory is *just* the text files that people edit and the README.

Does that seem like a good idea?

AlexisDeschamps:
That seems like a good idea from what I understand and if you want to implement it.

willkg:
@AlexisDeschamps ^^^ What do you think about that? Does that work better structurally?

willkg:
Thank you!

PR #4499, opened 2018-06-26
fix bug 1471299: collapse fs crashstorage classes

willkg:
This removes a bunch of unused fs crashstorage code. The remaining `FSPermanentStorage` isn't used in our server environments, but I've used it once or twice over the last couple of years and it's the default storage for the processor so I think we might as well keep it for now.

There really isn't much to test here since none of this is used in the server environment or the local development environment.


willkg:
Sorry about that--fixed two minor other things while I was at it. This is ready for review now.

socorro/external/fs/crashstorage.py
```diff
@@ -2,12 +2,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import datetime
+import gzip
 import json
```
AlexisDeschamps: I'm curious about the order of imports for within the same type of import. Is there a standard within the project? It seems that a lot of it alphabetical.

willkg: Yeah!

So, we do groups:

* python stdlib
* 3rd party libraries
* django things
* socorro things

In each group, we order them alphabetically.

We use flake8-isort to lint it. Though it looks like we have it shut off for everything except the e2e-tests directory.

https://github.com/timothycrosley/isort

willkg:
Thank you!

PR #4500, opened 2018-06-27
bug 1402838 Catch UnicodeEncodeError in SignatureIPCChannelError

is2ei:
I added `try ... catch` statement to catch `UnicodeEncodeError` in `SignatureIPCChannelError`.
Please be noted that I added `# -*- coding: utf-8 -*-` . This is for testing. Python will not accept multi byte character without the keyword.


socorro/signature/rules.py
```diff
@@ -725,12 +725,20 @@ def action(self, raw_crash, processed_crash, notes):
...
-        new_sig = new_sig.format(raw_crash['ipc_channel_error'][:100])
 
-        notes.append(
-            'Signature replaced with an IPC Channel Error, '
-            'was: "{}"'.format(processed_crash['signature'])
-        )
+        try:
+            new_sig = new_sig.format(raw_crash['ipc_channel_error'][:100])
+        except UnicodeEncodeError as ex:
+            new_sig = new_sig.format(ex)
```
willkg: Mmm... We don't want the exception to be part of the signature. That doesn't make any sense in regards to what signatures are.

I think instead of doing what you're doing here, I'd do what we do with the `AbortSignature` rule where it drops non-ascii characters using `socorro.lib.util.drop_unicode` (which has both a terrible name and a terrible docstring) and then goes with that.

socorro/unittest/signature/test_rules.py
```diff
@@ -1575,6 +1577,33 @@ def test_action_success(self):
         expected = ['Signature replaced with an IPC Channel Error, was: "foo::bar"']
         assert notes == expected
 
+    def test_action_non_ascii(self):
+        rule = SignatureIPCChannelError()
+
+        raw_crash = {
+            'ipc_channel_error': u'あわわ'
```
willkg: Instead of including unicode characters directly, I'd use the `\x` or `\u` ways to represent them. That's easier on tooling and editors. あ is `u'\u3042'` and わ is `u'\u308f'`.

Granted this is really frustrating--it'd be much better if tools, editors, fonts, and such supported unicode better.

socorro/signature/rules.py
is2ei: Ok! I'll try.

socorro/unittest/signature/test_rules.py
is2ei: I see! I will change it.

is2ei:
@willkg 
Thanks for your comment! It helps a lot :+1: 
I updated the change. Could you please check it?

socorro/signature/rules.py
```diff
@@ -725,11 +725,11 @@ def action(self, raw_crash, processed_crash, notes):
             new_sig = 'IPCError-browser | {}'
         else:
             new_sig = 'IPCError-content | {}'
-        new_sig = new_sig.format(raw_crash['ipc_channel_error'][:100])
+        new_sig = new_sig.format(drop_unicode(raw_crash['ipc_channel_error'][:100]))
```
willkg: I would do `drop_unicode` and then truncate to 100 characters. That way we have as many ascii characters as possible.

socorro/signature/rules.py
```diff
@@ -725,11 +725,11 @@ def action(self, raw_crash, processed_crash, notes):
...
 
         notes.append(
             'Signature replaced with an IPC Channel Error, '
-            'was: "{}"'.format(processed_crash['signature'])
+            'was: "{}"'.format(drop_unicode(processed_crash['signature']))
```
willkg: I'm pretty sure you don't need to call `drop_unicode` here. Did you hit problems which required us to do this?

is2ei: I thought `processed_crash['signature']` may contain multi byte character
https://bugzilla.mozilla.org/show_bug.cgi?id=1402838#c0

```
UnicodeEncodeError: 'ascii' codec can't encode characters in position 180-183: ordinal not in range(128)
  File "socorro/signature/__init__.py", line 98, in generate
    rule.action(raw_crash, processed_crash, notes)
  File "socorro/signature/signature_utilities.py", line 674, in action
    'was: "{}"'.format(processed_crash['signature'])
```

socorro/signature/rules.py
is2ei: Thanks! I will change it.

socorro/signature/rules.py
willkg: Oh noes! I messed up--this should be two separate bugs. The issue in the description stems from a previous rule creating a signature that had non-ASCII characters in it. That's a completely separate issue from the SignatureIPCChannelError rule creating a signature with non-ASCII characters in it.

For this PR, let's solve just the SignatureIPCChannelError issue. If we hit another problem where the signature has non-ASCII characters in it, we can use that sentry report to track down which rule is the culprit and I'll write up a bug to fix it then. It's possible it's already fixed.

Sorry about that!

is2ei:
@willkg Thanks! I fixed it.

willkg:
@is2ei When you say "I fixed it", what do you mean? It's still calling `drop_unicode` on `processed_crash['signature']`.

is2ei:
@willkg Oh sorry! I didn't notice the comment https://github.com/mozilla-services/socorro/pull/4500#discussion_r198676593 .
I will fix it after I finished daytime work. Maybe 10 hours later 😅

is2ei:
@willkg I removed `drop_unicode` from `processed_crash['signature']` .
Could you check it, please? :smile:

PR #4501, opened 2018-06-27
[WIP] fix bug 1344864: add startup icons signature page

AlexisDeschamps:
fixes bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1344864
This patch is a work in progress.

## what
- added icons to the signature report page that indicate startup crash statistics
  - these are the same icons found on the topcrashers page

## why
- being able to distinguish startup crashes is useful

## notes
I would love to hear thoughts on the following:
- the bug mentions adding a red banner, but I opted to try adding the icons from the top crashers page first because
  - having the same indicators everywhere leads to a more consistent experience
  - these icons theoretically reveal more information that just one banner
  - these icons could also be added to Super Search results, which has the similar space requirements to the top crashers page

- The template macro and part of startup statistics computation code was copied from top crashers files. I plan to modify the top crashers files to instead use the code I extracted in order to promote reusability.

## visuals
One can see the added icons by looking for the red rocket icon:
screen shot 2018-06-27 at 2 07 30 pm

### DIFF
--- a/webapp-django/crashstats/base/jinja2/macros/signature_startup_stats_icons.html
+++ b/webapp-django/crashstats/base/jinja2/macros/signature_startup_stats_icons.html
@@ -1,0 +1,40 @@
+{% macro signature_startup_stats_icons(startup_stats) %}
+{% if startup_stats %}
+<div class="signature-icons">
+    {% if startup_stats.startup_count > 0 %}
+    {% if startup_stats.startup_count == startup_stats.crash_count %}
+    <img src="{{ static('img/3rdparty/silk/flag_red.png') }}"
+        alt="red flag"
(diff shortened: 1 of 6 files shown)


willkg:
Signatures can be as long as 255 characters. What happens when you have a really long signature? Where do the icons end up?

AlexisDeschamps:
@willkg This is what it looks right now with a very long signature and no icons: https://screenshots.firefox.com/26Ym5wpidvixOVLf/localhost. Not sure if that would be worth trying to fix. I guess you could put the filters on the next line.

The icons should be right before the paragraph for the filters.

AlexisDeschamps:
Replaced by #4501.

PR #4502, opened 2018-06-28
bug 1460447: fix errant glom call

willkg:
This fixes a bug created when I switched the code from `tree_get` to
`glom`. `glom` requires the default to be a keyword argument.


willkg:
Thank you! Sorry about breaking this!

PR #4503, opened 2018-06-28
fix bug 1471973: fix signature generation docs

willkg:
This clarifies that you have to run the signature generation command line in
a Socorro environment and one way to start one.

@AlexisDeschamps Is this clearer?

### DIFF
--- a/socorro/signature/README.rst
+++ b/socorro/signature/README.rst
@@ -15,9 +15,17 @@ signature and the newly generated one.
 This can be used for testing signature generation changes, regression testing,
 and astounding your friends at parties.
 
-To use::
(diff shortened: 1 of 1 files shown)


AlexisDeschamps:
@willkg I think this is definitely clearer. Thanks! Though, I think having `socorro-cmd signature CRASHID [CRASHID ...]` was quite useful.

willkg:
The usage is covered by doing `socorro-cmd signature --help` and it shows more information. I was thinking that was sufficient and then we don't have to update it whenever we tweak the arguments.

Let me think about this some more.

willkg:
@AlexisDeschamps ^^^ What about that? Is that helpful?

PR #4505, opened 2018-06-29
fix bug 1472060: fix user creation for easier dev

willkg:
This fixes user creation in the base image so that we don't have to do
circus stunts to get around file permission issues between the container
user and the host user in a local dev environment.

The key here is ARG. ARG is the answer! ARG will solve all things!
ARG can make you cake on those days when you are sad and wondering if
maybe you've given up all hope. Everyone loves cake. ARG is good.


willkg:
I want to make sure the tests build with 10001:10001. If that works, then to test this:

1. run `make build`
2. run `make test`
3. run `docker-compose run processor bash` and create a file and make sure it's owned by you in the host

Pretty sure that's it.

We can follow this up with undoing the house-of-mirrors that is building static assets, but it's "ok" as it is now and I'm hesitant to do that just before I go on PTO. Someone else can tackle it if they like or I can do it when I get back.

willkg:
@Osmose I'm 95% sure if this is a good change (and all signs point to "yes" so far), then if we land this, it'll fix all the issues I'm having with #4494.

Makefile
```diff
@@ -4,14 +4,20 @@
 
 .PHONY: clean default docs help lint
 
+# Set these in the environment to override them. This is helpful for
+# development if you have file ownership problems because the user
+# in the container doesn't match the user on your host.
+SOCORRO_UID ?= 10001
+SOCORRO_GID ?= 10001
```
willkg: Because we're always passing `--build-arg` in `make build` and that rule gets used everywhere, it's best to have the defaults here.

However, this duplicates the same default values in the `Dockerfile`. I'm not sure what to do about that. We should never change these values since they're ops-approved. Maybe we can just leave it like this?

Makefile
```diff
@@ -4,14 +4,20 @@
...
+
 default:
 	@echo "You need to specify a subcommand. See \"make help\" for options."
 	@exit 1
 
 help:
 	@echo "Socorro make rules:"
 	@echo ""
-	@echo "  build            - build docker containers for dev"
+	@echo "  build            - build docker containers"
```
willkg: We use this rule for everything: dev, server environments, testing in CI, making waffles, etc. Figured I'd fix the help text while I was here.

docker/as_me.sh
```diff
@@ -1,29 +0,0 @@
...
-
-if [[ $1 == "--container" ]]; then
-    shift;
-    CONTAINER=$1
-    shift;
-else
-    CONTAINER=processor
-fi
-
-docker-compose run -u "${HOSTUSER}" ${CONTAINER} $@
```
willkg: This script was such a kludge. Glad to see it go.

docker/set_up_ubuntu.sh
```diff
@@ -24,18 +24,14 @@ apt-get install -y gcc apt-transport-https build-essential python-dev \
...
 # Stomp on the bash prompt with something more useful for development.
 cat > /etc/bash.bashrc <<EOF
-# Get the name for the current uid and drop error messages.
-MYUSER="\$(id -u -n 2> /dev/null)"
-
-# If id -u -n 2 didn't return a 0 exit code, then it's because there is
-# no name for that uid in the container. So set it to "you".
-if [[ "\$?" != "0" ]]; then
-    MYUSER="you"
-fi
```
willkg: This was a bunch of mumbo jumbo to deal with setting the userid part of the prompt so that it was helpful for people. So if you used the `as_me.sh` script, you'd get `you` and if you didn't, it'd be `app`.

We don't need to worry about that anymore. We'll always use `app` now.

docs/gettingstarted.rst
```diff
@@ -51,6 +51,17 @@ Quickstart
...
 
+   .. NOTE::
+
+      If you're on Linux and want the host user id to match the user id in the
+      containers, then you should do::
+
+        $ SOCORRO_UID=$(id -u) SOCORRO_GID=$(id -g) make build
+
+
+      Probably best to make a script out of that if you do it often.
```
willkg: This is the magic sauce that makes it all work for me. This is the point in the docs where we bring up `make build` the first time, so it seemed to make sense to put here. Maybe there are other good places?

docs/service/processor.rst
```diff
@@ -80,27 +80,27 @@ appropriate directory structure rooted at outputdir.
...
 You can get command help::
 
-  $ docker/as_me.sh ./socorro-cmd fetch_crash_data --help
+  $ docker-compose run processor ./socorro-cmd fetch_crash_data --help
 
 
-You should run this with ``docker/as_me.sh`` so that the files that get saved to
+You should run this with ``docker-compose run processor`` so that the files that get saved to
 the file system are owned by the user/group of the account you're using on your
 host.
```
willkg: Oh, whoops--don't need this anymore.

willkg:
Ok. I skimmed it. It's good for review.

docs/gettingstarted.rst
willkg: I spent some time trying to do two rules, but it's tricky because "make build" is a dependency for `.docker-build` and I don't know how to get it so that either rule can satisfy the requirement.

We could wrap the `${DC} build base` part in something that pulls the `my.env` environment. That's an interesting idea. We can resurrect the script you deleted when you did the config overhaul--it does this. That's an interesting idea and it uses `my.env` which we're pushing as the way to configure all things locally rather than another separate thing.

I'll try to do that.

docs/gettingstarted.rst
```diff
@@ -38,31 +38,45 @@ Quickstart
...
+
+
+3. (*Optional/Advanced*) Set UID and GID for Docker container user.
+
+   If you're on Linux or you want to set the UID/GID of the app user that
+   runs in the Docker containers, run::
+
+     $ make my.env
+
+
```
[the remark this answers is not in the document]
Osmose: Eh, it's two newlines in most (but not all) other spots in the file. I'll just leave it alone.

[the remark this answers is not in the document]
willkg: This is my fault. rst doesn't explicitly require two lines after literal blocks, but I find it's a lot easier to read in text files, so I usually do two lines.

I'm game for fixing everything to one line if you think that's helpful. I'm inclined to write a post-it on my monitor to remind myself to stop the madness but otherwise not do anything here.

willkg:
@Osmose Thank you for landing this!

PR #4506, opened 2018-06-29
fix bug 1344864: added startup stats to the signature summary

AlexisDeschamps:
fixes bug https://bugzilla.mozilla.org/show_bug.cgi?id=1344864

## why:
- being able to distinguish startup crashes from non-startup crashes is useful

## what:
- added icons and text to the signature summary tab that indicate startup crash statistics
  - these are the same icons as those found on the topcrashers page

## notes
I split the changes into three commits:
- first commit does the refactoring of topcrashers
  - needed to extract some of the functionality used to generate topcrashers data because part of it will be used to generate signature summary data
  - started generating data through classes instead of a dictionary
  - extracted computations into methods
  - this extraction should also be useful for https://bugzilla.mozilla.org/show_bug.cgi?id=1472335
- second commit adds the startup stats to the signature summary
- third commit adds review updated:
  - converted topcrashers signature icons into macros
    - started using these macros in the topcrashers page
    - started using these macros in signature summary
  - updated SignatureStats class
    - merged SignatureStartupStats into SignatureStats
    - started doing the previous results checks through properties instead of after the fact
    - made all properties be cached_property's
    - changed properties for startup, hang, and plugin stats to be booleans instead of numbers
    - added unit test
  - updated signature summary
    - added custom LESS classes to change color and layout
  - updated topcrashers
    - removed `get_signatures_stats` util method and started doing the work in views.py instead
    - change variable naming to be simpler and more informative

## visuals (updated)
startup crash signature
screen shot 2018-07-05 at 12 41 56 pm

non-startup crash signature
screen shot 2018-07-05 at 12 42 01 pm

### DIFF
--- a/webapp-django/crashstats/signature/jinja2/signature/signature_summary.html
+++ b/webapp-django/crashstats/signature/jinja2/signature/signature_summary.html
@@ -1,59 +1,55 @@
-{% if startup_stats %}
-<div class="panel tab-inner-panel">
-    <div class="status-message severity-warning">
-        {% if startup_stats.startup_count > 0 %}
-        {% if startup_stats.startup_count == startup_stats.count %}
-        <div>
-            <img src="{{ static('img/3rdparty/silk/flag_red.png') }}"
-                alt="red flag"
-                title="Startup Crash, all crashes happened during startup"
-                class="startup" height="16" width="16"
-            />
-            Startup Crash, all crashes happened during startup
-        </div>
-        {% else %}
-        <div>
-            <img src="{{ static('img/3rdparty/silk/flag_yellow.png') }}"
-                alt="yellow flag"
-                title="Potential Startup Crash, {{ startup_stats.startup_count }} out of {{ startup_stats.count }} crashes happened during startup"
-                class="startup" height="16" width="16"
-            />
-            Potential Startup Crash, {{ startup_stats.startup_count }} out of {{ startup_stats.count }} crashes happened during startup
-        </div>
-        {% endif %}
-        {% endif %}
+{% from "macros/signature_startup_icons.html" import startup_crash_icon,
+                                                     potential_startup_crash_icon,
+                                                     potential_startup_window_crash_icon,
+                                                     hang_crash_icon, plugin_crash_icon,
+                                                     browser_crash_icon %}
 
-        {% if startup_stats.startup_crash %}
-        <div>
-            <img
-            src="{{ static('img/icons/rocket_fly.png') }}"
-            alt="rocket"
-            title="Potential Startup Crash, more than half of the crashes happened during the first minute after launch"
-            class="startup" height="16" width="16"
-            />
-            Potential Startup Crash, more than half of the crashes happened during the first minute after launch
-        </div>
-        {% endif %}
-
-        {% if startup_stats.hang_count > 0 %}
-        <div>
-            <img src="{{ static('img/3rdparty/fatcow/stop16x16.png') }}" alt="stop sign" title="Hanged Crash" class="hang" height="16" width="16">
-            Hanged Crash
-        </div>
-        {% endif %}
-
-        {% if startup_stats.plugin_count > 0 %}
-        <div>
-            <img src="{{ static('img/3rdparty/fatcow/brick16x16.png') }}" alt="brick" title="Plugin Crash" class="plugin" height="16" width="16">
-            Plugin Crash
-        </div>
-        {% else %}
-        <div>
-            <img src="{{ static('img/3rdparty/fatcow/application16x16.png') }}" width="16" height="16" alt="application" title="Browser Crash" class="browser" />
-            Browser Crash
-        </div>
-        {% endif %}
+{% if signature_stats %}
+<div class="panel tab-inner-panel crash-type-indicators-panel
+            {% if signature_stats.is_startup_related_crash %} startup-crash-warning {% endif %}">
+    {% if signature_stats.is_startup_crash %}
+    <div class="crash-type-indicator-container">
+        {{ startup_crash_icon() }}
+        Startup Crash, all crashes happened during startup
     </div>
+    {% endif %}
+
+    {% if signature_stats.is_potential_startup_crash %}
+    <div class="crash-type-indicator-container">
+        {{ potential_startup_crash_icon(
+            num_startup_crashes=signature_stats.num_startup_crashes,
+            num_crashes=signature_stats.num_crashes,
+        ) }}
+        Potential Startup Crash, {{ signature_stats.num_startup_crashes }}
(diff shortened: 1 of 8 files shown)


webapp-django/crashstats/topcrashers/views.py
```diff
@@ -298,20 +209,18 @@ def topcrashers(request, days=None, possible_days=None, default_context=None):
         _range_type=range_type,
     )
 
-    if api_results['total'] > 0:
-        tcbs = api_results['facets']['signature']
-    else:
-        tcbs = []
+    tcbs = api_results
```
Osmose: This assignment is unnecessary.

Also, I'm a little worried about the noun mismatch here; `get_topcrashers_results` returns `SignatureStats` instances, instead of "results" (whatever that means). Then, we call the things inside `tcbs` "crashes", even though they're more stats about crashes.

Should we call these things TopCrashers?

webapp-django/crashstats/signature/views.py
```diff
@@ -404,6 +404,12 @@ def signature_summary(request, params):
     context = {}
 
     params['signature'] = '=' + params['signature'][0]
+    params['_aggs.signature'] = [
```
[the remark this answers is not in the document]
AlexisDeschamps: This makes the SuperSearch query get the data we need to display the startup statistics.

Osmose:
Moving to @willkg for when he gets back; I think a second set of eyes on this would be good, and deferring would help with some commerce stuff I need to work on.

webapp-django/crashstats/topcrashers/jinja2/topcrashers/topcrashers.html
```diff
@@ -128,104 +134,96 @@ <h2>
                     </tr>
                 </thead>
                 <tbody>
-                {% for crash in tcbs %}
+                {% for topcrasher_stats in topcrashers_stats %}
```
willkg: These two variable names differ in a single `s` in the middle of a long name. I'd much rather the variable be shorter and less similar. It'd make reading the rest of this code a lot easier for me. This is super easy to typo and make mistakes and the way we've got Jinja2 set up, we could miss typos.

Maybe something like:

```
{% for tcs in topcrashers_stats %}
```

if you're into abbreviations or perhaps:

```
{% for item in topcrashers_stats %}
```

if you're into that sort of thing.

webapp-django/crashstats/topcrashers/jinja2/topcrashers/topcrashers.html
```diff
@@ -128,104 +134,96 @@ <h2>
...
-                                    title="Startup Crash, all crashes happened during startup"
-                                    class="startup" height="16" width="16"
-                                />
-                                {% else %}
-                                <img src="{{ static('img/3rdparty/silk/flag_yellow.png') }}"
-                                    alt="yellow flag"
-                                    title="Potential Startup Crash, {{ crash.startup_count }} out of {{ crash.count }} crashes happened during startup"
-                                    class="startup" height="16" width="16"
-                                />
+                                {% if topcrasher_stats.is_startup_crash %}
```
willkg: Huge :clap: for moving the logic into the view and `SignatureStats` thing.

webapp-django/crashstats/topcrashers/views.py
```diff
@@ -150,30 +80,19 @@ def get_topcrashers_results(**kwargs):
...
-                hit['rank_diff'] = sig['rank'] - hit['rank']
-                hit['previous_percent'] = sig['percent']
-            else:
-                hit['diff'] = 'new'
-                hit['rank_diff'] = 0
-                hit['previous_percent'] = 0
+        previous_signatures = get_comparison_signatures(previous_range_results)
 
-    return search_results
+        signatures_stats = []
```
willkg: Unless I'm mistaken, I think you've already set this in line 61.

webapp-django/crashstats/topcrashers/jinja2/topcrashers/topcrashers.html
willkg: Or even:

```
{% for topcrashers_stats_item in topcrashers_stats %}
```

Also, this happens in a few other places in this PR.

AlexisDeschamps:
@willkg Thanks for the review. I added some updates in https://github.com/mozilla-services/socorro/pull/4506/commits/c2bb2bea29ce6a85a9eaab99047db6e354643a0a.

willkg:
Awesome! I'll work through testing this out today.

webapp-django/crashstats/signature/jinja2/signature/signature_summary.html
```diff
@@ -1,3 +1,58 @@
...
+    {% if signature_stats.is_potential_startup_crash %}
+    <div class="crash-type-indicator-container">
+        {{ potential_startup_crash_icon(
+            num_startup_crashes=signature_stats.num_startup_crashes,
+            num_crashes=signature_stats.num_crashes,
+        ) }}
+        Potential Startup Crash, {{ signature_stats.num_startup_crashes }}
+        out of {{ signature_stats.num_crashes }} crashes happened during startup
+    </div>
+    {% endif %}
```
willkg: I think we should show either `is_startup_crash` OR `is_potential_startup_crash` OR neither of them, but never both. I'd change this block to be an `else if` instead of an `if`.

AlexisDeschamps: The way the two properties are set up right now, they should be exclusive and never be true at the same time. `is_potential_startup_crash` needs to be under while `is_startup_crash` needs to be equal. I guess the wording makes it confusing, since one would expect a "potential startup crash" to also be a "startup crash".

With that in mind, would you make changes? I could rename `is_startup_crash` to `is_always_startup_crash`.

AlexisDeschamps: Yeah, that is what I would currently expect. The behavior should be the same as the current `topcrashers` page. The second icon red rocket icon is for a crash within one minute of launch, as opposed to pure startup.
So there are three startup crash icons:

1. All crashes happened directly during start up.
2. Some crashes happened directly during start up (should be exclusive with 1).
3. Some crashes happened during one minute of launch. This looks at a different source statistic than 1 and 2.

The user messaging for `#3` currently says `Potential Startup Crash`. I am using `is_startup_window_crash` in the code. I would personally change the user messaging to be `Potential Startup Window Crash` for better clarity. Do you think that is a change I could make?

willkg: Ahh... ok. I see what's going on now. I think this is fine. Thank you for clarifying!

PR #4508, opened 2018-07-05
bug 1469718: all the changes for sub dir socorro.external.boto

ceddy-cedd:
In testing the raw dump data there was an error in the test value
returning as a binary string. This value was failing in comparison to
the expected unicode value for python 3. Placing print statements in
class level functions didn’t tell us much.So we traced the function from
top to bottom and bottom to top. This pushed over the deep end as we
found ourselves in the boto service source code. We couldn’t find the
exact time this conversion of the test value occurred. To solve this
issue we made the assumption data being used in raw dump is binary. So
we change the test value to be a binary string and also changed the
expected comparison to be a binary string as well. The implementation of
the function iteritems() changed from python 2 to 3. Similar to the
range function the change in implementation went from generating a list
to using an iterator. Lastly objects of type ‘dict_values’ is not JSON
serializable. To solve this we wrap the call to values with a call to
list.

### DIFF
--- a/docker/run_tests_python3.sh
+++ b/docker/run_tests_python3.sh
@@ -21,6 +21,33 @@ WORKING_TESTS=(
     socorro/unittest/lib/test_transform_rules.py
     socorro/unittest/lib/test_util.py
     socorro/unittest/lib/test_vertools.py
+
+    socorro/unittest/external/boto/test_connection_context.py
+    socorro/unittest/external/boto/test_crash_data.py
+    socorro/unittest/external/boto/test_crashstorage.py
+    socorro/unittest/external/boto/test_upload_telemetry_schema.py
+
+    socorro/unittest/external/es/test_analyzers.py
+    socorro/unittest/external/es/test_connection_context.py
+    socorro/unittest/external/es/test_index_creator.py
+    socorro/unittest/external/es/test_new_crash_source.py
+    socorro/unittest/external/es/test_query.py
+
+    socorro/unittest/external/postgresql/test_adi.py
+    socorro/unittest/external/postgresql/test_base.py
+    socorro/unittest/external/postgresql/test_bugs.py
+    socorro/unittest/external/postgresql/test_connection_context.py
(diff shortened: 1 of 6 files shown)


docker/run_tests_python3.sh
```diff
@@ -11,6 +11,7 @@
 # This is the list of known working tests by directory/filename. When you
 # have tests in a directory/file working, add it to this list as a new line.
 WORKING_TESTS=(
+    #socorro/unittest/lib/*.py
```
Osmose: Could we remove the commented-out tests? Since they're commented out, they're not being run, and including them in the PR that we merge just adds more lines that aren't really useful for other developers. You can add them back in to your file after you commit and push the changes so that they're still on your laptop to help you keep track.

I can also remove them when merging this. Let me know if I should do that instead. :D

docker/run_tests_python3.sh
```diff
@@ -22,6 +23,53 @@ WORKING_TESTS=(
...
+    socorro/unittest/external/postgresql/test_signature_first_date.py
+
+    #socorro/unittest/external/rabbitmq/test_connection_context.py
+    #socorro/unittest/external/rabbitmq/test_crashstorage.py
+    socorro/unittest/external/rabbitmq/test_reprocessing.py
+    socorro/unittest/external/rabbitmq/test_rmq_new_crash_source.py
+
+
+
+
```
Osmose: These extra lines at the end aren't adding anything, can we remove them?

socorro/external/crashstorage_base.py
```diff
@@ -63,7 +64,7 @@ def as_file_dumps_mapping(self, crash_id, temp_path, dump_file_suffix):
         """convert this MemoryDumpMapping into a FileDumpsMappng by writing
         each of the dump to a filesystem."""
         name_to_pathname_mapping = FileDumpsMapping()
-        for a_dump_name, a_dump in self.iteritems():
```
Osmose: Aha, this one looks particularly tricky (since it's a dict subclass), nice work!

PR #4512, opened 2018-07-12
fix bug 1369498, 1477038: remove ADI

lonnen:
This removes most interfaces to ADI, ADU, and associated machinery.

TODO:
- [x] migrations to remove stored procedures
- [x] address `socorro/external/postgresql/raw_sql/*`

follow up actions:
* remove or shutdown the VM running hive externally


lonnen:
@willkg -- when you get back. Our documentation for writing migrations is... not great. You've recently landed one or two. What say you?

socorro/cron/crontabber_app.py
```diff
@@ -46,17 +39,14 @@
...
     ]
 )
 
 
 # Jobs that run in the -stage-new and -prod-new environments
 STAGE_NEW_JOBS = ', '.join(
-    DEFAULT_JOBS_BASE + [
-        'socorro.cron.jobs.fetch_adi_alt.RawADIMoverCronApp|1d|08:20'
-    ]
+    DEFAULT_JOBS_BASE + []
```
willkg: Two things. First, you could skip the `+ []` part since that's not really doing anything.

Second, if this is `STAGE_NEW_JOBS`, I wonder what prod is running now and it's probably worth writing up a bug and making sure all the environments are running the right set of cron jobs.

Also, also, it's worth noting that if we ditch the dependency security check cron app, then all our environments are the same which would be nice. That doesn't affect this PR at all, but I figured I'd mention it so someone remembers later.

socorro/cron/jobs/fetch_adi_from_hive.py
```diff
@@ -1,391 +0,0 @@
...
-* build
-* build_channel
-* product_guid
-* count
-
-For alternate crontabber apps for "acquiring" ADI data, see
-``fetch_adi_alt.py``
-
-NOTE(willkg): This runs on a box that is a maintenance nightmare. Avoid
-adjusting the contents of this script at all costs.
```
willkg: I'm going to miss this comment.

socorro/unittest/cron/jobs/test_matviews.py
```diff
@@ -132,10 +128,7 @@ def mock_utc_now():
...
         information = self._load_structure()
 
-        for app_name in ('product-versions-matview',
-                         'adu-matview',
-                         'build-adu-matview'):
-
+        for app_name in ('product-versions-matview',):
             assert app_name in information
             assert not information[app_name]['last_error']
             assert information[app_name]['last_success']
```
willkg: Since this is iterating over one item, we can nix the iteration.

willkg:
Migration-wise, I think we need to do the following:

1. drop the `raw_adi_logs` table
2. drop the `raw_adi` table
3. drop the `build_adu` table
4. drop the `product_adu` table
5. drop `backfill_adu(updateday date)`
6. drop `backfill_build_adu(updateday date)`
7. drop `backfill_matviews(firstday date, lastday date DEFAULT NULL::date, reportsclean boolean DEFAULT true, check_period interval DEFAULT '01:00:00'::interval)`
8. drop `update_adu(updateday date, checkdata boolean DEFAULT true)`
9. drop `update_build_adu(updateday date, checkdata boolean DEFAULT true)`

@lonnen Thank you SO much for this! Do you want to keep working on this? If not, I can take it. Either is fine by me.

lonnen:
I’ll hand it off now, and I would like to understand how migrations are
made at some point.

willkg:
Oh, my bad. I thought that since you had fixed the docs in the alembic `README`, you had figured it out. As a side note, I didn't realize there was a `README` in the alembic directory--that's super old stuff. I should have been updating it.

I have been maintaining documentation on migrations in the Socorro Mana page. It was there since it included instructions on running migrations in stage and prod which were infrastructure-specific. Reasons reasons. It's probably a good idea to move that into the RTD docs and update the alembic `README` accordingly.

Sorry about that!

lonnen:
No worries!

willkg:
I'm pretty sure we can remove the `ProductVersionsCronApp` and all the `MatViews` related cron app stuff, too. That crontabber app calls `update_product_versions`, but that gets called by ftpscraper and any time a release is added. I don't think there's a situation where product/release data is adjusted and `update_product_versions` is not called. Given that, we should nix the crontabber app and related bits, too.

willkg:
This leaves a few things undone:

1. it doesn't clean up the navigation bar which is silly now--we should do that in a new bug and overhaul the entire header; we should move the product/version pickers to the topcrashers page
2. we should redo how the `/` page works and the product home pages work--maybe even merge them all into a single front page that covers all the products

willkg:
@AlexisDeschamps I reviewed the first commit--that's the one that has most of the changes in it. Can you review the four commits after that? Those need looking at.

willkg:
Also, when this lands, we need to fix the infra and remove the crontabber jobs list setting.

After this makes it to prod, we can remove the ADI generating box in scl3 and the old -stage and -prod databases. (YAY!)

I'll write up issues for those now.

alembic/versions/eb8269f6bb85_bug_1369498_remove_adi.py
```diff
@@ -0,0 +1,38 @@
+"""bug 1369498 remove adi
+
+Remove ADI-related tables and stored procedures.
+
+Revision ID: eb8269f6bb85
+Revises: 0db05da17ae8
```
AlexisDeschamps: What do `eb8269f6bb85` and `0db05da17ae8` refer to? Are they randomly generated?

AlexisDeschamps: Ah, cool, thanks.

willkg:
I want to do a prod deploy with what's on stage now, then land this and let it simmer on stage for a few days.

Going to mark it as DO NOT MERGE so I don't forget.

willkg:
Thank you! The future is here!

lonnen:
🌅

PR #4515, opened 2018-07-16
fix bug 1475993: add debug view with ip info

willkg:
This adds infrastructure for a debug view for showing admin arbitrary things.

This uses that infrastructure to display the `REMOTE_ADDR` and
`HTTP_X_FORWARDED_FOR` headers to help us figure out what's going on with
rate limiting.


willkg:
To test this:

1. start the webapp
2. log in with an admin account
3. go to the admin
4. click on "Debug information"

It should show you `none` for `HTTP_X_FORWARDED_FOR` since you're running in a dev environment with no load balancer and whatever your ip address is for `REMOTE_ADDR`.

The only reason this PR is so "big" is that I decided to throw this in a new page we can "abuse" for debugging without affecting the Site Status page.

webapp-django/crashstats/manage/templates/admin/debug_view.html
```diff
@@ -0,0 +1,35 @@
...
+            <table>
+                <thead>
+                    <tr>
+                        <th>Key</th>
+                        <th>Value</th>
+                    </tr>
+                </thead>
+                <tbody>
+                    {% for key, val in debug_info.items %}
+                        <tr class="row1">
```
AlexisDeschamps: What does `class="row1"` do?

webapp-django/crashstats/manage/admin.py
```diff
@@ -197,3 +198,23 @@ def graphics_devices_lookup(request):
...
+    """This view is for ephemeral debugging of issues"""
+
+    context = {}
+
+    # Map of key -> val that will get displayed in a big table in the debug
+    # view in the order they were inserted
+    debug_info = OrderedDict()
+
+    # Add IP address related headers to figure out rate-limiting issues. #1475993
+    for meta_header in ['HTTP_X_FORWARDED_FOR', 'REMOTE_ADDR']:
```
sciurus: Can you also add HTTP_X_REAL_IP? We're currently doing `proxy_set_header X-Real-IP $remote_addr;` in nginx, and having it here will give me some extra wiggle room for debugging since the middleware doesn't touch it.

webapp-django/crashstats/manage/templates/admin/debug_view.html
willkg: This is a Django admin template, so I've been using the classes Django defined. When I did the Site Status page, I figured out which classes to use to get close to what I needed. But that was a while ago and I don't remember why I did what I did. I should probably write up a style guide. Either that or we should switch to our own style classes--probably a better idea, but probably not for this PR.

webapp-django/crashstats/manage/admin.py
willkg: Sure!

willkg:
Thank you!

PR #4517, opened 2018-07-18
fix bug 1476697: switch from X_FORWARDED_FOR to X_REAL_IP

willkg:
This switches the middleware from looking at the `X_FORWARDED_FOR` header
and all its interetingness to looking at the `X_REAL_IP` header which
is set by nginx module we're using.

### DIFF
--- a/webapp-django/crashstats/crashstats/middleware.py
+++ b/webapp-django/crashstats/crashstats/middleware.py
@@ -1,7 +1,7 @@
 from django.shortcuts import render
 
 
-class SetRemoteAddrFromForwardedFor(object):
+class SetRemoteAddrFromRealIP(object):
     """
     Middleware that sets REMOTE_ADDR based on HTTP_X_REAL_IP, if the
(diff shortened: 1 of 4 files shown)


willkg:
The middleware is pretty simple and the tests cover the cases pretty well (assuming they're correct). So I think this just requires looking over.

This changes which header we're looking at. Additionally, `X-Real-IP` (if it's there) is just a single ip address--not a curiously sorted list of ip addresses.

It's definitely worth verifying that I didn't typo the header.

webapp-django/crashstats/crashstats/middleware.py
```diff
@@ -3,16 +3,17 @@
 
 class SetRemoteAddrFromForwardedFor(object):
```
sciurus: Shouldn't we rename the middleware, since it no longer looks at ForwardedFor?

willkg: Yes! I'll do that now.

willkg:
Thank you!

PR #4519, opened 2018-07-20
Add Flask based simple alternative to Socorro

uglide:
Since Socorro is python based project users can bee interested in a Python-based alternative.


willkg:
Hi! I appreciate you wanting to recommend your project, but I'm not familiar with this project at all and it doesn't look like it's far enough along. If you like, feel free to re-recommend it in the future once it has a more active community around it.

Sorry!

PR #4520, opened 2018-07-20
fix bug 1477344: add minidumpsha256hash to allow list

willkg:
This fixes the crash report view so that it shows the `MinidumpSha256Hash` field in the metadata tab for everyone.

To test:

1. run the webapp
2. process a crash
3. view the crash report page -- in the metadata tab, you'll see `MinidumpSha256Hash` field when not logged in


willkg:
Thank you!

PR #4521, opened 2018-07-23
Update reprocessing docs

willkg:
This updates the command line to use for those commands. This also removes
the note which isn't relevant anymore now that we have the new infra
which autoscales.


willkg:
This is pretty cosmetic and low risk--it only affects site admin.

willkg:
Thank you!

PR #4523, opened 2018-07-25
fix bug 1461456: add raw access docs

willkg:
This adds documentation on how to get raw dump access on the site. It's still a bit buried because of the way our docs are. Unburying it will require changing the header and that'll come in a later PR.

This not only adds docs on raw dump access, it codifies our policy and what users agree to, and includes a link to generate a bug report. This gives us a single place to point people to.

Further, this unifies all documentation that exists on Crash Stats so it has a single navigation scheme and is easier to find, access, and use.


willkg:
There's a "Documentation" link in the footer which leads to the documentation home with links to all the docs.

One thing I was on the fence about is that if "how do I get raw dump access" is a first-class thing, then maybe it should have its own page and not be lumped in with random things on the docs home page.

What do you think? If you were looking to get raw dump access, is it easier to see it in the nav bar or table-of-contents? If someone sent you a url for the docs for raw dump access, is it better to be in the url anchor or the url path?

willkg:
One nice thing about these changes is that they don't mess up any existing docs urls--all those continue to work with these changes.

willkg:
I think I convinced myself that it should be a separate page. I'm going to fix that now.

willkg:
^^^ I'm done fiddling with it--this is good to go now.

webapp-django/crashstats/api/jinja2/api/documentation.html
```diff
@@ -1,152 +1,139 @@
-{% extends "crashstats_base.html" %}
-
-{% block page_title %}API Documentation{% endblock %}
-
-{% block product_nav_filter %}&nbsp;{% endblock %}
```
willkg: The API docs used to live in their own little world. This PR unifies them with all the other docs so they have the same nav header and such.

webapp-django/crashstats/base/jinja2/footer_nav.html
```diff
@@ -3,10 +3,8 @@
         <b>Mozilla Crash Reports</b> - Powered by <a href="https://github.com/mozilla-services/socorro">Socorro</a> - All dates are UTC
     </div>
     <ul>
-        <li><a href="{{ url('documentation:home') }}">User Documentation</a></li>
-        <li><a href="{{ url('api:documentation') }}">API</a></li>
+        <li><a href="{{ url('documentation:home') }}">Documentation</a></li>
         <li><a href="{{ url('crashstats:crontabber_state') }}">Crontabber State</a></li>
-        <li><a href="https://github.com/mozilla-services/socorro">Source</a></li>
```
willkg: We have a Source link in the footer already. It's silly to have two.

webapp-django/crashstats/base/jinja2/footer_nav.html
```diff
@@ -3,10 +3,8 @@
         <b>Mozilla Crash Reports</b> - Powered by <a href="https://github.com/mozilla-services/socorro">Socorro</a> - All dates are UTC
     </div>
     <ul>
-        <li><a href="{{ url('documentation:home') }}">User Documentation</a></li>
-        <li><a href="{{ url('api:documentation') }}">API</a></li>
```
willkg: All docs are being pulled into a single documentation area including the API docs.

webapp-django/crashstats/documentation/jinja2/documentation/docs_base.html
```diff
@@ -0,0 +1,56 @@
...
+        <div id="table-of-content">
+            <h2>Table of content</h2>
+        </div>
+
+        {% block doc_content %}{% endblock %}
+    </div>
+    {% endblock %}
+</div>
+
+{% endblock %}
```
willkg: This is effectively based on the old `supersearch_documentation_base.html` template. One change I made was to macro-ize the `navlink` thing which makes that code readable.

webapp-django/crashstats/documentation/jinja2/documentation/docs_base.html
```diff
@@ -0,0 +1,56 @@
...
+{% javascript 'jsonview' %}
+{% javascript 'documentation' %}
+{% endblock %}
+
+{% block page_title%}Crash Stats Documentation: {% block doc_title %}Home{% endblock %}{% endblock %}
+
+{% block content %}
+<div id="mainbody">
+    <div class="page-heading">
+        <h2>{{ self.page_title() }}</h2>
```
willkg: `page_title` here is an attribute on the template that pulls in the `page_title` block value. That's how you use block values multiple times in the same Jinja2 template.

webapp-django/crashstats/documentation/jinja2/documentation/memory_dump_access.html
```diff
@@ -0,0 +1,49 @@
...
+* When finished with a dump, delete it (within 30 days)
+* Do not share personal data from the dump publicly (on bugzilla, irc, etc)
+* Do not share personal data with non-employees, including partner organizations, without
+  consultation and explicit permission from Lonnen (lonnen@mozilla.com); a signed NDA and
+  other arrangements will be required of the other organization
+* It is fine to share stack traces, analysis, and other non-personal data
+* Access is conditional on employment and will be revoked upon departure from Mozilla
+* Access will be removed after a year of inactivity
+
+The most recent copy of this agreement is at: https://crash-stats.mozilla.com/documentation/pii_access/</pre>
```
willkg: This agreement text exists in Mana. This moves it to be in the Crash Stats docs where it's easier to access.

webapp-django/crashstats/documentation/jinja2/documentation/supersearch/supersearch_documentation_base.html
```diff
@@ -1,56 +0,0 @@
...
-    <div class="panel">
-        <div id="table-of-content">
-            <h2>Table of content</h2>
-        </div>
-
-        {% block doc_content %}{% endblock %}
-    </div>
-</div>
-
-{% endblock %}
```
willkg: All the docs now derive from `docs_base.html`, so we can nix this template.

willkg:
Thank you!

PR #4524, opened 2018-07-26
Fix URL to memory dump access document

jcristau:


willkg:
Whoops--thank you for catching this!

PR #4525, opened 2018-07-26
bug 1477782: implement auditgroups Django command

willkg:
This implements an auditgroups Django command and also fixes a bunch of issues I hit when trying to do that.

The command defaults to `dryrun=True` which makes sure that if there's a bug and someone goofs when running it, it doesn't do anything. We'll flip that after we're happy with what it's doing.

To test it:

1. set up the webapp and run:
   ```
   app@webapp:/app/webapp-django$ ./manage.py makesuperuser
   ```
   which will make you a user in the "Hackers" group.
2. then run the command:
   ```
   app@webapp:/app/webapp-django$ ./manage.py auditgroups
   ```

It won't do much because there isn't anything to do. You can create a new user and fiddle with the variations from there using the Django shell. For example:

```
$ docker-compose run webapp bash
app@webapp:/app$ cd webapp-django
app@webapp:/app/webapp-django$ ./manage.py shell
>>> from django.contrib.auth.models import Group, User
>>> from crashstats.tokens.models import Token
>>> user = User.objects.create(email='bob@mozilla.com')
>>> group = Group.objects.get(name='Hackers')
>>> 
```

The tests show how what fields to manipulate.


docker/run_tests_in_docker.sh
```diff
@@ -17,8 +17,8 @@ set -v -e -x
 PS4="+ (run_tests_in_docker.sh): "
 
 DC="$(which docker-compose)"
-APP_UID="10001"
-APP_GID="10001"
+SOCORRO_UID=${SOCORRO_UID:-"10001"}
+SOCORRO_GID=${SOCORRO_GID:-"10001"}
```
willkg: This fixes a thing I broke when I redid how uid/gid works when building images a couple of weeks ago. Oops.

Now running `make testshell` and running pytest in it works--it was kicking up permission errors for me.

docker/run_update_data.sh
```diff
@@ -25,6 +25,7 @@ docker-compose run -u "${HOSTUSER}" crontabber ${CRONTABBERCMD} \
                --crontabber.class-FTPScraperCronApp.products=${PRODUCTS}
 
 # Update featured versions data based on release data
+docker-compose run crontabber ${CRONTABBERCMD} --reset-job=featured-versions-automatic
```
willkg: Before running crontabber jobs like this, we need to reset them first. Otherwise they might decide not to run.

webapp-django/crashstats/authentication/management/commands/makesuperuser.py
```diff
@@ -1,6 +1,9 @@
-from __future__ import print_function
+"""
+Makes a user into a superuser/staff. This is helpful for local development
+environment.
+"""
```
willkg: While working on `auditgroups`, I needed `makesuperuser` to add me to the `Hackers` group. Then I looked at the code and was like, "ugh", and fixed a bunch of issues in the command code and the test.

webapp-django/crashstats/authentication/management/commands/migrate-users-csv.py
```diff
@@ -1,42 +0,0 @@
...
-        with open(options['csvfile']) as f:
-            reader = csv.reader(f)
-            for row in reader:
-                if first and not options['include_first_row']:
-                    first = False
-                else:
-                    assert len(row) == 2, len(row)
-                    combos.append(row)
-
-        migrate_users(combos, dry_run=options['dry_run'])
```
willkg: This was almost certainly checked in by mistake years ago. Removing it here.

webapp-django/crashstats/authentication/management/commands/migrate-users-csv.py
```diff
@@ -1,42 +0,0 @@
-import csv
```
willkg: We don't need this anymore--it's for migration users from their email alias to their LDAP account.

webapp-django/crashstats/authentication/migration.py
```diff
@@ -1,65 +0,0 @@
-from __future__ import print_function
```
willkg: We don't need this anymore--it's for migration users from their email alias to their LDAP account.

webapp-django/crashstats/authentication/tests/test_migration.py
```diff
@@ -1,103 +0,0 @@
-import contextlib
```
willkg: We don't need this anymore--it's for migration users from their email alias to their LDAP account.

webapp-django/crashstats/authentication/management/commands/auditgroups.py
```diff
@@ -0,0 +1,72 @@
...
+            self.stdout.write('"Hackers" group does not exist.')
+            return
+
+        # Go through the users and mark the ones for removal
+        users_to_remove = []
+        for user in hackers_group.user_set.all():
+            if not user.is_active:
+                users_to_remove.append((user, '!is_active'))
+            elif not user.email.endswith(VALID_EMAIL_DOMAINS):
+                users_to_remove.append((user, 'invalid email'))
```
AlexisDeschamps: Would you not maybe want to call this error `invalid email domain`?

webapp-django/crashstats/authentication/management/commands/makesuperuser.py
```diff
@@ -30,10 +37,24 @@ def handle(self, **options):
...
+            # Add user to Hackers group
+            try:
+                hackers_group = Group.objects.get(name='Hackers')
+                if user in hackers_group.user_set.all():
+                    self.stdout.write('{} is already in Hackers group.'.format(user.email))
+                else:
+                    hackers_group.user_set.add(user)
+                    self.stdout.write('{} added to Hackers group.'.format(user.email))
+                    user.save()
+
```
AlexisDeschamps: Not sure if you meant to put that empty line before the `except` clause.

webapp-django/crashstats/authentication/management/commands/migrate-users-csv.py
AlexisDeschamps: I guess this is because of the recent move to Mozilla Single Sign On?

willkg: This is leftover from the previous auth migration when we moved from Persona to Google Auth.

webapp-django/crashstats/authentication/management/commands/auditgroups.py
willkg: That's a good idea!

webapp-django/crashstats/authentication/management/commands/makesuperuser.py
willkg: Oops--I can remove this.

willkg:
Thank you!

PR #4526, opened 2018-07-27
fix bug 1478741: add ftpscraper wrapper

willkg:
This adds a ftpscraper wrapper script to be used in local development environment which keeps a log of ftpscraper and if the log is less than 7 days old, it'll just replay the log on the next run.

For me, it takes 10 minutes to run ftpscraper. On the second run, replaying the log takes about 10 seconds. This radically reduces the amount of time it takes to slick and rebuild my local development environment which I do a lot.

To test:

1. run `make setup` which slicks the db and everything
2. run `make updatedata` which runs ftpscraper and will create an ftpscraper log in `./.cache/ftpscraper.log`
3. run `make setup` which slicks the db again
4. run `make updatedata` which will replay the ftpscraper log

### DIFF
--- a/docker/run_ftpscraper_wrapper.sh
+++ b/docker/run_ftpscraper_wrapper.sh
@@ -9,18 +9,19 @@
 
 set -eo pipefail
 
-# Fetch and update release information for these products (comma-delimited)
+# Fetch and update release information for these products (comma-delimited).
 PRODUCTS="firefox,mobile"
 CRONTABBERCMD="./socorro-cmd crontabber"
 CACHEDIR="./.cache"
 CACHEFILE="${CACHEDIR}/ftpscraper.log"
 
+# If the cache file exists...
 if [[ -e "${CACHEFILE}" ]]; then
-    # FIXME(willkg): check to see if it's older than a week.
-    FILEMODDATE=$(stat -c %Y ${CACHEFILE})
+    FILEMODDATE=$(date -r ${CACHEFILE} '+%s')
     NOW=$(date +%s)
     FILEAGE=$(((NOW - FILEMODDATE) / 60 / 60 / 24))
 
+    # And it's less than a week old... replay.
     if [[ ${FILEAGE} -lt 7 ]]; then
         echo "Found recent (${FILEAGE}d) ftpscraper log--replaying...."
         docker-compose run crontabber ./socorro-cmd replay_ftpscraper "${CACHEFILE}"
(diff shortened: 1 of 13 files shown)


docker/run_ftpscraper_wrapper.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# Checks the cached ftpscraper log and if it's less than 7 days old, replays it
+# which is super duper fast.
+
+set -eo pipefail
+
+# Fetch and update release information for these products (comma-delimited)
+PRODUCTS="firefox,mobile"
+CRONTABBERCMD="./socorro-cmd crontabber"
+CACHEDIR="./.cache"
```
AlexisDeschamps: Not sure if a line similar to: `# The assumption is that you're running this from /app inside the container` could still be useful?

willkg: The problem is that `stat` on osx doesn't support the -c flag. I'll rework how this works.

willkg:
^^^ That fixes the code that gets the modified time from the `ftpscraper.log` file in seconds so it works on osx and linux.

AlexisDeschamps:
Nice, thanks. It seems that the error is fixed for me and the caching seems to work locally as well. I am going to look more at the code now.

AlexisDeschamps:
The part about`make updatedata` in `https://github.com/mozilla-services/socorro/blob/master/docs/gettingstarted.rst` mentions running the command "daily". Maybe the documentation could mention the caching?

willkg:
I had forgotten about the docs--thanks! ^^^ That updates them.

docs/gettingstarted.rst
```diff
@@ -107,6 +107,13 @@ Quickstart
    Depending on what you're working on, you might want to run this weekly or
    maybe even daily.
 
+   .. Note::
+
+      This runs ftpscraper which takes 10 minutes to run. To save your time
```
AlexisDeschamps: I think the formatting might not be working: https://github.com/mozilla-services/socorro/blob/bb66fd03c1ee576e5587cdfe8fb6794f354beddc/docs/gettingstarted.rst? Should the note text be within a gray box? Not sure if it's because there's 3 spaces instead of 2.

willkg: Yes! So, "note" is not in the reST standard--it's defined in Sphinx:

http://www.sphinx-doc.org/en/stable/markup/para.html

GitHub is rendering reST using just reST and not the additional Sphinx stuff. So it just renders as a block with a title on GitHub.

You can see an example Note as rendered in our docs on RTD here:

https://socorro.readthedocs.io/en/latest/gettingstarted.html#updating-release-data

willkg:
Thank you! This will save us GOBS of time.

PR #4527, opened 2018-07-30
fix bug 1479485: added report details api

AlexisDeschamps:
fixes bug https://bugzilla.mozilla.org/show_bug.cgi?id=1479485

## why

## what
- added a API endpoint to get the data needed for the details tab
of report indexes in JSON

## example
screen shot 2018-07-30 at 9 46 40 am


webapp-django/crashstats/crashstats/views.py
```diff
@@ -301,21 +301,12 @@ def report_index(request, crash_id, default_context=None):
...
     context['bug_product_map'] = settings.BUG_PRODUCT_MAP
 
-    process_type = 'unknown'
-    if context['report']['process_type'] is None:
-        process_type = 'browser'
-    elif context['report']['process_type'] == 'plugin':
-        process_type = 'plugin'
-    elif context['report']['process_type'] == 'content':
-        process_type = 'content'
-    context['process_type'] = process_type
```
AlexisDeschamps: Looks like this is not used at all in `report_index.html`.

willkg:
I want to adjust how we're doing the cupcake and switch it so we're doing the cupcake against a new `cupcake-dev` branch rather than the `master` branch. This does a couple of things:

1. it lets us work through the cupcake in a way that doesn't require us to think through every decision because we're landing it in a permanent-ish kind of way
2. because of that, we can throw together the cupcake faster and get to a point where we can see the whole cupcake and evaluate it as a whole

I created a `cupcake-dev` branch. I'm thinking I'm going to let it be for a couple of weeks instead of keeping it up to date with `master`. We can merge changes at some point in the future when the cupcake is done.

Can you redo this PR so that instead of it being against the `master` branch it's against the `cupcake-dev` branch? I think that just involves closing this PR and creating a new one against the `cupcake-dev` branch.

AlexisDeschamps:
Replaced by #4531.

PR #4528, opened 2018-07-31
fix bug 1479097: sentry cleanup in processor

willkg:
This does two big things:

1. clean up a bunch of processor code in regards to redundant setup
2. unify the two "let's capture something and send it to sentry" mechanisms we had and also fix them to use `exc_info` rather than exception values

This gives us two nice things:

1. the signature generation code no longer has anything about sentry in it--that's helpful for merging the two forks of signature generation
2. we can reuse send-this-to-sentry in other parts of the processor which was previous hard-ish

### DIFF
--- a/socorro/lib/raven_client.py
+++ b/socorro/lib/raven_client.py
@@ -2,8 +2,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from pkg_resources import resource_string
+import sys
 
+from pkg_resources import resource_string
 import raven
 
 # When socorro is installed (python setup.py install), it will create
@@ -20,3 +21,43 @@ def get_client(dsn, **kwargs):
     if not kwargs.get('release') and SOCORRO_REVISION:
         kwargs['release'] = SOCORRO_REVISION
     return raven.Client(**kwargs)
+
+
+def capture_error(sentry_dsn, logger, exc_info=None, extra=None):
+    """Capture an error in sentry if able
+
+    :arg sentry_dsn: the sentry dsn (or None)
+    :arg logger: the logger to use
+    :arg exc_info: the exception information as a tuple like from `sys.exc_info`
+    :arg extra: any extra information to send along as a dict
+
+    """
+    exc_info = exc_info or sys.exc_info()
+
+    if sentry_dsn:
+        extra = extra or {}
+
+        try:
+            # Set up the Sentry client.
+            client = get_client(sentry_dsn)
+            client.context.activate()
+            client.context.merge({'extra': extra})
+
+            # Try to send the exception.
+            try:
+                identifier = client.captureException(exc_info)
+                logger.info('Error captured in Sentry! Reference: {}'.format(identifier))
+
+                # At this point, if everything is good, the exceptions were
+                # successfully sent to sentry and we can return.
+                return
+            finally:
+                client.context.clear()
+        except Exception:
+            # Log the exception from trying to send the error to Sentry.
+            logger.error('Unable to report error with Raven', exc_info=True)
+
+    # Sentry isn't configured or it's busted, so log the error we got that we
+    # wanted to capture.
+    logger.warning('Sentry DSN is not configured and an exception happened')
+    logger.error('Exception occurred', exc_info=exc_info)
--- a/socorro/lib/raven_client.py
+++ b/socorro/lib/raven_client.py
@@ -46,7 +46,7 @@ def capture_error(sentry_dsn, logger, exc_info=None, extra=None):
             # Try to send the exception.
             try:
                 identifier = client.captureException(exc_info)
-                logger.info('Error captured in Sentry! Reference: {}'.format(identifier))
+                logger.info('Error captured in Sentry! Reference: %s' % identifier)
 
                 # At this point, if everything is good, the exceptions were
                 # successfully sent to sentry and we can return.
--- a/socorro/processor/mozilla_transform_rules.py
+++ b/socorro/processor/mozilla_transform_rules.py
@@ -863,10 +863,12 @@ class SignatureGeneratorRule(Rule):
         self.sentry_dsn = sentry_dsn
         self.generator = SignatureGenerator(error_handler=self._error_handler)
 
-    def _error_handler(self, raw_crash, processed_crash, extra):
+    def _error_handler(self, raw_crash, processed_crash, exc_info, extra):
         """Captures errors from signature generation"""
         extra['uuid'] = raw_crash.get('uuid', None)
-        raven_client.capture_error(self.sentry_dsn, self.config.logger, extra=extra)
+        raven_client.capture_error(
+            self.sentry_dsn, self.config.logger, exc_info=exc_info, extra=extra
+        )
 
     def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
         # Generate a crash signature and capture the signature and notes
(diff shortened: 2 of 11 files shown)


socorro/unittest/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+    def __eq__(self, other):
+        return True
+
+    def __lt__(self, other):
+        return True
+
+
+#: Sentinel that is equal to anything; simplifies assertions in cases where
+#: part of the value changes from test to test
+WHATEVER = EqualAnything()
```
willkg: I wrote this for Fjord ages ago. It's super handy in tests where parts change test-to-test. Then you can do things like "assert this tuple is like this except for this index and this index which I don't care about" succinctly.

socorro/unittest/processor/__init__.py
```diff
@@ -21,3 +24,16 @@ def create_basic_fake_processor():
...
+    config = CDotDict()
+    config.logger = Mock()
+    return config
+
+
+def get_basic_processor_meta():
+    processor_meta = DotDict()
+    processor_meta.processor_notes = []
+    processor_meta.quit_check = lambda: False
+    return processor_meta
```
willkg: I extracted these from the processor tests where they were written in like 20 times. Seemed silly. I wanted to turn them into pytest fixtures, but since we're still using unittest scaffolding for those tests, that wasn't possible.

socorro/unittest/processor/test_mozilla_transform_rules.py
```diff
@@ -7,33 +7,37 @@
 from StringIO import StringIO
 
 from configman.dotdict import DotDict as CDotDict
-from mock import Mock, patch
+from mock import call, Mock, patch
 
 from socorro.lib.datetimeutil import datetime_from_isodate_string
 from socorro.lib.util import DotDict
 from socorro.processor.mozilla_transform_rules import (
-    ProductRule,
```
willkg: This list was sorted and then I added `SignatureGeneratorRule`.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -857,10 +858,14 @@ def __init__(self, config):
...
             sentry_dsn = self.config.sentry.dsn
         except KeyError:
-            # DotDict raises a KeyError when things are missing
+            # NOTE(willkg): DotDict raises a KeyError when things are missing
             sentry_dsn = None
+        self.sentry_dsn = sentry_dsn
+        self.generator = SignatureGenerator(error_handler=self._error_handler)
 
-        self.generator = SignatureGenerator(sentry_dsn=sentry_dsn)
+    def _error_handler(self, extra):
```
willkg: I need to fix this--it needs to take the data it passed so it can add the crash id.

socorro/unittest/processor/test_mozilla_transform_rules.py
```diff
@@ -7,33 +7,37 @@
...
+    SignatureGeneratorRule,
     ThemePrettyNameRule,
+    TopMostFilesRule,
+    UserDataRule,
+    Winsock_LSPRule,
 )
+from socorro.signature.generator import SignatureGenerator
+from socorro.unittest import WHATEVER
 from socorro.unittest.testbase import TestCase
+from socorro.unittest.processor import get_basic_config, get_basic_processor_meta
```
AlexisDeschamps: I think the processor import might be supposed to go before `testbase`

socorro/unittest/processor/test_mozilla_transform_rules.py
```diff
@@ -1832,3 +1594,70 @@ def test_with_malformed_addons_field(self):
...
+
+
+class TestSignatureGeneratorRule:
+    def test_empty_raw_and_processed_crashes(self):
+        rule = SignatureGeneratorRule(get_basic_config())
+        raw_crash = {}
+        processed_crash = {}
+        processor_meta = get_basic_processor_meta()
+
+        ret = rule._action(raw_crash, {}, processed_crash, processor_meta)
```
AlexisDeschamps: What does `ret` mean as a variable name?

socorro/processor/mozilla_transform_rules.py
```diff
@@ -857,10 +858,17 @@ def __init__(self, config):
...
+            # NOTE(willkg): DotDict raises a KeyError when things are missing
             sentry_dsn = None
-
-        self.generator = SignatureGenerator(sentry_dsn=sentry_dsn)
+        self.sentry_dsn = sentry_dsn
+        self.generator = SignatureGenerator(error_handler=self._error_handler)
+
+    def _error_handler(self, raw_crash, processed_crash, exc_info, extra):
+        """Captures errors from signature generation"""
+        extra['uuid'] = raw_crash.get('uuid', None)
```
AlexisDeschamps: Does `.get` not default to `None`? Otherwise, is that to be more explicit?

socorro/lib/raven_client.py
```diff
@@ -20,3 +21,43 @@ def get_client(dsn, **kwargs):
...
+        try:
+            # Set up the Sentry client.
+            client = get_client(sentry_dsn)
+            client.context.activate()
+            client.context.merge({'extra': extra})
+
+            # Try to send the exception.
+            try:
+                identifier = client.captureException(exc_info)
+                logger.info('Error captured in Sentry! Reference: {}'.format(identifier))
```
AlexisDeschamps: Do you want to switch this line to use the other formatting method?

socorro/processor/processor_app.py
```diff
@@ -282,8 +263,15 @@ def _transform(self, crash_id):
...
-            self._capture_error(crash_id, exc_type, exc_value, exc_tb)
+            # PolyStorage can throw a PolyStorageException which is a sequence
+            # of exc_info items.
+            if isinstance(exc_value, collections.Sequence):
+                exc_info = exc_value
+            else:
+                exc_info = [(exc_type, exc_value, exc_tb)]
+
+            for exc_info_item in exc_info:
+                self._capture_error(crash_id, exc_info_item)
```
AlexisDeschamps: So we're doing `_capture_error` with both `sys.exc_info()` and all of its items individually?

socorro/unittest/processor/test_mozilla_transform_rules.py
willkg: "return" is a reserved word, so we can't use that. So I tend to abbreviate it to "ret". It's not a great name. I'd be game for alternatives if you have one.

socorro/unittest/processor/test_mozilla_transform_rules.py
willkg: It should! I'll fix that.

socorro/processor/mozilla_transform_rules.py
willkg: It depends on what `raw_crash` is. If it's a dict, then yes--it defaults to giving you `None`. If it's not, then who knows. In this case, I'm pretty sure it can be one of a variety of DotDict implementations.

socorro/processor/processor_app.py
willkg: The short answer is "yes".

The long answer is that this error handling is super muddled because (I claim) the interface here isn't well defined. Running `_transform()` could raise anything. In some cases, that exception could implement the sequence protocol and have a bunch of internal exceptions. The original code captured all of those and this version does, too.

Should it do that? I don't think so. I think we shouldn't have a notion of an exception that's a sequence of exceptions--that's weird. `PolyCrashStorage` should be handling the errors as they happen. That's a bigger change, though, and I'd need to skulk around the crashstorage code to figure out whether that's ok or not. Plus it'd probably involve rewriting a lot of tests.

willkg:
Thank you!

socorro/unittest/processor/test_mozilla_transform_rules.py
AlexisDeschamps: I see. I think, in the past, I have used `result`.

socorro/processor/processor_app.py
AlexisDeschamps: I see, thanks.

PR #4529, opened 2018-07-31
bug 1477782: adjust output of auditgroups

willkg:
This makes the output clearer in regards to inactivity amounts and also
adds a total at the end.

### DIFF
--- a/webapp-django/crashstats/authentication/management/commands/auditgroups.py
+++ b/webapp-django/crashstats/authentication/management/commands/auditgroups.py
@@ -68,7 +68,7 @@ class Command(BaseCommand):
                     )
 
         # Log or remove the users that have been marked
-        for user, reason in sorted(users_to_remove, key=lambda item: (item[1], item[0])):
+        for user, reason in users_to_remove:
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/authentication/management/commands/auditgroups.py
```diff
@@ -46,26 +51,30 @@ def audit_hackers_group(self, dryrun):
...
-                            user.email, len(active_tokens)
+                        'SKIP: %s (inactive %sd, but has active tokens: %s)' % (
+                            user.email, days, len(active_tokens)
                         )
                     )
 
         # Log or remove the users that have been marked
-        for user, reason in users_to_remove:
-            self.stdout.write('Removing: {} ({})'.format(user.email, reason))
+        for user, reason in sorted(users_to_remove, key=lambda item: (item[1], item[0])):
```
willkg: I just realized this isn't going to work because the number of days is variable. I'll nix it.

willkg:
I don't like `.format()` for a few reasons and I'd rather just standardize on the one I use the most.

willkg:
Thank you for doing this!

PR #4530, opened 2018-07-31
bug 1477782: change dryrun to persist

willkg:
This changes the dryrun argument for the auditgroups command to a "persist"
argument. It also fixes the argument value parsing which was busted before.


willkg:
Thank you!

PR #4531, opened 2018-07-31
fix bug 1479485: added report details api

AlexisDeschamps:
fixes bug https://bugzilla.mozilla.org/show_bug.cgi?id=1479485

## why

## what
- added a API endpoint to get the data needed for the details tab
of report indexes in JSON

## example
screen shot 2018-07-30 at 9 46 40 am


webapp-django/crashstats/api/views.py
```diff
@@ -529,3 +531,67 @@ def crash_verify(request):
...
+
+    all_fields = supersearch_models.SuperSearchFields().get()
+    descriptions = {}
+    for field in all_fields.values():
+        key = '{}.{}'.format(field['namespace'], field['in_database_name'])
+        descriptions[key] = '{} Search: {}'.format(
+            field.get('description', '').strip() or
+            'No description for this field.',
+            field['is_exposed'] and field['name'] or 'N/A',
+        )
```
willkg: If I read this correctly, the description keys will all be snake_case and namespaces, but the report data is going to be camelCase and possibly different namespaces. Does that sound right? If not, what am I missing?

Also, I don't see this `descriptions` variable passed anywhere. Is it supposed to get sent along with the data or something else?

webapp-django/crashstats/api/views.py
```diff
@@ -529,3 +531,67 @@ def crash_verify(request):
...
+
+    data = transform_report_details(
+        report=report,
+        raw_crash=raw_crash,
+        crashing_thread=crashing_thread,
+        parsed_dump=parsed_dump
+    )
+    return http.HttpResponse(json.dumps(data), content_type='application/json')
```
willkg: I looked at `json_view` because I wondered what it did especially in light of you using `HttpResponse` here and `JsonResponse` elsewhere. Seems like we could change this line to just return `data` and the `json_view` will handle it and also do pretty printing.

However, I don't think we can rely on `json_view` to handle the error situations. Those should get done like you're doing now. Might make sense to fix `json_view` to handle all the situations correctly.

Anyhow, I'm just thinking out loud. You could leave things as is.

AlexisDeschamps:
@willkg I added two commits if you want to take another look.

willkg:
We're out of time for the internship and I've got a bunch of stuff on my plate and I can't get to a real review of this any time soon.

This is a cupcake, though. There are a lot of pieces to this cupcake and until we have them all mixed together, it's hard to see the cupcake's form.

Given that, I'm going to merge this now. We can evaluate the cupcake when it's fully formed and someone has time to do it.

Thank you so much for the work here! The things I did look at look good!

PR #4533, opened 2018-08-01
Fix Token admin

willkg:
* makes tokens searchable via email address and notes
* fixes how tokens are represented--no longer do they show up as "Token"
  which is unhelpful
* makes them filterable by permissions (this could use work, but it's
  better than before)


willkg:
This affects a Django admin page that only admin can see and is rarely used. I can verify this on -stage. There's no risk here. self-r+

PR #4536, opened 2018-08-02
fix bug 1463756: make signature generator socorro-agnostic

willkg:
This reworks the socorro signature generation code to be ignorant of how socorro stores crash data (i.e. raw crash and processed crash).

This has three major advantages:

1. it's cleaner and more functional--we're not editing things in place
2. it's defined--we now have a schema for what signature generation is working on
3. it can be extracted as an external library for others to use


willkg:
I had forked signature generation a while ago and this unifies the two implementations. I made some changes here and made some changes to siggen to make it easier to unify the two. Then I took the siggen code and stomped on the Socorro code.

siggen is currently being used to generate signatures from crash ping data in telemetry. It's still "alpha-level" in the sense that I hadn't set the API and schema in stone, but it's working fine.

I compared signatures from these changes with what's in prod for the last week and there are no differences.

I processed a bunch of crashes.

Test coverage is decent.

Given all that, I'm going to read through all the code changes and then if that doesn't turn up anything, I'll land it.

setup.cfg
```diff
@@ -24,7 +24,7 @@ max-line-length = 100
 # -p no:django - disable the pytest-django plugin for Socorro tests
 addopts = -rsxX --tb=native -p no:django
 norecursedirs = .git docs config docker __pycache__
-testpaths = socorro/unittest/
+testpaths = socorro/
```
willkg: We now have tests in other places, so opening this up a bit allows those tests to run.

socorro-cmd
```diff
@@ -93,7 +93,7 @@ COMMANDS = [
     Group(
         'Miscellaneous', {
             'showcommands': showcommands_cmd,
-            'signature': 'socorro.signature.__main__.main'
+            'signature': 'socorro.signature.cmd_signature.main'
```
willkg: I moved the command code so it jives with extracting to an external library better. This does break the whole:

```
python -m socorro.signature ...
```

thing, but that's ok.

socorro/lib/util.py
```diff
@@ -33,28 +32,6 @@ def chunkify(iterable, n):
...
-        # Convert it back to a acsii string and drop any non-ascii characters
-        return text.encode('ascii', 'ignore')
-
-    # Python 3 ONLY, strip non ascii characters
-    ALLOWED_CHARS = [chr(c) for c in range(32, 127)]
-    text = ''.join([c for c in text if c in ALLOWED_CHARS])
-
-    return text
-
-
```
willkg: Moved, renamed, and rewrote this.

socorro/signature/generator.py
```diff
@@ -39,52 +38,58 @@
...
+                    old_sig = result['signature']
+                    rule.action(signature_data, result)
+
                     if self.debug:
-                        notes.append('%s: %s -> %s' % (
-                            rule.__class__.__name__, sig, processed_crash['signature']
-                        ))
+                        result['notes'].append(
+                            '%s: %s -> %s' % (
+                                rule.__class__.__name__,
```
willkg: We should pull this out into a variable since it's used a bunch of times.

socorro/signature/rules.py
```diff
@@ -418,7 +423,7 @@ def __init__(self):
     def _create_frame_list(self, crashing_thread_mapping, make_modules_lower_case=False):
         frame_signatures_list = []
         for a_frame in islice(
-            crashing_thread_mapping.get('frames', {}),
+            crashing_thread_mapping.get('frames', []),
```
willkg: Frames is a list of things--it should default to an empty list and not an empty dict.

socorro/signature/siglists_utils.py
```diff
@@ -43,6 +44,8 @@ def _get_file_content(source):
     with resource_stream(__name__, filepath) as f:
         for i, line in enumerate(f):
             line = line.strip()
+            if six.PY3:
+                line = line.decode('utf-8', 'strict')
```
willkg: We can use `strict` here because if the signature lists are busted, then that's a bug and we should fix them. If there are problems, they'll show up in the tests in CI.

socorro/signature/tests/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+    def __eq__(self, other):
+        return True
+
+    def __lt__(self, other):
+        return True
+
+
+#: Sentinel that is equal to anything; simplifies assertions in cases where
+#: part of the value changes from test to test
+WHATEVER = EqualAnything()
```
willkg: I copied this from elsewhere in Socorro code. It's important for the signature generation code to be self-contained and extractable. Hence the duplication.

socorro/signature/tests/test_rules.py
```diff
@@ -0,0 +1,1775 @@
...
+import pytest
+
+# NOTE(willkg): We do this so that we can extract signature generation into its
+# own namespace as an external library. This allows the tests to run if it's in
+# "siggen" or "socorro.signature".
+base_module = '.'.join(__name__.split('.')[:-2])
+rules = importlib.import_module(base_module + '.rules')
+
+
+class TestCSignatureTool:
```
willkg: These tests should be further cleaned up, but that's a job for a different day. There's a lot of "stuff" here and they could be reduced and still have the same coverage.

socorro/signature/tests/test_utils.py
```diff
@@ -0,0 +1,26 @@
...
+    ('123', '123'),
+    (u'123', '123'),
+
+    # Drop non-ascii characters
+    ('1\xc6\x8a23', '123'),
+    (u'1\u018a23', '123'),
+
+    # Drop non-printable characters
+    ('\r\n\t\b1 23', '1 23'),
+    (u'\r\n\t\b1 23', '1 23'),
```
willkg: Technically `\r\n\t` are all printable characters. `\b` is not.

socorro/signature/utils.py
```diff
@@ -0,0 +1,111 @@
...
+import six
+
+
+def int_or_none(data):
+    if data:
+        try:
+            return int(data)
+        except ValueError:
+            return None
+    return data
```
willkg: This should return `int` or `None` and nothing else. This should get fixed.

willkg:
^^^ That fixes the issues I found. Will wait for CI and then do a round of extract-to-siggen to see how that works.

willkg:
^^^ That adds code to update socorro-siggen and fixes some additional minor things. That's everything I wanted to do. This is good to land now.

willkg:
Self-r+. The code has been adequately vetted. We have the weekend to discover any issues, but at this point those will be new bugs.

PR #4537, opened 2018-08-03
fix bug 1473068: fix session refresh errors with xhr

willkg:
The search page does XHR requests when switching between searches. Other
pages do similar things. When the SessionRefresh middleware handles those
requests, it can return a 403 or a redirect, but the js can't handle that.

This fixes the SessionRefresh middleware to ignore those urls and not
refresh the session on them.


willkg:
I tested this out locally.

1. process some crashes
2. run the webapp
3. do a super search
4. wait 15+ minutes
5. click on the search button again which just does an xhr request

Without these changes, that xhr with fail with an HTTP 403. With the changes, the page works fine.

This fixes the super search page and the signature report page. It's possible there are other pages that are having problems. We can deal with those in new bugs.

willkg:
I verified this works locally. It's pretty straight-forward and relies on code I wrote for mozilla-django-oidc. I feel pretty comfortable with this.

self-r+.

PR #4538, opened 2018-08-03
fix bug 1480848: upgrade cryptography to 2.3

willkg:


willkg:
This is a minor bump with no API changes. self-r+.

PR #4540, opened 2018-08-06
fix bug 1473068: fix session refresh errors with xhr (buginfo)

willkg:
This adds another url used by the supersearch page to the exempt urls
list.


willkg:
self-r+

PR #4542, opened 2018-08-06
fix bug 1478055: Add product documentation

willkg:
This is based on my research for Socorro.

It's possible parts of this are wrong. It's very likely we'll want to change some of this as we add new products. It's a living document that we can adjust as we go along.


willkg:
self-r+

I think there's some other things we want to add, but this is good enough for now.

PR #4544, opened 2018-08-07
fix bug 1481282: handle c/cpp and rust frames separately

willkg:
This adjusts signature generation such that it normalizes frames
from Rust sources differently than frames from other sources (usually
C/C++).

This adds a bunch of additional tests from real data to make sure
things don't regress including a few comments about things that should
get fixed in the future.

For the most part, this doesn't change anything--it's just a refactoring
to enable future changes.

### DIFF
--- a/socorro/signature/rules.py
+++ b/socorro/signature/rules.py
@@ -73,9 +73,11 @@ class SignatureTool(object):
 
 
 class CSignatureTool(SignatureTool):
-    """This is the class for signature generation tools that work on breakpad C/C++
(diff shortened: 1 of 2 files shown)


willkg:
I ran this against a couple of thousand crash reports to see if it generates the exact same signatures. It does.

Generally, the changes are predominantly cosmetic. I did spend some time splitting up the normalization code so Rust-specific things were in `normalize_rust_frame` and C/C++-specific things were in `normalize_cpp_frame`. We can hone that more over time.

I think this looks good. self-r+

PR #4545, opened 2018-08-08
fix bug 1481828: fix proto_signature in SignatureGeneratorRule

willkg:
I recently refactored signature generation using the code from the
socorro-siggen fork. One thing I missed was making sure `proto_signature`
made it into the processed crash.

This fixes that bug, adds a test to make sure everything that needs
to be in the processed crash makes it there, and fixes the spelling
of "canonical".


willkg:
self-r+

PR #4546, opened 2018-08-08
bug 1476941: fix ProductRewrite rule to add note

willkg:
The `ProductRewrite` rule stomps on the incoming raw crash `ProductName`
with a new one in some cases. That sucks because it means we have no record
of what the original name was.

This adds a processor note so we at least have some record of what happened.

### DIFF
--- a/socorro/processor/mozilla_transform_rules.py
+++ b/socorro/processor/mozilla_transform_rules.py
@@ -363,10 +363,14 @@ class ProductRewrite(Rule):
         return '2.0'
 
     def _predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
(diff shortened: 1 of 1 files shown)


socorro/processor/mozilla_transform_rules.py
```diff
@@ -363,10 +363,14 @@ def version(self):
         return '2.0'
 
     def _predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
-            return raw_crash.get('ProductID') in self.PRODUCT_MAP
+        return raw_crash.get('ProductID') in self.PRODUCT_MAP
```
willkg: Cosmetic. This was indented wrong.

PR #4547, opened 2018-08-08
Fix docs on ProductName origins

willkg:


willkg:
self-r+. This fixes docs and has no other effect.

PR #4548, opened 2018-08-08
fix bug 1481302: fix API test drive button

willkg:
This adds the js that the page needs that got removed in the docs overhaul.


willkg:
This works locally and is easy to verify on -stage. If it's not right, it'll be very apparent but shouldn't make the page worse.

self-r+

PR #4549, opened 2018-08-08
fix bug 1480985: fix highlight_url view to validate schema

willkg:


willkg:
This is a benign change and low risk. self-r+

PR #4550, opened 2018-08-09
fix bug 1481282: correctly normalize frames based on source file

willkg:
Previously, the code assumed the source file was something like `"foo.rs"`.
That was horrifically wrong. This fixes that.


willkg:
This looks good. self-r+

PR #4551, opened 2018-08-10
fix bug 1482477: add product admin page

willkg:
This adds a Products model and a Django admin page to view it.


willkg:
This only affects the Django admin. It's low risk and mostly boilerplate. self-r+

PR #4552, opened 2018-08-10
fix bug 1481343: add new product listing home page

willkg:
This cleans up product home pages a bit, gets rid of legacy redirects, stops redirecting from `/` to `/home/product/Firefox`, and creates a new `/` that lists all active products.

The active products list is driven by the "products" table.

First problem is that it lists Fennec and FennecAndroid and it should only list one. The header drop down lists both, too, so I think I'm going to leave it for now and we can fix it later if it's a problem.

The second problem is that the list of products will be out of sync with the dropdown in the header. The dropdown in the header is driven by the "product_versions" table and only includes products that have active versions. This is fine now, but I suspect we're going to add products that don't have all the build information and active version information available. I think I'm going to leave this for now and we can fix it later if it's a problem.

The third problem is that my CSS skills are mediocre at best. I wrote some CSS based on what SUMO does with its product cards. It's not responsive. It's kind of goofy. It can definitely be minimized and made better by probably anyone on Earth. Having said that, I'm going to leave this as is for now and if someone jumps in with review comments, then yay! Otherwise we can improve it in a future PR.


webapp-django/crashstats/crashstats/jinja2/crashstats/home.html
```diff
@@ -0,0 +1,22 @@
...
+    <div class="panel">
+        <ul id="product-cards" class="card-grid">
+            {% for product in products %}
+            <li>
+                <a class="cf" href="{{ url('crashstats:product_home', product=product.product_name) }}">
+                    <span class="productname">{{ product.product_name }}</span>
+                </a>
+            </li>
+            {% endfor %}
+        </ul>
```
willkg: This is the HTML for the new product listing on the new home page.

webapp-django/crashstats/crashstats/static/crashstats/css/base/layout.less
```diff
@@ -117,3 +117,70 @@ body {
...
+            content: "";
+            // Fixes Opera/contenteditable bug:
+            // http://nicolasgallagher.com/micro-clearfix-hack/#comment-36952
+            line-height: 0;
+        }
+        &:after {
+            clear: both;
+        }
+    }
+}
```
willkg: This is the CSS I copied and pasted from SUMO and then edited down.

CSS is not my forte. If you have opinions on what to do here, please feel free to fix it.

@Osmose , @lonnen : Thoughts? ^^^

willkg:
screenshot_2018-08-10 crash data for firefox

willkg:
The home page needs a test. I'll do that on monday.

willkg:
Sorry about that. To test this, do the following:

1. get the branch
2. run `make build` (if you haven't done it this week)
3. run `make setup` to build a db and populate with basic data
4. run `make run` and go to `http://localhost:8000/`

willkg:
@Osmose I switched to your suggestions with some minor tweaks that I think make the blocks look a bit better. Thank you for this!

willkg:
Thank you!

PR #4554, opened 2018-08-10
fix bug 1482505: added React crash Details view 

AlexisDeschamps:
fixes bug https://bugzilla.mozilla.org/show_bug.cgi?id=1482505

## why

## what
- re-created the crash report Details tab using React
- added Enzyme unit tests for most of the new components

## image
image


Osmose:
Nice work!

PR #4556, opened 2018-08-14
fix bug 1483097: fix product home

willkg:
This fixes the product home (and the nav header) to work with products that may not have information in the `product_versions` table.

To test this, do:

1. `make setup`
2. `make run`
3. open up your browser and go to `http://localhost:8000`

You'll see a bunch of products which have no information in `product_versions`. The product home page won't show any Top Crasher links. If you process some crashes thus adding crash report data, there will then be Top Crasher links.

Hooray!


webapp-django/crashstats/crashstats/jinja2/crashstats/product_home.html
```diff
@@ -22,48 +22,54 @@ <h2>Crash Reports</h2>
...
+                                ) }}">
+                                    Top GPU Crashers
+                                </a>
+                            </li>
+                        </ul>
+                    </div>
+                    {% endfor %}
+                {% else %}
+                    <div class="release_channel">
+                        No version information available.
```
willkg: If we have no version information at all, we might as well display an informative message.

Part of me was tempted to include a knock-knock joke. Something like:

Knock knock.
Who's there?
No.
No who?
NO VERSION INFORMATION AVAILABLE!

I figured as much as I think that's amusing, I should stick with this message and write the joke in the PR comments.

PR #4557, opened 2018-08-15
fix bug 1483097: fix version sorting in product home page

willkg:
This fixes version sorting in the product home page so the right "featured" versions show up.

### DIFF
--- a/webapp-django/crashstats/crashstats/tests/test_utils.py
+++ b/webapp-django/crashstats/crashstats/tests/test_utils.py
@@ -409,3 +409,17 @@ def test_json_view_custom_status():
 ])
 def test_parse_version(version_string, expected):
     assert utils.parse_version(version_string) == expected
+
+
+def test_parse_version_sorted():
+    """Test that using parse_version produces version keys that sort correctly"""
+    versions = ['59.0', '59.0b2', '59.0.2', '59.0a1', '60p', '59.0.2esr']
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/crashstats/tests/test_utils.py
```diff
@@ -1,385 +1,411 @@
...
+    ('59.0.1b1', (59, 0, 1, 'b1')),
+
+    # esr
+    ('59.0.1esr', (59, 0, 1, 'zz', 'esr')),
+
+    # junk data
+    ('49p', (-1)),
+])
+def test_parse_version(version_string, expected):
+    assert utils.parse_version(version_string) == expected
```
willkg: I should add a test using `parse_version` as a key for sorting to make sure it sorts correctly now and to the end of time.

willkg:
The risks of this code being wrong are the following:

**Versions with structures that aren't accounted for in `parse_version` will either show up in the versions list when they're not supposed to or not show up when they are supposed to.** Someone will notice this if it's egregious and we can fix it. The site will continue to work fine.

**Exceptions might get thrown for product home pages that have versions that have a structure that aren't handled.** If this happens, pages related to that product won't work. We'll need to throw together a fix and push it out. Other products should work fine. I spent some time hardening the code and making sure it worked even with junk data, but it's possible I missed something.

I'm going to test this out on stage. self-r+.

PR #4558, opened 2018-08-15
fix bug 1483318: don't show inactive products

willkg:
This fixes the code so that it doesn't show inactive products in the listing. It also goes through and updates the "sort" column of the "products" table so that all products except specified ones are now inactive.

### DIFF
--- a/socorro/external/postgresql/staticdata.py
+++ b/socorro/external/postgresql/staticdata.py
@@ -84,11 +84,12 @@ class Products(BaseTable):
         'product_name', 'sort', 'rapid_release_version', 'release_name', 'rapid_beta_version'
     ]
     rows = [
-        ['Fennec', '3', '5.0', 'mobile', '999.0'],
-        ['Thunderbird', '2', '6.0', 'thunderbird', '999.0'],
         ['Firefox', '1', '5.0', 'firefox', '23.0'],
-        ['SeaMonkey', '6', '2.3', 'seamonkey', '999.0'],
-        ['FennecAndroid', '4', '5.0', '**SPECIAL**', '999.0'],
+        ['FennecAndroid', '2', '5.0', '**SPECIAL**', '999.0'],
+        ['Thunderbird', '80', '6.0', 'thunderbird', '999.0'],
+        ['SeaMonkey', '90', '2.3', 'seamonkey', '999.0'],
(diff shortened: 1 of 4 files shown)


socorro/external/postgresql/staticdata.py
```diff
@@ -84,11 +84,12 @@ class Products(BaseTable):
...
-        ['Fennec', '3', '5.0', 'mobile', '999.0'],
-        ['Thunderbird', '2', '6.0', 'thunderbird', '999.0'],
         ['Firefox', '1', '5.0', 'firefox', '23.0'],
-        ['SeaMonkey', '6', '2.3', 'seamonkey', '999.0'],
-        ['FennecAndroid', '4', '5.0', '**SPECIAL**', '999.0'],
+        ['FennecAndroid', '2', '5.0', '**SPECIAL**', '999.0'],
+        ['Thunderbird', '80', '6.0', 'thunderbird', '999.0'],
+        ['SeaMonkey', '90', '2.3', 'seamonkey', '999.0'],
+        ['Fennec', '100', '5.0', 'mobile', '999.0'],
+        ['InactiveProduct', '-1', '5.0', 'inactive', '999.0'],
```
willkg: Adding an inactive product here that should never show up helps us see if we broke something.

webapp-django/crashstats/crashstats/utils.py
```diff
@@ -286,13 +286,20 @@ def build_default_context(product=None, versions=None):
...
     # Build product information
     api = models.Products()
-    context['products'] = {
-        hit['product_name']: hit for hit in api.get()['hits']
-    }
+    # NOTE(willkg): using an OrderedDict here since Products returns products
+    # sorted by sort order and we don't want to lose that ordering
+    all_products = OrderedDict()
+    for item in api.get()['hits']:
+        if item['sort'] == -1:
```
willkg: I thought about adding an `is_active` filter to the model, but in the (hopefully near) future, we want to move those sqlalchemy models to Django models where they will be a lot easier to do things with. Seemed like it didn't make sense to figure out how to do filters now and implement it that way.

willkg:
The changes here are pretty benign. I spent some time testing them locally and everything seems fine. I'm going to self-r+ this and verify the changes on stage.

PR #4559, opened 2018-08-15
fix bug 1483564: comment out Python 3 tests

willkg:
The Python 3 tests step takes 4 minutes to run out of 18 minutes for an
entire CI pass. That's signficant.

We're putting Python 3 work on the back burner for now becuase we've got
other things to work on. It'll save us a lot of time to stop running
the Python 3 tests in CI for now. We'll re-enable them when Python 3
work is on the front burner again.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,10 +57,10 @@ jobs:
           working_directory: /socorro
           command: make test
 
-      - run:
-          name: Run tests in Python 3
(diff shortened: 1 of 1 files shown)


willkg:
I read through the Circle CI logs and verified it's skipping the Python 3 tests now. This one took 12m. Previous passing iterations took between 16m and 18m.

Let's go with this. self-r+

PR #4560, opened 2018-08-15
bug 1481696: add focus

willkg:
This adds Focus as a supported product.

This also fixes a problem in local development environments where if you
haven't set up ES, it causes the product home page to throw HTTP 500s.


willkg:
self-r+

PR #4561, opened 2018-08-15
fix bug 1477013: fix signature generation for rust trait methods

willkg:
This reimplements collapse, adds documentation, adds tests, and fixes the case where it shouldn't collapse a rust trait method.

### DIFF
--- a/socorro/signature/rules.py
+++ b/socorro/signature/rules.py
@@ -9,7 +9,7 @@ from glom import glom
 import ujson
 
 from . import siglists_utils
-from .utils import drop_bad_characters, parse_source_file
+from .utils import collapse, drop_bad_characters, parse_source_file
 
 
 SIGNATURE_MAX_LENGTH = 255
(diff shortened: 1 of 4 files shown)


willkg:
I pulled down 15,000 frames from crash reports in production and ran this version of collapse and the old version of collapse on them. There are two differences:

1. This version handles the Rust trait method situation correctly. So:
   ```
   <rayon_core::job::HeapJob<BODY> as rayon_core::job::Job>::execute
   ```
   no longer gets converted to:
   ```
   <T>::execute
   ```
   which is unhelpful.
2. This fixes a bug in handling context exceptions where it could drop rest of the string. So this:
   ```
   IPC::ParamTraits<nsTSubstring<char> >::Write(IPC::Message *,nsTSubstring<char> const &)
   ```
   used to be converted to:
   ```
   IPC::ParamTraits<nsTSubstring<T>::Write
   ```
   and is now more correctly converted to:
   ```
   IPC::ParamTraits<nsTSubstring<T> >::Write
                                   ^^
   ```

I also ran signature generation regressions against 1000 crashes from today:

```
app@processor:/app$ socorro-cmd fetch_crashids --num=1000 --date=today | socorro-cmd signature --different-only
Crash id: 3b05a0ad-9089-4008-a64a-8a4920180815
Original: OOM | large | NS_ABORT_OOM | IPC::ParamTraits<nsTSubstring<T>::Read
New:      OOM | large | NS_ABORT_OOM | IPC::ParamTraits<nsTSubstring<T> >::Read
Same?:    False
```

Only one difference.

socorro/signature/rules.py
```diff
@@ -109,89 +109,22 @@ def __init__(self, quit_check_callback=None):
...
-        function = self._collapse(
+        function = collapse(
             function,
             open_string='<',
-            replacement_open_string='<',
             close_string='>',
-            replacement_close_string='T>',
-            exception_substring_list=('name omitted', 'IPC::ParamTraits')
+            replacement='<T>',
+            exceptions=('name omitted', 'IPC::ParamTraits', ' as ')
```
willkg: I should check if "IPC::" ever shows up in Rust frames. I'm pretty sure it doesn't. If so, then we could nix that. Plus we could probably nix the `name omitted` exception, too.

socorro/signature/rules.py
```diff
@@ -109,89 +109,22 @@ def __init__(self, quit_check_callback=None):
...
-            function = self._collapse(
+            function = collapse(
                 function,
                 open_string='(',
-                replacement_open_string='',
                 close_string=')',
-                replacement_close_string='',
-                exception_substring_list=('anonymous namespace', 'operator')
+                replacement='',
+                exceptions=('anonymous namespace', 'operator')
```
willkg: I don't think either of these exceptions occur in Rust frames.

willkg:
This looks pretty good. Our ability to regression test signature generation changes is decent, so I claim this is low risk. self-r+

PR #4563, opened 2018-08-16
bug 1481696: add processor rule to fix Focus crashes

willkg:
This reworks the ProductRewrite rule so it logs what it did in the crash
report better. It also adjusts the rule to act as a "this is where we're
going to do all the product rewriting" place.

This adds product rewriting logic for Focus GV. That's using GeckoView,
but incoming crash reports are using `ProductName=Fennec`. So the ProductRewrite
rule was convering those to FennecAndroid. This will then convert it to
Focus if the `ProcessType=content`.

We should remove this as soon as we can, but it'll fix an existing issue
we have right now.

### DIFF
--- a/socorro/processor/mozilla_transform_rules.py
+++ b/socorro/processor/mozilla_transform_rules.py
@@ -351,8 +351,10 @@ class OutOfMemoryBinaryRule(Rule):
 
 
 class ProductRewrite(Rule):
-    """This rule rewrites the product name for products that fail to report
-    a useful product name
+    """Fix ProductName in raw crash for certain situations
+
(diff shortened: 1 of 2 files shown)


willkg:
To test, process a Focus crash report like this one:

https://crash-stats.mozilla.com/report/index/3a684db6-dbca-442c-ab42-3a22f0180805

It should show up as "Focus" now and not "FennecAndroid".

willkg:
Changes look good. I'm going to test this on stage by reprocessing all the FennecAndroid crashes for the last week. If there are any issues, they'll come out in that.

self-r+

PR #4564, opened 2018-08-16
fix bug 1483100: fix topcrashers report for build-less products

willkg:


webapp-django/crashstats/crashstats/decorators.py
```diff
@@ -79,8 +79,8 @@ def pass_default_context(view):
     """
     @functools.wraps(view)
     def inner(request, *args, **kwargs):
-        product = kwargs.get('product', None)
-        versions = kwargs.get('versions', None)
+        product = kwargs.get('product', request.GET.get('product', None))
+        versions = kwargs.get('versions', request.GET.get('versions', None))
```
willkg: This was half of the issue. The problem here was that this code used to pull the product and version from the url *path*, but the topcrashers report (and exploitability report) have that information in the url *querystring*. So I fixed this to try one, then the other.

webapp-django/crashstats/crashstats/utils.py
```diff
@@ -272,6 +272,7 @@ def get_recent_versions_for_product(product):
             'product': product,
             'version': major_to_versions[major_key][0],
             'is_featured': True,
+            'has_builds': False
```
willkg: This was the other half of the issue. The  faux versions needed a `has_builds` property that topcrashers uses.

willkg:
I'll test this out on stage where we've got Focus data now. self-r+

PR #4565, opened 2018-08-17
fix bug 1484072: fix supersearch for supported products

willkg:


willkg:
This works locally. We can verify it's correct on stage. It's pretty easy to see if it's working or not.

self-r+

PR #4566, opened 2018-08-21
bug 1483719: add support for GeckoView

willkg:


willkg:
I verified the upgrade and downgrade paths work. self-r+

PR #4567, opened 2018-08-22
fix bug 1485411: fix revision information acquisition

willkg:
In the old infrastructure, we had a "socorro_revision.txt" file that
was generated when we built the socorro package. We don't do that anymore.
Now we generate a version.json file when socorro is deployed.

This updates the revision acquiring code to use the correct file.


willkg:
Well, that was painful. :(

I'll verify this on stage. When the processor comes up, it should log the version info. I can trigger an exception in the webapp and make sure it's spitting out revision data.

PR #4568, opened 2018-08-23
fix bug 1485411: fix revision in sentry in stage

willkg:
When Socorro deploys to stage, there's no git tag, so the version is "".
Since that's a value in the dict, it never gets set to "unknown" so
then it doesn't get sent to sentry. It'd be better if we sent the
commit sha in this case.


willkg:
Looks better. We'll test it on stage again. self-r+

PR #4569, opened 2018-08-23
fix bug 1485188: user comments

willkg:
For test plan, see bug comments.


webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -343,35 +343,31 @@ <h2>{{ report.product }} {{ report.version }} Crash Report [@ {{ report.signatur
...
+                                    </td>
+                                </tr>
+                                <tr title="{{ fields_desc['processed_crash.user_comments'] }}">
+                                    <th scope="row">User Comments</th>
+                                    <td>
+                                        {% if report.user_comments %}
+                                        {{ report.user_comments | linebreaks }} - {{ sensitive_warning(your_crash) }}
+                                        {% endif %}
+                                    </td>
+                                </tr>
```
willkg: Previously, two of these fields were under the "must have pii" check. Now all three are. I also fixed the indentation because it was hard to read.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -691,11 +687,7 @@ <h2>Thread {{ thread.thread }}{% if thread.thread_name %}, Name: {{ thread.threa
                                         <i>empty key</i>
                                     {% endif %}
                                 </th>
-                                {% if key == 'Comments' %}
-                                <td><pre>{{ raw[key] | scrub_pii }}</pre></td>
-                                {% else %}
                                 <td><pre>{{ raw[key] }}</pre></td>
-                                {% endif %}
```
willkg: We're not scrubbing anything anymore, so we can nix this if/else block.

willkg:
I went through the test plan in my local development environment. The code looks ok--mostly it's removing things and then fixing a few tests that broke as a result. I'm going to land this and test this more extensively on stage.

self-r+

PR #4570, opened 2018-08-23
bug 1485477: clean up unused modules

willkg:
This removes modules in `socorro.lib` that are unused.


willkg:
I'll review this tomorrow. Mostly it's dead code removal, so there isn't much interesting here.

socorro/external/es/new_crash_source.py
```diff
@@ -1,103 +0,0 @@
...
-                doc_type=es_doctype,
-                fields=['crash_id'],
-                query=query,
-            )
-            for hit in res:
-                crash_ids.append(hit['fields']['crash_id'][0])
-                if self.config.cap and len(crash_ids) >= self.config.cap:
-                    break
-
-        return iter(crash_ids)
```
willkg: The problem with these scripts is that they run in a CLI in the infra and we don't really have access to that anymore and thus these aren't helpful.

We have a couple of different scripts that do this sort of thing that are API based so you can run them locally. 

So given that, it's good to remove them.

socorro/external/rabbitmq/purge_queue_app.py
```diff
@@ -1,48 +0,0 @@
...
-                rmq_channel.queue_delete(queue=self.config.queue_name)
-            except Exception:
-                self.config.logger.debug(
-                    "Could delete queue %s",
-                    self.config.queue_name
-                )
-                raise
-
-        self.config.logger.debug("Deleted queue %s", self.config.queue_name)
-        return SUCCESS
```
willkg: I wrote a queue consumer script for pigeon that's pretty straight-forward. If we ever need to purge a queue, we can use that.

socorro/external/rabbitmq/reprocess_crashlist.py
```diff
@@ -1,100 +0,0 @@
...
-                    body=uuid,
-                    properties=pika.BasicProperties(delivery_mode=2)
-                )
-                logging.debug('submitted %s' % uuid)
-
-        self.connection.close()
-
-
-if __name__ == '__main__':
-    sys.exit(ReprocessCrashlistApp.run())
```
willkg: We've got better tooling for this now.

socorro/lib/ver_tools.py
```diff
@@ -1,126 +0,0 @@
...
-    """
-    version_list = []
-    for part_count, version_part in enumerate(version_string.split('.')):
-        try:
-            groups = _version_part_re.match(version_part).groups()
-        except Exception:
-            raise NotAVersionException(version_string)
-        version_list.extend(t(x) for x, t in zip(groups, _normalize_fn_list))
-    version_list.extend(_padding_list * (max_version_parts - part_count - 1))
-    return version_list
```
willkg: We don't use this anywhere.

willkg:
self-r+

PR #4571, opened 2018-08-25
fix bug 1346891: rewrite BetaVersionRule to use webapp API

willkg:
This:

1. creates a webapp API called VersionString that returns ... well, ... version strings
2. creates a new `ExpiringCache` that has a ttl in addition to max size because we didn't have one and I wasn't enamored with the ones I saw (though `cachetools` might be a good choice)
3. rewrites the `BetaVersionRule` to use the new cache and new VersionString API
4. implements a `session_with_retries` and rewrites everything to use that rather than using requests directly

So... there's a lot of different things in this PR all lumped together, but it's all sort of related.

To test:

1. go to `http://localhost:8000/api/` and test the VersionString API using product/version/builds
2. set `processor.version_string_api` to `http://localhost:8000/api/VersionString` in your `my.env` file, then process a crash from the beta channel with a build that's in your db and make sure it gets the right version

### DIFF
--- a/socorro/external/postgresql/version_string.py
+++ b/socorro/external/postgresql/version_string.py
@@ -1,0 +1,49 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import logging
+
+from socorro.lib import MissingArgumentError, external_common
+from socorro.external.postgresql.base import PostgreSQLBase
+
+
+logger = logging.getLogger("webapi")
+
+
+class VersionString(PostgreSQLBase):
+    def get(self, **kwargs):
+        filters = [
+            ('product', None, 'str'),
+            ('version', None, 'str'),
+            ('build_id', None, 'int'),
+        ]
+        params = external_common.parse_arguments(filters, kwargs)
+        required = ('product', 'build_id', 'version')
+        for key in required:
+            if not params.get(key):
+                raise MissingArgumentError(key)
+
+        sql = """
+            SELECT
+                pv.version_string
+            FROM product_versions pv
+                LEFT JOIN product_version_builds pvb ON
+                    (pv.product_version_id = pvb.product_version_id)
+            WHERE pv.product_name = %(product)s
+            AND pv.release_version = %(version)s
+            AND pvb.build_id = %(build_id)s
+        """
+        results = self.query(sql, params)
+
+        # The query can return multiple results, but they're the same value. So
+        # we just return the first one.
+        version_string = [
+            row['version_string'] for row in results.zipped()
+        ]
+        if version_string:
+            version_string = [version_string[0]]
+
+        return {
+            'hits': version_string
+        }
--- a/socorro/lib/requestslib.py
+++ b/socorro/lib/requestslib.py
@@ -1,0 +1,33 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import urlparse
+
+import requests
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
+
+
+def session_with_retries(url):
+    """Generates a requests session that supports retries on HTTP 429
+
+    :arg url: url to use for requests
+
+    :returns: a requests Session instance
+
+    """
+    base_url = urlparse.urlparse(url).netloc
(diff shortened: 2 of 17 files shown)


willkg:
I'll review this on Monday and go from there.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -628,53 +633,35 @@ def _get_version_data(self, product, version, build_id):
         is 54.0). This database call returns the actual version number of said
         build (i.e. 54.0b3 for the previous example).
         """
```
willkg: We should document arguments, returns, and also that this can raise `requests.RequestException`.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -606,19 +608,22 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
-            database,
-        )
-        self._versions_data_cache = {}
+        self.version_string_api = config.version_string_api
+        self.cache = ExpiringCache(max_size=self.CACHE_MAX_SIZE, ttl=self.CACHE_TTL)
 
     def version(self):
         return '1.0'
 
+    # FIXME(willkg): add ttl cache here
```
willkg: We added it--so we can nix this comment.

willkg:
I think the changes in here look fine. I need to do and land a cloudops PR before landing this that has the configuration change that needs to be made.

Also, next week is release week, so I'm inclined to sit on this until after release week which would let us push fixes out without worrying about this one.

Given that, I'm going to tag this with DONOTMERGE for now.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -606,75 +608,68 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
-            self._versions_data_cache[key] = real_version
+            'build_id': build_id
+        })
 
-        return self._versions_data_cache.get(key)
+        if resp.status_code == 200:
+            hits = resp.json()['hits']
+            if hits:
+                self.cache[key] = hits[0]
+                return hits[0]
```
willkg: I need to fix this. If we get back an HTTP 200 and there are *no hits*, we should cache that response, too.

willkg:
^^^ I reworked caching. This should be more better in regards to several scenarios:

1. processor startup and how it's going to hit the webapp API endpoint
2. management of responses over time in relation to hitting the webapp API endpoint
3. updating the processor's understanding of the world over time

We can see how often the processor requests real versions from the webapp from the Datadog dashboard. We can also see how much memory the processors are using in Datadog. Thus we can make sure that's working ok on stage before moving to production.

socorro/external/postgresql/version_string.py
```diff
@@ -0,0 +1,49 @@
...
+        """
+        results = self.query(sql, params)
+
+        # The query can return multiple results, but they're the same value. So
+        # we just return the first one.
+        version_string = [
+            row['version_string'] for row in results.zipped()
+        ]
+        if version_string:
+            version_string = [version_string[0]]
```
willkg: All the API endpoints return `hits: list of things`, so I made this one do that, too. It's a weird shape in this specific case, but I figure uniformity is helpful. If it turns out to be a bad idea, we can change it.

socorro/lib/requestslib.py
```diff
@@ -0,0 +1,33 @@
...
+    base_url = urlparse.urlparse(url).netloc
+    scheme = urlparse.urlparse(url).scheme
+
+    retries = Retry(total=32, backoff_factor=1, status_forcelist=[429])
+
+    s = requests.Session()
+
+    # Set the User-Agent header so we can distinguish our stuff from other stuff
+    s.headers.update({
+        'User-Agent': 'socorro-requests/1.0'
```
willkg: We should have been doing this all along. It'll make life for a lot of people a lot easier.

socorro/processor/processor_2015.py
```diff
@@ -123,6 +123,11 @@ class Processor2015(RequiredConfig):
         from_string_converter=str_to_python_object,
         reference_value_from='resource.postgresql',
     )
+    required_config.add_option(
+        'version_string_api',
+        doc='url for the version string api endpoint',
+        default='https://crash-stats.mozilla.com/api/VersionString'
```
willkg: We need to set this to the webapp in -stage on -stage.

willkg:
This looks good. I'll talk with Miles and Brian about it, then do the ops PR, then land this.

willkg:
I split out and fixed the `session_with_retries` changes. Those were getting ornery and it made more sense to land them on their own.

I'm going to close this out and open a new PR with just the `BetaVersionRule` changes.

PR #4572, opened 2018-08-25
fix bug 1486083: remove /api/ProductBuildTypes

willkg:


webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -129,25 +128,31 @@ def test_cache_control(self):
...
+        ProductVersions.implementation().get.side_effect = mocked_get
 
-        url = reverse('api:model_wrapper', args=('ProductBuildTypes',))
+        url = reverse('api:model_wrapper', args=('ProductVersions',))
         response = self.client.get(url, {'product': settings.DEFAULT_PRODUCT})
         assert response.status_code == 200
         assert response['Cache-Control']
         assert 'private' in response['Cache-Control']
-        cache_seconds = ProductBuildTypes.cache_seconds
+        cache_seconds = ProductVersions.cache_seconds
```
willkg: This test was using `ProductBuildTypes` to test caching, so I rewrote it to use `ProductVersions` instead.

willkg:
This is straight-forward code removal. It leaves behind a `product_build_types` table. That's a whole different can of worms covered in bug 1387708 so I'm going to leave that alone.

self-r+

PR #4573, opened 2018-08-25
fix bug 1323033: add doc_values=True to ES mapping

willkg:
This adds `doc_values=True` to fields that are either not strings or explicitly
have `index=not_analyzed`. Per the ES docs, this will cause ES to store data
on disk for sorting and aggregations rather than in memory which should
reduce some of our memory problems when doing aggregations.

### DIFF
--- a/socorro/external/es/index_creator.py
+++ b/socorro/external/es/index_creator.py
@@ -2,16 +2,11 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import os
-
 from configman import Namespace, RequiredConfig
 from configman.converters import class_converter
 import elasticsearch
 
 from socorro.external.es.super_search_fields import SuperSearchFields
-
-
-DIRECTORY = os.path.dirname(os.path.abspath(__file__))
 
 
 # Elasticsearch indices configuration.
@@ -45,8 +40,10 @@ class IndexCreator(RequiredConfig):
     required_config.elasticsearch.add_option(
         'shards_per_index',
         default=10,
-        doc='number of shards to set in newly created indices. Elasticsearch '
-            'default is 5.',
+        doc=(
+            'number of shards to set in newly created indices. Elasticsearch '
+            'default is 5.'
+        )
     )
 
     def __init__(self, config):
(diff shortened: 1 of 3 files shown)


socorro/external/es/index_creator.py
```diff
@@ -2,18 +2,13 @@
...
-import os
-
 from configman import Namespace, RequiredConfig
 from configman.converters import class_converter
 import elasticsearch
 
 from socorro.external.es.super_search_fields import SuperSearchFields
 
 
-DIRECTORY = os.path.dirname(os.path.abspath(__file__))
```
willkg: This wasn't used anywhere, so I nixed it and the `os` import it used.

socorro/external/es/index_creator.py
```diff
@@ -45,8 +40,10 @@ class IndexCreator(RequiredConfig):
     required_config.elasticsearch.add_option(
         'shards_per_index',
         default=10,
-        doc='number of shards to set in newly created indices. Elasticsearch '
-            'default is 5.',
+        doc=(
+            'number of shards to set in newly created indices. Elasticsearch '
+            'default is 5.'
+        )
```
willkg: The parens make it easier for me to read this as "one string" rather than multiple strings. That's a me thing.

socorro/external/es/super_search_fields.py
```diff
@@ -142,6 +140,23 @@ def add_field_to_properties(properties, namespaces, field):
...
+            if storage_mapping.get('fields'):
+                for subfield in storage_mapping.get('fields', {}).keys():
+                    field_mapping = storage_mapping['fields'][subfield]
+                    if ((field_mapping.get('type') != 'string' or
+                         field_mapping.get('index') == 'not_analyzed')):
+                        field_mapping['doc_values'] = True
+
+            if ((storage_mapping.get('type') != 'string' or
+                 storage_mapping.get('index') == 'not_analyzed')):
+                storage_mapping['doc_values'] = True
```
willkg: This block is the meat of the change. We want to change the mapping for fields to add a `doc_values=True` iff:

1. the field is not a string type, OR
2. the field is a string type, but explicitly has `index=not_analyzed`

Fields can be nested. So for example, we have a field `abort_message` and we store an analyzed version of it, but also an unanalyzed version of it. We use the two for different search scenarios.

Given that fields can be nested (only one level of nesting), we want to make sure we get all the subfields.

This isn't glamorous code. Elasticsearch 2.0 switches the default from `doc_values=False` to `doc_values=True`. When/if we upgrade to something >= 2.0, we can nix this code because we won't need it anymore.

willkg:
I'm going to land this now so that it kicks in when stage builds new indexes which I think happens tomorrow. Then we can see how it fares.

self-r+

PR #4574, opened 2018-08-28
fix bug 1323033: fix doc_values for nested and boolean cases

willkg:
This fixes `doc_values` in Elasticsearch mappings for the `boolean` and nested
cases. `boolean` fields can't have `doc_values=True`. This isn't documented
anywhere, but appears to be the case as near as we can determine.

This also fixes nested cases where a field may have sub fields or be
of an `object` type with properties. The code now traverses the structures
adding `doc_values=True` where appropriate.

### DIFF
--- a/socorro/external/es/super_search_fields.py
+++ b/socorro/external/es/super_search_fields.py
@@ -10,6 +10,104 @@ from socorro.external.es.base import ElasticsearchBase
 from socorro.lib import datetimeutil, BadArgumentError
 
 
+def add_field_to_properties(properties, namespaces, field):
+    """Add a field to a mapping properties
+
+    An Elasticsearch mapping is a specification for how to index all
+    the fields for a document type. This builds that mapping one field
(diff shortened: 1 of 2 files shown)


socorro/external/es/super_search_fields.py
```diff
@@ -136,26 +136,53 @@ def add_field_to_properties(properties, namespaces, field):
...
+
+            # Handle subfields
+            if props.get('fields'):
+                for field in props.get('fields', {}).values():
+                    add_doc_values(field)
+
+            # Handle objects with nested properties
+            if props.get('properties'):
+                for field in props['properties'].values():
+                    add_doc_values(field)
```
willkg: These should be extracted. It'd be a lot easier to test and verify them if they weren't defined in the method they're used.

willkg:
This looks good. It fixes the curiosities that I saw when comparing the mapping on stage, the mapping on my local machine, and the mapping that was getting created via `get_mapping`.

PR #4575, opened 2018-08-28
bug 1323033: Apply doc_values=True to boolean fields

willkg:
Elasticsearch 1.4 seems to drop `doc_values=True` for boolean fields as
observed by the mapping we use to create the index has these properties,
but the mapping we get when querying Elasticsearch afterwards doesn't.

The documentation suggests boolean values should work with `doc_values=True`.

Given that, we're going to re-add it since it doesn't seem to break
anything even if it has no effect.


willkg:
Minor cosmetic changes. self-r+

PR #4576, opened 2018-08-28
fix bug 1486910: remove Releases api code

willkg:
This is no longer used, so it's time to remove it.


willkg:
Just dead code removal. self-r+

PR #4577, opened 2018-08-28
fix bug 1306643: document signature generation pipeline

willkg:
This adds a command that generates a restructured text file, the generated
file, and tweaks the docs to include that in the signature generation
chapter.

This also documents the signature generation rules.


willkg:
This is a good enough first pass. Mostly it only affects documentation.

PR #4578, opened 2018-08-29
bug 1474037: add support for FirefoxReality

willkg:


alembic/versions/f355b5458f82_bug_1474037_firefoxreality.py
```diff
@@ -0,0 +1,34 @@
...
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = 'f355b5458f82'
+down_revision = 'fab476fc2f5e'
+
+
+PRODUCTNAME = 'FirefoxReality'
+SORT = 4
```
willkg: I decided to try this out so it's easier to copy/paste migrations in the future. I thought about doing a function in `socorro.lib`, but those change over time and the migrations that use them don't, so I think that's a bad model to follow.

Anyhow, just trying this out to see what happens.

alembic/versions/f355b5458f82_bug_1474037_firefoxreality.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def upgrade():
+    op.execute("""
+        INSERT INTO products
+        (product_name, sort, release_name, rapid_beta_version, rapid_release_version)
+        VALUES
+        ('%s', %s, '%s', 1.0, 1.0)
+    """ % (PRODUCTNAME, SORT, PRODUCTNAME.lower()))
```
willkg: Alembic's `op.execute` doesn't support parameters. We could use their insert function which generates the insert statement. I looked at that for a bit and decided it's less risky for us to use something that's not deeply abstracted even if it involves interpolation of variables into SQL.

willkg:
self-r+

PR #4579, opened 2018-08-29
fix bug 1436063: set processor threads in my.env.dist

willkg:
In a local development environment, you probably don't want 16
processor threads running. You're much more likely to want something
like 2.

This makes that change to my.env.dist. Developers can change the
number to meet their specific needs.


willkg:
self-r+

This will make life nicer for new developers.

PR #4580, opened 2018-08-29
Fix signature generation with python 3

willkg:
* remove a couple of files from the list of files to test in python 3
  that don't exist anymore
* fix issues with signature generation code and tests that caused
  them to fail in python 3


willkg:
self-r+

PR #4581, opened 2018-08-30
fix bug 1487381: fix topcrasher IndexError bug

willkg:
If we add a new product, but it hasn't gotten any crash reports, then
there are no versions and topcrashers errors out.

The version it needs it solely to figure out which version to select
in the drop down menu. So we set it to "" if there are no versions.


willkg:
Tiny fix. Works super. self-r+

PR #4582, opened 2018-08-30
fix bug 1487387: add 63.0b1 to AuroraVersionFixitRule

willkg:


willkg:
I verified the build id and the version. Going to land this and test it on stage. self-r+

PR #4583, opened 2018-08-31
Add table of contents to API documentation

willkg:
SuperSearch is super super super long. Having a table of contents at the
top of the page makes it much easier to see which APIs are available
and also get to that part of the page quickly without excessive scrolling.


willkg:
Tags are balanced. Looks good. self-r+

PR #4584, opened 2018-08-31
bug 1485477: remove breakpad_pipe_to_json.py code

willkg:
We don't use this in Socorro, so we should remove it.


willkg:
Just code removal. self-r+

PR #4586, opened 2018-09-10
Adjust docs building so "make build" doesn't build docs image

willkg:
9 out of 10 socorro engineers don't build docs on a regular basis, so it
doesn't make sense to spend time building the image for building docs
every time someone runs "make build".

This adjusts things so building the docs image is done when only when
building the docs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,12 +38,12 @@ help:
 	@echo "See https://socorro.readthedocs.io/ for more documentation."
 
 clean:
-	-rm .docker-build
+	-rm .docker-build*
 	-rm -rf build breakpad stackwalk google-breakpad breakpad.tar.gz
 	-rm -rf .cache
 	cd minidump-stackwalk && make clean
 
-docs: my.env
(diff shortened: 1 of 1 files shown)


willkg:
The only thing CI will do to test this is to run `make build` during its setup. That should take slightly less time now. Otherwise CI passing doesn't really tell us anything.

Running `make clean` should clean the state of the project. Running `make docs` after that will build the `docs` image and then compile the docs using it. Running `make docs` again will just compile the docs. That all works.

willkg:
For some numbers, the `make build` step in CI takes about 7m30s (I looked at the last few runs). With this fix, the CI run took 5m30s. I'm surprised--that's a lot less time than I expected. It could be a fluke. It's hard to look at multiple data points, so we've just got this one.

However, given that this fix causes `make build` to do *less* work, my intuition is that it should reduce CI `make build` time even if this specific data point is a fluke. I'll keep an eye on things for a while and see what happens.

self-r+.

PR #4587, opened 2018-09-10
Add shell rule

willkg:


willkg:
Minor inconsequential, but handy, stuff. self-r+

PR #4589, opened 2018-09-11
fix bug 1490486: add *buildhub.json to ftpscraper ignore list

willkg:


willkg:
It's a minor well-defined change. self-r+

PR #4591, opened 2018-09-12
fix bug 1490662, 1420797, 1467856: add fields to super search

willkg:
This adds some fields to super search that have been requested.


willkg:
I went through and tested these with relevant crashes locally. self-r+

PR #4592, opened 2018-09-13
fix bug 1490622: add values for memory_error_correction

willkg:


socorro/external/es/super_search_fields.py
```diff
@@ -2795,9 +2795,17 @@ def get(self):
...
+        'form_field_choices': [
+            'Reserved',
+            'Other',
+            'Unknown',
+            'None',
+            'Parity',
+            'Single-bit ECC',
+            'Multi-bit ECC',
+            'CRC',
+            'Unexpected value',
```
willkg: These come from:

https://bugzilla.mozilla.org/show_bug.cgi?id=1490662#c

willkg:
self-r+

PR #4593, opened 2018-09-13
fix bug 1417612: use java_stack_trace if available

willkg:
When generating bugzilla submit urls, if the crash report has a
`java_stack_trace`, then that should be in the comment rather than
the crashing thread stack.

### DIFF
--- a/webapp-django/crashstats/crashstats/jinja2/crashstats/bugzilla_comment.txt
+++ b/webapp-django/crashstats/crashstats/jinja2/crashstats/bugzilla_comment.txt
@@ -4,7 +4,11 @@ This bug was filed from the Socorro interface and is
 report bp-{{ uuid }}.
 =============================================================
 
-{% if crashing_thread_frames -%}
+{% if java_stack_trace -%}
+Java stack trace:
+
+{{ java_stack_trace }}
+{% elif crashing_thread_frames -%}
 Top {{ crashing_thread_frames|length }} frames of crashing thread:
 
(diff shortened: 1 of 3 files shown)


webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -206,6 +206,7 @@ def bugzilla_submit_url(report, parsed_dump, crashing_thread, bug_product):
 
     comment = render_to_string('crashstats/bugzilla_comment.txt', {
         'uuid': report['uuid'],
+        'java_stack_trace': report.get('java_stack_trace', None),
```
willkg: The `java_stack_trace` is just a big string, so we don't need to do any further processing/adjusting of it.

webapp-django/crashstats/crashstats/jinja2/crashstats/bugzilla_comment.txt
```diff
@@ -4,7 +4,11 @@ This bug was filed from the Socorro interface and is
 report bp-{{ uuid }}.
 =============================================================
 
-{% if crashing_thread_frames -%}
+{% if java_stack_trace -%}
+Java stack trace:
+
+{{ java_stack_trace }}
```
willkg: This just uses the string verbatim--it doesn't truncate it. I figured I'd kick that can down the road because it involves figuring out "what's too long?" and "what should we do if it's too long?" This makes the situation better. We can make it more better later.

PR #4594, opened 2018-09-13
Cosmetic README tweaks

willkg:
I just need to trigger a stage deploy. Figured this was an easy way to do it.


willkg:
There's nothing testable in this, so I'm self-r+ and skipping CI.

PR #4595, opened 2018-09-14
fix bug 1474037: add rewrite rule for FirefoxReality

willkg:
Incoming FirefoxReality crash reports have the wrong ProductName, so they're
getting put in the FennecAndroid bin. This rewrites the ProductName for crash
reports that have `Android_Manufacturer=Oculus` to FirefoxReality.


willkg:
self-r+. I verified it locally and will double-check it on stage.

PR #4598, opened 2018-09-17
fix bug 1491431: get versions from bigger window

willkg:
For products that don't have build information in `product_versions` table,
we pull versions for crash reports from Elasticsearch. This increases the
window we look at when doing that.

### DIFF
--- a/webapp-django/crashstats/crashstats/tests/test_utils.py
+++ b/webapp-django/crashstats/crashstats/tests/test_utils.py
@@ -411,15 +411,13 @@ def test_parse_version(version_string, expected):
     assert utils.parse_version(version_string) == expected
 
 
-def test_parse_version_sorted():
+def test_sorted_versions():
     """Test that using parse_version produces version keys that sort correctly"""
     versions = ['59.0', '59.0b2', '59.0.2', '59.0a1', '60p', '59.0.2esr']
-    assert sorted(versions, key=utils.parse_version, reverse=1) == [
+    assert utils.sorted_versions(versions) == [
         '59.0.2esr',
         '59.0.2',
         '59.0',
         '59.0b2',
         '59.0a1',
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/crashstats/tests/test_utils.py
```diff
@@ -411,15 +411,13 @@ def test_parse_version(version_string, expected):
...
     """Test that using parse_version produces version keys that sort correctly"""
     versions = ['59.0', '59.0b2', '59.0.2', '59.0a1', '60p', '59.0.2esr']
-    assert sorted(versions, key=utils.parse_version, reverse=1) == [
+    assert utils.sorted_versions(versions) == [
         '59.0.2esr',
         '59.0.2',
         '59.0',
         '59.0b2',
         '59.0a1',
-        # junk data, so it goes to the bottom of the pile
```
willkg: I changed it so that it doesn't keep "junk versions" anymore--those get nixed.

webapp-django/crashstats/crashstats/utils.py
```diff
@@ -212,11 +213,47 @@ def parse_version(version):
...
+
+    :returns: sorted list of version strings
+
+    Example:
+
+    >>> sorted_versions(['63.0', '63.0.2', '62.0b1', '62.0esr', '47.0'])
+    ['63.0.2', '63.0', '62.0esr', '62.0b1', '47.0'])
+
+    """
+    # Build a list of (version, parsed) tuples
```
willkg: Oops--this is wrong. It's the other way around.

willkg:
The code changes here seem good. I'll double-check it on stage. It's likely we'll need to hone this further over time as Focus and FirefoxReality get going and we have more data affecting the version information.

self-r+.

PR #4599, opened 2018-09-18
fix bug 1346883: remove postgres from processor

willkg:
The processor no longer needs postgres to function. This removes the
configuration from the app and the service from the docker-compose file
(which only affects the local development environment).

There are also a couple of cosmetic tweaks to code because I just
couldn't help myself.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -32,7 +32,6 @@ services:
     depends_on:
       - statsd
       - localstack-s3
-      - postgresql
       - elasticsearch
       - rabbitmq
     volumes:
--- a/socorro/processor/processor_2015.py
+++ b/socorro/processor/processor_2015.py
@@ -108,17 +108,8 @@ class Processor2015(RequiredConfig):
 
     required_config = Namespace('transform_rules')
     required_config.add_option(
-        'database_class',
-        doc="the class of the database",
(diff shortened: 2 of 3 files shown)


socorro/processor/processor_2015.py
```diff
@@ -107,18 +107,9 @@ class Processor2015(RequiredConfig):
...
 
     required_config = Namespace('transform_rules')
-    required_config.add_option(
-        'database_class',
-        doc="the class of the database",
-        default='socorro.external.postgresql.connection_context.'
-                'ConnectionContext',
-        from_string_converter=str_to_python_object,
-        reference_value_from='resource.postgresql',
-    )
```
willkg: This removes the configuration that isn't being used anymore.

docker-compose.yml
```diff
@@ -32,7 +32,6 @@ services:
     depends_on:
       - statsd
       - localstack-s3
-      - postgresql
```
willkg: We no longer need the postgres container to depend on the postgresql service running in the local dev environment.

willkg:
To test this, I shut down all the docker containers, then brought up the processor and processed some crashes. Everything worked fine.

Once Circle is fine, I'll land this.

willkg:
self-r+!

PR #4601, opened 2018-09-20
fix bug 1468815, 1466974: switch to a single docker image

willkg:
Socorro has several separate services. Up to this point, each service had
its own Docker image that had to be built. This changes that and now all
services run with the same docker image, just with different entry point
commands.

This results in less time spent building images, uploading images,
and downloading images which should result in less time in CI and less
time during deploys.

This results in less comlexity--I have to go do something, which image
should I be doing it in?

Additionally, this fixes the Makefile PHONY-business and tweaks some
other things to clean them up.


willkg:
The test plan for this isn't very long, but it's somewhat work-intensive:

1. Run all the rules mentioned in `make help`: `clean`, `build`, `setup`, `updatedata`, `lint`, `test`, `testshell`, `shell`, `docs`, `run`, `stop`.
2. Build a local dev environment and make sure the processor, crontabber, and webapp all work:
   * `docker-compose up processor`
   * `docker-compose up crontabber`
   * `docker-compose up webapp`

After reviewing this, making sure it passes in CI (does it take longer?), and running through the above, the next step is to talk with @milescrabill and @sciurus to plan the infra changes required.

.circleci/config.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       - run:
           name: Lint
           working_directory: /socorro
-          command: docker run local/socorro_webapp ./docker/run_lint.sh
+          command: docker run local/socorro_app shell ./docker/run_lint.sh
```
willkg: The `local/socorro_app` image has an entrypoint now and that runs a script that takes a "service" as the first argument. `shell` is a "service" that runs whatever the other arguments are. So in this case it runs the `./docker/run_lint.sh` script.

.circleci/config.yml
```diff
@@ -91,10 +91,8 @@ jobs:
             # push on master or git tag
             if [ "${CIRCLE_BRANCH}" == "master" ] || [ -n "${CIRCLE_TAG}" ]; then
               echo "${DOCKER_PASSWORD}" | docker login -u="${DOCKER_USERNAME}" --password-stdin
-              for image in "socorro_processor" "socorro_webapp" "socorro_crontabber"; do
-                retry docker tag "local/${image}:latest" "mozilla/${image}:${DOCKER_TAG}"
-                retry docker push "mozilla/${image}:${DOCKER_TAG}"
-              done
+              retry docker tag "local/socorro_app:latest" "mozilla/socorro_app:${DOCKER_TAG}"
+              retry docker push "mozilla/socorro_app:${DOCKER_TAG}"
```
willkg: Yay! Less stuff to push!

docker/Dockerfile
```diff
@@ -19,3 +36,40 @@ RUN pip install -U 'pip>=8' && \
...
+# https://docs.docker.com/engine/reference/builder/#exec-form-entrypoint-example
+# To unset ENTRYPOINT, do:
+# ENTRYPOINT []
+
+# Processor:
+# CMD ["/app/docker/run_processor.sh"]
+# Webapp:
+# CMD ["/app/docker/run_webapp.sh"]
+# Crontabber:
+# CMD ["/app/docker/run_crontabber.sh"]
```
willkg: Oops--this is all old comment stuff.

docker/Dockerfile.docs
```diff
@@ -1,14 +1,15 @@
-FROM python:2.7.14-slim
+FROM python:2.7.15-slim
```
willkg: Oops--we want to use the same image as the rest of things, so this should get downgraded to 2.7.14.

docker/Dockerfile.docs
```diff
@@ -1,14 +1,15 @@
...
-COPY ./docker/set_up_ubuntu.sh /tmp/
-RUN DEBIAN_FRONTEND=noninteractive /tmp/set_up_ubuntu.sh
+WORKDIR /app/
+RUN groupadd --gid $groupid app && \
+    useradd -g app --uid $userid --shell /usr/sbin/nologin app
 
 # Install graphviz
-RUN DEBIAN_FRONTEND=noninteractive apt-get install -y graphviz
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y graphviz make
```
willkg: This removes a lot of stuff that the docs image doesn't require. Now it builds much faster.

docker/run_tests_in_docker.sh
```diff
@@ -62,63 +59,62 @@ if [ "$1" == "--shell" ]; then
            --env-file ./docker/config/test.env \
            --tty \
            --interactive \
+           --entrypoint="" \
            "${TESTIMAGE}" /bin/bash
+    exit $?
```
willkg: This `exit` causes this script to break out early so we can dedent the `else` block.

willkg:
Everything seems good here. I made the related changes in cloudops here:

https://github.com/mozilla-services/cloudops-deployment/pull/2574

@sciurus, @milescrabill: how does this look to you?

willkg:
This looks good. The cloudops changes were made and landed. I think we're good to test this out on stage now. self-r+

PR #4602, opened 2018-09-21
fix bug 1488774: remove cv/ref qualifiers in function names

willkg:
This adjusts the code to remove `const` at the end of function names
to also remove ref qualifiers `const&`, `&`, and `&&`. That way they
don't interact poorly with the code that removes types and specifiers.

### DIFF
--- a/socorro/signature/rules.py
+++ b/socorro/signature/rules.py
@@ -154,9 +154,10 @@ class CSignatureTool(SignatureTool):
 
     def normalize_cpp_function(self, function, line):
         """Normalizes a single cpp frame with a function"""
-        # If the function ends in "const", drop that
-        if function.endswith(' const'):
-            function = function[:-6]
+        # Drop member function cv/ref qualifiers like const, const&, &, and &&
+        for ref in ('const', 'const&', '&&', '&'):
+            if function.endswith(ref):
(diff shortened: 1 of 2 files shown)


socorro/signature/rules.py
```diff
@@ -154,9 +154,10 @@ def normalize_rust_function(self, function, line):
 
     def normalize_cpp_function(self, function, line):
         """Normalizes a single cpp frame with a function"""
-        # If the function ends in "const", drop that
-        if function.endswith(' const'):
-            function = function[:-6]
+        # Drop member function cv/ref qualifiers like const, const&, &, and &&
+        for ref in ('const', 'const&', '&&', '&'):
```
willkg: This covers things we've seen and what's here:

https://en.cppreference.com/w/cpp/language/member_functions

We can add others as they come up.

willkg:
This fixes the signature we're having problems with:

```
app@socorro:/app$ socorro-cmd signature 8137b58e-5c33-4171-96c9-b9def0180905
Crash id: 8137b58e-5c33-4171-96c9-b9def0180905
Original: &
New:      mozilla::jni::GlobalRef<T>::operator=
Same?:    False
```

PR #4605, opened 2018-09-22
fix bug 1478080: vendor crontabber

willkg:
This brings the necessary bits of crontabber back into Socorro and updates
it to pass linting and work with Python 3.

This doesn't bring in the tests since those are written for nose and there's
a lot of it and we really just want to test Socorro crontabber jobs--not an
entire library. If it turns out that was a bad decision, we can add (and
overhaul) them later.


willkg:
To test:

1. run `make build` to build the image without the crontabber python library
2. run `make setup` which runs the crontabber app with `--mark-success` to mark the existing jobs
3. run `make updatedata` to run ftpscraper and featured versions jobs
4. run the tests and the linter
5. run the crontabber app with `docker-compose up crontabber`

socorro/cron/connection_factory.py
```diff
@@ -0,0 +1,180 @@
...
+        reference_value_from='resource.postgresql',
+    )
+
+    # clients of this class may need to detect Exceptions raised in the
+    # underlying dbapi2 database module.  Rather that forcing them to import
+    # what should be a hidden module, we expose just the Exception. Clients
+    # can then just refer to it as ConnectionFactory.IntegrityError
+    IntegrityError = psycopg2.IntegrityError
+
+    #--------------------------------------------------------------------------
```
willkg: We can take these out.

socorro/cron/app.py
```diff
@@ -0,0 +1,1207 @@
...
+def line_splitter(text):
+    return [x.strip() for x in re.split('\n|,|;', text.strip())
+            if x.strip() and not x.strip().startswith('#')]
+
+
+def pipe_splitter(text):
+    return text.split('|', 1)[0]
+
+
+class CronTabberBase(RequiredConfig):
```
willkg: We can (and should) collapse this with `CronTabberApp`. We don't need to have both. Also, we can probably take out commands we don't use like "version"

willkg:
I think that's the best I can do to minimize the amount of stuff vendored. Anything more requires rewriting bits and I don't want to do that right now.

This looks good enough and everything I can do to test it locally I did. self-r+

PR #4607, opened 2018-09-24
fix bug 1493200: fix empty string signature generation

willkg:
There were two situations and a bug that were causing signature
generation to generate signatures of an empty string.

The bug was stomping on the signature regardless of what `CSignatureTool`
figured out. We only want to stomp on the signature if we have a valid
signature. That's fixed in this commit.

The two situations that weren't handled correctly both involve a space
between the last token and the previous token where the two tokens
together are the function name, args, and some `[clone .cold.xyz]` suffix.
This fixes that by making sure we're not losing bits.

### DIFF
--- a/socorro/signature/rules.py
+++ b/socorro/signature/rules.py
@@ -303,8 +303,8 @@ class CSignatureTool(SignatureTool):
 
             new_signature_list.append(a_signature)
 
-            # If we have some signature already and the signature does not match the prefix
-            # signatures regex, then it is the last one we add to the list.
+            # If the signature does not match the prefix signatures regex, then it is the last
+            # one we add to the list.
             if not self.prefix_signature_re.match(a_signature):
                 break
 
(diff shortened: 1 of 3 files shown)


willkg:
Output from signature generation on problematic crash reports.

```
Crash id: a1f5ddcb-acd4-4b02-8962-0fb950180919
Original: 
New:      [thunk]:mozilla::net::nsIOService::Release`adjustor{32}'
Same?:    False
```

and:

```
Crash id: cbb37a44-b845-48d5-bb60-d17f40180919
Original: 
New:      js::jit::Linker::newCode
Same?:    False
```

socorro/signature/rules.py
```diff
@@ -301,8 +303,8 @@ def _do_generate(self, source_list, hang_type, crashed_thread, delimiter=' | '):
 
             new_signature_list.append(a_signature)
 
-            # If the signature does not match the prefix signatures regex, then it is the last
-            # one we add to the list.
+            # If we have some signature already and the signature does not match the prefix
+            # signatures regex, then it is the last one we add to the list.
```
willkg: Oops--I should change this back.

willkg:
Looks good! self-r+

PR #4608, opened 2018-09-24
fixi bug 1389197: fix os_name to be "Unknown" and not empty string

willkg:
If an `os_name` can't be pulled from the minidump for whatever reason,
then set it to `"Unknown"` which works with the webapp interface better
rather than `""` which doesn't. This allows us to search for these
crashes and it makes it clearer that Socorro wasn't able to figure
out the `os_name` rather than who-knows-what-happened.

Also, this switches the `OSInfoRule` to use glom and getitem notation.


willkg:
Seems fine to me. self-r+

PR #4609, opened 2018-09-24
fix bug 1459216: drop signature_products and signature_products_rollup

willkg:
We don't use these tables anymore, so we should remove them.


willkg:
Looks good! self-r+

PR #4610, opened 2018-09-24
fix bug 1435076: remove Correlations model

willkg:
This got missed when removing all the other correlations-related things.


willkg:
This is just dead code removal. self-r+

PR #4612, opened 2018-09-26
fix bug 1457747: move graphics_device to Django

willkg:
This creates a new GraphicsDevice Django model, rewrites all the bits to use
that, removes the old GraphicsDevices pseudo-middleware, and creates a data
migration to move the data from the old table to the new table.

Now the webapp owns knowledge and maintenance of graphics devices.

The thing this doesn't do is delete the old table--we'll do that in a later
step once we know everything is good.


webapp-django/crashstats/crashstats/migrations/0001_initial.py
```diff
@@ -0,0 +1,14 @@
...
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+    ]
+
+    operations = [
+    ]
```
willkg: Having an empty initial commit made it muuuuuch easier to go back and forth with the state of the `crashstats` app while working on this PR.

webapp-django/crashstats/crashstats/migrations/0003_1457747_graphics_data_migration.py
```diff
@@ -0,0 +1,64 @@
...
+            'columns': ', '.join(columns)
+        }
+    )
+
+    print('')
+    count = 0
+    # Third, create new GraphicsDevice instances and save them to the db
+    for row in cursor.fetchall():
+        item = dict(zip(columns, row))
+        print('inserting', item)
```
willkg: This is here intentionally. It makes it easier to see what the migration did.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -32,15 +31,64 @@
...
+            name_pair = cache.get(cache_key)
+
+            if name_pair is None:
+                try:
+                    obj = self.get(adapter_hex=adapter_hex, vendor_hex=vendor_hexes[i])
+                except self.model.DoesNotExist:
+                    continue
+
+                name_pair = (obj.adapter_name, obj.vendor_name)
+                cache.set(cache_key, name_pair, 60 * 60 * 24)
```
willkg: This reimplementation is simpler than the original. It doesn't do any batching of queries. It (probably) isn't caching things helpfully. But it's a lot easier to read and I figured I'd see how it fared and optimize it later if need be.

As a side note, I'm pretty sure it's hierarchical vendor -> adapter. If so, then we might want to additionally do a query for just the vendor and show at least the vendor name even if there's no vendor/adapter in the db.

Things to think about another day.

webapp-django/crashstats/crashstats/tests/test_models.py
```diff
@@ -506,20 +443,7 @@ def mocked_get(**params):
...
-            'adapter_hex': set(['ahex1', 'ahex2']),
-            'vendor_hex': set(['vhex2', 'vhex1']),
-        }
-        assert params_called[0] == expected
-        expected = {
-            'adapter_hex': set(['ahex3']),
-            'vendor_hex': set(['vhex3']),
-        }
-        assert params_called[1] == expected
+        # FIXME(willkg): verify caching is working
```
willkg: This is intentionally left here. This test doesn't test the caching because I couldn't think of a good way to do it without mocking everything. Further, while I did implement caching, I'm not super convinced it's really important to have here. Anyhow, I figure someone else and/or future me can deal with this some other day and it helps to have a FIXME here to mark the issue.

willkg:
This is good enough. We'll find out how the migrations run when this goes to stage. Having said that, I have no idea if stage has migratable data. The admin page for Graphics Device kind of sucks in that it doesn't list what's in the table.

I'll land this tomorrow after asking for a dump of the contents of the table in stage.

willkg:
@sciurus did a data dump of stage and it definitely has data. I'm going to test the data migration with that.

Also, it occurred to me I could ditch some of the hand-written admin code for a Django model admin. I'm going to do that. Among other things, it makes it MUCH easier to see what's in the database.

webapp-django/crashstats/crashstats/migrations/0003_1457747_graphics_data_migration.py
```diff
@@ -0,0 +1,74 @@
...
+    print('')
+    insert_count = 0
+    skip_count = 0
+
+    # Third, create new GraphicsDevice instances and save them to the db.
+    # Looks like stage and prod have a *lot* of junk data. This is a lookup
+    # table so we're going to junk anything that doesn't have names.
+    for row in cursor.fetchall():
+        item = dict(zip(columns, row))
+        if item['vendor_name'] is None and item['adapter_name'] is None:
```
willkg: I figure if both of these columns are nulls, then this row isn't doing anything helpful and we should ditch it.

PR #4613, opened 2018-09-26
fix bug 1470704: fix rule metrics key

willkg:
Processor transform rules are part of the processor so the key should
start with "processor". This fixes that.


willkg:
Minor change, but brings these keys into spec with other Socorro keys. self-r+

PR #4614, opened 2018-09-26
fix bug 1488774: fix another cause of & signatures

willkg:
For this one kind of operator overloading, it ends up with this signature
that breaks the expectations for the `drop_prefix_and_return_type` code.
So instead of running that, we skip it. That prevents it from turning
into `"&"`.

### DIFF
--- a/socorro/signature/rules.py
+++ b/socorro/signature/rules.py
@@ -159,8 +159,11 @@ class CSignatureTool(SignatureTool):
             if function.endswith(ref):
                 function = function[:-len(ref)].strip()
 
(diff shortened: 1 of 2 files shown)


willkg:
Signature generation output:

```
app@socorro:/app$ socorro-cmd signature 56e7c06b-f3f4-47fa-a9f2-4fc670180911
Crash id: 56e7c06b-f3f4-47fa-a9f2-4fc670180911
Original: &
New:      JS::Heap<T>::operator JSObject* const&
Same?:    False
```

That's better.

willkg:
self-r+.

PR #4615, opened 2018-09-26
fix bug 1493200: fix the double-clone vexing variation

willkg:
It's possible for the function name to end with not one, but TWO
`[clone .cold.xxx]` things. This fixes the handling to cover that.


willkg:
self-r+

PR #4616, opened 2018-09-27
fix bug 1493200: fix an infinite loop

willkg:
The token merging should only happen if there are at least two tokens. This
fixes that.


sciurus:
Let's try it!

willkg:
The test case illuminates the issue and nicely proves correctness of the fix. self-r+

PR #4617, opened 2018-09-27
bug 1463121: add Signature Django model (step 1)

willkg:
This adds the `Signature` Django model, adds a Django model admin, adds a data
migration to copy everything from the old table to the new table, and
updates the `UpdateSignatureScronApp` to update both the old and new tables.


socorro/cron/jobs/update_signatures.py
```diff
@@ -48,9 +53,41 @@ class UpdateSignaturesCronApp(BaseCronApp):
...
+                sig[0]
+            )
+        except SQLDidNotReturnSingleRow:
+            sql = """
+            INSERT INTO crashstats_signature (signature, first_build, first_date)
+            VALUES (%s, %s, %s)
+            """
+            params = (signature, report_build, report_date)
+
+        execute_no_results(connection, sql, params)
```
willkg: The original does an "upsert", but I chose to do a basic implementation instead of an upsert because this is going to get converted to Django queryset stuff once we get crontabber into Django-land.

socorro/cron/jobs/update_signatures.py
```diff
@@ -48,9 +53,41 @@ class UpdateSignaturesCronApp(BaseCronApp):
...
+
     def run(self, end_datetime):
         # Truncate to the hour
-        end_datetime = end_datetime.replace(minute=0, second=0, microsecond=0)
+        # FIXME end_datetime = end_datetime.replace(minute=0, second=0, microsecond=0)
```
willkg: Oops--need to undo this. This made it a lot easier to test.

socorro/unittest/cron/jobs/test_update_signatures.py
```diff
@@ -26,17 +26,51 @@ def get(self, *args, **kwargs):
...
+        self._truncate_signatures()
+
     def _setup_config_manager(self):
         return super(UpdateSignaturesCronAppTestCase, self)._setup_config_manager(
             jobs_string='socorro.cron.jobs.update_signatures.UpdateSignaturesCronApp|1h'
         )
 
     def _truncate_signatures(self):
-        """Wipe the signatures table"""
+        """Wipe both signatures tables"""
```
willkg: Oops--can undo this change.

socorro/unittest/cron/jobs/test_update_signatures.py
```diff
@@ -86,7 +125,7 @@ def test_no_crashes_to_process(self, mock_supersearch):
         self.run_job_and_assert_success()
 
         # Assert that nothing got inserted
-        data = self.fetch_signatures_data()
+        data = self.fetch_signature_data()
```
willkg: I forgot to update this to check the new table, too. I'll do this now.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -87,6 +87,20 @@ class Meta:
...
+class Signature(models.Model):
+    """Bookkeeping table to keep track of when we first saw a signature"""
+    signature = models.TextField(
+        unique=True,
+        help_text='the crash report signature'
+    )
+    first_build = models.BigIntegerField(
+        help_text='the first build id this signature was seen in'
+    )
+    first_date = models.DateTimeField(
```
willkg: This was "first_report" in the old table, but I'm renaming it to "first_date". "report" is ambiguous and confusing.

willkg:
I think this covers everything.

I processed a bunch of crashes, ran the update-signatures crontabber job which populates the signatures table. Then I ran the migrations and that copied everything over to the new table. The update-signatures crontabber job updates the old and new tables at the same time. It inserts new signatures and updates existing ones with older dates and older builds.

New data is viewable in the Django admin.

This doesn't change any of the existing stuff that uses the data. That'll happen after this goes to prod and we know the data migration and update bits work.

self-r+

PR #4618, opened 2018-09-28
bug 1463121: redo data migration for signatures

willkg:
This redoes the data migration for signatures. The data on stage and prod
is *much* bigger than I thought it was. This reworks the code to do things
in batches and in bulk.

### DIFF
--- a/webapp-django/crashstats/crashstats/migrations/0005_1463121-signature-data-migration.py
+++ b/webapp-django/crashstats/crashstats/migrations/0005_1463121-signature-data-migration.py
@@ -3,6 +3,9 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import sys
+import time
+
 from django.db import connection, migrations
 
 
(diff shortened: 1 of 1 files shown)


willkg:
@milescrabill @sciurus Stage has 12m rows and this rewritten data migration takes about 30min on my local machine. That's definitely too long to run during a deploy.

I think the options are something like this:

1. we do the deploy, the migrations fail, ops runs this migration by hand
2. we junk the data migration and rewrite it as raw sql that ops runs by hand; this might run faster since it's not moving data around and there's only one dev so we don't *need* a data migration to update everyone else's data

Do you have a preference?

willkg:
Relatedly, I'm guessing we're going to have problems with other tables we're moving to Django-land. So whatever we decide to do here we should probably do going forward.

willkg:
I talked with @sciurus and he was all like, "I was a the yacht club talking to my friend and he said do the raw sql for lunch because it's super fresh." So I'm going to nix this for a different PR.

PR #4619, opened 2018-09-28
fix bug 1493827: make RecordReplay public and searchable

willkg:


willkg:
Looks good. I ran it with the example crash report. self-r+

PR #4620, opened 2018-09-28
bug 1463121: noop the data migration; create sql script

willkg:
We had problems running the data migration because there was waaaay
more data in that table than I thought there was.

This noops the data migration in favor of a sql script ops will run
by hand.


willkg:
@sciurus Does this look ok to you?

willkg:
Thank you!

PR #4621, opened 2018-09-28
fix bug 1495000: remove pcidatabase.com support

willkg:
The site is gone, so this removes the code that supported it.


willkg:
To test:

1. run the webapp, log in as an admin, view the Graphics Devices section in the admin
2. upload a pci-ids file
3. verify the file contents were loaded by looking at the other Graphics Devices section

webapp-django/crashstats/manage/admin.py
```diff
@@ -160,12 +160,11 @@ def graphics_devices(request):
...
-            if upload_form.cleaned_data['database'] == 'pcidatabase.com':
-                function = utils.pcidatabase__parse_graphics_devices_iterable
-            else:
-                function = utils.pci_ids__parse_graphics_devices_iterable
+            devices = utils.pci_ids__parse_graphics_devices_iterable(
+                upload_form.cleaned_data['file']
+            )
 
-            for item in function(upload_form.cleaned_data['file']):
+            for item in devices:
```
willkg: This is the interesting part of the change. Since we don't have two formats to support anymore, we can remove the dropdown from the form and the handling for it.

willkg:
Everything looks good here. It's mostly code removal. self-r+

PR #4623, opened 2018-09-28
fix bug 1493827: add RecordReplayError to search

willkg:


willkg:
I processed a crash with this field locally and verified it was searchable if you had PII. I think this is good to go. self-r+

PR #4625, opened 2018-09-29
fix bug 1491778: remove \r and \n from telemetry json blob

willkg:
The telemetry environment data comes in as a json blob. If the json blob
is malformed, then it won't display in the Telemetry Environment tab. One
way it appears to be malformed is that it has `\r` and `\n` characters in
it--the actual characters and not the escape sequences.

This removes them if they exist before trying to parse it as JSON.


willkg:
Scratch this--I see what's going on now.

PR #4630, opened 2018-10-03
bug 1463121: Add row counts for non-pg tables to debug view

willkg:
This will make it easier to verify that the `crashstats_signature` table
is growing at the same pace as the `signatures` table. Further, it shows
all the tables in the db so it'll surface other possible issues like whether
there are tables that shouldn't be there.


willkg:
This seems pretty straight-forward and it only affects a debug view in the admin that only I use. It works locally and I'll test it on stage before pushing it to prod. self-r+

PR #4631, opened 2018-10-03
Clean up transaction_executor

willkg:
When we vendored crontabber, we added a clause to the exception handling
in the `transaction_executor` which crontabber had in its version. This
fixes that code so it's not importing the postgres code. That way
it's generalized again.

Additionally, this removes code and tests for the `abandon_transaction`
feature that isn't used.

### DIFF
--- a/socorro/cron/jobs/bugzilla.py
+++ b/socorro/cron/jobs/bugzilla.py
@@ -80,12 +80,6 @@ def find_signatures(content):
     return signatures
 
 
-class NothingUsefulHappened(Exception):
-    """an exception to be raised when a pass through the inner loop has
-    done nothing useful and we wish to induce a transaction rollback"""
-    abandon_transaction = True
-
-
 @using_postgres()
 class BugzillaCronApp(BaseCronApp):
     """Updates Socorro's knowledge of which bugs cover which crash signatures
--- a/socorro/lib/transaction.py
+++ b/socorro/lib/transaction.py
@@ -113,14 +113,13 @@ class TransactionExecutorWithInfiniteBackoff(TransactionExecutor):
                         connection.commit()
                         return result
                     except Exception:
-                        connection.rollback()
                         last_failure = sys.exc_info()
+                        connection.rollback()
                         reraise(*last_failure)
 
             except self.db_conn_context_source.conditional_exceptions as x:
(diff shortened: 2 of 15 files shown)


socorro/cron/jobs/bugzilla.py
```diff
@@ -80,12 +80,6 @@ def find_signatures(content):
     return signatures
 
 
-class NothingUsefulHappened(Exception):
-    """an exception to be raised when a pass through the inner loop has
-    done nothing useful and we wish to induce a transaction rollback"""
-    abandon_transaction = True
```
willkg: This wasn't used anywhere, so I nixed it. It's the only thing that had a `abandon_transaction` property.

socorro/lib/transaction.py
```diff
@@ -2,14 +2,22 @@
...
 import time
 
 from configman.config_manager import RequiredConfig
 from configman import Namespace
 from six import reraise
 
-from socorro.external.postgresql.dbapi2_util import DBApiUtilNonFatalBaseException
+
+class NonFatalException(Exception):
+    """Exception for a non-fatal transaction that doesn't require reconnecting"""
```
willkg: We use the `TransactionExecutor` and friends across all the crash storage classes--not just the database. So I redid this part to undo tying it to our postgres stuff.

socorro/unittest/lib/test_transaction.py
```diff
@@ -236,9 +232,9 @@ def mock_function(connection):
             def mock_sleep(n):
                 _sleep_count.append(n)
 
-            # monkey patch the sleep function from inside transaction_executor
-            _orig_sleep = socorro.database.transaction_executor.time.sleep
-            socorro.database.transaction_executor.time.sleep = mock_sleep
+            # monkey patch the sleep function from inside transaction module
+            _orig_sleep = socorro.lib.transaction.time.sleep
+            socorro.lib.transaction.time.sleep = mock_sleep
```
willkg: I should rework this to use `mock.patch`.

willkg:
I think this is good. I'm going to wait to land this until after the prod deploy. self-r+

willkg:
Did a push to prod, so the decks are cleared for this to land.

PR #4632, opened 2018-10-04
bug 1463121: fix debug view db table counts to work in stage

willkg:


willkg:
self-r+. Will test in stage.

PR #4633, opened 2018-10-04
fix bug 1496495: remove implementation cache in webapp

willkg:
The implementation cache was helpful once, but I don't think it's helpful
anymore. It does cache mocks and other things, so it seems more trouble
than it's worth, so I'm removing it for now.


willkg:
We can re-add this later if we need to. self-r+

PR #4635, opened 2018-10-05
Fix fetch_crashids to work with other Crash STats sites

willkg:


willkg:
I just used this while reprocessing things on stage. It works super. self-r+

PR #4636, opened 2018-10-05
bug 1463121: signature step 2

willkg:
This does the following:

1. update top crashers to use the Django model
2. reimplement the `SignatureFirstDate` API endpoint using the Django model and rip out all the related middleware bits
3. fix `UpdateSignaturesCronApp` to only update the new table
4. remove the Signatures table model and bits
5. update tests and write any needed new ones

This doesn't delete the table. We can do that at a later point.


webapp-django/crashstats/crashstats/models.py
```diff
@@ -407,6 +402,20 @@ def _post(self, url, payload, method='post'):
...
+
+        for key, value in defaults.items():
+            kwargs[key] = kwargs.get(key) or value
+
+        params = self.kwargs_to_params(kwargs)
+        for param in params:
+            if aliases.get(param):
+                params[aliases.get(param)] = params[param]
+                del params[param]
+        return params
```
willkg: I extracted this out of `_get` so that I could use it elsewhere.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -900,7 +899,6 @@ class SignaturesByBugs(SocorroMiddleware):
 
 
 class SignatureFirstDate(SocorroMiddleware):
```
willkg: This is a Django-model using middleware. I think I'm going to go with this method of doing things for now. It supports the API documentation and other things, so it's convenient.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -927,50 +921,27 @@ class SignatureFirstDate(SocorroMiddleware):
...
-        return dates
+        # NOTE(willkg): The original implementation return first_build as a
+        # string, so we convert it here.
+        for hit in hits:
+            hit['first_build'] = str(hit['first_build'])
+
+        return {
+            'hits': hits,
+            'total': len(hits)
+        }
```
willkg: This didn't have any tests before. I should add one to at least verify the shape of the results.

willkg:
To test this:

1. Test the `/api/SignatureFirstDate` API endpoint. You can do this from the API documetnation page.
2. Test the Top Crashers page. This data affects the "First appearance" column.

That's it!

webapp-django/crashstats/crashstats/tests/test_models.py
```diff
@@ -34,6 +33,134 @@ def json(self):
...
+            ['vhex2', 'vhex3'],
+        )
+        assert len(r) == 2
+        expected = {
+            ('ahex2', 'vhex2'): ('A 2', 'V 2'),
+            ('ahex3', 'vhex3'): ('A 3', 'V 3'),
+        }
+        assert r == expected
+
+        # FIXME(willkg): verify caching is working
```
willkg: This was part of `TestModels` and I broke it out into its own class. Otherwise, the code is identical.

willkg:
I think this is good to go. I've tested it locally. I'll test it more thoroughly on stage.

PR #4637, opened 2018-10-08
fix bug 1497131: fix java stacktrace parsing for no message case

willkg:
This fixes java stacktrace parsing to handle cases where there's an exception
class, but no message. Previously, this would raise a `MalformedJavaStackTrace`
exception. Now it should parse correctly.


willkg:
self-r+

PR #4639, opened 2018-10-09
bug 1457747: rewrite GraphicsDeviceManager.get_pairs()

willkg:
This reworks `get_pairs` to do a single SQL query. It also ditches all the
caching work which I think is done at the wrong granularity.

### DIFF
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -44,39 +44,23 @@ logger = logging.getLogger('crashstats.models')
 
 
 class GraphicsDeviceManager(models.Manager):
-    def get_pair(self, adapter_hex, vendor_hex):
-        """Returns (adapter_name, vendor_name) or None"""
+    def get_pair(self, vendor_hex, adapter_hex):
+        """Returns (vendor_name, adapter_name) or None"""
         try:
-            obj = self.get(adapter_hex=adapter_hex, vendor_hex=vendor_hex)
-            return (obj.adapter_name, obj.vendor_name)
+            obj = self.get(vendor_hex=vendor_hex, adapter_hex=adapter_hex).values()
+            return (obj['vendor_name'], obj['adapter_name'])
         except self.model.DoesNotExist:
             return None
 
-    def get_pairs(self, adapter_hexes, vendor_hexes):
-        """Return dict where each tuple of (adapter_hex, vendor_hex)
-        corresponds to a (adapter_name, vendor_name) pair
-
-        """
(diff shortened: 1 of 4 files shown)


webapp-django/crashstats/crashstats/models.py
```diff
@@ -44,39 +44,23 @@
 
 
 class GraphicsDeviceManager(models.Manager):
-    def get_pair(self, adapter_hex, vendor_hex):
-        """Returns (adapter_name, vendor_name) or None"""
+    def get_pair(self, vendor_hex, adapter_hex):
```
willkg: I flipped the order of these arguments and the ones in `.get_pairs()` because there's a hierarchy to vendor and adapter and it makes more sense to me to do vendor first. Among other things, I keep putting vendor first and then have to correct it later. Now I don't have to do that anymore.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -44,39 +44,23 @@
...
-            names[key] = name_pair
+    def get_pairs(self, vendor_hexes, adapter_hexes):
+        """Returns dict of (vendor_hex, adapter_hex) -> (vendor_name, adapter_name)"""
+        # NOTE(willkg): graphics devices are hierarchical, but it's easier to do
+        # one query and get some extra stuff than to do one query per vendor
+        qs = self.filter(vendor_hex__in=vendor_hexes, adapter_hex__in=adapter_hexes).values()
+        names = {
+            (item['vendor_hex'], item['adapter_hex']): (item['vendor_name'], item['adapter_name'])
+            for item in qs
+        }
```
willkg: Previously, this was caching each pair. So instead of doing a SQL query for each pair, it'd:

* (cache hit) hit the cache
* (cache miss) hit the cache, get a miss, then do a SQL query, then cache the result

It'd do that on a pair basis, so a single HTTP request to the signature report view would generate, say, 15 of these and it'd do all this work 15 times.

Graphics driver data doesn't change ever. I think I updated it for the first time in 3 years last week. There's like 21,000 rows in the table, so I don't think it makes sense to pull the whole table and put that in memory, but maybe it does. We can think about that another time.

For now, I opted to reduce this down to a single SQL query per `.get_pairs()` call. It always does a single query. There's no caching of the results. The work involved is a lot less. That seemed like a good baseline from which we can measure and improve as we understand usage patterns.

willkg:
This seems good. I'm going to pull a few pages on stage, land this, and then see if the output of those pages change at all. self-r+

PR #4640, opened 2018-10-09
bug 1457747: fix KeyError issue in report_index view

willkg:
This fixes a `KeyError` from the last PR. Further, I adjusted the output of the hex vs. name in the report view to match what we do in the signature report.


willkg:
Looks good as far as I can tell. self-r+

PR #4641, opened 2018-10-10
bug 1493768: create BugAssociation Django model

willkg:
This creates a `BugAssociation` Django model and updates the bugzilla
cron job to save data to both the old and new tables.

This also adds a data migration script.

This also removes `test_run_job_based_on_last_success`. After we removed
the bugs table a couple of years ago, this test was adjusted and the
bugs table bits removed. As near as I can tell, this doesn't test anything
useful anymore.


socorro/cron/jobs/bugzilla.py
```diff
@@ -106,46 +109,56 @@ class BugzillaCronApp(BaseCronApp):
...
+                self.update_bug_data,
                 bug_id,
                 signature_set
             )
 
-    def inner_transaction(self, connection, bug_id, signature_set):
-        self.config.logger.debug("bug %s: %s", bug_id, signature_set)
+    def update_bug_data(self, connection, bug_id, signature_set):
+        self.update_bug_data_old_table(connection, bug_id, signature_set)
+        self.update_bug_data_new_table(connection, bug_id, signature_set)
```
willkg: The way I did this involves a lot of duplicated stuff, but it'll be much easier to remove in a couple of days.

willkg:
This seems good. We have the bits we need to verify it on stage, so I'm going to land it now and then we'll test out the data migration and move forward. self-r+

PR #4642, opened 2018-10-11
bug 1493768: redo webapp to use BugAssociation Django model

willkg:
This redoes the webapp to use the `BugAssociation` Django model instead of
the old `Bugs` middleware model.

There are a few things to point out.

First, the `Bugs` middleware model did this funny thing where if you passed
in signatures, then it would find all the bugs associated with those
signatures and then find all the signatures associated with all those bugs.
I redid this as `.get_bugs_and_related_bugs()`.

Second, some of the code that used the old `Bugs` middleware model relied on
this behavior. I redid that using `.get_bugs_and_related_bugs()`. Some of
the code filtered the results down to just the `(bug, signature)` sets for
signatures it asked for. I redid that code using just `.filter()`.

Third, I changed the shape of things, so I had to change templates to
look at the `bug_id` field instead of the `id` field.

This does not remove the old table, but it does remove all use of it.


willkg:
This touches a bunch of views, so testing is a bit involved and because it's using bug associations and that data is hard to acquire, it's hard to effectively test locally well. I did a pass locally, but will test more comprehensively on stage. :cry:

High level test plan:

1. (skip this step on stage) process a lot of crashes. like 100+
2. (skip this step on stage) reset bugzilla-associations cron job and then run it with `days_into_past` set to like 50
3. do a super search and find some signatures and make sure the signature facet shows bug ids
4. look at the Bugzilla tab in signature views for those signatures; it should have a "Bugs for ..." section and a "Related Crash Signatures" section with appropriate things in them
5. look at topcrasher report and make sure it shows bug ids
6. look at the Bugzilla tab of a specific crash report and make sure it shows appropriate bug ids
7. wait a bit and use the debug view in the admin to make sure the `crashstats_bugassociation` table count changes; it should change once an hour

webapp-django/crashstats/crashstats/models.py
```diff
@@ -880,10 +892,31 @@ class Bugs(SocorroMiddleware):
         ),
     }
 
+    def get(self, *args, **kwargs):
```
willkg: This returns bugs for the given signatures plus all the other bugs associated with signatures those bugs are associated with. Because of that, we use `.get_bugs_and_related_bugs()`.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -896,8 +929,31 @@ class SignaturesByBugs(SocorroMiddleware):
...
```
willkg: This returns the signatures for a set of bugs. Because of that, we just use `.filter()`.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -137,15 +137,12 @@ def report_index(request, crash_id, default_context=None):
...
-    bugs_api = models.Bugs()
-    hits = bugs_api.get(signatures=[context['report']['signature']])['hits']
-    # bugs_api.get(signatures=...) will return all signatures associated
-    # with the bugs found, but we only want those with matching signature
-    context['bug_associations'] = [
-        x for x in hits
-        if x['signature'] == context['report']['signature']
-    ]
-    context['bug_associations'].sort(key=lambda x: x['id'], reverse=True)
+    context['bug_associations'] = list(
```
willkg: The report view wants the list of bugs associated with this crash report's signature so we use `.filter()`.

webapp-django/crashstats/exploitability/views.py
```diff
@@ -115,10 +115,15 @@ def exploitability_report(request, default_context=None):
     # now, let's go back and fill in the bugs
     signatures = [x['signature'] for x in crashes]
     if signatures:
-        api = models.Bugs()
+        qs = (
```
willkg: The exploitability report wants the list of bugs for the list of signatures passed in so we use `.filter()`. Previously, it would get back bugs and related bugs and then only show the signatures that it was interested in so it was essentially ignoring the related bugs data.

webapp-django/crashstats/signature/views.py
```diff
@@ -571,17 +566,16 @@ def _transform_exploitability_summary(facets):
...
     signature = params['signature'][0]
     context['signature'] = signature
-
-    bugs_api = models.Bugs()
-    context['bugs'] = bugs_api.get(
-        signatures=[signature]
-    )['hits']
-
-    context['bugs'].sort(key=lambda x: x['id'], reverse=True)
+    context['bugs'] = list(
```
willkg: The signature report wants bugs and related bugs, so we use `.get_bugs_and_related_bugs()`.

PR #4643, opened 2018-10-11
bug 1493768: fix ordering of /api/Bugs results

willkg:
This was the one minor difference between the rewrite I did which is on stage and the old code which is in prod.


willkg:
self-r+

PR #4645, opened 2018-10-12
bug 1493768: add BugAssociationAdmin

willkg:


Makefile
```diff
@@ -103,7 +103,7 @@ testshell: my.env .docker-build
 
 .PHONY: run
 run: my.env
-	${DC} up processor webapp webpack oidcprovider
+	${DC} up processor webapp webpack
```
willkg: The `oidcprovider` service is a dependency of the `webapp`, so we don't need to run it here.

webapp-django/crashstats/crashstats/admin.py
```diff
@@ -84,7 +97,7 @@ class GraphicsDeviceAdmin(admin.ModelAdmin):
 
 
 @admin.register(Signature)
-class Signature(admin.ModelAdmin):
+class SignatureAdmin(admin.ModelAdmin):
```
willkg: This was a mistake. Not a big deal, but it was worth fixing.

willkg:
self-r+

PR #4646, opened 2018-10-12
bug 1498441: create Platform Django model and migration script

willkg:


webapp-django/crashstats/crashstats/models.py
```diff
@@ -107,6 +107,18 @@ class Meta:
         unique_together = ('vendor_hex', 'adapter_hex')
 
 
+class Platform(models.Model):
```
willkg: The original table name is `os_names`. The API endpoint was Platform, the middleware model is named "Platform".

"Windows 10" is an operating system name, but the table holds "Windows" which is more of a platform than a specific operating system. 

Given that, I went with "Platform".

webapp-django/crashstats/crashstats/models.py
```diff
@@ -107,6 +107,18 @@ class Meta:
...
+    """Lookup table for platforms"""
+    name = models.CharField(
+        max_length=20, blank=False, null=False,
```
willkg: This field has four values presently: "Windows", "Mac OS X", "Linux", and "Unknown". None of these are 20 characters, but I figured I'd give it a bit of room.

willkg:
Everything looks good. I verified the table creation migration, the data migration, and the admin page. self-r+

PR #4647, opened 2018-10-12
bug 1498441: rewrite everything to use new Platform model

willkg:
This rewrites everything to use the new Platform Django model.

This also removes the `/api/Platforms/` endpoint. I checked the dashboards
and it's not being used. It returned a static result, so anyone can just
hardcode that.

This removes the staticdata for the `OsName` model. In its place, it adds
a Django fixture and adds a line to the setup script to load that.

This also adds a schema migration to fix the `crashstats_platform` table
so that "name" is unique.


webapp-django/crashstats/topcrashers/views.py
```diff
@@ -300,6 +295,6 @@ def topcrashers(request, days=None, possible_days=None, default_context=None):
             label = option.capitalize()
         context['process_type_values'].append((value, label))
 
-    context['platform_values'] = settings.DISPLAY_OS_NAMES
+    context['platform_values'] = [item['name'] for item in platforms]
```
willkg: It's almost tempting to turn this entire table into a constant defined in the code. I'll think about that and maybe redo this that way.

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -183,27 +183,24 @@ def mocked_get(*args, **k):
...
+        url = reverse('api:model_wrapper', args=('CrontabberState',))
         with MetricsMock() as metrics_mock:
             response = self.client.get(url)
         assert response.status_code == 200
         assert metrics_mock.has_record(
             fun_name='incr',
             stat='webapp.api.pageview',
             value=1,
-            tags=['endpoint:apiPlatforms']
+            tags=['endpoint:apiCrontabberState']
```
willkg: We're getting rid of the `/api/Platforms`, so I had to switch this to use `/api/CrontabberState`.

webapp-django/crashstats/base/tests/test_utils.py
```diff
@@ -50,10 +50,11 @@ def test_SignatureStats(self):
...
         }
         platforms = [
-            {'code': 'win', 'name': 'Windows'},
-            {'code': 'mac', 'name': 'Mac OS X'},
-            {'code': 'lin', 'name': 'Linux'},
-            {'code': 'unknown', 'name': 'Unknown'}]
+            {'short_name': 'win', 'name': 'Windows'},
+            {'short_name': 'mac', 'name': 'Mac OS X'},
+            {'short_name': 'lin', 'name': 'Linux'},
+            {'short_name': 'unknown', 'name': 'Unknown'}
```
willkg: We switched platforms to have `name` and `short_name` rather than `name` and `code`.

webapp-django/crashstats/topcrashers/views.py
willkg: I'll think about that another day. For now, we'll just turn it into a table.

willkg:
This looks good enough. I'm going to wait a little bit to land it once the stage queue has cleared.

willkg:
self-r+

PR #4648, opened 2018-10-17
fix bug 1499714: further restrict VersionString by channel

willkg:
This fixes the problem with b99 where there are two different version
strings and the VersionString API code picked the "first" one, but it's
unsorted, so it's really a random one and random isn't awesome here.

### DIFF
--- a/socorro/external/postgresql/version_string.py
+++ b/socorro/external/postgresql/version_string.py
@@ -17,9 +17,10 @@ class VersionString(PostgreSQLBase):
             ('product', None, 'str'),
             ('version', None, 'str'),
             ('build_id', None, 'int'),
+            ('release_channel', None, 'str'),
         ]
         params = external_common.parse_arguments(filters, kwargs)
-        required = ('product', 'build_id', 'version')
+        required = ('product', 'build_id', 'version', 'release_channel')
         for key in required:
             if not params.get(key):
                 raise MissingArgumentError(key)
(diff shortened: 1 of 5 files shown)


socorro/external/postgresql/version_string.py
```diff
@@ -33,6 +34,7 @@ def get(self, **kwargs):
             WHERE pv.product_name = %(product)s
             AND pv.release_version = %(version)s
             AND pvb.build_id = %(build_id)s
+            AND pv.build_type = %(release_channel)s
```
willkg: That table has "build_type" and "build_type"enum" columns. We want the "build_type" one because the other appears to have been deprecated.

This doesn't matter a whole lot because I'm rewriting all that stuff soon.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -648,18 +648,16 @@ def _get_version_data(self, product, version, build_id):
...
 
         :raises requests.RequestException: raised if it has connection issues with
             the host specified in ``version_string_api``
 
         """
-        if not (product and version and build_id):
-            return None
-
-        key = '%s:%s:%s' % (product, version, build_id)
+        key = '%s:%s:%s:%s' % (product, version, build_id, release_channel)
```
willkg: I adjusted the caller so that it won't call this method with bad data.

willkg:
This looks good to me. I'm going to push it to stage and reprocess the affected crashes there and see where things are at. self-r+

PR #4649, opened 2018-10-17
fix bug 1499381: fix optgroups in dropdowns

willkg:
The navigation dropdown contents have changed significantly over time. This
tweaks the HTML to reduce some of the sparseness now that we've removed
a lot.


willkg:
self-r+

PR #4653, opened 2018-10-19
Update docs

willkg:
This updates documentation regarding how to run the shell and other command
line things. Many of these changed when we switched to a single image.


willkg:
Docs fixes. self-r+

PR #4654, opened 2018-10-22
More docs

willkg:


willkg:
These are docs and github's webhook stuff is down, so I'm going to merge this without the tests. self-r+

PR #4657, opened 2018-10-23
fix bug 1501346: fix api use of classnames

willkg:
This fixes the API code so that it can use model_name and model_name +
"Middleware" when looking up models for handling. This allows us
more leeway when naming middleware api classes.


willkg:
This looks fine. I went through and verified the API still works. It's interesting to note that only `ProductVersions` is still a valid endpoint and it's deprecated.

Anyhow, self-r+.

PR #4658, opened 2018-10-24
fix bug 1501780: rewrite BetaVersionRule to use Buildhub

willkg:
This removes the `VersionString` API which we no longer need since the
`BetaVersionRule` was the only user of it.

This removes the `AuroraVersionFixitRule` because we don't need it anymore.
That rule was in place because ftpscraper and the related stored procedures
didn't handle devedition aurora well causing the `BetaVersionRule` to muck
up version strings for b1 and b2. Buildhub doesn't have that problem so
we can get rid of it.


socorro/processor/processor_2015.py
```diff
@@ -115,9 +113,9 @@ class Processor2015(RequiredConfig):
         reference_value_from='resource.postgresql',
     )
     required_config.add_option(
-        'version_string_api',
-        doc='url for the version string api endpoint',
-        default='https://crash-stats.mozilla.com/api/VersionString'
+        'buildhub_api',
+        doc='url for the Buildhub api',
+        default='https://buildhub.prod.mozaws.net/v1/'
```
willkg: This should change to the full url. That way we can change the collection and bucket when testing more easily.

willkg:
This seems fine. I'm going to land it and see how it goes on -stage over the weekend. self-r+

PR #4661, opened 2018-10-25
fix bug 1484755: remove Focus rewrite rule

willkg:
We no longer need to rewrite incoming Focus crashes, so we can remove
this rule now.


willkg:
self-r+

PR #4662, opened 2018-10-25
fix bug 1502038: rework version sorting

willkg:
This handles the kinds of versions we have on Crash Stats including
"catch-all" versions like 62.0b which covers all the betas. Rewriting
it allows us to save it to the db as a string and also use it
elsewhere in the code.


willkg:
It's hard to test this more locally, but it'll be easy to verify it on stage.

willkg:
This looks ok. I want to land this and see how it does on stage and adjust accordingly. self-r+

PR #4663, opened 2018-10-25
Fix bug 1502038: tweak order of versions in navbar

willkg:


willkg:
This fixes a minor goof in ordering in the navbar and product home page. self-r+

PR #4664, opened 2018-10-25
fix bug 1501780: fix BetaVersionRule lookups for aurora

willkg:


willkg:
This fixes the aurora channel lookups for me. I'll land this and see where things lie on stage. self-r+

PR #4667, opened 2018-10-26
fix bug 1502038: handle bad versions

willkg:
The previous nav bar iteration handled bad versions better than the rewrite.
This fixes the rewrite so it's on par, though I think it's actually better
in that bad versions are junked at the start rather than let through.


willkg:
This looks ok. Let's test it on stage. self-r+

PR #4668, opened 2018-10-27
bug 1493687: move crontabber tables to Django

willkg:
This creates a new Django app named `cron` which has two tables in it. Then it reworks all the crontabber stuff to use those tables for bookkeeping.

This is an easier step than rewriting all of crontabber. The next step would be to implement basic crontabber stuff in Django and then start rewriting crontabber apps into Django cron apps. There are a couple of things we can do before we do that which will make that a lot easier, so I'm pushing that off until later.

To test this:

1. set up a dev environment which exercises many of the crontabber apps
2. run the crontabber container
3. check `/crontabber-state`
4. check out the new Django admin pages for `cron_job` and `cron_log`
5. run some other crontabber commands like the infamous `--list-jobs` and the quixotic `--reset`


socorro/cron/crontabber_app.py
```diff
@@ -832,13 +832,14 @@ def list_jobs(self, stream=None):
                     stream.write(' (in %s)\n' % timesince(_now, info['next_run']))
             else:
                 stream.write('none\n')
-            if info.get('last_error'):
+            if info.get('last_error') not in (None, '{}'):
```
willkg: In the previous table, `last_error` was a JSON field. In the new table, it's a text field. So we need to do this janky thing to match the possible values where `None` is if there was no entry and `{}` is an empty `last_error` value.

socorro/cron/crontabber_app.py
```diff
@@ -1013,16 +1014,18 @@ def _remember_success(
...
+                duration,
+                log_time
             ) VALUES (
+                %s,
                 %s,
                 %s,
                 %s
             )""",
-            (app_name, success_date, '%.5f' % duration),
+            (app_name, success_date, '%.5f' % duration, utc_now()),
```
willkg: Django `DateTimeField` has an `auto_add_now`, but it's implemented in Django-land and not db-land, so we need to add `utc_now()` here.

socorro/lib/transaction.py
```diff
@@ -48,9 +48,10 @@ def __call__(self, function, *args, **kwargs):
                 connection.commit()
                 return result
             except NonFatalException:
+                excinfo = sys.exc_info()
                 # This is a non-fatal exception so we don't need to reconnect
                 connection.rollback()
-                raise
+                reraise(*excinfo)
```
willkg: If `connection.rollback()` throws an error, then `raise` does the wrong thing. So this fixes it to capture the original error and reraise that.

socorro/unittest/cron/crontabber_tests_base.py
```diff
@@ -26,20 +24,12 @@
...
-        return None
-
-    @classmethod
-    def setUpClass(cls):
-        cls.tempdir = tempfile.mkdtemp()
-
-    @classmethod
-    def tearDownClass(cls):
-        if os.path.isdir(cls.tempdir):
-            shutil.rmtree(cls.tempdir)
```
willkg: None of the tests use the `tempdir` and the `shortDescription` thing was to fix a really irritating thing that nose did. Since we don't use nose and we're not using the `tempdir`, then we don't need this class at all.

socorro/unittest/cron/crontabber_tests_base.py
```diff
@@ -132,35 +115,68 @@ def setUpClass(cls):
...
-        try:
-            failed = True
-            tz, = cursor.fetchone()
-            if tz != 'UTC':
-                cursor.execute("""
-                   ALTER DATABASE %(dbname)s SET TIMEZONE TO UTC;
-                """ % cls.config.crontabber)
-            failed = False
-        finally:
-            failed and cls.conn.rollback() or cls.conn.commit()
```
willkg: We don't need this timezone-altering thing anymore.

socorro/unittest/external/postgresql/test_crontabber_state.py
```diff
@@ -42,7 +86,7 @@ def test_get(self):
                 '2013-02-09 00:16:00.893834',
                 '2012-12-24 22:27:07.316893',
                 6,
-                '{}',
+                '',
```
willkg: In the old table, `depends_on` was a `text[]` and now it's just a `text`. So this adjusts the test accordingly.

There are no dependencies in our current crontabber apps, but I figured I'd keep the `depends_on` stuff for now since it was harder to rip all of it out.

PR #4669, opened 2018-10-27
bug 1493687: fix /crontabber-state

willkg:


willkg:
This solves two problems:

First, `depends_on` has a screwy value that makes the crontabber-state JS sad. We don't have any dependencies anymore, so I tweaked the value as it's being thrown into the CrontabberState API result.

Second, because we're not copying jobs over, we end up with a `cron_job` table with jobs that have no dependencies and the JS that shows the crontabber-state page assumes there are dependencies when it's figuring out where to put everything. Instead of fixing that, I added a fake app that depends on everything.

Since we no longer have dependencies, we don't need to do `depends_on` things in the rewrite, so when we do the rewrite all these hacks can go away.

webapp-django/crashstats/crashstats/static/crashstats/js/socorro/crontabber_state.js
```diff
@@ -1,67 +1,9 @@
 /* global _, d3 */
 
-var margin = { top: 1, right: 20, bottom: 6, left: 10 };
```
willkg: I thought about redoing this file without d3, but it wasn't super valuable and I figured someone could do that later.

willkg:
This looks ok. It definitely fixes the `/crontabber-state` page and all the changes only affect that page and it's a page that not many people look at, so good enough. self-r+

PR #4670, opened 2018-10-29
fix bug 1501780: fix b99 and reduce buildhub hits

willkg:
ftpscraper would generate a product_versions record for X.0b99 which is a
"final beta". This fixes the `BetaVersionRule` to do a lookup for
`(product, build id, beta)` channel and if that returns nothing and the
version is possibly a final beta version, it will do a second lookup for
`(product, build id, release)` and if that returns something, it adds a b99
to the end of the version string.

Buildhub only stores build information for Firefox and Fennec, so we shouldn't
bother hitting it for other products. This fixes that.

This also splits up the monolithic "test_everything" into separate tests.

### DIFF
--- a/socorro/processor/mozilla_transform_rules.py
+++ b/socorro/processor/mozilla_transform_rules.py
@@ -625,6 +625,9 @@ class BetaVersionRule(Rule):
     #: If we know it's good, cache it for 24 hours because it won't change
     LONG_CACHE_TTL = 60 * 60 * 24
 
+    #: Products that are on Buildhub--we don't want to ask Buildhub about others
+    BUILDHUB_PRODUCTS = ['firefox', 'fennec', 'fennecandroid']
+
     def __init__(self, config):
         super(BetaVersionRule, self).__init__(config)
         # NOTE(willkg): These config values come from Processor2015 instance.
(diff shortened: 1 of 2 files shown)


socorro/processor/mozilla_transform_rules.py
```diff
@@ -625,6 +625,9 @@ class BetaVersionRule(Rule):
     #: If we know it's good, cache it for 24 hours because it won't change
     LONG_CACHE_TTL = 60 * 60 * 24
 
+    #: Products that are on Buildhub--we don't want to ask Buildhub about others
+    BUILDHUB_PRODUCTS = ['firefox', 'fennec', 'fennecandroid']
```
willkg: Putting "fennec" and "fennecandroid" in here because we're getting builds for two different build ids and crash reports for one of them have their product rewritten by the `ProductRewriteRule`. I want to get rid of that rule, but haven't done it, yet. In the meantime, seems prudent to handle both.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -697,34 +705,52 @@ def _get_version_data(self, product, build_id, channel):
 
         return None
 
+    def is_final_beta(self, version):
+        """Denotes whether this version could be a final beta"""
+        # NOTE(willkg): this started with 26.0 and 38.0.5 was an out-of-cycle
+        # exception
+        return version > 26.0 and (version.endswith('.0') or version == '38.0.5')
```
willkg: This logic was effectively copied from ftpscraper. The assumption is that it was right there, so it's probably right here, too.

willkg:
This looks ok. I'll land it and test it more on stage. self-r+

PR #4672, opened 2018-10-29
fix bug 1501780: fix the rc case

willkg:
Previously, we tossed on a b99, but it turns out that Buildhub had the right
version, but because we were doing a limit 1 and not sorting, it'd
intermittently come up with the right version and the wrong version. Yay!

This fixes that and ditches the whole b99 thing and now we're in a much better
place!

### DIFF
--- a/socorro/processor/mozilla_transform_rules.py
+++ b/socorro/processor/mozilla_transform_rules.py
@@ -674,10 +674,14 @@ class BetaVersionRule(Rule):
 
         session = session_with_retries(self.buildhub_api)
 
+        # Limit to 1 because otherwise we get a build for every locale and
+        # platform. Sort by target.version so that if it's an rc, we get the
+        # rc version rather than the release version.
         query = {
             'source.product': product,
             'build.id': '"%s"' % build_id,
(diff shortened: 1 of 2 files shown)


socorro/processor/mozilla_transform_rules.py
```diff
@@ -742,13 +746,13 @@ def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
                 # channel. So we check Buildhub asking about the release
-                # channel and if it's there, we tack on a b99.
+                # channel and get back an rc version.
                 version = processed_crash.get('version').strip()
                 if version and release_channel == 'beta' and self.is_final_beta(version):
                     real_version = self._get_version_data(product, build_id, 'release')
 
                     if real_version:
-                        processed_crash['version'] = real_version + 'b99'
+                        processed_crash['version'] = real_version
```
willkg: Yay! New situation is 2x better and solves bug #1295963 for sure.

willkg:
This looks good. I'll do more extensive testing on stage. self-r+

PR #4674, opened 2018-10-29
fix bug 1503033: remove "by build" optionality in topcrashers report

willkg:


willkg:
This is mostly code removal. self-r+

PR #4675, opened 2018-10-30
fix bug 1503034: change supersearch form to pull versions from supers…

willkg:
…earch

The supersearch form suggests and autocompletes versions in the version input
field. The list of versions used to come from the `ProductVersionMiddleware`
which pulled versions from the `product_versions` table for the Firefox product.

This reworks that to pull versions from a facet of crash reports for Firefox
over the last 3 months and no longer rely on the `product_versions` table.

### DIFF
--- a/webapp-django/crashstats/crashstats/utils.py
+++ b/webapp-django/crashstats/crashstats/utils.py
@@ -9,15 +9,15 @@ import re
 from past.builtins import basestring
 from collections import OrderedDict
 
-from six.moves import cStringIO
 from six import text_type
+from six.moves import cStringIO
 
 from django import http
 from django.conf import settings
 from django.core.cache import cache
 from django.utils import timezone
 
-from . import models
+from crashstats.crashstats import models
 import crashstats.supersearch.models as supersearch_models
 from socorro.lib.versionutil import (
     generate_version_key,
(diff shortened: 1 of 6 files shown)


webapp-django/crashstats/crashstats/utils.py
```diff
@@ -235,10 +238,44 @@ def get_versions_for_product(product):
...
+
+    :returns: list of version dicts sorted in reverse order or ``[]``
+
+    """
+    key = 'get_version_context_for_product:%s' % product.lower().replace(' ', '')
+    ret = cache.get(key)
+    if ret is not None:
+        return ret
+
+    versions = get_versions_for_product(product, use_cache=False)
```
willkg: This passes `use_cache=False` so that when this goes to recompute things, it's not inadvertently using a cached version of the versions which would result in double-caching and a longer period of time for the return for this function to expire out of cache.

webapp-django/crashstats/signature/tests/test_views.py
```diff
@@ -13,7 +15,24 @@
...
+
 class TestViews(BaseTestViews):
+    def setUp(self):
+        super(TestViews, self).setUp()
+        self._gvfp_patch = mock.patch('crashstats.crashstats.utils.get_versions_for_product')
+        mocked_gvfp = self._gvfp_patch.start()
+        mocked_gvfp.get.side_effect = mocked_gvfp
+
+    def tearDown(self):
+        super(TestViews, self).tearDown()
```
willkg: Oops--don't need this bit anymore.

willkg: And by "don't need this bit", I mean "there's two levels of mocking and one of them is unneeded".

I wanted to do this as a pytest fixture, but it requires us to un-unittest everything. Blech.

willkg:
This seems fine. I'll test it out on stage. It touches the following things:

* all pages that have the version dropdown nav
* supersearch page which uses the supersearch form
* signature page which uses the supersearch form

self-r+

PR #4676, opened 2018-10-30
fix bug 1503344: change java_stack_trace_full to java_stack_trace_raw

willkg:


willkg:
This has the awful effect of screwing up possibility of searching the existing `java_stack_trace_full` field. We can't just go and reprocess everything because older indexes won't have the right mapping.

Bug 1503591 covers researching how to add fields to existing indexes and building some tooling for it. I'm going to wait on landing this until I've looked into that.

willkg:
I'm closing this out. It's rotted pretty hard and this has huge consequences to land. I've got it in a local branch and can try it again another day if such need arises.

PR #4677, opened 2018-10-30
fix bug 1502477: add moz_crash_reason_raw

willkg:


webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -315,11 +315,32 @@ <h2>{{ report.product }} {{ report.version }} Crash Report [@ {{ report.signatur
...
-                                <th scope="row">MOZ_CRASH Reason</th>
-                                <td>{{ raw.MozCrashReason }}</td>
-                            </tr>
+                            {% if request.user.has_perm('crashstats.view_pii') or your_crash %}
+                                {% if report.moz_crash_reason_raw %}
+                                    <tr title="{{ fields_desc['processed_crash.moz_crash_reason_raw'] }}">
+                                        <th scope="row">MOZ_CRASH Reason (Raw)</th>
+                                        <td>{{ report.moz_crash_reason_raw }} - {{ sensitive_warning(your_crash) }}</td>
+                                    </tr>
+                                {% elif raw.MozCrashReason %}
```
willkg: We're not reprocessing all the crash reports--only the problematic ones. So until the old crash reports expire, we'll have some crash reports with a `processed.moz_crash_reason` and `processed.moz_crash_reason_raw` and some that don't. For the ones that don't, we want to degrade to the old way of doing things and show `raw.MozCrashReason`.

PR #4678, opened 2018-10-30
fix bug 1502477: add permission to search moz_crash_reason_raw

willkg:


willkg:
I ran the tests locally. This is a trivial change and there's some urgency to land it soon, so I'm going to merge it now. self-r+

PR #4679, opened 2018-10-31
Refactor transaction executor to catch and parse exceptions

lonnen:
Fixes 1503434 by refactoring the transaction executor's error handling to catch all exceptions and then compare the exception type against known cases that can be handled.

This is an expedient solution to keep progressing on py2 -> py3. A few other things I noticed but decided were out of scope: the TransactionExecutor is not really leveraged by its descendant classes here, and when the postgres crash storage is removed it can probably go away entirely. Connection contexts lack a common base class. Comments on the RMQ class attributes and variable names clearly differ from the implementations of other classes. It's unclear if `is_operational_error()` should be the whole source of truth about a retry should occur, of if the transaction executor should really own all the different cases and outcomes for different exceptions. At a future date it would be good to simplify all this with https://github.com/jd/tenacity or even rethink how robust we want this to be in light of our architecture. Process supervisors, monitoring, and queue properties may reduce or eliminate the need to be clever with error and connection handling here.

### DIFF
--- a/socorro/lib/transaction.py
+++ b/socorro/lib/transaction.py
@@ -117,38 +117,33 @@ class TransactionExecutorWithInfiniteBackoff(TransactionExecutor):
                         last_failure = sys.exc_info()
                         connection.rollback()
                         reraise(*last_failure)
-
-            except self.db_conn_context_source.conditional_exceptions as x:
-                # these exceptions may or may not be retriable the test is
-                # for is a last ditch effort to see if we can retry
-                if not self.db_conn_context_source.is_operational_exception(x):
-                    # If the logger exists, log the issue, otherwise print it
-                    # to stdout.
-                    if self.config.logger:
+            except Exception as x:
+                if type(x) in self.db_conn_context_source.conditional_exceptions:
+                    # these exceptions may or may not be retriable the test is
+                    # for is a last ditch effort to see if we can retry
+                    if not self.db_conn_context_source.is_operational_exception(x):
                         self.config.logger.critical(
                             'Unrecoverable %s transaction error',
                             self.connection_source_type,
                             exc_info=True
                         )
-                    else:
-                        print('Unrecoverable %s transaction error' % self.connection_source_type)
+                        reraise(*last_failure)
+
+                    self.config.logger.critical(
+                        '%s transaction error eligible for retry',
+                        self.connection_source_type,
+                        exc_info=True
+                    )
+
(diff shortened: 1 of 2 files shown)


willkg:
Huge +1 on everything you said in the description. The connection objects should derive from at least a mixin and they should determine the various retryability possibilities.

Regarding using tenacity or a retry library, the transaction executor enables retrying, but it also rolls back in certain error cases. I don't think tenacity can do that. I don't know if we need that feature anymore. I'm not sure whether to try to rewrite this now or just ditch it all in a processor rewrite. The crontabber apps use it, but they're about to go away soon, too.

I can look into this more soon.

socorro/lib/transaction.py
```diff
@@ -117,38 +117,33 @@ def __call__(self, function, *args, **kwargs):
...
-
-            except self.db_conn_context_source.conditional_exceptions as x:
-                # these exceptions may or may not be retriable the test is
-                # for is a last ditch effort to see if we can retry
-                if not self.db_conn_context_source.is_operational_exception(x):
-                    # If the logger exists, log the issue, otherwise print it
-                    # to stdout.
-                    if self.config.logger:
+            except Exception as x:
+                if type(x) in self.db_conn_context_source.conditional_exceptions:
```
willkg: This has sublty different behavior than what we had before. I think you need to do:

```
if self.db_conn_context_source.conditional_exceptions and isinstance(x, self.db_conn_context_source.conditional_exceptions):
```

which will correctly handle subclasses of exception classes in `conditional_exceptions`. Same thing for the other type checks.

I don't know offhand if there are cases of subclasses that need to get handled, but I'm not sure I could figure it out without knowing exactly what gets thrown where and Python is a crummy language for that kind of static analysis.

lonnen: Tried this yesterday and it isn't so straightforward. `isinstance` throws a type error for some of the classes. I'll need to dig into it a little.

willkg: Let me know if you want me to poke at it. I think I'll be switching to Python 3 migration in the next week.

lonnen: Take it if you'd like!

lonnen: Just close this out to let me know you're on it and I'll cease my end

willkg:
@lonnen and I chatted about this and some related issues. I'm going to expand this to "let's fix/redo/remove the various layers of error handling" covered by bug 1505231 and dependent bugs and tackle it in that direction. Closing this out for now.

PR #4681, opened 2018-10-31
bug 1503591: tweak create_recent_indices to have a dry run mode

willkg:
This also fixes `clear_recent_indices_app` to be a little less clunky in what
it's doing in regards to the args passed in. And it fixes `run_update_data.sh`
which was running crontabber and apps wrong.

### DIFF
--- a/docker/run_update_data.sh
+++ b/docker/run_update_data.sh
@@ -24,9 +24,9 @@ CRONTABBERCMD="./socorro/cron/crontabber_app.py"
 ./docker/run_ftpscraper_wrapper.sh
 
 # Update featured versions data based on release data
-${DC} run app shell bash -c ${CRONTABBERCMD} --reset-job=featured-versions-automatic
-${DC} run app shell bash -c ${CRONTABBERCMD} --job=featured-versions-automatic \
+${DC} run app shell ${CRONTABBERCMD} --reset-job=featured-versions-automatic
+${DC} run app shell ${CRONTABBERCMD} --job=featured-versions-automatic \
     --crontabber.class-FeaturedVersionsAutomaticCronApp.products=${PRODUCTS}
 
 # Create ES indexes for the next few weeks
-${DC} run app shell bash -c socorro/external/es/create_recent_indices_app.py
+${DC} run app shell ./socorro-cmd create_recent_indices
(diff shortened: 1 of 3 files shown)


socorro/external/es/index_creator.py
```diff
@@ -78,7 +85,10 @@ def create_socorro_index(self, es_index, mappings=None):
             mappings = SuperSearchFields(config=self.config).get_mapping()
 
         es_settings = self.get_socorro_index_settings(mappings)
-        self.create_index(es_index, es_settings)
+        if self.config.elasticsearch.dry_run:
+            print(json.dumps(es_settings, indent=2))
```
willkg: This is far more legible than `pprint`.

socorro/external/es/create_recent_indices_app.py
```diff
@@ -41,7 +47,10 @@ def main(self):
         today = date.today()
         current_monday = today - timedelta(days=today.weekday())
 
-        for week_diff in range(-2, self.config.elasticsearch_weeks_to_create):
+        past = self.config.weeks_to_create_past * -1
+        future = self.config.weeks_to_create_future
+
+        for week_diff in range(past, future):
```
willkg: Previously, you'd specify "create only 1 week", but it'd create 3 which makes no sense. I'm pretty sure that's my fault. This splits it into two arguments and lets you adjust accordingly.

docker/run_update_data.sh
```diff
@@ -24,9 +24,9 @@ CRONTABBERCMD="./socorro/cron/crontabber_app.py"
 ./docker/run_ftpscraper_wrapper.sh
 
 # Update featured versions data based on release data
-${DC} run app shell bash -c ${CRONTABBERCMD} --reset-job=featured-versions-automatic
-${DC} run app shell bash -c ${CRONTABBERCMD} --job=featured-versions-automatic \
+${DC} run app shell ${CRONTABBERCMD} --reset-job=featured-versions-automatic
+${DC} run app shell ${CRONTABBERCMD} --job=featured-versions-automatic \
```
willkg: Previously, these would just run crontabber twice because it wasn't passing in the arguments to crontabbercmd. This fixes it so that all the args get passed in so it first resets the featured-versions-automatic app and then runs it.

willkg:
Pretty sure this just affects the local dev environment. I ran `make updatedata` and that works fine. I'll wait until tests are done, but then

PR #4682, opened 2018-11-01
fix bug 1503250: fix BetaVersionRule lookups for beta/aurora

willkg:
Previously beta/aurora version lookups could get an rc version, but we
don't want that to ever happen, so this restricts those lookups to
non-rc versions.


willkg:
Looks good and matches the examples Philipp provided. self-r+

PR #4687, opened 2018-11-02
bug 1504067: first pass at removing unittest things

willkg:
This converts some classes to subclass object rather than a TestCase thing.
For those tests, I switched to using fixtures where I could.

This removes all the `self.assert*` type calls except one that checks num
queries. There's a pytest-django fixture that does the same thing, but we'd
have to convert the class first. Tricky.

This removes all the `unittest.TestCase` subclassing. There are still a bunch
that subclass Django's TestCase. That'll require more work to remove since we
have a bunch of scaffolding that is intertwined with that.


willkg:
This only affects the tests. It's a good-enough first pass. self-r+

PR #4688, opened 2018-11-02
fix bug 1504040: redo context to use Supersearch data

willkg:
This redoes the context to use supersearch data for all products
rather than just the products that don't have `product_versions` information.

This adds a `ProductVersionsMixin` for setting up the mock for setting
product versions in the tests.

This moves `BaseTestViews` into `conftest.py`. Seems like a better place
for scaffolding that a bunch of things use.


webapp-django/crashstats/crashstats/tests/conftest.py
```diff
@@ -0,0 +1,174 @@
...
+        self.content = content.strip()
+        self.status_code = status_code
+
+    @property
+    def text(self):
+        # Similar to content but with the right encoding
+        return six.text_type(self.content, 'utf-8')
+
+    def json(self):
+        return self.raw
```
willkg: This got moved from `crashstats.crashstats.tests.test_views`.

webapp-django/crashstats/crashstats/tests/conftest.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def tearDown(self):
+        self.mock_gvfp_patcher.stop()
+        super(ProductVersionsMixin, self).tearDown()
+
+    def set_product_versions(self, versions):
+        self.mock_gvfp.return_value = versions
+
+
+class BaseTestViews(ProductVersionsMixin, DjangoTestCase):
```
willkg: This got moved from `crashstats.crashstats.tests.test_views` and adjusted to handle product versions.

webapp-django/crashstats/topcrashers/tests/test_views.py
```diff
@@ -271,12 +273,6 @@ def test_with_unsupported_product(self):
         assert response.status_code == 404
 
     def test_without_any_signatures(self):
-        url = self.base_url + '?product=WaterWolf&version=19.0'
-        response = self.client.get(self.base_url, {
-            'product': 'WaterWolf',
-        })
-        assert url in response['Location']
```
willkg: This seems to test "does it redirect" which is tested elsewhere so we don't need to do it again.

willkg:
There's a bunch of cosmetic changes in this PR which is kind of annoying, but I added a note on the relevant non-cosmetic changes. Sorry about that.

This works super locally. I want to test it out on stage. self-r+

PR #4689, opened 2018-11-02
fix bug 1504281, 1492213: BetaVersionRule should degrade gracefully; add metrics

willkg:
This re-adds the `product_versions` lookup code for the case where Buildhub
doesn't know anything about that `(product, build_id, channel)`. This
is a temporary fix to deal with latency in Buildhub knowing about new
builds.

### DIFF
--- a/socorro/processor/mozilla_transform_rules.py
+++ b/socorro/processor/mozilla_transform_rules.py
@@ -6,14 +6,16 @@ import datetime
 import gzip
 import random
 import re
-from sys import maxint
 from past.builtins import basestring
+from sys import maxint
 import time
 from urllib import unquote_plus
 
+import markus
 from requests import RequestException
 import ujson
 
+from socorro.external.postgresql.dbapi2_util import execute_query_fetchall
 from socorro.lib import javautil
 from socorro.lib import raven_client
 from socorro.lib.cache import ExpiringCache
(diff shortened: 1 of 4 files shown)


willkg:
Gah--this is super tricky to test, but I tested it out locally using this crash:

http://crash-stats.mozilla.com/report/index/008ac878-ff8f-4a1e-9185-9b5280181102

That crash is for build id 20181101155334 which Buildhub does not currently have (it might at some point soon, so then this is hard to test again). Running ftpscraper will add entries to `product_version_builds` for that build to resolve it to `64.0b6`.

Without these changes, Buildhub shrugs and the crash report ends up with `64.0b0`. With these changes, Buildhub shrugs, but it's in `product_version_builds` and so it gets fixed to `64.0b6`.

willkg:
Looks good and works locally. I'll test it further on -stage. self-r+

sciurus:
So the processor talks to pg again?

willkg:
Maybe. This is a temporary fix. So if this did go in, it'd be for a couple of weeks. The alternative is to bring back the versionstring api in the web app. That’s a lot more code and work, but doable.

You and Miles were out so I figured I’d land the easier option now and test it on stage and possibly back it out if there were infra problems and talk to you about it on Monday.

willkg:
One of the tricky things here was that I wanted to test with a buildid that buildhub didn't have which is a temporary problem. 20181101155334 is in Buildhub now, so I can't test with it anymore.

PR #4690, opened 2018-11-05
bug 1504281: add admin page for checking buildhub status

willkg:
This adds an admin page that goes through what Socorro knows in
`product_versions` and checks what Buildhub knows. This will help us to suss out
missing data in Buildhub.


webapp-django/crashstats/manage/admin.py
```diff
@@ -218,9 +220,80 @@ def debug_view(request):
...
     context = {}
     api = ProductsMiddleware()
     context['products'] = api.get()['hits']
     context['title'] = 'Products'
 
     return render(request, 'admin/products.html', context)
+
+
+@superuser_required
+def buildhub_check(request):
```
willkg: This is "quick and dirty" and copies some code from the BetaVersionRule. It's temporary and will get removed when we remove the `product_versions` table and all its friends. At that point, we can replace it with a page that checks recent crash reports with "b0" to determine why they're stinky. (ha! get it--b0! stinky!)

willkg:
This is fine. It's an admin view that's temporary, so it's low risk. self-r+

PR #4691, opened 2018-11-06
bug 1504281: reverse order of versions in buildhub_check admin page

willkg:


willkg:
This is a minor change to an admin page that isn't tested. self-r+ and landing before CI.

PR #4693, opened 2018-11-06
fix bug 1505030: switch BetaVersionRule to use Buildhub ES API

willkg:
Buildhub has a Kinto API and an Elasticsearch API. This switches the
BetaVersionRule to use the Elasticsearch API which is MUCH faster
and will work with Buildhub2.

### DIFF
--- a/socorro/processor/mozilla_transform_rules.py
+++ b/socorro/processor/mozilla_transform_rules.py
@@ -4,9 +4,10 @@
 
 import datetime
 import gzip
+import json
+from past.builtins import basestring
 import random
 import re
-from past.builtins import basestring
 from sys import maxint
 import time
 from urllib import unquote_plus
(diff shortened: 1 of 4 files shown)


socorro/processor/mozilla_transform_rules.py
```diff
@@ -770,46 +771,68 @@ def _get_real_version(self, product, build_id, channel):
...
-                # For aurora and beta channel lookups, we want a non-"rc"
-                # version, but we can't specify that. So we ask for all the
-                # versions and winnow out the rc ones on our own.
-                #
-                # FIXME(willkg): This is silly and it'd be better to filter
-                # it out on Buildhub.
-                hits = [
-                    hit for hit in hits
-                    if 'rc' not in hit['target']['version']
-                ]
```
willkg: Glad to get rid of this. Blech.

socorro/unittest/processor/test_mozilla_transform_rules.py
```diff
@@ -1619,38 +1739,26 @@ def test_rc(self):
...
+                                    }
+                                }
+                            }]
                         }
-                    ]
-                }
-            )
+                    })
+
+            req_mock.post(self.API_URL, text=payload_callback)
```
willkg: This one was tricky. We can't differentiate between the two posts enough to do `req_mock.post` for each one, so we have on mock call that has a payload callback that checks the request and returns the appropriate response. Then we later assert that the Buildhub query requests have the right shape and happen in the correct order.

webapp-django/crashstats/manage/admin.py
```diff
@@ -271,31 +273,57 @@ def buildhub_check(request):
...
+                    }
+                }
+            },
+            'size': 1
         }
+
         buildhub_start = time.time()
-        resp = session.get(buildhub_api, params=query)
+        resp = session.post(buildhub_api, data=json.dumps(es_query))
         buildhub_time = buildhub_time + (time.time() - buildhub_start)
```
willkg: This change drops this page from taking 50+ seconds to run (of which all of that is taken up by querying Buildhub) to taking 1 second to run. It's pretty incredible.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -770,46 +771,68 @@ def _get_real_version(self, product, build_id, channel):
...
+        else:
+            # If it's the beta channel, we want to exclude rc versions
+            es_query['query']['bool']['must_not'] = {
+                'wildcard': {
+                    'target.version': '*rc*'
+                }
+            }
 
-        resp = session.get(self.buildhub_api, params=query)
+        resp = session.post(self.buildhub_api, data=json.dumps(es_query, sort_keys=True))
```
willkg: We sort the keys here since it makes it easier to test. It's a small dict, so I think the performance hit is pretty meh.

willkg:
This works pretty well for me locally. It'll be easier to test in stage with a wider range of inputs. self-r+

PR #4694, opened 2018-11-06
bug 1504281: add fennec to buildhub_check page

willkg:
This adds Fennec to the `buildhub_check` page so we can see what's going
on there, too.


willkg:
self-r+

PR #4696, opened 2018-11-07
fix bug 1505233: clean up Rule error handling

willkg:
This moves `Rule` from `socorro.lib.transform_rules` to `socorro.processor.rules.base`, guts it removing the error handling and some unused functionality, and then updates all the subclasses accordingly.


socorro/processor/general_transform_rules.py
```diff
@@ -4,26 +4,18 @@
...
-    def _action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
-
-        processed_crash.crash_id = raw_crash.uuid
-        processed_crash.uuid = raw_crash.uuid
-
-        return True
+    def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
+        if 'uuid' in raw_crash:
+            processed_crash['crash_id'] = raw_crash['uuid']
+            processed_crash['uuid'] = raw_crash['uuid']
```
willkg: The `uuid` gets added by the collector to the raw crash. We should never be processing crashes that don't have a `uuid`, so ... I'm not really sure why we would end up here ever.

socorro/processor/rules/base.py
```diff
@@ -0,0 +1,67 @@
...
+
+        """
+        rule_name = self.__class__.__name__
+        with metrics.timer('act.timing', tags=['rule:%s' % rule_name]):
+            if self.predicate(raw_crash, raw_dumps, processed_crash, processor_meta_data):
+                self.action(raw_crash, raw_dumps, processed_crash, processor_meta_data)
+
+    def close(self):
+        # FIXME(willkg): see if any of the rules use .close()
+        self.config.logger.debug('null close on rule %s', self.__class__)
```
willkg: I can nix this. There is a rule that has a `.close()` in it.

willkg:
The tests pass, linting passes, and I ran some crashes through it and it seems fine. There are a few differences here:

1. `action` and `predicate` may have been swallowing errors letting processing continue which are no longer swallowed
2. because of that, it's possible that bugs in the code will throw errors that weren't previously being thrown
3. because of that, it's possible that when this gets to stage, it'll start throwing a lot of errors and the processing queue will back up and things might go down

It'd be super if that didn't happen in production, so once this lands in stage, I'm going to let it hang out for a couple of days and fix anything that pops up. It's entirely possible that not much happens because @peterbe and @adngdb did a pass at fixing errors in rules a couple of years ago and that code has been in prod for ages. So hopefully this is a whole lot of worry for nothing.

willkg:
The code changes look ok. This is a lot of code removal, but it's mostly things that we don't use at all. There's some cosmetic changes in here, too.

The big change here is that `Rule` no longer catches and surfaces errors--that's now done by `Processor2015.process_crash`. It'll also add notes to the processed crash making it easier for us to identify and reprocess partially processed crashes.

socorro/processor/processor_2015.py
```diff
@@ -261,51 +269,47 @@ def process_crash(self, raw_crash, raw_dumps, processed_crash):
...
-                str(exception),
-                exc_info=True
-            )
-            processor_meta_data.processor_notes.append(
-                'unrecoverable processor error: %s' % exception
-            )
+                # NOTE(willkg): notes are public, so we can't put exception
+                # messages in them
+                processor_meta_data.processor_notes.append(
+                    'rule %s failed: %s' % (rule.__class__.__name__, exc.__class__.__name__)
```
willkg: This creates a line like "rule OSInfoRule failed: KeyError". That gives me something to search for to reprocess crashes that had issues but shouldn't reveal any PII in a public field.

willkg: The thing that calls `process_crash` also has error handling, but that error handling is tied in with processing the crash and saving it. I didn't want to further complicate that. Further, if we handled errors there, processing would halt on the first error thrown. That'd be great if we had a system where crashes that couldn't get through processing were put aside in a separate bin. But we don't have that. Our options are:

1. continue processing despite errors and end up with a partially processed crash
2. throw it back in the rabbitmq queue
3. drop it altogether and never process it again

We can't do option 3 because then we have crashes we've saved to S3 but don't have indexed and don't know anything about. Option 2 could lead to the queue filled with bad crashes and the processor grinding to a halt. Option 1 is unenthusing and could lead to confusion during analysis.

I'm going for option 1. The hope is that we'll catch the errors in sentry and fix the problems and push the fixes out. Since we're adding notes to the crashes in question, we can reprocess them once the problem is fixed.

We'll see how that goes.

socorro/unittest/processor/test_processor_2015.py
```diff
@@ -62,88 +108,3 @@ def test_process_crash_existing_processed_crash(self):
...
-            DotDict()
-        )
-
-        assert processed_crash.success
-        expected = (
-            'dwight; Processor2015; SigTrunc: signature truncated due to length'
-        )
-        assert processed_crash.processor_notes == expected
-        assert processed_crash.signature.startswith('shutdownhang')
-        assert len(processed_crash.signature) == 255
```
willkg: I have no idea what this was testing. Sure seemed like it was testing SigTrunc in signature generation. That's tested (better) elsewhere, so I nixed this.

willkg:
I'm going to land this now and watch stage for a while. self-r+

lonnen:
post-landing r+, this looks great. excited for it.

willkg:
Thank you!

PR #4697, opened 2018-11-07
bug 1406703: add comments for what's remaining

willkg:
This adds commented out lines for what remains to get working in Python 3.
This will make it easier to make sure we've got everything.


willkg:
CI doesn't test Python 3 stuff so this has no effect on the tests or stage or prod or anything. self-r+

PR #4698, opened 2018-11-07
fix bug 1469718: convert socorro/external/fs to Python 3

willkg:
This gets all the fs crashstorage tests working in Python 3.

### DIFF
--- a/socorro/external/fs/crashstorage.py
+++ b/socorro/external/fs/crashstorage.py
@@ -2,26 +2,22 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+from contextlib import contextmanager, closing
 import gzip
 import json
 import os
 
-from contextlib import contextmanager, closing
-
-try:
-    from cStringIO import StringIO
-except ImportError:
-    from StringIO import StringIO
-
 from configman import Namespace
+from six import BytesIO
+
 from socorro.external.crashstorage_base import (
     CrashStorageBase,
     CrashIDNotFound,
     FileDumpsMapping,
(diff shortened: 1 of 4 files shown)


willkg:
The fs crashstorage classes are used as the default for `FetchTransformSaveApp` and I'm pretty sure they're not used in our local dev environment or server environments. At some point, we might want to just remove these altogether.

socorro/external/fs/crashstorage.py
```diff
@@ -2,26 +2,22 @@
...
 
-from contextlib import contextmanager, closing
-
-try:
-    from cStringIO import StringIO
-except ImportError:
-    from StringIO import StringIO
-
 from configman import Namespace
+from six import BytesIO
```
willkg: This is tricky. In Python2, this ends up being StringIO and in Python3, this ends up being BytesIO.

socorro/external/fs/crashstorage.py
```diff
@@ -176,16 +171,18 @@ def _save_files(self, crash_id, files):
                 # probably already created, ignore
                 pass
 
-            for fn, contents in files.iteritems():
+            for fn, contents in files.items():
```
willkg: I ditched all the `iteritems` for `items`. In all of these cases, the dicts are really small, so using `items` isn't a big deal.

socorro/external/fs/crashstorage.py
```diff
@@ -176,16 +171,18 @@ def _save_files(self, crash_id, files):
...
                 with open(os.sep.join([parent_dir, fn]), 'wb') as f:
+                    print(repr(contents))
```
willkg: Oops--this should get removed.

willkg:
self-r+

PR #4699, opened 2018-11-07
fix bug 1469718: convert socorro/external/crashstorage_base.py to Python 3

willkg:


socorro/external/crashstorage_base.py
```diff
@@ -373,55 +365,55 @@ def ack_crash(self, crash_id):
...
             output.append(self.args[0])
         for e in self.exceptions:
-            output.append(repr(e[1]).encode('ascii', 'ignore'))
+            output.append(repr(e[1]))
+        if six.PY2:
+            # If this is Python2, encode everything into strings before joining
+            output = [
+                part.encode('ascii', 'ignore')
+                for part in output
+            ]
```
willkg: In Python 2, some of these are unicode. So we convert everything to ascii-encoded strings before we can join via a string.

In Python 3, there ain't no problems here.

socorro/external/crashstorage_base.py
```diff
@@ -544,7 +536,7 @@ def close(self):
                              exceptions raised by the subordinate storage
                              systems"""
         storage_exception = PolyStorageError()
-        for a_store in self.stores.itervalues():
+        for a_store in self.stores.values():
```
willkg: These are small dicts, so we don't need to call `itervalues`.

socorro/unittest/external/fs/test_fspermanentstorage.py
```diff
@@ -114,8 +114,8 @@ def test_get_raw_dump(self):
     def test_get_raw_dumps(self):
         self._make_test_crash()
         expected = MemoryDumpsMapping({
-            'foo': 'bar',
-            self.fsrts.config.dump_field: 'baz'
+            'foo': b'bar',
+            self.fsrts.config.dump_field: b'baz'
```
willkg: The values here are bytes--not strings.

socorro/unittest/external/test_crashstorage_base.py
```diff
@@ -206,17 +188,11 @@ def save_raw_crash(self, raw_crash, dumps, crash_id):
...
                 return values.pop(0)
 
-            crashstorage = MyCrashStorageTest(
-                config,
-                quit_check_callback=fake_quit_check
-            )
+            crashstorage = MyCrashStorage(config, quit_check_callback=fake_quit_check)
 
-            with mock.patch("__builtin__.open") as open_mock:
+            with mock.patch("socorro.external.crashstorage_base.open") as open_mock:
```
willkg: Python 3 doesn't have a `__builtins__`. I think there's another way to mock builtins, but instead I'm just abusing the `socorro.external.crashstorage_base` namespace which has the open we need to mock in it.

socorro/unittest/external/test_crashstorage_base.py
```diff
@@ -668,18 +619,13 @@ def test_benchmarking_crashstore(self):
 
 
 class TestDumpsMappings(TestCase):
-
     def test_simple(self):
         mdm = MemoryDumpsMapping({
-            'upload_file_minidump': 'binary_data',
-            'moar_dump': "more binary data",
+            'upload_file_minidump': b'binary_data',
+            'moar_dump': b'more binary data',
```
willkg: The values need to be bytes.

willkg:
self-r+

PR #4701, opened 2018-11-08
fix bug 1505797: remove socorro's DotDict

willkg:
This removes one of the DotDicts from the codebase. Now there's only the
configman DotDict.

One architectural change this required was in crashstorage. Previously,
`save_raw_and_processed` would take DotDicts for the `raw_crash` and
`processed_crash`. Now the processor converts the two crash structures
into dicts before calling `save_raw_and_processed`.

### DIFF
--- a/socorro/external/crashstorage_base.py
+++ b/socorro/external/crashstorage_base.py
@@ -10,35 +10,13 @@ import datetime
 import collections
 import copy
 import os
-from past.builtins import basestring
 import sys
 
 from configman import Namespace, RequiredConfig
 from configman.converters import class_converter, str_to_list
-from configman.dotdict import DotDict as ConfigmanDotDict
+from configman.dotdict import DotDict
 import six
 import markus
-
-from socorro.lib.util import DotDict as SocorroDotDict
-
-
-def socorrodotdict_to_dict(sdotdict):
-    """Takes a socorro.lib.util.DotDict and returns a dict
-
-    This does a complete object traversal converting all instances of the
-    things named DotDict to dict so it's deep-copyable.
-
-    """
-    def _dictify(thing):
-        if isinstance(thing, collections.Mapping):
-            return dict([(key, _dictify(val)) for key, val in thing.items()])
-        elif isinstance(thing, basestring):
-            return thing
-        elif isinstance(thing, collections.Sequence):
-            return [_dictify(item) for item in thing]
-        return thing
-
-    return _dictify(sdotdict)
 
 
 class MemoryDumpsMapping(dict):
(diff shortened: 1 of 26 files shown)


willkg:
One interesting thing here is that after I got the code passing linting and testing for both Python 2 and 3, I ran some crashes through processing and the processor died epicly. I'm guessing either there isn't much coverage or the relevant tests are mocked out they're not effective.

I think running the processor should have flushed out any issues.

socorro/external/crashstorage_base.py
```diff
@@ -10,36 +10,14 @@
...
-    def _dictify(thing):
-        if isinstance(thing, collections.Mapping):
-            return dict([(key, _dictify(val)) for key, val in thing.items()])
-        elif isinstance(thing, basestring):
-            return thing
-        elif isinstance(thing, collections.Sequence):
-            return [_dictify(item) for item in thing]
-        return thing
-
-    return _dictify(sdotdict)
```
willkg: I renamed this to `dotdict_to_dict` and move it to `socorro.lib.util`.

socorro/external/crashstorage_base.py
```diff
@@ -158,7 +136,7 @@ def redact(self, a_mapping):
                 for a_sub_key in sub_keys[:-1]:  # step through the subkeys
                     sub_mapping = sub_mapping[a_sub_key.strip()]
                 del sub_mapping[sub_keys[-1]]
-            except KeyError:
+            except (AttributeError, KeyError):
```
willkg: configman's DotDict can raise an `AttributeError` or a `KeyError` depending on things.

socorro/external/crashstorage_base.py
```diff
@@ -584,26 +563,14 @@ def save_processed(self, processed_crash):
     def save_raw_and_processed(self, raw_crash, dump, processed_crash, crash_id):
         storage_exception = PolyStorageError()
 
-        # Later we're going to need to clone this per every crash storage
-        # in the loop. But, to save time, before we do that, convert the
-        # processed crash which is a SocorroDotDict into a pure python
-        # dict which we can more easily copy.deepcopy() operate on.
-        processed_crash_as_dict = socorrodotdict_to_dict(processed_crash)
-        raw_crash_as_dict = socorrodotdict_to_dict(raw_crash)
-
```
willkg: I changed the architecture so `raw_crash` and `processed_crash` are dicts now. So we don't need to convert them or do any shenanigans with copying.

willkg:
The code changes themselves look fine. I can't tell if some other code was relying on the properties of `socorro.lib.util.DotDict` in some way that isn't apparent in the tests or minor processing.

Having said that, everything I've seen so far seems fine and I'm not sure what else to test or look at. Having it sit on stage for the weekend will be good.

self-r+

PR #4702, opened 2018-11-08
Add rust_begin_unwind to the irrelevant list

jrmuizel:
This should give us useful signature on crashes like: https://crash-stats.mozilla.com/report/index/f6ee9128-a2a5-4e5d-a634-5e61c0181108 and will should fix https://bugzilla.mozilla.org/show_bug.cgi?id=1505954

### DIFF
--- a/socorro/signature/siglists/irrelevant_signature_re.txt
+++ b/socorro/signature/siglists/irrelevant_signature_re.txt
@@ -76,6 +76,7 @@ org\.mozilla\.f.*-\d\.apk@0x
 panic_abort::
 PR_WaitCondVar
 RaiseException
+rust_begin_unwind
 rtc::FatalMessage
 RtlpAdjustHeapLookasideDepth
 RtlSleepConditionVariableCS


staktrace:
Might be better to put the panic_bounds_check frame on the sentinel list. See also https://bugzilla.mozilla.org/show_bug.cgi?id=1500401

willkg:
I'd add `core::panicking::panic_bounds_check` to the sentinel list and see where that puts us.

@jrmuizel Seems like you made the edit in the GitHub interface which is fine, but probably makes it harder for you to adjust it. Want me to close this PR out and try theme and variations in a new PR until we get a solution that works best?

jrmuizel:
I made https://github.com/mozilla-services/socorro/pull/4703 you can do with either pr as you see fit.

willkg:
I think rather than do both things, it's better to just do sentinels for these cases. I'm going to close this out in favor of pr #4703.

PR #4705, opened 2018-11-10
fix bug 1506228: fix socorro/unittest/cron tests to work in python 3

willkg:
This also redoes how the tests acquire CronTabberApp configuration because
that was a leftover mess from when I vendored crontabber.


willkg:
NB: I had most of this done on Friday. All I did just now was fix a thing, add a commit comment, and create a PR. I'm waiting for sweaters to dry.

socorro/cron/jobs/ftpscraper.py
```diff
@@ -290,9 +283,9 @@ def download(self, url):
         assert response.status_code == 200, response.status_code
         data = response.content
 
-        if is_caching:
-            with open(fn, 'w') as fp:
-                fp.write(data)
```
willkg: I took out the caching bits that I wrote a while ago, but forgot this part.

socorro/cron/jobs/ftpscraper.py
```diff
@@ -290,9 +283,9 @@ def download(self, url):
...
+        # All the things we're downloading should be text files
+        if not isinstance(data, six.text_type):
+            data = data.decode('utf-8')
```
willkg: This fixes the problem in Python 3 where data is a bytes and the we go do string things on it and it gets very sad.

socorro/unittest/cron/crontabber_tests_base.py
```diff
@@ -7,72 +7,50 @@
...
+    # Be sure to include defaults
+    overrides.insert(0, CronTabberApp.config_defaults)
+
+    return ConfigurationManager(
+        [crontabber_config],
+        values_source_list=overrides,
+        app_name='test-crontabber',
+        app_description='',
+        argv_source=[]
+    )
```
willkg: We had a more complicated version of this in the code in 3 different places. I merged them all together into this one and reduced it down a little.

We could probably improve this further, but ultimately, I want to "rewrite" crontabber in Django so this will all go away.

willkg:
self-r+

PR #4707, opened 2018-11-12
bug 1493687: remove /api/CrontabberState and /crontabber-state

willkg:
This removes the CrontabberState API endpoint and the `/crontabber-state`
view. We no longer support `depends_on` relationships between jobs, so
we don't need `/crontabber-state` anymore. This rewrites the one view
that looks at crontabber jobs to look at the database directly, so
we no longer need the API endpoint, either.

This also fixes the API tests to use a new `NoOpMiddleware` for testing
API infrastructure. That saves me the effort of rewriting those tests
every time I remove another API endpoint.


webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -13,17 +12,16 @@
 
 from crashstats.crashstats.models import (
     BugAssociation,
+    NoOpMiddleware,
```
willkg: I should have done this ages ago. It would have saved me a bunch of time. :(

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -84,65 +82,87 @@ def test_base_classes_raise_not_found(self):
...
+
+            # But even being signed in has a limit.
+            authenticated_limit = 6  # see above mentioned settings override
+            assert authenticated_limit > current_limit
+            for i in range(authenticated_limit * 2):
+                response = self.client.get(url, {'product': 'good'})
+            # Even if you're authenticated - sure the limit is higher -
+            # eventually you'll run into the limit there too.
+            assert response.status_code == 429
+
```
willkg: I moved all the api infrastructure tests to the same part of the class and the middleware tests after it. So this code is the same as what we had below, but it was in a different part of the file. The only change is that it now uses the `NoOpMiddleware`.

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -169,18 +189,6 @@ def mocked_get(*args, **k):
...
-        with MetricsMock() as metrics_mock:
-            response = self.client.get(url)
-        assert response.status_code == 200
-        assert metrics_mock.has_record(
-            fun_name='incr',
-            stat='webapp.api.pageview',
-            value=1,
-            tags=['endpoint:apiCrontabberState']
-        )
-
```
willkg: Moved upwards.

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -537,106 +545,11 @@ def mocked_supersearch_get(**params):
...
-            # But even being signed in has a limit.
-            authenticated_limit = 6  # see above mentioned settings override
-            assert authenticated_limit > current_limit
-            for __ in range(authenticated_limit * 2):
-                response = self.client.get(url)
-            # Even if you're authenticated - sure the limit is higher -
-            # eventually you'll run into the limit there too.
-            assert response.status_code == 429
-
```
willkg: This was moved upwards.

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -767,27 +680,6 @@ def mocked_supersearch_get(**params):
...
-        SuperSearch.implementation().get.side_effect = mocked_supersearch_get
-
-        url = reverse('api:model_wrapper', args=('SuperSearch',))
-        response = self.client.get(url)
-        assert response.status_code == 400
-        assert 'That was a bad thing to do' in response.content
-        response = self.client.get(url, {'product': 'foobaz'})
-        assert response.status_code == 400
-        assert 'foobaz' in response.content
```
willkg: I didn't understand this test and the mocking seemed like it created a mythical universe where up is down and left is right, so I moved it upwards, nixed the mocking, and redid it to verify what happens when required parameters were missing and what happens if `BadArgumentError` is thrown.

webapp-django/crashstats/settings/bundles.py
```diff
@@ -316,13 +310,6 @@
         ),
         'output_filename': 'js/api-documentation.min.js',
     },
-    'crontabber_state': {
-        'source_filenames': (
-            'underscore/underscore-min.js',
```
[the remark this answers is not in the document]
willkg: https://bugzilla.mozilla.org/show_bug.cgi?id=1506722

willkg:
This seems fine. I'll land it tomorrow when I can watch stage and fix errors.

willkg:
Looks good. self-r+

PR #4708, opened 2018-11-13
Remove webapp-django/bin/bootstrap.sh

willkg:
I don't know what era this belongs to, but it's not used now.


willkg:
This is pretty cosmetic. self-r+

PR #4709, opened 2018-11-13
fix bug 1506781: fix silent ujson errors

willkg:
minidump-stackwalker was failing with exit code 134 because there was no
raw crash data to work on. There was no raw crash data to work on because
ujson and json handle parsing errors differently. Particularly, ujson
would shrug when trying to JSON serialize a configman DotDict. That's
really unhelpful.

This fixes the underlying issue (need to convert configman DotDict
to a dict before serializing to JSON), purges ujson from the codebase,
and fixes some tests that were testing what happens if the constants
of the universe were different which will never be the case.

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -93,8 +93,6 @@ requests==2.20.0 \
 statsd==3.3.0 \
     --hash=sha256:c610fb80347fca0ef62666d241bce64184bd7cc1efe582f9690e045c25535eaa \
     --hash=sha256:e3e6db4c246f7c59003e51c9720a51a7f39a396541cb9b147ff4b14d15b5dd1f
-ujson==1.35 \
-    --hash=sha256:f66073e5506e91d204ab0c614a148d5aa938bdbf104751be66f8ad7a222f5f86
 python-dateutil==2.7.5 \
     --hash=sha256:063df5763652e21de43de7d9e00ccf239f953a832941e37be541614732cdfc93 \
     --hash=sha256:88f9287c0174266bb0d8cedd395cfba9c58e87e5ad86b2ce58859bc11be3cf02
--- a/socorro/processor/breakpad_transform_rules.py
+++ b/socorro/processor/breakpad_transform_rules.py
@@ -1,7 +1,7 @@
 import os
 import subprocess
 import threading
-import ujson
+import json
 import tempfile
 
 from contextlib import contextmanager, closing
@@ -9,9 +9,10 @@ from collections import Mapping
 
 from configman import Namespace
 from configman.converters import str_to_list
-from configman.dotdict import DotDict as DotDict
+from configman.dotdict import DotDict
 
 from socorro.lib.converters import change_default
+from socorro.lib.util import dotdict_to_dict
 from socorro.processor.rules.base import Rule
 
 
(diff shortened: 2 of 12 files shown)


requirements/default.txt
```diff
@@ -93,8 +93,6 @@ requests==2.20.0 \
 statsd==3.3.0 \
     --hash=sha256:c610fb80347fca0ef62666d241bce64184bd7cc1efe582f9690e045c25535eaa \
     --hash=sha256:e3e6db4c246f7c59003e51c9720a51a7f39a396541cb9b147ff4b14d15b5dd1f
-ujson==1.35 \
-    --hash=sha256:f66073e5506e91d204ab0c614a148d5aa938bdbf104751be66f8ad7a222f5f86
```
willkg: BEGONE UJSON!

socorro/processor/breakpad_transform_rules.py
```diff
@@ -73,50 +74,42 @@ class ExternalProcessRule(Rule):
...
 
     def _interpret_external_command_output(self, fp, processor_meta):
+        data = fp.read()
         try:
-            return ujson.load(fp)
+            return json.loads(data)
         except Exception as x:
+            self.config.logger.error(
+                '%s non-json output: "%s"' % (self.config.command_pathname, data[:100])
+            )
```
willkg: I dropped ujson here and also moved the `fp.read()` to a separate line. That lets us spit out some of it to the logs if it throws an error in `json.loads` which makes it muuuuuch more likely to be seen when doing development.

socorro/processor/breakpad_transform_rules.py
```diff
@@ -248,13 +232,10 @@ class BreakpadStackwalkerRule2015(ExternalProcessRule):
...
             self.config.temporary_file_system_storage_path,
-            "%s.%s.TEMPORARY.json" % (
-                crash_id,
-                threading.currentThread().getName()
-            )
+            '%s.%s.TEMPORARY.json' % (crash_id, threading.currentThread().getName())
         )
         with open(file_pathname, "w") as f:
-            ujson.dump(raw_crash, f)
+            json.dump(dotdict_to_dict(raw_crash), f)
```
willkg: This was the problematic line.

Last week, `raw_crash` would have been a Socorro DotDict. That subclasses dict, so `ujson.dump` would do the right thing with it. Then I purged Socorro DotDict from the system so then `raw_crash` here becomes a configman DotDict which does not subclass dict and thus `ujson.dump` doesn't serialize the data right.

This is exacerbated by ujson not throwing an error--it just serializes it weird:

```
app@socorro:/app$ python
Python 2.7.14 (default, Dec 12 2017, 06:55:29) 
[GCC 4.9.2] on linux2
>>> import ujson
>>> from configman.dotdict import DotDict
>>> d = DotDict({'foo': 'bar'})
>>> ujson.dumps(d)
'["foo"]'
>>> 
```

Further, this processor rule swallows errors, so they don't show up in logs or sentry or anywhere we're monitoring. Ugh.

So the fix here is to run the `raw_crash` through `dotdict_to_dict` which gets a Python dict and that serializes just fine. On top of that, I added some logging and I purged ujson from the system because it's frustrating and problematic to have two json serialization systems in play that do different things.

socorro/unittest/processor/test_breakpad_transform_rules.py
```diff
@@ -362,32 +362,29 @@ def test_external_fails_2(self, mocked_subprocess_module):
         processor_meta = get_basic_processor_meta()
 
         mocked_subprocess_handle = mocked_subprocess_module.Popen.return_value
-        mocked_subprocess_handle.stdout.read.return_value = int
+        # This data will fail in json.loads and throw an error
+        mocked_subprocess_handle.stdout.read.return_value = '{ff'
```
willkg: Previously, this would mock a file type's `read` method to return the `int` type. That'll never happen plus it broke my logging, so I changed it to a string that's not valid JSON.

socorro/unittest/processor/test_breakpad_transform_rules.py
```diff
@@ -362,32 +362,29 @@ def test_external_fails_2(self, mocked_subprocess_module):
...
 class TestBreakpadTransformRule2015(TestCase):
-
     def get_basic_config(self):
         config = get_basic_config()
         config.dump_field = 'upload_file_minidump'
         config.command_line = BreakpadStackwalkerRule2015.required_config.command_line.default
         config.kill_timeout = 5
         config.command_pathname = '/bin/stackwalker'
-        config.symbols_urls = 'https://localhost'
+        config.symbols_urls = ['https://example.com']
```
willkg: This was a bug from a long while ago. `symbols_urls` needs to be a list of strings. Otherwise you end up with a command line that has `--symbol-url=h --symbolurl=t --symbolurl=t ...`

willkg:
I went through this and annotated the interesting bits. This looks good to me. The purging of ujson might surface other issues, but I don't think so--it wasn't used that much.

I'm going to self-r+ this and test it more thoroughly on stage.

PR #4711, opened 2018-11-14
fix bug 1507190: mdsw metrics

willkg:
This adds metrics and logging for mdsw results because that's a good thing to have.

### DIFF
--- a/socorro/processor/breakpad_transform_rules.py
+++ b/socorro/processor/breakpad_transform_rules.py
@@ -30,24 +30,21 @@ class CrashingThreadRule(Rule):
 
         try:
             processed_crash.truncated = (
-                processed_crash['json_dump']
-                ['crashing_thread']['frames_truncated']
+                processed_crash['json_dump']['crashing_thread']['frames_truncated']
             )
         except KeyError:
             processed_crash.truncated = False
 
         try:
             processed_crash.address = (
-                processed_crash['json_dump']
-                ['crash_info']['address']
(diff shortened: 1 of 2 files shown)


socorro/processor/breakpad_transform_rules.py
```diff
@@ -261,13 +261,23 @@ def _execute_external_process(self, command_line, processor_meta):
...
+            processor_meta.processor_notes.append(msg)
+            self.config.logger.warning(msg)
 
         elif return_code != 0 or not stackwalker_data.success:
-            processor_meta.processor_notes.append(
-                'MDSW failed on "%s": %s' % (command_line, stackwalker_data.mdsw_status_string)
-            )
+            msg = 'MDSW failed with %s: %s' % (return_code, stackwalker_data.mdsw_status_string)
+            processor_meta.processor_notes.append(msg)
+            self.config.logger.warning(msg)
```
willkg: I changed this. Previously, it would put the command line in the processor notes. The command line is mostly static now since it's defined in code. Also, it seems nuts to me to put the command line in the processor notes which I think are public to everyone. Anyhow, the return code is much more helpful, so I switched this around to add the return code and status string to processor notes and also log them.

Logging is *SUPER* helpful for local development because I can see/grep it without having to fiddle through web pages.

willkg:
This looks good locally. I'll land it and check it on stage and build Datadog graphs and iterate further if needed. self-r+

PR #4712, opened 2018-11-14
bug 1506907: implement Product model

willkg:
This creates a Django model for products, creates a data migration to move
product data from the old table to the new one, and reimplements things to
use the new table.

We have to leave the old table in place because it's a foreign key of
the old `product_versions` table and some other tables. All that will go
away in a future commit.


webapp-django/crashstats/base/jinja2/crashstats_base.html
```diff
@@ -44,7 +44,7 @@ <h1>Product Navigation</h1>
             <li>
                 <label for="products_select" class="visually-hidden">Select Product</label>
                 <select id="products_select">
-                {% for product_name, product_info in products.items() %}
+                {% for product_name in products %}
```
willkg: This is a flat list of product names now.

webapp-django/crashstats/crashstats/fixtures/products.json
```diff
@@ -0,0 +1,83 @@
...
+  {
+    "model": "crashstats.product",
+    "pk": 9,
+    "fields": {
+      "product_name": "Fennec",
+      "sort": 100,
+      "is_active": true
+    }
+  }
+]
```
willkg: I don't know if it makes sense to have the fixture have all the current products, but I didn't want to think about it much so I just did it this way.

This only affects the local development environment.

webapp-django/crashstats/crashstats/jinja2/crashstats/home.html
```diff
@@ -3,7 +3,7 @@
 {% block content %}
 <div id="mainbody">
     <div class="panel product-grid">
-    {% for product_name, product_info in products.items() %}
+    {% for product_name in products %}
```
willkg: `products` is now a flat list of product names sorted by the `sort` column.

webapp-django/crashstats/crashstats/migrations/0010_1506907_product_migration.py
```diff
@@ -0,0 +1,59 @@
...
+        product = dict(zip(columns, row))
+        # NOTE(willkg): In the old table, we used sort=-1 to denote inactive
+        # products. In the new table, we have a dedicated field.
+        Product.objects.create(
+            product_name=product['product_name'],
+            sort=product['sort'] if product['sort'] != -1 else 100,
+            is_active=product['sort'] != -1
+        )
+        insert_count += 1
+    print('(inserted: %s)' % insert_count, end='')
```
willkg: With other tables, we do a sql script and run it manually after a deploy. That means we have to do these switch-overs in multiple steps and multiple deploys. That's fine, but this table has like 10 things in it and it's easier to do it all in one fell swoop if we can and here, we can.

webapp-django/crashstats/supersearch/forms.py
```diff
@@ -30,34 +30,22 @@ def make_restricted_choices(sequence, exclude=None):
...
-        **kwargs
-    ):
+    def __init__(self, all_fields, products, product_versions, platforms, user, *args, **kwargs):
         super(self.__class__, self).__init__(*args, **kwargs)
 
         self.all_fields = all_fields.copy()
 
-        # Default values loaded from a database.
-        product_names = list(set(x['product_name'] for x in products))
+        # Generate default values
```
willkg: `products` is passed in as a flat list of product names, so we don't have to do anything with it. I also changed `platforms` in the same way.

willkg:
This looks good to me. I ran through the pages in the webapp and made sure the search and topcrasher form fields were populated. I'll check it out on stage.

One thing this changes is how to add new products. We could do it in the admin now or by data migration. We'll need to figure out which we should standardize on.

Anyhow, self-r+.

PR #4714, opened 2018-11-16
fix bug 1506907: implement archivescraper

willkg:
This adds a `ProductVersion` Django model (and table) to store the data that we need for the processor's `BetaVersionRule` and also diagnosing issues in the archivescraper.

It also implements a new archivescraper that, like ftpscraper, scrapes archive.mozilla.org and, unlike ftpscraper, doesn't involve stored procedures, only scrapes the directories that have information that we need, and only runs for Firefox and Fennec.

It also adjusts the `BetaVersionRule` to do a second db lookup and tell us whether it matches the answer acquired from Buildhub or the first db lookup. This will help us know whether the data is good or whether it needs some tweaking. (My local testing suggests it's super good.)

When this lands, it'll deploy to stage, crontabber will kick off the `ArchiveScraperCronApp` crontabber job which will scrape archive.mozilla.org, and populate the `crashstats_productversion` table. The `BetaVersionRule` will push metrics regarding whether the new table has the right answers or not. We'll let this sit for a few days and by Monday our future will be clear.


socorro/cron/jobs/archivescraper.py
```diff
@@ -0,0 +1,351 @@
...
+    'beetmover',
+    'contrib',
+    'funnelcake',
+    'jsshell',
+    'logs',
+    'mar-tools',
+    'partner-repacks',
+    'source',
+    'update'
+]
```
willkg: These just lets archivescraper skip some of the GET requests that it knows will fail. We could not have this and it'd be fine. It's probably not a comprehensive list.

socorro/cron/jobs/archivescraper.py
```diff
@@ -0,0 +1,351 @@
...
+        VALUES (
+            %(product_name)s, %(release_channel)s, %(release_version)s,
+            %(version_string)s, %(build_id)s, %(archive_url)s
+        )
+        """
+        try:
+            cursor.execute(sql, params)
+            self.successful_inserts += 1
+        except psycopg2.IntegrityError:
+            # If it's an IntegrityError, we already have it and everything is fine
```
willkg: You're fine. I'm fine. Everything is fine.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -590,10 +591,68 @@ def __init__(self, config):
...
-            config,
-            database,
-        )
+        self.transaction = config.transaction_executor_class(config, database)
+
+        try:
+            sentry_dsn = self.config.sentry.dsn
+        except (AttributeError, KeyError):
+            sentry_dsn = None
+        self.sentry_dsn = sentry_dsn
```
willkg: This is added so we can surface errors with the lookup against `crashtstats_productversion` without otherwise disrupting processing and this rule.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -590,10 +591,68 @@ def __init__(self, config):
...
+
+    def _get_real_version_db2(self, product, channel, build_id):
+        """Return real version number from crashstats_productversion table
+
+        NOTE(willkg): This is not live, yet--we're doing this to see how it
+        performs compared to the existing product_version table.
+
+        :arg product: the product
+        :arg build_id: the build id as a string
+        :arg channel: the release channel
```
willkg: Oops--wrong order.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -590,10 +591,68 @@ def __init__(self, config):
...
+            WHERE product_name = %(product)s
+            AND release_channel = %(channel)s
+            AND build_id = %(build_id)s
+        """
+        params = {
+            'product': product,
+            'channel': channel,
+            'build_id': build_id
+        }
+        versions = self.transaction(execute_query_fetchall, sql, params)
```
willkg: The archivescraper puts the info in the "right place" in regards to channels so this lookup only has to happen once--we don't have to do a second lookup switching things around in certain cases and when the moon is full.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -590,10 +591,68 @@ def __init__(self, config):
...
+            # Flatten versions
+            versions = [version[0] for version in versions]
+
+            if versions:
+                if 'b' in versions[0]:
+                    # If we're looking at betas, then ignore "rc" versions
+                    versions = [version for version in versions if 'rc' not in version]
+
+                else:
+                    # If we're looking at the release channel, then only return "rc"
```
willkg: This should be "non-betas" because this isn't actually related to channels.

socorro/cron/jobs/archivescraper.py
```diff
@@ -0,0 +1,351 @@
...
+
+                # The build link text is something like "build1/" and we
+                # want just the number part, so we drop "build" and the "/"
+                rc_version_string = version_root + 'rc' + build_link['text'][5:-1]
+
+                # Whether or not this is the final build for a set of builds; for
+                # example for [build1, build2, build3] the final build is build3
+                final_build = (i + 1 == len(build_links))
+
+                if final_build:
```
willkg: This is some mildly crazy logic because of the way release candidates work. archivescraper generates data that's explicitly oriented for the `BetaVersionRule` to do a lookup with a single db query. As such, the release candidates for a final build are in the "release" channel on Buildhub, but in the "beta" channel in `crashstats_productversion` because the crash reports for those builds come in with `ReleaseChannel=beta`.

I don't know if that's a big deal, but it's possible that it makes this data bad for other people to use.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -778,45 +864,49 @@ def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
+                if real_version != maybe_version:
+                    self.config.logger.info(
+                        'betaversionrule: got different versions: %r %r vs. %r',
+                        raw_crash.get('uuid', None),
+                        real_version,
+                        maybe_version
+                    )
+                    self.metrics.incr('lookup_alternate', 1, tags=['outcome:different'])
+                else:
+                    self.metrics.incr('lookup_alternate', 1, tags=['outcome:same'])
```
willkg: We can run this on stage over the weekend and graph this and have a really good idea of how it fares in the real world.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -141,6 +141,37 @@ class Meta:
         ordering = ['sort']
 
 
+class ProductVersion(models.Model):
+    """Lookup table for product versions and build information"""
+    product_name = models.CharField(
+        max_length=50, blank=False, null=False,
+        help_text='ProductName of product as it appears in crash reports'
+    )
```
willkg: This is not a foreign key because I want this table to stand alone and be easy to insert things into and look things up without dealing with joins. We're using it solely for the `BetaVersionRule`. Once Buildhub2 is up, we could ditch this and use Buildhub2 exclusively.

willkg:
It's good as far as I can tell. It includes metrics and logging that'll let us figure out any issues on stage. Landing it now means I have a day to fix any issues that pop up and then we can run it through the weekend. self-r+

PR #4716, opened 2018-11-16
fix bug 1507016: fix default timeout for requests sessions

willkg:


willkg:
Testing it out with httpbin:

```
app@socorro:/app$ python
Python 2.7.14 (default, Dec 12 2017, 06:55:29) 
[GCC 4.9.2] on linux2
>>> from socorro.lib.requestslib import session_with_retries
>>> import time
>>> sess = session_with_retries(default_timeout=1.0)
>>> time.ctime()
'Fri Nov 16 16:30:09 2018'
>>> sess.get('https://httpbin.org/delay/10')
  File "/usr/local/lib/python2.7/site-packages/requests/sessions.py", line 537, in get
    return self.request('GET', url, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/requests/sessions.py", line 524, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/local/lib/python2.7/site-packages/requests/sessions.py", line 637, in send
    r = adapter.send(request, **kwargs)
  File "socorro/lib/requestslib.py", line 32, in send
    return super(HTTPAdapterWithTimeout, self).send(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/requests/adapters.py", line 516, in send
    raise ConnectionError(e, request=request)
requests.exceptions.ConnectionError: HTTPSConnectionPool(host='httpbin.org', port=443): Max retries exceeded with url: /delay/10 (Caused by ReadTimeoutError("HTTPSConnectionPool(host='httpbin.org', port=443): Read timed out. (read timeout=1.0)",))
>>> time.ctime()
'Fri Nov 16 16:30:27 2018'
>>> 
```

That sets up a session that times out after 1.0 seconds by default--both connection and read timeouts. It requests httpbin in a way that causes httpbin to delay the response by 10 seconds. That exceeds the timeout, so the session will kill the connection and retry the default number of times with the default backoff factor and then give up with a ConnectionError.

All of that happens in about 18 seconds here. That's what we want for a default.

This sets up the default as 5.0 seconds. I think that's decent. The normal default is None which means it never times out. Blech. Some things will take longer, but they should adjust their timeouts differently.

willkg:
self-r+

PR #4717, opened 2018-11-18
bug 1507916: fix archive scraper scraping

willkg:
Previously, archivescraper would only grab build information for a single
platform per `buildN/` directory. That turns out to be wrong, so this fixes
it to get build information for ALL platforms per `buildN/` directory.

However, that radically increases the runtime of archivescraper, to this
adds an optimization where it looks at the db for the maximum major version
for that product and only scrapes information for versions greater than
`(major_version - 3)`. That significantly reduces the run time.


socorro/cron/jobs/archivescraper.py
```diff
@@ -251,84 +295,85 @@ def scrape_candidates(self, product_name, url_path):
...
-                        # This is a release channel build, but it's not a final build,
-                        # so insert it as an rc beta build
-                        data['version_string'] = rc_version_string
-                        data['release_channel'] = 'beta'
+                        # We have the older build info file format, so we use that
+                        # structure
+                        data = {
+                            'product_name': product_name,
+                            'release_channel': data['moz_update_channel'],
+                            'release_version': data['moz_app_version'],
```
willkg: Oops--need to add the major version here.

willkg:
This has no unit tests--that's covered in another bug. I spent some time today testing it with and without existing `crashstats_productversion` data and testing it in conjunction with the `BetaVersionRule` and problematic crashes. It improves things. Particularly the runtime of archivescraper.

I'm going to land this now so it has time to run on stage before Monday. self-r+

PR #4718, opened 2018-11-19
bug 1507916: improve difference logging

willkg:
This adjusts the logging line for when the new `BetaVersionRule` lookup differs
from the existing `BetaVersionRule`. We can use this data to generate new
entries in the `crashstats_productversion` table.


willkg:
self-r+

PR #4719, opened 2018-11-19
bug 1507186: fix usage of basestring

willkg:
This fixes usage of basestring across the application and converts uses of
it to `six.text_type` or `six.string_type` depending on what's going on.

### DIFF
--- a/webapp-django/crashstats/api/templatetags/jinja_helpers.py
+++ b/webapp-django/crashstats/api/templatetags/jinja_helpers.py
@@ -1,18 +1,17 @@
+import datetime
 import urllib
 import warnings
-import datetime
-from past.builtins import basestring
-from six import text_type
 
 from django_jinja import library
 import jinja2
+import six
 
 from django.forms.widgets import RadioSelect
 
 
 @library.global_function
 def describe_friendly_type(type_):
-    if type_ is basestring:
+    if type_ is six.text_type:
         return "String"
     if type_ is int:
         return "Integer"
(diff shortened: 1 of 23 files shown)


webapp-django/crashstats/api/templatetags/jinja_helpers.py
```diff
@@ -1,18 +1,17 @@
...
 from django_jinja import library
 import jinja2
+import six
 
 from django.forms.widgets import RadioSelect
 
 
 @library.global_function
 def describe_friendly_type(type_):
-    if type_ is basestring:
```
willkg: `type_` here is an actual type. It's set to `basestring`, but I claim that's wrong since `basestring` should be used for type checking of a value, not equality checking.

So I changed that instance to `six.text_type` and all the other places where it's checked to be `six.text_type`. That'll be `unicode` in Python 2 and `str` in Python 3. I think that should work fine, but we'll know better when this hits stage and possible weird data.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -401,7 +398,7 @@ def fetch(
             self.cache_seconds
         ):
             name = implementation.__class__.__name__
-            cache_key = hashlib.md5(name + text_type(params)).hexdigest()
+            cache_key = hashlib.md5(name + six.text_type(params)).hexdigest()
```
willkg: This won't work in Python 3 because `md5` works on `bytes` but this has been in production for a while and not causing issues, so it's probably the case it's "ok for now". I'll fix this when I get the webapp tests working in Python 3.

willkg:
I think this looks ok. It works locally. I'll test it more thoroughly on stage. self-r+

PR #4720, opened 2018-11-20
bug 1507916: add missing versions

willkg:
This adds a script for adding the missing versions between 40 and 49
for Firefox that aren't available on archive.mozilla.org to the
`crashstats_productversion` table.

It's pretty easy to run:

```
$ docker-compose run app shell python scripts/insert_missing_versions.py
```

We'll run this on stage and prod and that'll add the missing versions and then the `BetaVersionRule` should be fine with this lookup.


willkg:
self-r+

PR #4721, opened 2018-11-20
fix bug 1508335: clean up BetaVersionRule lookups

willkg:
This has two sets of big changes:

The first is to clean up the scaffolding in the socorro tests in regards to use of Django-managed tables. Previously, setting up and removing those tables was done by individual tests. Now it's done as a mixin/pytest fixture. That will make things a lot easier.

Second, this reworks the `BetaVersionRule` so that it's now just checking the `crashstats_productversion` table which is managed by Django and fed by the new archivescraper crontabber job. This drops use of Buildhub which isn't really meeting our needs right now and it drops use of the old `product_versions` table which is sqlalchemy/alembic managed.


socorro/processor/mozilla_transform_rules.py
```diff
@@ -578,42 +574,28 @@ class BetaVersionRule(Rule):
...
-        # NOTE(willkg): These config values come from Processor2015 instance.
-        self.buildhub_api = config.buildhub_api
         self.cache = ExpiringCache(max_size=self.CACHE_MAX_SIZE, default_ttl=self.SHORT_CACHE_TTL)
         self.metrics = markus.get_metrics('processor.betaversionrule')
 
         # NOTE(willkg): These config values come from Processor2015 instance and are
         # used for lookup in product_versions
-        database = config.database_class(config)
-        self.transaction = config.transaction_executor_class(config, database)
+        self.conn_context = config.database_class(config)
```
willkg: I took out the transaction executor so this rule doesn't need it. Soon I'll be nixing the transaction executor for all things. Figured I'd do it here and see how it went.

socorro/processor/mozilla_transform_rules.py
```diff
@@ -578,42 +574,28 @@ class BetaVersionRule(Rule):
...
 
-        try:
-            sentry_dsn = self.config.sentry.dsn
-        except (AttributeError, KeyError):
-            sentry_dsn = None
-        self.sentry_dsn = sentry_dsn
```
willkg: We needed this to surface errors from the db2 lookup to sentry in a way that let the rule continue running. We don't need that anymore.

socorro/unittest/conftest.py
```diff
@@ -25,3 +28,165 @@ def test_something(metricsmock):
 
     """
     return MetricsMock()
+
+
+class DjangoTablesMixin(object):
```
willkg: This centralizes maintenance of Django-managed tables for non-Django tests. This is a temporary thing as most of the tests that need this are crontabber jobs and those will eventually either go away or get rewritten as Django things. The only one that should stick around is the BetaVersionRule tests. We can figure out what to do with those then.

socorro/unittest/cron/crontabber_tests_base.py
```diff
@@ -75,78 +76,16 @@ def _wind(data):
...
-        "log_time" timestamp with time zone NOT NULL,
-        "duration" double precision NOT NULL,
-        "success" timestamp with time zone NULL,
-        "exc_type" text NULL,
-        "exc_value" text NULL,
-        "exc_traceback" text NULL);
-        """)
-        self.conn.commit()
+        db_connection_factory = self.config.crontabber.database_class(self.config.crontabber)
+        self.conn = db_connection_factory.connection()
```
willkg: This comes over from `setUpClass`.

socorro/unittest/processor/test_mozilla_transform_rules.py
```diff
@@ -1266,503 +1267,169 @@ def test_missing_key_2(self):
...
+            'major_version': int(version_string.split('.')[0]),
+            'release_version': version_string.split('.')[0],
+            'version_string': version_string,
+            'archive_url': 'test',
+        }
+        cursor.execute(sql, params)
+        conn.commit()
 
-    def test_beta_channel_known_version(self):
+    def test_beta_channel_known_version(self, db_conn, django_tables):
```
willkg: Look, ma! No mocking!

willkg:
This looks good. Will monitor on stage for a while. self-r+

PR #4723, opened 2018-11-21
fix bug 1503383: remove alembic, sqlalchemy, and non-Django things

willkg:
This is mostly code removal, but covers the following:

* Remove `ftpscraper` and `featured_versions` crontabber apps. We no longer
  need these since we no longer use `product_versions` and friends tables.
* Remove `ProductVersionsMiddleware` and API endpoint.
* Remove alembic infrastructure and migrations.
* Remove sqlalchemy infrastructure and models.
* Remove all the stored procedures and non-Django tables.
* Add a new "socorro-cmd db" subcommand to create and drop the database.
* Remove setupdb app and rework tests and Makefile scripts to use
  Django to create tables.
* Update a bunch of documentation.
* Remove crontabber's `ConnectionFactory`--it's a duplicate of the postgresql `ConnectionContext`.
* Move `dbapi2_util` to `socorro/lib/dbutil.py`--this is to prep for removing the postgresql directory altogether.

After this, the database is managed by Django and there are no non-Django
bits.

This doesn't get rid of the postgresql directory. The only thing left of the postgresql crashstorage stuff is the `ConnectionContext` because it's used by crontabber and the `BetaVersionRule` in the processor. I'm going to rewrite crontabber soon (next?). We can fix the `BetaVersionRule`, too.

This doesn't clean up transaction executor stuff. That work is covered by a different bug.

I'm pretty sure this doesn't affect the infra/deploy code in the cloudops repo. There are no configuration changes or script changes. Everything uses existing entry points (for example, `make setup`, `make updatedata`, `make test`, ...).


willkg:
After this lands, we need to run `scripts/1503383-drop-everything.sql` on stage and prod. Running that will wipe out a bunch of tables. If there's stuff we want to keep, we should save it first. We already have a dump of `product_versions` and `product_version_builds` which contains information I can't find anywhere else. I don't think there's anything else we want to save. Most of those tables are lookup tables.

docker/run_setup_postgres.sh
```diff
@@ -6,35 +6,32 @@
...
-# that!" kind of thing
-echo "Dropping existing db and creating new db named (${DATABASE})..."
-./socorro/external/postgresql/setupdb_app.py \
-                  --database_name="${DATABASE}" \
-                  --dropdb \
-                  --createdb
+# Create the database if it doesn't exist
+echo "Dropping and creating db..."
+python socorro/scripts/db.py drop || true
+python socorro/scripts/db.py create || true
```
willkg: The `|| true` bit here is to ignore exit codes. It's ok if it can't drop a database that doesn't exist.

docker/run_tests.sh
```diff
@@ -28,23 +28,18 @@ ELASTICSEARCH_URL=${elasticsearch_url:-"http://elasticsearch:9200"}
...
-"${PYTHON}" "${SETUPDB}" --database_name=socorro_test --dropdb --logging.level=40 --unlogged --no_staticdata --createdb
-
-# Test the last migration
-"${ALEMBIC}" -c "${alembic_config}" downgrade -1
-"${ALEMBIC}" -c "${alembic_config}" upgrade heads
+# Set up socorro_test db
+./socorro-cmd db drop || true
+./socorro-cmd db create
+pushd webapp-django
+${PYTHON} manage.py migrate
```
willkg: We now use Django to build all the tables. Yay!

docker/run_update_data.sh
```diff
@@ -15,18 +15,12 @@ HOSTUSER=$(id -u):$(id -g)
...
-./docker/run_ftpscraper_wrapper.sh
-
-# Update featured versions data based on release data
-${DC} run app shell ${CRONTABBERCMD} --reset-job=featured-versions-automatic
-${DC} run app shell ${CRONTABBERCMD} --job=featured-versions-automatic \
-    --crontabber.class-FeaturedVersionsAutomaticCronApp.products=${PRODUCTS}
+# Fetch release data (verbosely)
+${DC} run app shell ${CRONTABBERCMD} --reset-job=archivescraper
+${DC} run app shell ${CRONTABBERCMD} --job=archivescraper \
+    --crontabber.class-ArchiveScraperCronApp.verbose
```
willkg: archivescraper is slow in a fresh local dev environment. It takes 20-30 minutes to do its thing. I have some ideas on how to speed that up, but haven't done that, yet. I'll think about it later.

scripts/1503383-drop-everything.sql
```diff
@@ -0,0 +1,46 @@
...
+DROP TABLE IF EXISTS product_release_channels;
+DROP TABLE IF EXISTS product_version_builds;
+DROP TABLE IF EXISTS product_versions;
+DROP TABLE IF EXISTS products;
+DROP TABLE IF EXISTS release_channels;
+DROP TABLE IF EXISTS release_repositories;
+DROP TABLE IF EXISTS releases_raw;
+DROP TABLE IF EXISTS special_product_platforms;
+
+COMMIT WORK;
```
willkg: This works locally. I did a pass a while back to make local == stage == prod, but there are probably some minor differences. If this fails, that's fine. We can fix it until it works in our leisure.

socorro/cron/mixins.py
```diff
@@ -90,71 +92,12 @@ def both_inits(self, *args, **kwargs):
 
 using_postgres = partial(
     with_transactional_resource,
-    'socorro.cron.connection_factory.ConnectionFactory',
+    'socorro.external.postgresql.connection_context.ConnectionContext',
```
willkg: I nixed `ConnectionFactory` since it was a duplicate of `ConnectionContext` from when I vendored crontabber.

socorro/scripts/db.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python
```
willkg: I figure we can use this subcommand for other things, too, if that's helpful.

socorro/unittest/conftest.py
```diff
@@ -30,163 +30,7 @@ def test_something(metricsmock):
     return MetricsMock()
 
 
-class DjangoTablesMixin(object):
```
willkg: I literally just wrote this earlier this week, so I'm a bit bummed to see it go so soon. But ... so it goes.

willkg:
I tested the `scripts/1503383-drop-everything.sql` script against a db generated in the local dev environment. I did a pass getting stage, prod, and the local dev environments the same a while back, so this might be sufficient. Or maybe not. We can adjust that script and run it until it works in our leisure.

I tested the basic local dev environment flows: 'make build`, `make setup`, `make updatedata`, `make lint`, `make test`, and `make docs`. Everything works fine.

I ran the processor and processed some crashes. The `BetaVersionRule` uses postgres. It works fine.

I ran crontabber. That is the biggest user of postgres. It worked fine. All the jobs work fine.

I ran the webapp. It has always used Django, so nothing interesting here. The API docs work fine.

I think we're good to go. If there are any problems, we'll see them on stage. Having said that, I'm pretty confident this is at least mostly fine.

self-r+.

adngdb:
That's one small step for a man, one giant leap for socorro.



Congrats Will! :-)

PR #4724, opened 2018-11-21
fix bug 1503383: fix crontabber

willkg:
Crontabber isn't running because the "main()" bit was taken out of the
Python module. Need to run it via socorro-cmd now.


willkg:
This isn't covered by the tests at all, so I'm going to just land it. self-r+

PR #4727, opened 2018-11-27
fix bug 1503434: redo retries and transactions

willkg:
Socorro has a family of `TransactionExecutor` classes which perform transaction
bookkeeping and also retry failed operations. The way it's implemented doesn't
work with Python 3. Rather than fix the issues, this splits the
TransactionExecutor into a retry function and a transaction_context context
manager and then redoes everything to use those.

This does a couple of things:

1. simplifies executing operations with transaction bookkeeping requirements
   so they're easier to reason about

2. simplifies retriable operations so they're easier to reason about

3. gets rid of one or both in places that don't need them


willkg:
I need to figure out a test plan for this to make sure the parts I changed maintain the stability properties they had before (or get better--that'd be cool, too).

There's no more "infinite retry" code anymore, so I don't think we have to worry about infinite loops. The code is a lot simpler than it was before and I tried to make it more explicit. That should make it easier to test and review.

Even if we muck up testing, it's likely that big issues will pop up on stage pretty quickly and be fixable and then it'll be "good enough" to push to prod and hone over time.

Anyhow, next steps with this:

1. make sure it passes Circle CI
2. write up a test plan
3. run through the test plan locally
4. push to stage and let it bake for a while

socorro/cron/crontabber_app.py
```diff
@@ -952,6 +909,8 @@ def _run_one(self, job_class, config, force=False):
         last_success = None
         now = utc_now()
         log_run = True
+
+        exc_type = exc_value = exc_tb = None
```
willkg: This needed to get hoisted up because if you hit ctrl-c, then it errors out because these names don't exist in that flow.

socorro/external/boto/connection_context.py
```diff
@@ -73,58 +72,39 @@ class ConnectionContextBase(RequiredConfig):
         reference_value_from='resource.boto'
     )
 
-    operational_exceptions = (
+    RETRYABLE_EXCEPTIONS = (
```
willkg: Here's the new model: the connection context defines an `is_retryable_exception` method which determines yes or no as to whether it's retryable. If it's retryable, then `retry` will recycle the connection and try again. That radically simplifies the two different classes of exceptions and the middling ground that `is_operational_exception` occupied.

socorro/external/boto/connection_context.py
```diff
@@ -73,58 +72,39 @@ class ConnectionContextBase(RequiredConfig):
...
-    conditional_exceptions = (
-        boto.exception.StorageResponseError
-    )
-
-    def is_operational_exception(self, x):
-        if "not found, no value returned" in str(x):
-            # the not found error needs to be re-tryable to compensate for
-            # eventual consistency.  However, a method capable of raising this
-            # exception should never be used with a transaction executor that
-            # has infinite back off.
```
willkg: We no longer do infinite backoffs. This is in the crontabber and processor, so if they fail in some long, but ephemeral issue, that's "ok" and we can recover from it. We don't want to wait infinitely.

socorro/external/boto/connection_context.py
```diff
@@ -254,29 +233,23 @@ def _convert_list_to_string(self, a_list):
     def _convert_string_to_list(self, a_string):
         return json.loads(a_string)
 
-    def commit(self):
-        """boto doesn't support transactions so this silently
-        does nothing"""
-
-    def rollback(self):
-        """boto doesn't support transactions so this silently
-        does nothing"""
```
willkg: boto is not transactional, so we don't need any transaction bookkeeping.

socorro/external/boto/connection_context.py
```diff
@@ -254,29 +233,23 @@ def _convert_list_to_string(self, a_list):
...
-        """boto doesn't support transactions, so it is never in
-        a transaction."""
-        return False
-
     def force_reconnect(self):
-        try:
-            del self.connection
-        except AttributeError:
-            # already deleted, ignorable
-            pass
```
willkg: Boto does its stuff via HTTPS, so I don't think there's a persistent connection to reconnect. Given that, I just nixed this code.

socorro/external/boto/crashstorage.py
```diff
@@ -64,35 +48,13 @@ class BotoCrashStorage(CrashStorageBase):
...
 
-    def is_operational_exception(self, x):
-        if "not found, no value returned" in str(x):
-            # the not found error needs to be re-tryable to compensate for
-            # eventual consistency.  However, a method capable of raising this
-            # exception should never be used with a transaction executor that
-            # has infinite back off.
-            return True
-        #elif   # for further cases...
-        return False
```
willkg: This got moved to the connection context.

socorro/external/es/connection_context.py
```diff
@@ -9,35 +9,8 @@
...
-        pass
-
-    def rollback(self):
-        pass
-
-    def close(self):
-        pass
-
-    def __getattr__(self, name):
-        return getattr(self._connection, name)
```
willkg: Elasticsearch is not transactional, so I took out this facade code.

socorro/external/es/connection_context.py
```diff
@@ -73,19 +46,6 @@ class ConnectionContext(RequiredConfig):
...
-        doc='a classname for the type of wrapper for ES connections',
-        reference_value_from='resource.elasticsearch',
-    )
-
-    # Operational exceptions are retryable, conditionals require futher
-    # analysis to determine if they can be retried or not.
-    operational_exceptions = (
-        elasticsearch.exceptions.ConnectionError,
-    )
-    conditional_exceptions = ()
```
willkg: `_submit_crash_to_elasticsearch` has its own retry logic, so we don't additionally use the `retry` function.

socorro/external/es/crashstorage.py
```diff
@@ -273,10 +249,7 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash, crash_id):
             'raw_crash': raw_crash
         }
 
-        self.transaction(
-            self._submit_crash_to_elasticsearch,
-            crash_document=crash_document
-        )
```
willkg: Because `_submit_crash_to_elasticsearch` had its own retry logic, that meant we had DOUBLE RETRYING!

socorro/external/es/crashstorage.py
```diff
@@ -453,11 +432,6 @@ def __init__(self, config, *args, **kwargs):
         self.raw_crash_redactor = config.raw_crash_es_redactor.redactor_class(
             config.raw_crash_es_redactor
         )
-        self.config.logger.warning(
-            "Beware, this crashstorage class is destructive to the "
-            "processed crash - if you're using a polycrashstore you may "
-            "find the modified processed crash saved to the other crashstores."
-        )
```
willkg: We handle this differently now, so we don't need this warning.

socorro/external/es/crashstorage.py
```diff
@@ -527,147 +501,3 @@ def save_raw_and_processed(self, raw_crash, dumps, processed_crash, crash_id):
             processed_crash,
             crash_id
         )
-
-
-class QueueWrapper(Queue):
```
willkg: I took out the bulk crash store code since we don't use it.

socorro/external/rabbitmq/connection_context.py
```diff
@@ -61,12 +59,6 @@ def __init__(self, config, connection,
         # adding some common semantics to aid the implementation of the fully
         # abstracted RabbitMQCrashStorage class.
 
-    def commit(self):
-        pass
-
-    def rollback(self):
-        pass
```
willkg: RabbitMQ is not transactional, so I nixed these.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -79,40 +63,18 @@ def __init__(self, config, namespace='', quit_check_callback=None):
             quit_check_callback=quit_check_callback
         )
 
-        self.config = config
```
willkg: This gets set in the superclass `__init__`, so we don't need to do it again here.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -79,40 +63,18 @@ def __init__(self, config, namespace='', quit_check_callback=None):
...
+            # Make message persistent
+            delivery_mode=2,
         )
 
-        if config.throttle == 100:
-            self.dont_queue_this_crash = lambda: False
-        else:
-            self.dont_queue_this_crash = (
-                lambda: randint(1, 100) > config.throttle
-            )
```
willkg: The throttle stuff was used by the old crashmover which we don't use anymore, so I nixed this.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -79,40 +63,18 @@ def __init__(self, config, namespace='', quit_check_callback=None):
...
-        if self.dont_queue_this_crash():
-            self.config.logger.info(
-                'Crash %s filtered out of RabbitMQ queue %s',
-                crash_id,
-                self.config.routing_key
-            )
-            return
         try:
             this_crash_should_be_queued = (
                 not self.config.filter_on_legacy_processing or
```
willkg: I could probably nix "legacy_processing" check and related code, too, but I figured testing that is involved and this PR is a little out of control at this point.

socorro/unittest/external/es/test_connection_context.py
```diff
@@ -1,45 +0,0 @@
...
-# License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/.
-
-import elasticsearch
-
-from socorro.external.es.connection_context import ConnectionContext
-from socorro.unittest.external.es.base import ElasticsearchTestCase
-
-
-class IntegrationTestConnectionContext(ElasticsearchTestCase):
```
willkg: This tested a lot of uninteresting boilerplate and also the wrapped faux connection which I rmoved, so I nixed it all.

socorro/lib/transaction.py
```diff
@@ -8,161 +8,107 @@
...
+    last_failure = None
+
+    for backoff_time in BACKOFF_TIMES:
+        try:
+            with connection_context() as conn:
+                return fun(conn, **kwargs)
+        except Exception as exc:
+            last_failure = sys.exc_info()
+            if connection_context.is_retryable_exception(exc):
+                connection_context.config.logger.warning(
```
willkg: This happens all the time in the processor when it's processing a crash it hasn't processed before which is usually the case, so this is kind of annoying.

I think I want to change this so it only logs on the last exception thrown rather than the interim exceptions.

willkg: On second thought, I think I'm going to nix the logging altogether. The caller can log the error if it wants to.

socorro/external/boto/connection_context.py
```diff
@@ -254,29 +233,23 @@ def _convert_list_to_string(self, a_list):
...
+
+    def is_retryable_exception(self, exc):
+        return (
+            # These are ephemeral problems that might go away with time
+            isinstance(exc, self.RETRYABLE_EXCEPTIONS) or
+
+            # This indicates that the item isn't there, but in the case of
+            # fetching raw crash data, it's possible it's not there *yet* so we
+            # want to give it a few more chances
+            'not found, no value returned' in str(exc)
```
willkg: Before these changes, this was only checked if the exception was a `StorageResponseError`. As near as I can tell, that can never happened. In the transaction executor, when it was handling the exception, the exception was a `CrashIDNotFound` which was neither in the operational or conditional lists, so it never went through this check.

The problem with the changes I made is that this check triggers when the processor is trying to get the processed crash which doesn't exist in most cases. So this slows down processing for every crash by 5 seconds. Yuck.

I'm going to nix this for now and if we need to re-add some kind of "retry if the crash data isn't there yet" thing, we can do it some other way.

willkg:
Test plan:

1. run the processor and process some crashes
2. shut off localstack-s3 and try to process a crash--processor can't get the crash, so it rejects it
3. turn on localstack-s3
4. in webapp, view a crash report
5. in webapp, view a non-existent crash report--should get "crash not found" page
6. set up a crash for processing and view it in webapp--should get the "waiting for crash" page and it should eventually find it and show the crash report

I think that's the important stuff.

willkg:
Here we go! Woo hoo! self-r+

PR #4730, opened 2018-11-28
fix bug 1510435: fix edge case where there's an rc, but no final

willkg:
This fixes a theoretical problem that we haven't seen, yet, where
archivescraper is scraping at a time when there's an rc build, but
no final build for a given version. In this scenario, archivescraper
would create a version entry, but that would be wrong and then
incoming crash reports would pick up the wrong version.

For example, say there's a 64.0b4 build1 build. That gets added
as 64.0b4rc1--that's fine. It would also get added as 64.0b4 (final)
but that's not correct because later, there'd be a 64.0b4 build2.
So any incoming crashes for 64.0b4rc1 would get the version 64.0b4.

This fixes it so that it checks the /releases/ page for all the
released version and if that version isn't on the page, it doesn't
create an entry for the final version. I think that's correct, but
since this hasn't happened, yet, I'm guessing.

We'll have to see if this works or not. If it doesn't, then I'm not
sure what other signal we can derive from archive.mozilla.org to
disambuguate the situation.

### DIFF
--- a/socorro/cron/jobs/archivescraper.py
+++ b/socorro/cron/jobs/archivescraper.py
@@ -88,6 +88,13 @@ NON_PLATFORM_SUBSTRINGS = [
 ]
 
 
+def key_for_build_link(link):
+    # The path is something like "build10/". We want to pull out the
+    # integer from that.
+    path = link['path']
+    return int(''.join([c for c in path if c.isdigit()]))
+
+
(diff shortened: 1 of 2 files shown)


willkg:
Pretty sure this is right. I can check it out by watching the versions in the admin page day-to-day. self-r+

PR #4731, opened 2018-11-28
bug 1497956: upgrade postgres to 9.5 in local dev environment

willkg:
This updates the local dev and test environments to use PostgreSQL 9.5.


willkg:
Yay! self-r+

PR #4733, opened 2018-11-29
bug 1511163: increase BugzillaCronApp timeout

willkg:
Today, the BugzillaCronApp is timing out when reading from Bugzilla. This
increases the timeout to 30 seconds for that app because it seems it needs
more time to read.


willkg:
self-r+

PR #4734, opened 2018-11-30
bug 1504067: remove unittest from socorro/unittest/ tests

willkg:
This removes use of `unittest.TestCase` from the `socorro/unittest/` tests. In
doing that, I also created a fixture to handle db stuff and moved the `req_mock`
fixture.  I fixed the Elasticsearch tests which were order dependent because
they were adjusting `a_processed_crash`. I also removed a bunch of unused code.


socorro/external/es/crashstorage.py
```diff
@@ -22,6 +22,30 @@
...
+        'completed_datetime',
+        'completeddatetime',
+    ]
+    for a_key in datetime_fields:
+        if a_key not in processed_crash:
+            continue
+
+        processed_crash[a_key] = string_to_datetime(processed_crash[a_key])
+
+
```
willkg: Hoisting this to the module level lets us use it in the unittests.

socorro/unittest/conftest.py
```diff
@@ -33,4 +40,26 @@ def test_something(metricsmock):
 @pytest.fixture
 def db_conn():
```
willkg: This now handles creating a connection, but also truncating certain tables before and after a test. crontabber apps use this, but also the `BetaVersionRule` processor rule tests.

socorro/unittest/cron/crontabber_tests_base.py
```diff
@@ -45,148 +43,31 @@ def get_config_manager(jobs=None, overrides=None):
     )
 
 
-class IntegrationTestBase(unittest.TestCase):
```
willkg: The bulk of this class isn't used anymore, so I nixed it.

socorro/unittest/cron/jobs/test_archivescraper.py
```diff
@@ -131,30 +127,9 @@ def config():
         yield class_config
 
 
-@pytest.fixture
-def req_mock():
-    with requests_mock.mock() as mock:
-        yield mock
```
willkg: This was put in `socorro/unittest/conftest.py`.

socorro/unittest/cron/jobs/test_archivescraper.py
```diff
@@ -131,30 +127,9 @@ def config():
...
-    def setup_method(self, method):
-        self.truncate_productversions()
-
-    def teardown_method(self, method):
-        self.truncate_productversions()
-
-    def truncate_productversions(self):
-        cursor = self.conn.cursor()
-        cursor.execute('TRUNCATE crashstats_productversion')
-        self.conn.commit()
```
willkg: Since the methods use the `db_conn` fixture, we can nix all this.

socorro/unittest/cron/jobs/test_bugzilla.py
```diff
@@ -78,28 +75,28 @@ def fetch_data(self):
...
-    def test_basic_run_job(self, requests_mocker):
-        requests_mocker.get(BUGZILLA_BASE_URL, json=SAMPLE_BUGZILLA_RESULTS)
+    def test_basic_run_job(self, db_conn, req_mock):
+        req_mock.get(BUGZILLA_BASE_URL, json=SAMPLE_BUGZILLA_RESULTS)
         config_manager = self._setup_config_manager(3)
 
         with config_manager.context() as config:
             tab = CronTabberApp(config)
-            tab.run_all()
+            tab.run_one('bugzilla-associations')
```
willkg: I switched the way this runs so that it only runs this one job.

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -207,29 +192,25 @@ def __init__(self, *args, **kwargs):
...
     def test_index_crash(self):
-        """Test indexing a crash document.
-        """
-
+        """Test indexing a crash document"""
         es_storage = ESCrashStorage(config=self.config)
 
         es_storage.save_raw_and_processed(
-            raw_crash=a_raw_crash,
+            raw_crash=deepcopy(a_raw_crash),
```
willkg: These calls all need to deepcopy the raw crash and processed crash. The Elasticsearch crashstorage modifies the processed crash.

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -368,10 +335,13 @@ def test_success(self, espy_mock):
         # IndexCreator but is part of the crashstorage workflow).
         assert espy_mock.client.IndicesClient.called
 
+        expected_processed_crash = deepcopy(a_processed_crash)
+        reconstitute_datetimes(expected_processed_crash)
```
willkg: Previously, this test was order dependent. Something was changing the processed crash and then that change carried on through other tests. After fixing that by deepcopying in every case, suddenly the expected document didn't match the actual document because the actual document has datetime strings turned into datetimes. So ew have to do this thing here now to get everything to match.

socorro/unittest/testbase.py
```diff
@@ -1,11 +0,0 @@
...
-# License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/.
-
-import unittest
-
-
-class TestCase(unittest.TestCase):
-
-    def shortDescription(self):
-        return None
```
willkg: This `shortDescription` thing was to fix a thing nose did. We don't use nose anymore, so we can nix this file altogether.

willkg:
This looks fine. self-r+

PR #4736, opened 2018-12-05
bug 1497957: upgrade to postgre 9.6

willkg:
This upgrades the local development environment and the test environment to
PostgreSQL 9.6.

We're in the process of upgrading to 9.5 right now, so we'll want to wait until that's done before we land this.


willkg:
Looks good! self-r+

PR #4737, opened 2018-12-07
fix bug 1512633: fix archivescraper to pick up esr builds

willkg:


socorro/cron/jobs/archivescraper.py
```diff
@@ -298,14 +298,19 @@ def scrape_candidates(self, product_name, archive_directory):
...
+                'Skipping anything before %s and not esr', major_version_minus_4
+            )
             version_links = [
                 link for link in version_links
-                # "63.0b7-candidates/" -> 63
-                if int(link['text'].split('.')[0]) >= major_version_minus_3
+                if (
+                    # "63.0b7-candidates/" -> 63
+                    int(link['text'].split('.')[0]) >= major_version_minus_4 or
+                    'esr' in link['text']
```
willkg: This will pick up *all* esr builds even ones that we're not doing updates on anymore. We can probably optimize that out, but it'd require another query to the `product_versions` table. Right now, it's not a big deal--there aren't that many additional directories to scrape. It'll be more onerous over time.

Having said that, the incremental archivescraper pass takes under 3 minutes on my laptop, so I think this is fine.

willkg:
Right now, stage has 60.0.3esr builds, but doesn't have 60.0.4esr builds. After this fix lands, archivescraper will pick up those builds and we'll know we're good.

willkg:
self-r+

PR #4739, opened 2018-12-11
fix bug 1426257: clean up _transform

willkg:
This is mostly cosmetic cleanup. The method was already doing most of
what we need it to do in regards to structure and exception handling.

### DIFF
--- a/socorro/processor/processor_app.py
+++ b/socorro/processor/processor_app.py
@@ -12,7 +12,7 @@ import sys
 from configman import Namespace
 from configman.converters import class_converter
 from configman.dotdict import DotDict
-from six import reraise
+import six
 
 from socorro.app.fetch_transform_save_app import FetchTransformSaveWithSeparateNewCrashSourceApp
 from socorro.external.crashstorage_base import CrashIDNotFound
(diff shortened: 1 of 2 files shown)


socorro/processor/processor_app.py
```diff
@@ -242,25 +242,30 @@ def _transform(self, crash_id):
...
                 exc_info = exc_value
             else:
                 exc_info = [(exc_type, exc_value, exc_tb)]
 
             for exc_info_item in exc_info:
                 self._capture_error(crash_id, exc_info_item)
+                self.config.logger.warning(
+                    'error in processing or saving crash %s',
+                    crash_id
+                )
```
willkg: This makes it easier to see in development.

socorro/processor/processor_app.py
```diff
@@ -242,25 +242,30 @@ def _transform(self, crash_id):
...
             for a_dump_pathname in dumps.values():
-                try:
-                    if 'TEMPORARY' in a_dump_pathname:
+                if 'TEMPORARY' in a_dump_pathname:
+                    try:
                         os.unlink(a_dump_pathname)
-                except OSError as x:
-                    self.config.logger.info('deletion of dump failed: %s', x)
+                    except OSError as x:
+                        self.config.logger.info('deletion of dump failed: %s', x)
```
willkg: I inverted the if block because if `TEMPORARY` isn't in the name, then there's no sense in doing the work to set up exception handling.

socorro/unittest/processor/test_processor_app.py
```diff
@@ -170,7 +170,7 @@ def mocked_save_raw_and_processed(*_):
         with pytest.raises(PolyStorageError):
             pa.transform('mycrashid')
 
-        config.logger.warning.assert_called_with(
+        config.logger.warning.assert_any_call(
```
willkg: There are now two logging calls, so we have to switch this to `assert_any_call` so it looks at all of them rather than the most recent.

willkg:
self-r+

PR #4741, opened 2018-12-12
remove dead code

willkg:


scripts/deploy_bug.py
```diff
@@ -15,8 +15,6 @@
 Then copy and paste the bits it prints out.
 """
 
-from __future__ import print_function
```
willkg: This is a thing we needed in Python 2, but in Python 3, print *is* a function so we don't need these anymore.

socorro/lib/util.py
```diff
@@ -2,49 +2,11 @@
...
 
-from __future__ import print_function
-
 import collections
 
-from itertools import islice
 import six
 
 
-def chunkify(iterable, n):
```
willkg: `more_itertools` has `chunked` which does the same thing, so might as well switch to that one.

socorro/scripts/add_crashid_to_queue.py
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/env python
```
willkg: These scripts get called from `socorro-cmd` and we don't expect anyone to use them directly, so we can nix all this runtime stuff.

webapp-django/crashstats/crashstats/migrations/0001_initial.py
```diff
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 # Generated by Django 1.11.15 on 2018-09-25 15:40
-from __future__ import unicode_literals
```
willkg: This affects what happens to unicode when printing to a terminal that doesn't specify its encoding. This import would cause Python 2 to do what Python 3 does. Python 3 does this by default, so we don't need these anymore.

Having said that, I think this is part of the Django template for migration files, so it's possible more will pop up over time.

webapp-django/crashstats/urls.py
```diff
@@ -1,5 +1,3 @@
-from __future__ import absolute_import
```
willkg: Python 2 did this screwy thing with implicit relative imports. I added this line to fix an issue in this file where the implicit relative imports broke things. Python 3 doesn't do implicit relative imports, so we don't need this line anymore.

willkg:
Looks good. self-r+

PR #4742, opened 2018-12-12
fix bug 1513595: switch webapp to gunicorn

willkg:
This switches the webapp from uwsgi to gunicorn which matches how Antenna runs.


willkg:
We need to add some stuff to the cloudops repo before landing this. I mimicked what Antenna does. This uses the following variables:

* `PORT`
* `GUNICORN_WORKERS`
* `GUNICORN_WORKER_CONNECTIONS`
* `GUNICORN_WORKER_CLASS`
* `GUNICORN_MAX_REQUESTS`
* `GUNICORN_MAX_REQUESTS_JITTER`

I'll put together a PR in the cloudops repo and we can discuss things there before this lands.

willkg:
@sciurus @milescrabill In the cloudops repo, we define all these variables for the collector, but not the webapp. For the webapp, it seems like the variables are defined in the script in the socorro repo. The only one that's "interesting" is probably `GUNICORN_WORKERS` the value of which is determined by some math and the processorcount.

Would you prefer to define these in cloudops land or in socorro land? I don't have a preference one way or the other.

willkg:
Perfect! I'll mimic what we do with the collector and then they'll be the same!

willkg:
The related cloudops PR landed, so I'm going to land this and fix any issues on stage as they happen.

PR #4743, opened 2018-12-14
bug 1513595: fix webapp logging with gunicorn

willkg:


willkg:
Antenna has these in one of the files. I figure it can't hurt. It might be a logging level issue, though. self-r+

PR #4744, opened 2018-12-14
fix bug 1513426: fix DepencencySecurityCheckCronApp

willkg:
It was using nsp which is dead. This switches it to use npm audit.


docker/Dockerfile
```diff
@@ -6,7 +6,7 @@ ARG userid=10001
 
 WORKDIR /app/
 RUN groupadd --gid $groupid app && \
-    useradd -g app --uid $userid --shell /usr/sbin/nologin app
+    useradd -g app --uid $userid --shell /usr/sbin/nologin --create-home app
```
willkg: `npm` gets mighty fussy if `$HOME` doesn't exist. I figured we might as well create it. That saves us from dealing with getting around whatever it is `npm` is doing.

docker/config/local_dev.env
```diff
@@ -105,7 +105,6 @@ OVERVIEW_VERSION_URLS=http://localhost:8000/__version__
 
 # crontabber
 # ----------
-crontabber.class-DependencySecurityCheckCronApp.nsp_path=/webapp-frontend-deps/node_modules/.bin/nsp
```
willkg: We don't need this anymore since we're using `npm` which is installed into the Docker container at `/usr/bin/npm`.

socorro/cron/jobs/monitoring.py
```diff
@@ -70,8 +70,9 @@ class DependencySecurityCheckCronApp(BaseCronApp):
 
     required_config = Namespace()
     required_config.add_option(
-        'nsp_path',
-        doc='Path to the nsp binary',
+        'npm_path',
+        doc='Path to the npm binary',
+        default='/usr/bin/npm'
```
willkg: Since this is installed in the Docker container, I made it a default.

socorro/unittest/cron/jobs/test_monitoring.py
```diff
@@ -157,65 +157,82 @@ def test_get_javascript_vulnerabilities_failure(self, mock_popen, app_config):
...
-                    'socorro-webapp-django@0.0.0',
-                    'jquery@2.1.0',
-                ],
-                'advisory': 'https://nodesecurity.io/advisories/328',
+        # Adapated from npm audit output for a jest issue
+        output = json.dumps({
+            "actions": [
+                # Skipping actions because we don't do anything with them.
+            ],
+            "advisories": {
```
willkg: The shape of the output is different than it used to be. I tried to map like things so the `Vulnerability` instance was similar.

willkg:
We don't need to make a change in the cloudops repo before landing this since we set an `npm_path` default. self-r+

PR #4745, opened 2018-12-14
fix bug 1514294: remove jest, enzyme, react, and webpack

willkg:
This removes scaffolding from the react cupcake that we decided not to go
forward with.

To test:

1. build the container
2. run the tests
3. run `make run` which runs the webapp and click around

If all that works, then building and testing all work fine.


willkg:
self-r+

PR #4746, opened 2018-12-14
fix bug 1463514: remove pinax-evenlog

willkg:
We don't use this app anymore, so remove it and the table it uses.

Also, this picked up a bunch of other table changes probably from
the Python 3 upgrade.


willkg:
self-r+

PR #4747, opened 2018-12-17
Bug 1514746 - add std::panicking::begin_panic<T> to the sentinels list

jcristau:


willkg:
I looked at 10 crashes and this change doesn't help much. For example:

```
app@fb5ccac425c2:/app$ socorro-cmd signature c6e9a60f-b87e-409a-bfbb-3770f0181217
Crash id: c6e9a60f-b87e-409a-bfbb-3770f0181217
Original: GeckoCrashOOL
New:      gkrust_shared::panic_hook
Same?:    False
```

That crash is here:

https://crash-stats.mozilla.com/report/index/e0f74b0f-7f76-446a-9c7c-260650181217

Seems like we want signature generation to start after `std::panicking::begin_panic`, so we should add that to the sentinels list, but curiously it's already there. I'll look into what's going on.

jcristau:
Thanks for the note about sentinels. I now get this from the socorro-siggen signature tool:

Crash id: 863f53d1-6e81-40bd-9252-839240181218
Original: GeckoCrashOOL
New:      webrender::device::gl::TextureUploader<T>::upload<T>
Same?:    False

PR #4749, opened 2018-12-19
fix bug 1511022: better signature generation debug logging

willkg:


willkg:
This adjusts the `--verbose` flag so that it prints stuff like this:

```
app@08c8e5aa8526:/app$ socorro-cmd signature -v e0f74b0f-7f76-446a-9c7c-260650181217
Crash id: e0f74b0f-7f76-446a-9c7c-260650181217
Original: GeckoCrashOOL
New:      GeckoCrashOOL
Same?:    True
Debug:    (5)
          SignatureGenerationRule: Using CSignatureTool
          SignatureGenerationRule: irrelevant; ignoring: "MOZ_CrashOOL"
          SignatureGenerationRule: not a prefix; stop iterating at: "GeckoCrashOOL"
          SignatureGenerationRule: change: "" -> "GeckoCrashOOL"
          SigFixWhitespace: change: "GeckoCrashOOL" -> "GeckoCrashOOL"
```

Makes it a lot easier to see how changes to the files affects generation.

For this pass, I added the scaffolding for doing debug logging and then added debug logging to the `CSignatureTool` which is where I most need it. We can add additional logging as we need it elsewhere.

socorro/signature/rules.py
```diff
@@ -507,20 +527,23 @@ def action(self, crash_data, result):
...
-        signature, signature_notes = self.c_signature_tool.generate(
+        signature, notes, debug_notes = self.c_signature_tool.generate(
             signature_list,
             crash_data.get('hang_type'),
             crash_data.get('crashing_thread')
         )
 
         if signature_list:
-            result['proto_signature'] = ' | '.join(signature_list)
+            result.extra['proto_signature'] = ' | '.join(signature_list)
```
willkg: Should double-check there's a test that makes sure the `proto_signature` makes it into the processed crash.

socorro/signature/rules.py
```diff
@@ -627,15 +651,14 @@ class SigFixWhitespace(Rule):
     * reduce consecutive spaces to a single space
 
     """
-
     WHITESPACE_RE = re.compile(r'\s')
     CONSECUTIVE_WHITESPACE_RE = re.compile(r'\s\s+')
 
     def predicate(self, crash_data, result):
-        return isinstance(result.get('signature'), six.string_types)
+        return isinstance(result.signature, six.string_types)
```
willkg: This seems silly since the signature is always a str and doesn't change. We should probably nix this.

socorro/signature/rules.py
```diff
@@ -646,20 +669,19 @@ def action(self, crash_data, result):
         # Reduce consecutive spaces to a single space
         sig = self.CONSECUTIVE_WHITESPACE_RE.sub(' ', sig)
 
-        result['signature'] = sig
+        result.set_signature(self.name, sig)
```
willkg: We could capture the original signature and only set it if it changed. That gives us better diagnostics about what happened.

socorro/signature/tests/test_rules.py
```diff
@@ -1326,50 +1280,39 @@ def test_action_success_remove_unwanted_parts(self, abort_msg, expected):
...
-        result = {
-            'signature': '',
-            'notes': []
-        }
+        result = generator.Result()
         assert rule.predicate({}, result) is True
 
-        result['signature'] = 42
+        result.signature = 42
         assert rule.predicate({}, result) is False
```
willkg: This test is silly.

willkg:
This looks good. self-r+

PR #4752, opened 2018-12-20
fix bug 1506722: re-add missing dependencies

willkg:
Somehow, when I generated package-lock.json after removing underscore.js, it
removed a bunch of stuff we still need. This broke the dependency checker.

This re-adds that.


willkg:
self-r+

PR #4754, opened 2018-12-21
bug 1515805: fix breakpad client building in image; add scripts

willkg:
Building the breakpad client requires using depot_tools which only work with
Python 2. This fixes the `build-breakpad.sh` script so it makes sure depot_tools
are using Python 2 in the Docker container.

This adds a `scripts/run_mdsw.sh` script which runs minidump-stackwalker just
like we do in the processor, but outside of the processor. This makes it
a lot easier to debug issues with minidump-stackwalker.

### DIFF
--- a/scripts/build-breakpad.sh
+++ b/scripts/build-breakpad.sh
@@ -35,6 +35,15 @@ cd ..
 export PATH
 PATH=$(pwd)/depot_tools:$PATH
 
+# depot_tools only work in Python 2 and it uses "/usr/bin/env python", so
+# we take advantage of depot_tools being first in the PATH and create a
+# symlink to the happy Python if "python" is Python 3.
+PYV=$(python -c "import sys; print(sys.version_info[0]);")
+if [ "${PYV}" == "3" ]; then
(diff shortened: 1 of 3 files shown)


scripts/build-breakpad.sh
```diff
@@ -35,6 +35,15 @@ cd ..
 export PATH
 PATH=$(pwd)/depot_tools:$PATH
 
+# depot_tools only work in Python 2 and it uses "/usr/bin/env python", so
+# we take advantage of depot_tools being first in the PATH and create a
+# symlink to the happy Python if "python" is Python 3.
+PYV=$(python -c "import sys; print(sys.version_info[0]);")
+if [ "${PYV}" == "3" ]; then
+  echo "'/usr/bin/env python' is Python 3, so making symlink to python2"
+  ln -s /usr/bin/python2 $(pwd)/depot_tools/python
```
willkg: This doesn't work if there is no `python2` binary, but having a `python2` is standard enough that we can deal with that if it ever pops up.

willkg:
self-r+

PR #4759, opened 2019-01-04
Re-add __future__ statement for siggen

willkg:
siggen still supports Python 2.7, so we need to re-add this `__future__`
line.


willkg:
self-r+

PR #4760, opened 2019-01-07
fix bug 1518143: add EOFError to handled errors in OutOfMemoryBinaryRule

willkg:
gzip can throw an EOFError when reading the blob. This adds handling
for that.


willkg:
self-r+

PR #4763, opened 2019-01-08
fix bug 1515805: update breakpad client to 1459e5df

willkg:
Once this lands, we can run the taskcluster task that uses this script to update the client binary. Once the binary is built, we'll need to trigger another stage deploy so that stage mdsw picks it up..


willkg:
If I do `make shell` and then `./scripts/build-breakpad.sh`, it pulls down the correct sha of the breakpad client and builds it. I think that's the best I can do to verify correctness of this change. self-r+

PR #4764, opened 2019-01-08
fix bug 1518576: fix bug description when filing bugs from crash reports

willkg:
Bugzilla switched to Markdown for formatting and the bugs that were created
from Bugzilla looked ter-rib-ble. This fixes that and reduces some of the
unnecessary boilerplate.


willkg:
self-r+

PR #4768, opened 2019-01-10
Fix bug 1515805: Update Breakpad version.

Osmose:
This should be the last change for this bug. I built it locally and it seemed to compile without an issue.

@willkg Any suggestions on what to land after the build to get this on stage?


willkg:
I'll verify this with crashes tomorrow.

For triggering the deploy, I vote tweaking the README or something in the docs.

PR #4771, opened 2019-01-14
Fix deploy_bug script for the Markdown world

willkg:


willkg:
This just affects writing up deploy bugs. If it's icky, I'll do another iteration of fixing later. self-r+

PR #4772, opened 2019-01-14
fix bug 1113583: remove index name caching code

willkg:
This removes index name caching code which never kicked in because
the cache was never updated.

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -206,10 +206,6 @@ class ESCrashStorage(CrashStorageBase):
         reference_value_from='resource.elasticsearch',
     )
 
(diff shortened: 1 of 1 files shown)


socorro/external/es/crashstorage.py
```diff
@@ -325,9 +321,8 @@ def _submit_crash_to_elasticsearch(self, crash_document):
         crash_id = crash_document['crash_id']
 
         # Attempt to create the index; it's OK if it already exists.
-        if es_index not in self.indices_cache:
```
willkg: Nothing updates `indices_cache`, so this code was just hanging out and never had any effect on anything.

willkg:
This is just code cleanup and doesn't affect behavior at all. self-r+

PR #4773, opened 2019-01-14
Fix bug 1511032: Move user info and documentation link to header.

Osmose:
The header was starting to get a little crowded, so I decided to put the user menu options within a dropdown:

screen shot 2019-01-14 at 3 20 02 pm

screen shot 2019-01-14 at 3 18 51 pm

screen shot 2019-01-14 at 3 19 43 pm

### DIFF
--- a/webapp-django/crashstats/base/jinja2/crashstats_base.html
+++ b/webapp-django/crashstats/base/jinja2/crashstats_base.html
@@ -30,6 +30,46 @@
             <input type="text" id="q" name="query" placeholder="Find Crash ID or Signature">
             <input type="submit" class="hidden">
         </form>
+
+        <div class="user-info">
+            {% if request.user and request.user.is_active %}
+                <button class="user-info-button">
+                    <img class="gravatar" src="{{ gravatar_url(request.user.email, size=64) }}">
+                    <span class="user-email">{{ request.user.email }}</span>
+                    <img class="menu-icon" src="{{ static('img/menu.svg') }}" >
+                </button>
+                <ul class="user-info-menu">
+                    <li>
+                        <a href="{{ url('profile:profile') }}"
+                          title="You are signed in as {{ request.user.email }}"
+                        >
+                            Your Profile
+                        </a>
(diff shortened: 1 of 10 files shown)


webapp-django/crashstats/base/jinja2/crashstats_base.html
```diff
@@ -106,7 +148,9 @@ <h1>Product Navigation</h1>
         </ul>
     {% endblock %}
 
-        <div class="search">
+        <div class="nav-links">
+            <a href="{{ url('documentation:home') }}">Documentation</a>
+            |
```
willkg: I forgot... I think we should have a "file a bug" kind of thing here. It should be much easier to file a bug against Socorro than it is now.

Having said that, I'm not sure how best to do that. Maybe add a "File a bug" here? Maybe there's a better place in the header? Maybe clearer/shorter phrasing?

willkg: We don't have to figure that out in this PR. I could do it in a follow-up one. But I figured I'd mention it in case you had ideas.

willkg:
The branch is out of date and it's failing on an eslint check. Otherwise I think this is fine.

I <3 this. Thank you so much for working on it!

webapp-django/crashstats/base/jinja2/crashstats_base.html
Osmose: I added a link to the [Bugzilla form for the Socorro component](https://bugzilla.mozilla.org/enter_bug.cgi?product=Socorro):

screen shot 2019-01-14 at 3 37 53 pm

Osmose:
Updated with lint fixes and a "File a bug" link.

Osmose:
You can tell I'm rusty when I just completely forget to run the tests "because it's just frontend changes".

Osmose:
I restored the deleted html files that were breaking the tests (since they're used in the error view) and rebased off master, and the tests pass for me now.

PR #4774, opened 2019-01-15
bug 1519475: switch worker in taskcluster task

willkg:
Bug 1519475 covers redoing how we build the breakpad client. This is
an interim fix to the task and how we document it so that I can run it.
This takes the urgency from figuring out a more permanent solution.

Also, in this change, I'm taking "ownership" and responsibility of the taskcluster task figuring then it's off of @luser's plate and he'll be happier for it. If that's not true, I'm happy to change it back.


willkg:
I did some timings and decided to just nix the whole pre-built binary thing. Closing this out.

PR #4775, opened 2019-01-15
fix bug 1519475: drop use of pre-built breakpad client binary

willkg:
The breakpad client binary was being built using a taskcluster task
that required scopes that I couldn't get. I was able to change the
worker so I could run the task, but it only saves a minute and a
half from build times. I don't think that warrants the additional
complexity maintaining a binary required so this nixes that and
now always builds breakpad client during build.


willkg:
CI took 10m23s. That's good enough. self-r+

PR #4776, opened 2019-01-15
fix bug 1519527: remove buildhub admin view

willkg:
We had a buildhub admin view to let us compare data we had in the
`product_versions` table with what was on Buildhub so as to figure out what
Buildhub was missing.

We decided to redo our `archive.mozilla.org` scraper and forego using Buildhub
for now. Further, we removed the `product_versions` table.

When I did that work, I should have removed the buildhub admin view, but
forgot. This fixes that.


willkg:
This is just code removal. self-r+

PR #4777, opened 2019-01-15
fix bug 1275576: remove distributor and distributor_version

willkg:
This removes two fields from the processed crash that aren't being used.


willkg:
This removes two fields that were never used and never exposed in super search. self-r+

PR #4778, opened 2019-01-15
Fix bug 1490622: Highlight linked line when coloring source code.

Osmose:
The original tests for this view did not run anymore due to the testcase class
name not starting with the word "Test", so this also removes the class and
refactors the tests to run under pytest.


willkg:
Linting kicked up an unused import. Otherwise this looks good to me. I can r+ it after fixing that.

PR #4779, opened 2019-01-16
fix bug 1519802, bug 1305956: fix cpu_info, cpu_arch/cpu_name, and cpu_count

willkg:
Previously, `cpu_info` was composed of `cpu_info` and `cpu_count` from the
`system_info` in the `json_dump`. Socorro indexed this and also `cpu_count`
so they were both available for searching. The report view showed the
`cpu_info` value.

But this is confusing in a few scenarios and there doesn't seem to be a good
reason to have a contrived value like this.

This changes it so `cpu_info` is just the `cpu_info` value from
`json_dump.system_info`. It also adds `cpu_count` to the crash report and the
report view.

This also changes the name from "Build Architecture Info" (which doesn't
make sense) to "CPU Info".

While doing that, I removed two deprecated items from Elasticsearch fields
and also moved the whole `cpu_name` -> `cpu_arch` deprecation another step.


willkg:
I went through and tested report view, searching, search facets with crash reports processed before this change and crash reports processed after this change.

I tested the signature summary page with crash reports processed before this change and crash reports processed after this change.

Some of the changes are there to use `cpu_arch` and if that's empty, then try `cpu_name`. In July 2019, we can stop doing that because all the crash reports will have a `cpu_arch`. I added notes about that in a bunch of places. Once this lands, I'll write up a bug about that.

I think that covers everything.

socorro/processor/general_transform_rules.py
```diff
@@ -16,25 +16,22 @@ def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
-        processed_crash['cpu_name'] = cpu_name
-        processed_crash['cpu_info'] = cpu_info
+        # This is the CPU that the product was built for
+        processed_crash['cpu_arch'] = glom(
+            processed_crash, 'json_dump.system_info.cpu_arch', default=''
+        )
+        # NOTE(willkg): "cpu_name" is deprecated and we can remove it in July 2019
+        processed_crash['cpu_name'] = glom(
+            processed_crash, 'json_dump.system_info.cpu_arch', default=''
+        )
```
willkg: Because we're still using it for Elasticsearch and the API, we need to keep `cpu_name` until then.

socorro/external/es/super_search_fields.py
```diff
@@ -1765,10 +1765,12 @@ def get(self):
...
         'description': (
             'The build architecture. Usually one of: "x86", "amd64" (a.k.a. x86-64), "arm", '
-            '"arm64".\n\nDuplicate of cpu_name, with a better name.'
+            '"arm64".'
         ),
         'form_field_choices': [],
         'has_full_version': False,
+        # NOTE(willkg): this uses "cpu_name" to search on, but that's deprecated;
+        # switch to "cpu_arch" after July 2019
         'in_database_name': 'cpu_name',
```
willkg: Uh... This is a mistake. This is continuing to populate `processed_crash.cpu_name` in the ES document, but in order to transition in July 2019, we need to also add `processed_crash.cpu_arch`. I'll add that now.

socorro/unittest/processor/test_general_transform_rules.py
```diff
@@ -147,8 +147,9 @@ def test_everything_we_hoped_for(self):
         # the call to be tested
         rule.act(raw_crash, raw_dumps, processed_crash, processor_meta)
 
-        assert processed_crash.cpu_info == 'GenuineIntel family 6 model 42 stepping 7 | 4'
         assert processed_crash.cpu_name == 'x86'
+        assert processed_crash.cpu_info == 'GenuineIntel family 6 model 42 stepping 7'
+        assert processed_crash.cpu_count == 4
```
willkg: `cpu_info` is no longer composed of `cpu_info | cpu_count`.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -218,21 +218,27 @@ <h2>{{ report.product }} {{ report.version }} Crash Report [@ {{ report.signatur
...
 
-                            <tr title="{{ fields_desc['processed_crash.cpu_name'] }}">
+                            <tr title="{{ fields_desc['processed_crash.cpu_arch'] }}">
                                 <th scope="row">Build Architecture</th>
-                                <td>{{ report.cpu_name }}</td>
+                                <td>{{ report.cpu_arch }}</td>
                             </tr>
                             <tr title="{{ fields_desc['processed_crash.cpu_info'] }}">
-                                <th scope="row">Build Architecture Info</th>
+                                <th scope="row">CPU Info</th>
```
willkg: This fixes bug 1305956.

willkg:
bug 1303758 covers removing `cpu_name` things from the ES document in july 2019.

willkg:
I'm not planning to do a deploy this week. I want to land this now so it can bake on stage for a couple of weeks and when I get back, I can verify that everything is working properly.

PR #4780, opened 2019-01-16
Exclude build/ from linting

willkg:
The `build/` directory holds breakpad-client-related build stuff that's
external to this project, but has a bunch of Python files that have
linting issues. This excludes them.


willkg:
self-r+

PR #4781, opened 2019-01-16
fix bug 1520570: switch to multi-stage docker image builds

willkg:
This moves building breakpad client into a separate docker image build stage.
That way all its intermediary build tools and files don't swamp the final
`socorro_app` image and make it super huge.

This also adds support for `NOCACHE=1 make build` which will build the
docker image with `--no-cache` flag. I do that so often that it seemed
prudent to codify it in the `Makefile` and save myself some time.

This also tweaks the `.dockerignore` file to ignore breakpad client
build bits during the `COPY` which might show up in a local dev environment
when working on updating the breakpad client.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,8 @@
 .git
+build/
+breakpad/
+breakpad.tar.gz
+depot_tools/
 **/*.swp
 **/*.pyc
 __pycache__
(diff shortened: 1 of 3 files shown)


willkg:
This drops the size of the resulting `socorro_app` image for me by a lot. I think that means it'll go back to its previous level.

We could maybe fiddle with this some more, but I think it's diminishing returns at this point and any big changes would likely make development on certain things harder. Having said that, I'm game for other ideas.

PR #4783, opened 2019-01-17
fix bug 1515325: handle when bug_ids is not a list of numbers

willkg:


willkg:
I verified that `/api/SignaturesByBug/?bug_id=e` now errors out normally rather than raising an HTTP 500. self-r+

PR #4784, opened 2019-01-17
fix bug 1299460: improve crash not found page

willkg:
This improves the crash report not found page. It explains several reasons
why the crash report may not be available and what to do about it. This
includes the possibility that the crash report is no longer on our system.


willkg:
I think this improves the text. The bug suggests we switch up the text based on the circumstances. While I can see how that's helpful, I'd much rather go for the low-maintenance solution of including the various situations on the page and letting the user figure out what their issue might be.

Given that this bug has been sitting around for a couple of years with no activity, I think this is a good enough fix.

self-r+

PR #4785, opened 2019-01-17
fix bug 1158386: add help text for all public API endpoints

willkg:


willkg:
This is better than before. We can improve it further as need be. self-r+

PR #4788, opened 2019-01-18
fix bug 1404225: fix status message to be HTML-safe

willkg:
This redoes how status messages work such that they can no longer contain
arbitrary HTML.

We do want bugs linked, so this redoes how that works, too and instead of
linking a url, it generates a bugzilla link from a "bug #xxx" formed text.


willkg:
Further, this documents the format in the admin so we know what it supports.

willkg:
This looks good. self-r+

PR #4792, opened 2019-01-30
bug 1511032: move bug link over

willkg:
This swaps the "file a bug" link with "super search" because the latter
is heavily used and it's better to put it on the right-most side for
easier mousing.


willkg:
self-r+

PR #4798, opened 2019-01-31
bug 1523284: add Fenix to crash stats site

willkg:


webapp-django/crashstats/crashstats/migrations/0016_1523284_fenix.py
```diff
@@ -0,0 +1,35 @@
+# Generated by Django 2.1.5 on 2019-01-31 16:28
+
+# Bug 1523284: Create Fenix Product
+
+from django.db import migrations
+
+
+PRODUCT = 'Fenix'
+SORT = 5
+IS_ACTIVE = True
```
willkg: I did it this way because it makes it easier to copy-and-paste for later product additions.

willkg:
This looks good. Still need to do the antenna side before this will have much effect. self-r+

PR #4799, opened 2019-01-31
fix bug 1508215: make graphics adapter/device keywords

willkg:
This fixes the analysis for graphics adapter id and graphics device id
so the values are not tokenized into smaller parts. Firefox sends
values like 0x8000 which tokenizes into a single token. Fennec, however,
sends values like "Imagination Technologies" which gets broken into two
tokens and then we end up with two facets which messes up the Graphics
Adapter section in the signature summary report.

This fixes that.

### DIFF
--- a/socorro/external/es/super_search_fields.py
+++ b/socorro/external/es/super_search_fields.py
@@ -858,6 +858,7 @@ FIELDS = {
         'permissions_needed': [],
         'query_type': 'enum',
         'storage_mapping': {
+            'analyzer': 'keyword',
             'type': 'string'
         }
     },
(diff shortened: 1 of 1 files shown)


willkg:
I tested this locally and it's definitely the right thing to do. It'll affect crashes in indexes starting with the one created after the prod deploy--it won't fix crashes in indexes before that. So even if we reprocess crashes, they won't get fixed. There's nothing I can do about that short of rewriting our Elasticsearch-related code.

Anyhow, self-r+ so that this gets on stage and I can verify it there.

PR #4800, opened 2019-02-01
fix bug 1389186: stop doing shell=True

willkg:
When running minidump-stackwalker as an external process, it'd use `shell=True`
because the command line was complicated and involved redirection. Instead
of doing that, we can use `shlex.split` to split it into arguments and do
the redirection in the `subprocess.Popen` call. Then we're not using `shell=True`
which has security implications.


willkg:
This seems good. I ran it locally and it continued to produce mdsw output. self-r+

PR #4803, opened 2019-02-01
bug 1357246: add de-nulling to processor

willkg:
We want to move the de-nulling step from the collector to the processor. This
makes the raw crash data more authentic in regards to what we got from the
client. Further, it reduces the work that the collector needs to do.

### DIFF
--- a/socorro/processor/general_transform_rules.py
+++ b/socorro/processor/general_transform_rules.py
@@ -7,6 +7,45 @@ from glom import glom
 from socorro.processor.rules.base import Rule
 
 
+class DeNullRule(Rule):
+    """Removes nulls from keys and values
+
+    Sometimes crash reports come in with junk data. This removes the egregious
+    junk that causes downstream processing and storage problems.
+
+    NOTE(willkg): While this changes the raw crash, that raw crash should never
+    get saved back to storage, so this doesn't overwrite the original material.
+
+    """
+    def de_null(self, s):
+        """Remove nulls from bytes and str values
+
+        :arg str/bytes value: The str or bytes to remove nulls from
+
+        :returns: str or bytes without nulls
+
+        """
+        if isinstance(s, bytes) and b'\0' in s:
+            return s.replace(b'\0', b'')
+
+        if isinstance(s, str) and '\0' in s:
+            return s.replace('\0', '')
+
+        # If it's not a bytes or a str, it's something else and we should
+        # return it as is
+        return s
+
+    def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
+        # Go through the raw crash and de-null keys and values
+        for key, val in raw_crash.items():
+            new_key = self.de_null(key)
+            if key != new_key:
+                del raw_crash[key]
+
+            new_val = self.de_null(val)
+            raw_crash[new_key] = new_val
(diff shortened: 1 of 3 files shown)


socorro/processor/general_transform_rules.py
```diff
@@ -7,6 +7,45 @@
...
+        """
+        if isinstance(s, bytes) and b'\0' in s:
+            return s.replace(b'\0', b'')
+
+        if isinstance(s, str) and '\0' in s:
+            return s.replace('\0', '')
+
+        # If it's not a bytes or a str, it's something else and we should
+        # return it as is
+        return s
```
willkg: This is identical to the code in antenna now. I spent some time looking at whether checking for null bytes and then replacing was faster than just replacing.

Just replacing:

```
$ python3 -m timeit -n 10000 -s "s = 'a' * 100000" "s.replace('\0', '')"
10000 loops, best of 3: 26.1 usec per loop
```

Test and then replace:

```
$ python3 -m timeit -n 10000 -s "s = 'a' * 100000" "if '\0' in s: s.replace('\0', '')"
10000 loops, best of 3: 0.986 usec per loop
```

I think the data we're fixing generally doesn't have null characters--it's all key/val string data from the `.extra` and annotations in the crash report. So test-and-replace it is!

socorro/unittest/processor/test_general_transform_rules.py
```diff
@@ -99,6 +101,61 @@
...
+        ('abc\u0000', 'abc'),
+        ('abc\0', 'abc'),
+        ('ab\0c\0', 'abc'),
+        (b'abc\0', b'abc'),
+        (b'a\0bc\0', b'abc'),
+    ])
+    def test_de_null(self, data, expected):
+        config = get_basic_config()
+        rule = DeNullRule(config)
+        assert rule.de_null(data) == expected
```
willkg: These tests were lifted from antenna.

socorro/unittest/processor/test_general_transform_rules.py
```diff
@@ -99,6 +101,61 @@
...
+
+        config = get_basic_config()
+        rule = DeNullRule(config)
+        rule.act(raw_crash, {}, {}, get_basic_processor_meta())
+
+        assert raw_crash == DotDict({
+            'key1': 'val1',
+            'key2': b'val2',
+            'key3': 'val3'
+        })
```
willkg: Currently, the raw crash is a DotDict. When we do a processor rewrite, we'll change that. So it's important to test both possibilities: dict and DotDict.

willkg: Hrm... that reminds me that the raw crash is read in by boto crashstorage and at that point it's tossed into a DotDict. If it has null keys then, that'll fail. I'll look into that.

willkg: Nevermind--this is fine. JSON can't store binary data so we're guaranteed for the keys to always be string types.

willkg:
This feels ok to me. Antenna is still de-nulling the data, so this code won't be fixing anything, though it will be running. I'll see how that goes on stage and proceed accordingly next week. self-r+

PR #4804, opened 2019-02-01
fix bug 1401274: nix the user-provided value

willkg:
This removes the user-provided value from the error message. It's a
possible security issue depending on what is done with the text. We
could escape it or change it, but I think that reduces its helpfulness
and it's better to just nix it.


willkg:
This is the easiest solution to the issue. It might make it a little harder for users to debug their API-using things, but probably not by much. If it does, someone will tell us and we can figure out what to do then.

self-r+

PR #4807, opened 2019-02-02
fix bug 1315109: don't show correlations tab if not supported

willkg:
This moves the list of supported products for correlations to the view
code and fixes the JS to build the options based on that. The correlations.js
code still has its own list of supported products, but we can fix that
another day.

This also adjusts the jinja template such that if the specified product
isn't supported by correlations, it won't show the correlations tab at
all.

### DIFF
--- a/webapp-django/crashstats/signature/jinja2/signature/signature_report.html
+++ b/webapp-django/crashstats/signature/jinja2/signature/signature_report.html
@@ -29,6 +29,7 @@
     data-fields="{{ fields | to_json }}"
     data-columns="{{ columns | to_json }}"
     data-sort="{{ sort | join(',') }}"
+    data-correlations-products="{{ correlations_products | to_json }}"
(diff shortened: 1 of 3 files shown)


willkg:
Testing this was a bit tricky. I had to process crash reports for Firefox and Thunderbird and then try the signature report out with the following settings:

1. go to signature report with no product filter -- shows tab, no errors in console
2. go to signature report with a product filter for Firefox -- shows tab, no errors in console
3. go to signature report with a product filter for Thunderbird -- does not show tab, no errors in console

It works fine.

Having said that, I find it confusing when the interface is context-sensitive. In this case the tab is there sometimes, but not other times. I'll land it this way for now, but might change it at some point depending on what users think.

willkg:
Seems ok. We'll see how it works for users and adjust as necessary. self-r+

PR #4808, opened 2019-02-02
fix bug 606630: show friendly build date

willkg:
This fixes the display of the build id in the report view so that it shows
the build id, a humanized version of the build id date, and then an explicit
link to the Buildhub data for that build.

This also adds the channel to the Buildhub link--that will focus the data
better. That'll help a little with bug #1056217.

This also moves the buildhub base url to settings--we shouldn't have had
it hardcoded in the template.

### DIFF
--- a/webapp-django/crashstats/crashstats/context_processors.py
+++ b/webapp-django/crashstats/crashstats/context_processors.py
@@ -3,7 +3,5 @@ from django.conf import settings as django_settings
 
 def settings(request):
     return {
-        'SETTINGS': django_settings,
-        'DEBUG': django_settings.DEBUG,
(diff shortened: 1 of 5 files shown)


webapp-django/crashstats/crashstats/context_processors.py
```diff
@@ -3,6 +3,7 @@
 
 def settings(request):
     return {
+        'SETTINGS': django_settings,
         'DEBUG': django_settings.DEBUG,
         'GOOGLE_ANALYTICS_ID': django_settings.GOOGLE_ANALYTICS_ID,
```
willkg: We could probably just add `SETTINGS` and then remove the other two since they're in the settings. Looks like `DEBUG` isn't used at all. The other one is used only once.

willkg:
Seems fine. self-r+

PR #4809, opened 2019-02-04
Update requirements

willkg:
This replaces pull #4802 .


willkg:
This covers a security update, too. self-r+

PR #4812, opened 2019-02-05
fix bug 1467578: support mozlog for processor, crontabber, and webapp

willkg:


willkg:
It's impossible to read mozlog format as a human, so I tweaked it so that server environments set up logging one way and local dev environments another. You can switch between the two by adding one of the following to your `my.env`:

```
LOCAL_DEV_ENV=True
```
for local dev environment mode (which is already in `docker/config/never_on_a_server.env` and

```
LOCAL_DEV_ENV=
```
for server mode.

That makes it easy to test this out.

I also adjusted the webapp console logging for local development environments to have timestamps and other bits. That makes it easier to see when things happen and the name of the logger that logged the message.

willkg:
self-r+

PR #4813, opened 2019-02-06
fix bug 1377632: fix crash signature bug summary

willkg:
This changes the bug summary from:

```
Crash in OOM | Small
```

to something easier to parse:

```
Crash in [@ OOM | Small]
```

which matches how we do crash signatures elsewhere.


willkg:
self-r+ with :fire:!

PR #4814, opened 2019-02-06
fix bug 1438001: drop flatpicker for HTML date/time fields

willkg:
This reworks the date/time fields on supersearch and signature report pages. Instead of using flatpicker which had problems with local time vs. UTC, this uses HTML date/time fields and now everything is specified in UTC.

Further, I added a "day ends" shortcut which I've wanted for a while which sets the "from" and "to" dates to their appropriate day end times.


willkg:
Here's my test plan:

1. go to super search, verify that the date and time are now in UTC
   1. click on the short cut links and make sure they change the "from" date appropriately
   2. click on the "day end" shortcut and make sure it sets the "from" time to 12:00AM and the "to" time to 11:59PM (note that whether it's 12 or 24-hour depends on your system locale)
   3. change the date and time, click on "Search" and make sure the UTC date and time are in the url as well as the Public API URL
   4. open that url in a new tab and make sure the "from" and "to" date and time get set correctly
2. go to a signature report, open the filters section, and verify that the date and time are now in UTC
   1. change the date and time and click "Search" and verify the UTC date and time are in the url
   2. open that url in a new tab and make sure the "from" and "to" date and time get set correctly

I think that covers everything. I don't see these widgets being used elsewhere in the crash stats site.

The plan is to review this code, land it, get it deployed to stage, and then ask the stability list for help verifying it works right and getting feedback.

webapp-django/crashstats/crashstats/form_fields.py
```diff
@@ -1,47 +0,0 @@
-# This Source Code Form is subject to the terms of the Mozilla Public
```
willkg: This file contained fields that were used in the past but aren't used anymore--this is all dead code now.

webapp-django/crashstats/crashstats/forms.py
```diff
@@ -27,10 +27,6 @@ class BaseForm(_BaseForm, forms.Form):
     pass
 
 
-class Html5DateInput(forms.DateInput):
-    input_type = 'date'
```
willkg: We don't use this input anymore.

webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -102,39 +102,6 @@ def time_tag(dt, format='%a, %b %d %H:%M %Z', future=False):
...
-                data-input
-            ><a data-toggle><i class="icon-calendar"></i></a>
-        </span>
-        """.format(
-            input_name=input_name,
-            # We need special formatting here because that's the only timezone
-            # format the JS library will correctly parse.
-            default_value=default_value.strftime('%Y-%m-%dT%H:%M:%SZ'),
-        )
-    )
```
willkg: This is only used in one place, so I folded the code into that place. It makes for less indirection.

webapp-django/crashstats/supersearch/jinja2/supersearch/macros/date_filters.html
```diff
@@ -18,6 +23,7 @@
     <a href="#" data-range="2w">14 days</a> -
     <a href="#" data-range="1M">1 month</a> -
     <a href="#" data-range="3M">3 months</a> -
-    <a href="#" data-range="6M">6 months</a>
+    <a href="#" data-range="6M">6 months</a> |
+    <a href="#" data-range="dayends">day ends</a>
```
willkg: The "dayends" thing sets the "from" time to 12:00am and the "to" time to 11:59pm. It's kind of nice.

webapp-django/crashstats/supersearch/jinja2/supersearch/macros/date_filters.html
```diff
@@ -1,12 +1,17 @@
 {% macro date_filters(start_date, end_date) %}
+  {# NOTE(willkg): the start_date and end_date here are probably not in UTC #}
```
willkg: I need to change this note to "This assumes the start_date and end_date are in UTC."

willkg:
I think this is good to land and get feedback from. self-r+

PR #4815, opened 2019-02-06
fix bug 1146956: add UTC label to date processed field

willkg:
Since the date processed field comes from Socorro, we can guarantee that it
is in UTC. We can do the same for the submitted datetime stamp. We can't do
that for other datetime stamps because they come from the client and who
knows what state it's in.

This adds a UTC label to the "Date processed" field in the report view.


willkg:
self-r+

PR #4816, opened 2019-02-06
bug 1525614: adjust server logging settings for webapp

willkg:
This tweaks the server logging settings for the webapp. It removes
`django.request` and adds `django.server`. That'll show incoming HTTP requests.

This also fixes all the loggers to use the same `LOGGING_LEVEL`.


willkg:
I can test this locally by switching `LOCAL_DEV_ENV` value in my `my.env` file and thus see the two modes. It looks better. 

This doesn't fix gunicorn, though. I'll look at that later. self-r+

PR #4818, opened 2019-02-08
Add missing versions to "make updatedata"

willkg:
This changes "make updatedata" to also run the insert_missing_versions.py
script after running the archivescraper crontabber job.


willkg:
self-r+

PR #4819, opened 2019-02-08
bug 1510658: track cache misses for BetaVersionRule

willkg:
This fixes the code to also track cache misses for the BetaVersionRule.


willkg:
self-r+

PR #4823, opened 2019-02-11
fix bug 1419899: implement __heartbeat__ and __lbheartbeat__

willkg:
This implements the `__heartbeat__` and `__lbheartbeat__` dockerflow endpoints
for the webapp. It reworks the existing `healthcheck` code and deprecates
the old version. We can remove that as soon as we update the infrastructure.

This also moves `__version__` from the crashstats app to the monitoring
app; seems to be a more appropriate place.

### DIFF
--- a/webapp-django/crashstats/crashstats/tests/test_views.py
+++ b/webapp-django/crashstats/crashstats/tests/test_views.py
@@ -1693,34 +1693,6 @@ class TestLogin(BaseTestViews):
         assert 'Insufficient Privileges' in smart_text(response.content)
 
 
-class TestDockerflow(object):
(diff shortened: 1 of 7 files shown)


willkg:
To test locally, hit these Dockerflow endpoints:

* http://localhost:8000/__version__ -- this worked before and should continue to work; it returns `{}` because there's no `version.json` file
* http://localhost:8000/__heartbeat__
* http://localhost:8000/__lbheartbeat__

These endpoints are deprecated, but still in use, so we should make sure they work, too:

* http://localhost:8000/monitoring/healthcheck
* http://localhost:8000/monitoring/healthcheck?elb=true

willkg:
self-r+

PR #4825, opened 2019-02-11
fix bug 1393416: stop copying Winsock_LSP to processed crash

willkg:
There was a processor rule for copying the `Winsock_LSP` value from the
raw crash to the processed crash, but nothing can use or access it
in the processed crash.

This removes that rule.


willkg:
self-r+

PR #4826, opened 2019-02-13
bug 1527728: fix versions in nav bar

willkg:
The versions in the nav bar are pulled from Super Search, so they're based
on the crash reports we've seen for that product in the last 60 days.

This fixes the query so that it's not cached in the `SuperSearchUnredacted` model
because it's cached in the `get_versions_for_product`. This also increases the
`_facets_size` so that it picks up versions which have recently been released
for which we don't have many crash reports, yet.

### DIFF
--- a/webapp-django/crashstats/crashstats/utils.py
+++ b/webapp-django/crashstats/crashstats/utils.py
@@ -369,19 +369,22 @@ def get_versions_for_product(product='Firefox', use_cache=True):
     now = timezone.now()
 
     # Find versions for specified product in crash reports reported in the last
-    # 6 months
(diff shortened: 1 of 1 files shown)


willkg:
stage has the same problem as prod where there's a 66.0b6, but no 66.0b7, yet. 66.0b7 just released yesterday, so there aren't enough crash reports to make it pop up in the list, yet.

If I land this quickly, I can verify whether my theories are correct and that this fixes the issue on stage.

lonnen:
🥇 👍

willkg:
This looks good to me and I think it'll fix the issue. Even if it doesn't fix the issue, I'm pretty sure it won't make anything worse. self-r+

PR #4827, opened 2019-02-15
fix bug 1528388: add breakpad binaries to socorro_app image

willkg:


willkg:
This won't affect server environments at all. I don't think it increases the image size much. Seems like 30mb or so locally. self-r+

PR #4828, opened 2019-02-21
bug 1529342: redo logging in socorro apps

willkg:
This nixes the `config.logger` and `config.metrics` items. This adds instance loggers to many components that are logging. One nice thing about this is that it means logged messages have a more helpfully named logger.

This also fixes all the logging-related tests to use pytest's caplog.

This is a first pass on removing configman--we had to switch to using a logger that's not carried around in config.

### DIFF
--- a/socorro/app/fetch_transform_save_app.py
+++ b/socorro/app/fetch_transform_save_app.py
@@ -27,7 +27,6 @@ sending new crash records to Postgres; sending the processed crash to HBase;
 the the submission of the crash_id to Elastic Search."""
 
 from functools import partial
-import logging
 import signal
 
 from configman import Namespace
@@ -35,9 +34,6 @@ from configman.converters import class_converter
 
 from socorro.lib.task_manager import respond_to_SIGTERM
 from socorro.app.socorro_app import App
-
-
-logger = logging.getLogger(__name__)
 
 
 class FetchTransformSaveApp(App):
@@ -86,7 +82,7 @@ class FetchTransformSaveApp(App):
     )
 
     def __init__(self, config):
-        super(FetchTransformSaveApp, self).__init__(config)
+        super().__init__(config)
         self.waiting_func = None
         # select the iterator type based on the "number_of_submissions" config
         self.source_iterator = {
@@ -234,12 +230,7 @@ class FetchTransformSaveApp(App):
                 # we're running all in the same thread, a failure here could
                 # derail the the whole processor. Best just log the problem
                 # so that we can continue.
-                logger.error(
-                    'Error completing job %s: %s',
-                    crash_id,
-                    x,
-                    exc_info=True
-                )
+                self.logger.error('Error completing job %s: %s', crash_id, x, exc_info=True)
 
     def _transform(self, crash_id):
         """this default transform function only transfers raw data from the
@@ -249,39 +240,23 @@ class FetchTransformSaveApp(App):
         try:
             raw_crash = self.source.get_raw_crash(crash_id)
         except Exception as x:
-            logger.error(
-                "reading raw_crash: %s",
-                str(x),
-                exc_info=True
-            )
+            self.logger.error("reading raw_crash: %s", str(x), exc_info=True)
             raw_crash = {}
         try:
             dumps = self.source.get_raw_dumps(crash_id)
         except Exception as x:
-            logger.error(
-                "reading dump: %s",
-                str(x),
-                exc_info=True
-            )
+            self.logger.error("reading dump: %s", str(x), exc_info=True)
             dumps = {}
         try:
             self.destination.save_raw_crash(raw_crash, dumps, crash_id)
-            logger.info('saved - %s', crash_id)
+            self.logger.info('saved - %s', crash_id)
         except Exception as x:
-            logger.error(
-                "writing raw: %s",
-                str(x),
-                exc_info=True
-            )
+            self.logger.error("writing raw: %s", str(x), exc_info=True)
         else:
             try:
                 self.source.remove(crash_id)
             except Exception as x:
-                logger.error(
-                    "removing raw: %s",
-                    str(x),
-                    exc_info=True
-                )
+                self.logger.error("removing raw: %s", str(x), exc_info=True)
 
     def quit_check(self):
         self.task_manager.quit_check()
@@ -299,10 +274,7 @@ class FetchTransformSaveApp(App):
                 quit_check_callback=self.quit_check
             )
         except Exception:
-            logger.critical(
-                'Error in creating crash source',
-                exc_info=True
-            )
+            self.logger.critical('Error in creating crash source', exc_info=True)
             raise
         try:
             self.destination = self.config.destination.crashstorage_class(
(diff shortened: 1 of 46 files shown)


socorro/app/fetch_transform_save_app.py
```diff
@@ -36,6 +37,9 @@
 from socorro.app.socorro_app import App
 
 
+logger = logging.getLogger(__name__)
```
willkg: Since the classes in here derive from `App` and that has a `self.logger`, we can use that logger rather than the module-level one here.

socorro/app/socorro_app.py
```diff
@@ -195,25 +224,28 @@ def fix_exit_code(code):
...
+
             config.executor_identity = (
                 lambda: threading.currentThread().getName()
             )
 
             # Log revision information
             revision_data = get_revision_data()
             revision_items = sorted(revision_data.items())
-            config.logger.info(
+            logger.info(
```
willkg: This should use a different logger--not a module-level one.

socorro/app/socorro_app.py
```diff
@@ -319,35 +330,13 @@ def filter(self, record):
...
-    # If the user is setting the root level to something below or equal to
-    # logging level, then we want to stop propagating some loggers. This only
-    # happens in local development environments.
-    if logging_root_level <= logging_level:
-        logging_config['root'] = {
-            'handlers': ['console'],
-            'level': logging_root_level,
         }
-        logging_config['loggers']['socorro']['propagate'] = 0
-        logging_config['loggers'][app_name]['propagate'] = 0
```
willkg: We needed this for tests and local dev environment. Using caplogpp lets us not need to do this. That lets us ditch the `logging_root_level` configuration, too.

socorro/app/socorro_app.py
```diff
@@ -375,64 +364,3 @@ def setup_metrics(config, local_unused, args_unused):
             raise ValueError('Invalid markus backend "%s"' % backend)
 
     markus.configure(backends=backends)
-
-    return markus.get_metrics('')
-
-
-class App(SocorroApp):
```
willkg: I folded this into `SocorroApp` and then renamed that as just `App` so we didn't have to make changes across the codebase.

socorro/cron/base.py
```diff
@@ -156,6 +157,7 @@ class BaseCronApp(RequiredConfig):
     def __init__(self, config, job_information):
         self.config = config
         self.job_information = job_information
+        self.logger = logging.getLogger(__name__ + '.' + self.__class__.__name__)
```
willkg: I thought about moving this to a lib module because it's kind of icky. So then we could have `get_instance_logger()` and `get_class_logger`, but it'd have to look at the stack to figure out the right class and `__name__` and I decided that's probably not wildly helpful. We could also make a mixin that has the methods, but that seemed unenthusing, too.

socorro/external/postgresql/connection_context.py
```diff
@@ -105,7 +107,7 @@ def __call__(self, name=None):
         try:
             yield conn
         finally:
-            # self.config.logger.debug('connection closed')
+            # self.logger.debug('connection closed')
```
willkg: We should nix this.

socorro/processor/symbol_cache_manager.py
```diff
@@ -50,7 +51,7 @@ def process_IN_DELETE(self, event):
                 sys.stdout.write('D')
                 sys.stdout.flush()
             elif self.verbosity == 2:
-                self.monitor.config.logger.debug('D  %s', event.pathname)
+                self.monitor.debug('D  %s', event.pathname)
```
willkg: Oops--this is wrong. It should be `self.monitor.logger.debug`.

socorro/unittest/__init__.py
```diff
@@ -17,3 +20,10 @@ def __lt__(self, other):
 #: Sentinel that is equal to anything; simplifies assertions in cases where
 #: part of the value changes from test to test
 WHATEVER = EqualAnything()
+
+
+@contextlib.contextmanager
+def mock_logging(logger_name, fun_name):
+    logger = logging.getLogger(logger_name)
+    with mock.patch.object(logger, fun_name) as mock_logging_fun:
+        yield mock_logging_fun
```
willkg: I used this originally, then switched to caplogpp, so we don't need this.

socorro/unittest/app/test_socorro_app.py
```diff
@@ -157,81 +157,3 @@ def main(self):
...
-class TestSocorroAppMetrics(object):
-    def test_metrics(self, metricsmock):
-        """Verify LoggingMetrics work"""
-        with metricsmock as mm:
-            AppWithMetrics.run(values_source_list=[configman_keys({})])
-
-            assert mm.has_record('incr', stat='increment_key', value=1)
-            assert mm.has_record('gauge', stat='gauge_key', value=10)
-            assert mm.has_record('timing', stat='timing_key', value=100)
-            assert mm.has_record('histogram', stat='histogram_key', value=1000)
```
willkg: We don't do this anymore, so we can nix this completely.

socorro/unittest/app/test_socorro_app.py
```diff
@@ -157,81 +157,3 @@ def main(self):
...
-
-
-def test_setup_logger():
-    """Verify setup_logger with level and root_level variations"""
```
willkg: We don't need this anymore, so we can nix this.

socorro/unittest/external/rabbitmq/test_crashstorage.py
```diff
@@ -149,19 +148,21 @@ def test__ack_crash(self):
...
         crash_store = RabbitMQCrashStorage(config)
         crash_store.acknowledgment_queue.put('b2')
         crash_store._consume_acknowledgement_queue()
 
-        config.logger.warning.assert_called_once_with(
-            'RabbitMQCrashStorage tried to acknowledge crash %s'
-            ', which was not in the cache',
-            'b2',
-            exc_info=True
+        recs = [(rec.message, rec.exc_info) for rec in caplogpp.records]
```
willkg: I kind of hate how caplog isn't helpful in regards to asserting things about the logged messages. I was tossing around wrapping that up with caplogpp into a library similar to the mock logger I wrote in Antenna:

https://github.com/mozilla-services/antenna/blob/master/testlib/loggingmock.py

I'll toss that in my list of things to think about.

socorro/unittest/processor/rules/test_memory_report_extraction.py
```diff
@@ -22,8 +21,6 @@ def get_example_file_data(filename):
 class TestMemoryReportExtraction(object):
     def get_config(self):
         fake_processor = create_basic_fake_processor()
-        # FIXME(willkg): we should switch logging testing to use pytest's caplog
```
willkg: Done! Yay!

socorro/unittest/processor/test_processor_app.py
```diff
@@ -300,11 +305,17 @@ def test_transform_polystorage_error_with_raven_configured_failing(self, mock_ra
...
-        assert (
-            config.logger.error.call_args_list == [
-                mock.call('Unable to report error with Raven', exc_info=True),
-                mock.call('Exception occurred', exc_info=first_exc_info),
-                mock.call('Unable to report error with Raven', exc_info=True),
-                mock.call('Exception occurred', exc_info=second_exc_info)
-            ]
-        )
+        expected = [
+            ('Unable to report error with Raven', WHATEVER),
```
willkg: I <3 `WHATEVER`. It makes some things so much easier to deal with.

socorro/app/fetch_transform_save_app.py
```diff
@@ -82,7 +82,7 @@ class FetchTransformSaveApp(App):
     )
 
     def __init__(self, config):
-        super(FetchTransformSaveApp, self).__init__(config)
+        super().__init__(config)
```
willkg: I should do a pass on converting `super()` calls to Python 3 format. It's much easier to deal with.

willkg:
Two thoughts:

1. One thing I didn't do was audit *what* was getting logged. Are those helpful messages? Are there other things that should get logged? Are messages getting logged at appropriate logging levels?
2. I'm not thrilled about instance-scoped loggers, but that seems ok for now and it's nice that you can see what's logging messages in the logged messages. We can always change that later--it's a small change.

This seems ok. I ran everything I could locally and ran the tests. I don't think that covers everything, so I think I want to let this run on stage for a few days. self-r+

PR #4829, opened 2019-02-21
bug 1529342: remove sighup handling

willkg:
We no longer run Socorro apps in a way that requires they handle `SIGHUP` and restart with updated configuration. Given that, we can remove this code.

### DIFF
--- a/socorro/app/socorro_app.py
+++ b/socorro/app/socorro_app.py
@@ -17,8 +17,6 @@ all the Socorro Apps derive.
 import logging
 import logging.config
 import logging.handlers
-import functools
-import signal
 import os
 import socket
 import sys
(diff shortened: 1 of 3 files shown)


socorro/app/socorro_app.py
```diff
@@ -150,17 +132,10 @@ def _app_instance_name(self):
...
-            restart = False
-            app_exit_code = cls._do_run(
-                config_path=config_path,
-                values_source_list=values_source_list
-            )
-        return app_exit_code
+        return cls._do_run(
+            config_path=config_path,
+            values_source_list=values_source_list
+        )
```
willkg: This seems silly. I can probably nix this.

willkg:
This looks good to me. self-r+

PR #4833, opened 2019-02-25
bug 1529342 - remove change default, executor_identity, ThreadedTaskManagerWithConfigSetup

willkg:
Remove `change_default`, `executor_identity` and `ThreadedTaskManagerWithConfigSetup`.


socorro/external/rabbitmq/crashstorage.py
```diff
@@ -39,8 +35,9 @@ class RabbitMQCrashStorage(CrashStorageBase):
         'rabbitmq_class',
         # we choose a pooled connection because we need thread safe connection
         # behaviors
-        default=ConnectionContextPooled,
+        default='socorro.external.rabbitmq.connection_context.ConnectionContextPooled',
```
willkg: I changed this to a Python module string that gets converted with the `class_converter` which is how we do all the other classes like this one.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -223,29 +220,28 @@ def _ack_crash(self, connection, crash_id, acknowledgement_token):
...
-    required_config.rabbitmq_class = change_default(
-        RabbitMQCrashStorage,
+    required_config.add_option(
         'rabbitmq_class',
-        ConnectionContext,
-    )
-    required_config.routing_key = change_default(
-        RabbitMQCrashStorage,
-        'routing_key',
-        'socorro.reprocessing'
```
willkg: We don't need this since the parent class already has a `routing_key` of `socorro.reprocessing`.

willkg:
Seems pretty straight-forward. self-r+

PR #4834, opened 2019-02-26
fix bug 1530674: fix version sorting to ignore invalid versions

willkg:
This adds a really basic "version validator" and tweaks the version
sorting code to ignore invalid versions.


willkg:
self-r+

PR #4835, opened 2019-02-27
bug 1529342: redo Elasticsearch-related configuration

willkg:
This does a couple of things:

1. remove Elasticsearch requirements from `TelemetryS3CrashStorage`
2. rework all the Elasticsearch-related classes fixing configuration and Law of Demeter type issues

This second one covers these:

* get rid of ElasticsearchConfig which duplicated options, had options that weren't used, and made it hard to reason about code that used it
* split SuperSearchFields into a few separate classes:
  * SuperSearchFieldsData which just has the fields in it
  * SuperSearchFields which takes a ConnectionContext
  * SuperSearchFieldsModel which is a valid middleware model that takes a config
* fix class names for test classes that weren't getting picked up by pytest and thus weren't being tested (and then fixed the tests because they were broken)
* folded IndexCreator into ConnectionContext
* added several getter methods to ConnectionContext to reduce Law of Demeter type problems
* fixed Query and SuperSearch to have configuration
* deleted the duplicate get_list_of_indices
* remove use of Elasticsearch from socorro.lib.search_common--now it's search-agnostic and no longer requires configuration

Why do this? Because in cleaning these issues up, it asserts ownership and responsibility of the classes involved and reduces the number of call-throughs. That makes it easier to read and write tests and set up the context these classes require. It'll make it easier to convert to a new configuration system in the future.


willkg:
I have two concerns with these changes. First, they're pretty extensive--this is a bigger PR than I would like. Second, because I'm moving classes around and adjusting configuration, I don't have a good feel for whether I've changed the overall configuration of the system. I don't think I have.

This runs fine in a local development environment, but may have issues in a server environment which has a different configuration.

I didn't change any of the configuration files, so I think this is ok, but it's hard to know for sure.

I need to write up a test plan for this.

socorro/cron/jobs/update_signatures.py
```diff
@@ -48,7 +47,8 @@ class UpdateSignaturesCronApp(BaseCronApp):
     )
     required_config.add_option(
         'elasticsearch_class',
-        default=ElasticsearchConfig,
+        default='socorro.external.es.connection_context.ConnectionContext',
```
willkg: In configuration, "class" indicates a Python dotted module name--not an actual class. I fixed this to take a Python dotted module name.

socorro/cron/jobs/update_signatures.py
```diff
@@ -98,8 +99,8 @@ def run(self, end_datetime):
 
         # Do a super search and get the signature, buildid, and date processed for
         # every crash in the range
-        all_fields = SuperSearchFields(config=self.config).get()
-        api = SuperSearch(config=self.config)
+        all_fields = SuperSearchFieldsData().get()
```
willkg: `SuperSearchFieldsData` is **just** the fields now and doesn't need configuration to instantiate.

socorro/external/boto/crashstorage.py
```diff
@@ -218,17 +218,10 @@ class TelemetryBotoS3CrashStorage(BotoS3CrashStorage):
         from_string_converter=class_converter,
         reference_value_from='resource.boto'
     )
-    required_config.elasticsearch = Namespace()
-    required_config.elasticsearch.add_option(
-        'elasticsearch_class',
-        default='socorro.external.es.connection_context.ConnectionContext',
-        from_string_converter=class_converter,
-        reference_value_from='resource.elasticsearch',
-    )
```
willkg: We don't need this anymore because `TelemetryS3CrashStorage` uses `SuperSearchFieldsData` which doesn't need a `ConnectionContext` or configuration.

socorro/external/es/base.py
```diff
@@ -32,49 +30,3 @@ def generate_list_of_indexes(from_date, to_date, index_format):
...
-    required_config.add_option(
-        'search_default_date_range',
-        default=7,  # in days
-        doc='the default date range for searches, in days'
-    )
-    required_config.add_option(
-        'search_maximum_date_range',
-        default=365,  # in days
-        doc='the maximum date range for searches, in days'
-    )
```
willkg: This had a number of duplicate options. Some options were defined here, but should have been elsewhere--closer to where they're used. This also includes the `IndexCreator` and `ConnectionContext`, but I merged those. At that point, there was no reason to keep this around.

socorro/external/es/connection_context.py
```diff
@@ -68,6 +101,18 @@ def connection(self, name=None, timeout=None):
...
+        """Return template for index names."""
+        return self.config.elasticsearch_index
+
+    def get_doctype(self):
+        """Return doctype."""
+        return self.config.elasticsearch_doctype
+
+    def get_timeout_extended(self):
+        """Return timeout_extended."""
+        return self.config.elasticsearch_timeout_extended
```
willkg: Getter methods for accessing things in the configuration for this class.

socorro/external/es/query.py
```diff
@@ -17,8 +19,16 @@
 from socorro.lib import datetimeutil, external_common
 
 
-class Query(object):
-    """Implement the /query service with ElasticSearch"""
+class Query(RequiredConfig):
```
willkg: All middleware models get `config` passed to them. Therefore they should all derive from `RequiredConfig` and specify what configuration they're looking for. This fixes `Query` accordingly.

socorro/external/es/super_search_fields.py
```diff
@@ -109,23 +118,12 @@ def add_doc_values(value):
             add_doc_values(field)
 
 
-class SuperSearchFields(object):
-    # Defining some filters that need to be considered as lists.
-    filters = [
-        ('form_field_choices', None, ['list', 'str']),
-        ('permissions_needed', None, ['list', 'str']),
-    ]
+class SuperSearchFieldsData(object):
```
willkg: Having a data-only class makes it easier for other things to use the fields without having to configure an Elasticsearch `ConnectionContext`.

socorro/external/es/super_search_fields.py
```diff
@@ -135,6 +133,15 @@ def get_fields(self):
     # can be conveniently subclassed by ``SuperSearchMissingFields``.
     get = get_fields
 
+
+class SuperSearchFields(SuperSearchFieldsData):
```
willkg: This class takes a `ConnectionContext`. This is not a middleware model.

socorro/external/es/super_search_fields.py
```diff
@@ -304,15 +308,39 @@ def test_mapping(self, mapping):
...
+                self.context.indices_client().delete(temp_index)
             except elasticsearch.exceptions.NotFoundError:
                 # If the index does not exist (if the index creation failed
                 # for example), we don't need to do anything.
                 pass
 
 
-class SuperSearchMissingFields(SuperSearchFields):
-    """Model that returns fields missing from super search fields"""
+class SuperSearchFieldsModel(RequiredConfig, SuperSearchFields):
```
willkg: This is the middleware model that takes a config.

socorro/external/es/super_search_fields.py
```diff
@@ -304,15 +308,39 @@ def test_mapping(self, mapping):
...
+    )
+
+    def __init__(self, config):
+        # NOTE(willkg): This doesn't call the super().__init__ but instead
+        # sets everything up itself.
+        self.config = config
+        self.context = self.config.elasticsearch_class(self.config)
+
+
+class SuperSearchMissingFieldsModel(SuperSearchFieldsModel):
```
willkg: Figured I'd rename `SuperSearchMissingFields` in kind.

socorro/external/es/supersearch.py
```diff
@@ -25,46 +26,27 @@
...
 
 
-class SuperSearch(SearchBase):
-
-    def __init__(self, *args, **kwargs):
-        self.config = kwargs.get('config')
-        self.es_context = self.config.elasticsearch.elasticsearch_class(
-            self.config.elasticsearch
-        )
+class SuperSearch(RequiredConfig, SearchBase):
```
willkg: `SuperSearch` is a middleware model, so it needs to define its configuration.

socorro/external/es/supersearch.py
```diff
@@ -25,46 +26,27 @@
...
-        super().__init__(*args, **kwargs)
+    def __init__(self, config):
+        """Create a SuperSearch instance.
 
-    def get_connection(self):
-        with self.es_context() as conn:
-            return conn
+        :arg context: an Elasticsearch ConnectionContext instance
 
-    def get_list_of_indices(self, from_date, to_date, es_index=None):
```
willkg: This was defined in `socorro.external.es.base` already, so I nixed this one.

socorro/lib/search_common.py
```diff
@@ -54,6 +54,14 @@
 }
 
 
+# Default date range for searches in days
+DEFAULT_DATE_RANGE = 7
+
+
+# Maximum date range for searches in days
+MAXIMUM_DATE_RANGE = 365
```
willkg: I took these out of configuration and made them constants. We're never going to change these. Seems like they make for better constants.

socorro/lib/search_common.py
```diff
@@ -103,10 +111,6 @@ class SearchBase(object):
         SearchFilter('_sort', default=''),
     )
 
-    def __init__(self, *args, **kwargs):
-        self.context = kwargs.get('config')
-        self.config = self.context
```
willkg: I nixed the config requirement for this class.

socorro/unittest/external/es/base.py
```diff
@@ -146,14 +132,14 @@ def get_tuned_config(self, sources, extra_values=None):
 
         return super().get_tuned_config(sources, values_source)
 
-    def get_base_config(self, es_index=None):
```
willkg: Passing in the class to get the tuned config for here makes this more generally useful and fixes a bunch of tests that were busted.

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -190,19 +188,8 @@ def setup_method(self, method):
...
-    def teardown_method(self, method):
-        """Remove indices that may have been created by the test """
-        try:
-            self.index_client.delete(self.config.elasticsearch.elasticsearch_index)
-
-        except elasticsearch.exceptions.NotFoundError:
-            # It's fine it's fine; 404 means the test didn't create any
-            # indices, therefore they can't be deleted.
-            pass
-        super().teardown_method(method)
```
willkg: The `super().teardown_method` already does this. No need to do it twice.

socorro/unittest/external/es/test_supersearch.py
```diff
@@ -1699,24 +1701,6 @@ def test_get_with_columns(self):
...
-        assert 'cpu_name' not in res['hits'][0]
-
-        # Test with 2 synonyme fields.
-        kwargs = {
-            '_columns': ['cpu_name', 'cpu_arch']
-        }
-        res = self.api.get(**kwargs)
-
-        assert 'cpu_name' in res['hits'][0]
-        assert 'cpu_arch' in res['hits'][0]
```
willkg: We don't have synonyms anymore, so I nixed this.

webapp-django/crashstats/supersearch/models.py
```diff
@@ -38,22 +38,6 @@
...
-    first having instanciated the base class with the necessary configuration
-    needed to reach Elasticsearch.
-
-    This way we can call `.get()` on an instance of this class, which
-    just returns the fields.
-    """
-
-    def __init__(self, *args, **kwargs):
-        # Deliberately don't do anything
-        pass
```
willkg: This was a hack to get around using just the fields from `SuperSearchFields`. Now we have `SuperSearchFieldsData`, so we can just use that.

willkg:
I want to land this after landing something else and then doing a prod deploy. That way this can sit on stage for the weekend.

willkg:
I did the prod push. This looks good. self-r+

PR #4836, opened 2019-02-28
bug 1513346: import Queue from non-six place

willkg:


willkg:
This is pretty benign. self-r+

PR #4839, opened 2019-03-09
Fix db setup

willkg:
Running "python socorro/scripts/db.py" does nothing--it just loads
the module. We need to run "socorro-cmd db" now. This fixes that.


willkg:
self-r+

PR #4840, opened 2019-03-09
fix bug 1528243: verify crashes are processed

willkg:
This creates a crontabber app that goes through all the raw crashes for a given day and verifies there are corresponding processed crashes.

It's a little wonky to run since it's a crontabber app. When we convert crontabber apps to Django command lines, it'll be less wonky.


willkg:
To test this:

```
$ make shell
app@030a58eeed80:/app$ socorro-cmd crontabber --reset-job=verifyprocessed
app@030a58eeed80:/app$ socorro-cmd crontabber --force --job=verifyprocessed --crontabber.class-VerifyProcessedCronApp.date=20190307
```

You can set up the local S3 bucket by creating files in a directory and then syncing them over with `./scripts/socorro_aws_s3.sh`.

socorro/cron/crontabber_app.py
```diff
@@ -1093,5 +1093,8 @@ def sentrytest(self):
     'socorro.cron.jobs.update_signatures.UpdateSignaturesCronApp|1h',
 
     # Dependency checking
-    'socorro.cron.jobs.monitoring.DependencySecurityCheckCronApp|1d',
+    'socorro.cron.jobs.monitoring.DependencySecurityCheckCronApp|1d|05:00',
+
+    # Processed crash verification
+    'socorro.cron.jobs.verify_processed.VerifyProcessedCronApp|1d|04:00',
```
willkg: On my local machine, this takes 7 minutes to run. That's entirely iterating through the 4096 entropy possibilities and doing a listing for each one. I suspect it'll take longer on stage and even longer in prod, but I don't think it'll take more than an hour.

socorro/cron/jobs/verify_processed.py
```diff
@@ -0,0 +1,151 @@
...
+        chunksize = 1
+
+        check_crashids_for_date = partial(
+            check_crashids,
+            boto_conn=boto_conn,
+            bucket_name=bucket_name,
+            date=date
+        )
+
+        pool = multiprocessing.Pool(num_workers)
```
willkg: This uses multiprocessing to split the task into smaller ones that run in multiple processes getting around Python's GIL and making better use of waiting around for network requests to happen. This can probably be tweaked--I picked 5 and 1 because that seemed fine.

socorro/cron/jobs/verify_processed.py
```diff
@@ -0,0 +1,151 @@
...
+        metrics.gauge('missing_processed', len(missing))
+        if missing:
+            for crash_id in missing:
+                self.logger.info('Missing: %s', crash_id)
+
+                with transaction_context(self.database) as conn:
+                    sql = """
+                    INSERT INTO crashstats_missingprocessedcrashes (crash_id, created)
+                    VALUES (%s, current_timestamp)
+                    """
```
willkg: For now, it's dumping the crash ids in a db table. I'll spin off a new bug to create a Django admin action for reprocessing items in the table and one for checking the items in the table. I think that gives us everything we need to follow up on things.

willkg:
self-r+

PR #4841, opened 2019-03-11
fix bug 1534402: fix error handling in multiprocessing code

willkg:
This adds Sentry error reporting for unhandled errors in the multiprocessing code. It also changes the number of workers and makes that configurable. This particular problem is network-i/o-heavy, so more worker processes speeds that up significantly.


willkg:
self-r+

PR #4844, opened 2019-03-12
bug 1534617: add "is_processed" check

willkg:
When looking at `MissingProcessedCrashes`, in the Django admin, this now
shows the collected date as well as whether the crash has been processed.
This will make it easier to deal with missing crashes.


willkg:
self-r+

PR #4845, opened 2019-03-14
bug 1527345: redo FetchTransformSaveApp and queuing

willkg:
Socorro only has one fetch-transform-save app: the processor. Given that, I collapsed the  `FetchTransformSaveAppWithNewCrashSource` into `FetchTransformSaveApp`. I removed unused bits.

I also renamed all the `new_crash_source` things as `queue` across files, configuration, classes and variable names.

I extracted the "consume from a queue" things from crash storage classes so they're less intertwined. That establishes an API for queuing which we can use for Pub/Sub.


docs/crashstorage/crashstorage.rst
```diff
@@ -144,7 +144,7 @@ socorro.external.rabbitmq
   pool.
 
 
-**socorro.external.rabbitmq.rmq_new_crash_source**
+**socorro.external.rabbitmq.crash_source**
```
willkg: Oops--that should be `crash_queue`.

socorro/app/fetch_transform_save_app.py
```diff
@@ -37,180 +37,71 @@
...
-    )
 
     def __init__(self, config):
         super().__init__(config)
         self.waiting_func = None
-        # select the iterator type based on the "number_of_submissions" config
-        self.source_iterator = {
-            'forever': self._infinite_iterator,
-            'all': self._all_iterator,
-        }.get(config.number_of_submissions, self._limited_iterator)
```
willkg: We only use `forever` except for a few tests which I reworked. Given that, I nixed the other iterators.

socorro/app/fetch_transform_save_app.py
```diff
@@ -37,180 +37,71 @@
...
-    # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-
-    def _create_iter(self):
-        # the actual mechanism of creating the iterator to be overridden in
-        # subclasses based on what they want to iterate over.  In this default,
-        # iteration will come from the crashstorage instance defined as
-        # "source".  This works for the crashmover where the generally used
-        # source is the file system walking class.
-        return self.source.new_crashes()
+        self.source_iterator = self._infinite_iterator
```
willkg: I can probably nix this line and rename `_infinite_iterator` to `source_iterator`.

socorro/app/fetch_transform_save_app.py
```diff
@@ -37,180 +37,71 @@
...
-    def _action_after_iteration_completes(self):
-        """an action to be done when the iterator returned by the
-        "_create_iter" method is exhausted.  This method is to be overridden
-        in base classes that might need to do something special at that point.
-        examples may be simple logging, waiting, or resetting the iteration
-        system for the next loop."""
-        pass
+    def get_crashids(self):
+        """Return the queue iterator."""
+        return self.queue.new_crashes()
```
willkg: I can probably nix this method, too.

socorro/app/fetch_transform_save_app.py
```diff
@@ -324,48 +207,3 @@ def main(self):
...
-            self.new_crash_source = self.source
-
-    def close(self):
-        super().close()
-        if self.source != self.new_crash_source:
-            try:
-                self.new_crash_source.close()
-            except AttributeError:
-                # the new_crash_source has no close, move on without it
-                pass
```
willkg: I folded all this into `FetchTransformSaveApp`.

socorro/external/rabbitmq/crashqueue.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+class RabbitMQCrashQueue(RequiredConfig):
+    """Crash queue."""
+
+    required_config = Namespace()
+    required_config.add_option(
+        'rabbitmq_class',
+        # we choose a pooled connection because we need thread safe connection
+        # behaviors
```
willkg: We don't need this comment.

socorro/external/rabbitmq/crashqueue.py
```diff
@@ -0,0 +1,167 @@
...
+            yield (
+                (body,),
+                {'finished_func': partial(self.ack_crash, body)}
+            )
+            queues.reverse()
+
+    def ack_crash(self, crash_id):
+        self.acknowledgment_queue.put(crash_id)
+
+    def _suppress_duplicate_jobs(self, crash_id, acknowledgement_token):
```
willkg: I think we should nix this code. We have multiple processor nodes, so the idea of "duplicate jobs" really needs to be global across the cluster and not local to a single node. I checked the logs on stage and prod and see one instance of "duplicate job:" in the last 6 days.

Duplicate jobs occur when someone triggers a priority processing or we're doing multiple rounds of reprocessing. The latter would be entirely my fault. Given the size of the priority queue most of the time, I don't think it happens much.

I'll leave it for now since this code isn't going to last long anyhow, but I don't think I'll have this in the Pub/Sub implementation.

socorro/external/rabbitmq/crashstorage.py
```diff
@@ -55,12 +40,6 @@ class RabbitMQCrashStorage(CrashStorageBase):
 
     def __init__(self, config, namespace='', quit_check_callback=None):
         super().__init__(config, namespace=namespace, quit_check_callback=quit_check_callback)
-
-        # Note: this may continue to grow if we aren't acking certain UUIDs.
-        # We should find a way to time out UUIDs after a certain time.
-        self.acknowledgement_token_cache = {}
-        self.acknowledgment_queue = Queue()
```
willkg: Consuming the queue is no longer in the domain of the crashstore, so we don't need to deal with acks.

willkg:
self-r+

PR #4846, opened 2019-03-14
Change Buildhub base URL

peterbe:
The old link points to the old web UI for Buildhub1. Changing it to the new one, which incidentally, isn't operating on a path prefix. 

This change would resolve https://github.com/mozilla/buildhub2/issues/366


peterbe:
@willkg Sorry for the laziness but I made this change without testing that the buildID links continue to work on the report index page. (is it still called report index page?)

willkg:
To answer your question, I call it the report view, but I talk to myself, so I don't know if other people have a different name for it.

Also, I figure it's probably fine and we'd test it on stage. That's easier than stashing the state of my local dev environment.

PR #4847, opened 2019-03-15
bug 1527345: implement Pub/Sub queue and scaffolding

willkg:
* add Pub/Sub emulator Docker container
* add configuration
* add `socorro-cmd pubsub` script to manipulate the arcane Pub/Sub energies of the universe
* add `PubSubCrashQueue` and tests


docker/config/never_on_a_server.env
```diff
@@ -27,3 +27,6 @@ resource.boto.secure=false
 # For webapp sessions, we need to allow cookies to be sent insecurely
 # since it's using HTTP and not HTTPS.
 SESSION_COOKIE_SECURE=False
+
+# Set Pub/Sub library to use emulator
+PUBSUB_EMULATOR_HOST=pubsub:5010
```
willkg: We don't want this set in server environments. Hence it being here.

docker/run_tests.sh
```diff
@@ -31,8 +31,13 @@ PYTHON="$(which python)"
 
 # Wait for postgres and elasticsearch services to be ready
 urlwait "${DATABASE_URL}" 10
+urlwait "http://${PUBSUB_EMULATOR_HOST}" 10
```
willkg: This variable is in the environment, but has no protocol since it defines a host and not a url. So, I added the protocol manually.

scripts/process_crashes.sh
```diff
@@ -46,6 +46,7 @@ mkdir "${DATADIR}" || echo "${DATADIR} already exists."
 
 # Add crash ids to queue
 ./socorro-cmd add_crashid_to_queue socorro.normal $@
+# ./socorro-cmd pubsub publish $@
```
willkg: When we switch to Pub/Sub, we can uncomment this line and comment out the above line.

setup.cfg
```diff
@@ -22,4 +22,9 @@ addopts = -rsxX --tb=native -p no:django
 norecursedirs = .git docs config docker __pycache__
 testpaths = socorro/
 # Transform all warnings into errors
-filterwarnings = error
+filterwarnings =
+    error
+    # grpcio kicks this up for some reason in pytest
+    ignore::ImportWarning
+    ignore:::grpcio[.*]
+    ignore:::google[.*]
```
willkg: These packages kick up a bunch of warnings. We have pytest set so that warnings -> errors. In this case, there's nothing we can do here, so we should just silence them for now.

socorro/external/pubsub/crashqueue.py
```diff
@@ -0,0 +1,150 @@
...
+        doc='The subscription name for the reprocessing queue.',
+        reference_value_from='resource.pubsub'
+    )
+
+    def __init__(self, config, namespace='', quit_check_callback=None):
+        self.config = config
+        self.quit_check_callback = quit_check_callback
+
+        if 'PUBSUB_EMULATOR_HOST' in os.environ:
+            self.subscriber = pubsub_v1.SubscriberClient()
```
willkg: If we're using the emulator, we don't need a service account file.

socorro/external/pubsub/crashqueue.py
```diff
@@ -0,0 +1,150 @@
...
+            self.priority_path,
+            self.standard_path,
+            self.reprocessing_path,
+            self.priority_path
+        ]
+
+        for queue_path in queues:
+            response = self.subscriber.pull(
+                queue_path,
+                max_messages=1,
```
willkg: This goes through all the queues and pulls at most one crash id. That way we're not sitting on a whole bunch of crash ids with ticking ack deadlines.

socorro/scripts/pubsub.py
```diff
@@ -0,0 +1,184 @@
...
+    'resource.pubsub.reprocessing_topic_name',
+    'resource.pubsub.reprocessing_subscription_name'
+]
+
+EPILOG = 'Requires %s, and %s to be set in the environment.' % (
+    ', '.join(ENV_VARS[:-1]), ENV_VARS[-1]
+)
+
+# Number of seconds Pub/Sub should wait for a message to be acknowledged
+ACK_DEADLINE = 120
```
willkg: I think 120 seconds works fine for local development. We can use a different value in stage and prod.

willkg:
This includes all the Pub/Sub code for the processor, but doesn't change the processor to use Pub/Sub for queuing. We'll do that later once all the pieces are done.

willkg:
self-r+

PR #4848, opened 2019-03-18
fix bug 1527346: switch webapp to use Pub/Sub

willkg:
This implements the necessary Pub/Sub bits in the webapp. I tried to keep it similar to how RabbitMQ was used with the hope that we could switch back and forth, but that didn't work out so well. The webapp is pretty hard-coded to use one or the other and I didn't want to put the work into abstracting it out since this is an interim problem that will go away soon.

Because the webapp is in Pub/Sub mode, this adjusts the processor to also be in Pub/Sub mode. That means we need to:

1. do the configuration changes in the infra
2. switch Antenna to Pub/Sub mode
3. land this
4. nix socorro-pigeon

all around the same time.

### DIFF
--- a/docker/config/test.env
+++ b/docker/config/test.env
@@ -3,6 +3,26 @@
 # put filesystem crashstorage stuff in /tmp/test_crashes
 resource.fs.fs_root=/tmp/test_crashes
 
+# Statsd
+STATSD_CLIENT=django_statsd.clients.null
+
+# Sentry
+SENTRY_DSN=
+RAVEN_DSN=
+
+# Boto
+AWS_ACCESS_KEY=something
+AWS_SECRET_ACCESS_KEY=something
+# NOTE(willkg): We want boto to use connect_to_region, so we make this an
+# empty string
+AWS_HOST=
+# Use test buckets
+resource.boto.bucket_name=crashstats-test
+TELEMETRY_BUCKET_NAME=telemetry-test
+
+# Elasticsearch configuration
+ELASTICSEARCH_URLS=http://example.com:9123
+
 # Postgres configuration
 database_url=postgres://postgres:aPassword@postgresql:5432/socorro_test
 DATABASE_URL=postgres://postgres:aPassword@postgresql:5432/socorro_test
@@ -24,6 +44,7 @@ resource.pubsub.reprocessing_subscription_name=test_reprocessing_sub
 # Django tests are meant to run with debug mode disabled.
 DEBUG=False
 
-# Use test buckets
-resource.boto.bucket_name=crashstats-test
-TELEMETRY_BUCKET_NAME=telemetry-test
+CACHE_IMPLEMENTATION_FETCHES=True
+DEFAULT_PRODUCT=WaterWolf
+BZAPI_BASE_URL=http://bugzilla.example.com/rest
+SECRET_KEY=fakekey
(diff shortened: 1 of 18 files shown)


willkg:
To test, I did the following:

1. ran `make build`, `make setup`, and `make updatedata`
2. ran `make lint` and `make test`
3. I did `make run`, then processed some crashes, and verified they were processed and could be seen in the webapp.
4. I logged in and reprocessed some crashes via the Reprocess tab.
5. I loaded the raw crash data for a crash report, then pulled up that crash report in the webapp which sends a priority request to process it. I waited a bit then verified that also got processed.

docker/config/test.env
```diff
@@ -3,6 +3,26 @@
...
+AWS_SECRET_ACCESS_KEY=something
+# NOTE(willkg): We want boto to use connect_to_region, so we make this an
+# empty string
+AWS_HOST=
+# Use test buckets
+resource.boto.bucket_name=crashstats-test
+TELEMETRY_BUCKET_NAME=telemetry-test
+
+# Elasticsearch configuration
+ELASTICSEARCH_URLS=http://example.com:9123
```
willkg: I moved some configuration from `webapp-django/crashstats/settings/test.py` into this `test.env` file. We have multiple places for test configuration and I want to keep pushing to reduce it further.

socorro/processor/processor_app.py
```diff
@@ -24,7 +24,7 @@
     'always_ignore_mismatches': True,
 
     'queue': {
-        'crashqueue_class': 'socorro.external.rabbitmq.crashqueue.RabbitMQCrashQueue',
+        'crashqueue_class': 'socorro.external.pubsub.crashqueue.PubSubCrashQueue',
```
willkg: This switches the processor to using `PubSubCrashQueue`.

socorro/unittest/external/pubsub/__init__.py
```diff
@@ -0,0 +1,116 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
willkg: I moved this from `test_crashqueue.py` so I could use it in the webapp, too.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -268,21 +260,12 @@ def config_from_configman():
...
     )
-    # The ReprocessingOneRabbitMQCrashStore crash storage, needs to have
-    # a "logger" in the config object. To avoid having to use the
-    # logger set up by configman as an aggregate, we just use the
-    # same logger as we have here in the webapp.
-    config.queuing.logger = logger
-    config.priority.logger = logger
-    config.crashdata.logger = logger
-    config.telemetrydata.logger = logger
-    return config
```
willkg: A while back, we moved the logger and metrics out of configuration. So we don't need this anymore.

webapp-django/pytest.ini
```diff
@@ -5,4 +5,12 @@
...
-filterwarnings = error
+filterwarnings =
+    error
+    # grpcio kicks this up for some reason in pytest
+    ignore::ImportWarning
+    ignore:::grpcio[.*]
+    ignore:::google[.*]
+    # configman kicks up errors if there's configuration stuff that isn't used
+    # and that breaks the webapp's configman glue
+    ignore:::configman[.*]
```
willkg: We have pytest set up to treat warnings as errors. However, some of the libraries we use kick up warnings and we don't want to treat *those* as errors. Until they get fixed, we'll configure pytest to ignore them.

willkg:
This looks good. I'll talk with Miles about setting up the infrastructure.

willkg:
The infra changes have happened. I'm going to land this and we'll see how it goes. self-r+

PR #4849, opened 2019-03-19
bug 1536147: make archivescraper faster

willkg:
This switches verifyprocessed to use `concurrent.futures` and then restructures archivescraper to also use `concurrent.futures`.

The end result of this on a fresh local dev environment is a massive performance improvement:

* **before**: between 30 and 40 min
* **after**: under 3 min

This will significantly improve development. I slick my environment once a week, so this will save me 30 minutes a week.

archivescraper runs every hour in server environments and takes about 10 min. Since crontabber is synchronous and only runs one job at a time, this blocks all other jobs from running. Improving performance in server environments will improve crontabber job start times.

Using `concurrent.futures` should fix our error handling, too. With the previous implementation using just multiprocessing, we had to have child processes report errors to sentry. I'm pretty sure with this new architecture, the child processes will report exceptions back to the parent which will send them to sentry.

### DIFF
--- a/socorro/cron/jobs/archivescraper.py
+++ b/socorro/cron/jobs/archivescraper.py
@@ -60,9 +60,14 @@ issues.
 
 """
 
+import concurrent.futures
+import copy
+from functools import partial
 import json
+import logging
 
 from configman import Namespace, class_converter
+import more_itertools
 import psycopg2
 from six.moves.urllib.parse import urljoin
 
@@ -79,7 +84,7 @@ with warnings.catch_warnings():
     from pyquery import PyQuery as pq
 
 
-# Substrings that indicate the thing is not a platform we want to traverse
+# Substrings that indicate the directory is not a platform we want to traverse
 NON_PLATFORM_SUBSTRINGS = [
     'beetmover',
     'contrib',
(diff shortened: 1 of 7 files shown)


socorro/cron/jobs/archivescraper.py
```diff
@@ -316,137 +254,272 @@ def scrape_candidates(self, product_name, archive_directory):
...
-                            'release_channel': data['target']['channel'],
-                            'major_version': int(data['target']['version'].split('.')[0]),
-                            'release_version': data['target']['version'],
-                            'build_id': data['build']['id'],
-                            'archive_url': urljoin(self.config.base_url, json_link)
-                        }
+        # build_data is a list of lists so we flatten that
+        return list(more_itertools.flatten(build_data))
+
+    def scrape_candidate_version(self, link, product_name, final_releases):
```
willkg: This method is pretty much the same, but the inputs and outputs were modified.

socorro/cron/jobs/archivescraper.py
```diff
@@ -316,137 +254,272 @@ def scrape_candidates(self, product_name, archive_directory):
...
-
-                        else:
-                            # Insert the rc beta build
-                            data['version_string'] = rc_version_string
-                            self.insert_build(**data)
+                        # Insert the rc beta build
+                        data['version_string'] = rc_version_string
+                        build_data.append(data)
+
+        return build_data
```
willkg: This method used to insert the builds. Now it returns a list of all the builds to be inserted.

socorro/cron/jobs/archivescraper.py
```diff
@@ -316,137 +254,272 @@ def scrape_candidates(self, product_name, archive_directory):
...
+            FROM crashstats_productversion
+            WHERE product_name = %s
+            """
+            params = (product_name,)
+            cursor.execute(sql, params)
+            data = cursor.fetchall()
+            if data:
+                return data[0][0]
+
+        return None
```
willkg: This is the same as before--just moved.

socorro/cron/jobs/archivescraper.py
```diff
@@ -316,137 +254,272 @@ def scrape_candidates(self, product_name, archive_directory):
...
+                %(version_string)s, %(build_id)s, %(archive_url)s
+            )
+            """
+            try:
+                cursor.execute(sql, params)
+            except psycopg2.IntegrityError:
+                # If it's an IntegrityError, we already have it and everything is fine
+                pass
+            except psycopg2.Error:
+                logger.exception('failed to insert')
```
willkg: This is the same as before--just moved.

socorro/cron/jobs/verify_processed.py
```diff
@@ -9,55 +9,41 @@
...
 from socorro.lib.dbutil import execute_no_results
-from socorro.lib.raven_client import capture_error
 from socorro.lib.transaction import transaction_context
 
 
 metrics = markus.get_metrics('crontabber.verifyprocessed')
 
 
-def check_crashids(entropy, boto_conn, bucket_name, date, sentry_dsn):
+def check_crashids(entropy, boto_conn, bucket_name, date):
```
willkg: With `concurrent.futures`, child processes will send exceptions back to the parent to handle. So we don't need the extra error handling here.

webapp-django/crashstats/crashstats/management/commands/dumppv.py
```diff
@@ -0,0 +1,48 @@
...
+        string_buffer = io.StringIO()
+        writer = csv.writer(string_buffer)
+        for pv in pvs:
+            writer.writerow(list(pv))
+
+        lines = string_buffer.getvalue().splitlines()
+        lines.sort()
+
+        with open(fn, 'w') as fp:
+            fp.write('\n'.join(lines) + '\n')
```
willkg: I added this because it'll make debugging archivescraper and testing changes easier. Django has `dumpdata`, but that dumps in JSON and has row ids and isn't convenient to sort (by lines) and compare with.

willkg:
I did some extensive testing comparing the output of this with the output of the original version. This inserts the exact same builds (barring errors from archive.mozilla.org), it just does the scraping concurrently.

This looks good so far. I'll keep an eye on it in stage. self-r+

PR #4852, opened 2019-03-20
bug 1400076: rework collector service to work right

willkg:
This reworks the collector service so it pulls in the latest collector
and runs it with the appropriate bits so that you have a complete
fully functional crash reporting pipeline.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -91,30 +91,28 @@ services:
   # External services
   # -----------------------------
 
-  # https://hub.docker.com/r/mozilla/antenna/
+  # https://hub.docker.com/r/mozilla/socorro_collector/
   #
-  # This seems weird, but for the purposes of Socorro, we should treat
-  # it as an external service.
+  # This pulls the latest Antenna and treats it as an external service.
   #
   # This uses the development ./bin/run_web.sh script from the Antenna
   # container since that creates a bucket before running Antenna.
-  antenna:
-    image: mozilla/antenna:latest
-    environment:
-      - LOGGING_LEVEL=INFO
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -91,30 +91,28 @@ services:
...
-    environment:
-      - LOGGING_LEVEL=INFO
-      - CRASHSTORAGE_CLASS=antenna.ext.s3.crashstorage.S3CrashStorage
-      - CRASHSTORAGE_ENDPOINT_URL=http://localstack-s3:5000
-      - CRASHSTORAGE_REGION=us-west-2
-      - CRASHSTORAGE_ACCESS_KEY=foo
-      - CRASHSTORAGE_SECRET_ACCESS_KEY=foo
-      - CRASHSTORAGE_BUCKET_NAME=dev_bucket
+  collector:
+    image: mozilla/socorro_collector:latest
```
willkg: We renamed the image from "antenna" to "socorro_collector" a while ago. The latter name is functional and easier for normal people to know what it is in the grand scheme of things.

Somehow, I never updated this file, so we've been using an *OLD* version of antenna for a long time. Not that me or anyone else ever really used it in this context because Pigeon made it difficult. But now it's easier with Pub/Sub, so it's more likely I'll use it.

docker-compose.yml
```diff
@@ -91,30 +91,28 @@ services:
...
+    env_file:
+      - docker/config/local_dev.env
+      - docker/config/never_on_a_server.env
+      - my.env
```
willkg: Using the normal env files makes this like the other parts of the pipeline which is handy since they effectively share configuration.

willkg:
This only affects local development. self-r+

PR #4854, opened 2019-03-21
Minor fixes

willkg:
* `--force` doesn't do anything in crontabber, so we don't need to use it
* remove `LOGGING_LEVEL=DEBUG` which makes the webapp very very verbose


willkg:
self-r+

PR #4856, opened 2019-03-21
Switch to readthedocs config file

willkg:
ReadTheDocs supports configuration from the web admin or from a config
file. This switches to the config file so we can track configuration in
version control with everything else.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,20 @@
+# .readthedocs.yml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  builder: html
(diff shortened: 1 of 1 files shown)


willkg:
self-r+

.readthedocs.yml
```diff
@@ -0,0 +1,20 @@
...
+# Required
+version: 2
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  builder: html
+  configuration: docs/conf.py
+
+# Optionally build your docs in additional formats such as PDF and ePub
+formats: []
```
willkg: This is how you say "just the html, please".

.readthedocs.yml
```diff
@@ -0,0 +1,20 @@
...
+
+# Optionally set the version of Python and requirements required to build your docs
+python:
+  version: 3.6
```
willkg: We want to be building with 3.6 because that's what we run.

PR #4858, opened 2019-03-21
bug 1527345: fix PubSubCrashQueue iterator

willkg:
Previously, the iterator would do one pass and then stop. The caller would toss
those crash ids in the work queue and then sleep for 7 seconds. That makes for
really lousy processed crashes throughput.

This fixes that by mimicking the `RabbitMQCrashQueue` and continuing to iterate
between queues until it hits a pass that yields no waiting crash ids. Then it
returns.

### DIFF
--- a/socorro/external/pubsub/crashqueue.py
+++ b/socorro/external/pubsub/crashqueue.py
@@ -113,10 +113,6 @@ class PubSubCrashQueue(RequiredConfig):
         ``finished_func`` as the only key in ``kwargs``. The caller should call
         ``finished_func`` when it's done processing the crash.
 
-        This does a single pass through the queues and pulls at most 10 from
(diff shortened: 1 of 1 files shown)


socorro/external/pubsub/crashqueue.py
```diff
@@ -113,10 +113,6 @@ def __iter__(self):
         ``finished_func`` as the only key in ``kwargs``. The caller should call
         ``finished_func`` when it's done processing the crash.
 
-        This does a single pass through the queues and pulls at most 10 from
-        each queue. It tries twice to pull from the priority queue since it's
-        a priority queue.
-
```
willkg: This isn't true anymore. This was true of a previous implementation that I didn't go with.

willkg:
self-r+

PR #4859, opened 2019-03-21
bug 1536903: fix queryset for processing in admin

willkg:
This was dumb. I thought I tested it, but I must have goofed the testing. Boo.


willkg:
self-r+

PR #4860, opened 2019-03-25
bug 1527345: reduce Pub/Sub pulling

willkg:
This reduces the number of Pub/Sub pulling operations by changing it to pull up
to 5 messages at a time from each queue. Further, it reduces the number of
times it pulls from the priority queue such that it only pulls from the
priority queue once per pass rather than twice per pass.

The theory is that this will increase the throughput for crash processing
on the standard queue which is where the vast majority of our messages
are.

### DIFF
--- a/socorro/external/pubsub/crashqueue.py
+++ b/socorro/external/pubsub/crashqueue.py
@@ -10,6 +10,11 @@ from configman import Namespace, RequiredConfig
 from google.cloud import pubsub_v1
 
 
+# Maximum number of messages to pull from a Pub/Sub topic in a single pull
+# request
+PUBSUB_MAX_MESSAGES = 5
+
+
 logger = logging.getLogger(__name__)
 
 
@@ -118,7 +123,6 @@ class PubSubCrashQueue(RequiredConfig):
             self.priority_path,
             self.standard_path,
             self.reprocessing_path,
-            self.priority_path
         ]
 
         while True:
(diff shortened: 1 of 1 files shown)


socorro/external/pubsub/crashqueue.py
```diff
@@ -10,6 +10,11 @@
 from google.cloud import pubsub_v1
 
 
+# Maximum number of messages to pull from a Pub/Sub topic in a single pull
+# request
+PUBSUB_MAX_MESSAGES = 5
```
willkg: Originally, this was 1. So every time, it does a synchronous blocking I/O pull request for a single message and gets either 0 or 1 messages and it does this a lot. Bleh.

The key issue here is we want standard items and priority items and reprocessing items to get processed evenly-ish. The tricky one is priority crash reports--we want them processed asap. So I think this changes it so that a priority crash report gets processed within a minute and that's probably fine since it doesn't happen often.

socorro/external/pubsub/crashqueue.py
```diff
@@ -118,15 +123,14 @@ def __iter__(self):
             self.priority_path,
             self.standard_path,
             self.reprocessing_path,
-            self.priority_path
```
willkg: Since we switched to a system where we process everything we save, the likelihood there's a crash report that's unprocessed when someone views it on Crash Stats drops significantly. Given that, I took out checking the priority queue twice in a pass. That'll reduce pull requests that yield 0 messages which are effectively wasted work.

willkg:
self-r+

willkg:
I looked at the logs for a bit and my wildly unscientific measuring is that it takes 5 seconds to process a crash and we process 15 crashes a minute on stage before this change. We'll see if this makes that better or worse.

PR #4861, opened 2019-03-25
Add code of coduct file

willkg:


willkg:
Tests won't do anything with this, so self-r+ without running CI.

PR #4862, opened 2019-03-26
bug 1539233: fix table and add missing processed crash check to verifyprocessed

willkg:
This fixes the Django model (and table) name to `MissingProcessedCrash`.

But more importantly, this adds some bookkeeping and updating so that verifyprocessed can let us know if a crash has since been processed. That makes it easier for us to delete them out of the table or manipulate them separately.


webapp-django/crashstats/crashstats/admin.py
```diff
@@ -174,6 +175,8 @@ def process_crashes(modeladmin, request, queryset):
 
 @admin.register(MissingProcessedCrashes)
 class MissingProcessedCrashesAdmin(admin.ModelAdmin):
+    """DEPRECATED."""
```
willkg: I'll take the deprecated things out in a later PR after everything has been pushed to prod.

willkg:
This seems fine. I want to land it on stage and let verifyprocessed run and see if that helps to clear out the 21,000 missing crashes. self-r+

PR #4863, opened 2019-03-26
bug 1539272: switch to 3.6.8-slim-stretch

willkg:
The last time I attempted this, I hit all kinds of problems with compiling the breakpad client and minidump-stackwalk and abandoned the effort. Since then, one of the breakpad client updates we did must have covered all the problems I had because this update was easy peasy.

### DIFF
--- a/docker/run_tests.sh
+++ b/docker/run_tests.sh
@@ -16,7 +16,7 @@
 # Failures should cause setup to fail
 set -v -e -x
 
-echo ">>> pytest"
+echo ">>> set up environment"
 # Set up environment variables
 
 # NOTE(willkg): This has to be "database_url" all lowercase because configman.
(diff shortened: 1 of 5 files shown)


docker/run_tests.sh
```diff
@@ -16,7 +16,7 @@
 # Failures should cause setup to fail
 set -v -e -x
 
-echo ">>> pytest"
+echo ">>> set up environment"
```
willkg: These notes were wrong, so I fixed them.

docker/run_tests_in_docker.sh
```diff
@@ -107,6 +107,6 @@ docker run \
        --tty \
        --interactive \
        --entrypoint= \
-       "${TESTIMAGE}" sh -c /app/docker/run_tests.sh
+       "${TESTIMAGE}" /app/docker/run_tests.sh
```
willkg: This additional `sh -c` caused all the configman environment variables to get dropped because they aren't valid shell variables.

docker/set_up_ubuntu.sh
```diff
@@ -35,6 +35,7 @@ PACKAGES_TO_INSTALL=(
     libcurl3-gnutls
     libcurl4-gnutls-dev
     wget
+    rsync
```
willkg: We additionally need rsync to build the breakpad client.

willkg:
This looks super! self-r+

PR #4865, opened 2019-03-29
bug 1539130: add PolicyException to auditgroups machinery

willkg:
This allows us to have exceptions to the "must be a Mozilla employee" rule
in the data access policy.


willkg:
stage is being used by Pub/Sub testing. Once we push all that to prod, then we can merge things again.

willkg:
This seems good enough base and has the minimal bits we need. self-r+

PR #4866, opened 2019-03-29
bug 1493687: rewrite crontabber as a django command

willkg:
This adds a new `cron` Django command that implements all the things we need from crontabber to schedule and backfill "cron jobs" which are implemented as regular Django commands.


willkg:
stage is being used by Pub/Sub testing. Once we push all that to prod, then we can merge things again.

docker/run_setup.sh
```diff
@@ -24,3 +24,4 @@ set -e
 
 # Initialize the crontabber bookkeeping for all configured jobs to success
 /app/socorro-cmd crontabber --mark-success=all
+/app/webapp-django/manage.py cron --mark-success=all
```
willkg: Over time, we're going to rewrite crontabber commands as cron commands. When we do that, the "job" will move from crontabber space to cron space so it'll only ever exist in one world. So we mark everything a success in each world here and that covers everything.

socorro/cron/base.py
```diff
@@ -17,116 +15,6 @@ class FrequencyDefinitionError(Exception):
...
-
-    try:
-        jobs = dict(zip(jobs.keys(), map(set, jobs.values())))
-        ordered_jobs = list(toposort_flatten(jobs))
-    except ValueError as e:
-        raise CircularDAGError(e)
-
-    return [map_[x] for x in ordered_jobs if x in map_]
-
-
```
willkg: This is dead code since we removed the whole idea of dependencies.

webapp-django/crashstats/cron/management/commands/cron.py
```diff
@@ -0,0 +1,464 @@
...
+
+    if unit == 'h':
+        number *= 60 * 60
+    elif unit == 'm':
+        number *= 60
+    elif unit == 'd':
+        number *= 60 * 60 * 24
+    elif unit:
+        raise FrequencyDefinitionError(value)
+    return number
```
willkg: This is a minor rewrite of the crontabber version. Same thing, same behavior, slightly different implementation. But this one has tests!

webapp-django/crashstats/cron/management/commands/cron.py
```diff
@@ -0,0 +1,464 @@
...
+
+
+def check_time(value):
+    """Validate it's an appropriate time in HH:MM form."""
+    if not VALID_TIME_RE.match(value):
+        raise TimeDefinitionError("Invalid definition of time %r" % value)
+
+    hh, mm = value.split(':')
+    if int(hh) > 23 or int(mm) > 59:
+        raise TimeDefinitionError("Invalid definition of time %r" % value)
```
willkg: This is a minor rewrite of the crontabber version. Same thing, same behavior, slightly different implementation. But this one has tests!

webapp-django/crashstats/cron/management/commands/cron.py
```diff
@@ -0,0 +1,464 @@
...
+    elif unit == 'm':
+        number *= 60
+    elif unit == 'd':
+        number *= 60 * 60 * 24
+    elif unit:
+        raise FrequencyDefinitionError(value)
+    return number
+
+
+VALID_TIME_RE = re.compile(r'^\d\d:\d\d$')
```
willkg: This enforces HH:MM format.

webapp-django/crashstats/cron/management/commands/cron.py
```diff
@@ -0,0 +1,464 @@
...
+    """Return list of matching job specs for this cmd.
+
+    :arg cmds: "all" returns list of all job_specs, cmds return single job_spec,
+        list of cmds returns list of job_specs
+
+    """
+    if cmds == ['all']:
+        return JOBS
+    if isinstance(cmds, str):
+        return JOBS_MAP.get(cmds)
```
willkg: I just double-checked the code and I don't think this can ever happen--it's always a list of cmds. Should fix this and the docstring.

webapp-django/crashstats/cron/management/commands/cron.py
```diff
@@ -0,0 +1,464 @@
...
+    time_ = job_spec.get('time')
+
+    if job.next_run is None:
+        if time_:
+            # Only run if this hour and minute is < now
+            h, m = [int(x) for x in time_.split(':')]
+            if now.hour > h:
+                return True
+            elif now.hour == h and now.minute >= m:
+                return True
```
willkg: Could probably redo this as:

```
if (now.hour, now.minute) >= (h, m):
```

That'll kick off at TIME or later.

willkg: Also, interestingly `check_time` isn't used anywhere. We might want to nix that and have it parse out the time. So then it'd return None for no time, (h, m) for a time, and raise TimeDefinitionError if it parses poorly.

webapp-django/crashstats/cron/management/commands/cron.py
```diff
@@ -0,0 +1,464 @@
...
+
+    return job.next_run < now
+
+
+def get_run_times(job_spec, last_success):
+    """Return generator of run_times for this job."""
+    now = timezone.now()
+
+    # If this is not a backfill job, just run it
+    if not job_spec.get('backfill', False):
```
willkg: This function runs iff it's time for the job to run. So if it's time for the job to run and it's not a backfill job, we can just run it.

webapp-django/crashstats/cron/management/commands/cron.py
```diff
@@ -0,0 +1,464 @@
...
+        # So, reset the hour/minute part to always match the
+        # intention.
+        h, m = [int(x) for x in job_spec['time'].split(':')]
+        when = when.replace(
+            hour=h,
+            minute=m,
+            second=0,
+            microsecond=0
+        )
+    seconds = convert_frequency(job_spec.get('frequency'))
```
willkg: If `frequency` is `None`, then `convert_frequency` will raise an error and this will fail. We should probably have a default frequency of `1d` or something like that.

webapp-django/crashstats/cron/management/commands/cron.py
```diff
@@ -0,0 +1,464 @@
...
+            help='Comma-delimited list of job names to reset or "all" for all.'
+        )
+        parser.add_argument(
+            '--job', type=str, default='',
+            help='Job name for job you want to run.'
+        )
+        parser.add_argument(
+            '--force', action='store_true',
+            help='Force a job to run even if it is not time.'
+        )
```
willkg: I thought about doing this with subparsers. That might be a good idea.

webapp-django/crashstats/cron/management/commands/cron.py
```diff
@@ -0,0 +1,464 @@
...
+            return self.cmd_mark_success(options['mark_success'])
+        elif options['reset_job']:
+            return self.cmd_reset_job(options['reset_job'])
+        elif options['job']:
+            job_args = {}
+            for arg in (options['job_arg'] or []):
+                if '=' in arg:
+                    key, val = arg.split('=', 1)
+                else:
+                    key, val = arg, True
```
willkg: If you do `--job-arg=key=True` and `--job-arg=key`, you get two different values of `True`. I'll have to think about that, but it seems kind of meh.

The other unenthusing thing here is that `key` is the variable name--not the arg name. So if you have an arg like:

```
parser.add_argument('--foo-bar')
```

then you would run that command with `--foo-bar`, but you'd pass in `foo_bar` here as the key.

I'll have to think about that, too.

webapp-django/crashstats/cron/management/commands/cron.py
```diff
@@ -0,0 +1,464 @@
...
+# the key in the db.
+3
+# cmd, frequency, time, backfill
+JOBS = [
+    {
+        # Test cron job
+        'cmd': 'crontest',
+        'frequency': '1d',
+    },
+]
```
willkg: Oh, hrm. I messed up. This needs to support arbitrary arguments to pass to the command. For example, with `auditgroups`, we want to pass in `--persist`.

webapp-django/crashstats/cron/management/commands/cronmarksuccess.py
```diff
@@ -0,0 +1,39 @@
...
+            cmd = job_spec['cmd']
+            self.stdout.write('Marking %s for success at %s...' % (cmd, now))
+
+            next_run = now
+            if job_spec.get('time'):
+                hh, mm = convert_time(job_spec['time'])
+                next_run = next_run.replace(hour=hh, minute=mm, second=0, microsecond=0)
+
+            job = Job.objects.get_or_create(app_name=cmd)[0]
+            job.first_run = now
```
willkg: Oh, whoops--this shouldn't stomp on `first_run`. I'll fix that.

willkg:
This looks good. If there's anything we're missing, it'll come up as we start rewriting jobs with the new system. self-r+

willkg:
Oh, poop. I forgot to commit all the tests. I'll have to do a follow-up PR. :(

PR #4867, opened 2019-03-29
bug 1540248: fix setting up docker-compose in CI

willkg:
The first CI step sets up docker-compose which is then used to set up
a test environment and run the tests. Something changed yesterday or
today and that no longer works.

This updates the first step to what's in CircleCI's documentation.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,23 +2,17 @@ version: 2
 jobs:
   build:
     docker:
-      - image: docker:stable
+      - image: mozilla/cidockerbases:docker-latest
     working_directory: /
+
     steps:
       - run:
+          name: Host info
+          command: uname -v
+
+      - run:
           name: Install essential packages
-          command: |
-            apk update
-            apk add --update --no-cache \
-                    ca-certificates \
-                    build-base \
-                    bash \
-                    make \
-                    git \
-                    openssh \
-                    docker \
-                    py-pip
-            pip install docker-compose
+          command: apt-get install make
 
       - checkout:
           path: /socorro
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -2,23 +2,17 @@ version: 2
 jobs:
   build:
     docker:
-      - image: docker:stable
+      - image: mozilla/cidockerbases:docker-latest
```
willkg: This is what Tecken uses. @mythmon manages it. It looks super!

.circleci/config.yml
```diff
@@ -2,23 +2,17 @@ version: 2
...
     working_directory: /
+
     steps:
+      - run:
+          name: Host info
+          command: uname -v
```
willkg: Figured I'd add this because it makes sleuthing a lot easier.

.circleci/config.yml
```diff
@@ -2,23 +2,17 @@ version: 2
...
-                    ca-certificates \
-                    build-base \
-                    bash \
-                    make \
-                    git \
-                    openssh \
-                    docker \
-                    py-pip
-            pip install docker-compose
+          command: apt-get install make
```
willkg: The new image is Ubuntu-based, so we switch to `apt-get`. We need `make` because that's how I've got everything working.

.circleci/config.yml
```diff
@@ -40,11 +34,17 @@ jobs:
 
       - setup_remote_docker
 
+      - run:
+          name: Get info
+          command: |
+            docker info
+            which docker-compose
+            docker-compose --version
```
willkg: Oh, whoops--I can nix this.

willkg: On second thought, I'll just leave it. The diagnostics help.

willkg:
self-r+

PR #4869, opened 2019-04-02
bug 1540858: fix /api/Reprocessing timeouts

willkg:
Previously, the code published crash ids one at a time waiting for each
to go through. This fixes it to send them in a batch, then make sure
they all went through.

### DIFF
--- a/socorro/external/pubsub/crashqueue.py
+++ b/socorro/external/pubsub/crashqueue.py
@@ -170,6 +170,11 @@ class PubSubCrashQueue(RequiredConfig):
         topic_name = self.config['%s_topic_name' % queue]
         topic_path = publisher.topic_path(project_id, topic_name)
 
+        # Queue up the batch
+        futures = []
         for crash_id in crash_ids:
-            future = publisher.publish(topic_path, data=crash_id.encode('utf-8'))
+            futures.append(publisher.publish(topic_path, data=crash_id.encode('utf-8')))
+
+        # Wait for everything in this group to get sent out
+        for future in futures:
             future.result()


socorro/external/pubsub/crashqueue.py
```diff
@@ -170,6 +170,11 @@ def publish(self, queue, crash_ids):
         topic_name = self.config['%s_topic_name' % queue]
         topic_path = publisher.topic_path(project_id, topic_name)
 
+        # Queue up the batch
+        futures = []
         for crash_id in crash_ids:
-            future = publisher.publish(topic_path, data=crash_id.encode('utf-8'))
+            futures.append(publisher.publish(topic_path, data=crash_id.encode('utf-8')))
```
willkg: The change here looks silly and uninteresting. What's going on is that every call to `.publish()` tosses something in a queue in the library to be sent out in a batch. Batch sending happens in a thread that's spun off and the batch accumulates until at some point it exceeds a time or a space limit at which point the thread sends the batch and fills in all the futures.

What this was doing before was publishing a crash id, that gets tossed in the batch, but then since we demanded a result from the future, the batch of a single crash id would get sent and the future filled. Then we'd do another one.

What this does now is call `.publish()` on each crash id which gets accumulated in the batch. We keep a list of all the futures. Once we're done iterating through all the crash ids, *then* we demand result from the first future which will cause the entire batch to get sent out and all the futures to be filled. That's faster in local development with the Pub/Sub emulator (8s vs 2s with 2 `/api/Reprocessing/` calls of 50 crashes each) and should be significantly faster in a server environment with the real Pub/Sub.

willkg:
self-r+

PR #4875, opened 2019-04-04
bug 1480214: run auditgroups weekly

willkg:


webapp-django/crashstats/authentication/management/commands/auditgroups.py
```diff
@@ -18,6 +20,19 @@
...
+    try:
+        return User.objects.get(username='auditgroups')
+    except User.DoesNotExist:
+        return User.objects.create_user(
+            username='auditgroups',
+            email='auditgroups@example.com',
+            first_name='SYSTEMUSER',
+            last_name='DONOTDELETE',
+            is_active=False,
+        )
```
willkg: So what's going on here is that in order to add an entry to `LogEntry`, it has to be from a user. So the auditgroups command needs to have a user to add things with. This creates a "system user" that's inactive to use for creating `LogEntry` items. That way when auditgroups removes someone from the group, we end up with an audit trail.

webapp-django/crashstats/authentication/models.py
```diff
@@ -24,3 +24,6 @@ class PolicyException(models.Model):
         help_text='Explanation for this exception.'
     )
     created = models.DateTimeField(default=timezone.now)
+
+    def __str__(self):
+        return 'PolicyException <%d, %s>' % (self.id, self.user.email)
```
willkg: This makes `PolicyException` instances more readable when they show up in the admin and elsewhere.

webapp-django/crashstats/crashstats/admin.py
```diff
@@ -35,8 +36,7 @@ class LogEntryAdmin(admin.ModelAdmin):
     list_display = [
         'action_time',
         'user_email',
-        'content_type',
-        'object_repr',
+        'get_edited_object',
```
willkg: This is a lot more useful than content_type + object_repr because the repr for users is useless.

webapp-django/crashstats/cron/__init__.py
```diff
@@ -16,6 +16,14 @@
         'cmd': 'crontest',
         'frequency': '1d',
     },
+
+    {
+        # Audit hackers group
+        'cmd': 'auditgroups',
+        'cmd_args': ['--persist'],
+        'frequency': '7d',
+        'time': '05:00'
```
willkg: This is pretty much all we need to run any Django command with arguments on a schedule. Neato!

webapp-django/crashstats/cron/management/commands/cronrun.py
```diff
@@ -66,10 +66,21 @@ def handle(self, **options):
...
+    @contextlib.contextmanager
+    def stdout_to_logger(self, cmd):
+        class StdoutLogger:
+            def write(self, txt):
+                logger.info('%s: %s' % (cmd, txt.strip()))
+
+        stdout = self.stdout
+        self.stdout = StdoutLogger()
+        yield
+        self.stdout = stdout
```
willkg: We want logging output when running on the servers which outputs mozlog format which goes on a trip to the Azores or something like that. However, Django commands don't use logging output--they use `self.stdout`, so we wrap that in a context when running jobs and magic happens.

webapp-django/crashstats/cron/management/commands/cronrun.py
```diff
@@ -122,7 +133,7 @@ def _run_one(self, job_spec, force=False, cmd_args=None):
                     # pass in a run_time argument
 
                     start_time = time.time()
-                    if job_spec.get('backfill', True):
+                    if job_spec.get('backfill', False):
```
willkg: This was a bug. It made all non-backfill jobs backfill jobs. Oops.

willkg:
self-r+

PR #4877, opened 2019-04-04
bug 1540871: remove rabbitmq

willkg:
This removes all the rabbitmq code, configuration, docker containers, and docs. This also cleans up some things in the docs near where the rabbitmq stuff was.


scripts/update_siggen.py
```diff
@@ -57,6 +57,7 @@ def main(argv=None):
 
     # Get git sha of source
     source_sha = subprocess.check_output(['git', 'rev-parse', 'HEAD']).strip()
+    source_sha = source_sha.decode('utf-8')
```
willkg: Whoops--this is a fix for a different thing. It's not a big deal it's here, but it was an accident.

willkg:
Looks super. self-r+

PR #4878, opened 2019-04-04
bug 1541890: fix local dev setup issues

willkg:
This fixes a broken link in the signature generation docs and also fixes some
issues with local dev setup. One of the big issues is that service containers
weren't started when you went to do "make setup", so then it would fail.
Sadface.


willkg:
self-r+

PR #4880, opened 2019-04-04
bug 1513346: remove python 2 bits from signature generation

willkg:
This removes Python 2 bits from signature generation. We no longer need to
support Python 2 with siggen.

### DIFF
--- a/socorro/signature/tests/siglists/test-invalid-sig-list.txt
+++ b/socorro/signature/tests/siglists/test-invalid-sig-list.txt
@@ -1,4 +1,4 @@
 # Test Signatures List.
 # Contains an invalid regular expression.
 
-lol[[wut\]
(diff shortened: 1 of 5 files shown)


socorro/signature/tests/siglists/test-invalid-sig-list.txt
```diff
@@ -1,4 +1,4 @@
 # Test Signatures List.
 # Contains an invalid regular expression.
 
-lol[[wut\]
+lol[wut\]
```
willkg: This kicked up a warning in Python 3 when compiled as a regexp. Since it's supposed to be invalid, but not kick up a warning (which fails our pytest setup), I adjusted it.

willkg:
Seems fine. self-r+

PR #4885, opened 2019-04-09
bug 1543176: fix sentry-cli

willkg:
This adds sentry-cli to the webapp packages list and removes the crontabber
packages list. Now we've only got one set of js packages to deal with.


willkg:
self-r+

PR #4889, opened 2019-04-11
bug 1542392: rewrite monitoring crontabber app as Django command

willkg:
This rewrites the monitoring crontabber app as a Django command.  "monitoring"
wasn't very descriptive of what it did, so I renamed it to depcheck.

### DIFF
--- a/webapp-django/crashstats/settings/base.py
+++ b/webapp-django/crashstats/settings/base.py
@@ -721,6 +721,11 @@ CRONTABBER_MAX_ONGOING_AGE_HOURS = 2
 # Number of seconds before we retry a failed job
 CRONTABBER_ERROR_RETRY_TIME = 300
 
+# Configuration for the depcheck command
+NPM_PATH = config('NPM_PATH', '/usr/bin/npm')
+SAFETY_PATH = config('SAFETY_PATH', '/usr/local/bin/safety')
+SAFETY_API_KEY = config('SAFETY_API_KEY', '')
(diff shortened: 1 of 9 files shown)


webapp-django/crashstats/settings/base.py
```diff
@@ -721,6 +721,11 @@ def filter(self, record):
 # Number of seconds before we retry a failed job
 CRONTABBER_ERROR_RETRY_TIME = 300
 
+# Configuration for the depcheck command
+NPM_PATH = config('NPM_PATH', '/usr/bin/npm')
+SAFETY_PATH = config('SAFETY_PATH', '/usr/local/bin/safety')
+SAFETY_API_KEY = config('SAFETY_API_KEY', '')
+PACKAGE_JSON_PATH = config('PACKAGE_JSON_PATH', '/app/webapp-django/package.json')
```
willkg: Three of these have path values to binaries in the docker image, so I just set them. The API key will be set in the ops code. We'll need to set that when this lands.

willkg:
This requires a `SAFETY_API_KEY` configuration variable. I wrote up bug 1543683 (https://bugzilla.mozilla.org/show_bug.cgi?id=1543683) that covers figuring out what to do with configuration (and other things) as we transition from configman app to Django command. Going to make this block on that.

willkg:
I'm working on that other bug. I'll add the required configuration next. I'm going to land this now so I can work on the next crontabber job while I'm waiting for other things.

willkg:
self-r+

PR #4890, opened 2019-04-12
Update docs

willkg:
This cleans up some text, removes some out-dated bits, and shuffles a few
things around to better places.


willkg:
This is a good-enough pass. self-r+

PR #4891, opened 2019-04-15
Bug 1544449 - Fix NoteXPCOMChild class name.

amccreight:
GCGraphBuilder got renamed to CCGraphBuilder a long time ago, but the prefix list was never updated.


amccreight:
I checked just now, and it looks like I renamed it in 2014, in bug 1033654.

PR #4892, opened 2019-04-15
bug 1514294: Remove webpack from webapp docs

jwhitlock:
The ``webpack`` container and build process was removed, and is no longer needed to run the ``webapp``.


willkg:
@jwhitlock Can you rebase this against master?

PR #4894, opened 2019-04-16
bug 1544919: Update processor docs and pubsub publish command

jwhitlock:
* Drop the 2-year-old note about extracting the processor code, which means we might get to it this year.
* Allow ``socorro-cmd pubsub publish`` to read IDs from ``stdin``, like ``socorro-cmd fetch_crash_data``.
* Move ``SOCORRO_API_TOKEN`` docs to a new section, higher in the doc.

### DIFF
--- a/docs/service/processor.rst
+++ b/docs/service/processor.rst
@@ -44,13 +44,47 @@ All helper scripts run in the shell in the container::
 
     $ make shell
 
+Some of the scripts require downloading production data from
+`crash-stats.mozilla.com <https://crash-stats.mozilla.com>`_, and it is
+useful to add an API token with higher permissions before entering the shell.
+
+
+.. _`API token`:
+
+Adding an API Token
+-------------------
+By default, the download scripts will fetch anonymized crash data, which does
+not include personally identifiable information (PII). This anonymized data can
+be used to test some workflows, but the the processor will not be able to
+analyze memory dumps or generate signatures.
+
+If you have access to memory dumps, you can fetch those with the crash data by
+using an API token with these permissions:
+
+* View Personal Identifiable Information
+* View Raw Dumps
+
+You can generate API tokens at `<https://crash-stats.mozilla.com/api/tokens/>`_.
+
+.. Note::
+
+   Make sure you treat any data you pull from production in accordance with our
+   data policies that you agreed to when granted access to it.
+
+Add the API token value to your ``my.env`` file::
+
+    SOCORRO_API_TOKEN=apitokenhere
+
+The API token is used by the download scripts (run inside ``$ make shell``),
+but not directly by the processor.
 
 process_crashes.sh
 ------------------
 
 You can use the ``scripts/process_crashes.sh`` script which will fetch crash
 data, sync it with the S3 bucket, and publish the crash ids to Pub/Sub
-for processing.
+for processing. If you have access to memory dumps and use a valid
+`API token`_, then memory dumps will be fetched for processing as well.
 
 It takes one or more crash ids as arguments.
 
@@ -111,7 +145,9 @@ fetch_crash_data
 ----------------
 
 This will fetch raw crash data from crash-stats.mozilla.com and save it in the
-appropriate directory structure rooted at outputdir.
+appropriate directory structure rooted at outputdir. If you have access to
+memory dumps and use a valid `API token`_, then memory dumps will be fetched
+for processing as well.
 
 Usage from host:
 
@@ -140,28 +176,6 @@ You can get command help:
 .. code-block:: shell
 
    app@socorro:/app$ ./socorro-cmd fetch_crash_data --help
-
-
-.. Note::
-
-   This script requires that you have a valid API token from the
-   crash-stats.mozilla.com environment that has the "View Raw Dumps" permission
-   in order to download personally identifiable information and dumps.
-
-   You can generate API tokens at `<https://crash-stats.mozilla.com/api/tokens/>`_.
-
-   Add the API token value to your ``my.env`` file::
-
-       SOCORRO_API_TOKEN=apitokenhere
-
-   If you don't have an API token, this will download some raw crash
-   information, but it will be redacted.
-
-
-.. Note::
-
-   Make sure you treat any data you pull from production in accordance with our
-   data policies that you agreed to when granted access to it.
 
 
 scripts/socorro_aws_s3.sh
@@ -233,6 +247,7 @@ Let's process crashes for Firefox from yesterday. We'd do this:
 
 .. code-block:: shell
 
+  # Set SOCORRO_API_TOKEN in my.env
   # Start bash in the socorro container
   $ make shell
 
--- a/docs/service/processor.rst
+++ b/docs/service/processor.rst
@@ -4,12 +4,6 @@
 Service: Processor
 ==================
 
-.. Note::
-
-   We're in the process of extracting the processor out of Socorro as a separate
-   project. There's no ETA for that work, yet.
-
-
 Running the processor
 =====================
 
--- a/socorro/scripts/pubsub.py
+++ b/socorro/scripts/pubsub.py
@@ -134,18 +134,36 @@ def status(config, args):
                 print('             %s' % crash_id)
 
 
-def publish_crashid(config, args):
+def publish_crashid(config, queue, crashids):
     publisher = pubsub_v1.PublisherClient()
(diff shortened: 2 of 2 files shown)


socorro/scripts/pubsub.py
```diff
@@ -142,7 +142,10 @@ def publish_crashid(config, args):
     topic_name = config['resource.pubsub.%s_topic_name' % queue]
     topic_path = publisher.topic_path(project_id, topic_name)
 
-    for crash_id in args.crashid:
+    if not args.crashid:
+        print('Reading crash IDs from stdin')
```
willkg: You don't need to change this, but generally, I think we refer to it as "crash id" or "crashid" all lowercase.

socorro/scripts/pubsub.py
```diff
@@ -142,7 +142,10 @@ def publish_crashid(config, args):
...
+    crashids_iterable = args.crashid or sys.stdin
```
willkg: I think you got this from `fetch_crash_data.py`. There's a better way to do this that better handles the case where they didn't pass anything in--then it won't hang indefinitely. We do that in `reprocess.py`:

https://github.com/mozilla-services/socorro/blob/master/socorro/scripts/reprocess.py#L63-L71

I think `fetch_crash_data.py` should use that code. Also, maybe it makes sense to lift that to `socorro/scripts/__init__.py` so any of the scripts can use it.

We don't have to to that in this PR. I'm just thinking out loud.

docs/service/processor.rst
```diff
@@ -50,13 +44,42 @@ All helper scripts run in the shell in the container::
...
+The API token is used by the data download scripts (run inside
+``$ make shell``), but not directly by the processor.
 
 process_crashes.sh
 ------------------
 
 You can use the ``scripts/process_crashes.sh`` script which will fetch crash
 data, sync it with the S3 bucket, and publish the crash ids to Pub/Sub
-for processing.
+for processing. This command requires a valid `API token`_.
```
willkg: I tried to do these commands so that they don't *require* a valid API token. You can still do things without one. The API token only helps if you have additional privileges--they let you pull down PII parts of the crash data. The old docs are muddled on this point.

I think I'd rephrase this as "If you have an account that has access to memory dumps, be sure to use an API token so you can download that data." Or something like that.

docs/service/processor.rst
```diff
@@ -50,13 +44,42 @@ All helper scripts run in the shell in the container::
...
+`crash-stats.mozilla.com <https://crash-stats.mozilla.com>`_, and it is
+useful to add an API token before entering the shell.
+
+
+.. _`API token`:
+
+Adding an API Token
+-------------------
+The data download scripts require an API token with the "View Raw Dumps"
+permission to acquire raw crash data. This crash data will not include
```
willkg: We discovered it requires two different permissions:

* View Raw Dumps
* View Personal Identifiable Information

If you're missing one or the other, then it won't download all the data. So this should be fixed.

socorro/scripts/pubsub.py
jwhitlock: Thanks, I was looking around for a way to do that - it was weird to have ``socorro_cmd pubsub publish`` just hang waiting for input, so I added a ``print`` message. I think this is a better solution, and I'll switch.

docs/service/processor.rst
jwhitlock: I thought ``View Raw Dumps`` was required, but it appears raw crash data is downloaded with ``SOCORRO_API_TOKEN=`` (blank token). I'm going to try to download some data so I'm clear on what the permissions do.

docs/service/processor.rst
jwhitlock: Do you mean the docs should mention both, or that the API should change, or something else?

docs/service/processor.rst
jwhitlock: OK it looks like you either need no API key, or an API key with both permissions. I'll update the docs to remove the middle tier of permissions.

**Blank token**
* Prints "No api token provided. Skipping dumps and personally identifiable information."
* Prints "Working on [uuid]..."
* ``[crashdata]/v2/raw_crash/[prefix]/[date]/[uuid]`` gets sanitized JSON data
* ``[crashdata]/v1/dump_names/[uuid]`` gets an empty list ``[]``

**Token with "View Raw Dumps"**
* Prints "Using api token: abcdxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
* Prints "Working on [uuid]..."
* ``[crashdata]/v2/raw_crash/[prefix]/[date]/[uuid]`` gets sanitized JSON data
* ``[crashdata]/v1/dump_names/[uuid]`` gets an empty list ``[]``

**Token with "View Personal Identifiable Information"**
* Prints "Using api token: abcdxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
* Prints "Working on [uuid]..."
* ``[crashdata]/v2/raw_crash/[prefix]/[date]/[uuid]`` gets complete JSON data
* Exception raised ``Exception: Something unexpected happened. status_code 403, content b"Binary response requires the 'View Raw Dumps' permission\n"``

**Token with both permissions**
* Prints "Using api token: abcdxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
* Prints "Working on [uuid]..."
* ``[crashdata]/v2/raw_crash/[prefix]/[date]/[uuid]`` gets complete JSON data
* ``[crashdata]/v1/dump_names/[uuid]`` gets a 1-item list ``["[the name]"]``
* ``[crashdata]/v1/dump/[dump uuid]`` gets a binary dump

jwhitlock:
I've force-pushed to:

* Add new [bug 1544919](https://bugzilla.mozilla.org/show_bug.cgi?id=1544919) to the commit messages
* Add ``list_from_arg_or_pipe``, to copy the method from ``reprocess`` that avoids blocking on stdin. I didn't promote it to ``__init__.py`` yet, because I'm not sure if this is the right scope for a utility function, and if you have an idea for a better name.
* Add the required permissions as a list, and reword a bit. It's now clearer that your options are no PII / no memory dumps, or PII and memory dumps.
* Update the back link text to "If you have access to memory dumps and use a valid
`API token`_, then memory dumps will be fetched for processing as well."

docs/service/processor.rst
willkg: This is goofy. The two permissions seem inextricably entwined. Seems like we should combine them or dump one or something.

Having said that, this is the only place where the awkwardness shows up that I can think of. Fiddling with permissions and such is non-trivial, so as a stop-gap, maybe we add a note to the top of the tokens page and push any other changes off until a rainy day.

https://crash-stats.mozilla.com/api/tokens/

Maybe something like:

"If you're creating an API token for downloading crash data, you need 'View Raw Dumps' and 'View Personal Identifiable Information'."

docs/service/processor.rst
willkg: I think the short-term least-effort solution is that the docs should mention both permissions. Maybe longer term after someone has really looked into it, we should fix the API/permissions structure.

docs/service/processor.rst
jwhitlock: People don't read the docs first?!? 😲

I think I'll tackle non-doc stuff in a new PR.

socorro/scripts/pubsub.py
```diff
@@ -134,18 +134,36 @@ def status(config, args):
...
     topic_name = config['resource.pubsub.%s_topic_name' % queue]
     topic_path = publisher.topic_path(project_id, topic_name)
 
-    for crash_id in args.crashid:
+    for crash_id in crashids:
         future = publisher.publish(topic_path, data=crash_id.encode('utf-8'))
         future.result()
         print('Published: %s' % crash_id)
+    else:
+        print('No crash ids provided.')
```
willkg: Oops--I should have caught this in review. This clause executes if `for` gets through the iterable without breaking out. That happens whether or not there were crashids to publish.

I think what we want is something like this:

```
if crashids:
    for crash_id in crashids:
        blahblahblah
else:
    print('No crash ids provided.')
```

Or alternatively:

```
if not crashids:
    print('No crash ids provided. I am sad.')
    return

for crash_id in crashids:
    blahblahblah()
```

jwhitlock: oh no I've learned ``else`` wrong. 🙈 

I think I can drop the clause entirely, since the detection for an empty list is done by the caller.

PR #4896, opened 2019-04-17
bug 1544919: Refresh webapp service docs

jwhitlock:
Update the docs to reflect my best guess at current workflows:

* Add note for signing up with ``oidcprovider`` in dev. I have further changes I want to make, but I first filed as issues in [mozilla-parsys](https://github.com/mozilla-parsys/docker-test-mozilla-django-oidc).
* Move static assets note into new section, with example ``docker-compose.override.yml`` and macOS note. Move instructions for production-like assets as a subsection.
* Simplify and organize the permission sections, which felt like an introduction to django.contrib.auth. Add a link to the official docs to cover content I removed.
* Permissions and default groups are now in ``signal.py``. I think they should be in a data migration, but that can be a future discussion / PR, depending on how often permissions are changed.
* Drop the "Troubleshooting" section, which referred to removed views and emailing tracebacks (it is easier to view the webapp logs for tracebacks).

### DIFF
--- a/docs/service/webapp.rst
+++ b/docs/service/webapp.rst
@@ -23,21 +23,6 @@ To ease debugging, you can run a shell in the container::
 Then you can start and stop the webapp, adjust files, and debug.
 
 
-.. Note::
-
-   The ``STATIC_ROOT`` is set to ``/tmp/crashstats-static/`` rather than
-   ``/app/webapp-django/static``. This alleviates permissions-related problems
-   because the process in the container runs as uid 10001 which is not the uid
-   of the user you're using on your host computer.
-
-   The problem this creates is that ``/tmp/crashstats-static/`` is ephemeral
-   and any changes there disappear when you stop the container.
-
-   If you want it persisted, you should mount that directory using ``volumes``
-   in a ``docker-compose.override.yml`` file.
-
-   https://docs.docker.com/compose/extends/
-
 
 Setting up authentication and a superuser
 =========================================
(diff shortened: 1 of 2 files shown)


docs/service/webapp.rst
```diff
@@ -23,21 +23,6 @@ To ease debugging, you can run a shell in the container::
...
-   of the user you're using on your host computer.
-
-   The problem this creates is that ``/tmp/crashstats-static/`` is ephemeral
-   and any changes there disappear when you stop the container.
-
-   If you want it persisted, you should mount that directory using ``volumes``
-   in a ``docker-compose.override.yml`` file.
-
-   https://docs.docker.com/compose/extends/
-
```
willkg: This is still true. It's something I keep forgetting to look into, but it's still the case. The `STATIC_ROOT` for the local dev environment is set in `docker/config/never_on_a_server.env`.

I can explain what's going on here, but it's sort of complicated and requires copious hand-waving.

willkg: Oh, my bad--you just moved it.

docs/service/webapp.rst
```diff
@@ -153,46 +132,45 @@ Here's an example in a template::
...
-  )
-  MANAGERS = ADMINS
+.. code-block:: yaml
 
+    version: "2"
+    services:
+      webapp:
+        volumes:
+          # Persist the static files folder
+          - ./static:/tmp/crashstats-static
```
willkg: For the record, this is my fault. It's terrible. I have some thoughts on how to fix it and make it less terrible, but I haven't done it, yet.

jwhitlock: I feel this is a limitation of Docker environments, and this isn't a bad solution, especially since you use long-running containers in your personal workflow. I'm happy to document now and improve later.

I've also struggled with this on Kuma. Our solution was [to set ``UID``](https://github.com/mozilla/kuma/blob/8ff9b279ce4b0c4f446b5c8591afdea16f43af66/.env-dist.dev#L3-L8), so the in-container user matched the host user. I'm not thrilled with that approach either, but it does make Linux work more like macOS.  For production, we have a derived image that explicitly gathers assets.

I've used the in-container directory trick as well, for installing npm packages and avoiding s system mismatch on ``node_modules``.

docs/service/webapp.rst
```diff
@@ -67,78 +52,72 @@ need to add the following entry to your ``/etc/hosts`` (or equivalent) file::
...
-   ./manage.py migrate auth
-   ./manage.py migrate
+Administering Users and Groups
+------------------------------
+The Django admin can be used to assign a user to a group, in order to grant
+them additional access in the web app.
 
+The admin can also be used to create new groups with associated permissions.
+However, since groups and permissions are stored in the database, additional
+work is needed to apply these changes to all environments.
```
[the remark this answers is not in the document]
jwhitlock: That sounds like a plan. I'd add "Hackers Plus" to the docs as well, unless you have a reason to omit it. The link to [user-facing docs](https://crash-stats.mozilla.com/documentation/memory_dump_access/) was a missing element for me, and completes the picture.

docs/service/webapp.rst
willkg: We allow people to set the uid/gid to use in `my.env`. That covers the permissions issues that differ between Linux and OSX. That's what step 3 here is about:

https://socorro.readthedocs.io/en/latest/localdevenvironment.html#setup-quickstart

If I recall correctly, the problem here is that the static assets get built in the image, but then in `docker-compose`, we stomp on `/app` with the current directory, so then the static assets that got built during `make build` disappear. So then you have two groups of static assets and it's hard to know which one you're adjusting. Or something along those lines.

jwhitlock:
I linked to the Quickstart section, to remind Linux users that they can adjust UID / GID. I reduced the permissions docs down to one section, and linked to the Django docs, the [in-product docs](https://crash-stats.mozilla.com/documentation/memory_dump_access/), and the code for details.

PR #4897, opened 2019-04-18
bug 1544497: fix comments tab in signature report

willkg:
Users shouldn't be able to view comments if they don't have `view_pii`
permissions.

### DIFF
--- a/webapp-django/crashstats/signature/jinja2/signature/signature_report.html
+++ b/webapp-django/crashstats/signature/jinja2/signature/signature_report.html
@@ -65,7 +65,9 @@
         <li><a href="#reports" class="reports" data-tab-name="reports">Reports</a></li>
         <li><a href="#graphs" class="graphs" data-tab-name="graphs">Graphs</a></li>
         <li><a href="#bugzilla" class="bugzilla" data-tab-name="bugzilla">Bugzilla</a></li>
-        <li><a href="#comments" class="comments" data-tab-name="comments">Comments</a></li>
+        {% if request.user.has_perm('crashstats.view_pii') %}
+          <li><a href="#comments" class="comments" data-tab-name="comments">Comments</a></li>
+        {% endif %}
         {% if product in correlations_products %}
           <li><a href="#correlations" class="correlations" data-tab-name="correlations">Correlations</a></li>
         {% endif %}
--- a/webapp-django/crashstats/signature/jinja2/signature/signature_report.html
+++ b/webapp-django/crashstats/signature/jinja2/signature/signature_report.html
@@ -65,9 +65,7 @@
         <li><a href="#reports" class="reports" data-tab-name="reports">Reports</a></li>
         <li><a href="#graphs" class="graphs" data-tab-name="graphs">Graphs</a></li>
         <li><a href="#bugzilla" class="bugzilla" data-tab-name="bugzilla">Bugzilla</a></li>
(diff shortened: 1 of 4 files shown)


webapp-django/crashstats/signature/jinja2/signature/signature_report.html
```diff
@@ -65,7 +65,9 @@ <h2>Signature report for <em>{{ signature }}</em></h2>
         <li><a href="#reports" class="reports" data-tab-name="reports">Reports</a></li>
         <li><a href="#graphs" class="graphs" data-tab-name="graphs">Graphs</a></li>
         <li><a href="#bugzilla" class="bugzilla" data-tab-name="bugzilla">Bugzilla</a></li>
-        <li><a href="#comments" class="comments" data-tab-name="comments">Comments</a></li>
+        {% if request.user.has_perm('crashstats.view_pii') %}
+          <li><a href="#comments" class="comments" data-tab-name="comments">Comments</a></li>
+        {% endif %}
```
willkg: This is a ui-nicety. Why show a tab the user can't see the data of.

I'd rather we did this in other ways--like maybe the signature report view should determine which tabs can be seen and then we iterate over that? I don't know.

webapp-django/crashstats/signature/tests/test_views.py
```diff
@@ -390,6 +390,13 @@ def mocked_supersearch_get(**params):
         assert 'term_counts' in struct
         assert len(struct['term_counts']) == 1
 
+    def test_signature_comments_no_permission(self):
+        """Verify comments are not viewable without view_pii."""
+        url = reverse('signature:signature_comments')
+
+        response = self.client.get(url, {'signature': 'whatever'})
+        assert response.status_code == 403
```
willkg: This verifies you can't view the comment data if you're not logged in. That's equivalent to logged-in, but no `view_pii` permission, so it covers both cases.

webapp-django/crashstats/signature/views.py
```diff
@@ -318,7 +318,10 @@ def signature_graphs(request, params, field):
 
 @pass_validated_params
 def signature_comments(request, params):
-    '''Return a list of non-empty comments. '''
+    """Return a list of non-empty comments."""
+    # Users can't see comments unless they have view_pii permissions.
+    if not request.user.has_perm('crashstats.view_pii'):
+        return http.HttpResponseForbidden()
```
willkg: This check makes sure no one can see the data--it throws a 403. So even if the tab was showing, it wouldn't work and people can't figure out the url and hit it directly. Etc.

webapp-django/crashstats/signature/jinja2/signature/signature_report.html
jwhitlock: I can imagine the version that iterates, but I'm not sure if that would be some Jinja code or a structure setup in the view. It would also make it harder to read. When the templates get too far away from the "View Source" version, I get uncomfortable.

The reason to show the tab would be to allow access to the page, but show the user a message on the Comments page saying "Comments may include personal information, and you aren't allowed to see them, see this page for details".

willkg:
Ooo--I like that idea better. I tweaked the fix so that the comments tab continues to show, but the handler handles 403 and shows a more helpful "you don't have access" message.

willkg:
@jwhitlock Thank you!

PR #4900, opened 2019-04-19
bug 1544919: More doc tweaks - README paths, process_app command

jwhitlock:
* For imported ``README``s, explicitly mention the directory path
* Fix the call to ``processor_app.py``, update the extracted output

There are issues with other documents, but I'll need more experience to suggest changes and fixes, and they have clear notes implying the document is out of date.


jwhitlock:
I suspect a lot of the elasticsearch stuff will change, and a lot will be deleted, so I read the prose and tried to verify the one command I could run - which didn't run. I'll revisit these docs in the context of modifying SuperSearch in the context of an ElasticSearch update.

PR #4901, opened 2019-04-19
bug 1544919: fix scripts/process_crashes.sh to take from stdin

willkg:
This fixes `./scripts/process_crashes.sh` to pull from stdin if there are no crashid args and stdin is not a tty.

```shell
$ make shell
app@afdfe895c928:/app$ ./scripts/process_crashes.sh 
Usage: process_crashes.sh CRASHID [CRASHID ...]
app@afdfe895c928:/app$ echo 4854438d-dd83-44d5-967b-155d70190418 | scripts/process_crashes.sh 
<copious output>
app@afdfe895c928:/app$ scripts/process_crashes.sh 4854438d-dd83-44d5-967b-155d70190418
<copious output>
```

### DIFF
--- a/docs/service/processor.rst
+++ b/docs/service/processor.rst
@@ -92,13 +92,13 @@ For example:
 
 .. code-block:: shell
 
-   app@socorro:/app$ ./scripts/process_crashes.sh ed35821d-3af5-4fe9-bfa3-dc4dc0181128
+   app@socorro:/app$ scripts/process_crashes.sh ed35821d-3af5-4fe9-bfa3-dc4dc0181128
 
 You can also use it with ``fetch_crashids``:
 
 .. code-block:: shell
 
-   app@socorro:/app$ ./socorro-cmd fetch_crashids --num=1 | ./scripts/process_crashes.sh
+   app@socorro:/app$ socorro-cmd fetch_crashids --num=1 | scripts/process_crashes.sh
 
 After running ``scripts/process_crashes.sh``, you will need to run the
 processor which will do the actual processing.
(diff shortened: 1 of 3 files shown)


docs/service/processor.rst
```diff
@@ -98,7 +98,7 @@ You can also use it with ``fetch_crashids``:
 
 .. code-block:: shell
 
-   app@socorro:/app$ socorro-cmd fetch_crashids --num=1 | xargs scripts/process_crashes.sh
+   app@socorro:/app$ ./socorro-cmd fetch_crashids --num=1 | ./scripts/process_crashes.sh
```
willkg: I'm pretty sure `.` is not on the `PATH`, so we always have to do `./socorro-cmd`. If that's not true, that's cool and we can remove them all from this file. But it felt weird to have half this way and half that way.

scripts/process_crashes.sh
```diff
@@ -30,8 +30,14 @@ function cleanup {
 trap cleanup EXIT
 
 if [[ $# -eq 0 ]]; then
-    echo "Usage: process_crashes.sh CRASHID [CRASHID ...]"
-    exit 1
+    if [ -t 0 ]; then
```
willkg: I never knew about `-t`. What it does is test to see if the following fd is a terminal. fd `0` is stdin. So then this is asking "is stdin a terminal?" If it is, then the user executed it. If it's not, then it's on the other side of a pipe from some command.

scripts/process_crashes.sh
```diff
@@ -30,8 +30,14 @@ function cleanup {
...
+        # If stdin is a terminal, then there's no input
+        echo "Usage: process_crashes.sh CRASHID [CRASHID ...]"
+        exit 1
+    fi
+
+    # stdin is not a terminal, so pull the args from there
+    set -- ${@:-$(</dev/stdin)}
```
willkg: So what this does is set the positional args to `$@` (the positional args) if there are any and default to data from stdin if not `$(</dev/stdin)`.

docs/service/processor.rst
jwhitlock: It's in my path:

```
$ make shell
app@19f44b639b99:/app$ echo $PATH
/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:.
```

That appears to be set in ``/etc/bash.bashrc``, but that also seems to set the prompt, which doesn't work for me 🤷‍♂️

willkg: It's probably overridden by `/home/app/.bashrc`. So, building the container adds the `/etc/bash.basrhc`, but then a few months later, I needed to do something that required a home directory so I changed adduser to build a home directory and while I was puzzled why "socorro" didn't appear in the prompt, I never looked into it until now. That's exciting!

willkg: We could probably have `docker/set_up_ubuntu.sh` delete `/home/app/.bashrc`. Then we get the right prompt and path for all users.

willkg: Oh, I know.... the problem I have is doing this:

```
$ docker-compose run app shell socorro-cmd
```

That won't find socorro-cmd. But doing this will:

```
$ docker-compose run app shell ./socorro-cmd
```

That's really irritating.

willkg:
^^^ That refixes the `./socorro-cmd` -> `socorro-cmd` and also adds a bit to `docker/set_up_ubuntu.sh` to remove `/home/app/.bashrc` which naughtily stomps on our nice prompt.

PR #4902, opened 2019-04-19
bug 1544919: Add FallbackToPipeAction

jwhitlock:
``FallbackToPipeAction`` centralizes the logic for taking ``crashids`` from either the positional arguments or a piped list, so both work:

```shell
cat crashids.txt | xargs socorro-cmd fetch_crash_data ./crashdata
cat crashids.txt | socorro-cmd fetch_crash_data ./crashdata
```

It works in the ``argparse`` framework so that omitting both is a parsing error, which displays the default usage message.

### DIFF
--- a/socorro/scripts/__init__.py
+++ b/socorro/scripts/__init__.py
@@ -99,6 +99,9 @@ class FallbackToPipeAction(argparse.Action):
         if option_strings:
             raise ValueError("This action does not work with named arguments")
         if nargs != '*':
+            # For nargs='*', the action is called with an empty list.
+            # For other values ('?', '+', 1), the action isn't called, so we
+            # can't fallback to reading from stdin.
             raise ValueError("nargs should be '*'")
         super().__init__(option_strings, dest, nargs=nargs, **kwargs)
 
--- a/socorro/scripts/__init__.py
+++ b/socorro/scripts/__init__.py
@@ -4,6 +4,7 @@
 
 import argparse
 import os
+import sys
 
 
 class WrappedTextHelpFormatter(argparse.HelpFormatter):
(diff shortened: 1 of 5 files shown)


jwhitlock:
I'm not 100% happy with this, but it is close enough for a Friday PR. I think the docstrings are overkill, and I expected to write some unit tests for ``FallbackToPipeAction``, but no other commands are tested. It can wait for Monday or later for review.

socorro/scripts/__init__.py
```diff
@@ -75,6 +76,66 @@ def __call__(self, parser, namespace, values, option_string=None):
         setattr(namespace, self.dest, value)
 
 
+class FallbackToPipeAction(argparse.Action):
```
willkg: There's a test for `FlagAction` here:

https://github.com/mozilla-services/socorro/blob/master/socorro/unittest/scripts/test_init.py#L12

It's probably worth adding a rough test for `FallbackToPipeAction`.

socorro/scripts/__init__.py
```diff
@@ -75,6 +76,66 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+
+        :param option_strings: Names for non-positional arguments.
+        :param dest: The destination attribute on the parser
+        :param nargs: The number of arguments, must be '*' for zero or more
+        :param kwargs: Additional parameters for the parser argument
+        """
+        if option_strings:
+            raise ValueError("This action does not work with named arguments")
+        if nargs != '*':
+            raise ValueError("nargs should be '*'")
```
willkg: Out of curiosity, why restrict it to 0 or more?

jwhitlock: Great question, and there should be a comment that explains. 

With ``*``, the action is called with an empty list. With ``?`` and some others I tested, it isn't called at all.

willkg: Oh, how weird. No need for a comment--I was just curious.

jwhitlock:
I've added tests for ``FallbackToPipeAction``. I'm usually aggressive with local pytest fixtures to remove duplication, but I followed the existing test style. Let me know if you want to see it with something like these:

```python
@pytest.fixture
def parser():
    return argparse.ArgumentParser()

@pytest.fixture
def fallback_parser(parser):
   parser.add_argument('testid', nargs='*', action=FallbackToPipeAction)
   return parser

@pytest.fixture
def mock_stdin():
    with mock.patch('sys.stdin', spec_set=['isatty', '__iter__']) as mock_stdin:
        yield mock_stdin
```

socorro/scripts/pubsub.py
```diff
@@ -134,36 +134,19 @@ def status(config, args):
                 print('             %s' % crash_id)
 
 
-def publish_crashid(config, queue, crashids):
+def publish_crashid(config, args):
```
willkg: I think I would have kept the arguments here as `queue` and `crashids` since it simplifies the setup for running this function--it's just working on a queue (str) and a list of crashids (list of crashids) rather than an argparse something or other.

I don't think we have to change this back. I don't think we test this function by itself.

jwhitlock: I prefer ``queue`` and ``crashids`` as well. It was ``args`` before I touched it, and I was matching the style of the other calls. I'm changing it back.

willkg: Oh noes! I'm sorry for flip-flopping on you.

PR #4903, opened 2019-04-23
bug 1546254: add django cmd support to socorro-cmd

willkg:


socorro-cmd
```diff
@@ -65,58 +84,47 @@ class Group:
 COMMANDS = [
     Group(
         'Crash processing utilities', {
-            'add_crashid_to_queue': 'socorro.scripts.add_crashid_to_queue.main',
```
willkg: Whoops--should have removed this a while back. That script doesn't exist anymore.

socorro-cmd
```diff
@@ -65,58 +84,47 @@ class Group:
...
             'clear_indices': 'socorro.external.es.clear_indices_app.ClearESIndicesApp',
-            'upload_telemetry_schema': (
+            'upload_telemetry_schema': import_path(
                 'socorro.external.boto.upload_telemetry_schema.UploadTelemetrySchema'
             ),
-            'db': 'socorro.scripts.db.main',
-            'pubsub': 'socorro.scripts.pubsub.main',
+            'db': import_path('socorro.scripts.db.main'),
+            'pubsub': import_path('socorro.scripts.pubsub.main'),
+            'depcheck': django_cmd('depcheck'),
```
willkg: I added `depcheck` figuring I run it periodically. Today's me doesn't think we should add all the Django commands that we use, but we should add the ones that we use for general local dev environment management.

For example, I'm reworking all the "clean elasticsearch indices" which includes a couple of apps, a crontabber job, a script, a house in kentucky, an obsolete lotto ticket, and a horse with no name. Seems kind of silly to have all that, but we need the script which wipes Elasticsearch state for the local dev environment. I think that should be in `socorro-cmd`.

Anyhow, I added depcheck because I needed one to test with and I do periodically run this since it covers the whole codebase.

socorro-cmd
```diff
@@ -49,6 +50,24 @@ def showcommands_cmd(argv):
...
+        manage.main(argv)
+    return _django_cmd
+
+
+def import_path(python_path):
+    """Import a name from a module."""
+    module_path, name = python_path.rsplit('.', 1)
+    __import__(module_path)
+    module = sys.modules[module_path]
+    return getattr(module, name)
```
willkg: This code used to be in `import_and_run`. One nice thing about having it here is that it tries to import the destination whenever this command is run--not just when this specific subcommand is run. That's how I noticed that one of the scripts was dead. I think I like that better since `socorro-cmd` is manually tested.

webapp-django/manage.py
```diff
@@ -8,11 +8,15 @@
...
     os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'crashstats.settings')
 
     from django.core.management import execute_from_command_line
 
-    execute_from_command_line(sys.argv)
+    execute_from_command_line(argv)
+
+
+if __name__ == '__main__':
+    main(sys.argv)
```
willkg: This change just adjusts this file so it's possible to run `main` with different arg sets.

willkg:
To test this, pull the PR branch down, then do this:

```shell
$ make shell
app@socorro:/app$ socorro-cmd
<copious output>
app@socorro:/app$ socorro-cmd depcheck --help
<depcheck output>
```

socorro-cmd
willkg: Oh, hot diggity! that's a new-in-Python-3 thing I didn't know about. I'll switch it.

willkg:
If you look at the paths for the subcommands in socorro-cmd, they're all over the code base. We're at a point where we have two cli front-ends for things and it's not obvious which local dev environment manipulation things are where. I feel like having some of the commands that you'd use when working on processor things in socorro-cmd where they're discoverable will help.

You can continue to use `manage.py` for Django-specific things if you like--it works the same. But I really like being able to say (especially in documentation) "the things you need to manipulate your local dev environment are in "socorro-cmd --help" regardless of where the implementation is".

PR #4905, opened 2019-04-23
bug 1545784: deal with "504 Deadline Exceeded"

willkg:
This adds some handling for "504 Deadline Exceeded" in the pubsub queue code and also fixes the processor app to send unhandled exceptions from queue code to Sentry.

### DIFF
--- a/socorro/external/pubsub/crashqueue.py
+++ b/socorro/external/pubsub/crashqueue.py
@@ -8,6 +8,7 @@ import os
 
 from configman import Namespace, RequiredConfig
 from google.cloud import pubsub_v1
+from google.api_core.exceptions import DeadlineExceeded
 
 
 # Maximum number of messages to pull from a Pub/Sub topic in a single pull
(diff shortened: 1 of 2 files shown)


socorro/external/pubsub/crashqueue.py
```diff
@@ -128,11 +129,17 @@ def __iter__(self):
...
+                try:
+                    response = self.subscriber.pull(
+                        sub_path,
+                        max_messages=PUBSUB_MAX_MESSAGES,
+                        return_immediately=True
+                    )
+                except DeadlineExceeded:
+                    # If we exceeded the timeout, treat it as if there was
+                    # nothing to receive and move on.
+                    continue
```
willkg: I think this is the best thing to do. It might be the case that we hit this a lot and we wouldn't know because we're silently ignoring it, but given that we've only hit a couple of these in the month the code has been live, I don't think that will be the case.

socorro/processor/processor_app.py
```diff
@@ -156,7 +156,7 @@ class ProcessorApp(FetchTransformSaveApp):
         reference_value_from='secrets.sentry',
     )
 
-    def _capture_error(self, crash_id, exc_info):
+    def _capture_error(self, exc_info, crash_id=None):
```
willkg: I switched the order of these so I could make `crash_id` optional. When processing crashes, it makes sense to have the `crash_id` as context--that makes errors in processing soooo much more helpful. But for errors in the queue code, it's not a thing.

willkg:
There's some urgency to having this landed because we'd like to have more signal about whether queuing is failing. I'm going to self-r+ this. If there are issues, we can roll forward.

PR #4906, opened 2019-04-24
bug 1540704: Update to Django 2.2

jwhitlock:
Update Django and related dependencies:

* Django 2.1.7 → [2.2](https://docs.djangoproject.com/en/2.2/releases/2.2/#id2): LTS version. There are some new database features, but they do not require changes in the webapp.
* pytz 2018.9 → 2019.1: Update to latest release
* sqlparse 0.3.0: New dependency for Django 2.2

There is a new ``PendingDeprecationWarning`` for ``CachedStaticFilesStorage``, scheduled for removal in Django 3.1. Rather than change the staticfiles storage, this adds a pytest exception for the warning.


willkg:
Relatedly, we have this wonky thing I wrote a while back that allows people to see LogEntry instances, but not change them:

https://github.com/mozilla-services/socorro/blob/master/webapp-django/crashstats/crashstats/admin.py#L61

It's super wonky. It'd be nice to fix that and if the comment is right, we should be able to now.

PR #4907, opened 2019-04-24
bug 1546433: refactor Elasticsearch index delete code

willkg:
This redoes all the code that has anything to do with deleting Elasticsearch indices. Now all create/delete code is part of `ConnectionContext`.
    
This adds `delete` to the `es` script so we can use the `es` script exclusively in the local dev environment.
    
This fixes test configuration so that we can test against the Elasticsearch service without messing up the local dev environment.
    
This adds an `es_conn` pytest fixture for testing with Elasticsearch.


docker/config/test.env
```diff
@@ -21,7 +21,8 @@ resource.boto.bucket_name=crashstats-test
 TELEMETRY_BUCKET_NAME=telemetry-test
 
 # Elasticsearch configuration
-ELASTICSEARCH_URLS=http://example.com:9123
+resource.elasticsearch.elasticsearch_index=testsocorro%Y%W
+resource.elasticsearch.elasticsearch_index_regex=^testsocorro[0-9]{6}$
```
willkg: This prevents the tests from stomping on data in the local dev environment which is kind of annoying.

docker/config/test.env
```diff
@@ -21,7 +21,8 @@ resource.boto.bucket_name=crashstats-test
 TELEMETRY_BUCKET_NAME=telemetry-test
 
 # Elasticsearch configuration
-ELASTICSEARCH_URLS=http://example.com:9123
```
willkg: This allows us to test against the actual Elasticsearch service in the local dev environment rather than mocking everything all the time. Switching to testing against a real Elasticsearch will save our :bacon: when we go to upgrade.

socorro/external/es/connection_context.py
```diff
@@ -56,9 +59,22 @@ class ConnectionContext(RequiredConfig):
...
+    required_config.add_option(
+        'elasticsearch_index_regex',
+        doc='regex that matches indexes--this should match elasticsearch_index',
+        default='^socorro[0-9]{6}$',
+        reference_value_from='resource.elasticsearch',
+    )
+    required_config.add_option(
+        'retention_policy',
+        default=26,
+        doc='number of weeks to retain an index',
```
willkg: These two options were added from the index cleaner.

socorro/scripts/es.py
```diff
@@ -23,7 +24,13 @@
...
 
 def get_conn():
-    cm = ConfigurationManager(ConnectionContext.get_required_config())
+    # Create a configuration manager that will only check the environment for
+    # configuration and not command line parameters
+
+    cm = ConfigurationManager(
+        ConnectionContext.get_required_config(),
+        values_source_list=[environment]
+    )
```
willkg: This fixes the `es` script so that configman isn't trying to look at `sys.argv` and then getting really sad that it doesn't like what it finds. This fixes it to *only* look at the environment.

willkg:
To test things:

```shell
$ make shell

# Create recent indices and delete them
app@socorro:/app$ socorro-cmd es create
<create output>
app@socorro:/app$ socorro-cmd es delete
<delete output>

# Create a bunch of old indices--far more than our retention policy allows
app@socorro:/app$ socorro-cmd es create --past=40

# Reset and run the crontabber job
app@socorro:/app$ socorro-cmd crontabber --reset-job=elasticsearch-cleanup
app@socorro:/app$ socorro-cmd crontabber --job=elasticsearch-cleanup
<output of it deleting old indices>

# Make sure we still have indices that haven't expired
app@socorro:/app$ socorro-cmd es list
```

willkg:
I think this feels good and the commands and tests all work. self-r+

PR #4908, opened 2019-04-24
bug 1542391: rewrite elasticsearch-cleanup crontabber job

willkg:
This rewrites the elasticsearch-cleanup crontabber job as a Django command and adds it to the cron job spec list.


willkg:
I went through and tested the `esclean` command using `socorro-cmd es` to set up indices. Seems to work super. self-r+

PR #4911, opened 2019-04-25
bug 1546978: Update permissions, refresh LogEntry admin view

jwhitlock:
The ``LogEntry`` view allows seeing admin actions (either manual or from tools) as a list. It should be read-only, for auditing purposes, but until Django 2.1, there wasn't a supported way to mark a view as read-only.

Use Django 2.1's default ``has_view_permission`` to drop the ``has_change_permission`` hack, and drop the unneeded ``has_module_permission`` method.

Borrow some ideas from [yprez/django-logentry-admin](https://github.com/yprez/django-logentry-admin/blob/master/logentry_admin/admin.py):

* Change the ``user_email`` column name to "Admin"
* Change the ``get_edited_object`` column to "Object", use the email address when the object is a ``User``, and link to admin details page if available
* Change the ``action_name`` column to "Action"
* Change the ``get_change_message`` column to "Change Message", and link to details

Here's the view on my local box:

Screenshot 2019-04-25 10 18 04

### DIFF
--- a/webapp-django/crashstats/crashstats/admin.py
+++ b/webapp-django/crashstats/crashstats/admin.py
@@ -59,19 +59,10 @@ class LogEntryAdmin(admin.ModelAdmin):
         return False
 
     def has_change_permission(self, request, obj=None):
-        # FIXME(willkg): If this always returned False, then this modeladmin
-        # doesn't show up in the index. However, this means you get a change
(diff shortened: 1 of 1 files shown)


jwhitlock:
Force-push to re-implement the "admin" column change, and change ``object_link`` to use ``format_html`` instead of ``escape`` and ``mark_safe``.

PR #4912, opened 2019-04-25
bug 1543097: Rename raven_client to sentry_client

jwhitlock:
In preparation for switching from raven to [sentry_sdk](https://docs.sentry.io/platforms/python/), rename ``raven_client.py`` to ``sentry_client.py``. Also, change code to mock ``get_client``, instead of ``raven.Client``. This name change will allow future PRs to focus on code changes and avoid wading through name switches. However, if you want the full picture, we can hold off reviewing or merging until the other changes are ready.


docker/config/test.env
```diff
@@ -8,7 +8,6 @@ STATSD_CLIENT=django_statsd.clients.null
 
 # Sentry
 SENTRY_DSN=
-RAVEN_DSN=
```
willkg: Oh, that's funny. Ha!

willkg: Oh, i think I see what's going on. We needed `RAVEN_DSN` for the webapp and `SENTRY_DSN` for the sentry-cli which wraps the deploy shell scripts. That's less funny, I suppose.

jwhitlock:
The deployment repo has been updated, thanks @milescrabill!

PR #4913, opened 2019-04-26
bug 1542395: implement verifyprocessed Django cmd

willkg:
This rewrites the verify-processed crontabber job as a Django command.

Since verifyprocessed accesses s3, I had to fix the configuration for s3 when running tests--it was all kinds of wrong. It's better now.


docker/config/test.env
```diff
@@ -13,10 +13,8 @@ RAVEN_DSN=
 # Boto
 AWS_ACCESS_KEY=something
 AWS_SECRET_ACCESS_KEY=something
-# NOTE(willkg): We want boto to use connect_to_region, so we make this an
-# empty string
-AWS_HOST=
-# Use test buckets
```
willkg: This was bogus. We really want to use the default boto ConnectionContext things since that's what moto will handily mock out. So this fixes that.

webapp-django/crashstats/crashstats/management/commands/bugassociations.py
```diff
@@ -79,20 +79,18 @@ def add_arguments(self, parser):
...
+            start_date = parse_datetime(start_date_arg)
             if not start_date:
                 # Try to parse it as a date
-                start_date = parse_date(start_date)
+                start_date = parse_date(start_date_arg)
             if not start_date:
-                raise CommandError(
-                    'Unrecognized last_success format: %s' % options.get('last_success')
-                )
+                raise CommandError('Unrecognized last_success format: %s' % start_date_arg)
```
willkg: I'm pretty sure this code was wrong before in the sense that line 88 stomps on `start_date` if the value isn't in date format. After that, everything is bananas.

I figured I'd fix it before I copy-and-paste the approach for verifyprocessed.

webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
```diff
@@ -0,0 +1,163 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
willkg: This file is virtually identical except I replaced RAW SQL with the Django model.

webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
```diff
@@ -0,0 +1,163 @@
...
+        check_crashids_for_date = partial(
+            check_crashids,
+            boto_conn=boto_conn,
+            bucket_name=bucket_name,
+            date=date,
+        )
+
+        missing = []
+        entropy = self.get_entropy()
+        if num_workers == 1:
```
willkg: I add this `if num_workers == 1` thing because it make testing easier.

webapp-django/crashstats/settings/test.py
```diff
@@ -23,10 +23,5 @@
     'django.contrib.auth.hashers.MD5PasswordHasher',
 )
 
-# Fix some implementations configuration
-SOCORRO_IMPLEMENTATIONS_CONFIG['resource']['boto'] = {  # noqa
-    'bucket_name': 'crashstats-test'
-}
```
willkg: This was all kinds of wrong and caused the tests to hang because it uses the local dev environment values for creating a boto ConnectionContext and that causes moto to be sad and fail.

willkg:
self-r+

PR #4915, opened 2019-04-29
bug 1542388: archivescraper fixes

willkg:
This fixes a few issues with archivescraper. One is a bug and the others are to make it produce better logging so we can see what's going on later.

### DIFF
--- a/docker/run_update_data.sh
+++ b/docker/run_update_data.sh
@@ -10,8 +10,7 @@
 set -eo pipefail
 
 # Fetch release data (verbosely)
-./socorro-cmd crontabber --reset-job=archivescraper
-./socorro-cmd crontabber --job=archivescraper
+webapp-django/manage.py archivescraper
 
 # Insert data that's no longer on archive.mozilla.org
 python ./scripts/insert_missing_versions.py
--- a/webapp-django/crashstats/crashstats/management/commands/archivescraper.py
+++ b/webapp-django/crashstats/crashstats/management/commands/archivescraper.py
@@ -133,10 +133,7 @@ class Downloader:
         if verbose and not self.verbose:
             return
 
-        if self.num_workers > 1:
-            self.msgs.append(msg)
-        else:
-            print(msg)
+        self.msgs.append(msg)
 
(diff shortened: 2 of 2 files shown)


docker/run_update_data.sh
```diff
@@ -10,8 +10,7 @@
 set -eo pipefail
 
 # Fetch release data (verbosely)
-./socorro-cmd crontabber --reset-job=archivescraper
-./socorro-cmd crontabber --job=archivescraper
```
willkg: This caused "make updatedata" to fail in the local dev environment. Oops!

webapp-django/crashstats/crashstats/management/commands/archivescraper.py
```diff
@@ -118,6 +118,25 @@ def __init__(self, base_url, num_workers, verbose):
...
+        :arg bool verbose: whether to write this only if in verbose mode
+
+        """
+        if verbose and not self.verbose:
+            return
+
+        if self.num_workers > 1:
+            self.msgs.append(msg)
+        else:
+            print(msg)
```
willkg: If there's one worker, then everything is happening in the parent process and we can skip the whole message buffer thing. If there's more than one worker, then parts are running in subprocesses.

Now that I'm thinking about this, I think I may nix the short-cut here. I'd rather it was getting tested in the tests.

webapp-django/crashstats/crashstats/management/commands/archivescraper.py
```diff
@@ -208,62 +226,10 @@ def get_json_links(self, path):
 
         return all_json_links
 
-    def scrape_candidates(self, product_name, archive_directory, major_version):
```
willkg: I moved this below the other method. This method is executed in the parent process and kicks off the worker processes so it makes sense (to me) to have it on the outside rather than buried in the class.

willkg:
Most of the changes in this PR are related to logging--what gets logged and how it's formatted. I'm going to self-r+ this and then verify it on stage. It runs once an hour, so it's pretty easy to do.

PR #4916, opened 2019-04-29
bug 1547807: cronrun fixes

willkg:
Series of minor fixes to cronrun and friends to make them a little easier to debug going forward.

### DIFF
--- a/webapp-django/crashstats/cron/admin.py
+++ b/webapp-django/crashstats/cron/admin.py
@@ -41,6 +41,7 @@ class LogAdmin(admin.ModelAdmin):
         'success',
         'pretty_error'
     ]
+    list_filter = ['app_name']
 
     def pretty_duration(self, obj):
         if not obj.duration:
--- a/webapp-django/crashstats/cron/management/commands/cronlist.py
+++ b/webapp-django/crashstats/cron/management/commands/cronlist.py
@@ -15,6 +15,17 @@ class Command(BaseCommand):
     def handle(self, **options):
         for job_spec in JOBS:
             self.stdout.write(job_spec['cmd'])
+
+            cmdline = [job_spec['cmd']]
+            if job_spec.get('cmd_args'):
+                cmdline.extend(job_spec['cmd_args'])
+            if job_spec.get('backfill'):
+                cmdline.append('--run-time=RUNTIME')
+            if job_spec.get('last_success'):
+                cmdline.append('--last-success=LASTSUCCESS')
+            cmdline = ' '.join(cmdline)
+            self.stdout.write('    cmdline:      %s' % cmdline)
+
             schedule = []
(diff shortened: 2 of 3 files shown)


webapp-django/crashstats/cron/admin.py
```diff
@@ -41,6 +41,7 @@ class LogAdmin(admin.ModelAdmin):
         'success',
         'pretty_error'
     ]
+    list_filter = ['app_name']
```
willkg: This makes it easier to look at all the log entries for a specific app. Makes it easier to diagnose issues.

webapp-django/crashstats/cron/management/commands/cronlist.py
```diff
@@ -15,6 +15,17 @@ class Command(BaseCommand):
...
+
+            cmdline = [job_spec['cmd']]
+            if job_spec.get('cmd_args'):
+                cmdline.extend(job_spec['cmd_args'])
+            if job_spec.get('backfill'):
+                cmdline.append('--run-time=RUNTIME')
+            if job_spec.get('last_success'):
+                cmdline.append('--last-success=LASTSUCCESS')
+            cmdline = ' '.join(cmdline)
+            self.stdout.write('    cmdline:      %s' % cmdline)
```
willkg: This code is sort of the same as the code in `cronrun`, but it's for displaying the command line--not running. So the output is a little different. Otherwise I would have surfaced it because it sort of computes the same thing and now we have it in two places. We can revisit this later.

Seeing the full command line will be AWESOME for backfill jobs and jobs that do things depending on their last success.

webapp-django/crashstats/cron/management/commands/cronrun.py
```diff
@@ -148,6 +148,8 @@ def _run_one(self, job_spec, force=False, cmd_args=None):
                         # the last_success as an argument
                         cmd_kwargs['last_success'] = format_datetime(last_success)
 
+                    logger.info('running: %s %s %s', cmd, cmd_args, cmd_kwargs)
```
willkg: This makes it easier to see all the actual run lines so we know which commands got which arguments.

webapp-django/crashstats/cron/management/commands/cronrun.py
```diff
@@ -240,7 +242,7 @@ def lock_job(self, cmd):
                 return
 
         # Can't lock it, so raise an error
-        raise OngoingJobError(job.ongoing)
+        raise OngoingJobError('%s: %s' % (job.app_name, job.ongoing))
```
willkg: This fixes the records in Sentry so we can see which job was ongoing.

willkg:
These fix some sharp edges with the new cronrun code. I'm excited about them. self-r+

PR #4917, opened 2019-04-29
Bug 1543097: Convert crash reporting from raven to sentry_sdk

jwhitlock:
Switch our crash reporting library from raven to sentry_sdk 0.7.14. This is a follow-on for PR #4912, which didn't avoid huge test diffs like I planned 😢.

## Switching from Raven to Sentry SDK

* The new SDK works best when initialized with [sentry_sdk.init()](https://docs.sentry.io/error-reporting/configuration/?platform=python) near the start of execution.
   - In the webapp, this is done in ``webapp-django/crashstats/settings/base.py``, and is very close to the documentation for Django integration.
   - For configman-based apps, ``setup_crash_reporting()`` is now called in the top-level ``App`` in `` socorro/app/socorro_app.py``, in the ``run`` method. This allows dropping ``sentry.sdk`` handling from various derived apps.
* The helper ``socorro/lib/sentry_client.py`` has a few changes:
   - The SDK uses a ``Hub`` to keep track of state, data scopes, and the client, and ``init`` sets up a global ``Hub`` that is used by default. The function ``get_client``, which initialized a raven client with a DSN, is replaced by ``get_hub``, which expected ``sentry_sdk.init()`` has been called, and does much less, but is still a good point for test mocking.
   - A new function ``is_enabled`` checks if ``sentry_sdk.init()`` has been run with a valid DSN. Tests mock this to return ``True`` rather than initialize with a fake DSN.
   - The function ``capture_error`` drops ``sentry_dsnn`` as the first argument. It uses ``is_enabled`` to decide whether to send the exception to Sentry or log it.
* The SDK methods have also changed from raven to sentry_sdk. ``captureException`` is ``capture_exception`` and takes a named parameter ``error``, ``captureMessage`` is ``capture_message``, etc.

## Filtering sensitive data from events

The previous Raven filters, added in PR #4357, sanitized the Sentry events.  Similar filtering is a requirement for switching to the new code. However, it is hard to verify the filtering implementation because there are no tests for the existing filtering (it is assumed Raven has those tests), the event format has changed significantly with the new SDK (apparently to standardize events across languages), and some collected data varies from development to deployment environments. This PR makes an effort to replicate the Raven filters, and I expect that there will be some additional filtering needed once we see events in the stage deployment environment.

In ``sentry_sdk``, a lot of sensitive data collection is covered by the ``init`` parameter ``send_default_pii``. It is disabled by default, and when enabled it selectively adds data:

* In the [Django integration](https://github.com/getsentry/sentry-python/blob/ea289c985b624050cf99f98933d614b8fa9ef007/sentry_sdk/integrations/django/__init__.py#L259-L281), it sends user data like an ID, email and username.
* The ``wsgi`` integration, included in the Django integration, includes the [user's IP address](https://github.com/getsentry/sentry-python/blob/a843018ba552882e646bc420c969d2a22b1c7856/sentry_sdk/integrations/wsgi.py#L226-L228), [headers with IP address info](https://github.com/getsentry/sentry-python/blob/a843018ba552882e646bc420c969d2a22b1c7856/sentry_sdk/integrations/wsgi.py#L101-L104), and [headers for cookies or auth](https://github.com/getsentry/sentry-python/blob/b47ed93901bbb7cfc35179ca3001f91bafaeb481/sentry_sdk/integrations/_wsgi_common.py#L126-L135). It will [collect cookies](https://github.com/getsentry/sentry-python/blob/b47ed93901bbb7cfc35179ca3001f91bafaeb481/sentry_sdk/integrations/_wsgi_common.py#L30-L31) for easier display.
* None of the [default integrations](https://github.com/getsentry/sentry-python/blob/b47ed93901bbb7cfc35179ca3001f91bafaeb481/sentry_sdk/integrations/__init__.py#L42-L51), which are used by processor and other configman-based tools, vary on ``send_default_pii``.

This PR explicitly sets ``send_default_pii=False``, so future devs will know we decided to omit this data rather than unknowingly accepted the default. Skipping cookies does the same work as the Raven filters for keywords ``sessionid``, ``csrftoken``, ``anoncsrf``, and ``sc``.

The [SanitizePasswordProcessor](https://github.com/getsentry/raven-python/blob/03559bb05fd963e2be96372ae89fb0bce751d26d/raven/processors.py#L155-L164) added some other keys that look like authorization data, such as ``password`` and ``api_key``. Additional code is needed to filter similar data with ``sentry_sdk``, which take a ``init`` param ``before_send`` to [filter events and breadcrumbs](https://docs.sentry.io/error-reporting/configuration/filtering/?platform=python).

When analyzing events in the development environment, only the web app appeared to have PII associated with the visitor, so the processing code lives in the web app. This PR adds a ``SentryProcessor`` class that runs a sequence of filters against an event, and includes some optional debug logging for development. The code could be trivially expanded to sanitize events from the processor app, or to sanitize breadcrumbs (data gathered during execution that may be useful if an exception occurs).

The current filters are:
* Filter SQL query breadcrumbs that include a sensitive column, such as ``email``, ``username``, ``session_key`` or ``tokens_token.key``, to avoid exposing PII or security data.
* Mask ``Auth-Token`` in HTTP headers
* Mask ``csrfmiddlewaretoken`` in POST data
* Mask ``code`` and ``state``, used in the OAuth flow, from query strings

## Future work

I was unable to get enough data in the development environment to write some planned filters. I hope to return to these with data from stage or production:
* Masking sensitive variables in call stacks (none found in dev)
* Masking sensitive environment from gunicorn breadcrumbs (not recorded in dev, may be Raven-specific)

This code should be merged at a time when we'll have a couple of working days to gather data in stage and create new filters as needed.

## Testing this PR

For my local testing, I added these to ``my.env``:

```
SENTRY_DSN=https://bad-9c7b6b6bcb7a45d7ae01fe8f60594083@sentry.127.0.0.1.nip.io/0
SENTRY_DEBUG=True
resource.sentry.debug=True
```

The ``SENTRY_DSN`` could be set to send development events to our Sentry collector for stage, which was done in testing for PR #4357, but I found the Sentry debug logging to be sufficient for development of the code.


socorro/app/socorro_app.py
```diff
@@ -206,6 +216,12 @@ def fix_exit_code(code):
 
             config_manager.log_config(mylogger)
 
+            # Add version to crash reports
+            version = (revision_data.get('version') or
```
jwhitlock: This is a copy of ``socorro.lib.revision_data.get_version`` that doesn't read the JSON from disk again. This may be a mis-optimization. I'm not sure if ``get_version`` should take an optional ``revision_data``, or I should just close my eyes and call ``get_version`` 🙈.

socorro/unittest/app/test_socorro_app.py
```diff
@@ -22,8 +22,9 @@ def test_instantiation(self):
         with pytest.raises(NotImplementedError):
             sa.main()
 
+    @mock.patch('socorro.app.socorro_app.setup_crash_reporting')
```
jwhitlock: This mock avoids``sentry_sdk`` choking when treating a ``MagicMock`` as a DSN string.

socorro/unittest/processor/test_processor_2015.py
```diff
@@ -31,28 +31,12 @@ def get_config(self):
...
-        def mock_Client(dsn):
-            client.dsn = dsn
-            client.captureException.side_effect = mock_capture_exception
-            return client
-
-        mock_get_client.side_effect = mock_Client
-
+    @patch('socorro.lib.sentry_client.get_hub', side_effect=Exception('fail'))
+    @patch('socorro.lib.sentry_client.is_enabled', return_value=False)
+    def test_rule_error_sentry_disabled(self, is_enabled, mock_get_hub):
```
jwhitlock: ``test_rule_error`` used to test both with a blank DSN and a fake DSN. When converting to a mocked ``is_enabled``, it made sense to split the test.

webapp-django/crashstats/cron/management/commands/cronrun.py
```diff
@@ -60,18 +59,14 @@ def add_arguments(self, parser):
...
-        try:
-            if options['job']:
-                job_args = options.get('job_arg') or []
-                # Re-add the -- because they're optional arguments; note that this
-                # doesn't support positional arguments
-                cmd_args = ['--%s' % arg for arg in job_args]
-                return self.cmd_run_one(options['job'], options['force'], cmd_args)
-            else:
-                return self.cmd_run_all()
-        except Exception:
```
jwhitlock: I dropped this ``try`` block because the default handler does what the ``except`` block does - log the uncaught exception to Sentry on exit.

socorro/cron/crontabber_app.py
```diff
@@ -597,16 +597,6 @@ class CronTabberApp(App, RequiredConfig):
...
 
-    # sentry
-    required_config.namespace('sentry')
-    required_config.sentry.add_option(
-        'dsn',
-        doc='DSN for Sentry',
-        default='',
-        reference_value_from='secrets.sentry',
-        secret=True
-    )
```
willkg: Nice--definitely want to have lifted this.

willkg:
Nice work!

socorro/app/socorro_app.py
jwhitlock: I went with the optional ``revision_data``.

socorro/app/socorro_app.py
```diff
@@ -107,6 +108,15 @@ class App(RequiredConfig):
         from_string_converter=str_to_list,
     )
 
+    # Sentry handles reporting unhandled exceptions.
+    required_config.namespace('sentry')
+    required_config.sentry.add_option(
+        'dsn',
+        doc='DSN for Sentry',
+        default='',
+        reference_value_from='secrets.sentry',
```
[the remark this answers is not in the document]
jwhitlock: Done in follow-on commit

jwhitlock:
I'm guessing another day to write and test a ``before_send`` function. To get an idea what we're currently collecting, see the [ZeroDivisionError](https://sentry.prod.mozaws.net/share/issue/cb5a2ea8805949a1a042e4d945edac64/) from stage. I may need to send some data from my developer environment to stage with this library for comparison.

socorro/lib/revision_data.py
```diff
@@ -46,15 +46,17 @@ def get_revision_data():
...
 
     This pulls revision data and then returns the best version-y thing
     available: the tag, the commit, or "unknown" if there's no revision
     data.
 
+    :arg revision_data: revision data, or None to fetch from disk
     :returns: string
 
     """
-    revision_data = get_revision_data()
```
willkg: I don't think the result of `get_version` will ever change during a run, so we could cache the result in the module and then return that. We do module-level caching in other places, too, so there's some precedent.

It's up to you.

webapp-django/crashstats/settings/base.py
```diff
@@ -586,25 +587,14 @@ def filter(self, record):
...
-        'processors': (
-            'raven.processors.SanitizeKeysProcessor',
-            'raven.processors.SanitizePasswordsProcessor',
-        )
-    }
+    sentry_sdk.init(
+        dsn=SENTRY_DSN,
+        release=SOCORRO_REVISION,
+        send_default_pii=True,
+        integrations=[DjangoIntegration()])
```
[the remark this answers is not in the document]
g-k: We should try to maintain raven's client side filtering to avoid accidentally collecting PII.

We have audited Sentry for storing PII (part of the reason we self-host it), can filter it server side, and can probably purge PII if we do collect it, but not collecting it in the first place would be best.

Looking at https://github.com/getsentry/sentry-python/issues/211 it looks like there isn't an equivalent for the raven filters and processors and we'd need to handle that in a `before_send` hook (looks like the raven filters are matching on field names and value regexes https://github.com/getsentry/raven-python/blob/master/raven/processors.py#L165).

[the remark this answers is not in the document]
g-k: jwhitlock pointed out some additional nuance, so I'm going to run this by the rest of the secops team.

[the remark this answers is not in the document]
g-k: Perfect timing on bringing this up right before the secops team meeting! So our consensus was that if we can filter out more PII client side we should do that, but it's also OK to continue collecting the same user fields since crash reports already contain user submitted PII and we need to associate users with their crashes.

And if the new sdk does less filtering we should try to preserve the behavior of raven's filters.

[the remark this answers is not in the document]
willkg: I just chatted with @g-k and he pointed out that web users don't opt-in to webapp crash reports to sentry and so they're not really the same thing as firefox crash reports to socorro.

I've been on Socorro for like 3 years and I don't think I've looked at any of those fields before. Sure, past is not prologue, but I think that's probably a good indicator.

I vote we go the easy route and do `set_default_pii=False`. If that turns out to be a problem in the future, we can revisit.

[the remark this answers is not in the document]
g-k: OK some things from further discussion with willkg:

* "webapp crashes -> sentry are really different than firefox crashes -> socorro" (especially in that they aren't opt-in)

* socorro usually goes out of its way to **not** associate Fx crashes with users (as in it's a legal or data requirement), so that's not a reason to collect PII

* willkg can't recall when he last used the PII fields so we can set that flag to false to not collect them

* we should re-implement the sanitizers from raven to preserve existing behavior (and upstream the change if sentry is open to it)

willkg:
I'd feel better if this baked on stage for a bit. I want to push the stuff that's on stage now to prod before we land this. I'm thinking we'll do a prod deploy tomorrow. I'm going to add the DO NOT MERGE label so I don't forget.

jwhitlock:
Today was research and a little code. Setting ``send_default_pii = True`` adds some data:

* In the [Django integration](https://github.com/getsentry/sentry-python/blob/ea289c985b624050cf99f98933d614b8fa9ef007/sentry_sdk/integrations/django/__init__.py#L259-L281), it sends user data like an ID, email and username.
* The ``wsgi`` integration, included in the Django integration, includes the [user's IP address](https://github.com/getsentry/sentry-python/blob/a843018ba552882e646bc420c969d2a22b1c7856/sentry_sdk/integrations/wsgi.py#L226-L228), [headers with IP address info](https://github.com/getsentry/sentry-python/blob/a843018ba552882e646bc420c969d2a22b1c7856/sentry_sdk/integrations/wsgi.py#L101-L104), and [headers for cookies or auth](https://github.com/getsentry/sentry-python/blob/b47ed93901bbb7cfc35179ca3001f91bafaeb481/sentry_sdk/integrations/_wsgi_common.py#L126-L135). It will [collect cookies](https://github.com/getsentry/sentry-python/blob/b47ed93901bbb7cfc35179ca3001f91bafaeb481/sentry_sdk/integrations/_wsgi_common.py#L30-L31) for easier display.
* None of the [default integrations](https://github.com/getsentry/sentry-python/blob/b47ed93901bbb7cfc35179ca3001f91bafaeb481/sentry_sdk/integrations/__init__.py#L42-L51), which are used by processor and other configman-based tools, vary on ``send_default_pii``.

We can turn off ``send_default_pii`` without losing useful debugging data ✅ .

The [raven processors](https://github.com/getsentry/raven-python/blob/d891c20f0f930153f508e9d698d9de42e910face/raven/processors.py#L68-L185) work by trying to automatically filter sensitive data by keywords across data, HTTP headers, and stacktraces. We could potentially port this, but the [Sentry dev says](https://forum.sentry.io/t/switching-to-sentry-python/4733/3):

> We don’t have a built-in password processor (mostly because it was never working perfectly for all usecases), but you can modify your events using [before_send](https://docs.sentry.io/learn/filtering/?platform=python)

This makes me think that copying the raven logic to our ``before_send`` and ``before_breadcrumb`` methods will potentially send sensitive data. Breadcrumbs, which contain SQL queries and gunicorn data, seem to be the likely candidates in testing. I think that we should add targeted filtering, and use some of the testing techniques from PR #4357, including some time in stage, and reading the event JSON closely.

jwhitlock:
I'll need to get a few more examples of filters before I can get a good API. I thought plain functions would be the simplest thing that could work, but since there is a little configuration and registration, classes with a ``__call__`` method might be better.

Filters to write:
* Call stack: Look for sensitive variable names like ``password``, ``csrf``, etc. Might be in locals or calling arguments
* Gunicorn logs: Anonomize sensitive headers like ``REMOTE_ADDR``

There are probably a few more. I'm getting the low-hanging fruit in the dev environment, and then looking at existing reports for more ideas. I'll push my work in progress for the curious, but no need to review the PR yet.

g-k:
:+1: I was assuming we had legal or compliance requirements to use the old SDK filters, but if they are broken or incomplete then we don't have to port that behavior over (especially since we can set `send_default_pii=False` now and filter server side).

jwhitlock:
I think @Osmose did a careful job of validating the raven-provided filters in PR #4357, and I want to make sure we're not adding data that was previously filtered.

The event format has changed with sentry_sdk, probably to standardize formatting across languages and platforms, which means I can't use existing raven-generated events to determine how to write my filters. The production stack (such as AWS load balancers) also adds some headers that I can't test locally, because I'm not sure how they will show up in the sentry_sdk event format. That's OK, we'll be mostly covered by ``send_default_pii=False``, and I'll be able to quickly add some more rules as needed after capturing an event in staging. But it means I won't be able to filter gunicorn headers in this PR, since I can't see them in development (which means they may not even be present in staging).

That leaves a couple more tasks that I can do in development:
* Sanitize ``POST``s, along with CSRF in form data and headers
* Sanitize API token usage
* Sanitize the call stack variables

I hope to open for review by Monday.

jwhitlock:
This is ready for review again.

I developed the sanitizers by turning up debugging in ``my.env``:

```
SENTRY_DSN=https://bad-9c7b6b6bcb7a45d7ae01fe8f60594083@sentry.127.0.0.1.nip.io/0
SENTRY_DEBUG=True
resource.sentry.debug=True
```

That fake DSN ensured that Sentry events were generated but went nowhere. I then added some ``raise Exception('here')`` lines, generated a Sentry event, formatted with ``jq``, looked for trouble, and wrote some code. Rinse and repeat.

The code is overly flexible:

- I added the ability to process breadcrumbs, which are added in response to SQL queries, logs, HTTP requests incoming and outgoing, etc. However, you can wait to sanitize these until the event is raised. It is possible that all the code to implement ``before_breadcrumb`` should be removed, and the code should focus on ``before_send``.
- I added ``before_send`` and ``before_breadcrumb`` handlers to the processor, but they just have optional debugging, and no filtering. I couldn't find any sensitive data outside of the crash data, which we're not filtering. That could mean all the filtering code belongs in the user-facing webapp, and shouldn't be in ``socorro/lib/sentry_client.py``.

The code is also too specific:

- I wrote a filter for each type of data: Headers, query strings, POST data, etc. A more generic filter could take a path in the event object using ``glom``, and then process it in a more generic way. However, each content type has wrinkles, and it was developed incrementally, so I went with a more targeted approach.

I'm not sure if I should tighten up the code, or refactor into some more general functions, or let it be for the first version.

As I mentioned earlier, the event format changed from raven to sentry_sdk, and the existing raven data is different than this data. If the gunicorn request details appear in staging, there may be more filtering to do, to omit IP address data from gunicorn header logs. I suspect there will be one or more rounds of changes.

jwhitlock:
I'm going to take another refactoring pass on this:

* Drop breadcrumb parsing, because it appears we won't need it
* Generalize the sanitizing algorithm, so that each data type isn't as much of a unique codepath

socorro/app/socorro_app.py
```diff
@@ -316,3 +346,24 @@ def setup_metrics(config):
...
+
+    if config.sentry and config.sentry.dsn:
+
+        before_breadcrumb = SentryProcessor(
+            'crumb',
+            log_name=SENTRY_LOG_NAME if config.sentry.debug_breadcrumbs else None)
+
+        before_send = SentryProcessor(
+            'event',
+            log_name=SENTRY_LOG_NAME if config.sentry.debug else None)
```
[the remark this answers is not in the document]
jwhitlock: Removed until needed.

jwhitlock:
I'm glad I had a chance to refactor this code, I'm a lot more comfortable with it. I've also edited the initial PR comment to cover the additional scope of filtering events for PII and secure information, so that future devs don't have to read all the 40+ comments.

We discussed in the meeting that this code is not urgent, and requires a couple of days to bake in stage. We can defer review and merging until we have that time. The code is in a good state to hang out on the shelf for a while, and I can rebase when needed.

webapp-django/crashstats/crashstats/tests/test_sentry.py
```diff
@@ -0,0 +1,323 @@
+"""Tests for Sentry event processing."""
+
+import pytest
+
+from ..sentry import (SanitizeBreadcrumbs,
```
willkg: I've been avoiding relative imports like the plague because it makes it hard to move things around and know from where they actually came in the first place.

We do use them in the signature generation code so that code can exist in multiple contexts. But other than that, I'd like to avoid them.

webapp-django/crashstats/crashstats/sentry.py
```diff
@@ -0,0 +1,327 @@
+"""Provides a before_send callable to sanitize Sentry events."""
```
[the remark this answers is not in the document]
willkg: Also, since this applies to the whole code base and we're importing this from settings, this shouldn't be in an app and should instead be in `webapp-django/crashstats/sentry.py`. That prevents potential bootstrap and circular import problems down the road.

Also, since it doesn't contain Sentry itself, but rather contains some utility things we wrote to use with Sentry, I'd call it `sentrylib.py` or `sentryutils.py`.

willkg:
Oh, I had one other thought. This adds a lot of code and while there are tests, they're tests for known scenarios and don't account for change in sentry-sdk. What happens when this code fails? How would we find out?

Maybe it makes sense to have the `before_send` in a big try/except block that logs the exception if anything happens? That gets us something in the logs that we can watch after deploys that might affect sentry things. It doesn't pro-actively notify us, though.

willkg:
> Exceptions in before_send

What if we wrap the `before_send` in a try/except that uses markus to send metrics to datadog whenever it fails and then have it re-raise? Something like this:

```
metrics = markus.get_metrics('sentry')

...

def before_send(whatever):
    try:
        whatever
    except Exception:
        markus.incr('before_send_exception')
        raise
```

If markus is configured at that point, it'll send something and we can see it in datadog.

Does that sound like a good idea? I'm not sure how else we can get signals out.

> Log level of before_send loggers

I suggested we switch it to debug because I thought we were looking at debug level in the local dev environment. Now that I know it's not logging at debug level, I take back my comment--let's leave it as what you have since if someone has debug mode enabled and it seems onerous and confusing to require multiple actions to get debugging to work.

willkg:
https://github.com/mozilla-services/socorro/pull/4917#discussion_r283934319

It's more readable to me if we adjust where the parens are. It's a tuple of three elements with the second element being a really long string. I think this is more readable:

```
        # Select a user by email
        (
            'email',
            (
                'SELECT "auth_user"."is_active" FROM "auth_user" WHERE'
                ' UPPER("auth_user"."email"::text) = UPPER(\'username@example.com\')'
            ),
            (
                '...'
            )
        )
```

When I was reviewing that code, I had to count parens to figure out where the three elements were demarcated. That seems really prone to error.

jwhitlock:
I think the logging is a little cleaner now. Instead of passing around a logger, there's a module-level logger that gets called when needed, and the logic to "turn on" the messages moves to settings. Take a look and see if you agree.

Incrementing a counter with ``markus`` will work to track exceptions. When we make it a library, I don't know if we want to require ``markus`` as a dependency. But, I think your code works for v1, and we can worry about the library later.

I'm continuing with migrating the return from ``(event, modified?)`` to just modified. Working through this change suggests another. The event is modified in place, and returned to signify that it wasn't dropped. None of our code drops events however, and it seems a misleading that we're modifying the input as well as returning it.

Instead, it seems better to raise an exception when then event (or breadcrumb) should be dropped instead of processed, catch the exception in ``SentryProcessor``, and return ``None``. But, none of the current filters drop events for crumbs, so I think I'm just going to drop the return value, and adjust docstrings to make it clear that the event is modified in place.

webapp-django/crashstats/tests/test_sentrylib.py
```diff
@@ -17,58 +17,46 @@
...
+            ' datetime.datetime(2019, 5, 3, 18, 7, 29, 852498)'
+            ' WHERE \"auth_user\".\"id\" = 123'),
+    }
 
     @pytest.mark.parametrize(
-        'keyword,message,expected', CASES,
-        ids=tuple(case[0] for case in CASES))
-    def test_filtered_queries(self, keyword, message, expected):
+        'keyword,sql', CASES.items(),
+        ids=tuple(case for case in CASES))
```
willkg: Does this line up right? Does `CASES.items()` return the same as `iter(CASES)`?

jwhitlock: In my experience, yes, and that's my understanding from the dictionary view docs: https://docs.python.org/3.6/library/stdtypes.html#dictionary-view-objects

jwhitlock:
My remaining task is to run in production mode, and see if I can get the sensitive breadcrumbs I'm seeing in the raven-generated events in stage. That may add another sanitizer, but I don't expect the current sanitizers to change, so this may be another good place to review.

I hope to squash all these review commits into the "Filer events for pii/security" commit, because I don't think there is a long-term benefit to all these follow-on commits. If you want me to squash before you review, let me know.

webapp-django/crashstats/tests/test_sentrylib.py
```diff
@@ -0,0 +1,302 @@
...
+    """Tests for SanitizeHeaders."""
+
+    # Test cases: names, URL, headers, expected headers
+    CASES = (
+        # Use an API token
+        ('Auth-Token',
+         'http://example.com/api/RawCrash/',
+         {'Auth-Token': '12345abcde', 'User-Agent': 'curl/7.54.0', 'Host': 'example.com'},
+         {'Auth-Token': '[filtered]', 'User-Agent': 'curl/7.54.0', 'Host': 'example.com'}),
+    )
```
willkg: This is another hard-to-parse thing. It's really hard to see a one-character difference that denotes whether something is in the tuple or not. This is better written as:

```
    CASES = (
        (
            'Auth-Token',
            'http://example.com/api/RawCrash/',
            {'Auth-Token': '12345abcde', 'User-Agent': 'curl/7.54.0', 'Host': 'example.com'},
            {'Auth-Token': '[filtered]', 'User-Agent': 'curl/7.54.0', 'Host': 'example.com'}
        ),
    )
```

webapp-django/crashstats/tests/test_sentrylib.py
```diff
@@ -0,0 +1,302 @@
...
+class TestSanitizePostData:
+    """Tests for SanitizePostData."""
+
+    # Test cases: names, URL, data, expected data
+    CASES = (
+        # Create a new token
+        ('csrfmiddlewaretoken',
+         'http://example.com/api/tokens/',
+         {'csrfmiddlewaretoken': 'base64_str', 'notes': 'Test', 'permissions': '18'},
+         {'csrfmiddlewaretoken': '[filtered]', 'notes': 'Test', 'permissions': '18'}),
```
willkg: This should get changed, too.

webapp-django/crashstats/tests/test_sentrylib.py
```diff
@@ -0,0 +1,302 @@
...
+class TestSanitizeQueryString:
+    """Tests for SanitizeQueryString."""
+
+    # Test cases: names, URL, querystring, expected filtered querystring
+    CASES = (
+        # code and state in the OIDC callback for OpenID flow
+        ('code,state',
+         'http://example.com/oidc/callback/',
+         'code=abc123&state=base64String',
+         'code=%5Bfiltered%5D&state=%5Bfiltered%5D'),
```
willkg: This should get changed, too.

webapp-django/crashstats/tests/test_sentrylib.py
```diff
@@ -0,0 +1,302 @@
...
+            ' "tokens_token"."key", "tokens_token"."expires",'),
+        # Update last login
+        'auth_user.id': (
+            'UPDATE \"auth_user\" SET \"last_login\" ='
+            ' datetime.datetime(2019, 5, 3, 18, 7, 29, 852498)'
+            ' WHERE \"auth_user\".\"id\" = 123'),
+    }
+
+    @pytest.mark.parametrize(
+        'keyword,sql', CASES.items(),
```
willkg: I've been putting spaces between the arguments. So then this would be:

```
'keyword, sql'
```

For this one, it's not a big deal, but for larger argument lists, it's a lot easier to read.

jwhitlock:
Fixes applied in follow-on commits. I'd like to squash these as well before merging.

My test plan:
* Use "Crash me now" to generate a webapp report in staging, look at the generated event
* Wait for a ``KeyError: 'java_stack_trace_'`` exception (2 in last 48 hours, might be possible to induce with re-submit), look at the generated event.

There's a good chance that there will be addition scrubbing needed in the deployed environments, based on crumbs that appear in the deployed ``raven`` events but not in the development environment with ``sentry_sdk``.

jwhitlock:
I was learning on this one too. Thanks for the feedback on coding styles, ``markus``, and the rest.

PR #4918, opened 2019-04-30
bug 1547411: remove crontabber app

willkg:
This removes the crontabber app which we no longer use. It keeps the
"crontabber" name as the name of the service/node because otherwise we have to
do a bunch of changes to the infrastructure.

"crontabber" is the name of the node and it runs the "cronrun" Django command.


willkg:
To test:

1. rebuild local dev environment (`make build`, `make setup`, `make updatedata`)
2. run crontabber service (`docker-compose up crontabber`)
3. build docs (`make docs`)
4. view `/monitoring/crontabber/` in browser -- redirects to `/monitoring/cron/`
5. view `/monitoring/cron/` in browser
6. search codebase for "crontabber"

Makefile
```diff
@@ -119,7 +119,3 @@ runservices: my.env
 .PHONY: stop
 stop: my.env
 	${DC} stop
-
-.PHONY: dependencycheck
-dependencycheck: my.env
-	${DC} run --rm crontabber shell ./docker/run_dependency_checks.sh
```
willkg: We can run this with `socorro-cmd depcheck` now, so we don't need this make rule. I updated the docs accordingly.

docs/flows/signature.rst
```diff
@@ -39,10 +39,9 @@ in the ``crashstats_signature`` PostgreSQL table.
 Where the data comes from
 =========================
 
-The ``socorro.crontabber.jobs.update_signatures.UpdateSignaturesCronApp``
-crontabber app looks at all the crash reports in Elasticsearch for some period
-of time, then goes through and updates the ``first_build`` and ``first_date``
-columns.
+The ``updatesignatures`` Django command looks at all the crash reports in
```
willkg: Oops--need to update the diagram, too.

webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
```diff
@@ -28,7 +28,7 @@
 PROCESSED_CRASH_TEMPLATE = 'v1/processed_crash/%s'
 
 
-metrics = markus.get_metrics('crontabber.verifyprocessed')
+metrics = markus.get_metrics('cron.verifyprocessed')
```
willkg: We'll have to update datadog dashboards after this deploys.

webapp-django/crashstats/monitoring/urls.py
```diff
@@ -3,14 +3,20 @@
...
 
 from crashstats.monitoring import views
 
 
 app_name = 'monitoring'
 urlpatterns = [
     url(r'^monitoring/$', views.index, name='index'),
-    url(r'^monitoring/crontabber/$', views.crontabber_status, name='crontabber_status'),
+    url(r'^monitoring/cron/$', views.cron_status, name='cron_status'),
+    # FIXME(willkg): DEPRECATED AS OF 4/30/2019
```
willkg: I guess I don't need to shout here.

willkg: Relatedly, we have a bunch of "redirect deprecated thing to new thing" things in the webapp. We should do a scrub pass and remove existing cruft.

docs/tests/system_checklist.rst
```diff
@@ -175,20 +175,16 @@ Checklist
     Crontabber
     ==========
 
-    Is crontabber working?
+    Is the crontabber node working?
 
-    * Check healthcheck endpoint (/monitoring/crontabber/)
```
jwhitlock: Should ``/monitoring/cron/`` be mentioned here or in ``cron.rst``?

willkg: Mmm... I think no. I want to phase it out, but I don't remember offhand where it's used if anywhere. It's much better to check the Django admin for Jobs and Logs. If we (admin) need an easier to interpret high-level view, we should add a Django admin page for it.

I think I'd nix related lines in the system checklist, too. I can do that now.

willkg: Oh, I already did it. Yay me!

willkg:
Thank you!

PR #4919, opened 2019-04-30
bug 1548158: remove postgres code

willkg:


willkg:
This is just code removal. self-r+

PR #4921, opened 2019-05-02
bug 1547807: add __str__ to Job and Log

willkg:
This makes it easier to know what's what in contexts like the Django admin.


willkg:
This makes the Django admin easier to read. Having said that, the `__str__` doesn't come up much. self-r+

PR #4922, opened 2019-05-02
bug 1544565: rename white/blacklist to allow/disallowlist

willkg:


socorro/external/es/crashstorage.py
```diff
@@ -441,7 +441,7 @@ class ESCrashStorageRedactedJsonDump(ESCrashStorageRedactedSave):
     """
     required_config = Namespace()
     required_config.add_option(
-        name="json_dump_whitelist_keys",
+        name="json_dump_allowlist_keys",
```
willkg: This is configuration, but we don't define it anywhere--we use the defaults from here.

webapp-django/crashstats/api/views.py
```diff
@@ -122,7 +122,7 @@ def clean(self):
 
 
 # Names of models we don't want to serve at all
-BLACKLIST = (
+API_DONT_SERVE_LIST = (
```
willkg: This is a more descriptive name.

willkg:
I think this is good. It's mostly internal, so it doesn't affect anyone. self-r+

PR #4923, opened 2019-05-02
bug 1546672: fix supersearch versions suggestions to include X.Yb

willkg:


webapp-django/crashstats/crashstats/utils.py
```diff
@@ -389,16 +389,26 @@ def get_versions_for_product(product='Firefox', use_cache=True):
...
         version = item['term']
         try:
             # This generates the sort key but also parses the version to
             # make sure it's a valid looking version
             versions.append((generate_version_key(version), version))
+
+            # Add X.Yb to betas set
+            if 'b' in version:
+                beta_version = version[:version.find('b') + 1]
+                betas.add(beta_version)
```
willkg: This is nice to do here because this work will also get cached. So, in general, this is a good move.

webapp-django/crashstats/crashstats/utils.py
```diff
@@ -432,22 +442,19 @@ def get_version_context_for_product(product):
...
-
     # Map of major version (int) -> list of (key (str), versions (str)) so
     # we can get the most recent version of the last three majors which
     # we'll assume are "featured versions"
     major_to_versions = OrderedDict()
     for version in versions:
+        # In figuring for featured versions, we don't want to include the
+        # catch-all-betas X.Yb faux version
+        if version.endswith('b'):
+            continue
```
willkg: Liz suggested that we shouldn't include esr versions in the featured versions list. That's part of the issue in bug 1545426. If we wanted to nix esr, then we'd do that here, too.

willkg:
I verified this with my local dev environment. One tricky thing here is that the version data is cached, so we have to wait an hour for the cache to expire. When this deploys to stage, we'll have to wait a while to see if it worked out.

self-r+

PR #4924, opened 2019-05-02
bug 1545426: don't include ESR versions in featured versions

willkg:


willkg:
I tested this locally by processing the problematic 66.0.3esr reports and verifying that without this change, they show up in featured versions and with this change, they go away.

The version data is cached, so once this deploys, it'll take an hour for the cache to expire and version information to get re-calculated. It should work after that happens.

self-r+

PR #4925, opened 2019-05-02
bug 1543988: update breakpad client to 4d550cceca107f36c4bc1ea1126b7d…

willkg:
…32cc50f424

This updates the breakpad client to match the crash reporter.


willkg:
This change is really straight-forward. We update the sha in `build-breakpad.sh`, then build the Docker image to make sure it and minidump-stackwalk build, then process some crashes as a spot-check to make sure minidump-stackwalk works.

Then we land it which deploys it to stage. We'll get better data from processing tons of crashes on stage.

willkg:
self-r+

PR #4926, opened 2019-05-03
bug 1548830: Fix API token examples

jwhitlock:
The RawCrash API uses ``crash_id``, not ``uuid``, for the querystring parameter. Fix the example to use the correct parameter.

### DIFF
--- a/webapp-django/crashstats/tokens/jinja2/tokens/home.html
+++ b/webapp-django/crashstats/tokens/jinja2/tokens/home.html
@@ -137,13 +137,13 @@
         </p>
         <p>Here's an example:</p>
         <p class="example">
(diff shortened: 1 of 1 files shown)


jwhitlock:
Things I didn't change but maybe could:

* In the Django example, the sample token ends with an 'x', so that it isn't a valid token. The curl version could be changed to match, or change into a string that makes it more obvious that your string goes here.
* The example could use the user's token and a valid crash ID, if available, so that it works on copy-and-paste. JS code would be needed to get the same "hide secrets" functionality as the API token display. Maybe "Show the whole token" can also swap in real values into the example...

jwhitlock:
👍 Let's wait a second person to reach out that they were confused about the sample code.

PR #4927, opened 2019-05-06
bug 1545446: Allow CDN for fira-sans in CSP

jwhitlock:
Loading Fira Sans requires a stylesheet and fonts, so https://code.cdn.mozilla.net is added to the CSP expected paths.

In local testing, this causes the pages to use Fira Sans. Once the font is loaded into Firefox, it is also used in production, although with CSP violations.

The font seems OK if thin to me, and the navigation text gets blurry on my normal DPI external monitor. We may want to adjust weights and sizes to those suggested on https://design.firefox.com/photon/visuals/typography.html

### DIFF
--- a/webapp-django/crashstats/settings/base.py
+++ b/webapp-django/crashstats/settings/base.py
@@ -751,6 +751,7 @@ CSP_SCRIPT_SRC = (
 CSP_STYLE_SRC = (
     "'self'",
     "'unsafe-inline'",
+    'https://code.cdn.mozilla.net',
 )
 CSP_IMG_SRC = (
     "'self'",
@@ -760,6 +761,10 @@ CSP_IMG_SRC = (
 )
 CSP_CONNECT_SRC = (
     "'self'",
+)
+CSP_FONT_SRC = (
+    "'self'",
+    'https://code.cdn.mozilla.net',
 )
 
 CSP_REPORT_URI = (


jwhitlock:
Here's Fira Sans on a normal DPI display on https://crash-stats.mozilla.com/search/ :

PR 4927 - Low DPI png

And on a High DPI display:

PR 4927 - High DPI

jwhitlock:
Is there a web interface for normandy? https://github.com/mozilla/normandy appears to be just an API, without browser-facing styles.

willkg:
The frontend is here: https://github.com/mozilla/delivery-console

jwhitlock:
Fira-Sans is brought into this, and other Mozilla projects, as Photon's suggested font for web sites. Normandy's admin uses React, and included the rules since the [project start](https://github.com/mozilla/delivery-console/pull/1) in March 2018, in a library called [photon-ant](https://www.npmjs.com/package/photon-ant). There were some style tweaks next month in https://github.com/mozilla/delivery-console/pull/53 to use Normandy-specific styles less and rely on ``photon-ant`` more. And then three months later, ``photon-ant`` was dropped in https://github.com/mozilla/delivery-console/pull/186, in order to use styles provided by React components instead (see [screenshots](https://github.com/mozilla/delivery-console/pull/186#issuecomment-397854635)). 

Normandy's choices feel like a mix of "don't fight the rest of our React tools" (which doesn't apply to the webapp), combined with an unwillingness to evolve Photon for a functional / admin-only website (which is closer to Socorro's webapp).

Another way to stop the CSP violation is to not load Fira Sans at all. This will not help [bug 1547316](https://bugzilla.mozilla.org/show_bug.cgi?id=1547316), which appears to be caused by different font choices across operating systems. But, we could tackle that independently of this CSP issue.

@willkg, how do you feel about removing Fira Sans, so that everyone gets the fallback fonts?  Here's where that is defined:

https://github.com/mozilla-services/socorro/blob/3fbb6bf4d70ada416ba18b80f29942cded0204bb/webapp-django/crashstats/crashstats/static/crashstats/css/base/variables.less#L6

I'd go a step further and reduce the font list to a boring "web safe" list:

```
@base-font: Helvetica, Arial, sans-serif;
```

In either case, I think we close this PR, it sounds like a dead end.

willkg:
I agree. I think let's nix this PR and go back to web safe fonts. If that turns out to be a problem, someone can write up a bug and we can get some design help to figure out a better solution.

Thanks, @jwhitlock !

PR #4928, opened 2019-05-07
bug 1513346: Convert six to native Python 3

jwhitlock:
Convert most ``six`` usage to native Python 3 code.

The remaining code is for ``six.reraise``. This could be replaced with ``from future.utils import raise_ as reraise``. Reading the code, it appears Python 3's exception chaining would be better for some cases, but I'm not up-to-date on Python 3 exceptions, and need to do my homework.

### DIFF
--- a/webapp-django/crashstats/authentication/management/commands/makesuperuser.py
+++ b/webapp-django/crashstats/authentication/management/commands/makesuperuser.py
@@ -9,10 +9,10 @@ environment.
 
 from django.contrib.auth.models import Group, User
 from django.core.management.base import BaseCommand, CommandError
-from django.utils.six.moves import input
 
 
 def get_input(text):
+    """Get user input, or mock it for tests."""
     return input(text).strip()
 
 
--- a/scripts/update_siggen.py
+++ b/scripts/update_siggen.py
@@ -9,8 +9,6 @@ import shutil
 import subprocess
 import sys
 
-import six
-
 
 DESCRIPTION = """
 Updates socorro-siggen extracted library with files from Socorro's
@@ -75,7 +73,7 @@ def main(argv=None):
     else:
         print('No previous sha to determine git commits.')
 
-    confirm = six.moves.input('Continue to copy? [y/N]: ')
+    confirm = input('Continue to copy? [y/N]: ')
     if confirm.strip().lower() != 'y':
         print('Exiting...')
         return 1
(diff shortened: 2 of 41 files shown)


webapp-django/crashstats/authentication/management/commands/makesuperuser.py
```diff
@@ -9,11 +9,11 @@
 
 from django.contrib.auth.models import Group, User
 from django.core.management.base import BaseCommand, CommandError
-from django.utils.six.moves import input
 
 
 def get_input(text):
-    return input(text).strip()
+    """Get user input, or mock it for tests."""
+    return input(text)
```
willkg: We dropped the `.strip()` here. Having the `.strip()` handles copy-and-paste issues where the user inadvertently adds whitespace.

jwhitlock: Ah you caught that 😁 . I thought it was weird we were double-striping the output:

https://github.com/mozilla-services/socorro/blob/b684523243ca23b9e85578e3cea855ff775d7cbf/webapp-django/crashstats/authentication/management/commands/makesuperuser.py#L31

jwhitlock:
``reraise`` will be a bit more complicated, because of weirdness like this:

https://github.com/mozilla-services/socorro/blob/989913ddbfd0efa202758666fc780f091ee48981/socorro/processor/processor_app.py#L243-L259

It is definitely worth its own PR.

webapp-django/crashstats/authentication/management/commands/makesuperuser.py
willkg: Oh weird. I thought I would have done the stripping in `get_input()` and not at the usage point. Ok--past me probably knew best maybe or something.

willkg:
Oh, right. `PolyCrashStorage`. Yeah, that might be a little trickier. I think it's time we nix `PolyCrashStorage`. I'll mull over that.

webapp-django/crashstats/authentication/management/commands/makesuperuser.py
jwhitlock: My guess is that tests were failing with a mocked ``get_input()``

jwhitlock: Or maybe pipe a list of emails, with possible whitespace, although I'm not sure it would work, and why would you do that?

jwhitlock: Adding it back, this PR should not introduce code changes.

PR #4929, opened 2019-05-07
bug 1549540: surface signal issues from minidump-stackwalk

willkg:
This adds `SIGKILL` and `SIGABRT` to logging and processor notes so we can more
easily find crash reports which didn't process well.

We can add other signals as they pop up in the exit code graphs on Datadog.

### DIFF
--- a/socorro/processor/breakpad_transform_rules.py
+++ b/socorro/processor/breakpad_transform_rules.py
@@ -289,9 +289,9 @@ class BreakpadStackwalkerRule2015(ExternalProcessRule):
             msg = 'MDSW failed with %s: %s' % (
                 return_code, stackwalker_data.mdsw_status_string
             )
+            # subprocess.Popen with shell=False returns negative exit codes
+            # where the number is the signal that got kicked up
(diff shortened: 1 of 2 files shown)


socorro/processor/breakpad_transform_rules.py
```diff
@@ -280,14 +281,21 @@ def _execute_external_process(self, command_line, processor_meta):
...
 
         elif return_code != 0 or not stackwalker_data.success:
-            msg = 'MDSW failed with %s: %s' % (return_code, stackwalker_data.mdsw_status_string)
+            msg = 'MDSW failed with %s: %s' % (
+                return_code, stackwalker_data.mdsw_status_string
+            )
+            # subprocess.Popen with shell=False returns negative exit codes
+            # where the number is the signal that got kicked up
+            if return_code == -6:
+                msg = msg + ' (SIGABRT)'
```
willkg: We can add other signals as we see them. This is the only one we're seeing now.

willkg:
self-r+

PR #4931, opened 2019-05-08
bug 1545446: Remove Fira-Sans, reduce font list

jwhitlock:
Stop loading Fira Sans from the CDN, rather than add a CSP exception, as tried and discussed in PR #4927. 

Reduce the default font list to a boring set. Here's what Helvetica looks like:

PR 4931 - Helvetica

### DIFF
--- a/webapp-django/crashstats/crashstats/static/crashstats/css/base/fonts.less
+++ b/webapp-django/crashstats/crashstats/static/crashstats/css/base/fonts.less
@@ -1,5 +1,3 @@
-@import "https://code.cdn.mozilla.net/fonts/fira.css";
-
 @font-face {
     font-family: 'icons';
     src: url('/static/crashstats/fonts/icons.eot?#iefix2013') format('embedded-opentype'),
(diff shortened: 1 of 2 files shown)


willkg:
I think this is fine, but it messes up the text alignment in the input date/time fields in super search so that I see what Dana is setting in https://bugzilla.mozilla.org/show_bug.cgi?id=1547316 .

I can't figure out what's causing that. The baseline for the text inside the input date/time fields and the UTC next to them is too high. I don't understand enough CSS to know what's going on, though. I fiddled with it for a bit, but didn't get anywhere.

image

I think I want to hold off on landing this until we can fix that, too.

jwhitlock:
@willkg which font does your browser end up using? Can you try production with just "Fira Sans" removed, and see what font is used?

jwhitlock:
@willkg reported in a meeting that he was using Helvetica, which means my macOS Helvetica is different from his Linux Helvetica 😢. Time to read [Styling HTML Forms](https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/Styling_HTML_forms)...

PR #4933, opened 2019-05-09
bug 1550043: rough mdswshell and docs

willkg:
This creates a rough mdswshell in a container that lets you run minidump-stackwalk
as the processor does, adjust the code, compile it, run gdb, etc.

It adds some rough docs, too.


Makefile
```diff
@@ -19,6 +19,7 @@ ifeq (1, ${NOCACHE})
 DOCKER_BUILD_OPTS := --no-cache
 endif
 
+DOCKER := $(shell which docker)
```
willkg: Oops--I can take this out.

docker/Dockerfile
```diff
@@ -14,14 +14,21 @@ RUN DEBIAN_FRONTEND=noninteractive /tmp/set_up_ubuntu.sh
 
 
 FROM socorro_image_base as socorro_breakpad
+
 WORKDIR /mdsw/
 
+# Install some helpful debugging things
+RUN apt-get -y install gdb vim
```
willkg: This will only exist in this intermediary container and it doesn't make it into the final socorro_app image that we use in stage and prod.

docker/run_mdswshell.sh
```diff
@@ -0,0 +1,59 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Sets up a shell in a container for building and debugging minidump-stackwalk.
+
+set -e
+
+INDOCKER=${INDOCKER:-0}
+
+if [[ ${INDOCKER} -eq 1 ]]; then
```
willkg: This script is so cool. It runs outside the docker container to set everything up, then reruns itself *inside* the docker container to finish things off. I keep meaning to rewrite the tests stuff this way, too.

docker/run_mdswshell.sh
```diff
@@ -0,0 +1,59 @@
...
+    # Save the virtualenv in /app because that's mounted locally and saves some
+    # time between invocations
+    python -m venv /app/mdsw_venv/
+    source /app/mdsw_venv/bin/activate
+    pip install -r requirements/default.txt -c requirements/constraints.txt
```
willkg: The socorro_breakpad intermediary docker image kicks off *before* it installs Python requirements. That's good for building the socorro_app image, but it sucks here because we want a shell in a container with all the things we need to build `minidump-stackwalk` **and** the Python requirements so we can use the scripts to download crash data and run `minidump-stackwalk` against it.

Given that this is a debugging tool, I decided to go for "gross, but it works". Because installing a lot of Python requirements every time you invoke `run_mdswshell.sh` stinks, I'm "saving that state" in a virtualenv. It's gross because now users have a virtualenv that's only used for mdswshell in their socorro root directory which make grepping and a bunch of other things work poorly. We can yak-shave on this some other day. My current prediction is that I will use the `mdswshell` and maybe two other people in the whole world, so gross is ok.

willkg:
To test this out (and make sure it doesn't create problems for processing or other things):

1. `make clean`
2. `make build`
3. `make test`
4. `make run` and then process some crashes and compare the signatures with what we have on prod
5. `make shell` and run `scripts/run_mdsw.sh` and make sure it runs

willkg:
This looks good enough. self-r+

PR #4935, opened 2019-05-09
bug 1513346: Migrate six.reraise to standard raise

jwhitlock:
For most uses of ``six.reraise``, the goal is to do a little work, then raise the exception being handled. For these cases, a plain ``raise`` works.

For ``transaction_context``, the ``six.reraise`` is also replaceable with ``raise``. There is a inner handler to log exceptions raised by the rollback process, with a note from before the Python 3 conversion. It's not clear to me what the note is suggesting once the code is converted to Python 3. Logging an exception has the side effect of sending an issue to Sentry, which may be the desired action (create an issue for the rollback exception, and a second issue for the original exception). If two issues are not needed, the inner handler could be eliminated, to rely on default Python 3 exception chaining. For now, I added a ``raise`` in the inner handler, so a rollback exception would be also be chained to the original exception.

For ``transaction.retry``, I changed the code so that a plain ``raise`` could be used. This changes the behaviour of the final loop, to raise the exception immediately rather than after a final connection reset, a ``time.sleep(0)``, and an optional ``quit_check()``. This required a change to the tests.

After these changes, Socorro code doesn't use ``six``. It is still required by many libraries, so it moves to constraints.


willkg:
I'm going to get to this tomorrow. Can you rebase it against master?

socorro/external/es/supersearch.py
```diff
@@ -475,8 +472,8 @@ def get(self, **kwargs):
                     # Not an ElasticsearchParseException exception
                     pass
 
-                # Re-raise the original exception with the correct traceback
-                six.reraise(exc_type, exc_value, exc_tb)
+                # Re-raise the original exception
+                raise
```
willkg: Nice!

socorro/lib/transaction.py
```diff
@@ -43,15 +40,15 @@ def transaction_context(connection_context):
...
             try:
                 conn.rollback()
             except Exception:
                 # NOTE(willkg): This exception is effectively getting
                 # swallowed. We're assuming it's a red herring for whatever
                 # threw the original exception. In Python 3, we can chain them.
                 conn.logger.exception('cannot rollback')
-            six.reraise(*excinfo)
+                raise
+            raise
```
willkg: That is super! Plus I think we can nix the comment.

socorro/lib/transaction.py
```diff
@@ -81,24 +78,21 @@ def retry(connection_context, quit_check, fun, **kwargs):
     :returns: varies
 
     """
-    last_failure = None
-
-    for backoff_time in BACKOFF_TIMES:
+    for retry_round, backoff_time in enumerate(BACKOFF_TIMES):
+        is_last_round = retry_round == (len(BACKOFF_TIMES) - 1)
```
willkg: I think I would pre-calculate `len(BACKOFF_TIMES) - 1` and stick that in a variable. Something like this:

```
max_attempts = len(BACKOFF_TIMES) - 1

for retry_round, backoff_time in enumerate(BACKOFF_TIMES):
    is_last_round = retry_round == max_attempts
    ...
```

But either way is fine and gets the same behavior.

socorro/lib/transaction.py
```diff
@@ -81,24 +78,21 @@ def retry(connection_context, quit_check, fun, **kwargs):
...
         try:
             with connection_context() as conn:
                 return fun(conn, **kwargs)
         except Exception as exc:
-            last_failure = sys.exc_info()
+            if is_last_round:
+                # If the last attempt failed, then raise the error.
+                raise
```
willkg: Nice!

socorro/processor/processor_app.py
```diff
@@ -241,7 +240,7 @@ def _transform(self, crash_id):
                 self.logger.warning('error in processing or saving crash %s', crash_id)
```
willkg: We only run the processor with a `PolyCrashStorage`, so I think `exc_value` is always a `PolyStorageWhatever` sequence. However, in the case where it isn't, this code looks like it'll raise `exc_value` twice. We can probably rework this like this:

```
# PolyCrashStorage can throw a PolyStorageError which is a sequence of exceptions
# so we should capture each of those separately.
if isinstance(exc_value, collections.Sequence):
    for exc_info_item in exc_value:
        self._capture_error(exc_info_item, crash_id)
        self.logger.warning('error in processing or saving crash %s', crash_id)
```

We don't need to turn the non-sequence into a sequence and capture it and then raise it to get captured again later.

Does that make sense? I'm not sure this is a big deal because I don't think we'll ever run the processor differently. It probably makes the code a little easier to understand. Maybe. I don't know. Wednesday.

jwhitlock: It sounds like you are saying the behaviour should be:

* If the exception is a ``PolyStorageError``:
  - Send each contained exception to Sentry via ``.capture_error``
  - Log each error at Warning level
  - (*change*) Do not re-raise
* If the exception is something else:
  - (*change*) Do not send to Sentry (yet?)
  - (*change*) Do not log
  - Let it propagate as normal

Did I understand correctly? We'll need to change a few tests to match this.

socorro/lib/transaction.py
```diff
@@ -81,24 +75,21 @@ def retry(connection_context, quit_check, fun, **kwargs):
...
-    for backoff_time in BACKOFF_TIMES:
+    max_attempts = len(BACKOFF_TIMES) - 1
+    for retry_round, backoff_time in enumerate(BACKOFF_TIMES):
         try:
             with connection_context() as conn:
                 return fun(conn, **kwargs)
         except Exception as exc:
-            last_failure = sys.exc_info()
+            if retry_round == max_attempts:
+                # If the last attempt failed, then raise the error.
```
jwhitlock: hmm I think "last attempt" could imply "previous attempt", changing to "final attempt"

socorro/processor/processor_app.py
willkg: The PolyStorageError is the special case this needs to handle because it's actually a sequence of errors each of which need to be captured and then the PolyStorageError itself should get re-raised. If it's not a PolyStorageError, then we don't need to do anything special with it--we can just reraise it.

With some tweaks, I think the code should be:

```
       except Exception as exc:
            # PolyCrashStorage can throw a PolyStorageError which is a sequence
            # of exc_info items, so we need to capture each one
            if isinstance(exc, collections.Sequence):
                for exc_info_item in exc:
                    self._capture_error(exc_info_item, crash_id)
                    self.logger.warning('error in processing or saving crash %s', crash_id)

            # Re-raise the original exception with the correct traceback
            raise
```

jwhitlock: OK, then there is no reason to have a generic exception handler, just look for ``PolyStorageError``:

```
except PolyCrashStorage as poly_error:
    # Capture and log the exceptions raised by storage backends
    for exception in poly_error:
        self._capture_error(exception, crash_id)
        self.logger.warning('error in processing or saving crash %s', crash_id)
    raise
```

and we leave other exceptions uncaptured:

* If the exception is a ``PolyStorageError``:
  - Send each contained exception to Sentry via ``.capture_error``
  - Log each error at Warning level
  - Re-raise the ``PolyStorageError``
* If the exception is something else:
  - (*change*) Do not send to Sentry (yet?)
  - (*change*) Do not log
  - Let it propagate as normal

willkg: I think that's fine, but I don't know for sure if `PolyStorageError` is the only sequence error. You'd have to skulk through the code to figure that out. It's probably the case.

jwhitlock: I took a look, this appears to be the only one:

* There are no sequence errors in the [standard library](https://docs.python.org/3.7/library/exceptions.html)
* ElasticSearch exceptions like [TransportError](https://github.com/elastic/elasticsearch-py/blob/206f5e2754b01789facc0d50dac5177444afd2f6/elasticsearch/exceptions.py#L37-L87) are simple derived classes from standard exceptions
* Configman [config exceptions](https://github.com/mozilla/configman/blob/master/configman/config_exceptions.py) and [value source exceptions](https://github.com/mozilla/configman/blob/master/configman/value_sources/source_exceptions.py) are also trivial derived classes from ``Exception``
* Our code doesn't raise any other sequence-derived exceptions, but we do have this interesting code that raises an integer:

https://github.com/mozilla-services/socorro/blob/eb37a3215d992be987c2d3308d4eb42b666efd78/socorro/lib/cache.py#L107-L108

I think it would be safe, and better practice to make this code less generic. I'll add a new commit that changes this code, and the tests around exceptions that aren't `PolyStorageError`.

willkg: Sounds good! :+1:

jwhitlock:
@willkg rebased on master. The new ``PolyStorageError`` handler is in the latest commit.

PR #4936, opened 2019-05-09
bug 1549717: fix std::out_of_range error in minidump-stackwalk

willkg:
Some crash reports cause minidump-stackwalk to kick up a `std::out_of_range` error when trying to figure out the cpu microcode by converting a string to an int. The value that's causing problems is `0xffffffff`. Gabriele says the firmware holds a 32-bit unsigned int, so stoi isn't going to work.
    
This fixes the problem by changing the code from `stoi` to `stoul` per his suggestion.

You can see the problem with a03f0937-f28f-427d-88ce-a17b80190509.

Without this fix:

```
$ make shell
app@socorro:/app$ ./scripts/run_mdsw.sh a03f0937-f28f-427d-88ce-a17b80190509
...
terminate called after throwing an instance of 'std::out_of_range'
  what():  stoi
```

With the fix, it completes as expected and produces JSON output.


scripts/run_mdsw.sh
```diff
@@ -23,24 +23,33 @@ DATADIR=./crashdata_mdsw_tmp
...
-    echo "Usage: run_mdsw.sh CRASHID"
-    exit 1
+    if [ -t 0 ]; then
+        # If stdin is a terminal, then there's no input
+        echo "Usage: run_mdsw.sh CRASHID"
+        exit 1
+    fi
+
+    # stdin is not a terminal, so pull the args from there
+    set -- ${@:-$(</dev/stdin)}
```
willkg: This follows the same code I wrote in `scripts/process_crashes.sh`. It makes `run_mdsw.sh` sooooooo much better since you can run it against a file of crashids.

minidump-stackwalk/stackwalker.cc
```diff
@@ -567,8 +567,9 @@ bool ConvertStackToJSON(const ProcessState& process_state,
       if (ContainsModule(modules_with_corrupt_symbols, frame->module)) {
         frame_data["corrupt_symbols"] = true;
       }
-      assert(!frame->module->code_file().empty());
-      frame_data["module"] = PathnameStripper::File(frame->module->code_file());
+      if (!frame->module->code_file().empty()) {
```
gabrielesvelto: In mozilla-central's minidump-analyzer I just store the empty string in this case. Either way works if downstream code assumes that a missing `module` means it's empty.

gabrielesvelto: Actually I was wrong, we don't store it at all, we just store the index so this is 100% fine.

willkg: Thank you!

willkg:
This looks good, so I'm going to land it and see how it performs over the weekend.

PR #4937, opened 2019-05-10
Add rust_begin_panic to the sentinal list.

jrmuizel:
Should help with https://crash-stats.mozilla.org/report/index/89167b96-0d9f-48e0-ae05-a85460190510#tab-details

### DIFF
--- a/socorro/signature/siglists/signature_sentinels.txt
+++ b/socorro/signature/siglists/signature_sentinels.txt
@@ -18,3 +18,4 @@ core::panicking::panic_fmt
 std::panicking::begin_panic
 std::panicking::begin_panic<T>
 std::panicking::begin_panic_fmt
+rust_begin_panic


willkg:
I think `panic_fmt` would be a better sentinel. That's in the sentinels list with the module name. The problem is that the symbol we're getting from the symbols file is missing the module name, so then it doesn't match what's in the sentinels list.

The original bug is this one:

https://bugzilla.mozilla.org/show_bug.cgi?id=1544246

which then led to this one:

https://bugzilla.mozilla.org/show_bug.cgi?id=1544416

with one of the key comments being this:

https://bugzilla.mozilla.org/show_bug.cgi?id=1544416#c3

I'm hoping they fix it in Rust so we don't have to deal with this, but we may end up having to deal with it by adding the symbols without the module names to the sentinels list. In the meantime, I'm going to close this.

PR #4938, opened 2019-05-13
bug 1549297: fix android manufacturer faceting

willkg:
In the signature report, one of the facets is on android manufacturer. Because
it wasn't being indexed as a keyword, it was getting tokenized and then faceted
on the tokens. This fixes that.

### DIFF
--- a/socorro/external/es/super_search_fields.py
+++ b/socorro/external/es/super_search_fields.py
@@ -1204,6 +1204,7 @@ FIELDS = {
         'permissions_needed': [],
         'query_type': 'enum',
         'storage_mapping': {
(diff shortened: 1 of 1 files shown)


willkg:
I've done this before with other fields. To test:

1. using master branch, process reports that show the problem--in this case, i processed reports from the signature report from the bug
2. verify pulling up the signature report page shows the problem
3. slick the dev environment
4. switch to this branch, process same reports
5. verify they're fixed in the signature report

All that works correctly.

Since this is an index change, we won't see any differences in stage or prod until it creates a new index on sunday.

self-r+

PR #4940, opened 2019-05-16
bug 1543097: Sanitize ELB headers 

jwhitlock:
The ELB adds several headers, including ``X-Forwarded-For`` and ``X-Real-IP``, that include the user's IP address. They are moved to ``sentry.interfaces.Http`` when saved in Sentry (load the JSON for a Sentry issue to see).

### DIFF
--- a/webapp-django/crashstats/tests/test_sentrylib.py
+++ b/webapp-django/crashstats/tests/test_sentrylib.py
@@ -296,7 +296,7 @@ class TestBeforeSend:
     @pytest.mark.parametrize(
         'names, url, headers, expected_headers',
         TestSanitizeHeaders.CASES,
-        ids=tuple(case[0] for case in TestSanitizePostData.CASES))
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/tests/test_sentrylib.py
```diff
@@ -163,6 +163,25 @@ class TestSanitizeHeaders:
...
+        # Deployed behind an AWS Elastic Load Balancer (ELB)
+        (
+            'X-Forwarded-For,X-Real-IP',
+            'https://example.com/siteadmin/crash-me-now/',
+            {
+                'Host': 'example.com',
+                'X-Forwarded-For': '203.0.113.19, 203.0.113.19',
+                'X-Forwarded-Port': '443',
+                'X-Forwarded-Proto': 'https',
+                'X-Real-Ip': '203.0.113.19'
```
jwhitlock: [RFC 5737](https://tools.ietf.org/html/rfc5737) reserves this IP block for documentation.

PR #4941, opened 2019-05-16
Document deploys and PII request processing

willkg:
Tried to document PII access request processing and deploying in a way that isn't going to be wrong in a month, but has enough details that it's doable by someone from reading the docs.

### DIFF
--- a/docs/howto.rst
+++ b/docs/howto.rst
@@ -81,7 +81,8 @@ For more complex crash sets, pass a search URL to generate the list::
 Processing requests for PII access
 ==================================
 
-People file bugs in Bugzilla asking to be granted access to PII.
+People file bugs in Bugzilla asking to be granted access to PII using
+`these instructions <https://crash-stats.mozilla.org/documentation/memory_dump_access/>`_.
 
 Process for handling those:
 
@@ -97,7 +98,7 @@ Process for handling those:
    agreement. Make sure to copy and paste the terms in the bug comments as well
    as the url to where it exists. Tag the reporter with a needinfo.
 
-5. Reply in the bug asking th emanager to verify the reporter requires access to
+5. Reply in the bug asking the manager to verify the reporter requires access to
    PII on Crash Stats for their job. Tag the manager with a needinfo.
 
 Then wait for those needinfos to be filled. Once that's done:
@@ -113,6 +114,8 @@ Then reply in the bug something like this::
     back in again. Let us know if you have any problems!
 
     Thank you!
+
+and mark the bug FIXED.
 
 That's it!
 
@@ -169,7 +172,7 @@ Wait for needinfos to resolve.
 10. Depending on what went out, watch Sentry, the logs, Datadog, and other
     things for problems.
 
-    Datadog and Sentry will sent notifications for problems.
+    Datadog and Sentry will send notifications for problems.
 
 11. Once everything is good to go, comment that everything is good and note
     anything that happened that we might want to know in the future. Then mark
--- a/docs/howto.rst
+++ b/docs/howto.rst
@@ -2,6 +2,8 @@
 How to
 ======
 
+.. contents::
+
 Run security checks for dependencies
 ====================================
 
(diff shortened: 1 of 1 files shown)


docs/howto.rst
```diff
@@ -75,3 +77,100 @@ For more complex crash sets, pass a search URL to generate the list::
...
+
+   I structure the comment like this::
+
+       Tagging NNN for deploy to prod
+
+       COPY-AND-PASTE-CHANGES
+
+   Then the changes show up in the releases in GitHub. For example:
+
+   https://github.com/mozilla-services/socorro/releases/tag/377
```
willkg: Tecken has a make-tag.py script that automates this. But it only does this and since we have a deploy bug with the changes, I'm not sure it helps so much that it offsets maintaining that script.

docs/howto.rst
```diff
@@ -75,3 +77,100 @@ For more complex crash sets, pass a search URL to generate the list::
...
+
+Stage autodeploys any time we land something in the master branch. You can see
+notifications on IRC or Slack. Ask for where it is these days.
+
+To deploy to prod:
+
+1. Check out the master branch and make sure you have the latest commits.
+2. Run ``scripts/deploy_bug.py`` to get the summary and description text for the
+   deploy bug.
+3. `Create a bug <https://bugzilla.mozilla.org/enter_bug.cgi?format=__standard__&product=Socorro&component=Infra>`_
```
willkg: We didn't used to do deploy bugs, but we do now. It's a really handy record later on for knowing what went out, how we verified it, any curiosities in the deploy, etc. I've looked at deploy bugs a lot after the fact.

Plus because it's a bug, we can link people to it which has been handy, too.

willkg:
@jwhitlock What do you think? Is this good enough for a first pass?

docs/howto.rst
```diff
@@ -75,3 +77,100 @@ For more complex crash sets, pass a search URL to generate the list::
...
+2. Make sure they've logged into Crash Stats prod. If they have, there will be a
+   user account with their LDAP username.
+
+3. Look them up in phonebook and find their manager.
+
+4. Reply in the bug asking the reporter to agree to the memory dump access
+   agreement. Make sure to copy and paste the terms in the bug comments as well
+   as the url to where it exists. Tag the reporter with a needinfo.
+
+5. Reply in the bug asking th emanager to verify the reporter requires access to
```
jwhitlock: ``th emanager`` → ``the manager``

docs/howto.rst
```diff
@@ -75,3 +77,100 @@ For more complex crash sets, pass a search URL to generate the list::
...
+       $ git push --tags REMOTE TAG
+
+8. I do ``git show NNN`` and copy the output into a comment in Bugzilla.
+
+9. Watch IRC/Slack for deploy notifications and make sure the deploy succeeded.
+
+10. Depending on what went out, watch Sentry, the logs, Datadog, and other
+    things for problems.
+
+    Datadog and Sentry will sent notifications for problems.
```
jwhitlock: ``sent`` → ``send``

docs/howto.rst
```diff
@@ -75,3 +77,100 @@ For more complex crash sets, pass a search URL to generate the list::
     app@socorro:app$ socorro-cmd fetch_crashids --num=all --url="https://crash-stats.mozilla.org/search/?product=Sample&date=%3E%3D2019-05-07T22%3A00%3A00.000Z&date=%3C2019-05-07T23%3A00%3A00.000Z" > crashids
     app@socorro:app$ cat crashids | socorro-cmd reprocess
 
+
+Processing requests for PII access
+==================================
+
+People file bugs in Bugzilla asking to be granted access to PII.
```
jwhitlock: I think there should be a link to [Access to memory dumps](https://crash-stats.mozilla.com/documentation/memory_dump_access/). It would make sense in this opening paragraph, since it has the instructions for the requester, including the Bugzilla link to get it to the right place. Alternatively, link below where you mention the terms to copy and paste.

docs/howto.rst
```diff
@@ -75,3 +77,100 @@ For more complex crash sets, pass a search URL to generate the list::
...
+   PII on Crash Stats for their job. Tag the manager with a needinfo.
+
+Then wait for those needinfos to be filled. Once that's done:
+
+1. Log into Crash Stats.
+2. Go into the admin.
+3. Look up the user.
+4. Add the user to the "Hackers" group.
+
+Then reply in the bug something like this::
```
jwhitlock: and mark the bug fixed?

willkg:
Good points! I'll fix those and then land this. Thank you!

willkg:
Thank you!

PR #4947, opened 2019-05-22
bug 1407212: remove unused super_search_fields code

willkg:
This removes some code from super_search_fields that we're not using and haven't for a long time. Figured I'd do a pass on this before refactoring things to reduce the amount of code changes.

To test:

1. run `make lint` and `make test`
2. do `make run` and process a bunch of crashes
3. bring up the webapp and perform searches and run topcrashers


willkg:
@jwhitlock This PR has a bunch of eyeball-numbing changes, so it's good to have a second set of eyes to catch errors. Can you skim it?

socorro/lib/search_common.py
```diff
@@ -442,8 +441,6 @@ def convert_to_type(value, data_type):
         value = datetimeutil.string_to_datetime(value)
     elif data_type == 'date' and not isinstance(value, datetime.date):
         value = datetimeutil.string_to_datetime(value).date()
-    elif data_type == 'json' and isinstance(value, str):
```
jwhitlock: 👍. Also, as tested, ``date`` is unused:

https://github.com/mozilla-services/socorro/blob/ff6dbb40e1c32d6b2954e19807445e9d99306c6d/socorro/unittest/external/es/test_super_search_fields.py#L266-L267

willkg: Funny! I want to keep `date`. Bug 1376453 covers pulling the date out of the submission timestamp allowing people to facet on that. That's a P5, but it'd help us since it would simplify some of the goofiness of `fetch_crashids` and some other things. I'll switch that to a P3.

willkg:
Thank you!

PR #4949, opened 2019-05-23
Bug 1553665 - Add libc to the signature prefixes.

emilio:
See crashes like:

 * https://bugzilla.mozilla.org/show_bug.cgi?id=1549316
 * https://bugzilla.mozilla.org/show_bug.cgi?id=1553665

I did try to test this manually but the `make shell` stuff ends up abruptly
with:

  /app/docker/socorro_entrypoint.sh: Permission denied

And I tried a bunch of things without luck.

So I hope someone can poke at it for me instead

### DIFF
--- a/socorro/signature/siglists/irrelevant_signature_re.txt
+++ b/socorro/signature/siglists/irrelevant_signature_re.txt
@@ -37,8 +37,8 @@ google_breakpad::ReceivePort::WaitForMessage
 KiFastSystemCallRet
 libandroid_runtime\.so@0x
 libbinder\.so@0x
-libc\.so@
-libc-2\.5\.so@
+libc\.so(\.\d+)?@0x
+libc-\d+\.\d+(\.\d+)?\.so@0x
 libEGL\.so@
 libdvm\.so\s*@\s*0x
 libgui\.so@0x


willkg:
I really appreciate you tackling this! I'm sorry the instructions didn't work. I haven't seen that error before, so I'm not sure what might be happening.

I did a facet on signatures that have "libc" in them and see these variations:

* libc-N.N.so@0x
* libc-N.N.N.so@0x
* libc.so.N@0x

Instead of doing a `.*` catchall which since it's a greedy operator will catch things we don't want it to catch, I think it's better to do the variations more explicitly:

```
libc\.so\.\d+@0x
libc-\d+\.\d+(\.\d+)?\.so@0x
```

That gives us:

```
app@socorro:/app$ socorro-cmd signature 0f928817-04cd-4ba8-a7cd-289c10190505
Crash id: 0f928817-04cd-4ba8-a7cd-289c10190505
Original: libc-2.24.so@0x840b1
New:      libc-2.24.so@0x840b1 | mozilla::layers::UpdateYCbCrTextureClient
Same?:    False
```

Looking at the changes to other signatures, I think libc isn't helping at all.

Instead of keeping the libc frames in the signature, let's dump them. So that's a change to the irrelevant list:

https://github.com/mozilla-services/socorro/blob/a501d1e09f3b04dcb668f08fc9e86bba3fbb4e77/socorro/signature/siglists/irrelevant_signature_re.txt#L40

And it looks like there are already some libc entries there, so let's drop those two libc entries and add the ones above.

That gives us this:

```
app@socorro:/app$ socorro-cmd signature 0f928817-04cd-4ba8-a7cd-289c10190505
Crash id: 0f928817-04cd-4ba8-a7cd-289c10190505
Original: libc-2.24.so@0x840b1
New:      mozilla::layers::UpdateYCbCrTextureClient
Same?:    False

app@socorro:/app$ socorro-cmd signature b206311b-2b61-497c-a0b7-2978b0190522
Crash id: b206311b-2b61-497c-a0b7-2978b0190522
Original: libc-2.29.9000.so@0xa6735
New:      mozilla::layers::MappedYCbCrTextureData::CopyInto
Same?:    False
```

Some of the other signatures I was looking at aren't wildly better because they're full of other library stuff.

Sorry about the bad advice before.

Can you add these two lines to the `irrelevant_signature_re.txt` list? Do they cover all the libc variants you're seeing?

willkg:
If it's easier, I'm happy to take this over, test it more thoroughly, and get it landed today. Just let me know.

emilio:
No worries, sorry, was swamped with other stuff today, will do that in a bit.

emilio:
Should be done.

Instead of `libc\.so\.\d+@0x` I used `libc\.so(\.\d+)?@0x` so that libc.so which is currently ignored keeps being ignored. I think it's the right thing to do, since I know some systems have an unversioned libc. But let me know if I goofed and I'll happily remove the parenthesis and the questionmark :)

emilio:
Ah, and thanks a lot for the advice! ^.^

socorro/signature/siglists/irrelevant_signature_re.txt
```diff
@@ -37,8 +37,8 @@ google_breakpad::ReceivePort::WaitForMessage
 KiFastSystemCallRet
 libandroid_runtime\.so@0x
 libbinder\.so@0x
-libc\.so@
-libc-2\.5\.so@
+libc\.so(\.\d+)?@0x
```
willkg: :+1:

emilio:
No worries, thanks for all the help! :)

PR #4950, opened 2019-05-23
bug 1553314: move and rename rules and pipeline

willkg:
We have a `socorro/processor/rules/` directory with the base class and one other rule. The rest of the rules were in `socorro/processor/` which makes no sense and is inconsistent and unhelpful. This moves all the rules so they're in the same directory.

We have a `processor_2015.py` module. What does that have in it? The name doesn't really tell you. It has a `Processor2015` class in it. What's that? It's not the processor app--that's in a different module. This renames `processor_2015.py` to `processor_pipeline.py` and `Processor2015` to `ProcessorPipeline`. I think that's marginally better.


willkg:
This is work I did like 4 months ago and has been sitting in a branch. Figured I'd land it now on its own.

willkg:
This is pretty benign, but I think it cleans up some confusion when reading through the code and is worth landing. self-r+

PR #4952, opened 2019-05-24
bug 1553314: remove configuration from Rules

willkg:
This adjusts the `ProcessorPipeline` to own configuration of the rules in the
pipeline. Configuration is now defined on that class and it passes in necessary
data through init args to the rules that require it.  In this way, rules are
just transformation classes and aren't configman-style classes.

This adds logging of the pipeline rules and their configuration. This will
make it a lot easier to notice and debug rules problems in the future.

Doing this also fixed configuration for the `JitCrashCategorizeRule` which was
totally busted and probably has been for some time. Now rule configuration is
namespaced which allows us to configure both the `BreakpadStackwalkerRule2015`
and `JitCrashCategorizeRule`.

This also removed all the confusing duplicated (and triplicated in some places)
configuration which was confusing and causing problems.

I also nixed the `ExternalProcessRule` and folded it into the rules that subclassed it. Those three rules were hard to follow because of all the back and forth between super/subclasses and selective overriding. It was way easier to just break them up and move the shared bits to a function.

To test:

1. run tests
2. process some crashes
   1. does the rule output on startup look sane?
   2. did BreakpadStackwalkerRule2015 run? do crashes get signatures?
   3. did the BetaVersionRule run? you need to make sure to process a Firefox beta channel crash to check this.

Note that the `JitCrashCategorizeRule` isn't quite fixed, yet. It still won't do anything because it's kicking off before signature generation, but depends on the signature to run. I want to fix that in a follow-up PR because it hasn't run in so long that I don't know what it's supposed to do or whether it works at all.

### DIFF
--- a/docker/config/local_dev.env
+++ b/docker/config/local_dev.env
@@ -70,15 +70,15 @@ resource.boto.resource_class=socorro.external.boto.connection_context.HostPortS3
 # there's only one processor node. For server environments, we probably want to store that
 # in a volume. These three vars are all affected.
 companion_process.symbol_cache_path=/tmp/symbols/cache
-processor.symbol_cache_path=/tmp/symbols/cache
-processor.symbol_tmp_path=/tmp/symbols/tmp
+processor.breakpad.symbol_cache_path=/tmp/symbols/cache
+processor.breakpad.symbol_tmp_path=/tmp/symbols/tmp
 
 # Drop kill_timeout to 30 because this is a dev environment and 5 minutes is
 # a long time
 processor.kill_timeout=30
 
 # Set symbols_urls to something helpful for local dev
-processor.symbols_urls=https://s3-us-west-2.amazonaws.com/org.mozilla.crash-stats.symbols-public/v1
+processor.breakpad.symbols_urls=https://s3-us-west-2.amazonaws.com/org.mozilla.crash-stats.symbols-public/v1
 
 # Stackwalker is in a different place in the new infra and local dev
 processor.command_pathname=/stackwalk/stackwalker
--- a/socorro/processor/processor_pipeline.py
+++ b/socorro/processor/processor_pipeline.py
@@ -58,87 +58,32 @@ from socorro.processor.rules.mozilla import (
     UserDataRule,
 )
 
-DEFAULT_RULES = [
-    # fix the raw crash removing null characters
-    DeNullRule,
-    # rules to change the internals of the raw crash
-    ProductRewrite,
-    ESRVersionRewrite,
-    PluginContentURL,
-    PluginUserComment,
-    # rules to transform a raw crash into a processed crash
-    IdentifierRule,
-    MinidumpSha256Rule,
-    BreakpadStackwalkerRule2015,
-    ProductRule,
-    UserDataRule,
-    EnvironmentRule,
-    PluginRule,
-    AddonsRule,
-    DatesAndTimesRule,
-    OutOfMemoryBinaryRule,
-    JavaProcessRule,
-    MozCrashReasonRule,
-    # post processing of the processed crash
-    CrashingThreadRule,
-    CPUInfoRule,
-    OSInfoRule,
-    BetaVersionRule,
-    ExploitablityRule,
-    FlashVersionRule,
-    OSPrettyVersionRule,
-    TopMostFilesRule,
-    ThemePrettyNameRule,
-    MemoryReportExtraction,
-    # a set of classifiers to help with jit crashes
-    JitCrashCategorizeRule,
-    # generate signature now that we've done all the processing it depends on
-    SignatureGeneratorRule,
-]
-
 
 class ProcessorPipeline(RequiredConfig):
-    """this class is a generalization of the Processor into a rule processing
-    framework. This class is suitable for use in the 'processor_app'
-    introducted in 2012."""
+    """Processor pipeline for Mozilla crash ingestion."""
 
     required_config = Namespace('transform_rules')
-    required_config.add_option(
+
+    # BreakpadStackwalkerRule2015 configuration
+    required_config.breakpad = Namespace()
+    required_config.breakpad.add_option(
         'dump_field',
         doc='the default name of a dump',
         default='upload_file_minidump',
     )
-    required_config.add_option(
+    required_config.breakpad.add_option(
         'command_pathname',
         doc='the full pathname to the external program to run (quote path with embedded spaces)',
-        # NOTE(willkg): This is the path for the RPM-based Socorro deploy. When
-        # we switch to Docker, we should change this.
-        default='/data/socorro/stackwalk/bin/stackwalker',
+        default='/stackwalk/stackwalker',
     )
-    required_config.add_option(
-        'result_key',
-        doc=(
-            'the key where the external process result should be stored '
-            'in the processed crash'
-        ),
-        default='stackwalker_result',
-    )
-    required_config.add_option(
-        'return_code_key',
-        doc=(
-            'the key where the external process return code should be stored '
-            'in the processed crash'
-        ),
-        default='stackwalker_return_code',
-    )
-    required_config.add_option(
+    required_config.breakpad.add_option(
         name='symbols_urls',
         doc='comma-delimited ordered list of urls for symbol lookup',
         default='https://localhost',
         from_string_converter=str_to_list,
         likely_to_be_changed=True
     )
-    required_config.add_option(
+    required_config.breakpad.add_option(
         'command_line',
         doc='template for the command to invoke the external program; uses Python format syntax',
         default=(
@@ -150,12 +95,12 @@ class ProcessorPipeline(RequiredConfig):
             '{dump_file_pathname}'
         )
     )
-    required_config.add_option(
+    required_config.breakpad.add_option(
         'kill_timeout',
         doc='amount of time to let mdsw run before declaring it hung',
         default=600
     )
-    required_config.add_option(
+    required_config.breakpad.add_option(
         'symbol_tmp_path',
         doc=(
             'directory to use as temp space for downloading symbols--must be '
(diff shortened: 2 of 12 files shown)


docker/config/local_dev.env
```diff
@@ -70,15 +70,15 @@ resource.boto.resource_class=socorro.external.boto.connection_context.HostPortS3
 # there's only one processor node. For server environments, we probably want to store that
 # in a volume. These three vars are all affected.
 companion_process.symbol_cache_path=/tmp/symbols/cache
-processor.symbol_cache_path=/tmp/symbols/cache
-processor.symbol_tmp_path=/tmp/symbols/tmp
+processor.breakpad.symbol_cache_path=/tmp/symbols/cache
+processor.breakpad.symbol_tmp_path=/tmp/symbols/tmp
```
willkg: These variables are now namespaced so that `BreakpadStackwalkerRule2015` configuration is separate from `JitCrashCategorizeRule` configuration.

socorro/processor/processor_pipeline.py
```diff
@@ -58,87 +58,32 @@
...
-    FlashVersionRule,
-    OSPrettyVersionRule,
-    TopMostFilesRule,
-    ThemePrettyNameRule,
-    MemoryReportExtraction,
-    # a set of classifiers to help with jit crashes
-    JitCrashCategorizeRule,
-    # generate signature now that we've done all the processing it depends on
-    SignatureGeneratorRule,
-]
```
willkg: The `ProcessorPipeline` owns configuration for and instantiating these, so I figure it should own the list as well.

socorro/processor/processor_pipeline.py
```diff
@@ -58,87 +58,32 @@
...
-        ),
-        default='stackwalker_result',
-    )
-    required_config.add_option(
-        'return_code_key',
-        doc=(
-            'the key where the external process return code should be stored '
-            'in the processed crash'
-        ),
-        default='stackwalker_return_code',
```
willkg: We don't need to set `return_key` and `return_code_key` anymore since I nixed the `ExternalCommandRule` and some of the generalizations.

socorro/processor/processor_pipeline.py
```diff
@@ -150,33 +95,61 @@ class ProcessorPipeline(RequiredConfig):
...
-    required_config.add_option(
-        'temporary_file_system_storage_path',
+    required_config.breakpad.add_option(
+        'tmp_storage_path',
         doc='a path where temporary files may be written',
         default=tempfile.gettempdir(),
     )
-    required_config.add_option(
+
+    # JitClassCategorizationRule configuration
```
willkg: Oops--that's the wrong name. Should be `JitCrashCategorizeRule`.

socorro/processor/processor_pipeline.py
```diff
@@ -150,33 +95,61 @@ class ProcessorPipeline(RequiredConfig):
...
+        doc='full pathname to external program; quote path with embedded spaces',
+        default='/stackwalk/jit-crash-categorize'
+    )
+    required_config.jit.add_option(
+        'command_line',
+        doc='template for command line; uses Python format syntax',
+        default=(
+            'timeout -s KILL {kill_timeout} {command_pathname} {dump_file_pathname}'
+        )
+    )
```
willkg: Previously, the `JitCrashCategorizeRule` was picking up the wrong configuration, but because it ran before signature generation, it never did anything so we never noticed. Yay.

This puts the configuration in its own namespace so it'll work now.

socorro/processor/processor_pipeline.py
```diff
@@ -58,87 +58,32 @@
...
-        default='stackwalker_result',
-    )
-    required_config.add_option(
-        'return_code_key',
-        doc=(
-            'the key where the external process return code should be stored '
-            'in the processed crash'
-        ),
-        default='stackwalker_return_code',
+        default='/stackwalk/stackwalker',
```
willkg: Since we're using Docker across the board now, we can set this to the path in the Docker container and remove the note I added many moons ago.

socorro/processor/processor_pipeline.py
```diff
@@ -201,11 +174,71 @@ def __init__(self, config, rules=None, quit_check_callback=None):
         else:
             self.quit_check = lambda: False
 
-        rule_set = rules or list(DEFAULT_RULES)
+        self.rules = rules or self.get_ruleset(config)
+        for rule in self.rules:
+            self.logger.info('Loaded rule: %r' % rule)
```
willkg: This prints out the rules and their configuration at startup. It's super handy.

socorro/processor/rules/base.py
```diff
@@ -4,24 +4,21 @@
...
     """Base class for transform rules
 
     Provides structure for calling rules during the processor pipeline and also
     has some useful utilities for rules.
 
     """
 
-    def __init__(self, config=None, quit_check_callback=None):
-        self.config = config
-        self.quit_check_callback = quit_check_callback
```
willkg: This would take a `quit_check_callback`, but none of the rules used it, so I'm nixing it.

socorro/processor/rules/breakpad.py
```diff
@@ -63,188 +60,73 @@ def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
         processed_crash['minidump_sha256_hash'] = raw_crash['MinidumpSha256Hash']
 
 
-class ExternalProcessRule(Rule):
```
willkg: I nixed this class and folded the bits into `BreakpadStackwalkerRule2015` and `JitCrashCategorizeRule` as follows:

1. moved `dot_save` to a module-level function
2. moved `_execute_external_process` to a module-level function that also takes as an argument `interpret_command_output`--both subclasses used this, but had different interpret functions
3. folded `_interpret_command_output` into `BreakpadStackwalkerRule2015` since that's the subclass that uses it; `JitCrashCategorizeRule` had its own
4. folded `save_results` into `JitCrashCategorizeRule` since that's the only one that used it
5. folded `action` into `JitCrashCategorizeRule`; `BreakpadStackwalkerRule2015` defined its own `action`

socorro/processor/rules/breakpad.py
```diff
@@ -254,22 +136,41 @@ def _temp_raw_crash_json_file(self, raw_crash, crash_id):
...
+        try:
+            return json.loads(data)
+        except Exception as x:
+            self.logger.error(
+                '%s non-json output: "%s"' % (command_pathname, data[:100])
+            )
+            processor_meta.processor_notes.append(
+                "%s output failed in json: %s" % (command_pathname, x)
+            )
+        return {}
```
willkg: This came from `ExternalCommandRule`.

socorro/processor/rules/breakpad.py
```diff
@@ -354,42 +255,29 @@ def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
-    required_config = Namespace()
-    required_config.add_option(
-        'command_line',
-        doc='template for the command to invoke the external program; uses Python format syntax',
-        default='timeout -s KILL 30 {command_pathname} {dump_file_pathname} '
-    )
-    required_config.add_option(
-        'command_pathname',
-        doc='the full pathname to the external program to run (quote path with embedded spaces)',
-        default='/data/socorro/stackwalk/bin/jit-crash-categorize',
```
willkg: Note that this is the old path from when we were running RPM-based Socorro in CentOS. The new path is `/stackwalk/jit-crash-categorize`.

socorro/processor/rules/breakpad.py
```diff
@@ -413,19 +301,36 @@ def predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
...
+        output, return_code = execute_external_process(
+            crash_id=crash_id,
+            command_pathname=self.command_pathname,
+            command_line=command_line,
+            processor_meta=processor_meta,
+            interpret_output=self._interpret_output
+        )
+
+        dot_save(processed_crash, 'classifications.jit.category', output)
+        dot_save(processed_crash, 'classifications.jit.category_return_code', return_code)
```
willkg: This came from `ExternalCommandRule`.

socorro/processor/processor_pipeline.py
```diff
@@ -201,11 +174,71 @@ def __init__(self, config, rules=None, quit_check_callback=None):
...
+
+        :arg config: configman DotDict config instance
 
-        self.rules = []
-        for a_rule_class in rule_set:
-            self.rules.append(a_rule_class(config))
+        :returns: pipeline of rules
+
+        """
+        return [
```
willkg: I verified we didn't drop any rules between this list and the old `DEFAULT_RULES` list.

socorro/processor/rules/breakpad.py
```diff
@@ -63,188 +60,73 @@ def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
-            stderr=subprocess.DEVNULL,
+            current_mapping = current_mapping[key_fragment]
+        except KeyError:
+            current_mapping[key_fragment] = {}
+            current_mapping = current_mapping[key_fragment]
+    current_mapping[key_parts[-1]] = value
+
+
+def execute_external_process(crash_id, command_pathname, command_line,
+                             processor_meta, interpret_output):
```
willkg: This function needs a docstring. Plus it doesn't use `crash_id`, so we can nix that argument.

socorro/processor/rules/breakpad.py
```diff
@@ -63,188 +60,73 @@ def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
-        processed_crash,
-        processor_meta
-    ):
-        # FIXME(willkg): we could replace this with glom
-        self.dot_save(processed_crash, self.config.result_key, external_command_output)
-        self.dot_save(processed_crash, self.config.return_code_key, return_code)
-
-    def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
-        crash_id = raw_crash['uuid']
-        command_parameters = dict(self.config)
```
willkg: This is how `kill_timeout` and `command_pathname` got into the parameters for interpolation.

willkg:
This was pretty intense. I want to land this now so it has time to hang out on stage for the weekend. Then on Tuesday, we can compare processed crashes between stage and prod and verify nothing changed for the worse.

self-r+

PR #4953, opened 2019-05-24
bug 1553838: fix JitCrashCategorizeRule

willkg:
The JitCrashCategorizeRule only kicks off for crashes with certain
properties including requiring certain things in the signature. If this
runs *before* signature generation, then it'll never do anything since
the predicate will never be met. This fixes that.

To test:

1. process d02ac76e-eee6-4206-8026-c08f10190523 using master branch; then check http://localhost:8000/api/UnredactedCrash/?crash_id=d02ac76e-eee6-4206-8026-c08f10190523 and look for "classifications" key--it's not there
2. process it using this branch and re-check--it is there :tada:


willkg:
self-r+

PR #4954, opened 2019-05-25
bug 1553314: default to old keys for environment variables

willkg:
This defaults to the old keys for environment variables that got changed until
we can change them in stage/prod.


willkg:
We can remove the `reference_value_from` lines after https://github.com/mozilla-services/cloudops-deployment/pull/3186 lands. With this, stage won't fail all through the long weekend and we'll have something to verify between stage and prod on Tuesday.

willkg:
self-r+

PR #4955, opened 2019-05-27
Revert "bug 1553314: default to old keys for environment variables"

willkg:
This reverts commit ea27aad459ff2f6ada808c37e06821e7e1989bc8.

We don't need this anymore since the config change landed, so I'm reverting it.


willkg:
self-r+

I'll watch stage to see if it works out and if not revert the revert.

PR #4958, opened 2019-05-30
bug 1555744: support --host and --processed in fetch_crash_data

willkg:
This makes `fetch_crash_data` more useful by allowing it to pull from any host and also adding support for pulling down processed crash data.

### DIFF
--- a/socorro/scripts/fetch_crash_data.py
+++ b/socorro/scripts/fetch_crash_data.py
@@ -27,13 +27,13 @@ identifiable information::
 
     SOCORRO_API_TOKEN=xyz
 
+Make sure the auth-token matches the host you're fetching data from.
+
 To create an API token for Socorro in -prod, visit:
 
     https://crash-stats.mozilla.org/api/tokens/
 
 """
-
-HOST = 'https://crash-stats.mozilla.org'
 
 
 class CrashDoesNotExist(Exception):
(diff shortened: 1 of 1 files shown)


socorro/scripts/fetch_crash_data.py
```diff
@@ -135,17 +137,53 @@ def fetch_crash(fetchdumps, outputdir, api_token, crash_id):
...
+        '--host', default='https://crash-stats.mozilla.org',
+        help='host to pull crash data from; this needs to match SOCORRO_API_TOKEN value'
+    )
     parser.add_argument(
         '--dumps', '--no-dumps', dest='fetchdumps', action=FlagAction, default=True,
         help='whether or not to save dumps'
     )
+    parser.add_argument(
+        '--processed', '--no-processed', dest='fetchprocessed', action=FlagAction, default=False,
+        help='whether or not to save processed crash data'
```
willkg: I have this defaulting to `False` since the dominant use for this is to pull down crash data for processing--not to compare processed crash data.

socorro/scripts/fetch_crash_data.py
```diff
@@ -135,17 +137,53 @@ def fetch_crash(fetchdumps, outputdir, api_token, crash_id):
...
 def main(argv=None):
     parser = argparse.ArgumentParser(
         formatter_class=WrappedTextHelpFormatter,
         description=DESCRIPTION.strip(),
         epilog=EPILOG.strip(),
     )
+    parser.add_argument(
+        '--host', default='https://crash-stats.mozilla.org',
+        help='host to pull crash data from; this needs to match SOCORRO_API_TOKEN value'
+    )
```
willkg: Adding `--host` brings this in line with `fetch_crashids` which already has this flag.

One thing we could do is move the `DEFAULT_HOST` up to `socorro/scripts/__init__.py` as a constant used across scripts that set the host. I don't know how much that helps, but I remember we discovered an unenthusing proliferation of host setting when we adjusted .com to .org.

willkg:
I'm going to self-r+ this. I've had a need for it a couple of times and it only affects the local dev environment and if there are issues, we can just fix them going forward.

PR #4959, opened 2019-05-31
bug 1544552: Add `BaseAllocator` to the prefix signature list

glandium:
It avoid crash reports signatures stopping at e.g. `BaseAllocator::malloc`. (e.g. bug 1544552)

### DIFF
--- a/socorro/signature/siglists/prefix_signature_re.txt
+++ b/socorro/signature/siglists/prefix_signature_re.txt
@@ -16,6 +16,7 @@ alloc::raw_vec::capacity_overflow
 _alloca_probe
 __android_log_assert
 arena_
+BaseAllocator
 BaseGetNamedObjectDirectory
 __clear_cache
 .*calloc


glandium:
Once merged, is it possible to reprocess crashes that stop on `BaseAllocator::something`?

willkg:
This change gets us this:

```
pp@socorro:/app$ socorro-cmd signature 81ca1575-d4cf-4713-8628-cdac50190523
Crash id: 81ca1575-d4cf-4713-8628-cdac50190523
Original: arena_t::MallocSmall | BaseAllocator::malloc
New:      arena_t::MallocSmall | BaseAllocator::malloc | moz_xmalloc | Gecko_CSSValue_InitSharedList
Same?:    False
```

That looks better.

willkg:
@glandium There are 408k crash reports in the last month that have 'BaseAllocator' in the signature. That's a big reprocessing job, but we could do that. I'll make a note to do this when we push the change to prod.

glandium:
FWIW, I've seen some reports with BaseAllocator in the signature that don't stop on the BaseAllocator symbol, so those wouldn't need to be reprocessed. I have no idea why that's the case, btw.

willkg:
I noticed that, too. There's `BaseAllocator::free` which is matching the `.*free` line in the prefix file. I didn't check for others.

The reprocess script I have can do "reprocess all the crash reports that match this supersearch url", but can't do "reprocess all the crash reports that match this supersearch url, but not this supersearch url". I don't think supersearch can do "I want all these, except not these", unless we do a regexp search, but Elasticsearch has a weird syntax I'm not familiar with for regexps.

I'll spend a bit of time trying to winnow down the number of crash reports to reprocess. That's a good idea. Thank you!

willkg:
I pushed this to prod just now. I'm reprocessing reports.

willkg:
I reprocessed 255k crash reports. It took about 4 hours. Let me know if you see anything that needs looking at.

glandium:
This makes things much better, thanks. There's something else that stops some signatures, still, though: https://crash-stats.mozilla.org/report/index/909df5df-3eea-454b-98b4-250200190317

willkg:
I'll look into that one today.

PR #4964, opened 2019-06-11
bug 1552898: add product_details and validation test

willkg:


product_details/README.rst
```diff
@@ -0,0 +1,48 @@
...
+This lets us use the GitHub interface for editing, reviewing, and merging
+changes.
+
+Products can have a file in this directory. The file is in JSON format.
+Here's an example:
+
+.. code-block:: json
+
+   {
+       'active_versions': ['69.0a1', '68.0b8', '67.0.2']
```
willkg: Oo--that should be ".

webapp-django/crashstats/crashstats/tests/test_utils.py
```diff
@@ -4,13 +4,53 @@
...
+    if not isinstance(data['active_versions'], list):
+        raise Exception(
+            '%s: "active_versions" value is not a list of strings' % fn_basename
+        )
+
+    for item in data['active_versions']:
+        if not isinstance(item, str):
+            raise Exception(
+                '%s: value %r is not a str' % (fn_basename, item)
+            )
```
willkg: These tests are designed to make it clearer to the contributor and the reviewer exactly what the problem is with minimal time-wastage.

willkg: If this file grows in responsibility and other things, we should build a schema and validate it against the schema. I figured I wouldn't do that, yet, but it's a good next step if we do more with these files.

willkg:
I need to land this first step so I have something to build and test the second step with.

Just to reiterate, this is a stop-gap solution to solve the immediate problem that Crash Stats has where the way it calculates active versions doesn't work if there's enough junk data.

A better long-term solution could be to fix product-details.mozilla.org to have a standard schema, support all Mozilla products, and have documentation somewhere about adding new products. Another better long-term solution is this information being surfaced in some other system for all Mozilla products.

Those are way outside what I can affect. So for now, we're going to have our own data set and interested people can update the information we've got. That solves our immediate need. We can switch to something better when it rolls around.

self-r+

PR #4965, opened 2019-06-12
bug 1552898: implement manual featured versions

willkg:
This implements manually-curated featured versions. They exist in the
`product_details/` directory. People can edit and create pull requests using the
GitHub interface. Using GitHub for process flow significantly reduces the
complexity of the implementation. This is probably a good enough first pass
and we can adjust that as needs necessitate.
    
If a file exists, Crash Stats will use that to set the featured versions which
show up on the product home page and in the "featured version" part of the
current versions drop down in the menu. If a file does not exist or can't be
retrieved, Crash Stats will calculate the featured versions based on the
existing crash report data. Featured versions are cached for an hour after
which they're figured again, so if anything goes wrong, we just have to wait
a bit.

### DIFF
--- a/webapp-django/crashstats/crashstats/jinja2/crashstats_base.html
+++ b/webapp-django/crashstats/crashstats/jinja2/crashstats_base.html
@@ -81,7 +81,7 @@
           <li class="version_select">
             <label for="product_version_select" class="visually-hidden">Select Version:</label>
             <select id="product_version_select" >
-              <optgroup label="Active">
+              <optgroup label="Featured">
                 <option value="Current Versions">Current Versions</option>
                 {% set pvs = filter_featured_versions(active_versions[product]) %}
                 {% if pvs %}
--- a/webapp-django/crashstats/crashstats/utils.py
+++ b/webapp-django/crashstats/crashstats/utils.py
@@ -7,9 +7,12 @@ import datetime
 import functools
 import isodate
 import json
+import logging
 import random
 import re
 from urllib.parse import urlencode
+
+import requests
 
 from django import http
 from django.conf import settings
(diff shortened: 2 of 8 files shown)


willkg:
Because I tweaked the `FennecAndroid.json` file, we can't see this locally--we have to land it in master first. :cry:

I'll verify it on stage before going further.

webapp-django/crashstats/crashstats/jinja2/crashstats_base.html
```diff
@@ -81,7 +81,7 @@ <h1>Product Navigation</h1>
           <li class="version_select">
             <label for="product_version_select" class="visually-hidden">Select Version:</label>
             <select id="product_version_select" >
-              <optgroup label="Active">
+              <optgroup label="Featured">
```
willkg: I spent some time fixing some cases of "active" which should have been "featured".

webapp-django/crashstats/crashstats/utils.py
```diff
@@ -417,6 +423,33 @@ def get_versions_for_product(product='Firefox', use_cache=True):
...
+    logger.debug('trying %s for featured versions', url)
+    try:
+        resp = requests.get(url)
+        if resp.status_code == 200:
+            data = resp.json()['featured_versions']
+            logger.info('using %s for featured versions', url)
+            return data
+
+    except (requests.ConnectionError, json.decoder.JSONDecodeError, KeyError):
+        return None
```
willkg: I need to add some tests for this.

willkg:
I think this seems fine. I'll check it on stage and we can roll forward with it. self-r+

PR #4966, opened 2019-06-12
bug 1552898: fix product details url

willkg:
This fixes the `PRODUCT_DETAILS_BASE_URL`. I fixed part of it in the previous PR, but missed the domain, so it's returning 404s right now.


willkg:
I tested this locally, so this should fix stage. self-r+

PR #4968, opened 2019-06-13
bug 1559153: remove depcheck cronrun cmd

willkg:
With GitHub security alerts, we don't need this anymore, so time to remove it.


willkg:
This is mostly code removal. Yay! One less thing to maintain. self-r+

PR #4969, opened 2019-06-20
Update Fennec release version to 67.0.3

rvandermeulen:


willkg:
I checked it 40 minutes later and it had updated. Yay!

PR #4972, opened 2019-06-24
Update docs

willkg:
* remove remaining depcheck bits since we don't have that anymore
* move docs on creating database migrations using Django from Mana
* add blurb on Elasticsearch migrations


willkg:
self-r+

PR #4973, opened 2019-06-26
bug 1561697: add mozilla::ipc::WriteIPDLParam to prefix list

willkg:
This adds `mozilla::ipc::WriteIPDLParam` to the prefix list so signature generation continues beyond it. By itself, it's too general a bucket.  It'd help to know what the `T` actually is, but that breaks normalization.  Continuing to the next frame should clarify the bucket.


willkg:
Using the example crash report:

```
app@socorro:/app$ socorro-cmd signature 4fb422eb-61e3-425f-917c-2bb7f0190624
Crash id: 4fb422eb-61e3-425f-917c-2bb7f0190624
Original: mozilla::ipc::WriteIPDLParam<T>
New:      mozilla::ipc::WriteIPDLParam<T> | mozilla::dom::PContentParent::SendWindowPostMessage
Same?:    False
```

willkg:
I'm pretty sure this is a step in the right direction. We can always do more later. self-r+

PR #4975, opened 2019-07-01
bug 1561628: change deploy process

willkg:
This nixes the deploy docs. I've moved them to Mana.

This removes `deploy_bug.py` in favor of `release.py` which can help create a
deploy bug and also create and push a tag. That reduces some of the steps we
have to do going forward for deploys.


docs/howto.rst
```diff
@@ -107,62 +107,3 @@ Then reply in the bug something like this::
...
-9. Watch IRC/Slack for deploy notifications and make sure the deploy succeeded.
-
-10. Depending on what went out, watch Sentry, the logs, Datadog, and other
-    things for problems.
-
-    Datadog and Sentry will send notifications for problems.
-
-11. Once everything is good to go, comment that everything is good and note
-    anything that happened that we might want to know in the future. Then mark
-    the bug FIXED.
```
willkg: I moved this to Mana. I want to standardize on having deploy docs, infra notes, and a runbook on pages in Mana for every project we maintain. That's all tied together. For example, the runbook for "what to do when a deploy goes south" should be with other infra things.

scripts/release.py
```diff
@@ -0,0 +1,258 @@
+#!/usr/bin/env python
```
willkg: This file is "copied" into this repository just like it'll be copied with the others. Copying it means there's one less thing to install to maintain these projects. Because it only uses things in Python 3.6+ stdlib, it doesn't need a virtual env or anything like that and doesn't have to be run from inside a Docker container.

scripts/release.py
```diff
@@ -0,0 +1,258 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, you can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""
+This script handles releases for this project.
+
+This has two subcommands: ``make-bug`` and ``make-tag``. See the help text for
+both.
+
+This requires Python 3 to run.
```
willkg: Oh, whoops--it requires git, too. I'll fix that upstream.

willkg:
Thank you!

PR #4976, opened 2019-07-02
Monthly update for July

jwhitlock:
Same as pyup's PR #4974, with these changes:

* Keeps moto at 1.3.8, since moto 1.3.9 pinned PyYAML to 3.13, and I want to keep PyYAML 5.1. I may try to submit an upstream PR to fix moto.
* pytest 5.0.0 treats ``DeprectatedWarning``s as errors. I think this is unintentional, and they instead want ``pytest.PytestDeprecationWarning`` as errors, as mentioned in the [changelog](https://docs.pytest.org/en/latest/changelog.html#pytest-5-0-0-2019-06-28). The ``DeprecationWarning`` is from the boto package, and doesn't appear to be in boto3, but we closed [bug 1433148](https://bugzilla.mozilla.org/show_bug.cgi?id=1433148) as WONTFIX.
* Moves pytest-django update, which includes updates for pytest 5.0.0.
* Updates constraint google-api-core from 1.8.2 to 1.13.0
* Removes constraints that no longer apply. I believe a lot of these were included in Python 2.x, and are no longer needed.


willkg:
Pretty sure we'll ditch moto when we switch to GCP, so I've been deferring updating from boto to boto3 and dealing with moto. If we don't want to wait, we could fix the s3mock bug in antenna and then switch to that. Alternatively, we could drop mocking altogether and just move data back and forth with the localstack-s3 bucket. I don't know if those are good ideas, but maybe?

For pytest, we switched on "treat all warnings as errors". We have to manually filter out the ones we want to keep as warnings.

Regarding constraints, I rebuild the constraints file every once in a while. I forget when I did it last, but it was probably in 2018. One nice thing about constraints is that stuff that is in the file, but not required by something in `default.txt` gets ignored. One not-so-nice thing is that we probably do more work updating things that aren't needed anymore. I don't have good ideas on what to do about that other than rebuild it periodically, but rebuilding it takes a while.

jwhitlock:
> Pretty sure we'll ditch moto when we switch to GCP, so I've been deferring updating from boto to boto3 and dealing with moto. If we don't want to wait, we could fix the s3mock bug in antenna and then switch to that. Alternatively, we could drop mocking altogether and just move data back and forth with the localstack-s3 bucket. I don't know if those are good ideas, but maybe?

I'm happy to stick with "this code is doomed, don't bother", since it seems like GCP is on track for end of year, and we should focus on the transition tasks. If we encounter similar issues or a compelling reason to keep AWS support, we can change our minds.

> For pytest, we switched on "treat all warnings as errors". We have to manually filter out the ones we want to keep as warnings.

I'm pretty happy with the targeted pattern that just catches importing ``imp``:

https://github.com/mozilla-services/socorro/blob/a1de252470ae07adafae5c5ee4b03303cfb02899/setup.cfg#L31-L32

My goal is that we remove this exception the same time we're removing all the "boto" references in the codebase.

> Regarding constraints, I rebuild the constraints file every once in a while. I forget when I did it last, but it was probably in 2018. One nice thing about constraints is that stuff that is in the file, but not required by something in default.txt gets ignored. One not-so-nice thing is that we probably do more work updating things that aren't needed anymore. I don't have good ideas on what to do about that other than rebuild it periodically, but rebuilding it takes a while.

It looks like the constraints file has been updated in 2018, but I don't think unneeded lines have been removed since it was created on Halloween 2017. My changes removes constraints that are no longer needed, which will make it easier to only update needed constraints in the future.

My process was to install the requirements in a fresh virtualenv, then install ``constraints.txt`` to see what new requirements were installed. I then removed these lines from ``constraints.txt`` and repeated until no extra lines remained. Finally, I ran ``make build`` to ensure that there weren't any OS-specific requirements needed by the images.

PR #4977, opened 2019-07-02
bug 1562665: Use nip.io to avoid /etc/hosts change

jwhitlock:
The [nip.io DNS service](https://nip.io), plus a [docker-compose links alias](https://docs.docker.com/compose/compose-file/compose-file-v2/#links), allows the OIDC provider to work without updating ``/etc/hosts``.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -61,6 +61,8 @@ services:
       - pubsub
       - memcached
       - oidcprovider
+    links:
+      - "oidcprovider:oidcprovider.127.0.0.1.nip.io"
     command: ["webapp", "--dev"]
     ports:
       - "8000:8000"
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -61,6 +61,8 @@ services:
       - pubsub
       - memcached
       - oidcprovider
+    links:
+      - "oidcprovider:oidcprovider.127.0.0.1.nip.io"
```
willkg: Is this the thing that lets services inside a container refer to the `oidcprovider` service as `oidcprovider.127.0.0.1.nip.io`?

jwhitlock: Yep, this is a link alias, of the form ``"container-name:alias"``. There may also be a way to do this with the [network declaration](https://docs.docker.com/compose/compose-file/compose-file-v2/#aliases) (and this may be the only way in docker-compose v4), but it is unclear if you need to manually define the whole network, so I'm assigning that research to future @jwhitlock.

Future @jwhitlock: this might work, without manual networks:

```
services:
  oidcprovider:
    networks:
      default:
        aliases:
          - oidcprovider.127.0.0.1.nip.io
```

PR #4978, opened 2019-07-03
bug 1563261: fix /monitoring/heartbeat/ Elasticsearch connection

willkg:
Previously, the webapp and the crash storage connection context built
Elasticsearch clients differently. When we updated urllib3 to 1.25.3, the one
the webapp built for `/monitoring/heartbeat/` started failing with SSL handshake
errors. This fixes that one to use `RequestsHttpConnection` which the rest
of Socorro uses.


willkg:
@jwhitlock Does this look ok to you?

willkg:
I grepped the codebase for other places where we create `Elasticsearch` clients and these are the only two.

PR #4983, opened 2019-07-15
bug 1563222: Optionally create OIDC account

jwhitlock:
Update the ``oidcprovider`` run script to create an account if the variables ``OIDC_EMAIL``, ``OIDC_USERNAME``, and ``OIDC_PASSWORD`` are defined in ``my.env``. This allows a developer to skip the "Sign Up" process when logging in locally.

I tried to write this as a script that could be run from the command line. This works well for the ``webapp`` because its database persists between restarts, but the ``oidcprovider`` container throws away the database when it is is shut down. To implement the command, I needed to 1) start the container, 2) wait for migrations to complete, and 3) use ``docker-compose exec`` to run in the running container. Integrating the user creation in the run script was easier and less error prone.

One alternate method would be to add a test user to ``oidcprovider``'s ``fixtures.json``, which would create the user at startup as well. This would mean a developer could use the documented test user to get started, but that test user wouldn't have the developer's email. "Sign up" could still be used to create a user with their email.

To test this PR out, add some variables to ``my.env``:

```
OIDC_EMAIL=admin@example.com
OIDC_USERNAME=admin
OIDC_PASSWORD=password
```

Then rebuild and restart:
```
docker-compose stop
docker-compose build --pull oidcprovider
make runservices run
```

In another terminal window, you can see the account created with ``docker-compose logs -f oidcprovider``:

```
...
oidcprovider_1    | + python manage.py loaddata fixtures.json
oidcprovider_1    | Installed 9 object(s) from 1 fixture(s)
oidcprovider_1    | + '[' -n admin ']'
oidcprovider_1    | + '[' -n password ']'
oidcprovider_1    | + '[' -n admin@example.com ']'
oidcprovider_1    | + ./manage.py shell -c 'from django.contrib.auth.models import User; user, _ = User.objects.get_or_create(email="admin@example.com", defaults={"username": "admin"}); user.set_password("password"); user.save()'
oidcprovider_1    | + python manage.py runserver 0.0.0.0:8080
oidcprovider_1    | Watching for file changes with StatReloader
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -75,6 +75,8 @@ services:
       context: .
       dockerfile: ./docker/Dockerfile.oidcprovider
     image: local/socorro_oidcprovider
+    env_file:
+      - my.env
     ports:
       - "8080:8080"
 
--- a/docker/Dockerfile.oidcprovider
+++ b/docker/Dockerfile.oidcprovider
@@ -6,3 +6,4 @@
 FROM mozillaparsys/oidc_testprovider
 
 COPY ./docker/config/oidcprovider-fixtures.json /code/fixtures.json
+COPY ./docker/run_oidcprovider.sh /code/bin/run.sh
--- a/docker/config/my.env.dist
+++ b/docker/config/my.env.dist
@@ -19,6 +19,15 @@
 producer_consumer.number_of_threads=2
 
 # ---------------------------------------------
+# webapp settings
+# ---------------------------------------------
+
+# Auto-created user when starting the development-only OpenID Connect (OIDC) provider
+# OIDC_EMAIL=admin@example.com
+# OIDC_USERNAME=admin
+# OIDC_PASSWORD=password
+
+# ---------------------------------------------
 # crash-stats.mozilla.org settings
 # ---------------------------------------------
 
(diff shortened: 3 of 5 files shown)


willkg:
I also don't like adding more to the fixtures here. It's not very flexible.

I'm not wild about setting environment variables--that also isn't very flexible for scripting that involves multiple accounts.

If the oidcprovider is saving state in a sqlite3 db, can we volume mount the db and get persistence across restarts that way?

jwhitlock:
Possibly. The file is at ``/code/db.sqlite3``, but the ``/code`` folder is not mounted. I'd need to play around a bit to see if I could make that database persist across restarts.

willkg:
We just discussed on irc that the container state persists if you do `docker-compose stop`, but not if you do `docker-compose down` which deletes containers. I think if we stick with "stop", we're good.

We can mount a specific file, but we need to make sure it exists on the host with the right permissions before launching the oidcprovider container and I think that's tricky to do at least with the various ways I launch oidcprovider.

So, I think I don't want to go the route of env vars and instead go with a django command for creating accounts that we can upstream. I'm game for fiddling with that. I think it's sort of like the `createsuperuser` command we have already.

jwhitlock:
Closing in favor of #4991

PR #4985, opened 2019-07-19
bug 1567528: Filter MozCrashReason with eval msg

jwhitlock:
Filter a ``MozCrashReason`` prefix that may contain sensitive data in the full message.


jwhitlock:
Tested on relevant crash report in dev. There are crashes in stage to test on as well.

lonnen:
Yep, r+ waiting on circle

PR #4986, opened 2019-07-22
bug 1564551: Expand and test is_valid_model_class

jwhitlock:
If a non-class is passed as the model, return ``False`` instead of raising a ``TypeError``. This can happen if a name is passed that is a valid non-class attribute of one of the model modules, such as ``/api/time``.

This regression was introduced in PR #4982, which hasn't yet been deployed to production, so I continued to use [bug 1564551](https://bugzilla.mozilla.org/show_bug.cgi?id=1564551). I can open a new bug if we want to track regressions in new bugs.

To avoid future regressions, add tests to confirm behavior for a variety of inputs.

### DIFF
--- a/webapp-django/crashstats/api/tests/test_views.py
+++ b/webapp-django/crashstats/api/tests/test_views.py
@@ -19,17 +19,27 @@ from moto import mock_s3_deprecated
 import pyquery
 import pytest
 
-from crashstats.api.views import api_models_and_names, MultipleStringField, TYPE_MAP
+from crashstats.api.views import (
+    api_models_and_names,
+    is_valid_model_class,
+    MultipleStringField,
+    TYPE_MAP,
+)
 from crashstats.crashstats.models import (
     BugAssociation,
     NoOpMiddleware,
     ProcessedCrash,
     Reprocessing,
     RawCrash,
+    SocorroMiddleware,
     UnredactedCrash,
 )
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -50,6 +60,21 @@ def test_dedent_left(self):
...
+class TestIsValidModelClass():
+    """Test that is_valid_model_class validates API models."""
+
+    @pytest.mark.parametrize("model", (SuperSearch, SuperSearchUnredacted))
+    def test_valid(self, model):
+        assert is_valid_model_class(model)
+
+    @pytest.mark.parametrize(
+        "not_model",
+        ("SuperSearch", int, contextlib, SocorroMiddleware, ESSocorroMiddleware)
```
jwhitlock: Non-valid arguments:

* ``"SuperSearch"`` - a string, not a class
* ``int`` - a class, but not a subclass of ``SocorroMiddleware``
* ``contextlib`` - a module, not a class. This is how the bug would manifest in production.
* ``SocorroMiddleware`` - A disallowed subclass of ``SocorroMiddleware``
* ``ESSocorroMiddleware`` - A disallowed subclass of ``SocorroMiddleware``

jwhitlock:
Future non-urgent cleanup - calculate the list of valid models and their names once at app startup, and use in ``model_wrapper``, ``documentation``, and tests.

PR #4990, opened 2019-07-24
bug 1563230: wrap cronrun in a timeout

willkg:
This prevents cronrun from hanging for more than an hour.

### DIFF
--- a/socorro/processor/processor_pipeline.py
+++ b/socorro/processor/processor_pipeline.py
@@ -81,7 +81,7 @@ class ProcessorPipeline(RequiredConfig):
         "command_line",
         doc="template for the command to invoke the external program; uses Python format syntax",
         default=(
-            "timeout -s KILL {kill_timeout} {command_pathname} "
+            "timeout --signal KILL {kill_timeout} {command_pathname} "
             "--raw-json {raw_crash_pathname} "
             "{symbols_urls} "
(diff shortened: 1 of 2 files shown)


socorro/processor/processor_pipeline.py
```diff
@@ -81,7 +81,7 @@ class ProcessorPipeline(RequiredConfig):
         "command_line",
         doc="template for the command to invoke the external program; uses Python format syntax",
         default=(
-            "timeout -s KILL {kill_timeout} {command_pathname} "
+            "timeout --signal KILL {kill_timeout} {command_pathname} "
```
willkg: While I was fixing this bug, I figured I'd expand this argument to the long name so it's clearer.

socorro/processor/processor_pipeline.py
```diff
@@ -91,7 +91,7 @@ class ProcessorPipeline(RequiredConfig):
     )
     required_config.breakpad.add_option(
         "kill_timeout",
-        doc="amount of time to let mdsw run before declaring it hung",
+        doc="amount of time in seconds to let mdsw run before declaring it hung",
```
willkg: While I was fixing this bug, I figured I'd clarify that the value is in seconds here so future me doesn't have to look it up again.

willkg:
self-r+

PR #4991, opened 2019-07-24
bug 1563222: add createuser script to oidcprovider container

willkg:
This lets you do the following:

```
docker-compose up -d oidcprovider
docker-compose exec oidcprovider /code/manage.py createuser USER PASSWORD EMAIL
```

Then you'll have a user account in the running oidcprovider container. Couple this with `makesuperuser` and that's all you need to script account creation for local dev environment purposes.

This fixes my needs without a lot of mumbo-jumbo. The account that gets created exists only in the running container--it's not persisted to the host or across other oidcprovider containers.

To test:

1. `docker-compose stop oidcprovider`
2. `docker-compose build --no-cache oidcprovider`
3. `docker-compose up -d oidcprovider`
4. `docker-compose exec oidcprovider /code/manage.py USER PASSWORD EMAIL`

This also handles the case where the account already exists--it lets you know and doesn't error out.

### DIFF
--- a/docker/Dockerfile.oidcprovider
+++ b/docker/Dockerfile.oidcprovider
@@ -6,3 +6,7 @@
 FROM mozillaparsys/oidc_testprovider
 
 COPY ./docker/config/oidcprovider-fixtures.json /code/fixtures.json
+RUN mkdir --parents /code/oidcprovider/management/commands/
+RUN touch /code/oidcprovider/management/__init__.py
+RUN touch /code/oidcprovider/management/commands/__init__.py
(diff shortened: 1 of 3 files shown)


willkg:
@jwhitlock What do you think? This solves my immediate need. We can also make it persist across containers, but that's a little trickier since we need to mount the `db.sqlite` file into the container which requires us to create it on the host first so we have something to mount. We can do that in the `Makefile`, but we need to make sure we're always starting the oidcprovider using the `Makefile`. It gets tricky to make it "just work" all the time.

PR #4992, opened 2019-07-25
bug 1563222: Drop confirmation of OIDC provider

jwhitlock:
Use better names for the OIDC test clients, and skip confirmation entirely, to further streamline local development.

I piggybacked on [bug 1563222](https://bugzilla.mozilla.org/show_bug.cgi?id=1563222), because I noticed the weird test client name on the confirmation screen while reviewing PR #4991. If you'd like to track under a new bug, I can file it.


jwhitlock:
Rebased. Since CI liked it before the rebase, and this change doesn't impact tests, merging before CI is done.

PR #4993, opened 2019-07-25
Fix fetch_crash_data to optionally not download raw crash data

willkg:
This fixes `fecth_crash_data` to also have a `--raw`/`--no-raw` argument which
makes it easier to use it to *just* download processed crash data.

### DIFF
--- a/socorro/scripts/fetch_crash_data.py
+++ b/socorro/scripts/fetch_crash_data.py
@@ -47,7 +47,9 @@ def create_dir_if_needed(d):
         os.makedirs(d)
 
 
-def fetch_crash(host, fetchdumps, fetchprocessed, outputdir, api_token, crash_id):
+def fetch_crash(
+    host, fetchraw, fetchdumps, fetchprocessed, outputdir, api_token, crash_id
+):
     """Fetch crash data and save to correct place on the file system
 
     http://antenna.readthedocs.io/en/latest/architecture.html#aws-s3-file-hierarchy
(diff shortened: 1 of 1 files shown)


socorro/scripts/fetch_crash_data.py
```diff
@@ -60,31 +62,32 @@ def fetch_crash(host, fetchdumps, fetchprocessed, outputdir, api_token, crash_id
...
-    with open(fn, "w") as fp:
-        json.dump(raw_crash, fp, cls=JsonDTEncoder, indent=2, sort_keys=True)
+        # Save raw crash to file system
+        raw_crash = resp.json()
+        fn = os.path.join(
+            outputdir, "v2", "raw_crash", crash_id[0:3], "20" + crash_id[-6:], crash_id
+        )
+        create_dir_if_needed(os.path.dirname(fn))
+        with open(fn, "w") as fp:
+            json.dump(raw_crash, fp, cls=JsonDTEncoder, indent=2, sort_keys=True)
```
willkg: This whole section was merely indented a level and put under an "if" block.

socorro/scripts/fetch_crash_data.py
```diff
@@ -196,6 +207,10 @@ def main(argv=None):
     else:
         args = parser.parse_args(argv)
 
+    if args.fetchdumps and not args.fetchraw:
+        print("You cannot fetch dumps without also fetching the raw crash. Exiting.")
+        return 1
+
```
willkg: The raw crash has a field that tells us what dumps came with the crash report. So if the user wants to download dumps, but not the raw crash, we can't really do that since without the raw crash, we have no idea what dumps exist.

willkg:
self-r+

PR #4994, opened 2019-07-25
Update release.py

willkg:
This updates `release.py` to the latest. It looks a little goofy. We used black to reformat the python code, so we made the changes in this repo and then copied it back to the socorro-release repo. This mostly picks up the SHA change from the socorro-release repo.


willkg:
self-r+

PR #4995, opened 2019-07-26
bug 1503344: start putting data in java_stack_trace_raw

willkg:
We have this problem where I named the field with `_full` at the end not knowing about the whole `_full` thing our Elasticsearch crashstorage/supersearch thing does. We periodically get errors from that poor naming choice.

This does the first step in renaming the field to `_raw`--it additionally puts the data in that field and deprecates the original one. It also adds a note to remove the old field in January 2020.

If we had a way of doing migrations, that'd be better. But we don't, so we're left with this. :(


socorro/external/es/super_search_fields.py
```diff
@@ -2248,9 +2248,10 @@ def get(self):
             "type": "string",
         },
     },
+    # FIXME(willkg): Remove java_stack_trace_full after January 2020.
     "java_stack_trace_full": {
         "data_validation_type": "str",
-        "description": "Raw JavaStackTrace value.",
+        "description": "(Deprecated) Raw JavaStackTrace value.",
```
willkg: Both fields are going to show up in supersearch for a while. I figured I'd do that so people can switch to the new one over time and search both of them.

socorro/unittest/processor/rules/test_mozilla.py
```diff
@@ -597,7 +597,7 @@ def test_everything_we_hoped_for(self):
         rule.act(raw_crash, raw_dumps, processed_crash, processor_meta)
 
         # The entire JavaStackTrace blob
-        assert processed_crash["java_stack_trace_full"] == raw_crash["JavaStackTrace"]
+        assert processed_crash["java_stack_trace_raw"] == raw_crash["JavaStackTrace"]
```
willkg: I changed the tests to only test the new one. That makes for slightly less work in January 2020 when cleaning up.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -338,6 +338,7 @@ <h2>{{ report.product }} {{ report.version }} Crash Report [@ {{ report.signatur
                   <td>{{ report.address }}</td>
                 </tr>
                 {% if request.user.has_perm('crashstats.view_pii') or your_crash %}
+                  {# FIXME(willkg): switch this to java_stack_trace_raw after January 2020 #}
```
willkg: We have to show the old one here because that's the one that has data for the last 6 months. When we delete the old one, we can switch to the new one here.

willkg:
self-r+

PR #4996, opened 2019-07-26
bug 1559223, 1542964: add modules_in_stack to processed crash

willkg:
This adds a new "modules_in_stack" field which is a semicolon-delimited
set of "module/debugid" strings for modules that are in the stack of the
crashing thread.

The value is indexed using the semicolon_analyzer, so each string ends
up as a single term. This allows us to use supersearch "startswith" for
finding crash reports with a specified module in the stack (e.g.
"libfenix.so") and "has term" for finding crash reports with a specified
"module/debugid" in the crashing thread of the stack. The latter helps
us reprocess crash reports for which we just uploaded symbols.

### DIFF
--- a/socorro/external/es/super_search_fields.py
+++ b/socorro/external/es/super_search_fields.py
@@ -2730,6 +2730,20 @@ FIELDS = {
         "query_type": "string",
         "storage_mapping": {"type": "string"},
     },
+    "modules_in_stack": {
+        "data_validation_type": "str",
+        "description": "Set of module/debugid strings that show up in stack of the crashing thread.",
+        "form_field_choices": [],
+        "has_full_version": False,
+        "in_database_name": "modules_in_stack",
+        "is_exposed": True,
+        "is_returned": True,
+        "name": "modules_in_stack",
+        "namespace": "processed_crash",
+        "permissions_needed": [],
+        "query_type": "string",
+        "storage_mapping": {"analyzer": "semicolon_keywords", "type": "string"},
+    },
     "moz_crash_reason": {
         "data_validation_type": "str",
         "description": (
--- a/socorro/processor/rules/mozilla.py
+++ b/socorro/processor/rules/mozilla.py
@@ -527,6 +527,59 @@ class TopMostFilesRule(Rule):
                 return
 
 
+class ModulesInStackRule(Rule):
+    """
+    Adds value with semi-colon separated set of "module/debugid" strings for
+    all the modules that show up in the stack of the crashing thread.
+    """
+
+    # Filenames should contain A-Za-z0-9_. and that's it.
+    BAD_FILENAME_CHARACTERS = re.compile(r"[^a-z0-9_\.]", re.IGNORECASE)
+
+    # Debug ids are hex strings
+    BAD_DEBUGID_CHARACTERS = re.compile(r"[^a-f0-9]", re.IGNORECASE)
+
+    def format_module(self, item):
+        filename = item.get("filename", "")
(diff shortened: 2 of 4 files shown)


willkg:
To test:

1. slick your local dev environment so you pick up the new Elasticsearch index mapping
2. process some crashes
3. do a supersearch for "modules_in_stack" and "exists"--it should find crashes
4. do a supersearch for "modules_in_stack" and "startswith" and "libxul.dll"--it should find crashes (it's possible it might not if you didn't process any Windows crash reports)
5. look at the unredacted data for a crash report (crash report -> "raw dump" tab -> unredacted link is at bottom of page) and make sure "modules_in_stack" is a ";" delimited string of "module/debugid" strings

That's pretty much it.

One thing I wasn't sure about was whether to add "modules_in_stack" to the details tab in the report view. Maybe it's helpful? Probably not?

socorro/external/es/super_search_fields.py
```diff
@@ -2730,6 +2730,20 @@ def get(self):
...
+        "form_field_choices": [],
+        "has_full_version": False,
+        "in_database_name": "modules_in_stack",
+        "is_exposed": True,
+        "is_returned": True,
+        "name": "modules_in_stack",
+        "namespace": "processed_crash",
+        "permissions_needed": [],
+        "query_type": "string",
+        "storage_mapping": {"analyzer": "semicolon_keywords", "type": "string"},
```
willkg: Using this analyzer splits the string into a series of terms, so it works nicely in supersearch.

socorro/processor/rules/mozilla.py
```diff
@@ -527,6 +527,59 @@ def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
+        except (KeyError, IndexError):
+            # If those things aren't in the raw crash, then this rule doesn't
+            # have anything to do.
+            return
+
+        module_to_module_info = {
+            item["filename"]: self.format_module(item)
+            for item in json_dump.get("modules", [])
+            if "filename" in item
+        }
```
willkg: This builds a map of "module" -> "module/debug_id" for all the modules that were in memory when the crash happened.

socorro/processor/rules/mozilla.py
```diff
@@ -527,6 +527,59 @@ def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
+
+        modules_in_stack = {
+            module_to_module_info.get(frame.get("module")) for frame in stack
+        }
```
willkg: This builds a set of all the "module/debug_id" strings for the frames in the stack of the crashing thread. Because we're using `.get()`, it's possible that one of the items in this set is `None`.

socorro/processor/rules/mozilla.py
```diff
@@ -527,6 +527,59 @@ def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
+        modules_in_stack = sorted(module for module in modules_in_stack if module)
```
willkg: This gets rid of `None` if it's in the set. It also sorts it because stable results are easier to test.

socorro/processor/rules/mozilla.py
```diff
@@ -527,6 +527,59 @@ def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
+        if modules_in_stack:
+            processed_crash["modules_in_stack"] = ";".join(modules_in_stack)
```
willkg: This joins the "module/debug_id" strings with ";" giving us a final value and putting that in the processed crash.

willkg:
There really isn't much to this. Seems good. self-r+

PR #4997, opened 2019-07-29
bug 1569582: add Bug_1541161 annotation to super search

willkg:


socorro/external/es/super_search_fields.py
```diff
@@ -393,6 +393,23 @@ def get(self):
...
+            "NewChannelFromURIWithProxyFlagsInternal()."
+        ),
+        "form_field_choices": None,
+        "has_full_version": False,
+        "in_database_name": "Bug_1541161",
+        "is_exposed": True,
+        "is_returned": True,
+        "name": "bug_1541161",
+        "namespace": "raw_crash",
+        "permissions_needed": ["crashstats.view_pii"],
```
willkg: This has stack traces, but it's not clear whether that's totally devoid of pii to me, so I'm going to make this require pii for now.

willkg:
I tested this with the example crash report in my local dev environment. self-r+

PR #4999, opened 2019-08-05
bug 1571432: downgrade sentry-sdk to 0.9.5

willkg:
We're having memory usage spikes in the processor after upgrading a
series of dependencies last week. This downgrades sentry-sdk to see if
that fixes it.


willkg:
We can see the memory usage behavior change in stage, but it looks like it takes about 6 hours to know for sure one way or the other. I'll land this now and then we'll see what happens.

willkg:
self-r+

PR #5002, opened 2019-08-16
Remove vestiges of Python 3 migration

willkg:
When we were doing the Python3 migration, we added a USEPYTHON
environment variable to determine which Python binary to use. We don't
need that anymore--all the related machinery is gone.


willkg:
self-r+

PR #5017, opened 2019-10-07
bug 1586675: add strcmp variations

willkg:


socorro/signature/rules.py
```diff
@@ -205,7 +205,10 @@ def normalize_frame(
         module_offset=None,
         offset=None,
         normalized=None,
-        **kwargs,  # eat any extra kwargs passed in
+        # eat any extra kwargs passed in
+        # fmt: off
+        **kwargs
+        # fmt: on
```
willkg: siggen supports Python 3.5, which doesn't support trailing commas in argument lists. If we take out the `fmt` markers, then black adds a trailing comma after the `**kwargs`.

https://black.readthedocs.io/en/stable/the_black_code_style.html#the-black-code-style

willkg:
self-r+

PR #5022, opened 2019-10-22
Update Fennec versions for new releases going out today

rvandermeulen:


willkg:
I know doing this by hand is not great and I really appreciate you doing it. Fenix may start putting out a product details file on archive.mozilla.org. If that happens, I think we should revisit this and maybe go back to using product details. I'll think about that more in 2020 probably.

rvandermeulen:
I agree 100% that we should ultimately make publishing version information
to product-details a requirement for appearing in Socorro.

PR #5023, opened 2019-10-23
bug 1590741: update oidcprovider to v0.9.3

willkg:
This updates the oidcprovider image we're using to the v0.9.3 tag. That
version has the createuser stuff we had, so we no longer need that code.


docker/Dockerfile.oidcprovider
```diff
@@ -1,12 +1,4 @@
-# Derived from the "testprovider" container from
-# https://github.com/mozilla-parsys/docker-test-mozilla-django-oidc.
-# Only the redirect_urls specified in "fixtures.json" are being
-# modified to fit the needs of the docker setup.
-
-FROM mozillaparsys/oidc_testprovider
+FROM mozilla/oidc-testprovider:oidc_testprovider-v0.9.3
```
willkg: Existing tags are here:

https://hub.docker.com/r/mozilla/oidc-testprovider/tags?page=1&name=testprovider

We want a specific tag and not `latest` because it's nice to intentionally upgrade rather than accidentally upgrade.

docker/Dockerfile.oidcprovider
```diff
@@ -1,12 +1,4 @@
-# Derived from the "testprovider" container from
-# https://github.com/mozilla-parsys/docker-test-mozilla-django-oidc.
-# Only the redirect_urls specified in "fixtures.json" are being
-# modified to fit the needs of the docker setup.
```
willkg: I didn't understand this comment, so I winnowed it down.

willkg:
This only affects the local dev environment and it worked for me! self-r+

PR #5028, opened 2019-11-06
bug 1594468: move libc* lines from irrelevant list to prefix list

willkg:
We've done a lot of symbol uploading for libc binaries for Linux
distributions. Instead of skipping these frames, we want to include them
in the signature and move past them.

```
app@socorro:/app$ socorro-cmd signature b206311b-2b61-497c-a0b7-2978b0190522
Crash id: b206311b-2b61-497c-a0b7-2978b0190522
Original: mozilla::layers::MappedYCbCrTextureData::CopyInto
New:      libc-2.29.9000.so@0xa6735 | mozilla::layers::MappedYCbCrTextureData::CopyInto
Same?:    False
```


willkg:
self-r+

PR #5032, opened 2019-11-12
bug 1433148: redo retry

willkg:
This replaces `socorro.lib.transaction.retry` with `socorro.lib.util.retry` which is what we use in Antenna. This also removes some vestigial bits and intermediary classes that don't need to be there anymore. 

Why do this? The ultimate goal here is to update the Socorro S3 code to use boto3. Rather than just rewrite the existing code to use boto3, I'm replacing the existing code with the code we're using in Antenna. That code handles AWS credentials better, is battle-tested, is simpler, and uses boto3. Seems like a good idea to unify the S3 code in both code bases. Replacing the `retry` code is a small, self-contained step.


willkg:
I need to self-review this. I'll do it tomorrow.

.editorconfig
```diff
@@ -21,7 +21,7 @@ indent_size = 2
 indent_size = 2
 
 [*.py]
-max_line_length = 100
+max_line_length = 88
```
willkg: This fixes the line length in my editor to match black and flake8.

socorro/external/boto/connection_context.py
```diff
@@ -33,7 +33,7 @@ def default(self, obj):
         raise NotImplementedError("Don't know about {0!r}".format(obj))
 
 
-class ConnectionContextBase(RequiredConfig):
+class S3ConnectionContext(RequiredConfig):
```
willkg: `S3ConnectionContext` was an intermediary class and the only subclass of `ConnectionContextBase`, so I folded it up into `ConnectionContextBase`.

socorro/external/boto/connection_context.py
```diff
@@ -228,39 +241,6 @@ def _convert_string_to_list(self, a_string):
     def __call__(self):
         yield self
 
-    def force_reconnect(self):
-        pass
```
willkg: `force_reconnect` wasn't used by anything.

socorro/external/boto/crashstorage.py
```diff
@@ -15,17 +16,25 @@
...
-class BotoCrashStorage(CrashStorageBase):
+LOGGER = logging.getLogger(__name__)
+
+
+def wait_time_generator():
+    for i in [1, 1, 1, 1, 1]:
+        yield i
+
+
+class BotoS3CrashStorage(CrashStorageBase):
```
willkg: `BotoS3CrashStorage` was the only subclass of `BotoCrashStorage`, so I merged the two.

socorro/external/boto/crashstorage.py
```diff
@@ -54,20 +63,23 @@ def __init__(self, config, namespace="", quit_check_callback=None):
         )
         self.connection_source = config.resource_class(config)
 
+        # Function for calling a function retriably
+        self.retryable = retry(
```
willkg: This creates a decorator that wraps functions in a retryable. We use it later in this file to wrap all the `do_xyz` function calls.

socorro/external/boto/crashstorage.py
```diff
@@ -54,20 +63,23 @@ def __init__(self, config, namespace="", quit_check_callback=None):
...
+            retryable_exceptions=self.connection_source.RETRYABLE_EXCEPTIONS,
+            wait_time_generator=wait_time_generator,
+            module_logger=LOGGER,
+        )
+
     @staticmethod
-    def do_save_raw_crash(boto_connection, raw_crash, dumps, crash_id):
+    def do_save_raw_crash(conn, raw_crash, dumps, crash_id):
```
willkg: This is cosmetic. `boto_connection` is super long and not overly more informative than just `conn`. Swapping for the shorter name reduces some lines and simplifies things.

socorro/external/boto/crashstorage.py
```diff
@@ -76,102 +88,87 @@ def do_save_raw_crash(boto_connection, raw_crash, dumps, crash_id):
...
-            self.quit_check,
-            self.do_save_raw_crash,
-            raw_crash=raw_crash,
-            dumps=dumps,
-            crash_id=crash_id,
-        )
+        """Save raw_crash, dump_names, and dumps."""
+        retryable_save = self.retryable(self.do_save_raw_crash)
+        with self.connection_source() as conn:
+            retryable_save(conn, raw_crash=raw_crash, dumps=dumps, crash_id=crash_id)
```
willkg: This is functionally similar to the old code except it doesn't use the `quit_check_callback` stuff. Because the wait time generator is `1, 1, 1, 1, 1`, it waits at most 5 seconds. Given that, I decided it's ok to drop the quit check callback stuff for now.

socorro/external/boto/crashstorage.py
```diff
@@ -221,15 +204,6 @@ class TelemetryBotoS3CrashStorage(BotoS3CrashStorage):
...
 
-    required_config = Namespace()
-    required_config.add_option(
-        "resource_class",
-        default="socorro.external.boto.connection_context.RegionalS3ConnectionContext",
-        doc="fully qualified dotted Python classname to handle Boto connections",
-        from_string_converter=class_converter,
-        reference_value_from="resource.boto",
-    )
-
```
willkg: This is redundant since the superclass already sets the default for `resource_class` as `RegionalS3ConnectionContext`.

socorro/lib/util.py
```diff
@@ -25,3 +29,138 @@ def _dictify(thing):
         return thing
 
     return _dictify(sdotdict)
+
+
+class MaxAttemptsError(Exception):
```
willkg: This is copied from Antenna verbatim.

socorro/unittest/external/boto/test_crash_data.py
```diff
@@ -35,14 +34,13 @@ def test_get_processed(self, boto_helper):
 
     @mock_s3_deprecated
     def test_get_processed_not_found(self, boto_helper):
-        with mock.patch("socorro.lib.transaction.BACKOFF_TIMES", [0]):
```
willkg: This made it so that the test would pass faster when raising errors. We don't have `BACKOFF_TIMES` anymore, so I removed these mocks.

socorro/unittest/external/boto/test_crash_data.py
```diff
@@ -35,14 +34,13 @@ def test_get_processed(self, boto_helper):
...
-            boto_helper.get_or_create_bucket("crashstats")
+        boto_helper.get_or_create_bucket("crashstats")
 
-            boto_s3_store = self.get_s3_store()
-            with pytest.raises(CrashIDNotFound):
```
willkg: I need to check if `CrashIDNotFound` will trigger retrying. Seems like this test expects it to.

socorro/unittest/lib/test_util.py
```diff
@@ -5,8 +5,9 @@
 import copy
 
 from configman.dotdict import DotDict
+import pytest
 
-from socorro.lib.util import dotdict_to_dict
+from socorro.lib.util import dotdict_to_dict, retry, MaxAttemptsError
```
willkg: These bits are copied verbatim from Antenna.

willkg:
I need to check if `CrashIDNotFound` will trigger retrying. Seems like this test expects it to. I think `crash_data` classes need it to because of race conditions between submission and processing. Marking this as WIP until I do that.

socorro/unittest/external/boto/test_crash_data.py
willkg: Neither `CrashIDNotFound` nor the two response errors `KeyNotFound` and `boto.exceptions.StorageError` trigger retrying. This test suggests it should have, but I think either the code changed at some point between when this test was written and now or this test was written curiously.

So... to summarize, retrying isn't triggered and I don't think it should at this time, so we're all good here.

willkg:
I've worked through the outstanding issue and this is ready for review.

After this lands, I'd like it to sit on stage for a few days to see how it behaves.

@jwhitlock Whatever reviewing you want to do, you're good to go. Just let me know when you're done with it.

socorro/external/boto/crashstorage.py
jwhitlock: I agree, losing interactive abilities for 5 seconds is probably not a big deal. I think it would be possible to port ``quit_check_callback`` by implementing a  ``sleep_function`` for ``retry``.

willkg:
I like that idea of doing the quit check in the sleep function. That's nice!

willkg:
Thank you!

PR #5034, opened 2019-11-14
Update contributing docs

willkg:
This fixes some things in the contributing docs:

1. we don't have an auto-closer anymore
2. we use black for Python code formatting
3. might as well include component=General in the link rather than talk about it


willkg:
This is a docs change. self-r+

PR #5035, opened 2019-11-14
Move "pip check" from lint step to build step

willkg:
Moving "pip check" to the build step makes it fail faster if there are
issues.


willkg:
Thank you!

PR #5037, opened 2019-11-18
bug 1597334: enable docker-layer-caching

willkg:
This enables docker-layer-caching to reduce build times in Circle CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,8 @@ jobs:
       - store_artifacts:
           path: /socorro/version.json
 
-      - setup_remote_docker
+      - setup_remote_docker:
+          docker_layer_caching: true
 
       - run:
           name: Get info
--- a/README.rst
+++ b/README.rst
@@ -10,11 +10,11 @@ displaying crash reports from clients using the `Breakpad libraries
 Support
 =======
 
-This is a very Mozilla-specific product. We do not currently have the capacity
-to support external users.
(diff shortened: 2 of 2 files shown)


willkg:
First CI run passed in 13:32:

https://circleci.com/gh/mozilla-services/socorro/4975

The `Setup a remote Docker engine` step doesn't indicate that docker-layer-caching is enabled. I'm going to tweak the README and then push that as a commit and see if that uses the layer cache.

willkg:
Second CI run used layer caches and passed in 4:41:

https://circleci.com/gh/mozilla-services/socorro/4976

The "build docker images" step took 32s.

I like this.

@sciurus, @jwhitlock: Does this look ok to you two?

jwhitlock:
I [mentioned on the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1597334#c3) that we may want to remove the apt caches, which may make it more likely that there will be no file system changes and Docker will use the cache for later layers. This test suggests it doesn't matter. It could also be tricky to get that right with the `vim` install in a later step.

willkg:
The `Dockerfile` was one of the first ones I wrote and it's likely junky in parts. If you wanted to spend some time streamlining it, that'd be super. Less complexity, less back-and-forth between bash, smaller image size, fewer steps, better cacheability--all those things are valuable. We can then build a better set of conventions and use that across all our projects.

willkg:
Thank you!

PR #5038, opened 2019-11-18
bug 1597334: Adjust Dockerfile for size

jwhitlock:
Make some changes to the Docker build size:

* Following [Docker best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run), remove the apt cache at the end of package installs. This reduced the final image size almost 1%. It does require that `apt-get update` is run again when installing dev tools in the socorro_breakpad build, however, so it can potentially slow down builds without caches.

And, some more suggested by @willkg:

* Move breakpad package installs to ``socorro_breakpad``, to reduce final image size.
* Fix quoting in ``set_up_ubuntu.sh``.
* Combine some commands to reduce layers.
* Run ``collectstatic`` as app user, rather than ``chown`` afterwards, to eliminate a duplicate layer.

### DIFF
--- a/docker/set_up_ubuntu.sh
+++ b/docker/set_up_ubuntu.sh
@@ -29,20 +29,12 @@ PACKAGES_TO_INSTALL=(
     # For sentry-cli
     gawk
 
-    # For building breakpad
-    pkg-config
-    libcurl3
-    libcurl3-gnutls
-    libcurl4-gnutls-dev
-    wget
-    rsync
-
     # For nodejs and npm
     curl
 )
 
 # Install Ubuntu packages
-apt-get install -y ${PACKAGES_TO_INSTALL[@]}
+apt-get install -y "${PACKAGES_TO_INSTALL[@]}"
 
 # Install nodejs and npm from Nodesource's 8.x branch
 curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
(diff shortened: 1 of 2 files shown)


jwhitlock:
I think I'm done messing with this.

One of the additional changes was to make sure that `/app` was owned by user `app` as well, in case it needs to create a new file or folder. In the dev environment (at least on macOS), this usually has the working directory mounted on top of it, so I had to make sure I was running the image to check it:

```
$ docker run -it --rm --entrypoint /bin/bash local/socorro_app
app@socorro:/app$ ls -lart /app
```

It may be a bug that `/stackwalk/` is not owned by user `app`, just the files in it.

For me, this drops the image from 1.2 GB to 1.14 GB. Not really worth spending a day on it 😢

docker/set_up_ubuntu.sh
```diff
@@ -29,20 +29,12 @@ PACKAGES_TO_INSTALL=(
...
-    wget
-    rsync
-
     # For nodejs and npm
     curl
 )
 
 # Install Ubuntu packages
-apt-get install -y ${PACKAGES_TO_INSTALL[@]}
+apt-get install -y "${PACKAGES_TO_INSTALL[@]}"
```
willkg: shellcheck pointed this out to me. It needs quotes so the array contents don't get additionally broken up if there are spaces in the values.

docker/set_up_ubuntu.sh
```diff
@@ -59,7 +54,7 @@ MYUSER="\$(id -u -n)"
 # Set the prompt to use the username we just figured out plus the container
 # name which is in an environment variable. If there is no CONTAINERNAME,
 # then use the host name.
-PS1="\${MYUSER}@socorro:\w\$ "
+PS1="\${MYUSER}@socorro:\\w\$ "
```
willkg: shellcheck pointed this out to me, too. The additional `\` escapes the `\` so then `\w` ends up in the `PS1` intentionally.

jwhitlock:
Thanks @willkg for the review and the improvements!

PR #5039, opened 2019-11-19
bug 1433148: switch from boto to boto3

willkg:
This switches the code from boto to boto3. Because that effectively required rewriting some of the code, I ended up rewriting all of it.
    
The S3Connection class is based on the one we have in Antenna with some changes:
    
* uses configman instead of everett
* has a `load_file` method
* nixes the bucket check code
* has better tests
    
Because `S3Connection` supports the local dev environment and server environments and a variety of different AWS authentication methods, this removes all the other connection context variants since we don't need them anymore.
    
This redoes the tests for crash_data and crashstorage.
    
This adds docstrings for methods along with notes and what the methods might raise.
    
Since `S3Connection` handles retrying, this removes the retrying in the crashstorage code.
    
This changes some of the configuration. We no longer have `AWS_HOST`, `AWS_PORT`, and `AWS_SECURE`. Those have been replaced with `AWS_ENDPOINT_URL`. This only affects the local dev environment.
    
This adds `ResourceWarning` to the warnings to ignore in pytest. boto3 raises a `ResourceWarning` since it doesn't explicitly close connections.

This gets rid of moto in favor of tests hitting localstack-s3 service directly.

Test plan:

1. `make build`, `make setup`, `make lint`, `make test`
2. process some crashes and make sure they show up in webapp
   1. `make run` in one terminal
   2. `make shell` in another and then `socorro-cmd fetch_crashids --num=5 | ./scripts/process_crashes.sh`
   3. wait a bit and then open one of the links
3. run verifyprocessed script: `cd webapp-django && ./manage.py verifyprocessed`
4. run upload_telemetry_schema script:
   1. `socorro-cmd upload_telemetry_schema`
   2. verify it created the `.json` file with `./scripts/socorro_aws_s3.sh ls --recursive s3://telemetry_bucket/`
5. check a crash with `http://localhost:8000/api/CrashVerify/?crash_id=UUID`
6. list the bucket and make sure paths start with `v1/` and `v2/`:
   1. `./scripts/socorro_aws_s3.sh ls --recursive s3://dev_bucket/`
   2. `./scripts/socorro_aws_s3.sh ls --recursive s3://telemetry_bucket/`

I think that's it. It verifies that keys didn't change, that everything works with itself, that the processor can load and save crash data, the telemetry crash storage works, uploading the telemetry schema works, and verifyprocessed works.


willkg:
I have a few more things to do, then this will be ready for review.

docker/config/local_dev.env
```diff
@@ -54,14 +54,11 @@ resource.elasticsearch.elasticsearch_urls=http://elasticsearch:9200
...
 # ----
 
-resource.boto.host=localstack-s3
-resource.boto.port=5000
+resource.boto.endpoint_url=http://localstack-s3:5000/
 resource.boto.access_key=foo
 secrets.boto.secret_access_key=foo
 resource.boto.bucket_name=dev_bucket
 resource.boto.temporary_file_system_storage_path=/tmp
-resource.boto.calling_format=boto.s3.connection.OrdinaryCallingFormat
```
willkg: We had to switch calling format before because we had periods in the bucket names. That's no longer true, so we can use the defaults (Antenna does) and we don't have to fiddle with calling formats anymore.

docker/run_recreate_s3_buckets.sh
```diff
@@ -12,12 +12,22 @@ set -e
...
+TELEMETRY_BUCKET="$(getenv 'destination.telemetry.bucket_name')"
+
+
 cd /app
 
 echo "Dropping and recreating S3 crash bucket..."
-(./scripts/socorro_aws_s3.sh rb s3://dev_bucket/ --force || true) 2> /dev/null # Ignore if it doesn't exist
-./scripts/socorro_aws_s3.sh mb s3://dev_bucket/
+(./scripts/socorro_aws_s3.sh rb "s3://${BUCKET}/" --force || true) 2> /dev/null # Ignore if it doesn't exist
+./scripts/socorro_aws_s3.sh mb "s3://${BUCKET}/"
```
willkg: This un-hardcodes these values. Now they're picked up from the environment.

docker/run_tests.sh
```diff
@@ -34,6 +35,7 @@ echo ">>> wait for services"
 urlwait "${DATABASE_URL}" 10
 urlwait "http://${PUBSUB_EMULATOR_HOST}" 10
 urlwait "${ELASTICSEARCH_URL}" 10
+urlwait "${CRASHSTORAGE_ENDPOINT_URL}" 10
```
willkg: Previously, all the boto/s3 things were mocked out with moto or other things. This PR ditches moto in favor of talking to the localstack-s3 service directly. These changes make sure that service is running.

requirements/default.txt
```diff
@@ -155,11 +155,9 @@ django-npm==1.0.0 \
 markus==2.1.0 \
     --hash=sha256:3408818d4d78ca0f4085c6ea32b02f4b3d2a2520b169ddec3167b50f0b7c700d \
     --hash=sha256:bfa276c69b25006e37449eed447bc27105bf6f26c9ca27d01ddefbd1077b44ea
-awscli==1.16.276 \
-    --hash=sha256:6ae2d5bcce2e16906ad34bee99f3ba3d7a1633c8f0f53ef97e7575499b5a88eb \
-    --hash=sha256:6c63da10a322f1245f90df99b549074d97b08f0e6f69d8ce7f06330b4822b792
-moto==1.3.13 \
-    --hash=sha256:95d48d8ebaad47fb5bb4233854cf1cf8523ec5307d50eb1e4017ce10f1960b66
```
willkg: I ditched moto, removed boto, and moved boto3 to the `requirements/default.txt` file.

scripts/socorro_aws_s3.sh
```diff
@@ -58,8 +57,6 @@ aws_secret_access_key = ${AWS_SECRET_ACCESS_KEY}
 EOF
 fi
 
-AWSOPTIONS="--endpoint-url=http://${AWS_HOST}:${AWS_PORT}/"
-
-echo "S3 container bucket is ${AWS_BUCKET}"
```
willkg: I took this line out because it's confusing when you're making and removing buckets.

setup.cfg
```diff
@@ -33,8 +33,8 @@ filterwarnings =
     ignore::ImportWarning
     ignore:::grpcio[.*]
     ignore:::google[.*]
-    # boto imports imp not importlib, but boto3 update not planned, see 1433148
-    ignore:.*the imp module is deprecated:DeprecationWarning
```
willkg: We no longer use boto, so we can ditch this.

setup.cfg
```diff
@@ -33,8 +33,8 @@ filterwarnings =
...
+    # boto3 kicks up ResourceWarning: unclosed socket, https://github.com/boto/boto3/issues/454
+    ignore:unclosed:ResourceWarning
```
willkg: However, boto3 kicks up a `ResourceWarning`, so we have to ignore that in the tests otherwise it prints a bunch of stuff out that's annoying and unhelpful.

socorro/external/boto/connection_context.py
```diff
@@ -2,38 +2,72 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import contextlib
-import datetime
-import json
-import socket
+import io
```
willkg: This file is based on the `S3Connection` from Antenna. It has the following differences:

* uses configman rather than everett
* drops the health check code
* adds a `load_file` function
* uses different wait times for retrying (Antenna needs to ensure the crash gets to the destination, the processor can just fail and try again later)
* includes `S3Connection` specific tests

Probably best to read this file on its own and not as a diff.

socorro/external/boto/crash_data.py
```diff
@@ -95,7 +95,7 @@ def get(self, **kwargs):
             raise MissingArgumentError("uuid")
 
         try:
-            return self.get_unredacted_processed(params.uuid)
+            return self.get_unredacted_crash_report(params.uuid)
```
willkg: I'm going to undo this and related changes to the `TelemetryBotoS3CrashStorage`.

socorro/external/boto/crashstorage.py
```diff
@@ -28,13 +30,101 @@ def wait_time_generator():
         yield i
 
 
+class CrashidMissingDatestamp(Exception):
+    """Indicates the crash id is invalid and missing a datestamp."""
```
willkg: We have a `socorro.lib.ooid.is_crash_id_valid`. It validates the structure, but doesn't verify the datestamp at the end is a valid datestamp. I thought about fixing that which would allow me to reduce this code a bit. I decided it was outside the scope of this change.

At some point, we should go through `socorro.lib.ooid` and similar things in Antenna and unify them into a single library that is in both projects.

socorro/external/boto/crashstorage.py
```diff
@@ -28,13 +30,101 @@ def wait_time_generator():
...
+
+    """
+    datestamp = date_from_ooid(crashid)
+    if datestamp is None:
+        # We should never hit this situation unless the crashid is not valid
+        raise CrashidMissingDatestamp("%s is missing datestamp" % crashid)
+    return datestamp
+
+
+def build_keys(name_of_thing, crashid):
```
willkg: This is still titled `build_keys` even though it returns a list of a single key in all cases. I think it's likely we'll want to do a migration pass removing the entropy from the paths and figure it's easier to leave this as is.

If, however, you think that's not worth waiting for, I can spend some time de-pluralizing this in this PR.

socorro/external/boto/crashstorage.py
```diff
@@ -61,116 +151,99 @@ def __init__(self, config, namespace="", quit_check_callback=None):
...
-        except conn.ResponseError as x:
-            raise CrashIDNotFound("%s not found: %s" % (crash_id, x))
-
     def get_raw_crash(self, crash_id):
-        """Retrieve a raw_crash."""
-        retryable_get = self.retryable(self.do_get_raw_crash)
-        with self.connection_source() as conn:
-            return retryable_get(
-                conn, crash_id=crash_id, json_object_hook=self.config.json_object_hook
+        """Get the raw crash file for the given crash id."""
```
willkg: I should add `:returns:` and `:raises:` to these docstrings.

socorro/external/boto/crashstorage.py
```diff
@@ -28,13 +30,101 @@ def wait_time_generator():
         yield i
 
 
+class CrashidMissingDatestamp(Exception):
```
willkg: There's a `CrashIDNotFound` exception and then this one `CrashidMissingDateStamp`. I should change this to `CrashIDMissingDatestamp` for consistency.

socorro/external/boto/crashstorage.py
```diff
@@ -61,116 +153,119 @@ def __init__(self, config, namespace="", quit_check_callback=None):
...
             for dump_name in dump_names:
                 if dump_name in (None, "", "upload_file_minidump"):
                     dump_name = "dump"
-                dumps[dump_name] = conn.fetch(crash_id, dump_name)
+                path = build_keys(dump_name, crash_id)[0]
+                dumps[dump_name] = self.conn.load_file(path)
             return dumps
-        except conn.ResponseError as x:
+        except self.conn.KeyNotFound as x:
+            # FIXME--different errors
```
willkg: This is an errant fixme.

socorro/unittest/conftest.py
```diff
@@ -58,54 +60,86 @@ def caplogpp(caplog):
...
+        )
+        client = session.client(
+            service_name="s3",
+            config=Config(s3={"addressing_style": "path"}),
+            endpoint_url=os.environ.get("resource.boto.endpoint_url"),
+        )
+        return client
+
+    def __enter__(self):
+        return self
```
willkg: This should set `_buckets_seen` to an empty set.

socorro/unittest/conftest.py
```diff
@@ -58,54 +60,86 @@ def caplogpp(caplog):
...
+        session = boto3.session.Session(
+            aws_access_key_id=os.environ.get("resource.boto.access_key"),
+            aws_secret_access_key=os.environ.get("secrets.boto.secret_access_key"),
+        )
+        client = session.client(
+            service_name="s3",
+            config=Config(s3={"addressing_style": "path"}),
+            endpoint_url=os.environ.get("resource.boto.endpoint_url"),
+        )
+        return client
```
willkg: `__init__` should create the client and keep it.

socorro/unittest/external/boto/__init__.py
```diff
@@ -26,7 +29,7 @@ def get_config(cls, values_source=None):
         app_name="testapp",
         app_version="1.0",
         app_description="",
-        values_source_list=[values_source],
+        values_source_list=[environment, values_source],
```
willkg: This changes tests such that they use environment variables from `docker/config/test.env` rather than who knows what.

socorro/unittest/external/boto/test_connection_context.py
```diff
@@ -2,249 +2,56 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
willkg: This is taken from Antenna with the addition of the `load_file` tests.

socorro/unittest/external/boto/test_crash_data.py
```diff
@@ -13,75 +12,75 @@
...
 
-class TestSimplifiedCrashData(object):
+class TestSimplifiedCrashData:
     def get_s3_store(self):
         return SimplifiedCrashData(config=get_config(SimplifiedCrashData))
 
-    @mock_s3_deprecated
     def test_get_processed(self, boto_helper):
-        boto_helper.set_contents_from_string(
-            bucket_name="crashstats",
```
willkg: The old code hardcoded the bucket used.

socorro/unittest/external/boto/test_upload_telemetry_schema.py
```diff
@@ -2,30 +2,16 @@
...
-    @mock_s3_deprecated
-    def test_bucket_not_found(self, caplogpp):
-        # If the bucket isn't found, the script should tell the user and return
-        # a non-zero exit code
-        config = get_config(UploadTelemetrySchema)
-        app = UploadTelemetrySchema(config)
-
-        assert app.main() == 1
-        recs = [rec.message for rec in caplogpp.records]
-        assert "Failure: The crashstats S3 bucket must be created first." in recs
```
willkg: I nixed a lot of the "let's do special things if the bucket isn't there". It's an edge case that only happens when setting things up or if you've misconfigured the bucket name. In all those cases, it should be sending something to Sentry or the logs. It should be really obvious without additional special code we have to maintain that the bucket is either misconfigured or not there.

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -702,8 +700,12 @@ def test_bad_uuid(self, client):
         data = json.loads(resp.content)
         assert data == {"error": "unknown crash id"}
 
-    @mock_s3_deprecated
-    def test_elastcsearch_has_crash(self, client):
+    def test_elastcsearch_has_crash(self, boto_helper, client):
+        bucket = settings.SOCORRO_CONFIG["resource"]["boto"]["bucket_name"]
+        boto_helper.create_bucket(bucket)
+        telemetry_bucket = settings.SOCORRO_CONFIG["telemetrydata"]["bucket_name"]
+        boto_helper.create_bucket(telemetry_bucket)
```
willkg: Because we're not using moto and mocking out all s3 things, we need to create the buckets so that they don't throw a "bucket isn't there" error.

willkg: I should centralize this.

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -782,21 +788,23 @@ def test_processed_has_crash(self, client):
...
             "platform": "Linux",
             "signature": "now_this_is_a_signature",
             "uuid": uuid,
         }
 
-        boto_helper = BotoHelper()
-        boto_helper.get_or_create_bucket("telemetry-test")
-        boto_helper.set_contents_from_string(
+        boto_helper.upload_fileobj(
             bucket_name="telemetry-test",
```
willkg: Oops--hardcoded bucket name.

webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
```diff
@@ -27,24 +28,42 @@
...
 
+# Number of seconds until we decide a worker has stalled
+WORKER_TIMEOUT = 10 * 60
+
 
 metrics = markus.get_metrics("cron.verifyprocessed")
 
 
-def check_crashids(entropy, boto_conn, bucket_name, date):
+def check_crashids(entropy_chunk, date):
```
willkg: The boto S3 client was pickleable, but the boto3 S3 client is not. The problem that creates is that when the main process spins off a bunch of subprocesses to do the work, it can no longer pass along a boto3 S3 client. Thus each subprocess has to create its own client.

Creating a client takes 0.5s for some reason. We used to create like 4096 processes. So that's another 2000 seconds or so. That makes the script take a *LOT* longer to run.

The other thing I noticed in the "the bucket is empty" case, is that this script spends an awful lot of time creating and handling processes. `verifyprocessed` takes 2m to run in the master branch with nothing in the bucket. That's kind of silly.

To reduce the cost of creating boto3 S3 clients and the overhead of creating and handling processes, I changed this to chunk entropy values. Now each process handles 5 entropy values rather than 1. That significantly reduces the work required for the "there's nothing in the bucket" case and it takes 1m30s or so for me to run.

Socorro is using Python 3.6. Once we get to 3.7, it might be even better to switch to asyncio since this is mostly waiting on network i/o.

webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
```diff
@@ -74,25 +93,19 @@ def get_entropy(self):
...
-            for result in map(check_crashids_for_date, entropy):
+            for result in map(check_crashids_for_date, entropy_chunked):
                 missing.extend(result)
         else:
             with concurrent.futures.ProcessPoolExecutor(
                 max_workers=num_workers
             ) as executor:
                 for result in executor.map(
-                    check_crashids_for_date, entropy, timeout=300
+                    check_crashids_for_date, entropy_chunked, timeout=WORKER_TIMEOUT
```
willkg: In early rounds of making this faster, I was exceeding the timeout, so I stuck it in a constant with a comment.

webapp-django/crashstats/crashstats/tests/test_verifyprocessed.py
```diff
@@ -32,61 +38,72 @@ def test_get_entropy(self):
         assert entropy[0] == "000"
         assert entropy[-1] == "fff"
 
-    @mock_s3_deprecated
-    def test_no_crashes(self, boto_helper):
+    def test_no_crashes(self, boto_helper, monkeypatch):
         """Verify no crashes in bucket result in no missing crashes."""
-        boto_helper.get_or_create_bucket(BUCKET_NAME)
+        monkeypatch.setattr(Command, "get_entropy", get_small_entropy)
```
willkg: This used to go through all 4096 values of entropy. We don't need to do that for the tests, so I finally got around to mocking that to a smaller number.

webapp-django/crashstats/settings/base.py
```diff
@@ -588,21 +588,8 @@ def filter(self, record):
...
-                "resource.boto.resource_class",
-                "socorro.external.boto.connection_context.RegionalS3ConnectionContext",
-            ),
-            "host": config("resource.boto.host", None),
-            "port": config("resource.boto.port", None),
-            "secure": config("resource.boto.secure", True),
-            "calling_format": config(
-                "resource.boto.calling_format",
-                "boto.s3.connection.OrdinaryCallingFormat",
-            ),
```
willkg: We don't use host, port, or secure anymore. Those are replaced with endpoint_url and only used in the dev environment.

We never used prefix, so I nixed that.

The resource_class won't change anymore, so I hardcoded that.

We no longer need calling_format.

None of these changes should affect server environments.

willkg:
This is ready for review!

webapp-django/crashstats/settings/base.py
```diff
@@ -588,21 +588,8 @@ def filter(self, record):
...
-            # NOTE(willkg): In the local dev environment, we need to use a
-            # HostPortS3ConnectionContext which requires these additional configuration bits. The
-            # defaults are taken from the config sections of the relevant classes.
-            "resource_class": config(
-                "resource.boto.resource_class",
-                "socorro.external.boto.connection_context.RegionalS3ConnectionContext",
-            ),
-            "host": config("resource.boto.host", None),
-            "port": config("resource.boto.port", None),
-            "secure": config("resource.boto.secure", True),
```
jwhitlock: There's a `resource.boto.secure=false` that  can probably be removed from `docker/config/never_on_a_server.env` now

socorro/external/boto/connection_context.py
```diff
@@ -47,280 +81,135 @@ class S3ConnectionContext(RequiredConfig):
...
+            session_kwargs["aws_access_key_id"] = self.config.access_key
+            session_kwargs["aws_secret_access_key"] = self.config.secret_access_key
+        session = boto3.session.Session(**session_kwargs)
+
+        kwargs = {
+            "service_name": "s3",
+            "region_name": self.config.region,
+            # NOTE(willkg): We use path-style because that lets us have dots in
+            # our bucket names and use SSL.
+            "config": Config(s3={"addressing_style": "path"}),
```
jwhitlock: Is the change of addressing style needed? Other comments suggest that we no longer use periods in bucket names.

socorro/external/boto/connection_context.py
```diff
@@ -2,38 +2,72 @@
...
+    We have this problem where we're binding IAM credentials to the EC2 node
+    and on startup when boto3 goes to get the credentials, it fails for some
+    reason and then degrades to hitting the https://s3..amazonaws.net/
+    endpoint and then fails because that's not a valid endpoint.
+
+    This sequence increases the wait times and adds some jitter.
+
+    """
+    for i in [5, 5, 5, 5, 5]:
+        yield i + random.uniform(-2, 2)  # nosec
```
jwhitlock: what does `# nosec` do?

webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
```diff
@@ -27,24 +28,42 @@
...
+                crash_id = raw_crash_key.split("/")[-1]
+
+                try:
+                    s3_client.head_object(
+                        Bucket=bucket, Key=PROCESSED_CRASH_TEMPLATE % crash_id
+                    )
+                except s3_client.exceptions.ClientError as exc:
+                    # If we got back a 404: Not Found, then the processed crash isn't
+                    # there. If we got something else back, re-raise it.
+                    if "404" in str(exc):
```
jwhitlock: 🤢 Maybe the Tecken version would be slightly better:

```
if exc.response["Error"]["Code"] == "404"
```

socorro/external/boto/crashstorage.py
```diff
@@ -61,116 +153,118 @@ def __init__(self, config, namespace="", quit_check_callback=None):
...
     def save_processed(self, processed_crash):
-        """Save processed_crash."""
-        retryable_save = self.retryable(self.do_save_processed)
-        with self.connection_source() as conn:
-            retryable_save(conn, processed_crash=processed_crash)
+        """Save the processed crash file."""
+        crash_id = processed_crash["uuid"]
+        data = dict_to_str(processed_crash).encode("utf-8")
+        path = build_keys("processed_crash", crash_id)[0]
+        self.conn.save_file(path, data)
```
jwhitlock: I find this easier to read than the old `do_save_processed` / `save_processed` code. Nice change!

docker/run_recreate_s3_buckets.sh
```diff
@@ -12,12 +12,22 @@ set -e
...
 # resource.boto.telemetry_bucket_name
 
+# First convert configman environment vars which have bad identifiers to ones
+# that don't
+function getenv {
+    python -c "import os; print(os.environ['$1'])"
+}
+
+BUCKET="$(getenv 'resource.boto.bucket_name')"
+TELEMETRY_BUCKET="$(getenv 'destination.telemetry.bucket_name')"
```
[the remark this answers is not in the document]
willkg: Oh, right! That can be removed now.

webapp-django/crashstats/settings/base.py
willkg: Yes! I'll do this.

socorro/external/boto/connection_context.py
willkg: I sometimes run `bandit` on the code. The `nosec` tells bandit that, yes, this is a call to random, but it's ok and not a security issue.

webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
willkg: Mmm... yes. It probably would be. I'll fix it.

socorro/external/boto/connection_context.py
willkg: I forgot about this. I did write comments elsewhere about this. I'll have to double-check to see what the deal is, but I'm pretty sure we're fine now.

If we change this code, I'll make sure to change Antenna to match. At a minimum, I'll update the comment.

willkg: This line was in Antenna from the beginning (September 2016) and at that time, we did have periods in the bucket names. Pretty sure we don't anymore because periods are the worst.

"virtual" `addressing_style` also doesn't work with the local dev environment because it starts building paths like `http://dev_bucket.localstack-s3:3000/` and that won't work.

However, it looks like the default for addressing_style is "auto" where it'll try "virtual' first and if that fails, use "path":

https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html?highlight=addressing_style#configuration-file

Looks like that was changed in botocore 1.9.0 in 2018:

* https://github.com/boto/botocore/blob/develop/CHANGELOG.rst#190
* https://github.com/boto/botocore/pull/1387

I'll remove these three lines from Socorro and Antenna.

willkg:
Big thank you for reviewing this!

PR #5040, opened 2019-11-21
bug 1539305: update to python 3.7.5

willkg:


socorro/lib/threaded_task_manager.py
```diff
@@ -171,7 +171,7 @@ def _responsive_join(self, thread, waiting_func=None):
         while True:
             try:
                 thread.join(1.0)
-                if not thread.isAlive():
+                if not thread.is_alive():
```
willkg: `isAlive` is deprecated, so I switched these to `is_alive`.

socorro/unittest/external/test_crashstorage_base.py
```diff
@@ -155,7 +155,7 @@ def test_polyerror(self):
             )
             assert (
                 str(x)
-                == "hell,NameError('dwight',),KeyError('wilma',),AttributeError('sarita',)"
+                == "hell,NameError('dwight'),KeyError('wilma'),AttributeError('sarita')"
```
willkg: This is comparing the str of an exception with a string. The str output changed, so I updated the expected text.

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -346,7 +347,7 @@ def test_quick_search(self):
         # Test with a signature.
         response = self.client.get(url, {"query": "moz"})
         assert response.status_code == 302
-        target = reverse("supersearch:search") + "?signature=%7Emoz"
+        target = reverse("supersearch:search") + "?signature=~moz"
```
willkg: `%7E` is `~`. I think something changed between 3.6 and 3.7 such that the `response["location"]` isn't urlencoding the `~` anymore. I didn't look into it.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -23,13 +22,6 @@
 from socorro.external.crashstorage_base import CrashIDNotFound
 
 
-# To prevent running in to a known Python bug
-# (http://bugs.python.org/issue7980)
-# we, here at "import time" (as opposed to run time) make use of time.strptime
-# at least once
-datetime.datetime.strptime("2013-07-15 10:00:00", "%Y-%m-%d %H:%M:%S")
```
willkg: This bug was a Python 2.7 bug and isn't an issue anymore, so I nixed this.

willkg:
Thank you!

PR #5048, opened 2019-12-03
Update current Fennec versions

lizzard:
Release day for 68.3 Fennec.


willkg:
It takes a few minutes for the caches in prod to expire, then it'll pick this up. We don't need to deploy it for it to take effect.

PR #5049, opened 2019-12-03
bug 1599779: support other crashid forms in signature command

willkg:
This only affects the signature command line utility and doesn't affect
signature generation. It makes it easier to copy/paste crash ids from
various sources (bugzilla, urls, etc) to test signature generation.

To test:

```
app@socorro:/app$ socorro-cmd signature https://crash-stats.mozilla.org/report/index/0b794045-87ec-4649-9ce1-73ec10191120
Crash id: 0b794045-87ec-4649-9ce1-73ec10191120
Original: NXMapRemove
New:      NXMapRemove | -[NSControl sendAction:to:]
Same?:    False
```

### DIFF
--- a/socorro/signature/cmd_signature.py
+++ b/socorro/signature/cmd_signature.py
@@ -10,7 +10,7 @@ import sys
 import requests
 
 from .generator import SignatureGenerator
-from .utils import convert_to_crash_data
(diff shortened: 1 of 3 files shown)


willkg:
self-r+

PR #5052, opened 2019-12-04
bug 1578781: remove markdown from docs

willkg:
We used to use markdown to include the QA docs, but we no longer use markdown for the QA docs, so we don't need the markdown support for Sphinx.

To test:

```
$ make build-docs
$ make docs
```


willkg:
self-r+

I'll double-check RTD in an hour after merging this.

PR #5055, opened 2019-12-10
bug 1602120: sqs local dev

willkg:
This updates localstack image to 0.10.5 which is the latest one. It also switches the localstack-s3 service port from 5000 to 4572 which is the default one used by that image.

This also adds:
    
* configuration for `sqs_endpoint_url`
* `sqs_cli` script
* changes `resource.boto.endpoint_url` to `resource.boto.s3_endpoint_url` to differentiate it with the `sqs_endpoint_url`
* adds localstack-sqs service and relevant startup bits
* moves click to a default dependency since we use it in `sqs_cli`
* fixes some weird configuration things in `docker/run_tests.sh`
* removes some unused env vars

To test:

1. do `make build`
2. do `make runservices` -- this should start the localstack-sqs service
2. do `make test` -- this should run the tests with the fixed configuration


docker/config/test.env
```diff
@@ -19,13 +19,7 @@ resource.elasticsearch.elasticsearch_index=testsocorro%Y%W
 resource.elasticsearch.elasticsearch_index_regex=^testsocorro[0-9]{6}$
 
 # Postgres configuration
-database_url=postgres://postgres:aPassword@postgresql:5432/socorro_test
 DATABASE_URL=postgres://postgres:aPassword@postgresql:5432/socorro_test
-database_hostname=postgresql
-database_port=5432
-database_username=postgres
-database_password=aPassword
-database_name=socorro_test
```
willkg: These database variables aren't used anywhere, so I removed them.

docker/config/test.env
```diff
@@ -19,13 +19,7 @@ resource.elasticsearch.elasticsearch_index=testsocorro%Y%W
 resource.elasticsearch.elasticsearch_index_regex=^testsocorro[0-9]{6}$
 
 # Postgres configuration
-database_url=postgres://postgres:aPassword@postgresql:5432/socorro_test
```
willkg: I switched the `docker/run_tests.sh` script to use `DATABASE_URL` instead.

docker/run_tests.sh
```diff
@@ -20,22 +20,28 @@ set -v -e -x
...
 PYTEST="$(which pytest)"
 PYTHON="$(which python)"
 
-echo ">>> wait for services"
-# Wait for postgres and elasticsearch services to be ready
+echo ">>> wait for services to be ready"
 urlwait "${DATABASE_URL}" 10
 urlwait "http://${PUBSUB_EMULATOR_HOST}" 10
 urlwait "${ELASTICSEARCH_URL}" 10
-urlwait "${CRASHSTORAGE_ENDPOINT_URL}" 10
```
willkg: `CRASHSTORAGE_ENDPOINT_URL` is an Antenna variable--we shouldn't have been using that here.

requirements/default.txt
```diff
@@ -7,6 +7,9 @@ configman==1.3.0 \
     --hash=sha256:bc96b4c974aefabf4605d82dfab409a5aed10b4ae0b05ae2535eda415a3f7b67 \
     --hash=sha256:8e45c20c7b73173bd4d0194521f982201b70558d6f6e945f11053201110b83e0 \
     --hash=sha256:4000037487d65ec05e3d4183298f0a5799d7c71a47e8753f4edeede586baeb50
+click==7.0 \
+    --hash=sha256:2335065e6395b9e67ca716de5f7526736bfa6ceead690adf616d925bdc622b13 \
+    --hash=sha256:5b94b49521f6456670fdb30cd82a4eca9412788a93fa6dd6df72c94d5a8ff2d7
```
willkg: `sqs_cli` uses click, so we need this to be a default dependency. I think I want to rewrite the other scripts to also use click, but that's a future low-priority task.

socorro/scripts/sqs_cli.py
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/env python
```
willkg: This is pretty much the same one we have in Antenna, but Socorro-ized.

willkg:
These changes only affect the local development environment and CI. CI passed and this works fine in my local dev environment. If we hit issues with other local dev environments, we can fix them as they come up. For those reasons, self-r+.

PR #5056, opened 2019-12-11
bug 1602121: fix CrashQueue classes and redo docs

willkg:
This moves some CrashQueue bits up to a base class so we can implement other crash queue classes without having to copy and paste a bunch of stuff.

This also moves around the crashstorage documentation and cleans some of that up, too.


willkg:
There's some minor code adjustments with the crash queue stuff, but it's pretty basic.

There's a bunch of documentation changes, but mostly it updates what we already had. In the future, I'd like to rewrite it and have more of automatically generated from the code. Everett makes that easy. I have some code that can extract configuration information from configman, but I never got it working well enough.

Anyhow, my current thinking is that at some point, we'll switch from configman to Everett and then we can automatically generate a lot of this documentation from the source code which eases maintenance of the docs.

In the meantime, self-r+.

PR #5057, opened 2019-12-12
bug 1603231: fix verifyprocessed

willkg:
The verifyprocessed command failed because the `check_past_missing()` method hadn't been updated to boto3 things. There were no tests, so it wasn't caught.

This adds tests and fixes the code.

### DIFF
--- a/webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
+++ b/webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
@@ -146,10 +146,13 @@ class Command(BaseCommand):
         no_longer_missing = []
 
         for crash_id in crash_ids:
-            bucket = s3_client.get_bucket(bucket_name)
-            processed_crash_key = bucket.get_key(PROCESSED_CRASH_TEMPLATE % crash_id)
-            if processed_crash_key is not None:
+            try:
+                key = PROCESSED_CRASH_TEMPLATE % crash_id
+                s3_client.head_object(Bucket=bucket_name, Key=key)
                 no_longer_missing.append(crash_id)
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
```diff
@@ -146,10 +146,13 @@ def check_past_missing(self):
         no_longer_missing = []
 
         for crash_id in crash_ids:
-            bucket = s3_client.get_bucket(bucket_name)
```
willkg: `get_bucket()` is boto code. This needed to be updated for boto3. Oops.

webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
```diff
@@ -146,10 +146,13 @@ def check_past_missing(self):
...
-            processed_crash_key = bucket.get_key(PROCESSED_CRASH_TEMPLATE % crash_id)
-            if processed_crash_key is not None:
+            try:
+                key = PROCESSED_CRASH_TEMPLATE % crash_id
+                s3_client.head_object(Bucket=bucket_name, Key=key)
                 no_longer_missing.append(crash_id)
+            except s3_client.exceptions.ClientError as exc:
+                if exc.response["Error"]["Code"] != "404":
+                    raise
```
willkg: If it's a 404, then the processed crash isn't there and this crash is still missing--no biggie. If it's an exception, but not a 404, then we don't know what's going on, so re-raise it.

willkg:
verifyprocessed is failing on both stage and prod with the same error. It runs every morning, so I'll land this now and check stage tomorrow and verify it then.

self-r+

PR #5058, opened 2019-12-12
bug 1603236: fix unknown property indexing failure

willkg:
Let's say we have two crash reports that have a field that isn't in super search fields. The first crash report has a field value that's a big string. The second one has a nested value. When the first is
indexed, Elasticsearch will infer that field to be a string. When the second gets indexed, it'll raise an "unknown property" error and not index the crash report.

This fixes that situation by adding code that recognizes what just happened, drops the field from the document, and retries indexing. This adds a test, too. This also adds metrics so we can see how often
these indexing errors happen.

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -201,6 +201,7 @@ class ESCrashStorage(CrashStorageBase):
     # have been tested with Elasticsearch 1.4.
     field_name_string_error_re = re.compile(r"field=\"([\w\-.]+)\"")
     field_name_number_error_re = re.compile(r"\[failed to parse \[([\w\-.]+)]]")
+    field_name_unknown_property_error_re = field_name_number_error_re
 
     def __init__(self, config, namespace="", quit_check_callback=None):
         super().__init__(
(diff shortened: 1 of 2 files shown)


socorro/external/es/crashstorage.py
```diff
@@ -201,6 +201,7 @@ class ESCrashStorage(CrashStorageBase):
     # have been tested with Elasticsearch 1.4.
     field_name_string_error_re = re.compile(r"field=\"([\w\-.]+)\"")
     field_name_number_error_re = re.compile(r"\[failed to parse \[([\w\-.]+)]]")
+    field_name_unknown_property_error_re = field_name_number_error_re
```
willkg: These regular expressions are so janky. But the messages are not structured, so I'm not sure whether there are better options. :(

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -598,12 +599,14 @@ def mock_index(*args, **kwargs):
...
-                    "NumberFormatException[For input string: "
-                    '"18446744073709480735"]; ',
+                    (
+                        "RemoteTransportException[[i-f94dae31][inet[/172.31.1.54:"
+                        "9300]][indices:data/write/index]]; nested: "
+                        "MapperParsingException[failed to parse "
+                        "[processed_crash.bogus-field]]; nested: "
+                        "NumberFormatException[For input string: "
+                        '"18446744073709480735"]; '
+                    ),
```
willkg: If these tests weren't mocked out, then they would tell us what happens when we update to a recent version of Elasticsearch. We should probably redo these tests so they run against an actual Elasticsearch.

willkg:
This uses existing scaffolding to solve this new problem. We haven't seen this problem in stage because the stage-submitter always generates crashes in multipart/form-data format. (We should fix that. :cry: )

I can reproduce it in stage, though. I'm going to land this now, wait for it to deploy, and verify it there. self-r+

PR #5059, opened 2019-12-12
bug 1603308: fix verifyprocessed

willkg:
This fixes a couple of things in verifyprocessed command:

1. It fixes the day it checks. Before it would check "today", but since it ran at 4:00am, there wasn't much to check, so it probably hasn't been checking everything in ages. Plus since it's checking "today", it's picking up some raw crashes that are being processed and thinks they're missing. Bleh.

2. It adds checking for Elasticsearch and considers a crash missing if it's either not on S3 or not in Elasticsearch. This will plug a hole where some things could end up on S3, but due to bugs in the Elasticsearch crashstorage code, never make it to Elasticsearch and thus never be searchable. Bleh.

In the process of doing that, I had to rework how the tests work with Elasticsearch. The code changes are much better. While doing that, I hit bugs in how the webapp tests mock the socorro middleware model class implementations, so I fixed that, too.

To test:

1. run `make setup` to clear out all storage things
2. run `./scripts/process_crashes.sh CRASHID` which will pull down crash data and put the raw crash in S3--don't run the processor, yet
3. run `cd webapp; ./manage.py verifyprocessed` which will see that as a missing crash
4. run the processor
5. run `socorro-cmd es delete` to and then `socorro-cmd es create` to wipe out Elasticsearch storage so now the crash is on S3 but not Elasticsearch
6. run `cd webapp; ./manage.py verifyprocessed` which will see that the crash is still missing
7. process the crash
8. run `cd webapp; ./manage.py verifyprocessed` which will see that the crash is no longer missing--yay!


socorro/scripts/es.py
```diff
@@ -68,6 +69,18 @@ def cmd_list():
...
+def cmd_list_crashids(index):
+    """List crashids for index."""
+    es_conn = get_conn()
+    with es_conn() as conn:
+        search = Search(using=conn, index=index, doc_type=es_conn.get_doctype(),)
+        search = search.fields("processed_crash.uuid")
+        results = search.execute()
+        print("Crashids in %s:" % index)
+        for hit in results:
+            print(hit["processed_crash.uuid"][0])
```
willkg: This is helpful since it prints all the crash ids in an ES index. It makes it easier to answer questions like, "The thing I'm looking for isn't coming up in search in this test. What's in there?"

socorro/unittest/conftest.py
```diff
@@ -144,3 +152,26 @@ def boto_helper():
     """
     with BotoHelper() as boto_helper:
         yield boto_helper
+
+
+@pytest.fixture
+def es_conn():
```
willkg: We had this fixture in two different places. I centralized it here.

socorro/unittest/conftest.py
```diff
@@ -144,3 +152,26 @@ def boto_helper():
...
+    manager = ConfigurationManager(
+        ESConnectionContext.get_required_config(), values_source_list=[environment]
+    )
+    conn = ESConnectionContext(manager.get_config())
+
+    # Create two indexes--this week and last week
+    template = conn.config.elasticsearch_index
+    conn.create_index(utc_now().strftime(template))
+    conn.create_index((utc_now() - datetime.timedelta(weeks=1)).strftime(template))
+    conn.refresh()
```
willkg: Indexing crash reports will create missing indexes, but when setting up a test, it's easier if the test harness does it. Another way to do this that wouldn't have required me to fix the connection_context tests would be to create an ESHelper like we have the BotoHelper which has convenience methods for doing things. I like that idea a lot, but I'm sort of rushing this work out so I didn't want to spend time on that.

It's definitely a good future direction. It'll clean up a chunk of test code and make it much easier to remove a whole bunch of mocking.

socorro/unittest/external/es/test_connection_context.py
```diff
@@ -15,8 +15,11 @@
 
 class TestConnectionContext:
     def test_create_index(self, es_conn):
-        # Verify there are no indices first
-        assert list(es_conn.get_indices()) == []
+        # Delete any existing indices first
+        for index in es_conn.get_indices():
+            es_conn.delete_index(index)
+        es_conn.refresh()
+        es_conn.health_check()
```
willkg: Because I changed `es_conn` to create two indexes, we have to delete them for this test to run.

socorro/unittest/external/es/test_connection_context.py
```diff
@@ -40,6 +43,12 @@ def test_delete_index(self, es_conn):
         assert index_name not in list(es_conn.get_indices())
 
     def test_delete_old_indices(self, es_conn):
+        # Delete any existing indices first
+        for index in es_conn.get_indices():
+            es_conn.delete_index(index)
+        es_conn.refresh()
+        es_conn.health_check()
```
willkg: Because I changed `es_conn` to create two indexes, we have to delete them for this test to run.

webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
```diff
@@ -173,9 +199,15 @@ def handle(self, **options):
             if not check_date:
                 raise CommandError("Unrecognized run_time format: %s" % check_date_arg)
         else:
-            check_date = timezone.now() - datetime.timedelta(days=1)
+            check_date = timezone.now()
+
+        # Look at the previous day because we want to look at a whole day.
+        check_date = check_date - datetime.timedelta(days=1)
```
willkg: This is a little annoying since if your'e running it manually on the command line and give it a date, feels like you expect it to check the crash reports on that date, but instead it does the previous day. However, that's how cronrun is doing it and it's running it most of the time, so I figured I'd cater to that and make a note in the docs.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -694,6 +693,7 @@ class ProcessedCrash(SocorroMiddleware):
 
 
 class UnredactedCrash(ProcessedCrash):
+    implementation = SimplifiedCrashData
```
willkg: Subclassing and not specifying the `implementation` messes with the way the implementations get mocked out in the tests. We should fix the tests to do a better job of patching, but it was easier to do this for now.

webapp-django/crashstats/crashstats/tests/conftest.py
```diff
@@ -86,13 +85,10 @@ def set_product_versions(self, versions):
         self.mock_gvfp.return_value = versions
 
 
-class BaseTestViews(ProductVersionsMixin, DjangoTestCase):
+class SuperSearchFieldsMock:
```
willkg: I extracted this out as a mixin because it wasn't cleaning itself up correctly and this is more correct.

willkg:
self-r+

PR #5060, opened 2019-12-13
bug 1603308: fix date range for ES lookup

willkg:
Previously, verifyprocessed would do an ES search using the default date range which is a week ago to today. If the crash it was looking at was outside of that range, it wouldn't get found.

This fixes it to look at the day the crash id says the collector picked it up.

### DIFF
--- a/webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
+++ b/webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
@@ -24,6 +24,7 @@ from django.utils.dateparse import parse_date, parse_datetime
 from crashstats.crashstats.configman_utils import get_s3_context
 from crashstats.crashstats.models import MissingProcessedCrash
 from crashstats.supersearch.models import SuperSearchUnredacted
+from socorro.lib.ooid import date_from_ooid
 
 
 RAW_CRASH_PREFIX_TEMPLATE = "v2/raw_crash/%s/%s/"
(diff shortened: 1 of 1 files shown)


webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
```diff
@@ -53,8 +54,12 @@ def is_in_s3(s3_client, bucket, crash_id):
...
     """Is the processed crash in Elasticsearch."""
+    crash_date = date_from_ooid(crash_id)
+    start_date = crash_date.strftime("%Y-%m-%d")
+    end_date = (crash_date + datetime.timedelta(days=1)).strftime("%Y-%m-%d")
     params = {
         "uuid": crash_id,
+        "date": [">=%s" % start_date, "<=%s" % end_date],
         "_results_number": 1,
         "_columns": ["uuid"],
         "_facets": [],
```
willkg: This gives us ES queries like this where the date range is based on the date portion of the crash id.

```
{'query': {'filtered': {'query': {'match_all': {}}, 'filter': {'bool': {'must': [{'bool': {'must': [{'range': {'processed_crash.date_processed': {'lte': '2019-12-14T00:00:00+00:00'}}}, {'range': {'processed_crash.date_processed': {'gte': '2019-12-13T00:00:00+00:00'}}}]}}, {'term': {'processed_crash.uuid': '0000ddbd-1438-4ad4-8d0f-be8992191213'}}]}}}}, 'from': 0, 'size': 1, 'fields': ['processed_crash.uuid']}
```

willkg:
`verifyprocessed` would have worked fine when running for *today*, but will fail if we ever fiddle with the `cron_job` record to get it to run on older days or if it has to do some serious backfill.

willkg:
I'll continue testing this on stage. self-r+

PR #5061, opened 2019-12-13
bug 1603308: move update missing task into updatemissing command

willkg:
This splits verifyprocessed into two tasks. Checking existing known missing processed crashes to see if they've since been processed only needs to happen once a day and it doesn't need to backfill since it checks all the known missing crashes.

Moving that to a separate task will reduce the runtime for verifyprocessed. Further, because verifyprocessed *is* a backfill task, if it needs to run for a bunch of days, it doesn't repeat the unnecessary work of updatemissing.

### DIFF
--- a/webapp-django/crashstats/crashstats/management/commands/updatemissing.py
+++ b/webapp-django/crashstats/crashstats/management/commands/updatemissing.py
@@ -6,8 +6,6 @@
 This command checks known missing crashes to see if they've since been processed.
 """
 
-import markus
-
 from django.core.management.base import BaseCommand
 
 from crashstats.crashstats.configman_utils import get_s3_context
@@ -17,13 +15,6 @@ from crashstats.crashstats.management.commands.verifyprocessed import (
 )
 from crashstats.supersearch.models import SuperSearchUnredacted
 from crashstats.crashstats.models import MissingProcessedCrash
-
-
-# Number of seconds until we decide a worker has stalled
-WORKER_TIMEOUT = 10 * 60
-
-
-metrics = markus.get_metrics("cron.verifyprocessed")
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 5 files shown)


webapp-django/crashstats/crashstats/management/commands/updatemissing.py
```diff
@@ -0,0 +1,65 @@
...
+from crashstats.crashstats.management.commands.verifyprocessed import (
+    is_in_s3,
+    is_in_elasticsearch,
+)
+from crashstats.supersearch.models import SuperSearchUnredacted
+from crashstats.crashstats.models import MissingProcessedCrash
+
+
+# Number of seconds until we decide a worker has stalled
+WORKER_TIMEOUT = 10 * 60
```
willkg: Oops--this doesn't have workers, so I can remove this.

webapp-django/crashstats/crashstats/management/commands/updatemissing.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+metrics = markus.get_metrics("cron.verifyprocessed")
```
willkg: This doesn't do any metrics, so I can remove this, too.

webapp-django/crashstats/crashstats/management/commands/updatemissing.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class Command(BaseCommand):
+    help = "Check known missing crashes to see if they've been processed."
+
+    def check_past_missing(self):
```
willkg: This was taken verbatim from verifyprocessed. It works the same.

webapp-django/crashstats/crashstats/tests/test_updatemissing.py
```diff
@@ -0,0 +1,79 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
willkg: These tests come from `test_verifyprocessed.py` and are otherwise unaltered.

webapp-django/crashstats/cron/__init__.py
```diff
@@ -48,6 +48,12 @@
         "time": "04:00",
         "backfill": True,
     },
+    {
+        # Check past missing crashes and update status daily at 3:00am
+        "cmd": "updatemissing",
+        "frequency": "1d",
+        "time": "03:00",
```
willkg: Running it at 3:00am is sort of an arbitrary time. We could run it afterwards. I'm not sure it matters a whole lot.

willkg:
I'll test this further on stage. self-r+

PR #5062, opened 2019-12-13
bug 1603308: fix cronrun to save place after every run_time

willkg:
If cronrun is working on a backfill job that has a bunch of run_times to get through, then it'll go through all those run_times and if everything is successful, it'll update the "last_success" field for that job.

However, since the run_times for a backfilled job start with the "last_success" datetime, if a job fails half-way through, it'll start all over again at the beginning. Ugh.

This fixes it so that cronrun saves its place after every successful run_time. If it fails (ran out of time, some random problem, whatever), it'll start the run_times at the last successful run_time. This will
drastically improve the situation where a backfill job has to catch up from many failures.

### DIFF
--- a/webapp-django/crashstats/cron/management/commands/cronrun.py
+++ b/webapp-django/crashstats/cron/management/commands/cronrun.py
@@ -117,8 +117,6 @@ class Command(BaseCommand):
         logger.info("about to run %s", cmd)
 
         now = timezone.now()
-        log_run = True
-        exc_type = exc_value = exc_tb = None
         start_time = None
         run_time = None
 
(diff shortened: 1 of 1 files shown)


webapp-django/crashstats/cron/management/commands/cronrun.py
```diff
@@ -117,8 +117,6 @@ def _run_one(self, job_spec, force=False, cmd_args=None):
         logger.info("about to run %s", cmd)
 
         now = timezone.now()
-        log_run = True
-        exc_type = exc_value = exc_tb = None
```
willkg: We don't need any of these since we'll now call `self._log_run` after every successful run_time and in the event of failure. We no longer call it in the `finally` block after everything has happened.

willkg:
self-r+

PR #5063, opened 2019-12-13
bug 1603308: adjust verifyprocessed entropy chunk size and timeout

willkg:
In prod, it's taking too long for a worker to work through an entropy chunk. This lowers the chunk size from 5 to 4 and increases the timeout from 10 minutes to 15 minutes. The hope is that this gives workers enough time to finish an entropy chunk and not timeout.


willkg:
I'll be able to see any adverse change in timings on stage and can tweak accordingly. Once things look ok on stage, I'll push this to prod. self-r+

willkg:
I'll be able to see any adverse change in timings on stage and can tweak accordingly. Once things look ok on stage, I'll push this to prod. self-r+

willkg:
I'll be able to see any adverse change in timings on stage and can tweak accordingly. Once things look ok on stage, I'll push this to prod. self-r+

PR #5064, opened 2019-12-13
bug 1603308: adjust Elasticsearch check to do it in batches

willkg:
We can search for a bunch of items in Elasticsearch at the same time rather than one at a time. This adjusts the code to batch Elasticsearch lookups.

### DIFF
--- a/webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
+++ b/webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
@@ -55,25 +55,29 @@ def is_in_s3(s3_client, bucket, crash_id):
     return True
 
 
-def is_in_elasticsearch(supersearch, crash_id):
-    """Is the processed crash in Elasticsearch."""
-    crash_date = date_from_ooid(crash_id)
+def check_elasticsearch(supersearch, crash_ids):
+    """Checks Elasticsearch and returns list of missing crash ids.
+
+    Crash ids should all be on the same day.
+
+    """
+    crash_ids = [crash_ids] if isinstance(crash_ids, str) else crash_ids
+    crash_date = date_from_ooid(crash_ids[0])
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
```diff
@@ -92,16 +96,25 @@ def check_crashids(entropy_chunk, date):
...
         for page in page_iterator:
-            for item in page.get("Contents", []):
-                raw_crash_key = item["Key"]
-                crash_id = raw_crash_key.split("/")[-1]
+            # NOTE(willkg): Keys look like /v2/raw_crash/ENTRPOY/DATE/CRASHID
+            crash_ids = [
+                item["Key"].split("/")[-1] for item in page.get("Contents", [])
+            ]
+
+            if not crash_ids:
```
willkg: I don't understand how it's possible to get a page that has no items in it. I would think the page_iterator would be smarter than to have us request an empty page. It's weird.

webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
```diff
@@ -92,16 +96,25 @@ def check_crashids(entropy_chunk, date):
...
 
-                elif not is_in_elasticsearch(supersearch, crash_id):
-                    missing.append(crash_id)
+            # Check Elasticsearch in batches
+            for crash_ids_batch in chunked(crash_ids, 100):
+                missing_in_es = check_elasticsearch(supersearch, crash_ids_batch)
+                missing.extend(missing_in_es)
 
-    return missing
+    return list(set(missing))
```
willkg: Because we're appending to `missing` for both the S3 and ES checks, it's possible for a crash id to be in there twice. So we turn it into a set to drop duplicates and back to a list to send back to the main process.

willkg:
self-r+

PR #5065, opened 2019-12-16
bug 1604317: remove missing processed check from admin

willkg:
This removes the check to see if a missing processed crash is still
missing from the Django admin page. The check only checked S3 and not
elasticsearch, so it wasn't right. Further, it took FOREVER to run
making the Django admin page super slow.


willkg:
This only affects the django admin--it doesn't affect anything else. self-r+

PR #5066, opened 2019-12-17
Add a convenient shell for the webapp

willkg:
The shell includes readline completion, all the models, and some other
things in the space. Makes some debugging easier.


willkg:
I think I threw this together for @jwhitlock or he threw it together for me. Someone wrote it. I figure it's easier to just check it in at this point.

willkg:
It can only be used in development, so self-r+.

PR #5069, opened 2019-12-18
bug 1602121: add queue consuming and publishing for AWS SQS

willkg:
This adds the bits to Socorro to consume from and publish to AWS SQS. This adds tests and some configuration, too.

The default continues to be Google Pub/Sub. This makes no changes to how things run in production. After this lands, we'll figure out the migration plan.

Next steps on this:

1. I need to self-review it.
2. I need to set this up with real AWS SQS queues and Antenna and make sure the whole thing works.

### DIFF
--- a/socorro/unittest/external/sqs/test_crashqueue.py
+++ b/socorro/unittest/external/sqs/test_crashqueue.py
@@ -8,10 +8,7 @@ import pytest
 
 from socorro.external.sqs.crashqueue import SQSCrashQueue
 from socorro.lib.ooid import create_new_ooid
-from socorro.unittest.external.sqs import get_sqs_config
-
-
-VISIBILITY_DEADLINE = 2
+from socorro.unittest.external.sqs import get_sqs_config, VISIBILITY_TIMEOUT
 
 
 class TestSQSCrashQueue:
@@ -61,8 +58,7 @@ class TestSQSCrashQueue:
         for args, kwargs in new_crashes:
             kwargs["finished_func"]()
 
-        # Wait beyond the ack deadline in the grossest way possible
-        time.sleep(VISIBILITY_DEADLINE + 1)
+        time.sleep(VISIBILITY_TIMEOUT + 1)
 
         # Now call it again and make sure we get nothing back
         new_crashes = list(crash_queue.new_crashes())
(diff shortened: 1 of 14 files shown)


socorro/unittest/external/sqs/test_crashqueue.py
```diff
@@ -0,0 +1,94 @@
...
+        # nor have the leases expired
+        second_new_crashes = list(crash_queue.new_crashes())
+        assert second_new_crashes == []
+
+        # Now ack the crash_id and we don't get it again
+        for args, kwargs in new_crashes:
+            kwargs["finished_func"]()
+
+        # Wait beyond the ack deadline in the grossest way possible
+        time.sleep(VISIBILITY_DEADLINE + 1)
```
willkg: This isn't right. This is a visibility deadline that's not related to the default visibility for the queue.

willkg:
I was able to set everything up using localstack-sqs and got that working. Then I set up three standard queues in AWS SQS and (trivially) got it working with those.

I want to fix the webapp so the crashqueue class it's using is configurable. Then the entire migration is configuration driven and we can back it out more easily.

willkg:
To test against localstack-sqs, add this to my.env:

```
# Configure webapp and processor to use localstack-sqs
queue.crashqueue_class=socorro.external.sqs.crashqueue.SQSCrashQueue

# Configure collector service to use localstack-sqs
CRASHPUBLISH_CLASS=antenna.ext.sqs.crashpublish.SQSCrashPublish
CRASHPUBLISH_ACCESS_KEY=foo
CRASHPUBLISH_SECRET_ACCESS_KEY=foo
CRASHPUBLISH_REGION=us-west-2
CRASHPUBLISH_QUEUE_NAME=local_dev_standard
CRASHPUBLISH_ENDPOINT_URL=http://localstack-sqs:4576/
```

That matches the configuration in `docker/config/local_dev.env`.

After landing this, I'll work on the migration documentation.

webapp-django/crashstats/settings/base.py
```diff
@@ -567,6 +567,12 @@ def filter(self, record):
     implementations_database_url = database_url
 implementations_config = dj_database_url.parse(implementations_database_url)
 
+# The CrashQueueBase class to use for submitting priority and reprocessing
+# requests
+CRASHQUEUE = config(
+    "queue.crashqueue_class", "socorro.external.pubsub.crashqueue.PubSubCrashQueue"
+)
+
```
willkg: This continues the trend of having the webapp use processor environment configuration variables. We can't do it the other way around--configman doesn't work like that. Not having duplicated values makes life easier.

At some point in the future, we'll switch from configman to something else and I'm hoping some of these problems will go away then. We'll see.

willkg:
I did a minor change to how the webapp loads the crash queue class. That will affect server environments. Everything else in here won't because it's not enabled.

After landing this, I'll keep an eye on stage to make sure it's still working.

self-r+

PR #5070, opened 2019-12-23
bug 1605716: add processor.process_crash timing

willkg:
This adds a new timing `processor.process_crash` that covers the length of time it takes to download the raw crash data, process it, and save the processed crash.


willkg:
self-r+

PR #5071, opened 2019-12-31
bug 1303758: remove cpu_name

willkg:
This finishes the transition from "cpu_name" to "cpu_arch". The cpu_arch
field holds the architecture the binaries were built for.

To test:

1. process some crashes
2. verify the "Build Architecture" and "CPU Count" fields have correct values in the report view
3. verify the bugzilla links are built properly
4. verify that the `cpu_arch` and `cpu_count` fields can be searched and faceted on


willkg:
Past me left present me decent instructions for doing this. I tested it in my local dev environment. We should land this and make sure stage looks ok and builds a new index correctly.

self-r+

PR #5072, opened 2019-12-31
bug 1503344: remove java_stack_trace_full

willkg:
This finishes the transition from `java_stack_trace_full` (which was poorly
named and caused problems in the Elasticsearch crash storage code) to
`java_stack_trace_raw` (which doesn't).


willkg:
self-r+

PR #5073, opened 2019-12-31
bug 1544427: change --persist to --dry-run in auditgroups

willkg:
This makes auditgroups consistent with other commands.

### DIFF
--- a/webapp-django/crashstats/authentication/management/commands/auditgroups.py
+++ b/webapp-django/crashstats/authentication/management/commands/auditgroups.py
@@ -42,10 +42,8 @@ class Command(BaseCommand):
     help = "Audits Django groups and removes inactive users."
 
     def add_arguments(self, parser):
-        # FIXME(willkg): change this to default to False after we've tested
-        # it.
         parser.add_argument(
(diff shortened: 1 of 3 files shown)


willkg:
self-r+

webapp-django/crashstats/authentication/management/commands/auditgroups.py
```diff
@@ -42,10 +42,8 @@ class Command(BaseCommand):
     help = "Audits Django groups and removes inactive users."
 
     def add_arguments(self, parser):
-        # FIXME(willkg): change this to default to False after we've tested
-        # it.
         parser.add_argument(
-            "--persist", action="store_true", help="persists recommended changes to db"
+            "--dry-run", action="store_true", help="Whether or not to do a dry run."
```
willkg: Oops--this should be a `store_false`. Bah.

willkg: My bad--I copy-and-pasted this from the other command, so this is fine. Bah to the bah.

PR #5074, opened 2019-12-31
bug 1606450: truncate long string values for Elasticsearch

willkg:
Elasticsearch gets sad when crash reports contain values which have more than 32,766 utf-8 encoded characters. This adds some code to truncate those fields before saving to Elasticsearch.

This fixes saving `java_stack_trace` and `java_stack_trace_raw` for StackOverflow errors which can get up to 100k characters long and make Elasticsearch sad.

To test:

1. run the processor
2. process a crash report for a Java Stack Overflow error like 844d5678-e15f-4974-bcf1-d51d90191230
3. watch the logs and make sure it saves in Elasticsearch without kicking up a maxbyteslengthexceeded error

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -17,7 +17,10 @@ from socorro.lib.datetimeutil import JsonDTEncoder, string_to_datetime
 
 
 # Maximum size in characters for a keyword field value
-MAX_KEYWORD_FIELD_VALUE_SIZE = 10000
+MAX_KEYWORD_FIELD_VALUE_SIZE = 10_000
+
+# Maximum size in utf-8 encoded characters for a string field value
+MAX_STRING_FIELD_VALUE_SIZE = 32_766
 
 
 def reconstitute_datetimes(processed_crash):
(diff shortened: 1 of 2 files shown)


socorro/external/es/crashstorage.py
```diff
@@ -17,7 +17,10 @@
 
 
 # Maximum size in characters for a keyword field value
-MAX_KEYWORD_FIELD_VALUE_SIZE = 10000
+MAX_KEYWORD_FIELD_VALUE_SIZE = 10_000
```
willkg: I switched to using `_` to split groups. It makes it easier to read.

socorro/external/es/crashstorage.py
```diff
@@ -17,7 +17,10 @@
...
+
+# Maximum size in utf-8 encoded characters for a string field value
+MAX_STRING_FIELD_VALUE_SIZE = 32_766
```
willkg: This number comes from Elasticsearch. When we upgrade to a newer version of Elasticsearch, we can see what the maximum size for a term is and adjust this accordingly.

socorro/external/es/crashstorage.py
```diff
@@ -158,8 +160,43 @@ def truncate_keyword_field_values(fields, data):
...
+
+    :arg dict fields: the super search fields schema
+    :arg dict data: the data to look through
+
+    """
+    string_fields = [
+        field
+        for field in fields.values()
+        if (field.get("storage_mapping") or {}).get("type", "") == "string"
+    ]
```
willkg: This will pick up keyword values, too. I didn't see a good way of building a set of non-overlapping fields we want to operate on with what we have now. When we redo how Elasticsearch crash storage works, we can clean this whole thing up. Until then, this is a hack-ish thing.

socorro/external/es/crashstorage.py
```diff
@@ -158,8 +160,43 @@ def truncate_keyword_field_values(fields, data):
...
+        new_value = value
+
+        # First truncate down to MAX_STRING_FIELD_VALUE_SIZE
+        if len(new_value.encode("utf-8")) > MAX_STRING_FIELD_VALUE_SIZE:
+            new_value = new_value[:MAX_STRING_FIELD_VALUE_SIZE]
+
+        # If the utf-8 encoded bytes is still larger, whittle off unicode
+        # characters until it fits
+        while len(new_value.encode("utf-8")) > MAX_STRING_FIELD_VALUE_SIZE:
+            new_value = new_value[:-1]
```
willkg: We do it this way so that we don't remove a byte and break a unicode character.

willkg:
self-r+

PR #5075, opened 2019-12-31
bug 1581567: validate crash ids in /api/Reprocessing

willkg:
This fixes `/api/Reprocessing` to validate crash ids before submitting them. This reduces failures in the processor when it gets crash ids that aren't crash ids.

To test:

Use the /api/Reprocessing API and send in an invalid crash id. Easiest thing to do is test it with the API reference docs.

http://localhost:8000/api/#Reprocessing


willkg:
self-r+

PR #5076, opened 2019-12-31
bug 1597730: remove quit check

willkg:
We used to do deploys by stopping the processor service, updating the code, and then starting it again. That was a long time ago. Now we create a bunch of new processor nodes and then kill off the old processor nodes. Because of that, we no longer need the quit check code that allows the processor to shut down nicely.

To test the local dev environment:

1. run the processor
2. process a crash
3. hit ctrl-c and shut down the processor


willkg:
self-r+

PR #5078, opened 2020-01-03
bug 1552976: fix correlations tab

willkg:
When the report view page is loaded, the correlations tab is hidden. JavaScript in the page goes to fetch correlation data and if it gets some, then it unhides the tab so as to show correlations data. The
problem here is that unhiding the tab causes the tabs to the right to jump.

This changes the code so that the tab is always shown and when there's no correlations data, it states there's no correlations data. This fixes the UI jumping which keeps biting me. It also reduces confusion around why the tab isn't showing up in some cases.

While doing this, I also unhid the Telemetry Environment tab. That will now show all the time and when there's no TelemetryEnvironment data, it'll state there's no data.

### DIFF
--- a/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
+++ b/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
@@ -77,12 +77,10 @@
             <li class="ui-state-default ui-corner-top">
               <a href="#extensions" class="ui-tabs-anchor"><span>Extensions</span></a>
             </li>
-            {% if raw.TelemetryEnvironment %}
-              <li class="ui-state-default ui-corner-top">
-                <a href="#telemetryenvironment" class="ui-tabs-anchor"><span>Telemetry Environment</span></a>
-              </li>
-            {% endif %}
-            <li class="ui-state-default ui-corner-top correlations hidden">
+            <li class="ui-state-default ui-corner-top">
+              <a href="#telemetryenvironment" class="ui-tabs-anchor"><span>Telemetry Environment</span></a>
+            </li>
+            <li class="ui-state-default ui-corner-top">
               <a href="#correlation" class="ui-tabs-anchor"><span>Correlations</span></a>
             </li>
             {% if request.user.has_perm('crashstats.reprocess_crashes') %}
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -77,12 +77,10 @@ <h2>{{ report.product }} {{ report.version }} Crash Report [@ {{ report.signatur
...
             </li>
-            {% if raw.TelemetryEnvironment %}
-              <li class="ui-state-default ui-corner-top">
-                <a href="#telemetryenvironment" class="ui-tabs-anchor"><span>Telemetry Environment</span></a>
-              </li>
-            {% endif %}
-            <li class="ui-state-default ui-corner-top correlations hidden">
+            <li class="ui-state-default ui-corner-top">
+              <a href="#telemetryenvironment" class="ui-tabs-anchor"><span>Telemetry Environment</span></a>
+            </li>
```
willkg: Now the TelemetryEnvironment tab always shows and if there's no data, the page will say there's no data available.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -77,12 +77,10 @@ <h2>{{ report.product }} {{ report.version }} Crash Report [@ {{ report.signatur
...
+            <li class="ui-state-default ui-corner-top">
```
willkg: This changes the correlations tab to always show. If there's no correlations data, then the page will say there's no data available.

willkg:
To test this, do the following:

1. Process a Firefox crash report that has correlations. For example, this crash report: debc96f3-4f50-413c-baa4-1032c0200103

   View it in the report view.

   The Correlations tab should show and it should show correlations data in it.

   The TelemetryEnvironment tab should show. There should be data in it.

2. Process a Firefox crash report that doesn't have correlations. For example, this report: 39c51d41-b1b0-4e61-99f3-fd70d0200103

   View it in the report view.

   The Correlations tab should show and it says there's no correlations data.

   The TelemetryEnvironment tab should show. There should be data in it.

3. Process a Fenix crash report--these don't have correlations data. For example, this report: aba2c0b1-8f40-405b-93cc-6985e0200103

   View it in the report view.

   The Correlations tab should show (this is the change), and it should state there's no correlations data available.

   The TelemetryEnvironment tab should show and it should state there's no Telemetry environment data available.

willkg:
The code changes are pretty small. I was worried there might be related bits elsewhere in the codebase, but I didn't find anything else. I'm pretty sure I got everything. Local manual testing was successful. We don't have any UI automated testing. I think that's the best we can do. If there are problems, we can fix them going forward.

self-r+

PR #5079, opened 2020-01-03
bug 1605806: fix language around requesting pii access

willkg:
Previously, the documentation only talked about memory dumps, but the process for requesting access covers both memory dumps and private user data.

This fixes the language so that's clearer.


willkg:
self-r+

PR #5080, opened 2020-01-03
bug 1543767: fix topcrashers summary line

willkg:
The topcrashers summary line was confusing because the math was being done against a "total_number_of_crashes" that was not the total number of crashes but instead the user-provided number of signatures to look at. This fixes everything by pulling the total number from the search
results which is the total number of crash reports that match the given filters.

To test:

1. process like 1000 crashes--enough to get a Firefox version to look at that has more than 50 crash reports
2. look at the top crashers page and make sure it's something like this:

   ```
   The report covers 259 (58.07%) of all 446 crashes during this period. 
   ```

   and not something like this:

   ```
   The report covers 4277 (8554.0%) of all 50 crashes during this period. 
   ```

### DIFF
--- a/webapp-django/crashstats/topcrashers/views.py
+++ b/webapp-django/crashstats/topcrashers/views.py
@@ -27,7 +27,7 @@ def datetime_to_build_id(date):
 
 
 def get_topcrashers_stats(**kwargs):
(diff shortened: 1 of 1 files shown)


willkg:
self-r+

PR #5081, opened 2020-01-03
bug 1605429: fix stackwalker to work with ModuleSignatureInfo as object

willkg:
Recently, the crash reporter was changed so it sends a JSON blob with the annotations in it. When that happened, the type of the `ModuleSignatureInfo` value changed from a JSON-encoded string to an
object. stackwalker expected a JSON-encoded string, so if it was an object, it'd ignore it. Thus it didn't capture the signed-by information for modules.

This fixes that by supporting both JSON-encoded string and `Json::Value` values of `ModuleSignatureInfo`.

### DIFF
--- a/minidump-stackwalk/stackwalker.cc
+++ b/minidump-stackwalk/stackwalker.cc
@@ -603,14 +603,20 @@ static ModuleCertMap GetModuleCertMap(const Json::Value& raw_root)
   ModuleCertMap result;
 
   const Json::Value& infoString = raw_root["ModuleSignatureInfo"];
-  if (!infoString || !infoString.isString()) {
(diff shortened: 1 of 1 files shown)


willkg:
To test, use these two crash reports:

* a2cf2e3a-f2ba-4dfd-be8f-bdcf40191220: the `ModuleSignatureInfo` is an object
* 24c81bb4-fe10-4766-a3f1-b675f0200103: the `ModuleSignatureInfo` is a JSON-encoded string

After processing both, you should see a "Signed by" column in the modules tab of the report view. If you look at prod, only the second one has a "Signed by" column.

willkg:
self-r+

PR #5082, opened 2020-01-03
bug 1581927: fix cronrun to run jobs despite OngoingJobError

willkg:
Previously, if `cronrun` hit an `OngoingJobError`, it would stop running any jobs. This fixes it to continue through the list of jobs that are ready to run even if one of them throws an `OngoingJobError`.


willkg:
self-r+

PR #5083, opened 2020-01-07
bug 1607369: switch to dependabot and remove pyup configuration

willkg:


willkg:
Shouldn't land this until Greg has removed socorro from pyup, lest we get a slew of pyup PRs.

willkg:
Greg disabled socorro in pyup, so this is good to go. self-r+

PR #5146, opened 2020-01-08
bug 1607519: fix Caused By handling in java stack trace parser

willkg:
Previously, we only saw "Caused By:" lines that were indented by one tab. Now we're seeing them unindented which is kicking up a `MalformedJavaStackTrace` exception.

This fixes the parsing to handle both situations.

### DIFF
--- a/socorro/lib/javautil.py
+++ b/socorro/lib/javautil.py
@@ -98,11 +98,7 @@ def parse_java_stack_trace(text):
                 stage = ADDITIONAL
 
         elif stage is ADDITIONAL:
-            # Verify this has a tab at the beginning of the line
-            if line[0] != "\t":
-                raise MalformedJavaStackTrace("additional line missing tab")
-
-            # Drop first tab from the line
-            new_exc.additional.append(line[1:])
+            # Append all the rest of the stack trace as is
+            new_exc.additional.append(line)
 
     return new_exc
--- a/socorro/unittest/lib/test_javautil.py
+++ b/socorro/unittest/lib/test_javautil.py
@@ -100,8 +100,8 @@ def test_parse_suppressed():
     assert java_exc.exception_message == "msg"
     assert java_exc.stack == ["at org.File.function(File.java:100)"]
     assert java_exc.additional == [
-        "Suppressed: Exception2: msg2",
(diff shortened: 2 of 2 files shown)


socorro/lib/javautil.py
```diff
@@ -98,11 +98,7 @@ def parse_java_stack_trace(text):
                 stage = ADDITIONAL
 
         elif stage is ADDITIONAL:
-            # Verify this has a tab at the beginning of the line
-            if line[0] != "\t":
-                raise MalformedJavaStackTrace("additional line missing tab")
```
willkg: I took out the "does it start with a tab?" bits. The result is that the additional section is closer to the original plus it doesn't enforce the "starts with a tab" thing which isn't always the case and is causing `MalformedJavaStackTrace` exceptions now.

socorro/unittest/lib/test_javautil.py
```diff
@@ -100,8 +100,8 @@ def test_parse_suppressed():
     assert java_exc.exception_message == "msg"
     assert java_exc.stack == ["at org.File.function(File.java:100)"]
     assert java_exc.additional == [
-        "Suppressed: Exception2: msg2",
-        "\tat org.File.function(File.java:101)",
+        "\tSuppressed: Exception2: msg2",
+        "\t\tat org.File.function(File.java:101)",
```
willkg: Because we took out the "let's remove the first tab from additional bits", we needed to update the existing tests.

socorro/unittest/lib/test_javautil.py
```diff
@@ -131,13 +149,6 @@ def test_parse_caused_by():
         "",
         # Line without a tab in STACK stage
         ("Exception: msg\n" "\tat org.File.function(File.java:100)\n" "badline"),
-        # Line without a tab in ADDITIONAL stage
-        (
-            "Exception: msg\n"
-            "\tat org.File.function(File.java:100)\n"
-            "\tSuppressed: Exception: msg\n"
-            "badline"
-        ),
```
willkg: This was a malformed case because `badline` didn't have a tab at the beginning of the line. That's no longer something we consider malformed, so I nixed this test case.

willkg:
self-r+

PR #5148, opened 2020-01-08
bug 1607806: make ModuleSignatureInfo always a JSON-encoded string

willkg:
This fixes `raw_crash["ModuleSignatureInfo"]` to always have a JSON-encoded string value so the rest of the processor doesn't have to handle both cases.


willkg:
self-r+

PR #5149, opened 2020-01-09
bug 1607806: fix converting ModuleSignatureInfo values

willkg:
`raw_crash["ModuleSignatureInfo"]` is a `DotDict` which doesn't work with `json.dumps`. This fixes the code to convert it to a dict and then `json.dumps` it.

### DIFF
--- a/socorro/lib/util.py
+++ b/socorro/lib/util.py
@@ -19,7 +19,7 @@ def dotdict_to_dict(sdotdict):
 
     def _dictify(thing):
         if isinstance(thing, collections.Mapping):
-            return dict([(key, _dictify(val)) for key, val in thing.items()])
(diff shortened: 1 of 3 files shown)


socorro/lib/util.py
```diff
@@ -19,7 +19,7 @@ def dotdict_to_dict(sdotdict):
 
     def _dictify(thing):
         if isinstance(thing, collections.Mapping):
-            return dict([(key, _dictify(val)) for key, val in thing.items()])
+            return {key: _dictify(val) for key, val in thing.items()}
```
willkg: I did this while I was looking at this code. It's not really related to the rest of the PR.

willkg:
self-r+

PR #5165, opened 2020-01-15
bug 1608111: initial crash report v0 specification

willkg:
This adds a specification for the format of crash reports submitted to Socorro. It's specific to what Socorro accepts and how it accepts it.

This is a v0 specification. In Firefox 73, we added support for a `extra` field which has a JSON-encoded value covering all the crash report annotations. I want to cover that, too, but I think that's crash report v1 and I'll write that up after this one is approved and merged.

There are a few other things that might be interesting to cover:

1. important-to-include annotations and what they mean: ProductName, ReleaseChannel, Version, ...
2. links to documentation on minidump structure

What other things should get covered here?

### DIFF
--- a/docs/spec_crashreport.rst
+++ b/docs/spec_crashreport.rst
@@ -1,0 +1,218 @@
+===========================
+Specification: Crash report
+===========================
+
+This is a specification for the format for submitting crash reports.
+
+.. contents::
+   :local:
+
+crash report: v0
+================
+
+v0 refers to whatever we had prior to when we moved crash annotations into a
+single JSON-encoded field in `bug 1420363
+<https://bugzilla.mozilla.org/show_bug.cgi?id=1420363>`_. That work landed in
+December 2019 and is in Firefox 73.
+
+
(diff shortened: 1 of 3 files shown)


docs/spec_crashreport.rst
```diff
@@ -0,0 +1,218 @@
...
+------------------------------
+
+The content type header of the crash report must be ``multipart/form-data``
+and specify the multipart/form-data boundary.
+
+The content length of the crash report must be set. The value is the length
+of the body.
+
+If the HTTP body is gzip compressed, then the ``Content-Encoding`` must be
+set to ``gzip``.
```
gabrielesvelto: nit: You're mentioning this in the following paragraph

willkg: I tweaked it a bit. This sentence is about headers. The second sentence below is about the body. I'm trying to keep the sections self-contained even if that causes some redundancy in the whole document. Makes it easier for future me to remember all the bits.

Thank you!

PR #5167, opened 2020-01-17
bug 1608111: add single JSON-encoded value bits to specification

willkg:
Originally, I was thinking the JSON-encoded value bits constitute a new version of the specification. However, since we decided that the collector would accept crash reports and the processor would fix them, I don't think it's necessary to have a new version of the specification.

Thus, this updates the specification with the new bits plus it adds a very minimal "debugging" section.

### DIFF
--- a/docs/spec_crashreport.rst
+++ b/docs/spec_crashreport.rst
@@ -7,13 +7,14 @@ This is a specification for the format for submitting crash reports.
 .. contents::
    :local:
 
-crash report: v0
-================
 
-v0 refers to whatever we had prior to when we moved crash annotations into a
-single JSON-encoded field in `bug 1420363
-<https://bugzilla.mozilla.org/show_bug.cgi?id=1420363>`_. That work landed in
-December 2019 and is in Firefox 73.
+crash report specification
+==========================
+
+History:
+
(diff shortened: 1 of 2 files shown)


docs/spec_crashreport.rst
```diff
@@ -69,6 +70,45 @@ Example::
...
+2. The ``Content-Disposition`` must specify a ``name``. The value should
+   be ``extra``. The value must be in ASCII.
+
+3. The ``Content-Type`` must be ``application/json``.
+
+4. The value of this field must a JSON-encoded string of crash report
+   annotations.
+
+   Each annotation value must be a string. If the value is an object, it must
+   be JSON-encoded.
```
gabrielesvelto: It's hard to express this one in a way that's easy to understand. Maybe something like:
```
Annotations cannot contain JSON-encoded objects directly, they must be escaped so that they're stored as a single string.
```
I'm not sure if it's better though :-/

willkg:
@gabrielesvelto: ^^^ That change adds an example and tries to clarify how to treat annotation values that are objects. Is that clearer?

willkg:
Thank you!

PR #5169, opened 2020-01-21
bug 1610249: switch . to : in chown calls

willkg:
While GNU coreutils chown supports `.` (albeit with a bunch of caveats), we should update all uses of `.` to `:` because that's the current way to separate uid and gid.


willkg:
self-r+

PR #5170, opened 2020-01-23
bug 1610520: listing and removing data from S3 and Elasticsearch

willkg:
To test:

1. process a bunch of crash reports for 2020-01-01
2. run:
   ```
   python 1610520/es_tcid.py > crashids.txt
   ```
   to generate a list of crash ids
3. run: 
   ```
   python 1610520/remove_field.py --parallel crashids.txt TelemetryClientId TelemetryServerURL
   ```
4. run:
   ```
   python 1610520/remove_field.py --parallel crashids.txt TelemetryClientId TelemetryServerURL
   ```
   again--it'll tell you the things were fixed already

### DIFF
--- a/1610520/es_tcid.py
+++ b/1610520/es_tcid.py
@@ -1,0 +1,64 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""
+Spits out crash ids for all crash reports in Elasticsearch that have a
+TelemetryClientId field.
+
+Usage:
+
+    python 1610520/es_tcid.py > crashids.txt
+
+It prints some lines with a "#" to make it easier to see what it did. To
+remove those, do:
+
+    python 1610520/es_tcid.py | grep -v "#" > crashids.txt
+
+"""
+
+from configman import ConfigurationManager
+from configman.environment import environment
+from elasticsearch_dsl import Search
+
+from socorro.external.es.connection_context import ConnectionContext
+
+
+def get_conn():
+    # Create a configuration manager that will only check the environment for
+    # configuration and not command line parameters
+
+    cm = ConfigurationManager(
+        ConnectionContext.get_required_config(), values_source_list=[environment]
+    )
+    config = cm.get_config()
+    return ConnectionContext(config)
+
+
+def cmd_list_crashids():
+    es_conn = get_conn()
+    indices = es_conn.get_indices()
+
+    print("# %s indexes." % len(indices))
+    print("# %r" % indices)
+    total = 0
+    for index in indices:
+        print("# working on %s..." % index)
+        with es_conn() as conn:
+            search = Search(using=conn, index=index, doc_type=es_conn.get_doctype())
+            search = search.filter("exists", field="TelemetryClientId")
+            search = search.fields(["processed_crash.uuid"])
+            results = search.scan()
+            for hit in results:
+                print(hit["processed_crash.uuid"][0])
+                total += 1
+
+    print("# total found %d" % total)
+
+
+def main():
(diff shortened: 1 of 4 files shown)


willkg:
In testing those scripts, I double-checked the following:

1. When a crash report is processed, we save raw and crash data as a document in Elasticsearch. When I pull that document out of Elasticsearch, is it the same as the document we put in?

   Yes, the `_source` field has the original document with the one difference being the timestamp values. The ones from Elasticsearch have timezone information.

2. When we save the adjusted document back to Elasticsearch, is it still like the document we put in?

   Yes and the only difference is that the document in Elasticsearch no longer includes the `TelemetryClientId` and `TelemetryServerURL`.

It's hard to verify this in stage since we don't have the original data before the change.

@jwhitlock Can you review these scripts? Do they look sane to you?

1610520/es_tcid.py
```diff
@@ -0,0 +1,68 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""
+Spits out crash ids for all crash reports in Elasticsearch that have a
+TelemetryClientId field.
+
+Usage:
+
+    python 1610520/es_tcid.py > crashids.txt
```
jwhitlock: I'd prefer a path like ``scripts/es_tcid_1610520.py``. I'm guessing you intend to delete the script in a week or so, which is why the path is like it is.

1610520/remove_field.py
```diff
@@ -0,0 +1,214 @@
...
+    data = json.loads(raw_crash_as_string)
+    should_save = False
+    for field in fields:
+        if field in data:
+            del data[field]
+            should_save = True
+
+    if should_save:
+        s3_client.upload_fileobj(
+            Fileobj=io.BytesIO(dict_to_str(data).encode("utf-8")), Bucket=bucket, Key=path
```
jwhitlock: I learned this week that `utf-8` is the default for `.encode()`. I'm trying to decide if 2020 is the year to use the default, or to keep it explicit.

1610520/remove_field.py
```diff
@@ -0,0 +1,214 @@
...
+
+        # Fix the data in S3 and then Elasticsearch
+        try:
+            fix_data_in_s3(fields, bucket, s3_client, crashid)
+            fix_data_in_es(fields, es_conn, crashid)
+        except Exception:
+            # If this throws an exception, print it out and move on. Then we'll finish
+            # all the fixing for the first pass and can address the problematic crash
+            # reports in a second pass.
+            logger.exception("COUGH")
```
jwhitlock: I'd include the crash ID in the exception message, to avoid confusion if stdout and stderr print in a strange order. However, if you know more about the logger config, that may be unnecessary.

1610520/es_tcid.py
willkg: This brings back scars of finding and deleting scripts that had no documentation and no clear reason what they were doing or why or what they fixed. It took a long time to remove them. They were everywhere in the codebase. We probably still have a bunch.

Thus my scars compelled me to make sure these scripts were clearly connected to that bug and in a directory by themselves and not mixed in with everything else.

How about `scripts/1610520/es_tcid.py`? Would that be ok?

1610520/remove_field.py
willkg: Interesting! I think that changed with Python 3. I remember the default being whatever the encoding in the environment was set to and one could never rely on anything and it sucked.

1610520/remove_field.py
willkg: I like that idea. I'll change that.

1610520/es_tcid.py
jwhitlock: That's better, if you intend to keep the script around forever.

willkg:
Thank you!

PR #5219, opened 2020-02-18
bug 1616253: fix aggregation for ipc_shutdown_state

willkg:
Previously, the values were getting tokenized when indexed and that
caused problems when looking at `ipc_shutdown_state` facet. This fixes
that by switching it to the keyword analyzer which doesn't tokenize.

To test, process crash reports with `ipc_shutdown_state` values. Then use supersearch and do a facet on that. The facet options should be something like:

* `SendFinishShutdown (sent)`

and not something like:

* `sendfinishshutdown`
* `sent`


willkg:
self-r+

PR #5223, opened 2020-02-20
bug 1616194: use json_dump.lsb_release.description for platform pretty version

willkg:
For crash reports from Linux, the platform pretty version has been "Linux" which is pretty unhelpful.

This changes it so that it uses the value in `json_dump.lsb_release.description` which comes from `/etc/lsb_release` file in Linux. When this is available, it's a far more useful string like "Ubuntu 18.04 LTS".

To test:

1. process a Linux crash that has a `json_dump.lsb_release.description` value; for example bp-53090204-c74f-4041-9caf-84ee30200218
2. the platform pretty version (shows up as "OS" on the details tab of the crash report) will be "Ubuntu 18.04 LTS"

### DIFF
--- a/socorro/processor/rules/mozilla.py
+++ b/socorro/processor/rules/mozilla.py
@@ -10,6 +10,7 @@ import time
 from urllib.parse import unquote_plus
 
 from configman.dotdict import DotDict
+from glom import glom
 import markus
 
 from socorro.lib import javautil
@@ -727,9 +728,16 @@ class BetaVersionRule(Rule):
 
 
 class OSPrettyVersionRule(Rule):
-    """This rule attempts to extract the most useful, singular, human
-    understandable field for operating system version. This should always be
-    attempted.
+    """Populate os_pretty_version with most readable operating system version string.
+
+    This rule attempts to extract the most useful, singular, human understandable field
+    for operating system version. This should always be attempted.
+
+    For Windows, this is a lookup against a map.
+
+    For Mac OSX, this pulls from os_name and os_version.
+
+    For Linux, this uses json_dump.lsb_release.description if it's available.
 
     """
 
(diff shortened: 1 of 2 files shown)


socorro/processor/rules/mozilla.py
```diff
@@ -758,34 +766,41 @@ def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
 
-        if not processed_crash.get("os_version"):
+        os_version = processed_crash.get("os_version") or ""
+        if not os_version:
             # The version number is missing, there's nothing more to do.
             return True
 
-        version_split = processed_crash.os_version.split(".")
-
+        version_split = os_version.split(".")
```
willkg: While doing this, I also fixed this rule so it works with dicts and `DotDict`.

socorro/unittest/processor/rules/test_mozilla.py
```diff
@@ -1384,90 +1384,100 @@ def test_aurora_channel(self):
...
+            ("Windows NT", "15.2", "Windows Unknown"),
+            # A valid version of Mac OS X
+            ("Mac OS X", "10.18.324", "OS X 10.18"),
+            # An invalid version of Mac OS X
+            ("Mac OS X", "12.1", "OS X Unknown"),
+            # Generic Linux
+            ("Linux", "0.0.12.13", "Linux"),
+        ],
+    )
+    def test_everything_we_hoped_for(self, os_name, os_version, expected):
```
willkg: While working on this, I redid the tests so they're separate parametrized tests rather than a single test testing a bazillion different things.

socorro/unittest/processor/rules/test_mozilla.py
```diff
@@ -1384,90 +1384,100 @@ def test_aurora_channel(self):
...
+        assert processed_crash["os_pretty_version"] == expected
 
-        processed_crash = DotDict()
-        processed_crash.os_name = None
-        processed_crash.os_version = None
-
-        rule.act(raw_crash, raw_dumps, processed_crash, processor_meta)
-        assert "os_pretty_version" in processed_crash
-        assert processed_crash["os_pretty_version"] is None
+    def test_dotdict(self):
```
willkg: Since I rewrote it to use dicts, we need to make sure it works with `DotDict` still until we remove that.

socorro/unittest/processor/rules/test_mozilla.py
```diff
@@ -1384,90 +1384,100 @@ def test_aurora_channel(self):
...
-        processed_crash.os_name = "Windows NT"
-        processed_crash.os_version = "NaN"
+    def test_none(self):
+        raw_crash = copy.deepcopy(canonical_standard_raw_crash)
+        raw_dumps = {}
+        processor_meta = get_basic_processor_meta()
 
+        rule = OSPrettyVersionRule()
+
+        processed_crash = {"os_name": None, "os_version": None}
```
willkg: In the earlier tests, if `os_name` or `os_version` were `None`, then they were omitted from the `processed_crash`. So this explicitly tests what happens if the values were `None` rather than omitted.

willkg:
self-r+

PR #5226, opened 2020-02-27
Improve graphics device data flow docs

willkg:
This improves the graphics device flow docs by including information from the related admin page and also adding a link to the parsing code.

This makes it easier for other groups to do what we're doing--acquiring the graphics device lookup data and parse it.


willkg:
self-r+

PR #5265, opened 2020-03-06
bug 1619962: crash report deletion script

willkg:
This implements a script that can be run in a server environment to remove crash report data from crashstorage destinations.

Usage: `python scripts/permadelete_crash_data.py <CRASHIDSFILE>`

This has no tests, yet. I tested it manually. Need to add tests for the bug.

### DIFF
--- a/scripts/delete_crash_data.py
+++ b/scripts/delete_crash_data.py
@@ -108,7 +108,7 @@ def s3_delete(crashid):
     key = "v2/raw_crash/%(entropy)s/%(date)s/%(crashid)s" % {
         "entropy": crashid[0:3],
         "date": "20" + crashid[-6:],
-        "crashid": crashid
+        "crashid": crashid,
     }
     obj = s3_fetch_object(s3_client, bucket, key)
     if obj:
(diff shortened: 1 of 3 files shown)


scripts/delete_crash_data.py
```diff
@@ -0,0 +1,236 @@
...
+    doc_type = es_conn.get_doctype()
+
+    with es_conn() as conn:
+        try:
+            search = Search(using=conn, doc_type=doc_type)
+            search = search.filter("term", **{"processed_crash.uuid": crashid})
+            results = search.execute().to_dict()
+            hits = results["hits"]["hits"]
+            if hits:
+                return results["hits"]["hits"][0]
```
willkg: This could be `return hits[0]`.

willkg:
I tested it locally by processing a bunch of crash reports so they're all in local crash storage, then running the script a few times.

I'm going to have Brian run it in stage. If that works, I'll merge this and note in the bug to add tests later.

willkg:
Worked in stage! self-r+

PR #5266, opened 2020-03-10
Update Fennec versions for 68.6 release

jcristau:
Drop nightly version as users were migrated to Fenix.  Beta is kept at 68.6b for now as 68.7b hasn't shipped (and might never ship).


rvandermeulen:
@willkg Can you confirm that removing the Nightly version number will have the desired affect of causing "FennecAndroid 68.5a1" to be removed from https://crash-stats.mozilla.com/home/product/FennecAndroid so only the Beta and Release versions remain?

willkg:
I think it should work. It's definitely easy to test and probably easy to fix if it doesn't work out the way we want.

willkg:
Looks like it worked to me:

image

PR #5283, opened 2020-03-16
bug 1619646: add handling for UnicodeEncodeError in truncating

willkg:
The Elasticsearch crash storage code tries to be really forgiving for bad data. One of the things it does is truncate absurdly long string values. Strings are unicode, but are stored in Elasticsearch as utf-8, so the truncation code encodes as utf-8 to get the length.

This handles the situation where the string value can't be encoded as utf-8. In that case, probably best to consider it bad dta and replace it with "BAD DATA"

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -187,13 +187,18 @@ def truncate_string_field_values(fields, data):
         new_value = value
 
         # First truncate down to MAX_STRING_FIELD_VALUE_SIZE
-        if len(new_value.encode("utf-8")) > MAX_STRING_FIELD_VALUE_SIZE:
-            new_value = new_value[:MAX_STRING_FIELD_VALUE_SIZE]
+        try:
+            if len(new_value.encode("utf-8")) > MAX_STRING_FIELD_VALUE_SIZE:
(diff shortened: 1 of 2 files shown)


willkg:
To test, process b7e5b25a-a5e9-4d91-9ea3-ce7090200316.

socorro/external/es/crashstorage.py
```diff
@@ -187,13 +187,18 @@ def truncate_string_field_values(fields, data):
...
+                new_value = new_value[:MAX_STRING_FIELD_VALUE_SIZE]
+
+            # If the utf-8 encoded bytes is still larger, whittle off unicode
+            # characters until it fits
+            while len(new_value.encode("utf-8")) > MAX_STRING_FIELD_VALUE_SIZE:
+                new_value = new_value[:-1]
+        except UnicodeEncodeError:
+            # If we hit a UnicodeEncodeError converting the unicode to utf-8, then the
+            # string value is likely junk and we don't want it in Elasticsearch.
+            new_value = "BAD DATA"
```
willkg: Replacing with "BAD DATA" might be unenthusing, but it doesn't happen often, so I figure go for the easy fix now and a harder fix later.

willkg:
self-r+

PR #5284, opened 2020-03-17
bug 1623064: rework save_processed and save_raw_and_processed

willkg:
Previously, we had a save_processed and a `save_raw_and_processed`. `save_processed` is only called by `save_raw_and_processed` and `save_raw_and_processed` is misnamed--it doesn't save the raw crash since doing that would irrecoverably change our record of what we got from the collector.

This cleans up naming and reduces these two functions down to a single `save_processed_crash`. We pass in the `raw_crash` and `processed_crash` so crashstorage can combine them into a crash report, but no crashstorage should ever stomp on the original `raw_crash` data.

Since this changes `save_raw_and_processed` into `save_processed_crash` and that's part of metrics names, this will change those metrics and we need to update the dashboards.

To test:

1. run `make test` and `make lint`
2. run the processor and process some crashes

### DIFF
--- a/docs/tests/system_checklist.rst
+++ b/docs/tests/system_checklist.rst
@@ -94,7 +94,7 @@ Checklist
 
       To check for errors grep for "ERRORS".
 
-    * Check Datadog "processor.save_raw_and_processed" for appropriate
+    * Check Datadog "processor.save_processed_crash" for appropriate
       environment.
 
       localdev: Check the logging in the console
@@ -104,13 +104,13 @@ Checklist
     Is the processor saving to ES? S3?
 
     * Check Datadog
-      "processor.es.ESCrashStorageRedactedJsonDump.save_raw_and_processed.avg"
+      "processor.es.ESCrashStorageRedactedJsonDump.save_processed_crash.avg"
 
       stage: https://app.datadoghq.com/dash/187676/socorro-stage-perf
       prod: https://app.datadoghq.com/dash/65215/socorro-prod
 
     * Check Datadog
-      "processor.s3.BotoS3CrashStorage.save_raw_and_processed" for
+      "processor.s3.BotoS3CrashStorage.save_processed_crash" for
       appropriate environment.
 
       stage: https://app.datadoghq.com/dash/187676/socorro-stage-perf
--- a/socorro/external/crashstorage_base.py
+++ b/socorro/external/crashstorage_base.py
@@ -2,9 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-"""This is the base of the crashstorage system - a unified interfaces for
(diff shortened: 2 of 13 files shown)


docs/tests/system_checklist.rst
```diff
@@ -104,13 +104,13 @@ Checklist
     Is the processor saving to ES? S3?
 
     * Check Datadog
-      "processor.es.ESCrashStorageRedactedJsonDump.save_raw_and_processed.avg"
+      "processor.es.ESCrashStorageRedactedJsonDump.save_processed_crash.avg"
 
       stage: https://app.datadoghq.com/dash/187676/socorro-stage-perf
       prod: https://app.datadoghq.com/dash/65215/socorro-prod
 
     * Check Datadog
```
willkg: This is wrong. We need to update the system checklist, but that's a job for another day.

willkg:
This looks ok. However, I have to update dashboards, so I'll wait until tomorrow, skim it again, then land it.

socorro/external/crashstorage_base.py
```diff
@@ -195,140 +188,98 @@ def is_mutator(self):
...
-                        of a dict.
-            dumps - a dict of dump name keys and binary blob values
-            crash_id - the crash key to use for this crash"""
-        pass
+        :param raw_crash: Mapping containing the raw crash meta data. It is often saved
+            as a json file, but here it is in the form of a dict.
+        :param dumps: A dict of dump name keys and binary blob values.
+        :param crash_id: the crash report id
 
-    def save_raw_crash_with_file_dumps(self, raw_crash, dumps, crash_id):
```
willkg: Nothing used this except the test that tested it, so I removed it.

socorro/external/crashstorage_base.py
```diff
@@ -195,140 +188,98 @@ def is_mutator(self):
...
-        parameters:
-           crash_id - the id of a processed_crash to fetch"""
+        :param crash_id: crash report id
+
+        :returns: DotDict
+
+        """
         raise NotImplementedError("get_unredacted_processed is not implemented")
 
     def remove(self, crash_id):
```
willkg: Nothing implements this, so we could probably remove this, too. Having said that, it'd be handy to have so we didn't have to implement it by hand in the deletion scripts.

socorro/external/crashstorage_base.py
```diff
@@ -195,140 +188,98 @@ def is_mutator(self):
...
-        """delete a crash from storage
+        """Delete crash report data from storage
 
-        parameters:
-           crash_id - the id of a crash to fetch"""
-        raise NotImplementedError("remove is not implemented")
+        :param crash_id: crash report id
 
-    def new_crashes(self):
```
willkg: This and `ack_crash` were from the days when crashstorage was also used for queueing crashes for processing. I split that out into a separate crashqueue hierarchy, so this is vestigial now.

willkg:
self-r+

After landing this, I need to update the stage dashboards.

PR #5285, opened 2020-03-18
bug 1607519: fix JavaStackTrace parsing

willkg:
There was a case where JavaStackTrace parsing would hit a multiline message where one of the lines was empty. In this case, the parser would get confused and state the JavaStackTrace was malformed.

This fixes that by restricting the check to just None allowing for empty lines.


willkg:
This fixes the issue with malformed JavaStackTrace crash reports. For example, it fixes this one: https://crash-stats.mozilla.org/report/index/d66fb046-8458-4da9-85b4-068840200305

self-r+

PR #5286, opened 2020-03-19
bug 1608116: add DeNoneRule

willkg:
This adds a `DeNoneRule` that will remove keys from the raw crash that have a `None` value. These come from crash annotations the crash reporter packs into a JSON-encoded value in the crash report. Previously, I think these would end up with empty strings.

The field we seem to be having the most problems with is `BuildID`.

This also adds some metrics code for the `DeNoneRule` and `DeNullRule` so we can see how often they're kicking off so as to identify bugs in the crash reporter.

### DIFF
--- a/socorro/processor/rules/general.py
+++ b/socorro/processor/rules/general.py
@@ -3,8 +3,12 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 from glom import glom
+import markus
 
 from socorro.processor.rules.base import Rule
+
+
+METRICS = markus.get_metrics("processor")
 
 
 class DeNullRule(Rule):
@@ -12,9 +16,6 @@ class DeNullRule(Rule):
 
     Sometimes crash reports come in with junk data. This removes the egregious
     junk that causes downstream processing and storage problems.
-
-    NOTE(willkg): While this changes the raw crash, that raw crash should never
-    get saved back to storage, so this doesn't overwrite the original material.
 
     """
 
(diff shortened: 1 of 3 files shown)


socorro/processor/rules/general.py
```diff
@@ -3,19 +3,20 @@
...
+
+
 class DeNullRule(Rule):
     """Removes nulls from keys and values
 
     Sometimes crash reports come in with junk data. This removes the egregious
     junk that causes downstream processing and storage problems.
 
-    NOTE(willkg): While this changes the raw crash, that raw crash should never
-    get saved back to storage, so this doesn't overwrite the original material.
```
willkg: This isn't needed anymore. I nixed `save_raw_and_processed` and changed the contract so that `save_processed_crash` never saves raw crash data.

socorro/processor/rules/general.py
```diff
@@ -37,14 +38,43 @@ def de_null(self, s):
         return s
 
     def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
+        had_nulls = False
+
         # Go through the raw crash and de-null keys and values
-        for key, val in raw_crash.items():
+        for key, val in list(raw_crash.items()):
```
willkg: It's possible for this rule to change the thing being iterated over. Given that, we need to convert the iterator to a new list that doesn't change.

socorro/unittest/processor/rules/test_general.py
```diff
@@ -110,7 +111,31 @@
...
+        rule.action(raw_crash, None, {}, {})
+        assert raw_crash == expected
+
+    def test_denone_with_dotdict(self):
+        raw_crash = DotDict({"foo": "bar", "baz": None})
+        expected = DotDict({"foo": "bar"})
+
+        rule = DeNoneRule()
+        rule.action(raw_crash, None, {}, {})
+        assert raw_crash == expected
```
willkg: I kept the `DotDict` test separate because at some point, we're going to remove `DotDict` usage.

willkg:
I tested this by processing 4e8c8ddd-6a1f-4905-9f02-94b840200108 which has a `BuildId` of `None`.

willkg:
self-r+

willkg:
I was tossing around unifying all the "lets fix the raw crash" rules, but decided they all have different reasons for existing and maybe it's better to keep them as separate rules.

PR #5287, opened 2020-03-19
bug 1622932: fix featured versions

willkg:
The featured versions list is calculated for most products based on existing crash reports. This allows the list to be dynamic and change quickly which is helpful during releases. However, this means that calculating the list is vulnerable to junk data and bugs.
    
This adjusts the calculation slightly in that it adds a minimum count threshold. I picked 50 after looking at all the existing products on Socorro. Going lower increases the likelihood that junk data affects the featured versions. Going higher increases the likelihood we hit an issue with less-used products where versions take a really long time to show up.
    
We can tweak the number a little as circumstances come up, but I think if products have a problem with it, they should do what Fennec is doing and manually maintain a list of featured versions.
    
Note that the `VERSIONS_COUNT_THRESHOLD` is set to 0 for the local dev environment.

### DIFF
--- a/webapp-django/crashstats/crashstats/utils.py
+++ b/webapp-django/crashstats/crashstats/utils.py
@@ -342,12 +342,6 @@ def enhance_raw(raw_crash):
             raw_crash["AdapterDeviceName"] = result[1]
 
 
-#: Number of days to look at for versions in crash reports. This is set
-#: for two months. If we haven't gotten a crash report for some version in
-#: two months, then seems like that version isn't active.
-VERSIONS_WINDOW_DAYS = 60
-
-
 def get_versions_for_product(product="Firefox", use_cache=True):
     """Returns list of recent version strings for specified product
 
@@ -376,15 +370,16 @@ def get_versions_for_product(product="Firefox", use_cache=True):
     now = timezone.now()
 
     # Find versions for specified product in crash reports reported in the last
-    # VERSIONS_WINDOWS_DAYS days and use a big _facets_size so that it picks up versions
+    # VERSIONS_WINDOW_DAYS days and use a big _facets_size so that it picks up versions
     # that have just been released that don't have many crash reports, yet
+    window = settings.VERSIONS_WINDOW_DAYS
     params = {
         "product": product,
         "_results_number": 0,
         "_facets": "version",
         "_facets_size": 1000,
         "date": [
-            ">=" + (now - datetime.timedelta(days=VERSIONS_WINDOW_DAYS)).isoformat(),
+            ">=" + (now - datetime.timedelta(days=window)).isoformat(),
             "<" + now.isoformat(),
         ],
     }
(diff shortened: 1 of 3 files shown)


webapp-django/crashstats/crashstats/utils.py
```diff
@@ -342,17 +342,11 @@ def enhance_raw(raw_crash):
...
-#: for two months. If we haven't gotten a crash report for some version in
-#: two months, then seems like that version isn't active.
-VERSIONS_WINDOW_DAYS = 60
-
-
 def get_versions_for_product(product="Firefox", use_cache=True):
     """Returns list of recent version strings for specified product
 
     This looks at the crash reports submitted for this product over
-    VERSIONS_WINDOW_DAYS days and returns the versinos of those crash reports.
```
willkg: Fixed this typo. It was just sitting there. Its very existence was a blight upon this repository. Fixing it is probably the most important thing I did all year.

webapp-django/crashstats/crashstats/utils.py
```diff
@@ -376,47 +370,47 @@ def get_versions_for_product(product="Firefox", use_cache=True):
...
     # Since we're caching the results of the search plus additional work done,
-    # we don't need to cache the fetch
+    # we don't want to cache the fetch
     ret = api.get(**params, dont_cache=True)
     if "facets" not in ret or "version" not in ret["facets"]:
         return []
 
     # Get versions from facet, drop junk, and sort the final list
-    betas = set()
-    versions = []
```
willkg: This code was weird, so I reduced it down a bit.

webapp-django/crashstats/crashstats/utils.py
```diff
@@ -376,47 +370,47 @@ def get_versions_for_product(product="Firefox", use_cache=True):
...
+    versions = set()
     for item in ret["facets"]["version"]:
+        if item["count"] < settings.VERSIONS_COUNT_THRESHOLD:
+            continue
```
willkg: This is the actual fix. It's pretty minimal. Minimal ftw.

webapp-django/crashstats/settings/base.py
```diff
@@ -659,4 +659,13 @@ def filter(self, record):
 # This is the number of versions to display if a particular product
 # has no 'featured versions'. Then we use the active versions, but capped
 # up to this number.
-NUMBER_OF_FEATURED_VERSIONS = config("NUMBER_OF_FEATURED_VERSIONS", 4)
+NUMBER_OF_FEATURED_VERSIONS = config("NUMBER_OF_FEATURED_VERSIONS", 4, cast=int)
```
willkg: Given that this wasn't cast and thus was probably wrong, it's probably not used anywhere anymore. We might be able to remove it. But I think I'm going to think about that another day.

willkg:
self-r+

PR #5289, opened 2020-03-23
bug 1596106: security fixes

willkg:
This fixes a couple of issues raised by ZAP.

Static assets are served by whitenoise which is adding this header:

```
Access-Control-Allow-Origin: *
```
    
Since we're not hosting static assets on a CDN, we don't need this so we should disable it.

This adds data-no-csrf attribute to three forms:
    
* quick search form in the navbar
* super search form on the super search page
* signature search form
    
These forms all use GET method and they're search forms, but ZAP doesn't know that, so we're adding the attribute so ZAP doesn't report them as failures.


willkg:
I'm going to land this, see what ZAP says, and verify that it didn't break anything. self-r+

PR #5290, opened 2020-03-24
bug 1624403: reduce localstack containers

willkg:
We can run one localstack container that handles both services, so this fixes things accordingly.

Further this updates localstack to 0.10.8.

The issue is intermittent and I can't reproduce it locally. Hopefully one of these two things reduces CI failures because the SQS container isn't really up, yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,7 +125,7 @@ run: my.env
 
 .PHONY: runservices
 runservices: my.env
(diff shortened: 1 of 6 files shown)


willkg:
So theoretically given that this passed it's probably not _worse_. self-r+

PR #5291, opened 2020-03-24
Update to CircleCI config 2.1

willkg:
As a side-effect, this switches CircleCI pipelines on.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 jobs:
   build:
     docker:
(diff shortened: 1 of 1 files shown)


willkg:
Weird. The "configuration" tab seems to say it's still using the 2 config and not 2.1. I verified that the config file is valid, so I'm puzzled.

Then I decided to opt-in to the new user interface because why not throw more problems in the fire to burn and the new user interface differentiates between original config file and the "compiled config file". So maybe it's fine.

This didn't break anything as far as I can tell, so I'm going to land it. self-r+

PR #5292, opened 2020-03-25
bug 1624403: fix bucket and queue names

willkg:
AWS S3 bucket names can't have underscores in them. This fixes the names in the local dev environment to comply with that.

AWS SQS queue names *can* have underscores, but I figured it's easier on my brain if names all have the same conventinos so I changed those, too.

This only affects the local dev environment.


willkg:
This only affects CI and local development. self-r+

PR #5294, opened 2020-03-25
Add Fenix feature verions

rocketsroger:


willkg:
Once merged, it may take as much as an hour for the caches in production to expire, but after that you should be able to see the changes in production.

rocketsroger:
> Once merged, it may take as much as an hour for the caches in production to expire, but after that you should be able to see the changes in production.

Understood.  Thanks!

PR #5295, opened 2020-03-25
bug 1619638: remove java_stack_trace_full before indexing in Elasticsearch

willkg:
Elasticsearch crashstorage has problems with the `java_stack_trace_full` field. We removed it a while back, but crash reports processed before we removed it still have the field in the processed crash file.

This adds it to the `forbidden_keys` list for Elasticsearch. We can remove the forbidden key in August 2020.

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -456,7 +456,7 @@ class ESCrashStorageRedactedSave(ESCrashStorage):
         default="socorro.external.crashstorage_base.Redactor",
         from_string_converter=class_converter,
     )
-    required_config.add_option(
+    required_config.es_redactor.add_option(
         name="forbidden_keys",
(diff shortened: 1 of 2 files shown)


socorro/external/es/crashstorage.py
```diff
@@ -520,7 +520,7 @@ class ESCrashStorageRedactedJsonDump(ESCrashStorageRedactedSave):
         default="socorro.external.crashstorage_base.Redactor",
         from_string_converter=class_converter,
     )
-    required_config.add_option(
+    required_config.es_redactor.add_option(
```
willkg: I'm 90% positive these had the wrong namespace and thus weren't doing anything. The redactor works currently since the configuration is done in the processor app in a giant configuration thing.

willkg:
self-r+

PR #5296, opened 2020-03-26
bug 1624403: switch SQS check to use waitfor script

willkg:
This adds a waitfor script that does an HTTP connection and waits for a certain response code. This switches the SQS check to use that so we can avoid HTTP 502 bad gateway which passes urlwait, but hoses CI.

### DIFF
--- a/docker/run_tests_in_docker.sh
+++ b/docker/run_tests_in_docker.sh
@@ -25,9 +25,11 @@ SOCORRO_GID=${SOCORRO_GID:-"10001"}
 BASEIMAGENAME="python:3.7.6-slim-stretch"
 TESTIMAGE="local/socorro_app"
 
-# Start services in background (this is idempotent)
+# Start services in background (this is idempotent)--make sure to start
+# localstack first since it needs some extra time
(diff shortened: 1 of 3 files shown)


docker/run_tests_in_docker.sh
```diff
@@ -25,9 +25,11 @@ SOCORRO_GID=${SOCORRO_GID:-"10001"}
 BASEIMAGENAME="python:3.7.6-slim-stretch"
 TESTIMAGE="local/socorro_app"
 
-# Start services in background (this is idempotent)
+# Start services in background (this is idempotent)--make sure to start
+# localstack first since it needs some extra time
 echo "Starting services needed by tests in the background..."
-${DC} up -d elasticsearch localstack postgresql statsd
+${DC} up -d localstack
+${DC} up -d elasticsearch postgresql statsd
```
willkg: I'm not sure this has any effect since passing `-d` makes it "happen in the background and return almost instantly".

willkg:
Jeremy Bearimy+

PR #5297, opened 2020-03-26
bug 1624403: fix case where URLError has no code

willkg:


willkg:
Ok. Let's see whether this helps. self-r+

PR #5298, opened 2020-03-26
bug 1625154: fix updatemissing to work across date bounds

willkg:
updatemissing can't find processed crash data in Elasticsearch where the crash came in on one day and was processed the next day. This happens around midnight.

This increases the date range to look at and also nixes the `_results_number` which I don't think we need to pass in.

To test:

1. process some crashes
2. use `socorro-cmd es list` to get the list of indexes, then use `socorro-cmd es list_crashids <INDEX>` to figure out which index to delete, then `socorro-cmd es delete <INDEX>` to delete the index with crashes in it
3. run `cd webapp-django` then `./manage.py verifyprocessed` which creates a missing processed crash item
4. run `./manage.py updatemissing` and it'll say it didn't update any missing crashes
5. reprocess the crashes
6. run `./manage.py updatemissing` and it'll say it found them


willkg:
self-r+

PR #5299, opened 2020-03-26
bug 1624911: change Fenix nightly versions from "Nightly..." to "0.0a1"

willkg:
This allows grouping of Fenix nightly crash reports until they change the version number to something Socorro understands better.


willkg:
This looks good. Once I land and deploy it, I'll run it by @rocketsroger . We can fix and reprocess if there are issues.

self-r+

PR #5300, opened 2020-03-27
bug 1624870: improve product support docs and tips

willkg:
This improves the product support documentation. It also moves the product documentation to the Socorro ReadTheDocs docs.

This tweaks the UX for tips by making them clearer, shorter, and more obvious.


willkg:
I think this is good.

Generally, I'm moving Crash Stats and Socorro at Mozilla user-centric docs to ReadTheDocs where it's easier to write and maintain as part of the whole system. There are some docs that won't work for--like the API docs which are automatically generated and have example things.

self-r+

PR #5301, opened 2020-03-27
Modify Fenix feature versions

rocketsroger:


product_details/Fenix.json
```diff
@@ -1,3 +1,3 @@
 {
-    "featured_versions": ["1.0.2013", "1.0.2012", "1.0.2011", "75.0", "74.0"]
+    "featured_versions": ["1.0.2013", "1.0.2012", "0.0a", "75.0", "75.0a", "74.0"]
```
willkg: I think you're adding nightlies to this list, so this needs to have the full nightly version:

```
     "featured_versions": ["1.0.2013", "1.0.2012", "0.0a1", "75.0", "75.0a1", "74.0"] 
```

rocketsroger: Understood.  Thanks

PR #5302, opened 2020-03-27
bug 1622976: first pass at socorro -> telemetry docs

willkg:


willkg:
This is a good first pass. We can hone it going forward. self-r+

PR #5325, opened 2020-04-01
bug 1626032: fix validate_and_test script

willkg:
This fixes the `validate_and_test` script so it runs. It's still erroring out on `SubmittedFromInfobar`, but that's a different problem.

To test:

```
$ make shell
app@socorro:/app$ python socorro/schemas/validate_and_test.py 
```

### DIFF
--- a/socorro/schemas/validate_and_test.py
+++ b/socorro/schemas/validate_and_test.py
@@ -39,8 +39,7 @@ class MockedTelemetryBotoS3CrashStorage(TelemetryBotoS3CrashStorage):
         self._all_fields = resp.json()
         self.conn = MockConn()
 
-    def save_processed_crash(self, raw_crash, processed_crash):
-        super().save_processed_crash(raw_crash, processed_crash)
+    def get_last_data(self):
         return self.conn.last_data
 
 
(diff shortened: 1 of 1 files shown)


willkg:
This script is run by developers after making schema changes. It has no tests which is how it broke after I made the `save_raw_and_processed` -> `save_processed_crash` changes. I'm going to write up a separate bug for adding a basic test to make sure it still vaguely works.

For now, I need to fix the immediate issue and move on.

This works in my local dev environment except for the problem with `SubmittedFromInfobar` which is bug 1626048.

socorro/schemas/validate_and_test.py
```diff
@@ -18,51 +19,69 @@
...
-        self._all_fields = r.json()
+        resp = requests.get(API_BASE.format("SuperSearchFields"))
+        print(resp.url)
+        self._all_fields = resp.json()
+        self.conn = MockConn()
 
     def save_processed_crash(self, raw_crash, processed_crash):
-        self.combined = super().save_processed_crash(raw_crash, processed_crash)
+        super().save_processed_crash(raw_crash, processed_crash)
+        return self.conn.last_data
```
willkg, line 44: I need a comment here about how this changes the API set by crash storage base.

willkg:
self-r+

PR #5326, opened 2020-04-01
bug 1626618: change BetaVersionRule to only work on supported products

willkg:
We've got lots of products that aren't like Firefox and Fennec and adding the "b0" at the end is disasterous. I don't even know how to spell that word it's so bad.

This changes the rule to only operate on Firefox, Fennec, and FennecAndroid.


willkg:
I verified this fixes the Fenix 4.2.0 crash reports. self-r+

PR #5330, opened 2020-04-06
bug 1626801: move CxxThrowException to prefix list

willkg:
Having `CxxThrowException` in the prefix and show up in the signature fixes some bucketing.


willkg:
```
Crash id: 4c1410c4-c35e-4a04-b767-de09a0200401
Original: CxxThrowException
New:      CxxThrowException | goingime.x64.dll | CThreadMgrEventSink::OnSetFocus
Same?:    False
```

self-r+

PR #5337, opened 2020-04-06
bug 1626048: fix SubmittedFromInfobar values

willkg:
The values are `"true"` and `true` and it should be `"1"`. The crash reporter was fixed to always send `"1"`, but we have to handle crash reports for all the broken versions.

### DIFF
--- a/socorro/processor/rules/mozilla.py
+++ b/socorro/processor/rules/mozilla.py
@@ -35,10 +35,10 @@ MAXINT = 9223372036854775807
 class ConvertModuleSignatureInfoRule(Rule):
     """Make ModuleSignatureInfo to a string.
 
-    For a while, crash reports submitted as a JSON blob had ModuleSignatureInfo appended
-    to the end of them as an object. This JSON-encodes that object so that it's always a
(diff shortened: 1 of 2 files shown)


socorro/processor/rules/mozilla.py
```diff
@@ -35,10 +35,10 @@
...
     """Make ModuleSignatureInfo to a string.
 
-    For a while, crash reports submitted as a JSON blob had ModuleSignatureInfo appended
-    to the end of them as an object. This JSON-encodes that object so that it's always a
-    JSON-encoded string. That way, the rest of the processor doesn't have to handle both
-    cases. Bug #1607806.
+    For a while, crash reports with annotations submitted as a JSON blob had
+    ModuleSignatureInfo appended to the end of them as an object. This JSON-encodes that
+    object so that the value is always a JSON-encoded string. That way, the rest of the
+    processor doesn't have to handle both cases. Bug #1607806
```
willkg, line 41: I improved this docstring and made it clearer.

willkg:
self-r+

PR #5343, opened 2020-04-08
bug 1628311: fix scripts/run_mdsw.sh

willkg:
It was looking at the wrong environment variable for the command pathname. That changed a while ago, but the change missed this script.


willkg:
self-r+

PR #5344, opened 2020-04-14
Adjust core::result::unwrap_failed entry in the prefix list

jcristau:
Looks like it sometimes occurs without the extra `<T>`, e.g.
https://bugzilla.mozilla.org/show_bug.cgi?id=1628886


willkg:
This is missing the bookkeeping and I figured it's probably easier for you if I just went ahead and did it rather than have you update the commit and PR. For some things, the bookkeeping isn't a big deal. But for signature generation, it's so squirrely and mystical that having the bookkeeping that ties back to the "why we did this and what did we test" makes future debugging and decision-making easier.

Thank you for doing this!

jcristau:
Thanks for taking care of this, I'll go through bugzilla next time.

PR #5346, opened 2020-04-14
Update minimist to 1.2.3 or later

willkg:
This updates minimist to a version that isn't known vulnerable.


willkg:
I loaded the site locally and went through it. We don't have any JS testing, so maybe I would have caught issues, but maybe not. I do know that the things I went through worked and that's probably good enough of a smoketest.

self-r+

PR #5351, opened 2020-04-17
bug 1409547: add basic tests to scripts

willkg:
This rewrites `db` and `es` to use click. It adds a basic "does this run" test, too.

I didn't do the other scripts in that directory because I think I'm going to remove them when I work on bug 1631108.


willkg:
self-r+

PR #5356, opened 2020-04-27
bug 1383113: switch to getitem notation for general rules

willkg:
I had switched most of the general rules code already. This updates the tests.

To test:

1. run tests


willkg:
self-r+

PR #5357, opened 2020-04-27
bug 1383113: remove DotDict usage in breakpad rules

willkg:
This removes the last bit of DotDict usage in the breakpad processor rules that I missed in an earlier pass.


willkg:
self-r+

PR #5388, opened 2020-05-05
bug 1622932: explicitly exclude "1024" version

willkg:
None of our products are going to hit 1024 any time soon. We don't have the problem with other versions, so for now, let's just exclude version 1024 and push off a better fix which will likely require a lot more work until later.


willkg:
self-r+ with shame.

PR #5390, opened 2020-05-06
bug 1633501: remove vestigal user_id bits in processor

willkg:
Seems like there was a UserId in raw crashes a while ago and someone wrote code in the processor to stomp on those values. We don't need this anymore, so we can remove it.


willkg:
I looked at a sample of 1000 raw crashes for Firefox and didn't see any "UserID" fields. I think we're good here.

willkg:
self-r+

PR #5391, opened 2020-05-07
bug 1625515: redo version parsing and sorting

willkg:
This converts versions into semver `VersionInfo` instances and uses those for sorting. Generally, this works, but there are some hacks:
    
1. Firefox/Fennec versions aren't valid semver, so I changed the parser we had before to parse those versions and generate semver `VersionInfo` instances.

   * 75.0a1 -> 75.0.0-alpha.1
   * 75.0b2 -> 75.0.0-beta.2

2. ESR releases need to sort *after* normal regular releases. The only way to do that with semver that I could figure is to mark everything with prerelease bits so that they sort correctly.
    
   * 68.0.2esr -> 68.0.2-xsr
    
   Note that "xsr" parses alphabetically after alpha, beta, and release.
    
3. Firefox/Fennec have release candidates of betas so we have to add "rc" to everything.
    
   * 75.0b2 -> 75.0.0-beta.2.rc.999
   * 75.0.1 -> 75.0.1-release.rc.999
   * 68.0.2esr -> 68.0.2-xsr.rc.999
    
4. The "b" ending thing needs to sort before all the betas--it's a catch-all for all betas.

I tested this by pulling all the versions for 10,000 crash reports for Firefox and Fenix and then making sure that `generate_semver` works and that the resulting lists of versions sorted correctly.

After this lands, I'll need to do some more verification on stage.

### DIFF
--- a/socorro/lib/versionutil.py
+++ b/socorro/lib/versionutil.py
@@ -6,14 +6,15 @@ import semver
 
 
 class VersionParseError(Exception):
-    """Raised if the version isn't parseable"""
+    """Raised if the version isn't parseable."""
 
 
 def generate_semver(version):
     """Convert a version to semver.
 
-    This converts to a semver that (ab)uses the prerelease section to denote the channel
-    and in doing that, sorts Firefox versions correctly.
+    This converts the version string to a semver that (ab)uses the prerelease section to
+    denote the channel and in doing that, sorts Firefox, Fennec, and Fenix versions
+    correctly.
 
     :param version: the version string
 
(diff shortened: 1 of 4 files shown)


socorro/lib/versionutil.py
```diff
@@ -2,88 +2,80 @@
...
-    if len(version) < 2:
-        return False
 
-    # First part should consist of one or more ascii digits
-    if not (len(version[0]) > 0 and version[0].isdigit()):
-        return False
+def generate_semver(version):
+    """Convert a version to semver.
 
-    return True
```
willkg, line 27: This is crap validation and we "get it for free" by try to parse the version. So I nixed this.

socorro/lib/versionutil.py
```diff
@@ -2,88 +2,80 @@
...
-            # We use 999 so that the release always sorts after the release
-            # candidates
-            rc = 999
-
-        if "pre" in version:
-            # Treat "pre" like we do rc, but call it 1 if there's no number
-            version, rc = version.split("pre")
-            if not rc:
-                rc = 1
+            prerelease.insert(0, "rc.999")
```
willkg, line 44: I should add tons of comments to this section.

willkg:
self-r+

PR #5393, opened 2020-05-11
bug 1608873: switch from hg annotate to hg file link

willkg:
The file link comes up much faster than the annotate link does. If users want annotations, they can easily switch to the annotate link after the file link comes up.


willkg:
self-r+

PR #5394, opened 2020-05-11
bug 1619962: rework deletion scripts to use click and add basic test

willkg:
This reworks the deletion scripts to use click for argument parsing. It also adds basic tests.


willkg:
self-r+

PR #5395, opened 2020-05-12
bug 1616718: fix code to use bugzilla api token if available

willkg:
This adds a `BZAPI_TOKEN` setting. The `bugassociations` job and `BugzillaBugInfo` model will use it if it's available.

To test:

1. process some crashes
2. do `make shell` and `cd webapp-django` and `manage.py cronrun --job=bugassociations`

   It should work fine and not log that it's using a token.

3. Set up an account and API token on https://bugzilla-dev.allizom.org/ and set in `my.env`:

   ```
   BZAPI_BASE_URL=https://bugzilla-dev.allizom.org/rest
   BZAPI_TOKEN=yourtoken
   ```

4. run `manage.py cronreset bugassociations` and then `manage.py cronrun --job=bugassociations`

   It should run and log that it used a token.


willkg:
self-r+

PR #5396, opened 2020-05-13
bug 1630968: remove deprecated monitoring endpoints

willkg:
We no longer use these endpoints for monitoring anymore, so we can remove them.


willkg:
self-r+

PR #5398, opened 2020-05-18
Remove "object" superclass

willkg:
Python 2 required subclassing from `object` in order to use new-style class functionality. In Python 3, everything is a new-style class, so you don't need to specify `object` anymore.


willkg:
self-r+

PR #5403, opened 2020-05-27
bug 1641316: fix bugzilla bug creation for java errors

willkg:
If the crash report represents a java error, then there's a `JavaStackTrace` annotation and Crash Stats puts that in the crash report. However, the backticks were in the wrong place in respects to if/for blocks and so it was missing the end set.

While fixing that, I tweaked the description a bit.


webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -177,6 +177,8 @@ def bugzilla_submit_url(report, parsed_dump, crashing_thread, bug_product):
         },
     )
 
+    print(repr(comment))
```
willkg, line 180: Oops... Need to remove this.

willkg:
self-r+

PR #5429, opened 2020-06-01
bug 1611131: improve language around protected data and access

willkg:
Over the years, Crash Stats has evolved how it talks about data that can only be seen by people who have been vetted and granted access to it.

This goes through and changes all the language to talk about "protected data" and "protected data access". This updates the url where the protected data access policy exists and provides a redirect for the original policy url. This reduces wasted space on the report view and adds a statement at the top explaining what can be seen and why with links to documentation on requesting protected data access. This adds exclamation marks to some sensitive fields--we should add more as we find them.

The goals of this work are three-fold:

1. use consistent language
2. make it clearer what is and isn't protected
3. make it clearer what users can see depending on whether they're logged in or not and whether they've been granted access to protected data

### DIFF
--- a/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
+++ b/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
@@ -1,8 +1,9 @@
-{% macro sensitive_warning(your_crash) -%}
+{% macro protected_warning(your_crash) -%}
+  <img src="{{ static('img/3rdparty/silk/exclamation.png') }}" alt="protected" title="Protected Data" />
   {% if your_crash %}
-    You can only see this because it is your crash!
+    Protected data: You can see it because it is your crash.
   {% else %}
-    This is super sensitive data! Be careful how you use it!
+    Protected data: Be careful how you use it.
   {% endif %}
 {%- endmacro %}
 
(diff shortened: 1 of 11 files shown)


willkg:
We should add a test verifying the redirect from the original memory dump policy url still works.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -41,6 +42,19 @@
       <h2>{{ report.product }} {{ report.version }} Crash Report [@ {{ report.signature }} ]</h2>
     </div>
 
+    <div class="protected-info">
+      {% if your_crash %}
+        You are seeing public and protected data for your crash.
```
willkg, line 47: This bit is only in the report view because users sometimes look at their own reports.

webapp-django/crashstats/crashstats/static/crashstats/css/base/layout.less
```diff
@@ -15,7 +15,7 @@ html {
             display: none;
         }
         .page-heading {
-            margin: 4em 2em 2em;
+            margin: 1em 1em 1em;
```
willkg, line 18: This reduces some of the wasted space.

webapp-django/crashstats/crashstats/static/crashstats/css/base/layout.less
```diff
@@ -173,6 +173,12 @@ body {
     }
 }
 
+.protected-info {
+    margin: 1em 1em 1em;
+    display: flex;
+    align-items: center;
+}
```
willkg, line 180: This is a new element that's added to the header on relevant pages to show the protected data status of the page and why/why not a user is seeing protected data.

willkg:
self-r+

PR #5439, opened 2020-06-03
Modify Fenix feature versions

rocketsroger:


product_details/Fenix.json
```diff
@@ -1,3 +1,3 @@
 {
-    "featured_versions": ["5.0.0", "4.3.0", "4.2.0", "0.0a1", "76.0a1", "75.0a1"]
+    "featured_versions": ["5.1.0", "5.0.0", "4.3.0", "0.0a1", "77.0.0-b", "76.0.0-b", "75.0.0-b"]
```
willkg, line 2: Ugh. It's weird that it's `-b`. It should be `-beta` like the rest of the versions. Everything is so terrible.

PR #5440, opened 2020-06-05
bug 1497353: remove not_analyzed fields

willkg:
This removes the "dump" field which wasn't used and wasn't exposed.

This also switches the "uuid" field from `not_analyzed` to `keyword`. I did some searches and facets and it still works fine.


willkg:
self-r+

PR #5441, opened 2020-06-05
bug 1624345: stop saving random data to elasticsearch crashstorage

willkg:
Previously, we would put everything in the raw and processed crash in the document and index that in Elasticsearch--even things that were not specified in super_search_fields. This caused multiple problems with strings that were invalid keys. We worked through those. This also caused the problem where we have a bunch of unknown user-provided data in Elasticsearch.

The one thing where this helps is we can sometimes retroactively make a field searchable and have some data in Elasticsearch already. This doesn't happen often enough to be compelling.

This changes the Elasticsearch crash storage such that it removes anything in the raw and processed crash that isn't specified in super_search_fields. This moves us another step towards a world where we have a well defined schema of what we're storing.

### DIFF
--- a/socorro/external/crashstorage_base.py
+++ b/socorro/external/crashstorage_base.py
@@ -6,7 +6,6 @@
 
 import datetime
 import collections
-import copy
 import logging
 import os
 import sys
@@ -178,14 +177,6 @@ class CrashStorageBase(RequiredConfig):
     def close(self):
         """Close resources used by this crashstorage."""
         pass
-
-    def is_mutator(self):
-        """Whether this storage class mutates the crash or not
-
-        By default, crash storage classes don't mutate the crash.
-
-        """
-        return False
 
     def save_raw_crash(self, raw_crash, dumps, crash_id):
         """Save raw crash data.
@@ -476,19 +467,14 @@ class PolyCrashStorage(CrashStorageBase):
         :param raw_crash: the raw crash data
         :param processed_crash: the processed crash data
 
+        NOTE(willkg): Crash storage systems that implement this should
+        not mutate the raw and processed crash structures!
+
         """
         storage_exception = PolyStorageError()
         for a_store in self.stores.values():
             try:
-                actual_store = getattr(a_store, "wrapped_object", a_store)
-                if hasattr(actual_store, "is_mutator") and actual_store.is_mutator():
-                    my_raw_crash = copy.deepcopy(raw_crash)
-                    my_processed_crash = copy.deepcopy(processed_crash)
-                else:
-                    my_raw_crash = raw_crash
-                    my_processed_crash = processed_crash
-
-                a_store.save_processed_crash(my_raw_crash, my_processed_crash)
+                a_store.save_processed_crash(raw_crash, processed_crash)
             except Exception:
                 store_class = getattr(a_store, "wrapped_object", a_store.__class__)
                 crash_id = processed_crash.get("uuid", "NONE")
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import copy
 import json
 import re
 import time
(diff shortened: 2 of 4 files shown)


socorro/external/crashstorage_base.py
```diff
@@ -179,14 +178,6 @@ def close(self):
...
         pass
 
-    def is_mutator(self):
-        """Whether this storage class mutates the crash or not
-
-        By default, crash storage classes don't mutate the crash.
-
-        """
-        return False
-
```
willkg, line 189: The Elasticsearch crashstorage was the only one that mutated the raw and processed crash data. Because we're changing it so that Elasticsearch crashstorage now creates new data structures, we don't need this mutator stuff anymore.

socorro/external/es/crashstorage.py
```diff
@@ -115,29 +116,24 @@ def is_valid_key(key):
...
+    for field in FIELDS.values():
+        if field["namespace"] != namespace:
+            continue
 
-    """
-    if not data:
-        return
+        key = field["in_database_name"]
+        if key in doc:
+            new_doc[key] = copy.deepcopy(doc[key])
```
willkg, line 134: We need the deepcopy for `json_dump`. That should be the only field that's deep. The rest of them are strings and ints and this should be a fast operation.

socorro/external/es/crashstorage.py
```diff
@@ -115,29 +116,24 @@ def is_valid_key(key):
     return bool(VALID_KEY.match(key))
 
 
-def remove_bad_keys(data):
```
willkg, line 118: We don't need `remove_bad_keys` anymore because we're only indexing a document with keys that are explicitly specified in super_search_fields.

socorro/external/es/crashstorage.py
```diff
@@ -270,15 +267,15 @@ def get_index_for_crash(self, crash_date):
 
         return index
 
-    def save_processed_crash(self, raw_crash, processed_crash):
-        """Save processed crash report to Elasticsearch"""
-        # Massage the crash such that the date_processed field is formatted
-        # in the fashion of our established mapping.
-        reconstitute_datetimes(processed_crash)
+    def prepare_processed_crash(self, raw_crash, processed_crash):
```
willkg, line 270: There are a bunch of subclasses that redact and adjust the data. Previously, these all overrode `save_processed_crash`, but they need to modify the data before calling that because that also indexes the final document.

This changes things so we have a prepare step first that everything can override. The order that prep steps happen isn't a big deal. After prep is done, then `save_processed_crash` kicks in.

socorro/external/es/crashstorage.py
```diff
@@ -290,6 +287,16 @@ def save_processed_crash(self, raw_crash, processed_crash):
         convert_booleans(FIELDS, raw_crash)
         convert_booleans(FIELDS, processed_crash)
 
+    def save_processed_crash(self, raw_crash, processed_crash):
+        """Save processed crash report to Elasticsearch"""
+        # Generate indexable raw and processed crash data and leave everything not
+        # listed in FIELDS out
+        raw_crash = extract_indexable(FIELDS, "raw_crash", raw_crash)
+        processed_crash = extract_indexable(FIELDS, "processed_crash", processed_crash)
```
willkg, line 295: `extract_indexable` creates a deep copy of the data.

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -47,7 +47,6 @@
     "cpu_arch": "arm",
     "crashedThread": 8,
     "date_processed": "2012-04-08 10:56:41.558922",
-    "dump": "...",
```
willkg, line 50: We no longer have this field.

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -98,24 +97,7 @@
     },
 }
 
-a_processed_crash_with_no_stackwalker = deepcopy(a_processed_crash)
```
willkg, line 101: I changed the tests that used this variable so that they build the processed crash as part of the test.

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -285,66 +286,14 @@ def test_success(self, espy_mock):
...
-        additional = {
-            "doc_type": "crash_reports",
-            "id": crash_id,
-            "index": "socorro_integration_test_reports",
-        }
-
-        sub_mock.index.assert_called_with(body=document, **additional)
-
-    @mock.patch("socorro.external.es.connection_context.elasticsearch")
-    def test_success_with_no_stackwalker_class(self, espy_mock):
```
willkg, line 307: Other tests test this scenario where there's no `json_dump` field.

willkg:
self-r+

PR #5442, opened 2020-06-09
bug 1497353: make telemetry_environment not searchable

willkg:
This changes `telemetry_environment` to be like `json_dump` and not searchable.

I verified that you can specify it as a column and get it back in the response, so everything continues to work except that it's no longer searchable.


willkg:
self-r+

PR #5447, opened 2020-06-30
bug 1645442: switch master to main

willkg:


setup.py
```diff
@@ -11,37 +11,33 @@
 
 def read(fname):
     fpath = os.path.join(os.path.dirname(__file__), fname)
-    with codecs.open(fpath, 'r', 'utf8') as f:
+    with codecs.open(fpath, "r", "utf8") as f:
```
willkg, line 14: It used to be that we built socorro as a Python package and then installed it from that. We don't do that anymore. I'm pretty sure we can nix this file altogether, but figured I'd look into that later.

willkg:
self-r+

PR #5482, opened 2020-07-02
Tweak Makefile

willkg:
This ditches the explicit help for help that's generated from the Makefile. This makes it a bit easier to maintain since you don't have to remember to update the docs far away from the things they're documenting.


willkg:
self-r+

PR #5495, opened 2020-07-27
bug 1649810: upgrade to python 3.8.5

willkg:
This upgrades to 3.8.5. In doing that, we switched from stretch to buster as the base image. That required a tweak to the debian packages we install.

The switch from 3.7 to 3.8 also required us to fix a test which had a line that now kicks up a syntax error.

It also required us to fix a test that relied on an iterator over dict keys being stable even if you're deleting the underlying structure. I'm not sure what changed between Python 3.7 and 3.8 to cause the test to fail, but I didn't spend a lot of time looking since the fix is a good defensive thing to do anyhow.

To test:

1. build
2. run tests
3. test the processor by processing crashes with minidumps


willkg:
CI failed to build the first time around, but I couldn't reproduce it locally. Looking at the logs, it might have been a transient fetching the breakpad code? I'm not sure. I reran and everything was fine.

I've never seen that before in CI, so I'm going to defer thinking about it more. If it pops up again, I'll write up a bug.

self-r+

PR #5496, opened 2020-07-29
Modify Fenix feature versions

rocketsroger:


willkg:
This seems to get rid of all the 5.x version variations. Is that what you intended to do?

rocketsroger:
> This seems to get rid of all the 5.x version variations. Is that what you intended to do?

Yes, now we've released, there will be no more 5.x version variations.

willkg:
Thank you!

PR #5519, opened 2020-08-10
bug 1657668, 1558916: rework product support

willkg:
This reworks product support in Crash Stats. Previously, it was split across a DB table and a bunch of hard-coded stuff across the codebase. Now it's based solely on product settings files in `product_details/` directory.

This makes it easier for other people to submit changes to product settings--just edit the file in GitHub. Further, this reduces maintenance--we don't have to maintain database entries across multiple environments.

One big change here is that `featured_versions` changes now ride the deploys--Crash Stats no longer looks at GitHub for data.

This PR provides:

1. infrastructure for the new `product_details/` -based support
2. a test to make sure changes to product settings files are correct so we can catch issues in CI
3. rework product home page to use the new system and drop the `crashstats_product` table
4. rework featured versions figuring to use the new system (and fix bug 1558916)

To test:

1. (wait for cache to expire)
2. visit home page
3. visit product home pages--make sure featured versions are correct for each one
4. visit supersearch page


requirements/default.txt
```diff
@@ -196,3 +196,6 @@ elasticsearch-dsl==0.0.11 \
 semver==2.10.2 \
     --hash=sha256:21e80ca738975ed513cba859db0a0d2faca2380aef1962f48272ebf9a8a44bd4 \
     --hash=sha256:c0a4a9d1e45557297a722ee9bac3de2ec2ea79016b6ffcaca609b0bc62cf4276
+enforce-typing==1.0.0.post1 \
+    --hash=sha256:90347a61d08e7f7578d9714b4f0fd8abd9b6bc48c8ac8d46d7f290d413afabb7 \
+    --hash=sha256:d3184dfdbfd7f9520c884986561751a6106c57cdd65d730470645d2d40c47e18
```
willkg, line 201: We use this to enforce types data for dataclass instantiations. It's not a great library. It doesn't seem to get much maintenance. There don't seem to be better alternatives a this time.

This is good enough until we need to do nested data structures. At that point, this fails and we might want to switch from Python data classes to pydantic thing.

webapp-django/crashstats/crashstats/jinja2/crashstats/home.html
```diff
@@ -1,5 +1,9 @@
 {% extends "crashstats_base.html" %}
 
+{% block page_title %}
+  Crash Stats Homepage
+{% endblock %}
```
willkg, line 5: This has been messed up for a long while. Figured I'd fix it here since I noticed it while testing.

willkg:
self-r+

I'll test this heavily in stage where we've got more representative data.

PR #5520, opened 2020-08-10
bug 1658023: switch fenix to automatically figure out featured versions

willkg:


willkg:
self-r+

Once this gets to stage, I'll verify it with Roger.

PR #5521, opened 2020-08-10
bug 1658023: rework featured_versions to be based on MAJOR.MINOR

willkg:
Firefox and Fennec have trains based on MAJOR number of the version, but
Fenix has trains based on MAJOR.MINOR. This adjusts the figuring for
featured versions to take that into account.


willkg:
self-r+

I'll verify this on stage where we have more data and run it by Roger.

PR #5522, opened 2020-08-10
bug 1658023: handle Firefox versions correctly

willkg:
This fixed featured versions figuring for Firefox and Fennec where
version numbers can be like 79.0b2.


willkg:
self-r+

This fixes the issue locally. I'll verify it in stage where we have more data.

PR #5523, opened 2020-08-11
bug 1628387: remove google analytics

willkg:
I haven't looked at anything in GA ever. So we don't really use it anymore. This removes it.


willkg:
self-r+

PR #5527, opened 2020-08-18
bug 1659795: add fission indicators to signature/topcrashers reports

willkg:
This adds indicators to the signature report summary and topcrashers
report to indicate if all of the crash reports have DOMFissionEnabled as
well as an indicator if some of the crash reports have
DOMFissionEnabled.

This also removes the remnants of font-awesome stuff we had before which
wasn't in use and replaces it with an updated font-awesome. font-awesome
is used for the fission icon.

### DIFF
--- a/webapp-django/crashstats/crashstats/jinja2/macros/signature_startup_icons.html
+++ b/webapp-django/crashstats/crashstats/jinja2/macros/signature_startup_icons.html
@@ -1,17 +1,25 @@
+{% macro fission_icon() %}
+  <i class="status-icon fas fa-atom" title="Fission Crash"></i>
+{%- endmacro %}
+
+{% macro potential_fission_icon() %}
+  <i class="status-icon fas fa-atom" title="Potential Fission Crash"></i>
+{%- endmacro %}
+
 {% macro startup_crash_icon() %}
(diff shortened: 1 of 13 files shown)


webapp-django/crashstats/crashstats/jinja2/macros/signature_startup_icons.html
```diff
@@ -1,17 +1,25 @@
+{% macro fission_icon() %}
+  <i class="status-icon fas fa-atom" title="Fission Crash"></i>
+{%- endmacro %}
+
+{% macro potential_fission_icon() %}
+  <i class="status-icon fas fa-atom" title="Potential Fission Crash"></i>
+{%- endmacro %}
```
willkg, line 7: I think I'd rather use fontawesome for all the icons rather than involve multiple 3rd party sets. However, that's not something I wanted to deal with in this pass.

So for now, we've got different kinds of icons and they sort of work together.

willkg:
I think this is right. I tweaked some of the "show this icon when ..." conditions. This seems like it makes more sense to me since some of the conditions overlapped.

Anyhow, we'll find out. self-r+

PR #5528, opened 2020-08-18
bug 1659795: change criteria for "potential fission crash"

willkg:
This increases the criteria for "potential fission crash" to "at least
50% of crashes" to reduce false positives.


willkg:
self-r+

PR #5529, opened 2020-08-19
bug 1657465: support creating github issues

willkg:
This reworks the "create a bug" stuff so that each product can specify multiple (text, url template) combinations for where to file bugs. Because it's a url template, it can be a Bugzilla url or a GitHub url or whatever else.

In doing that, I tied the report view to products so I had to fix some tests that had the product hard-coded to something that wasn't in the test fixtures.

This also tweaks `fetch_crashids` so it can take a product--I was getting a little tired of specifying supersearch urls.

### DIFF
--- a/webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
+++ b/webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
@@ -5,7 +5,7 @@
 import datetime
 import json
 import re
-from urllib.parse import urlencode, parse_qs
+from urllib.parse import urlencode, parse_qs, quote_plus
 
 from django_jinja import library
 import humanfriendly
@@ -143,19 +143,17 @@ def show_bug_link(bug_id):
 
 
 @library.global_function
-def bugzilla_submit_url(raw_crash, report, parsed_dump, crashing_thread, bug_product):
-    url = "https://bugzilla.mozilla.org/enter_bug.cgi"
-
+def generate_create_bug_url(template, raw_crash, report, parsed_dump, crashing_thread):
     # Some crashes has the `os_name` but it's null so we
     # fall back on an empty string on it instead. That way the various
     # `.startswith(...)` things we do don't raise an AttributeError.
     op_sys = report.get("os_pretty_version") or report["os_name"] or ""
 
-    # At the time of writing, these pretty versions of the OS name
-    # don't perfectly fit with the drop-down choices that Bugzilla
-    # has in its OS drop-down. So we have to make some adjustments.
+    # At the time of writing, these pretty versions of the OS name don't perfectly fit
+    # with the drop-down choices that Bugzilla has in its OS drop-down. So we have to
+    # make some adjustments.
     if op_sys.startswith("OS X "):
-        op_sys = "Mac OS X"
+        op_sys = "macOS"
     elif op_sys == "Windows 8.1":
         op_sys = "Windows 8"
     elif op_sys in ("Windows Unknown", "Windows 2000"):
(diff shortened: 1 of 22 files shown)


webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -178,33 +176,25 @@ def bugzilla_submit_url(raw_crash, report, parsed_dump, crashing_thread, bug_pro
...
-    if len(kwargs["short_desc"]) > 255:
-        kwargs["short_desc"] = kwargs["short_desc"][: 255 - 3] + "..."
+        kwargs["title"] = "[Fission] " + kwargs["title"]
 
-    # People who are new to bugzilla automatically get the more
-    # basic, "guided format". for entering bugs. This unfortunately
-    # means that all the parameters we pass along gets lost when
-    # the user makes it to the second page. Let's prevent that.
-    # See https://bugzilla.mozilla.org/show_bug.cgi?id=1238212
-    kwargs["format"] = "__default__"
```
willkg, line 204: We don't need this part anymore--it's just part of the template now.

webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -178,33 +176,25 @@ def bugzilla_submit_url(raw_crash, report, parsed_dump, crashing_thread, bug_pro
 
     kwargs = {
         "bug_type": "defect",
-        "keywords": "crash",
-        "product": bug_product,
```
willkg, line 182: These two items are just part of the template now.

webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -143,19 +143,17 @@ def show_bug_link(bug_id):
...
 
-    # At the time of writing, these pretty versions of the OS name
-    # don't perfectly fit with the drop-down choices that Bugzilla
-    # has in its OS drop-down. So we have to make some adjustments.
+    # At the time of writing, these pretty versions of the OS name don't perfectly fit
+    # with the drop-down choices that Bugzilla has in its OS drop-down. So we have to
+    # make some adjustments.
     if op_sys.startswith("OS X "):
-        op_sys = "Mac OS X"
+        op_sys = "macOS"
```
willkg, line 156: When working on this, I discovered that this op sys had changed in Bugzilla--there isn't a "Mac OS X" anymore. So I fixed it.

The rep_platform is wrong, however. I'll write up a bug for that later.

webapp-django/crashstats/productlib.py
```diff
@@ -30,6 +30,10 @@ class Product:
     # entries
     featured_versions: List[int]
 
+    # The list of [link name, link url] bug links for creating new bugs from the report
+    # view of a crash report
+    bug_links: List[List[str]]
```
willkg, line 35: This field probably isn't validated very well. We should double-check the validation in the test.

webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
willkg: https://bugzilla.mozilla.org/show_bug.cgi?id=1659961

willkg:
There's a ton of stuff here to test:

1. Does validation of the `bug_links` in product files work?
   * add a field to the list, verify tests fail
   * change one of the strings to an int, verify tests fail
   * add a typo to a variable name (e.g. `bug_type` -> `bg_type`) and verify tests fail
2. Do all the products have links like what they had before?
   * process a crash for each product and compare links with prod
   * Note that I fixed `op_sys` for Mac OS X so it now works correctly which it doesn't do in prod
   * Note that some products have been disabled in bugzilla, so it's impossible to create bugs for them with existing bug links with the new code or the old code (e.g. Fennec)

I manually tested the first item. We could probably add automated tests for them, too, but we're getting into testing the tests territory and I'm not enthused about that here. I think I'm going to defer adding tests for the tests until it becomes a problem.

I did some testing for the second item and things looked ok. It'll be easier to test in stage where I can compare stage and prod more easily since there's more data. After this lands, we'll want to make a bunch of changes to the product bug links anyhow.

willkg:
self-r+

PR #5530, opened 2020-08-20
bug 1657465: redo Fenix "create a bug" links

willkg:
This redoes the "create a bug" links for Fenix per Roger's suggestions.

Couple of unexciting things:

1. There's no link back to the crash report. With bugzilla, it replaces "bp-xxxxxxxxx" with a link to the crash report on Crash Stats. GitHub doesn't do this (obviously). So we'd need to explicitly create the link. This PR doesn't do that.
2. For Fenix crashes, it'd probably help if device information were included. I'm not sure what information is specifically helpful, though.

I figure we can improve after we get some more feedback.

### DIFF
--- a/product_details/Fenix.json
+++ b/product_details/Fenix.json
@@ -7,23 +7,19 @@
   "bug_links": [
     [
       "Fenix",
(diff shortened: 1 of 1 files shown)


willkg:
self-r+

PR #5531, opened 2020-08-20
bug 1657465: fix crash report link in create-a-bug links

willkg:
To test:

1. process a fenix crash report so you have something to work with
2. look at the report view for that crash report and at the Bugzilla tab (that should get renamed) and click on the create a bug links
3. verify that there is now a crash report url and not the bp-xxxxxx text in the issue description


willkg:
self-r+

PR #5535, opened 2020-09-01
bug 1661605: move minidump-stackwalk to separate repository

willkg:
We moved minidump-stackwalk code and build tools to
https://github.com/mozilla-services/minidump-stackwalk .

When that repo is tagged, it generates a Docker image and posts it to
https://hub.docker.com/r/mozilla/socorro-minidump-stackwalk/tags .

The Socorro build process now pulls the binaries from that image rather
than build stackwalker and friends every time during the build process.

This simplifies building Socorro, building minidump-stackwalk,
maintaining stackwalker, and debugging all the bits.


docker/Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM python:3.8.5-slim@sha256:9c84459051ee7e9d386b24ee5468352d52b6789a5d4a8cf6a649a8a1c6ad5636 as socorro_image_base
+FROM mozilla/socorro-minidump-stackwalk:2020.08.31@sha256:fa842d7a3fd84c01acceba9360562f57f0be8a59a21f90fa788689e2feab63ea as breakpad
```
willkg, line 1: It's easier to add this as a `FROM` so as to get the tag and sha256. You could do `COPY --from=REF`, but then it's deep in the Dockerfile and harder (in my opinion) to find and update. Plus I have no idea if dependabot and friends know what to do with it there.

docker/Dockerfile
```diff
@@ -1,4 +1,6 @@
...
+
+FROM python:3.8.5-slim@sha256:9c84459051ee7e9d386b24ee5468352d52b6789a5d4a8cf6a649a8a1c6ad5636
```
willkg, line 3: We don't need to name this image in the multi-stage because we no longer jump back and forth with it.

willkg:
self-r+

PR #5565, opened 2020-09-02
bug 1473068: add signature urls to OIDC_EXEMPT_URLS

willkg:
This adds the two prefix url matches to `OIDC_EXEMPT_URLS` to cover
signature graphs and aggregations so they don't try to redirect to OIDC
login.


willkg:
self-r+

PR #5566, opened 2020-09-03
bug 1603663: add cleanse_cronlog

willkg:
This adds a `cleanse_cronlog` command and schedules it to run once a week.
It'll delete cron log entries older than 6 months.


willkg:
self-r+

PR #5568, opened 2020-09-09
bug 1663670: handle zlib errors when decompressing memory report

willkg:
To test:

1. process 48c08f87-06c1-4e30-8793-24bf70200908 without this fix and it'll spit out a traceback in the logs, but nothing in the processor notes
2. process 48c08f87-06c1-4e30-8793-24bf70200908 with this fix and it won't spit out anything in the logs, but it will add a processor note

### DIFF
--- a/socorro/processor/rules/mozilla.py
+++ b/socorro/processor/rules/mozilla.py
@@ -8,6 +8,7 @@ import json
 import re
 import time
 from urllib.parse import unquote_plus
+from zlib import error as ZlibError
 
 from configman.dotdict import DotDict
 from glom import glom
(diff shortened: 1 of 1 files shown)


socorro/processor/rules/mozilla.py
```diff
@@ -8,6 +8,7 @@
 import re
 import time
 from urllib.parse import unquote_plus
+from zlib import error as ZlibError
```
willkg, line 11: For some reason, the exception class is `error` which is unhelpful so I named it something better.

willkg:
self-r+

PR #5569, opened 2020-09-10
Improve signature generation docs

willkg:
* cleanup and improve overview
* remove duplicate content
* add philosophy on signature generation which helps guide future
  changes
* add example cases where adding a symbol to the irrelevant list is the
  right thing to do
* fix the pipeline docs generator command to make the rule name bold
  for easier reading


willkg:
self-r+

PR #5573, opened 2020-09-11
bug 1664509: only show Buildhub link if product has buildhub data

willkg:
Most products don't have data on Buildhub. For products that don't, Crash Stats shouldn't show a Buildhub link because it doesn't make any sense.


willkg:
self-r+

PR #5576, opened 2020-09-15
bug 1643339: don't search indices older than retention policy

willkg:
This fixes supersearch so that regardless of the specified date range,
it'll drop the indices that are older than the retention policy.

These indices shouldn't exist, so supersearch would have searched them,
failed, then spent an inordinate amount of time trying to fix the query
by dropping bad indices one at a time.

### DIFF
--- a/socorro/external/es/connection_context.py
+++ b/socorro/external/es/connection_context.py
@@ -114,6 +114,10 @@ class ConnectionContext(RequiredConfig):
         """Return doctype."""
         return self.config.elasticsearch_doctype
 
+    def get_retention_policy(self):
+        """Return retention policy in weeks."""
+        return self.config.retention_policy
+
     def get_timeout_extended(self):
         """Return timeout_extended."""
         return self.config.elasticsearch_timeout_extended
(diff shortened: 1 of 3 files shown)


socorro/external/es/connection_context.py
```diff
@@ -196,15 +200,11 @@ def delete_expired_indices(self):
...
+        cutoff = cutoff.strftime(self.config.elasticsearch_index)
 
         was_deleted = []
         for index_name in self.get_indices():
-            # strptime ignores week numbers if a day isn't specified, so we append
-            # '-1' and '-%w' to specify Monday as the day.
-            index_date = datetime.datetime.strptime(
-                index_name + "-1", self.config.elasticsearch_index + "-%w"
-            )
-            if index_date >= cutoff:
```
willkg, line 207: I reworked this. We can convert the cutoff to a string and then do string comparisons and skip the whole Monday weirdness.

socorro/external/es/supersearch.py
```diff
@@ -169,6 +171,15 @@ def get(self, **kwargs):
         # Find the indices to use to optimize the elasticsearch query.
         indices = self.get_indices(params["date"])
 
+        if "%" in self.context.get_index_template():
```
willkg, line 174: If there's no `%` in the template, then it's not date based and the cutoff isn't pertinent.

At the moment, this is only an issue with one of the tests. I figured it's easier to add this check here than try to figure out whether we should support non-date-based templates because that has big ramifications on test setup and a bunch of other stuff.

socorro/unittest/external/es/test_supersearch.py
```diff
@@ -1435,11 +1435,19 @@ def test_get_with_beta_version(self):
...
+            end = end - datetime.timedelta(days=1)
+        start = end - datetime.timedelta(days=10)
+        end = end.strftime("%Y-%m-%dT%H:%M:%S")
+        start = start.strftime("%Y-%m-%dT%H:%M:%S")
         config = self.get_base_config(
             cls=SuperSearchWithFields, es_index="socorro_test_reports_%W"
         )
         api = SuperSearchWithFields(config=config)
-        params = {"date": [">2000-01-01T00:00:00", "<2000-01-10T00:00:00"]}
+        params = {"date": [">%s" % start, "<%s" % end]}
```
willkg, line 1450: With the new code, indices between 2000-01-01 and 2000-01-10 would get dropped since they're _way_ before the retention policy.

willkg:
self-r+

PR #5577, opened 2020-09-15
bug 1665126: add note about user-agent

willkg:
We really want users to be specifying a User-Agent if they're using the APIs. It helps us help them. This fixes the docs to specify that.

Further, it fixes the supersearch documentation to work better when there's no version data.

While I was there, I also made some updates to the system checklist.


willkg:
self-r+

PR #5578, opened 2020-09-17
bug 1641059: update stackwalker to 2020.09.17

willkg:


willkg:
This updates to stackwalker 2020.09.17 which picks up some compilation optimizations and is (according to my menial benchmarks) mildly faster.

PR #5585, opened 2020-09-30
bug 1668381: add _XReply to irrelevant list

willkg:
_XReply is an x11 function that waits for a reply packet and triggers an
error if it gets back an XError which is what we see in crash reports,
so it's not very interesting.


willkg:
self-r+

PR #5594, opened 2020-10-01
bug 1668581: redo test scaffolding

willkg:
This redoes the test scaffolding to use a service defined in
docker-compose that doesn't volume mount the local directory to run the
tests in when running in CI. Previously, we had this whole setup where
we'd copy the local directory into a data container and then mount from
that and it was complex and ended up being harder to maintain.

To test:

1. run `make test`
2. run `make lint`
3. run `make test-ci`


willkg:
self-r+

PR #5595, opened 2020-10-06
bug 1668517: switch to pip-compile

willkg:
This switches from having a ``requirements/default.txt`` file and a
``requirements/constraints.txt`` file to having a ``requirements.in`` file which
compiles to a ``requirements.txt`` file with pip-compile.

This solves a bunch of irritating maintenance issues and should reduce
overall load and make it easier to switch to the new pip resolver.


willkg:
self-r+

PR #5596, opened 2020-10-06
bug 1669163: fix formatting for reason/moz_crash_reason in bug links

willkg:
`MozCrashReason` only gets generated in cases where the `reason` is `SIGSEGV` or `EXCEPTION_BREAKPOINT` so the reason isn't helpful. This changes the template to show `MozCrashReason` or `Reason` or nothing.

Bugzilla comments use Markdown for formatting, so we need to put the `MozCrashReason`/`Reason` in backticks so the values aren't kicking off Markdown formatting inadvertently.

To test:

1. process a crash report with a `Reason`, but no `MozCrashReason` and check bug link--should show `Reason` in backticks
2. process a crash report with a `Reason` and a `MozCrashReason` and check bug link--should show `MozCrashReason` in backticks


willkg:
self-r+

PR #5597, opened 2020-10-06
Set docker version in CircleCI config

willkg:
This explicitly sets the Docker version. If we don't set the version,
CircleCI defaults to using 17.03.0 which is a version that CircleCI is
deprecating.

https://discuss.circleci.com/t/old-linux-machine-image-remote-docker-deprecation/37572


relud (approved):

willkg:
Thank you!

PR #5599, opened 2020-10-19
bug 1669163: fix reason in bug report creation

willkg:
The reason was being escaped by the bug creation template, but it
shouldn't be. This marks the relevant fields as safe.


willkg:
Note: This template is used to build a bug description which is used to build a link. It's not rendered in an HTML context at any point.

self-r+

PR #5609, opened 2020-11-09
bug 1669247: add dockerhub auth

willkg:
When doing things in CircleCI that interact with docker hub, we want to
log in so we're not affected by rate-limiting.


willkg:
This worked in the PR. The docker hub credentials aren't available in the CircleCI environment when running for a PR based on a branch in a fork. That's perfect!

self-r+

PR #5612, opened 2020-11-11
bug 1676654: update node.js and js libs

willkg:
We were using an old version of node. This updates it to a more recent one. Socorro uses node to compile and transform static assets.

This updates most of the js libraries with a couple of exceptions:
    
1. d3: Socorro is using uglify to minify JS and it doesn't support ES6
   and kicks up a parser error when working on d3 that looks like an ES6
   thing. I think we need to switch from uglify to something else, but
   that's a much bigger project than I want to deal with now, so I'm
   pushing off updating d3 for now.
    
2. select2: Updating to a recent version kicks up type errors. I think
   we have to update our code. I pushed that off to a new bug.

To test, build it locally, process some crashes so you have things to see in the signature report and supersearch, and then go through web-pages.


webapp-django/package.json
```diff
@@ -10,31 +10,31 @@
...
+    "less": "3.12.2",
+    "prettier": "2.1.2",
+    "uglify-js": "3.11.5"
   },
   "dependencies": {
     "@sentry/cli": "^1.40.0",
     "Select2": "github:select2/select2#3.5.4",
-    "ace-builds": "1.3.3",
+    "ace-builds": "1.4.12",
     "d3": "5.1.0",
```
willkg, line 25: I couldn't update d3 because uglify doesn't like ES6. I wrote up a bug for that.

webapp-django/package.json
```diff
@@ -10,31 +10,31 @@
...
+    "eslint-config-prettier": "6.15.0",
+    "eslint-plugin-prettier": "3.1.4",
+    "graceful-fs": "4.2.4",
+    "less": "3.12.2",
+    "prettier": "2.1.2",
+    "uglify-js": "3.11.5"
   },
   "dependencies": {
     "@sentry/cli": "^1.40.0",
     "Select2": "github:select2/select2#3.5.4",
```
willkg, line 23: I couldn't update Select2. I think we have to make some code changes to make it work. I wrote up a bug for that.

webapp-django/crashstats/crashstats/static/crashstats/js/polyfill/fetch.js
```diff
@@ -85,54 +85,54 @@
...
 
-  Headers.prototype.getAll = function(name) {
+  Headers.prototype.getAll = function (name) {
     return this.map[normalizeName(name)] || [];
   };
 
-  Headers.prototype.has = function(name) {
-    return this.map.hasOwnProperty(normalizeName(name));
+  Headers.prototype.has = function (name) {
+    return Object.prototype.hasOwnProperty.call(this.map, normalizeName(name));
```
willkg, line 102: I fixed all of these per the eslint docs' suggestion.

willkg:
I'll test this more thoroughly on stage where there's more data. Seems ok so far, though.

PR #5615, opened 2020-11-18
bug 1676593: fix adapter_driver_version sorting in signature report

willkg:
jquery-tablesorter gets the value type wrong some how and thus sorts the values in the column wrong. If we "hard-code" it to be "text", then it sorts alphabetically which is correct.

There are a bunch of tables in Crash Stats, but I'm fixing this one specifically figuring we can generalize later if that's appropriate.


willkg:
@rehandalal: Can you look this over for "Oh dear--don't do that in JS." kinds of things?

rehandalal:
Besides the mild panic attack of having to read jQuery code nothing looks egregious here.

willkg:
Thank you!

PR #5625, opened 2020-12-08
bug 1681388: clean up protected data language and links

willkg:
This cleans up the language about protected data near things like minidump links. The goal is to make it much clearer what is expected of the user in regards to protected data.

This also switches out the old terms "personal data" for the new term "protected data" in the data policy.


willkg:
self-r+

PR #5627, opened 2020-12-09
bug 1678739: make remote_type public

willkg:
I can't tell why `remote_type` was originally protected data. The values in that field suggest it's fine to make public and it's a public field in the crash ping data set so I'm making it public in Crash Stats.


willkg:
self-r+

PR #5629, opened 2020-12-09
bug 1681388: specify minidump sharing policy near links

willkg:
This fixes the raw info and minidump links so it's clearer what they
link to--it always confuses me and I work on the site.

This also modifies the note that minidumps are protected data by
additionally stating they should not be shared with third-parties.


willkg:
self-r+

PR #5632, opened 2020-12-10
bug 1681390: add Hackers Plus users

willkg:
This fixes the protected data users admin page to show users in the
Hackers group and the Hackers Plus group.

### DIFF
--- a/webapp-django/crashstats/manage/admin.py
+++ b/webapp-django/crashstats/manage/admin.py
@@ -134,15 +134,25 @@ def graphics_devices(request):
 
 @superuser_required
 def protected_data_users(request):
-    # Get all users in the "Hackers" group
+    email_addresses = []
+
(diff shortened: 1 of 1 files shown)


webapp-django/crashstats/manage/admin.py
```diff
@@ -134,15 +134,27 @@ def graphics_devices(request):
...
         hackers_group = []
 
-    email_addresses = [
-        user.email for user in hackers_group.user_set.all() if user.is_active
-    ]
+    email_addresses.extend(
+        [user.email for user in hackers_group.user_set.all() if user.is_active]
+    )
+
+    # Get all users in the "Hackers" group
```
willkg, line 149: Bah. Bad comment.

willkg:
self-r+

PR #5644, opened 2021-01-07
bug 1685178: fix signature generation for unknown in dll frames

willkg:
We changed `dump_syms` to yield a better symbol for cases where we have a
dll without symbols. The better symbol is structured like:

```
<unknown in foo.dll>
```

The `< >` triggers the collapsing types in the CSignatureTool
normalization code. This adds an exception so that these frames don't
get converted to `<T>`.

### DIFF
--- a/socorro/signature/rules.py
+++ b/socorro/signature/rules.py
@@ -163,12 +163,15 @@ class CSignatureTool(SignatureTool):
         function = function.replace("`anonymous namespace'", "(anonymous namespace)")
 
         # Collapse types
+        #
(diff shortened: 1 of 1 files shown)


willkg:
```
app@socorro:/app$ socorro-cmd signature e1ff24f1-27cc-4134-aae7-339570201227
Crash id: e1ff24f1-27cc-4134-aae7-339570201227
Original: <T>
New:      <unknown in msmpeg2vdec.dll> | BaseThreadInitThunk
Same?:    False
```

PR #5645, opened 2021-01-07
bug 1684164: add application_build_id to search list

willkg:
Fenix puts the GeckoView build id in `BuildID` and the Fenix build id in
`ApplicationBuildID` in crash reports. The GeckoView `BuildID` is
well-formed--it's a long in the form of `YYYYMMDDHHMMSS`. The Fenix
`ApplicationBuildID` is a monotonically increasing number, but isn't
structured on a datestamp. We can treat it like a build id and let
people search with it as if it's a long.

This adds it as a public field that's searchable.

This also adds it to the Details and Metadata tabs of the report view.


willkg:
This adds a super search field and a processor rule to populate the field that looks at. So it won't work with everything we've got already--only crash reports received going forward.

I'll land this and let it sit on stage for a couple of days and then test it, but I might have to wait until next week when a new index with the new mapping is created.

socorro/external/es/super_search_fields.py
```diff
@@ -379,6 +379,20 @@ def get(self):
...
+        "form_field_choices": [],
+        "has_full_version": False,
+        "in_database_name": "application_build_id",
+        "is_exposed": True,
+        "is_returned": True,
+        "name": "application_build_id",
+        "namespace": "processed_crash",
+        "permissions_needed": [],
+        "query_type": "number",
+        "storage_mapping": {"type": "long"},
```
willkg, line 394: This is soooooo much "string" boilerplate. Some day it'd be great to fix that.

This is structured the same as `build_id`. I figured users are likely to want to do the same things with it.

socorro/processor/rules/mozilla.py
```diff
@@ -87,6 +89,12 @@ def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
         processed_crash["ReleaseChannel"] = raw_crash.get("ReleaseChannel", "")
         processed_crash["build"] = raw_crash.get("BuildID", "")
 
+        # NOTE(willkg): ApplicationBuildID is for Fenix which sends the gecko view build
+        # id in BuildID.
```
willkg, line 93: Fenix is the only product that does this, but it's possible other products may also do this going forward. I don't remember where the data that populates `ApplicationBuildID` comes from specifically. We can adjust the comment when we know more.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -188,6 +188,14 @@ <h2>{{ report.product }} {{ report.version }} Crash Report [@ {{ report.signatur
...
                   </td>
                 </tr>
+                {% if report.product == "Fenix" %}
+                  <tr title="{{ fields_desc['processed_crash.application_build_id'] }}">
+                    <th scope="row">Application Build ID</th>
+                    <td>
+                      {{ report.application_build_id }}
+                    </td>
+                  </tr>
+                {% endif %}
```
willkg, line 198: The ApplicationBuildID isn't structured the same as BuildID, plus Fenix isn't in Buildhub. So there's no point in showing a Buildhub link or extracting a date from the value.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -699,6 +699,7 @@ class RawCrash(SocorroMiddleware):
         "AdapterRendererIDs",
         "AdapterSubsysID",
         "AdapterVendorID",
+        "ApplicationBuildID",
```
willkg, line 702: Adding this here makes the field show up in the Metadata tab of the crash report.

PR #5646, opened 2021-01-07
bug 1674791: update to Python 3.9.1

willkg:
In the previous attempt to upgrade from Python 3.8 to 3.9, I hit
problems with building lxml and sasl. Since then, the lxml problem went
away.

sasl is a dependency of pyhs2. That library is both unused (we don't use
Hive) and unmaintained (since 2016). I removed that which removed the
sasl and thrift dependencies.

That unblocked upgrading to Python 3.9.


willkg:
Since we're not building lxml (comes as a wheel now) and sasl (nixed from requirements list), builds should take less time now.

PR #5650, opened 2021-01-08
bug 1671276: process breadcrumbs and show on details page

willkg:
The crash reporter in android-components which Fenix uses captures
breadcrumbs and puts it in the Breadcrumbs annotation.

This adds a processor rule to validate it and add it to the processed
crash and then display it in the Details of the report view.

This marks it as protected data, so if you don't have access to that,
you can't see it.

This doesn't make it searchable. I decided not to deal with that until I
know more about how people want to search it (or even if people do want
to search it).

### DIFF
--- a/socorro/processor/rules/mozilla.py
+++ b/socorro/processor/rules/mozilla.py
@@ -267,6 +267,8 @@ class DatesAndTimesRule(Rule):
 
 
 class JavaProcessRule(Rule):
+    """Move Java-crash-related bits over."""
+
     def predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
         return bool(raw_crash.get("JavaStackTrace", None))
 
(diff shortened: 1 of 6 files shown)


socorro/unittest/processor/rules/test_mozilla.py
```diff
@@ -674,6 +676,30 @@ def test_bad_seconds_since_last_crash(self):
...
+)
+def test_validate_breadcrumbs(data, expected):
+    if expected is VALID:
+        validate_breadcrumbs(data)
+
+    else:
+        with pytest.raises(MalformedBreadcrumbs, match=expected):
+            validate_breadcrumbs(data)
+
+
```
willkg, line 702: Dammit, @willkg! You forgot to test the rule!

socorro/processor/rules/mozilla.py
```diff
@@ -288,6 +290,67 @@ def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
+    for i, item in enumerate(data):
+        if not isinstance(item, dict):
+            raise MalformedBreadcrumbs(f"item {i} not a dict")
+
+        # missing_keys is the required_keys minus the intersection of required_keys and
+        # the item's keys
+        missing_keys = required_keys - (required_keys & set(item.keys()))
+        if missing_keys:
+            missing_keys = ", ".join(sorted(missing_keys))
+            raise MalformedBreadcrumbs(f"item {i} missing keys: {missing_keys}")
```
willkg, line 323: At some point, I want to do a bunch of schema stuff and when I do that, we'll redo this with that.

webapp-django/crashstats/crashstats/static/crashstats/css/base/tables.less
```diff
@@ -63,54 +63,6 @@ table {
         }
     }
 
-    &.captioned-data-table {
```
willkg, line 66: This isn't used anywhere, so I nixed it.

PR #5651, opened 2021-01-11
bug 1643339: re-fix cutoff policy code

willkg:
When someone does a supersearch and specifies a date range, we don't
want to try to search indices that are outside the retention policy. In
September, I wrote some code that I thought correctly removed indices
that weren't valid, but I didn't think about how our index templates
include a week number, but not a year.

This fixes that code so it's not longer doing bad comparisons with index
names and instead figures out the set of valid indices and intersects
that with the set getting searched.

This is the correct fix for the test failure I saw last week.

### DIFF
--- a/socorro/external/es/supersearch.py
+++ b/socorro/external/es/supersearch.py
@@ -22,6 +22,31 @@ ELASTICSEARCH_PARSE_EXCEPTION_REGEX = re.compile(
 )
 
 
+def prune_invalid_indices(indices, policy, template):
+    """Prunes given indices by ones that were prior to the cutoff
+
+    :arg list indices: list of indices to prune
+    :arg int policy: number of days to retain indices--prior to this is the cutoff
+    :arg str template: index template
+
+    :returns: list of valid indices
+
+    """
+    now = utc_now()
+    marker = now - policy
+
+    # Generate the set of valid indexes between the cutoff and now.
+    valid_indices = set()
+    while marker <= now:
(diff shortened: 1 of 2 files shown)


socorro/external/es/supersearch.py
```diff
@@ -22,6 +22,31 @@
...
+    now = utc_now()
+    marker = now - policy
+
+    # Generate the set of valid indexes between the cutoff and now.
+    valid_indices = set()
+    while marker <= now:
+        valid_indices.add(marker.strftime(template))
+        # NOTE(willkg): have to increment by 1 so we pick up week numbers for weeks
+        # that span a new year
+        marker = marker + datetime.timedelta(days=1)
```
willkg, line 44: We could cache the set of valid indices for the day, but I didn't want to spend time thinking about testing the caching code. We can do that in a future PR.

socorro/unittest/external/es/test_supersearch.py
```diff
@@ -1452,9 +1452,7 @@ def test_get_against_nonexistent_index(self):
         res = api.get(**params)
         assert res["total"] == 0
         assert len(res["hits"]) == 0
-        # NOTE(willkg): There are at least two missing indexes--we don't really care
-        # about how many there are as long as it detected there are missing indexes
-        assert len(res["errors"]) >= 2
+        assert len(res["errors"]) == 3  # 3 weeks are missing
```
willkg, line 1455: This returns the test to the original form I did in September.

PR #5653, opened 2021-01-12
Fix elasticsearch tests

willkg:
Did a pass on cleaning up the Elasticsearch tests.

* fixed test setup and test teardown to create and delete test indices
* switched the default index template to include a %W
* removed unnecessary mocking
* moved some methods around to better easier-to-access places
* fixed index_crash and index_many_crashes to use crashstorage so
  they're not wildly divergent from how we index crashes
* added some docstrings and comments clarifying what's going on
* some other cleanup

I extracted this from my work on bug 1686103 to make that an easier PR to get through.


willkg:
That test passes for me locally, so I'll have to figure out what's going on. It's probably a race condition with the state of ES. Boo hiss.

socorro/external/es/connection_context.py
```diff
@@ -110,6 +110,20 @@ def get_index_template(self):
...
+        Index names are generated by filling in the date parameters for the
+        specified date in the index template.
+
+        :arg datetime date: the datetime to use to generate the index template
+
+        :returns: index name as a str
+
+        """
+        template = self.get_index_template()
+        return date.strftime(template)
```
willkg, line 125: A different version of this was in crashstorage. I moved it to here and generalized it because it makes writing tests a lot easier.

socorro/external/es/super_search_fields.py
```diff
@@ -20,6 +20,34 @@
...
+            field_full_name = ".".join((namespace, key))
+        else:
+            field_full_name = key
+
+        if "properties" in field:
+            fields.update(parse_mapping(field["properties"], field_full_name))
+        else:
+            fields.add(field_full_name)
+
+    return fields
```
willkg, line 48: This function was defined in the function that it was being used in. I extracted it to make it easier to use and test.

socorro/unittest/external/es/base.py
```diff
@@ -50,26 +50,13 @@ def test_with_version(self):
...
-    """Convenience class for SuperSearch to pass all fields
-
-    SuperSearch's get method requires to be passed the list of all fields. This
-    class does that automatically so we can just use ``get()``.
-
-    """
-
-    def get(self, **kwargs):
-        kwargs["_fields"] = copy.deepcopy(FIELDS)
-        return super().get(**kwargs)
```
willkg, line 63: Most of the tests that used this didn't really need it and I think it moves important information about the test too far from the test. So I moved this to `test_supersearch.py` where this is helpful since that file is testing Supersearch queries only.

For the rest of the files using this, I changed the `api.get()` calls to pass in `_fields`.

socorro/unittest/external/es/base.py
```diff
@@ -100,25 +87,36 @@ def get_tuned_config(self, sources, extra_values=None):
...
+        ]
+        for index_name in to_create:
+            print(f"setup: creating index: {index_name}")
+            self.es_context.create_index(index_name)
+
+    def teardown_method(self):
+        for index_name in self.es_context.get_indices():
+            print(f"teardown: delete test index: {index_name}")
+            self.es_context.delete_index(index_name)
+        super().teardown_method()
```
willkg, line 116: I changed the `elasticsearch_index` variable to `testsocorro_%W` and added an appropriate `elasticsearch_index_regex` so that all the tests use that and it makes cleanup much easier and more correct.

The next evolution of this code is probably to add an elasticsearch pytest fixture that does this instead of embedding it all mysteriously in the test case class.

socorro/unittest/external/es/base.py
```diff
@@ -138,55 +136,46 @@ def get_base_config(self, cls=ConnectionContext, es_index=None):
 
         return self.get_tuned_config(cls, extra_values=extra_values)
 
-    def index_crash(self, processed_crash=None, raw_crash=None, crash_id=None):
+    def index_crash(
+        self, processed_crash=None, raw_crash=None, crash_id=None, refresh=True
+    ):
+        """Index a single crash and refresh"""
```
willkg, line 142: This used to access Elasticsearch directly which means that indexing crashes didn't go through our crash indexing code and thus the tests were pretty seriously diverged from reality.

I reworked this and `index_crash_bulk` to use the crashstorage code to save crashes.

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -132,30 +133,27 @@ def test_invalid_key(self, key):
...
+    def setup_method(self):
+        super().setup_method()
         self.config = self.get_tuned_config(ESCrashStorage)
 
-        # Helpers for interacting with ES outside of the context of a
-        # specific test.
-        self.es_client = elasticsearch.Elasticsearch(
-            hosts=self.config.elasticsearch.elasticsearch_urls
-        )
-        self.index_client = elasticsearch.client.IndicesClient(self.es_client)
```
willkg, line 144: This duplicates things that should already be set up via superclasses.

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -132,30 +133,27 @@ def test_invalid_key(self, key):
...
         es_storage.save_processed_crash(
-            raw_crash=deepcopy(a_raw_crash),
-            processed_crash=deepcopy(a_processed_crash),
+            raw_crash=raw_crash,
+            processed_crash=processed_crash,
         )
 
         # Ensure that the document was indexed by attempting to retreive it.
-        assert self.es_client.get(
-            index=es_storage.es_context.get_index_template(),
```
willkg, line 157: Previously, the template variable didn't include any date placeholders, so you didn't need to interpolate it. That's not true now, so I switched it to use `get_index_for_date()`.

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -179,65 +177,27 @@ def test_index_crash_indexable_keys(self):
...
-            index=self.config.elasticsearch.elasticsearch_index,
+        doc = self.conn.get(
+            index=self.es_context.get_index_for_date(utc_now()),
             id=processed_crash["uuid"],
         )
 
         # Verify keys that aren't in super_search_fields aren't in the raw or processed
         # crash parts
         raw_crash = doc["_source"]["raw_crash"]
-        assert raw_crash == {"BuildID": "20200506000000"}
```
willkg, line 190: These assertions are testing keys and whether keys that should be removed were removed. I changed the code to ignore the values.

socorro/unittest/external/es/test_query.py
```diff
@@ -24,20 +23,23 @@
...
+        )
+        self.index_crash(
+            processed_crash={"date_processed": datestamp, "product": "EarthRaccoon"}
+        )
         self.es_context.refresh()
 
         query = {"query": {"match_all": {}}}
         res = self.api.get(query=query)
-        assert res
-        assert "hits" in res
```
willkg, line 40: These are unneeded. Asserting `res["hits"]["total"]` implicitly asserts these other two things.

socorro/unittest/external/es/test_query.py
```diff
@@ -49,30 +51,17 @@ def test_get(self):
             }
         }
         res = self.api.get(query=query)
-        assert res
-        assert "hits" in res
         assert res["hits"]["total"] == 1
 
-    @mock.patch("socorro.external.es.connection_context.elasticsearch")
```
willkg, line 56: I nixed the mocking--we can implement this test without it and it makes the test more useful.

socorro/unittest/external/es/test_super_search_fields.py
```diff
@@ -40,93 +40,95 @@ def test_get_fields(self):
...
+        expected = [
+            "field_z",
+            "namespace1.field_a",
+            "namespace1.field_b",
+            "namespace1.subspace1.field_d",
+            "namespace2.subspace1.field_b",
+        ]
+
+        assert missing_fields["hits"] == expected
+        assert missing_fields["total"] == 5
```
willkg, line 131: We're doing deletion in teardown now. The rest of these changes are mostly dedenting and making the test clearer without iterating over a range of 0, 1.

PR #5655, opened 2021-01-14
bug 1686103: don't index fields that aren't in the mapping

willkg:
A while back, I changed `save_processed_crash` to drop any keys that weren't in super_search_fields. That works fine if the index mapping matches super_search_fields. However, it fails whenever we add a new field and deploy--all the indexes will not have the new field. With the current code, the new field will get indexed, ES will guess at the storage_mapping, and make it a string. That breaks aggregations going forward if the storage_mapping for that field is different than what it is in super_search_fields.

This changes the `save_processed_crash` code to drop any keys that aren't in super_search_fields OR aren't in the mapping for the index the crash will be indexed in. This fixes the issues we're having.

One consequence of this is that when adding a new field, it will have no effect until a new index with the new mapping is created over the weekend.

### DIFF
--- a/socorro/scripts/es.py
+++ b/socorro/scripts/es.py
@@ -8,11 +8,13 @@ indices.
 """
 
 import datetime
+from pprint import pprint
 
 import click
 from configman import ConfigurationManager
 from configman.environment import environment
 from elasticsearch_dsl import Search
+from elasticsearch.client import IndicesClient
 
 from socorro.external.es.base import generate_list_of_indexes
 from socorro.external.es.connection_context import ConnectionContext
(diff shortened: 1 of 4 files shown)


socorro/scripts/es.py
```diff
@@ -104,6 +106,37 @@ def cmd_list_crashids(ctx, index):
...
+    conn = context.connection()
+    search = Search(
+        using=conn,
+        index=index,
+        doc_type=context.get_doctype(),
+    )
+    search = search.query("match", crash_id=crashid)
+    results = search.execute()
+    for item in results:
+        pprint(item.to_dict())
```
willkg, line 137: These two new commands make it easier to see what's going on in ES land.

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -190,6 +192,84 @@ def test_index_crash_indexable_keys(self):
...
+        Indexing a crash that has keys that aren't in the mapping for the index
+        should cause those keys to be removed from the crash.
+
+        """
+        # The test harness creates an index for this week and last week. So let's create
+        # one for 4 weeks ago.
+        now = utc_now()
+        four_weeks_ago = now - timedelta(days=28)
+
+        field = "user_comments"
```
willkg, line 207: I used this field because I'm pretty sure it'll be around for a while. If we ever drop this field, we'll need to update this test.

socorro/unittest/external/es/test_crashstorage.py
```diff
@@ -190,6 +192,84 @@ def test_index_crash_indexable_keys(self):
...
+        es_storage = ESCrashStorage(config=self.config)
+
+        # Create a crash for this week and save it
+        now_uuid = "00000000-0000-0000-0000-000000120408"
+        raw_crash = {
+            "BuildID": "20200506000000",
+        }
+        processed_crash = {
+            field: "this week",
+            "date_processed": date_to_string(now),
```
willkg, line 230: The ESCrashStorage figures out which index to put the crash in based on the `processed_crash["date_processed"]` field which is added by the processor.

PR #5656, opened 2021-01-20
bug 1682931: reduce JavaStackTrace confusion

willkg:
If you don't have protected data access, then on the Details tab, you
see the "JavaStackTrace (redacted)" key with the java_stack_trace value.
It doesn't show up at all on the Metadata tab.

If you do have protected data access, then on the Details tab, you see
JavaStackTrace (Raw) key with the java_stack_trace_raw value and on the
Metadata tab, you see JavaStackTrace with the java_stack_trace_raw
value.

If you have protected data access, this is confusing because you're
seeing both the raw value in the Details tab and the sanitized value in
the Metadata tab and they have the same value which makes it seem like
Crash Stats is not sanitizing the value correctly.

This fixes that by changing the "Metadata" tab to "Crash annotations",
splitting that data into "public" and "protected" sections, and adding
the "protected data" label to the protected data section.

Further, on the Details tab, I added the "protected data" label to
all the items that are protected data including the "Java Stack Trace
(raw)" field if it's showing.

I also improved the description for "Java Stack Trace (raw)" so it
notes that it's protected data.

I think this clarifies what value you're seeing and what you can do with
it. Further, I think this makes the "Metadata" tab a lot clearer along
with a link to the CrashAnnotations.yaml document that describes the
fields there.

### DIFF
--- a/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
+++ b/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
@@ -77,7 +77,7 @@
               <a href="#details" class="ui-tabs-anchor"><span>Details</span></a>
             </li>
             <li class="ui-state-default ui-corner-top">
-              <a href="#metadata" class="ui-tabs-anchor"><span>Metadata</span></a>
+              <a href="#annotations" class="ui-tabs-anchor"><span>Crash Annotations</span></a>
             </li>
             <li class="ui-state-default ui-corner-top">
               <a href="#bugzilla" class="ui-tabs-anchor"><span>Bugzilla</span></a>
@@ -108,7 +108,7 @@
             <table class="record data-table vertical hardwrapped">
               <tbody>
 
-                {# Crash report metadata #}
+                {# Crash Annotations #}
 
                 <tr title="{{ fields_desc['processed_crash.signature'] }}">
                   <th scope="row">Signature</th>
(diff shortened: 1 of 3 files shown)


webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -77,7 +77,7 @@ <h2>{{ report.product }} {{ report.version }} Crash Report [@ {{ report.signatur
               <a href="#details" class="ui-tabs-anchor"><span>Details</span></a>
             </li>
             <li class="ui-state-default ui-corner-top">
-              <a href="#metadata" class="ui-tabs-anchor"><span>Metadata</span></a>
+              <a href="#annotations" class="ui-tabs-anchor"><span>Crash Annotations</span></a>
```
willkg, line 80: This is so much clearer than "metadata".

It does break links and users will be confused for a bit. I could add a note about that, but I think I'm going to wait until someone actually has confusion before doing anything about it.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -806,10 +818,21 @@ <h2>Thread {{ thread.thread }}{% if thread.thread_name %}, Name: {{ thread.threa
           </div>
           <!-- /details -->
 
-          <div id="metadata" class="ui-tabs-hide">
+          <div id="annotations" class="ui-tabs-hide">
+            <div>
+              These are the crash annotations from the submitted crash report. You
+              can see the annotations and descriptions at <a href="https://hg.mozilla.org/mozilla-central/file/tip/toolkit/crashreporter/CrashAnnotations.yaml">https://hg.mozilla.org/mozilla-central/file/tip/toolkit/crashreporter/CrashAnnotations.yaml</a>.
```
willkg, line 824: I think telling people where the data comes from helps bridge the "what is this and how is it calculated" gap we have now. I think we can do a lot more on this front, though.

PR #5658, opened 2021-01-23
bug 1687987: stop using removed fields

willkg:
I did a pass on the collector and removed fields it was adding to the
raw crash that we didn't need. This implements the other side of that
and removes those fields from super search and the processed crash.

This removes `type_tag`, `legacy_processing`, `throttle_rate`, and `timestamp`.

This also reworks the `DatesAndTimesRule` processor rule which used
timestamp to now use submitted_timestamp. I also had to fix the tests
because they're using fictitious data which doesn't make sense and when
I removed timestamp, that broke the tests.

I also switched from using
`socorro.lib.datetimeutil.datetime_from_isodate_string` to using
`datetime.datetime.fromisoformat` which is new in Python 3.7 and is the
inverse of `datetime.datetime.isoformat()` which we use to generate these
values.


socorro/unittest/processor/rules/test_mozilla.py
```diff
@@ -534,50 +508,26 @@ def test_bad_timestamp_and_no_crash_time(self):
...
-        expected = datetime_from_isodate_string("1970-01-01 00:00:00+00:00")
-        assert processed_crash["client_crash_date"] == expected
-        assert processed_crash["install_age"] == -1335439892
-        assert processed_crash["uptime"] == 0
+        assert processed_crash["crash_time"] == expected_timestamp
+        assert processed_crash["client_crash_date"] == (
+            datetime.datetime.fromtimestamp(expected_timestamp, UTC)
+        )
+        assert processed_crash["install_age"] == 1079701
+        assert processed_crash["uptime"] == 20155
```
willkg, line 522: "uptime" changes here because originally it was based on things based on "timestamp" and we swapped "timestamp" out for "submitted_timestamp".

PR #5659, opened 2021-01-26
bug 1675560: show java exception information

willkg:
Socorro gets a `JavaStackTrace` unstructured text blob which gets shown on the Details tab of the report view. Because it's unstructured, there isn't much we can do with it without parsing it which I can't really do well.

Now we also have `JavaException` which is a structured form of the Java exception.

This adds code for processing that in the processor so we have the forms we want and the data is validated. This also adds code to display that in the report view in the Details tab above the breadcrumbs.


willkg:
To test:

1. process a java crash
2. view the Details tab of the crash report page; the exception message should be REDACTED
3. log in with an account that has protected data access
4. view the Details tab of the crash report page; the exception message should be whatever the message is

socorro/lib/javautil.py
```diff
@@ -102,3 +107,38 @@ def parse_java_stack_trace(text):
...
+    :arg dict data: the JavaException value
+
+    :returns: a deep-copied copy of the data with PII redacted
+
+    """
+    data = copy.deepcopy(data)
+    for item in data["exception"]["values"]:
+        stacktrace = item["stacktrace"]
+        if "value" in stacktrace:
+            stacktrace["value"] = "REDACTED"
```
willkg, line 143: This is the exception message. It can contain PII in cases where the exception is a problem with substring or something like that. Then the message contains the string it was working on and that string could be a url.

socorro/schemas/java_exception.json
```diff
@@ -0,0 +1,75 @@
...
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}
```
willkg, line 75: I fixed the indentation to 2 spaces.

socorro/schemas/telemetry_socorro_crash.json
```diff
@@ -0,0 +1,1122 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
```
willkg, line 2: I changed the indentation to 2 spaces and renamed the file to something more specific. Otherwise, it's identical.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -691,44 +691,46 @@ <h2>{{ report.product }} {{ report.version }} Crash Report [@ {{ report.signatur
...
-            {% if (request.user.has_perm('crashstats.view_pii') or your_crash) and report.breadcrumbs %}
-              <div id="breadcrumbs">
-                <h2>Breadcrumbs</h2>
-                <p>
-                  {{ protected_warning(your_crash) }} -
-                  {% if report.breadcrumbs|length > 10 %}
-                    Latest 10 breadcrumbs (out of {{ report.breadcrumbs|length }})
-                  {% else %}
-                    All {{ report.breadcrumbs|length }} breadcrumbs
+            {% if java_exception_stacks %}
```
willkg, line 694: I moved breadcrumbs to below showing stack data.

I modeled showing stacks here to be like showing crashing thread stacks.

I think it's kind of gross looking, but Socorro doesn't have good primitives for this (imo) and I didn't want to figure that out in addition to everything else. I think it behooves us to do a ui/ux refresh and deal with it in that.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -691,44 +691,46 @@ <h2>{{ report.product }} {{ report.version }} Crash Report [@ {{ report.signatur
...
+
+                  {% if stacktrace.value %}
+                    <h3>
+                      Message: {{ stacktrace.value }}
+                    </h3>
+                    <p>
+                      <img src="{{ static('img/3rdparty/silk/exclamation.png') }}" alt="protected" title="Protected Data" />
+                      Note: Unredacted exception messages are protected data.
+                      <a href="{{ url('documentation:protected_data_access') }}">Protected data policy.</a>
+                    </p>
```
willkg, line 708: This was tricky. The only part of the stacktrace that's protected data is the message. I hope the way I've got it structured and the messaging I'm using gets that across.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -815,6 +817,48 @@ <h2>Thread {{ thread.thread }}{% if thread.thread_name %}, Name: {{ thread.threa
               </div>
               <!-- /frames -->
             {% endif %}
+
+            {% if (request.user.has_perm('crashstats.view_pii') or your_crash) and report.breadcrumbs %}
```
willkg, line 821: This is the same as it was before--just moved to this position in the document.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -129,6 +130,16 @@ def report_index(request, crash_id, default_context=None):
 
     context["parsed_dump"] = parsed_dump
 
+    if "java_exception" in context["report"]:
```
willkg, line 133: Oops--I meant to add a comment here.

socorro/lib/javautil.py
willkg: If we ever discover some other part of the JavaException has PII, we'd need to fix the processor code here and then reprocess all those crashes.

On the one hand, that's probably a lot of crashes.

On the other hand, processing Java crashes is really fast since we don't have to run the stackwalker which has to download and parse symbols in order to walk the stack.

I should add a note here about that.

socorro/processor/rules/mozilla.py
```diff
@@ -268,24 +268,49 @@ class JavaProcessRule(Rule):
     """Move Java-crash-related bits over."""
 
     def predicate(self, raw_crash, raw_dumps, processed_crash, proc_meta):
-        return bool(raw_crash.get("JavaStackTrace", None))
+        return bool(raw_crash.get("JavaStackTrace", None)) or bool(
+            raw_crash.get("JavaException", None)
+        )
```
willkg, line 273: Part of me wonders whether we should move all the Java crash related rules to a new file. I'm deferring doing that until the next Java-specific change we make.

socorro/processor/rules/mozilla.py
```diff
@@ -627,15 +652,15 @@ def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
...
-    have more than one in a list.  However, in all the years that this existed
-    it was never expanded to use more than just one.  Meanwhile, the code
+    have more than one in a list. However, in all the years that this existed
+    it was never expanded to use more than just one. Meanwhile, the code
     ambiguously would sometimes give this as as single value and other times
     return it as a list of one item.
 
     This rule does not try to reproduce that imbiguity and avoids the list
-    entirely, just giving one single value.  The fact that the destination
+    entirely, just giving one single value. The fact that the destination
```
willkg, line 663: I'm not sure what made these changes. I don't think I did it by hand, so maybe it was something automated?

socorro/schemas/java_exception.json
```diff
@@ -0,0 +1,75 @@
...
+          "type": "array",
+          "description": "Array of exception stacktraces",
+          "items": {
+            "type": "object",
+            "required": ["stacktrace"],
+            "properties": {
+              "stacktrace": {
+                "type": "object",
+                "description": "A single stacktrace",
+                "required": ["frames", "type", "module"],
```
willkg, line 23: NullPointerException is an example of an exception that has no value. So we don't want to require "value" here.

socorro/schemas/java_exception.json
```diff
@@ -0,0 +1,75 @@
...
+                "properties": {
+                  "frames": {
+                    "type": "array",
+                    "description": "Array of frames",
+                    "items": {
+                      "type": "object",
+                      "description": "A single frame",
+                      "required": ["module", "function", "in_app", "lineno"],
```
willkg, line 31: I looked at 10,000 JavaException values. Some of them had frames that don't have a filename. I don't know offhand what causes that in Java land. Anyhow, "filename" shouldn't be required here.

socorro/schemas/java_exception.json
```diff
@@ -0,0 +1,75 @@
...
+                      "properties": {
+                        "module": {
+                          "type": "string",
+                          "description": "Module for the frame"
+                        },
+                        "function": {
+                          "type": "string",
+                          "description": "Function in the module"
+                        },
+                        "in_app": {
```
willkg, line 41: I have no idea what in_app means. I'll ping Roger about it and add a description here.

PR #5661, opened 2021-01-27
bug 1669983: move scripts to bin/

willkg:
This moves scripts that were in `scripts/` and `docker/` to `bin/`. This makes Socorro more like other projects I'm maintaining and reduces some of the mental overhead for maintenance.

While doing this, I also:

* improved usage documentation
* changed the `set` line in bash scripts to make them safer
* removed scripts we didn't need anymore

One thing is that this copied `scripts/run_migrations.sh` to `bin/`. The deploy code uses this script, so once this lands, I'll go fix the deploy code and then do another PR here to remove `scripts/run_migrations.sh`.


bin/run_webapp.sh
```diff
@@ -26,11 +29,6 @@ if [ "$1" == "--dev" ]; then
     cd /app/webapp-django/ && ${CMDPREFIX} python manage.py runserver 0.0.0.0:8000
 
 else
-    echo "GUNICORN_WORKERS=${GUNICORN_WORKERS}"
-    echo "GUNICORN_WORKER_CLASS=${GUNICORN_WORKER_CLASS}"
-    echo "GUNICORN_MAX_REQUESTS=${GUNICORN_MAX_REQUESTS}"
-    echo "GUNICORN_MAX_REQUESTS_JITTER=${GUNICORN_MAX_REQUESTS_JITTER}"
-    echo "PORT=${PORT}"
```
willkg, line 33: I added `set -x` so it spits out the variables as they're set and we don't need to additionally echo them. That reduces a lot of error-prone code.

docker/run_mdswshell.sh
```diff
@@ -1,58 +0,0 @@
-#!/bin/bash
```
willkg, line 1: I moved the minidump-stackwalk code to a separate repo and should have removed this file.

docker/run_build_docs.sh
```diff
@@ -1,11 +0,0 @@
...
-
-# This Source Code Form is subject to the terms of the Mozilla Public
-# License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/.
-
-# Change directory to docs directory to use that Makefile and cwd
-cd docs
-
-# Build the docs
-make html
```
willkg, line 11: This file does nothing interesting, so I nixed it.

PR #5662, opened 2021-01-27
bug 1669983: fix unbound variable in run_web.sh

willkg:


bin/run_webapp.sh
```diff
@@ -21,7 +21,7 @@ GUNICORN_MAX_REQUESTS_JITTER=${GUNICORN_MAX_REQUESTS_JITTER:-"1000"}
 CMDPREFIX="${CMDPREFIX:-}"
 
 
-if [ "$1" == "--dev" ]; then
+if [ "${1:-}" == "--dev" ]; then
```
willkg, line 24: When in production, there's no arg passed, so this was an unbound variable and raised an error and the webapp would die. Oops.

PR #5676, opened 2021-02-08
bug 1691393: drop defaults for http requests

willkg:
This drops the `backoff_factor` to 0.1 and the default timeout to 3. This makes it less likely that trying to request a resource from a
non-responsive service (*cough*Bugzilla*cough*) doesn't end up exceeding Gunicorn's timeout value at which point Gunicorn kills the process.

I also dropped the waitlist generator from `[2, 2, 2, 2, 2]` to `[1, 1,1, 1, 1]` which might alleviate a similar issue in the crashstorage code where it's trying to push something to AWS S3 and has to retry a few times.

### DIFF
--- a/socorro/lib/requestslib.py
+++ b/socorro/lib/requestslib.py
@@ -35,18 +35,27 @@ class HTTPAdapterWithTimeout(HTTPAdapter):
 
 def session_with_retries(
     total_retries=5,
-    backoff_factor=0.2,
+    backoff_factor=0.1,
     status_forcelist=(429, 500),
-    default_timeout=5.0,
+    default_timeout=3.0,
 ):
     """Returns session that retries on HTTP 429 and 500 with default timeout
 
     :arg int total_retries: total number of times to retry
 
-    :arg float backoff_factor: number of seconds to increment by between
-        attempts
+    :arg float backoff_factor: number of seconds to apply between attempts
 
-        For example, 0.1 will back off 0.1s, then 0.2s, then 0.3s, ...
+        The sleep amount is calculated like this::
+
+            sleep = backoff_factor * (2 ** (num_retries - 1))
+
+        For example, backoff_factor 0.1 will back off :
+
+        * 0.1
+        * 0.2
+        * 0.4
+        * 0.8
+        * 1.6 ...
 
     :arg tuple of HTTP codes status_forcelist: tuple of HTTP codes to
         retry on
--- a/socorro/lib/util.py
+++ b/socorro/lib/util.py
@@ -41,7 +41,7 @@ class MaxAttemptsError(Exception):
 
 def wait_time_generator():
     """Return generator for wait times."""
-    yield from [2, 2, 2, 2, 2]
+    yield from [1] * 5
 
 
 def retry(
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -948,9 +948,6 @@ class BugzillaBugInfo(SocorroCommon):
     # This is for how long we cache the metadata of each individual bug.
     BUG_CACHE_SECONDS = 60 * 60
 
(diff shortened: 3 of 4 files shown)


bin/run_webapp.sh
```diff
@@ -18,6 +18,7 @@ GUNICORN_WORKERS=${GUNICORN_WORKERS:-"1"}
 GUNICORN_WORKER_CLASS=${GUNICORN_WORKER_CLASS:-"sync"}
 GUNICORN_MAX_REQUESTS=${GUNICORN_MAX_REQUESTS:-"10000"}
 GUNICORN_MAX_REQUESTS_JITTER=${GUNICORN_MAX_REQUESTS_JITTER:-"1000"}
+GUNICORN_TIMEOUT=${GUNICORN_TIMEOUT:-"60"}
```
willkg, line 21: Oops--I didn't mean to do this. Will remove it.

socorro/lib/requestslib.py
```diff
@@ -35,18 +35,27 @@ def send(self, *args, **kwargs):
 
 def session_with_retries(
     total_retries=5,
-    backoff_factor=0.2,
+    backoff_factor=0.1,
```
willkg, line 38: Earlier, I thought the backoff was linear so it'd go 0.2, 0.4, 0.6, 0.8, 1.0. That's wrong--it's exponential. So with a 0.2 backoff factor, it goes 0.2, 0.4, 0.8, 1.6, 3.2.

This changes it to 0.1 which goes 0.1, 0.2, 0.4, 0.8, 1.6.

socorro/lib/requestslib.py
```diff
@@ -35,18 +35,27 @@ def send(self, *args, **kwargs):
...
     status_forcelist=(429, 500),
-    default_timeout=5.0,
+    default_timeout=3.0,
```
willkg, line 40: After reading the urllib3 docs, I decided a default timeout of 5 is too high and we should use 3. In the case where the service is unresponsive, this kicks in for every retry. So if you have a max_retries of 5 and the default connection timeout is 5, that's 25 seconds sitting around waiting for something that's probably not going to come up in the 30 seconds gunicorn gives us before it gives up and kills the process.

Changing it to 3 reduces that to 15 seconds.

socorro/lib/requestslib.py
```diff
@@ -35,18 +35,27 @@ def send(self, *args, **kwargs):
...
+
+            sleep = backoff_factor * (2 ** (num_retries - 1))
+
+        For example, backoff_factor 0.1 will back off :
+
+        * 0.1
+        * 0.2
+        * 0.4
+        * 0.8
+        * 1.6 ...
```
willkg, line 58: This improves the docs so it's clearer it's not a linear progression.

socorro/lib/util.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, msg, ret=None):
 
 def wait_time_generator():
     """Return generator for wait times."""
-    yield from [2, 2, 2, 2, 2]
+    yield from [1] * 5
```
willkg, line 44: Changed this from 2, 2, 2, 2, 2 to 1, 1, 1, 1, 1.

There was another case in sentry where the crashstorage code was retrying pushing something to AWS S3. We don't need to wait 2 seconds between iterations.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -948,9 +948,6 @@ class BugzillaBugInfo(SocorroCommon):
     # This is for how long we cache the metadata of each individual bug.
     BUG_CACHE_SECONDS = 60 * 60
 
-    # How patient we are with the Bugzilla REST API
-    BUGZILLA_REST_TIMEOUT = 5  # seconds
```
willkg, line 952: This should use the defaults, so I removed this which just complicated things and wasn't used right.

PR #5679, opened 2021-02-09
bug 1688203: improve crash report view summary in Matrix

willkg:
This is a first pass at adding summary information for the page so when
systems like Matrix/Slack show a summary of the url of a crash report
view page, it has something more helpful to show.


willkg:
I need to test this in stage, so I'm going to land it and iterate from there in future PRs.

PR #5682, opened 2021-02-13
bug 1692504: add What's New docs section

willkg:
This adds a What's New docs section that's built from a `WHATSNEW.rst`
file in the repo root.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ socorro.egg-info/
 dist/
 webapp-django/.env
 .pytest_cache
+.python-version
 
 # docker things
 .docker-build*
--- a/WHATSNEW.rst
+++ b/WHATSNEW.rst
@@ -1,0 +1,18 @@
+=========================
+What's new in Crash Stats
+=========================
+
+Important changes to the Crash Stats service.
+
+
+No longer collecting Email data (2021-02-11)
+============================================
+
+*February 11th, 2021*
+
+As of February 11th, 2021, Socorro is no longer collecting Email data. Email
+data in crash reports collected on and after February 11th, 2021 will have the
+Email data stripped from the crash report before being saved and processed.
+
+One consequence of this is that new crash reports will no longer show up in
+Your Profile.
--- a/webapp-django/crashstats/crashstats/static/crashstats/css/base/typography.less
+++ b/webapp-django/crashstats/crashstats/static/crashstats/css/base/typography.less
@@ -43,7 +43,7 @@ em {
     font-style: italic;
 }
 
-b {
+strong, b {
     font-weight: bold;
 }
 
--- a/webapp-django/crashstats/crashstats/static/crashstats/css/components/panel.less
+++ b/webapp-django/crashstats/crashstats/static/crashstats/css/components/panel.less
@@ -9,7 +9,7 @@
         .shadow-none();
     }
 
-    .title {
+    div.title {
         background: @tan repeat-x left top;
         display: flex;
         font-size: 12px;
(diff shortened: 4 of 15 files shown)


.gitignore
```diff
@@ -17,6 +17,7 @@ socorro.egg-info/
 dist/
 webapp-django/.env
 .pytest_cache
+.python-version
```
willkg, line 20: This isn't really related to this PR, but it keeps irritating me.

WHATSNEW.rst
```diff
@@ -0,0 +1,18 @@
...
+============================================
+
+*February 11th, 2021*
+
+As of February 11th, 2021, Socorro is no longer collecting Email data. Email
+data in crash reports collected on and after February 11th, 2021 will have the
+Email data stripped from the crash report before being saved and processed.
+
+One consequence of this is that new crash reports will no longer show up in
+Your Profile.
```
willkg, line 18: This allows us to add service-related things to this file and they'll show up on the What's New page as well as in the docs on readthedocs.

This seems like a good-enough system. I'm not entirely sure what kinds of things should go in this file, but for now I'm saying it's "service changes". For example, dropping the Email field is a big service change. Other possible examples:

* adding/removing an API
* adding/removing a report
* switching to a new authentication system

We've done things like that in the past. It might be interesting to backfill the news.

webapp-django/crashstats/crashstats/static/crashstats/css/base/typography.less
```diff
@@ -43,7 +43,7 @@ em {
     font-style: italic;
 }
 
-b {
+strong, b {
```
willkg, line 46: reST uses `strong`, but that wasn't showing up as bold. I didn't spend a lot of time looking into why. This fix seemed good enough for now.

webapp-django/crashstats/crashstats/static/crashstats/css/components/panel.less
```diff
@@ -9,7 +9,7 @@
         .shadow-none();
     }
 
-    .title {
+    div.title {
```
willkg, line 12: This applied to anything with class title. That doesn't work for the reST pages because the first h1 has class title and then shows up poorly. I figured it was easier to reduce the scope here.

webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -79,7 +79,7 @@ def timestamp_to_date(timestamp, fmt="%Y-%m-%d %H:%M:%S"):
 
 
 @library.filter
-def time_tag(dt, format="%a, %b %d %H:%M %Z", future=False):
+def time_tag(dt, format="%a, %b %d, %Y at %H:%M %Z", future=False):
```
willkg, line 82: This isn't really related to this PR, but I bumped into it when fiddling with things. Not having the year is really irritating for me. Socorro shows things over a 6 month period which can span years.

webapp-django/crashstats/documentation/views.py
```diff
@@ -33,6 +38,29 @@ def home(request, default_context=None):
...
+    :returns: HTML document as string
+
+    """
+    path = Path(settings.SOCORRO_ROOT) / "WHATSNEW.rst"
+
+    with open(path, "r") as fp:
+        data = fp.read()
+        parts = docutils.core.publish_parts(data, writer_name="html")
+
+    return parts["html_body"]
```
willkg, line 54: I thought about doing this with the protected data access policy, too. Then it can also exist in three places (and be easier to maintain). I might do that another day.

PR #5688, opened 2021-02-18
bug 1262424: don't create/save django session for API endpoints

willkg:
This stops creating/saving django sessions for API endpoint requests.
This also nixes the anoncsrf cookie since we don't need that with the
API.

To test:

1. run the webapp
2. log in and create an API token
3. use curl to hit the SuperSearch endpoint a few times:
   ```
   curl -v -H "Auth-Token: TOKEN" http://localhost:8000/api/SuperSearch/
   ```
4. verify that no cookies are set in the response
5. verify that ```select count(*) from django_session``` doesn't change

### DIFF
--- a/webapp-django/crashstats/api/views.py
+++ b/webapp-django/crashstats/api/views.py
@@ -7,6 +7,9 @@ import inspect
 import json
 import re
 
+from ratelimit.decorators import ratelimit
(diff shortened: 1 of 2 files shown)


willkg:
On second thought, this is junk. I'm going to rethink this.

PR #5689, opened 2021-02-18
bug 1262424: don't create/save django session for API endpoints

willkg:
This stops creating/saving django sessions for API endpoint requests. This also nixes the anoncsrf cookie since we don't need that with the API.

To test:

1. run the webapp
2. log in and create an API token
3. use curl to hit the SuperSearch endpoint a few times:
   ```
   curl -v -H "Auth-Token: TOKEN" http://localhost:8000/api/SuperSearch/
   ```
4. verify that no cookies are set in the response
5. verify that ```select count(*) from django_session``` doesn't change

### DIFF
--- a/webapp-django/crashstats/api/views.py
+++ b/webapp-django/crashstats/api/views.py
@@ -3,9 +3,13 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import datetime
+from functools import wraps
 import inspect
 import json
 import re
+
+from ratelimit.decorators import ratelimit
+from session_csrf import anonymous_csrf_exempt
 
 from django import http
 from django import forms
@@ -16,8 +20,6 @@ from django.core.validators import ProhibitNullCharactersValidator
 from django.shortcuts import render
 from django.urls import reverse
 from django.views.decorators.csrf import csrf_exempt
-
-from ratelimit.decorators import ratelimit
 
 from crashstats.api.cleaner import Cleaner
 from crashstats.crashstats import models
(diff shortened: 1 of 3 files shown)


willkg:
This differs from the previous PR. This one implements the logic in the API module rather than in the tokens middleware which was an awkward place for it.

willkg:
Because I moved the logic to the API module, this should only affect the API. I'll test it out with tools when this lands in stage.

webapp-django/crashstats/api/views.py
```diff
@@ -114,7 +116,53 @@ def is_valid_model_class(model):
...
+    of the existence of the anonymous_csrf_exempt decorator.
+
+    """
+
+    @wraps(fun)
+    def _no_csrf(request, *args, **kwargs):
+        ret = fun(request, *args, **kwargs)
+        # Remove any csrf bits from Django or django-session-csrf so they don't persist
+        if hasattr(request, "_anon_csrf_key"):
+            del request._anon_csrf_key
```
willkg, line 156: What's going on here is that we have `@anonymous_csrf_exempt`, but that only exempts CHECKING the csrf--not doing the bookkeeping and adding a cookie. This `_anon_csrf_key` gets set by `process_request` in the middleware before the view code runs. So we remove it after the view runs and then the middleware doesn't add the cookie in `process_response`.

webapp-django/crashstats/api/views.py
```diff
@@ -114,7 +116,53 @@ def is_valid_model_class(model):
...
+        ret = fun(request, *args, **kwargs)
+
+        # Clear the session if nothing was saved to it
+        session_keys = [
+            key for key in request.session.keys() if not key.startswith("_auth_user")
+        ]
+        if not session_keys:
+            request.session.clear()
+            request.session.flush()
+        return ret
```
willkg, line 134: Because the token middleware logs the "user" in using the token, that creates a session. This code checks to see if the view added anything to the session and if not, clears it. That way, the session never makes it to the db and never goes in the cookie.

webapp-django/crashstats/settings/base.py
```diff
@@ -106,6 +106,7 @@ def path(*dirs):
     "django.middleware.common.CommonMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
+    "crashstats.tokens.middleware.APIAuthenticationMiddleware",
```
willkg, line 109: As near as I can tell, we should have the authentication middlewares happen before things that look at whether the user is authenticated. If that's not true for Socorro, it's not documented anywhere. If this causes problems, I'll add a comment when I fix it.

PR #5690, opened 2021-02-18
bug 1262424: fix another session issue and a bug

willkg:
This fixes a bug where if you were using API endpoints from your browser
(like from the API docs), it would whack your loging.

This fixes another bug wehre it was still creating session cookies and
not properly removing them.

### DIFF
--- a/webapp-django/crashstats/api/views.py
+++ b/webapp-django/crashstats/api/views.py
@@ -124,11 +124,12 @@ def clear_empty_session(fun):
         # Run the function
         ret = fun(request, *args, **kwargs)
 
-        # Clear the session if nothing was saved to it
+        # Clear the session if nothing was saved to it and it was generated using a
+        # token login
         session_keys = [
             key for key in request.session.keys() if not key.startswith("_auth_user")
         ]
-        if not session_keys:
+        if not session_keys and getattr(request.user, "token_login", False):
             request.session.clear()
             request.session.flush()
         return ret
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/tokens/middleware.py
```diff
@@ -65,6 +65,7 @@ def process_request(self, request):
         # we use as long as it's something
         user.backend = "django.contrib.auth.backends.ModelBackend"
         user.has_perm = partial(has_perm, token.permissions.all())
+        user.token_login = True
```
willkg, line 68: This allows us to differentiate between whether this "user" logged in via normal login or via an auth token.

webapp-django/crashstats/api/views.py
```diff
@@ -151,18 +152,22 @@ def no_csrf_i_mean_it(fun):
...
         return ret
 
     return _no_csrf
 
 
 @anonymous_csrf_exempt
 @csrf_exempt
-@no_csrf_i_mean_it
 @clear_empty_session
+@no_csrf_i_mean_it
```
willkg, line 170: We need `no_csrf_i_mean_it` to run and remove the `csrf_token` that django adds to the session. That way the session will get cleared.

webapp-django/crashstats/api/views.py
```diff
@@ -124,11 +124,12 @@ def _clear_empty_session(request, *args, **kwargs):
...
         ret = fun(request, *args, **kwargs)
 
-        # Clear the session if nothing was saved to it
+        # Clear the session if nothing was saved to it and it was generated using a
+        # token login
         session_keys = [
             key for key in request.session.keys() if not key.startswith("_auth_user")
         ]
-        if not session_keys:
+        if not session_keys and getattr(request.user, "token_login", False):
```
willkg, line 132: We only want to wipe the session if we know it was created because the "user" logged in using an auth token.

willkg:
I've almost hit the point where this is complicated enough that it's not worth doing. If this doesn't fix it, I think I'll back everything out.

PR #5704, opened 2021-03-01
bug 1166058: implement update_graphics_pci

willkg:
This removes the graphics device admin page that lets us upload and
parse the `pci.ids` file That requires the max body upload size to be
higher than the file size and it's not. Rather than fiddle with that, I
redid everything to run as a cron job.

In doing that, I discovered we were parsing the file wrong, so I fixed
that, too.

I updated the graphicsdevice data flow docs.


webapp-django/crashstats/crashstats/management/commands/update_graphics_pci.py
```diff
@@ -0,0 +1,73 @@
...
+            resp.text.splitlines(), debug=debug_mode
+        )
+
+        total_created = 0
+        total_updated = 0
+        total_skipped = 0
+
+        for item in devices:
+            # print(item)
+            # input("foo")
```
willkg, line 48: I should take these out.

webapp-django/crashstats/crashstats/tests/test_utils.py
```diff
@@ -3,6 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import copy
+import os
 import json
```
willkg, line 7: What has two thumbs and doesn't know the alphabet?

:+1: :+1: this guy!

webapp-django/crashstats/crashstats/utils.py
```diff
@@ -633,3 +634,88 @@ def ratelimit_rate(group, request):
...
+
+def string_hex_to_hex_string(snippet):
+    """The PCIDatabase.com uses shortened hex strings (e.g. '919A')
+    whereas in Socorro we use the full represenation, but still as a
+    string (e.g. '0x919a').
+    Also, note that when converting the snippet to a 16 base int, we
+    can potentially lose the leading zeros, but we want to make sure
+    we always return a 4 character string preceeded by 0x.
+    This function tries to make that conversion.
+    """
```
willkg, line 647: I should fix this docstring while I'm moving this function.

PR #5705, opened 2021-03-09
bug 1696910: fix data links and redo "raw data and minidumps tab"

willkg:
This redoes the "Raw data and minidumps" tab so that the links are at
the top and a syntax-highlighted and more usable version of the
minidump-stackwalk output is below it.

This fixes the data links so they're generated in the view code using
API endpoints rather than an odd crashstats view. This fixes the
RawCrash model to have a class method for figuring out the binary
filename rather than a static template.

This redoes how we do list-style CSS by adding a couple of
tailwind-inspired classes and putting them in the typography base file.

### DIFF
--- a/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
+++ b/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
@@ -1002,9 +1002,36 @@
           <!-- /modules -->
 
           <div id="rawdump" class="ui-tabs-hide">
-            <div class="code">{{ raw_stackwalker_output }}</div>
+            <h3>Download raw and processed crash data</h3>
+            {% if request.user.has_perm('crashstats.view_pii') %}
+              <p>
+                {{ protected_warning(your_crash) }}
+                {% if not your_crash %}
+                  Raw and processed crash reports contain protected data.
+                {% endif %}
+              </p>
+              <ul class="list-inside list-disc">
+                {% for name, url in raw_data_urls %}
+                  <li>{{ name }}: <a href="{{ url }}">{{ url }}</a></li>
+                {% endfor %}
+              </ul>
+            {% elif request.user and request.user.is_active %}
+              <p>
+                You do not have access to protected data.
+                You need to be logged in and have access to protected data to see links to crash report data.
+                See <a href="{{ url('documentation:protected_data_access') }}">protected data access documentation</a>
+                for more information.
+              </p>
+            {% else %}
+              <p>
+                You are not logged in.
+                You need to be logged in and have access to protected data to see links to crash report data.
+                See <a href="{{ url('documentation:protected_data_access') }}">protected data access documentation</a>
+                for more information.
+              </p>
+            {% endif %}
 
-            <h3>Download Raw Crash Data and Minidumps</h3>
+            <h3>Download dumps and memory report</h3>
             {% if request.user.has_perm('crashstats.view_rawdump') %}
               <p>
                 {{ protected_warning(your_crash) }}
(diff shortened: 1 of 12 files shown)


webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -1013,28 +1011,42 @@ <h3>Download Raw Crash Data and Minidumps</h3>
                   See <a href="{{ url('documentation:protected_data_access') }}">Protected data policy</a> for details.
                 {% endif %}
               </p>
-              <ul>
-                {% for name, url in raw_dump_urls %}
+              <ul class="list-inside list-disc">
+                {% for name, url in raw_data_urls %}
```
willkg, line 1015: I don't know what "raw" means anymore.

I moved generating the urls into the view code, so all this does is spit out what it knows about.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -1013,28 +1011,42 @@ <h3>Download Raw Crash Data and Minidumps</h3>
...
+                {% if not your_crash %}
+                  minidump-stackwalk output contains sensitive and exploitability information.
+                {% endif %}
+              {% else %}
+                You do not have access to protected data so the minidump-stackwalk output is redacted.
+                See <a href="{{ url('documentation:protected_data_access') }}">protected data access documentation</a>
+                for more information.
+              {% endif %}
+            </div>
+            <div id="minidump-stackwalk-json" data-minidumpstackwalk="{{ raw_stackwalker_output }}"></div>
```
willkg, line 1048: This displays the minidump-stackwalk output the same way we do the Telemetry environment output. It's got some nice features that make it easier to see a HUGE PILE O' JSON.

webapp-django/crashstats/crashstats/jinja2/crashstats/signup.html
```diff
@@ -10,34 +10,32 @@ <h2>Signing up for an account on Crash Stats</h2>
 
     <div class="panel">
       <div class="body">
-        <div class="prose">
```
willkg, line 13: I got rid of "prose" as a class and switched to using tailwind-inspired classes.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -418,12 +418,14 @@ class SocorroMiddleware(SocorroCommon):
...
     API_BINARY_RESPONSE = {}
-    # and thus no needed binary filename
-    API_BINARY_FILENAME = None
 
     # By default no special permissions are needed for binary response
     API_BINARY_PERMISSIONS = ()
 
+    @classmethod
+    def get_binary_filename(cls, params):
+        return None
```
willkg, line 427: I made this a class method. It's only used by RawCrash and it's used because the various dumpy things all have different names. Making it a class method allows us to return a differently shaped name depending on what's being requested.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -188,14 +188,47 @@ def report_index(request, crash_id, default_context=None):
     context["public_raw_keys"].sort(key=lambda s: s.lower())
     context["protected_raw_keys"].sort(key=lambda s: s.lower())
 
+    data_urls = []
```
willkg, line 191: I switched link creation to something that's more predictable and uses the existing API.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -188,14 +188,47 @@ def report_index(request, crash_id, default_context=None):
...
+                    crash_id,
+                ),
+            )
+        )
+
     if request.user.has_perm("crashstats.view_rawdump"):
-        context["raw_dump_urls"] = [
-            ("dump", reverse("crashstats:raw_data", args=(crash_id, "dmp"))),
+        # Add link for each minidump and memory report
+        # FIXME(willkg): does this show even if the crash report thas no minidump?
```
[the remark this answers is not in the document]
willkg: It was so busted before--it was printing a link regardless of whether the crash report actually had a minidump or not. So I fixed that.

PR #5714, opened 2021-03-17
bug 1699122: make cache key in __heartbeat__ unique

willkg:
This changes the cache_key that we use to test cache set/get with such
that it's unique. I think this will prevent race conditions from two
webapp things doing the cache set/get test at the same time from bumping
into one another creating a spurious error.

### DIFF
--- a/webapp-django/crashstats/monitoring/views.py
+++ b/webapp-django/crashstats/monitoring/views.py
@@ -4,6 +4,7 @@
 
 import datetime
 import json
+import uuid
 
 import elasticsearch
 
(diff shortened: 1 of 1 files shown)


webapp-django/crashstats/monitoring/views.py
```diff
@@ -81,9 +82,9 @@ def dockerflow_heartbeat(request):
     assert Permission.objects.all().count() > 0
 
     # We should also be able to set and get a cache value
-    cache_key = "__healthcheck__"
+    cache_key = "__healthcheck__" + str(uuid.uuid4())
```
willkg, line 85: This creates a string like `__healthcheck__6a9b3d64-d68a-40d6-8801-8a6b4db27a9c`.

webapp-django/crashstats/monitoring/views.py
```diff
@@ -81,9 +82,9 @@ def dockerflow_heartbeat(request):
...
     cache.set(cache_key, 1, 10)
-    assert cache.get(cache_key)
+    assert cache.get(cache_key), f"cache_key {cache_key} not available"
```
willkg, line 87: This adds a message when the assertion fails that's easier to glance at in Sentry and understand what's going on.

PR #5715, opened 2021-03-18
bug 1699239: end Fennec support

willkg:
This removes all Fennec/FennecAndroid things. We no longer support
collecting and processing Fennec crash report data.


socorro/processor/rules/mozilla.py
```diff
@@ -466,30 +466,6 @@ def action(self, raw_crash, raw_dumps, processed_crash, processor_meta):
                 processed_crash["memory_report"] = memory_report
 
 
-class ProductRewrite(Rule):
```
willkg, line 469: Who has two thumbs and is not going to miss this?

:+1: :+1: this guy.

PR #5739, opened 2021-04-08
bug 1701357: fix topcrasher report process type filter

willkg:
After the process_type changes, the TopCrasher report had a "parent"
filter and a "browser" filter. If you were looking at TopCrashers by
process type, that split the parent process crash reports into two
separate lists. That would eventually fix itself as data expired, but
it's incredibly inconvenient right now.

This changes the TopCrashers report filters by changing Supersearch to
treat "process_type=parent" and "process_type=browser" filters as "crash
reports where process_type is parent OR has no ProcessType".

That makes TopCrashers report "parent" type filter work correctly
regardless of when the crash report was processed.

### DIFF
--- a/socorro/lib/search_common.py
+++ b/socorro/lib/search_common.py
@@ -6,7 +6,9 @@
 Common functions for search-related external modules.
 """
 
+from dataclasses import dataclass
 import datetime
+from typing import Any, Optional
 
 from socorro.lib import BadArgumentError, datetimeutil
 import socorro.lib.external_common as extern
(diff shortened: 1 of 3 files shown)


socorro/lib/search_common.py
```diff
@@ -59,20 +61,20 @@
...
-        self.name = name
-        self.value = value
-        self.operator = operator
-        self.data_type = data_type
-        self.operator_not = operator_not
+    name: str
+    value: str
+    operator: Optional[str] = None
+    data_type: Optional[str] = None
+    operator_not: bool = False
```
willkg, line 70: Switching this to dataclasses makes this code much easier to debug because it includes `__repr__` and `__str__` methods.

socorro/lib/search_common.py
```diff
@@ -292,24 +294,37 @@ def fix_date_parameter(self, parameters):
     def fix_process_type_parameter(parameters):
         """Correct the process_type parameter.
 
-        If process_type is 'browser', replace it with a 'does not exist'
-        parameter. Do nothing in all other cases.
+        Previously, crash reports that didn't have a ProcessType annotation
```
willkg, line 297: We can remove this staticmethod when we remove all the "browser" shenanigans, but for now it helps us bridge the gap between the old way and the new way.

webapp-django/crashstats/settings/base.py
```diff
@@ -291,7 +291,6 @@ def filter(self, record):
 PROCESS_TYPES = (
     "any",
     "parent",
-    "browser",
```
willkg, line 294: Taking this out here means that it won't show up as a filter option in the TopCrashers report. That should reduce some confusion for people using that report.

I'm pretty sure that's the only place this list shows up, but Socorro is an antediluvian place of malevolence and horror.

PR #5742, opened 2021-04-12
bug 1704529: clean up field value truncation code

willkg:
This fixes the case where `async_shutdown_timeout` values weren't getting truncated.

Before indexing data into Elasticsearch, the crashstorage code truncates the values for "string" and "keyword" fields. The way it figured out which fields were which was messy, missed a bunch of fields, and didn't cache intermediary work which would always produce the same results.

This refactors that code up so it's a little better and fixes the immediate issue in bug #1704529. The `get_fields_by_item()` function isn't great, but it's ok for now.

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -15,7 +15,11 @@ from elasticsearch.exceptions import NotFoundError
 import markus
 
 from socorro.external.crashstorage_base import CrashStorageBase, Redactor
-from socorro.external.es.super_search_fields import FIELDS, parse_mapping
+from socorro.external.es.super_search_fields import (
+    FIELDS,
+    parse_mapping,
+    get_fields_by_item,
+)
 from socorro.lib.datetimeutil import JsonDTEncoder, string_to_datetime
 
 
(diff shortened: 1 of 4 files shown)


willkg:
Because this fixes a case where not truncating causes Elasticsearch to return an error and we're only truncating to the largest possible size, I don't expect this to change how much we're indexing.

Also, this adds a "crash_document_size" metric so we can toss that in grafana and see total sizes of things being indexed.

socorro/external/es/crashstorage.py
```diff
@@ -118,68 +88,78 @@ def is_valid_key(key):
...
+            data[field_name] = value[:max_size]
 
 
-def truncate_string_field_values(fields, data):
-    """Truncates string field values greater than MAX_STRING_FIELD_VALUE_SIZE length
+def truncate_string_field_values(data, fields, max_size):
+    """Truncates string field values greater than max_size length in bytes
 
-    Note: This modifies the data dict in-place and only looks at the top level.
+    This truncates unicode string values to less than 32,766 bytes.
```
willkg, line 116: This should say "max_size" and not the number.

socorro/external/es/crashstorage.py
```diff
@@ -118,68 +88,78 @@ def is_valid_key(key):
...
 
-        if value != new_value:
-            data[field_name] = new_value
+        # Remove bytes until we either run out of bytes or we can decode to a unicode
+        # string
+        while value_bytes:
+            try:
+                new_value = value_bytes.decode("utf-8")
+                break
+            except UnicodeDecodeError:
```
willkg, line 157: I can rewrite this to use a function and it'd be clearer.

PR #5744, opened 2021-04-14
bug 1704872: fix operands for percent_of_total_crashes_diff calculation

willkg:
The percent of total crashes diff between previous and current should go
"up" when moving towards 100 and "down" when moving towards 0. Given
that, We want current - previous to figure out the difference.

This also adds "helpful" up/down arrows to make the TopCrashers lines
easier to skim without having to read the hover text.

### DIFF
--- a/webapp-django/crashstats/topcrashers/jinja2/topcrashers/topcrashers.html
+++ b/webapp-django/crashstats/topcrashers/jinja2/topcrashers/topcrashers.html
@@ -146,6 +146,7 @@
                     <td title="This is a new signature">new</td>
                   {% else %}
                     <td title="A change of {{ topcrashers_stats_item.percent_of_total_crashes_diff | round(2) }}% from {{ topcrashers_stats_item.previous_percent_of_total_crashes | round(2) }}%">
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/topcrashers/jinja2/topcrashers/topcrashers.html
```diff
@@ -145,7 +145,8 @@ <h2>
                   {% if topcrashers_stats_item.percent_of_total_crashes_diff == 'new' %}
                     <td title="This is a new signature">new</td>
                   {% else %}
-                    <td title="A change of {{ topcrashers_stats_item.percent_of_total_crashes_diff | round(2) }}% from {{ topcrashers_stats_item.previous_percent_of_total_crashes | round(2) }}%">
+                  <td title="A change of {{ topcrashers_stats_item.percent_of_total_crashes_diff | round(2) }}% from {{ topcrashers_stats_item.previous_percent_of_total_crashes | round(2) }}%">
```
willkg, line 148: Oops--this got dedented.

PR #5747, opened 2021-04-15
bug 1699492: fix mutation issues in signature generation

willkg:
Signature generation used to mutate the stack frames adding a
"normalized" key to them. Not only do we not use that, but there's
another bit of code elsewhere that used to remove it because it was
unhelpful.

This gets rid of the normalized caching altogether. It adds a test to
make sure we're not mutating in the future. It reduces some code around
generating signatures from a stack which suggested that these tools had
the same API but was a lie.

### DIFF
--- a/socorro/signature/rules.py
+++ b/socorro/signature/rules.py
@@ -58,33 +58,12 @@ class Rule:
         return True
 
 
-class SignatureTool:
-    """Stack walking signature generator base class
-
-    This defines the interface for classes that take a stack and generate a
-    signature from it.
-
-    Subclasses should implement the ``_do_generate`` method.
-
-    """
-
-    def generate(self, source_list, hang_type=0, crashed_thread=None, delimiter=" | "):
-        signature, notes, debug_notes = self._do_generate(
-            source_list, hang_type, crashed_thread, delimiter
-        )
-
-        return signature, notes, debug_notes
-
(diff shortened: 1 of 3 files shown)


socorro/signature/rules.py
```diff
@@ -58,33 +58,12 @@ def action(self, crash_data, result):
         return True
 
 
-class SignatureTool:
```
willkg, line 61: This suggested that subclasses followed a standard API. There are two subclasses and they both had very different `_do_generate` methods and thus, there was no standard API. Given that, I removed this untruthful thing since it's not helping any.

socorro/signature/rules.py
```diff
@@ -226,15 +201,7 @@ def normalize_frame(
         names of the fields from the jsonMDSW frame output. This allows this
         function to be invoked by passing a frame as ``**a_frame``.
 
-        Sometimes, a frame may already have a normalized version cached. If
-        that exists, return it instead.
```
willkg, line 230: Because we change the normalization code periodically, it's not helpful to cache the normalized form. Because of that, I wrote code elsewhere to remove the normalized data. So... logically if we're just going to remove it and go out of our way not to use it, we shouldn't be keeping it in the first place. Thus I removed caching of the normalized form.

PR #5750, opened 2021-04-16
bug 1705573: fix correlations tab in signature report

willkg:
This fixes the correlations tab in the signature report so it shows up
all the time (no more hidden tabs) and also moves the "is this a
correlations product" check to JS.

### DIFF
--- a/webapp-django/crashstats/signature/views.py
+++ b/webapp-django/crashstats/signature/views.py
@@ -13,6 +13,7 @@ from django.urls import reverse
 from csp.decorators import csp_update
 from socorro.lib import BadArgumentError
 
+from crashstats import productlib
 from crashstats.crashstats import models, utils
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/signature/views.py
```diff
@@ -374,7 +375,12 @@ def signature_correlations(request, params):
         elif all("esr" in version for version in params["version"]):
             context["channel"] = "esr"
 
-    context["product_name"] = "Firefox"
+    default_product_name = productlib.get_default_product().name
+    product_name = params.get("product", default_product_name)
+    if isinstance(product_name, (list, tuple)):
+        product_name = product_name[0]
```
willkg, line 381: This un-hard-codes the product so if we ever add correlations support for products other than Firefox, it'll work correctly without us having to fix a bunch of hard-coded places.

PR #5752, opened 2021-04-20
bug 1706076, 1706171: fix indexing problems

willkg:
It started out looking into why a test about capturing crash document sizes was flip-flopping between two values and ended up a mild rewrite of a bunch of that code and related code.

The problem stems from us wanting to index data that:

1. is explicitly the data we want to index and nothing more
2. prevent problems indexing data where we have incompatible types between what Elasticsearch inferred and what we later specified in the mapping
3. index data in such a way that we can search across multiple indices

That's all very complicated, but we had sorted most of it out last year.

The problem is that while the raw crash is a one-level data structure, the processed crash is nested and the code that builds the document to be indexed didn't work for nested structures. On top of that, we had a bunch of fields specified with no `storage_mapping`, so they weren't being added to the mapping when we create indices and so those fields would never get indexed.

This PR fixes a bunch of things:

1. removes fields where `is_exposed` and `is_returned` are False indicating that they're not used in SuperSearch, RawCrash, or ProcessedCrash APIs
2. fixes the code that builds the indexable crash document from the raw crash, the processed crash, and the list of valid fields to take into account nested fields
3. fixes the storage_mapping for a bunch of fields including the PHC fields which weren't getting indexed since like January 14th, 2021

To test:

1. run the tests
2. process a bunch of crash reports
3. run some super searches and look at aggregations of json_dump and memory_measure sub fields


socorro/external/es/crashstorage.py
```diff
@@ -222,9 +249,9 @@ def get_keys_for_fields(self):
         return {
             "%s.%s" % (field["namespace"], field["in_database_name"])
             for field in FIELDS.values()
+            if field.get("storage_mapping")
```
willkg, line 252: The mappings are built from the storage mapping for fields. If the field has no storage_mapping set, then nothing gets added to the mapping and if the field is indexed, Elasticsearch infers what to do with it. Which is totally not the situation we want to be in.

This PR changes how it all works. From now on, if it's got no storage_mapping specified, then it doesn't get indexed.

willkg:
This seems fine, however I want it to sit on stage over the weekend and I want to do a prod deploy tomorrow. So I'm going to wait to land it.

PR #5755, opened 2021-04-21
bug 973894: fix super search so all fields can use exists filter

willkg:
This fixes super search so all fields can use "exist/does not exist" filter. Previously, only string fields could.

This makes it a lot easier to search for crash reports that have or don't have a specific field. Previously, you'd have to hack it in some cases and in others, it wasn't possible.

For example, `ProcessType` annotation isn't included in main process crash reports. Previously, you'd have to do a facet on the `process_type` field and then subtract the total number from the total number of crash reports in that same datetime range because there  was no way to search for "crash reports that have no `process_type` field".

I tested this out by processing 50 Firefox crashes and then looking at "exists/does not exist" for these fields:

* boolean: `record_replay`
* str: `phc_kind`
* long: `phc_usable_size`
* enum: `bios_manufacturer`

### DIFF
--- a/socorro/lib/search_common.py
+++ b/socorro/lib/search_common.py
@@ -35,17 +35,18 @@ import socorro.lib.external_common as extern
 """
 OPERATOR_NOT = "!"
 OPERATORS_BASE = [""]
+OPERATORS_EXISTENCE = ["__null__"]
 OPERATORS_BOOL = ["__true__"]
-OPERATORS_STRING = ["__null__", "=", "~", "$", "^", "@"]
+OPERATORS_STRING = ["=", "~", "$", "^", "@"]
 OPERATORS_NUMBER = [">=", "<=", "<", ">"]
 OPERATORS_MAP = {
-    "str": OPERATORS_STRING + OPERATORS_BASE,
-    "int": OPERATORS_NUMBER + OPERATORS_BASE,
-    "date": OPERATORS_NUMBER,
-    "datetime": OPERATORS_NUMBER,
-    "bool": OPERATORS_BOOL,
-    "enum": OPERATORS_BASE,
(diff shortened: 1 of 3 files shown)


socorro/lib/search_common.py
```diff
@@ -35,17 +35,18 @@
 """
 OPERATOR_NOT = "!"
 OPERATORS_BASE = [""]
+OPERATORS_EXISTENCE = ["__null__"]
 OPERATORS_BOOL = ["__true__"]
-OPERATORS_STRING = ["__null__", "=", "~", "$", "^", "@"]
```
willkg, line 39: Previously, only string fields supported exists/not-exists. Now they all do.

socorro/lib/search_common.py
```diff
@@ -35,17 +35,18 @@
...
-    "bool": OPERATORS_BOOL,
-    "enum": OPERATORS_BASE,
-    "default": OPERATORS_BASE,
+    "str": OPERATORS_STRING + OPERATORS_EXISTENCE + OPERATORS_BASE,
+    "int": OPERATORS_NUMBER + OPERATORS_EXISTENCE + OPERATORS_BASE,
+    "date": OPERATORS_NUMBER + OPERATORS_EXISTENCE,
+    "datetime": OPERATORS_NUMBER + OPERATORS_EXISTENCE,
+    "bool": OPERATORS_BOOL + OPERATORS_EXISTENCE,
+    "enum": OPERATORS_EXISTENCE + OPERATORS_BASE,
+    "default": OPERATORS_EXISTENCE + OPERATORS_BASE,
```
willkg, line 49: `OPERATORS_BASE` has to come _after_ everything else since it's the empty string.

PR #5768, opened 2021-05-03
bug 1546800: add supersearch status page to admin

willkg:
This adds a super search status page to the admin where we can see the
list of indices available as well as the latest mapping. The latter will
help debug various super search problems because the mapping specifies
which fields are indexed and how.

### DIFF
--- a/socorro/external/es/super_search_fields.py
+++ b/socorro/external/es/super_search_fields.py
@@ -7,6 +7,7 @@ import logging
 
 from configman import class_converter, Namespace, RequiredConfig
 import elasticsearch
+from elasticsearch_dsl import Search
 
 from socorro.external.es.base import generate_list_of_indexes
 from socorro.lib import datetimeutil, BadArgumentError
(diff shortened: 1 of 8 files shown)


socorro/external/es/super_search_fields.py
```diff
@@ -192,6 +192,26 @@ def get_connection(self):
...
+    def get_supersearch_status(self):
+        """Return list of indices, latest index, and mapping.
+
+        :returns: dict with keys "indices", "latest_index", "mapping"
+
+        """
+        conn = self.get_connection()
+        index_client = elasticsearch.client.IndicesClient(conn)
+        indices = self.context.get_indices()
+        latest_index = sorted(indices)[-1]
```
willkg, line 204: It's possible if Socorro is creating indexes in the future that the latest index isn't the current index. I'm going to ignore that for now but keep it in the back of my head.

socorro/external/es/super_search_fields.py
```diff
@@ -192,6 +192,26 @@ def get_connection(self):
...
+        doctype = self.context.get_doctype()
+
+        mapping = index_client.get_mapping(index=latest_index)
+        properties = mapping[latest_index]["mappings"][doctype]["properties"]
+        return {
+            "indices": indices,
```
willkg, line 210: I was thinking it might be nice to know the index names and number of documents in them. I think I'm going to implement that now.

webapp-django/crashstats/api/views.py
```diff
@@ -95,14 +95,18 @@ def __init__(self, model, *args, **kwargs):
...
+    # It's very sensitive and we don't want to expose it
     "Query",
-    # because it's an internal thing only
+    # This is only used to rapidly process crash reports that haven't been processed
+    # yet, but someone is trying to look at them
     "PriorityJob",
+    # This is used by another API to verify crash data
     "TelemetryCrash",
+    # This is internal and used in the Django admin
+    "SuperSearchStatus",
```
willkg, line 109: This shouldn't be a list. Instead it should be an attribute of the individual classes and default to "private". That's a task for a different day.

webapp-django/crashstats/manage/admin.py
```diff
@@ -100,6 +104,18 @@ def supersearch_fields_missing(request):
     return render(request, "admin/supersearch_fields_missing.html", context)
 
 
+@superuser_required
+def supersearch_status(request):
+    context = {}
+    status = SuperSearchStatus().get()
+    context["indices"] = status["indices"]
+    context["latest_index"] = status["latest_index"]
+    context["mapping"] = json.dumps(status["mapping"], indent=2)
```
willkg, line 113: This is a cheap/easy way to make the structure readable in the web page. Django has a `json_script` template filter, but that puts it in `<script>` tags and we don't want to do that.

PR #5783, opened 2021-05-25
bug 1032227: rework sorting for address field

willkg:
This fixes the address field so that it sorts correctly client-side and server-side.

It pulls in a minidump-stackwalk that left-pads the address data according to the minidump width for the architecture of the process. Once the addresses are left-padded, they sort alphanumerically correctly. Given that, this removes the pre-existing hexdigit sorting (which converted from a hex string to an int and sorted on that).

While doing that, I hit some bugs with sorting other things like graphics adapter values, so I fixed those, too, by generalizing the
sorting fix. This should fix sorting for all the tables in the signature report for the address, graphcs device, and adapter fields.


webapp-django/crashstats/settings/bundles.py
```diff
@@ -194,10 +194,7 @@
         "output_filename": "js/select2.min.js",
     },
     "tablesorter": {
-        "source_filenames": (
-            "tablesorter/dist/js/jquery.tablesorter.js",
-            "crashstats/js/socorro/tablesorter_utils.js",
```
willkg, line 199: I took out tablesorter-utils because we're going to switch to sorting all the hex strings as strings and left-padding the values.

webapp-django/crashstats/signature/jinja2/signature/signature_aggregation.html
```diff
@@ -3,7 +3,7 @@
     <thead>
       <tr>
         <th scope="col">Rank</th>
-        <th scope="col">{{ aggregation | replace('_', ' ') | capitalize }}</th>
+        <th data-field-name="{{ aggregation }}" scope="col">{{ aggregation | replace('_', ' ') | capitalize }}</th>
```
willkg, line 6: This fixes sorting fields that jquery-tablesorter has difficulties sussing the type and sort method for. I generalized the `fieldOverrides` nonsense, but need to capture the field name in a value that I can carry through.

We use the aggregation because it's the super search field name in snake-case.

webapp-django/crashstats/signature/jinja2/signature/signature_summary.html
```diff
@@ -276,8 +276,8 @@ <h2>Graphics Adapter</h2>
     <table class="tablesorter data-table">
       <thead>
         <tr>
-          <th class="header">Vendor</th>
-          <th class="header">Adapter</th>
+          <th data-field-name="adapter_vendor" class="header">Vendor</th>
+          <th data-field-name="adapter_device" class="header">Adapter</th>
```
willkg, line 280: These two fields sometimes sorted correctly, but if all the values were hex strings, they wouldn't sort at all. This fixes that by telling the sort-fixing code the names of these fields so we can override the sorter used.

webapp-django/crashstats/signature/static/signature/js/signature_tab.js
```diff
@@ -109,12 +109,47 @@ SignatureReport.Tab.prototype.buildUrl = function (params, option) {
...
+      var fieldName = elem.getAttribute('data-field-name');
+
+      if (fieldName in fieldOverride) {
+        headers[counter] = fieldOverride[fieldName];
+      }
+      counter += 1;
+    });
+
+  return headers;
+};
```
willkg, line 141: I generalized the code we had in tablesorter-utils and lifted it to here where we can use it across all the tables in the signature report.

webapp-django/crashstats/signature/static/signature/js/signature_tab.js
```diff
@@ -109,12 +109,47 @@ SignatureReport.Tab.prototype.buildUrl = function (params, option) {
...
 SignatureReport.Tab.prototype.onAjaxSuccess = function (contentElement, data) {
   contentElement.empty().append($(data));
   contentElement.addClass('loaded');
+
   if (this.dataDisplayType === 'table') {
-    $('.tablesorter').tablesorter();
+    $('.tablesorter').each(function (index, elem) {
+      var headers = SignatureReport.Tab.prototype.generateTablesorterHeaders($(elem));
+      $(elem).tablesorter({ headers: headers });
+    });
```
willkg, line 152: The summary page does a request and on ajax success, the data is a whole bunch of tables, so I fixed this to run the tablesorter stuff on each table. That helps us fix the headers for each table individually.

webapp-django/crashstats/supersearch/tests/test_views.py
```diff
@@ -422,8 +422,8 @@ def mocked_supersearch_get(**params):
             assert "NightTrain" in params["product"]
 
             assert "address" in params
-            assert "0x0" in params["address"]
-            assert "0xa" in params["address"]
+            assert "0x00000000" in params["address"]
+            assert "0xa0000000" in params["address"]
```
willkg, line 426: address values are now left-padded, so I figured I'd fix these tests accordingly.

PR #5845, opened 2021-08-06
bug 1722626: make mac memory pressure annotations searchable and public

willkg:
ints:

* MacMemoryPressureSysctl:
  * make public in report view
  * add to super search fields                                    
  * always seems to be "1", but the description in the CrashAnnotations.yaml suggests it should be 0, 2, or 4; maybe a bug in their code or the description
* MacAvailableMemorySysctl:
  * make public in report view
  * add to super search fields
  * integer comparison
  * seems to be a valid int > 0                                             

strings:                                                                      

* MacMemoryPressure:                                                          
  * make public in report view
  * add to super search fields                                              
* MacMemoryPressureNormalTime:                                                
  * make public in report view
* MacMemoryPressureWarningTime:                                               
  * make public in report view
* MacMemoryPressureCriticalTime:                                              
  * add to report view

I tested with these crash reports:

Example crash reports:

* 345172f8-2b2f-44de-a097-897c40210806
* cad5a48d-a5a3-49f1-88d0-917510210806
* 4871483d-759d-40a7-ab6a-e48330210806
* bd1c4d4f-331a-4617-b5c6-f70050210806


socorro/processor/rules/mozilla.py
```diff
@@ -38,16 +38,31 @@ class CopyFromRawCrashRule(Rule):
...
+        # Mac memory pressure annotations
+        ("int", "MacMemoryPressureSysctl", "mac_memory_pressure_sysctl"),
+        ("int", "MacAvailableMemorySysctl", "mac_available_memory_sysctl"),
+        ("string", "MacMemoryPressure", "mac_memory_pressure"),
+        ("string", "MacMemoryPressureNormalTime", "mac_memory_pressure_normal_time"),
+        ("string", "MacMemoryPressureWarningTime", "mac_memory_pressure_warning_time"),
+        (
+            "string",
+            "MacMemoryPressureCriticalTime",
+            "mac_memory_pressure_critical_time",
```
willkg, line 54: I looked at writing code that converted CamelCase to snake_case, but crash annotation names are all over the place and I was pretty sure I was going to hit edge cases. I looked for libraries and didn't see any I liked. Then I decided it's better to be able to grep through the code for specific field names and find all the places the field exists than it is to automate generating the name.

socorro/processor/rules/mozilla.py
```diff
@@ -56,6 +71,17 @@ def action(self, raw_crash, dumps, processed_crash, processor_meta):
...
+                try:
+                    processed_crash[processed_key] = int(raw_crash[raw_key])
+                except ValueError:
+                    processor_meta["processor_notes"].append(
+                        f"{raw_key} has a non-int value"
+                    )
+
+            elif value_type == "string":
+                processed_crash[processed_key] = raw_crash[raw_key]
+
```
willkg, line 84: There are probably some rules we can remove and move the annotation moving here. I'll think about that later.

socorro/external/es/super_search_fields.py
```diff
@@ -472,6 +484,105 @@ def flag_field(
     }
 
 
+def keyword_field(
```
willkg, line 487: I like the feel of these field generating functions and it definitely reduces some risk of copy-and-paste errors. Next step is probably to apply them across all the super search fields and see which field behavior categories we're missing.

Also, it might be useful to add an "override" kind of thing. I'm on the fence about that.

PR #5852, opened 2021-08-20
bug 1726032: fix nits in ADRs

willkg:
* change timezone to Etc/UTC
* capitalize a title


willkg:
We're in a code freeze, but these changes are entirely to docs/ and nothing else.

PR #5853, opened 2021-08-30
Update chat channel in docs

willkg:
This updates the docs to point to the new crashreporting matrix channel
and removes the mention of the tools-socorro mailing list which is dead.


willkg:
This needs to wait for codefreeze to be over.

PR #5865, opened 2021-09-20
bug 1730948: improve raw data tab

willkg:
This makes the raw data tab a little easier:

1. it drops use of "raw crash" which is an internal Socorro name for
   "crash annotations" which is more widely understood
2. it removes the API urls which reduces some of the clutter
3. it breaks the dumps out into a separate section


willkg:
This is still not great, but it's a slight improvement that I could make now.

PR #5892, opened 2021-11-08
bug 1697051: add crash_report_keys field

willkg:
This adds a field that we can search using contains/does-not-contain and
aggregate on to help us answer questions like:

1. Which crash reports have file named xyz? For example, which have a
   memory_report.
2. Which crash reports have crash annotation XYZ? For example, which
   have a new crash annotation field we're not indexing yet.

This also adds a processor note with "invalidkeys" to indicate when the
crash report has keys that are malformed.


willkg:
I processed 100 crash reports to look for invalidkeys and didn't see anything, so I think we're good here.

PR #5917, opened 2021-12-01
bug 1743510: capture minidump-stackwalk output and surface it

willkg:
This reworks `execute_process` so that the `MinidumpStackwalkRule` can
surface stderr as `mdsw_stderr`. It adds that to the Debug tab so it's
visible.

The combination of the exit code plus the last lines of stderr will make
debugging minidump-stackwalk easier plus it tells us a lot more about
what's bad about specific minidumps.

### DIFF
--- a/socorro/processor/processor_pipeline.py
+++ b/socorro/processor/processor_pipeline.py
@@ -103,11 +103,13 @@ class ProcessorPipeline(RequiredConfig):
             "format syntax"
         ),
         default=(
-            "timeout --signal KILL {kill_timeout} {command_path} "
-            "--raw-json {raw_crash_path} "
+            "timeout --signal KILL {kill_timeout} "
+            "{command_path} "
+            "--raw-json={raw_crash_path} "
+            "--symbols-cache={symbol_cache_path} "
+            "--symbols-tmp={symbol_tmp_path} "
             "{symbols_urls} "
-            "--symbols-cache {symbol_cache_path} "
-            "--symbols-tmp {symbol_tmp_path} "
+            "--verbose=error "
             "{dump_file_path}"
         ),
     )
--- a/socorro/processor/rules/breakpad.py
+++ b/socorro/processor/rules/breakpad.py
@@ -88,26 +88,23 @@ def tmp_raw_crash_file(tmp_path, raw_crash, crash_id):
 
 
 def execute_process(command_line):
-    """Executes process and returns output and return_code.
+    """Executes process and returns completed process data.
 
     :param command_line: the complete command line to run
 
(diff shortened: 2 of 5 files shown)


socorro/processor/rules/breakpad.py
```diff
@@ -141,11 +138,13 @@ def __init__(
...
-            "--raw-json {raw_crash_path} "
+            "timeout --signal KILL {kill_timeout} "
+            "{command_path} "
+            "--raw-json={raw_crash_path} "
+            "--symbols-cache={symbol_cache_path} "
+            "--symbols-tmp={symbol_tmp_path} "
             "{symbols_urls} "
-            "--symbols-cache {symbol_cache_path} "
-            "--symbols-tmp {symbol_tmp_path} "
+            "--verbose=error "
```
willkg, line 147: It's unenthusing that this is in both places, but the configuration is in a different file and I like having the defaults here because it's at hand.

At some point, I'll figure out a better way to do this.

socorro/processor/processor_pipeline.py
```diff
@@ -103,11 +103,13 @@ class ProcessorPipeline(RequiredConfig):
...
         default=(
-            "timeout --signal KILL {kill_timeout} {command_path} "
+            "timeout --signal KILL {kill_timeout} "
+            "{command_path} "
             "--raw-json {raw_crash_path} "
-            "{symbols_urls} "
             "--symbols-cache {symbol_cache_path} "
             "--symbols-tmp {symbol_tmp_path} "
+            "{symbols_urls} "
+            "--verbose error "
```
willkg, line 112: Oops--I need to switch this to `foo=bar` format.

socorro/processor/rules/breakpad.py
```diff
@@ -408,8 +420,32 @@ def expand_commandline(self, dump_file_path, raw_crash_path):
         }
         return self.command_line.format(**params)
 
+    def execute_process(self, command_line):
```
willkg, line 423: This is the original `execute_process`. I figure `BreakpadStackwalkerRule2015` is deprecated, so I want to stop touching it and creating churn that could break it.

socorro/unittest/processor/rules/test_breakpad.py
```diff
@@ -368,6 +368,11 @@ def test_temp_file_context(self, mocked_unlink):
         mocked_unlink.reset_mock()
 
 
+class ProcessCompletedMock:
+    def __init__(self, **kwargs):
+        self.__dict__.update(kwargs)
```
willkg, line 373: I should undo the laziness here.

PR #5936, opened 2021-12-10
bug 1745454: redo cors support; fix api responses

willkg:
This redoes CORS support to use django-cors-headers middleware. This
adds support for OPTIONS requests as well, which fix preflight issues.

This fixes all the responses from the API to be JSON so we're not doing
some in JSON and some in HTML.

### DIFF
--- a/webapp-django/crashstats/api/tests/test_views.py
+++ b/webapp-django/crashstats/api/tests/test_views.py
@@ -115,10 +115,10 @@ class TestViews(BaseTestViews):
         response = self.client.get(url)
         assert response.status_code == 404
 
-    def test_CORS(self):
-        """any use of model_wrapper should return a CORS header"""
+    def test_option_CORS(self):
+        """OPTIONS request for model_wrapper returns CORS headers"""
         url = reverse("api:model_wrapper", args=("NoOp",))
-        response = self.client.get(url, {"product": "good"})
+        response = self.client.options(url, HTTP_ORIGIN="http://example.com")
         assert response.status_code == 200
         assert response["Access-Control-Allow-Origin"] == "*"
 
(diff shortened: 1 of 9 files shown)


webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -115,10 +115,10 @@ def test_base_classes_raise_not_found(self):
         response = self.client.get(url)
         assert response.status_code == 404
 
-    def test_CORS(self):
-        """any use of model_wrapper should return a CORS header"""
+    def test_option_CORS(self):
+        """OPTIONS request for model_wrapper returns CORS headers"""
         url = reverse("api:model_wrapper", args=("NoOp",))
-        response = self.client.get(url, {"product": "good"})
```
willkg, line 121: This should be an OPTIONS request because it's effectively testing preflight.

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -203,7 +203,7 @@ def test_ProcessedCrash(self):
         url = reverse("api:model_wrapper", args=("ProcessedCrash",))
         response = self.client.get(url)
         assert response.status_code == 400
-        assert response["Content-Type"] == "application/json; charset=UTF-8"
```
willkg, line 206: I changed all the error responses so they use `JsonResponse` and thus the content-type changed.

webapp-django/crashstats/api/views.py
```diff
@@ -226,88 +226,106 @@ def model_wrapper(request, model_name):
...
+    function = None
     if request.method == "POST":
         function = instance.post
-    else:
+    elif request.method == "GET":
         function = instance.get
-    if not function:
-        return http.HttpResponseNotAllowed([request.method])
+    elif request.method == "OPTIONS":
+        function = instance.options
```
willkg, line 235: I added explicit support for OPTIONS here.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -447,6 +450,9 @@ def _post(self, url, payload, method="post"):
         # set dont_cache=True here because the request depends on the payload
         return self.fetch(url, method=method, data=payload, dont_cache=True)
 
+    def _options(self, method="options"):
+        return {}, False
```
willkg, line 454: I don't understand the second bit here. This mimics what `fetch` returns and there's no comments around why that returns a dict and a bool. I need to follow up on this. But I'll do it another day.

PR #5938, opened 2021-12-12
bug 828452: add signature generation api

willkg:
This adds an API endpoint for signature generation: /api/CrashSignature/

It's an HTTP POST API where the payload is one or more jobs in the form
that the signature generation library takes.

This adds django-rest-framework and reimplements the CrashVerify API to
use it. It's also used to implement the new CrashSignature API.

This adjusts the API documentation code to support SocorroAPIView
classes (which are django-rest-framework classes), though the support is
not great. I'll work on improving this over time. It's probably the
case we'll wipe out model_wrapper and turn everything into a
django-rest-framework API class.


webapp-django/crashstats/api/jinja2/api/documentation.html
```diff
@@ -105,17 +105,17 @@ <h2><a href="#{{ endpoint.name }}">{{ endpoint.name }}</a></h2>
             {% endfor %}
           </tbody>
         </table>
-      {% else %}
-        <p>
-          <em>Takes no parameters</em>
-        </p>
```
willkg, line 111: I decided this isn't wildly helpful and it's erroneous for API endpoints that have complicated payloads.

webapp-django/crashstats/api/jinja2/api/documentation.html
```diff
@@ -105,17 +105,17 @@ <h2><a href="#{{ endpoint.name }}">{{ endpoint.name }}</a></h2>
...
-        <button type="submit">Run Test Drive!</button>
-        <img src="{{ static('img/ajax-loader16x16.gif') }}" alt="Loading..." class="loading-ajax">
-        <button type="button" class="close">&times; Close</button>
+        {% if endpoint.test_drive %}
+          <button type="submit">Run Test Drive!</button>
+          <img src="{{ static('img/ajax-loader16x16.gif') }}" alt="Loading..." class="loading-ajax">
+          <button type="button" class="close">&times; Close</button>
+        {% else %}
+          <em>Test drive not supported.</em>
+        {%endif %}
```
willkg, line 118: API endpoints with complicated POST payloads don't work with the test drive stuff, so I added indication that it wasn't supported.

It'd be interesting to get it working at some point, but not today.

webapp-django/crashstats/api/views.py
```diff
@@ -92,29 +101,14 @@ def __init__(self, model, *args, **kwargs):
             self.fields[name] = field_class(required=required)
 
 
-# Names of models we don't want to serve at all
-API_DONT_SERVE_LIST = (
```
willkg, line 96: I never liked this. It's far away from the things it's talking about so to create new API things, you have to make sure to edit this file in a different Django app otherwise your thing ends up public. That seems unhelpful.

Instead, I moved this idea to `IS_PUBLIC` which is a class attribute.

webapp-django/crashstats/api/views.py
```diff
@@ -415,9 +409,9 @@ def _describe_model(model_name, model):
     params = list(model_inst.get_annotated_params())
     params.sort(key=lambda x: (not x["required"], x["name"]))
     methods = []
-    if model.get:
+    if hasattr(model, "get"):
         methods.append("GET")
-    elif model.post:
+    elif hasattr(model, "post"):
         methods.append("POST")
```
willkg, line 415: This should be `getattr(model, "get", None)` because there are some models where `get = None`.

webapp-django/crashstats/api/views.py
```diff
@@ -463,8 +458,27 @@ def dedent_left(text, spaces):
...
+
+    @wraps(fun)
+    def _handle_ratelimit(*args, **kwargs):
+        try:
+            return fun(*args, **kwargs)
+        except Ratelimited:
+            # If the view is rate limited, we throw a 429.
+            raise Throttled()
+
+    return _handle_ratelimit
```
willkg, line 476: ratelimit returns a 403 which is wrong for our uses, so this wraps views and swaps a 403 for a 429.

PR #5939, opened 2021-12-12
bug 1745663: remove mdsw status from signature

willkg:
The mdsw status is always "OK", so it's not helpful in signature
generation.


willkg:
I started looking at what signatures change with this change and decided I missed something and this is a bad idea. Closing it out.

PR #5940, opened 2021-12-13
bug 1745663: fix error codes in crash signatures

willkg:
When minidump-stackwalk encounters an error (empty minidump, no threads
stream, etc), the MinidumpStackwalkerRule should stomp on the
mdsw_status_string with that error code and then signature generation
can grab it to use in the signature.

This updates rust-minidump to ebb603f5:

```
ebb603f Start top-level errors with an error name.
```

Note that this changes some crash signatures.

With breakpad stackwalker, we'd have signatures like:

* EMPTY: no crashing thread identified; ERROR_NO_MINIDUMP_HEADER
* EMPTY: no crashing thread identified; ERROR_NO_THREAD_LIST

With rust-minidump minidump-stackwalker and these changes, we'll now
have:

* EMPTY: no crashing thread identified; EmptyMinidump
* EMPTY: no crashing thread identified; MissingThreadList

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -27,8 +27,8 @@ RUN groupadd --gid $groupid app && \
 USER app
 
 # From: https://github.com/luser/rust-minidump
-ARG MINIDUMPREV=6ccdead2bffb0c9db60037b1ea6923c69afba023
-ARG MINIDUMPREVDATE=2021-12-10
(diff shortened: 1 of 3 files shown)


willkg:
I'll keep an eye on crash signatures with EMPTY in them and we'll see how they change over the next 24 hours and whether we caught everything.

PR #5961, opened 2022-01-05
bug 1748484: add fake sentry service

willkg:
This adds a fake sentry service (kent) and configures the local dev
environment to use it.

### DIFF
--- a/docker/Dockerfile.fakesentry
+++ b/docker/Dockerfile.fakesentry
@@ -13,7 +13,7 @@ ENV PYTHONUNBUFFERED=1 \
     PYTHONDONTWRITEBYTECODE=1
 
 RUN pip install -U 'pip>=8' && \
-    pip install --no-cache-dir 'kent==0.1.0'
+    pip install --no-cache-dir 'kent==0.2.0'
 
 USER app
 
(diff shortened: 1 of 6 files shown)


docker/Dockerfile.fakesentry
```diff
@@ -0,0 +1,21 @@
...
+
+RUN addgroup -g $groupid app && \
+    adduser --disabled-password --gecos "" --home /home/app --ingroup app --uid $userid app && \
+    chown app:app /app/
+
+ENV PYTHONUNBUFFERED=1 \
+    PYTHONDONTWRITEBYTECODE=1
+
+RUN pip install -U 'pip>=8' && \
+    pip install --no-cache-dir 'https://github.com/willkg/kent/archive/refs/heads/main.zip'
```
willkg, line 16: This is currently tied to main tip. I'll change that once I push out the first version. I want this to go through CI first to catch any bugs there and then I want to write a webapp integration test.

docker/config/local_dev.env
```diff
@@ -38,7 +38,7 @@ resource.metrics.markus_backends=markus.backends.logging.LoggingMetrics,markus.b
 # sentry
 # ------
 
-# secrets.sentry.dsn=
+secrets.sentry.dsn=http://public@fakesentry:8090/1
```
willkg, line 41: These are all from the point-of-view of the webapp, crontabber, and processor containers.

PR #5965, opened 2022-01-05
bug 1748248: fix mac_crash_info field to require records

willkg:
If `num_records` is 0, then treat that as if there was no `mac_crash_info` data in the `json_dump`. This allows rust-minidump to always emit the field, but Socorro will only extract it into the processed crash if there are records.

### DIFF
--- a/socorro/unittest/processor/rules/test_mozilla.py
+++ b/socorro/unittest/processor/rules/test_mozilla.py
@@ -740,37 +740,53 @@ class TestMacCrashInfoRule:
     @pytest.mark.parametrize(
         "processed, expected",
         [
+            # These shouldn't result in a mac_crash_info
             ({}, False),
             ({"json_dump": {}}, False),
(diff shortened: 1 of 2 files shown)


socorro/unittest/processor/rules/test_mozilla.py
```diff
@@ -740,37 +740,53 @@ class TestMacCrashInfoRule:
...
-        assert rule.predicate(raw_crash, dumps, processed, processor_meta) == expected
+        rule.action(raw_crash, dumps, processed, processor_meta)
+
+        assert ("mac_crash_info" in processed) == expected
 
     def test_mac_crash_info_action(self):
         raw_crash = {}
         dumps = {}
-        processed_crash = {"json_dump": {"mac_crash_info": {"key": "val"}}}
+        processed_crash = {
```
willkg, line 765: This also fixes the test to use values that are more correct.

PR #5972, opened 2022-01-11
bug 1749561: fix supersearch refine links

willkg:
Previously, if you do a supersearch and refined the search by clicking
on one of the field values in the results or facet results, it would add
that as an "is" query to the search. That's fine, except when there's a
filter for that field. In that case, the "is" query is OR'd with the
existing filter and I can't think of a single case where that's what you
want.

Now the refine links replace any existing filters for the field with the
"is" query.

Further, it fixes product, version, and product_type refinements to use
a term query (no prefix) instead of an "is" query ("=" prefix) so that
it works correctly with the supersearch form and site navigation.

Then I moved the code to macros because the template is really messy and
hard to read and we have a lot of complex-looking repeated logic
everywhere.

### DIFF
--- a/webapp-django/crashstats/supersearch/jinja2/supersearch/macros/links.html
+++ b/webapp-django/crashstats/supersearch/jinja2/supersearch/macros/links.html
@@ -1,0 +1,21 @@
+{% macro signature_link(name, val) %}
+  <a href="{{ url('signature:signature_report') }}{{ change_query_string(
+    _no_base=True,
+    signature=val,
+    page=None,
+    _columns=None,
+    _facets=None
+    ) }}"
+    title="View signature report"
+  >
+    {{ val }}
+  </a>
+{% endmacro %}
+
+{% macro refine_link(name, val, display_text=None, use_is_query=True) %}
+  {% set filter_val = '=%s' % val if use_is_query else val %}
+  {% set display_text = val if display_text == None else display_text %}
+  <a href="{{ change_query_string(_no_base=True, **{name: filter_val}) }}" class="term" data-field="{{ name }}" data-content="{{ val }}" title="Refine search">
+    {{ display_text }}
+  </a>
+{% endmacro %}
--- a/webapp-django/crashstats/supersearch/jinja2/supersearch/search_results.html
+++ b/webapp-django/crashstats/supersearch/jinja2/supersearch/search_results.html
@@ -1,4 +1,5 @@
 {% from "macros/pagination.html" import pagination %}
+{% from "supersearch/macros/links.html" import signature_link, refine_link with context %}
 
 {% if query.total > 0 %}
   <ul id="search_results-nav">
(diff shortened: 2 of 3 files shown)


webapp-django/crashstats/supersearch/jinja2/supersearch/search_results.html
```diff
@@ -30,35 +31,19 @@
...
-                       signature=crash.signature,
-                       page=None,
-                       _columns=None,
-                       _facets=None
-                       ) }}"
-                       title="Click to leave this search and go to this signature report"
-                    >
-                      {{ crash[column] }}
-                    </a>
+                    {{ signature_link(column, crash[column]) }}
```
willkg, line 34: `crash[column]` is the same value as `crash.signature` here.

webapp-django/crashstats/supersearch/jinja2/supersearch/search_results.html
```diff
@@ -30,35 +31,19 @@
...
+                    {{ refine_link(column, crash[column], "Add term") }}
                   {% elif is_list(crash[column]) %}
                     {% for val in crash[column] %}
-                      <a href="{{ current_url }}&amp;{{ column }}=%3D{{ val | urlencode }}" class="term" data-field="{{ column }}" data-content="{{ val }}" title="Add this term to the search form">
-                        {{ val }}
-                      </a>
+                      {{ refine_link(column, val) }}
                     {% endfor %}
+                  {% elif column in ('product', 'version', 'platform', 'process_type') %}
+                    {{ refine_link(column, crash[column], use_is_query=False) }}
```
willkg, line 42: I added this section because these fields are special. `product` and `version` also affect site navigation and other things and the value has to be a full value and can't start with `=`. `platform` and `process_type` go in the supersearch form and if the values start with `=`, then that doesn't work right.

webapp-django/crashstats/supersearch/jinja2/supersearch/macros/links.html
```diff
@@ -0,0 +1,21 @@
...
+    title="View signature report"
+  >
+    {{ val }}
+  </a>
+{% endmacro %}
+
+{% macro refine_link(name, val, display_text=None, use_is_query=True) %}
+  {% set filter_val = '=%s' % val if use_is_query else val %}
+  {% set display_text = val if display_text == None else display_text %}
+  <a href="{{ change_query_string(_no_base=True, **{name: filter_val}) }}" class="term" data-field="{{ name }}" data-content="{{ val }}" title="Refine search">
```
willkg, line 18: I changed the title of the link. This one is shorter and clearer.

webapp-django/crashstats/supersearch/jinja2/supersearch/search_results.html
```diff
@@ -90,22 +75,14 @@
...
-                     signature=hit.term,
-                     page=None,
-                     _columns=None,
-                     _facets=None
-                     ) }}"
-                     title="Click to leave this search and go to this signature report"
-                  >
-                    {{ hit.term }}
-                  </a>
+                {% if facet == 'signature' %}
```
willkg, line 78: I reversed this conditional so the first block is if `facet == 'signature'`.

webapp-django/crashstats/supersearch/jinja2/supersearch/search_results.html
```diff
@@ -90,22 +75,14 @@
...
+                  {{ signature_link(facet, hit.term) }}
                   &nbsp;
-                  <a href="{{ current_url }}&amp;{{ facet }}=%3D{{ hit.term | urlencode }}" class="term" data-field="{{ facet }}" data-content="{{ hit.term }}" title="Add this term to the search form">Add term</a>
+                  {{ refine_link(facet, hit.term, "Add term") }}
+                {% elif facet in ('product', 'version', 'platform', 'process_type') %}
+                  {{ refine_link(facet, hit.term, use_is_query=False) }}
```
willkg, line 83: I added this section because these are special fields that affect site navigation and the supersearch form.

webapp-django/crashstats/supersearch/tests/test_views.py
```diff
@@ -216,7 +216,9 @@ def mocked_supersearch_get(**params):
         assert '<th scope="col">Bugs</th>' in smart_text(response.content)
         assert "123456" in smart_text(response.content)
         # Test links on terms are existing
-        assert "product=%3DWaterWolf" in smart_text(response.content)
+        assert "build_id=%3D888981" in smart_text(response.content)
```
willkg, line 219: Adding a test for `build_id` here tests the refine link where `use_is_query=True`.

willkg:
I need to add a test for urlencoding the value to make sure that's correct.

PR #5977, opened 2022-01-25
bug 1752001: replace sentry-cli with sentry-wrap.py

willkg:
We use sentry-cli to get a heads-up when migrations fail during deploys. sentry-cli requires npm to install. That's a problem when we move to building JS/CSS files in a separate frontend container.

This drops sentry-cli in favor of a new sentry-wrap.py Python script.

### DIFF
--- a/bin/sentry-wrap.py
+++ b/bin/sentry-wrap.py
@@ -1,0 +1,69 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Wraps a command such that if it fails, an error report is sent to the Sentry service
+# specified by SENTRY_DSN in the environment.
+#
+# Usage: python bin/sentry-wrap.py -- [CMD]
+
+
+import argparse
+import os
+import shlex
+import subprocess
+import sys
+
+import sentry_sdk
+from sentry_sdk import capture_exception, capture_message
+
+
+def main():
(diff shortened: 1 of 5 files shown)


bin/sentry-wrap.py
```diff
@@ -0,0 +1,69 @@
...
+    sentry_dsn = os.environ.get("SENTRY_DSN")
+
+    if not sentry_dsn:
+        print("SENTRY_DSN is not defined. Exiting.")
+        sys.exit(1)
+
+    sentry_sdk.init(sentry_dsn)
+
+    cmd = " ".join(args.cmd)
+    cmd_args = shlex.split(cmd)
```
willkg, line 44: This might not be quite right, but I figure we can deal with this when complex scenarios arise.

bin/sentry-wrap.py
```diff
@@ -0,0 +1,69 @@
...
+    print(f"Running: {cmd_args}")
+
+    try:
+        ret = subprocess.run(cmd_args, capture_output=True, timeout=args.timeout)
```
willkg, line 48: If this times out, it throws a `TimeoutException` or something like that and that'll get handled by `capture_exception()` below.

docker/set_up_ubuntu.sh
```diff
@@ -27,9 +27,6 @@ PACKAGES_TO_INSTALL=(
     # For scripts
     git
 
-    # For sentry-cli
-    gawk
```
willkg, line 31: This was used by the sentry-cli bash hook.

PR #5992, opened 2022-02-01
bug 1703252: set id default to AutoField and undo autofield migrations

willkg:
This undoes the migrations created in
f30f9ab549e7922e6974fe8fdf4960499331d3aa and changes the
DEFAULT_AUTO_FIELD setting set in
6ce4d969c38f074598dd2f6afb9e3b9a10a952a3.

This sets the autofield default to AutoField which is what it was
before. This requires no db migrations.

Any migrations previously applied should be undone _before_ landing this
change.

```
python webapp-django/manage.py migrate authentication 0001
python webapp-django/manage.py migrate crashstats 0022
python webapp-django/manage.py migrate cron 0005
python webapp-django/manage.py migrate status 0004
python webapp-django/manage.py migrate tokens 0001
```


willkg:
Jason ran the anti-migrations in stage, so we should be all set. Merging this now.

PR #5996, opened 2022-02-04
bug 1753550: recover after cache.set error

willkg:
Apparently we have some data we want to cache that's larger than the
memcached max size and that's causing cache.set() to error out.

This wraps that in a try/except block and emits a cache_set_error metric
so the code "recovers", but also lets us monitor the situation in case
we want to do something different when we learn more.

### DIFF
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -21,6 +21,9 @@ from django.urls import reverse
 from django.utils.encoding import iri_to_uri
 from django.utils.module_loading import import_string
 
+import markus
+from pymemcache.exceptions import MemcacheServerError
+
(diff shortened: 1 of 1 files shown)


webapp-django/crashstats/crashstats/models.py
```diff
@@ -384,7 +390,10 @@ def fetch(
         implementation_method = getattr(implementation, method)
         result = implementation_method(**params)
         if cache_key:
-            cache.set(cache_key, result, self.cache_seconds)
+            try:
+                cache.set(cache_key, result, self.cache_seconds)
+            except MemcacheServerError:
+                metrics.incr("cache_set_error")
```
willkg, line 396: We probably want to know what other crash ids this happens with. I'll figure out how to log this later.

PR #6001, opened 2022-02-07
bug 1743704: remove jit crash classifier

willkg:
The JitCrashClassifierRule runs _after_ signature generation, but
signature generation has a SignatureJitRule that prefixes "jit" to
signatures that have "jit_cateogry" which comes from
"classification.jit.category" which is set by JitCrashClassifierRule.
That's circular and thus signatures will never have "jit" prefix unless
they were reprocessed!

jit-crash-classifier comes from the old Breakpad-based
minidump-stackwalk which we stopped using in December
2021 in favor of rust-minidump. rust-minidump doesn't have a
jit-crash-classifier equivalent (yet). In a discussion about this
situation, it was said that the jit-crash-classifier often has false
positives, so it's not useful.

This changes the telemetry_socorro_crash.json schema by removing two
fields one of which is required, so I increased the version number to 3.

We don't need to edit the `telemetry_socorro_crash.json` schema. It's ok if Socorro processor omits the fields.

I didn't remove the minidump-stackwalk bits from the Docker image because we've still got the `BreakpadStackwalkerRule2015` code. Bug 1754078 covers removing that. Frankly, I'm surprised and chagrined that the tests pass even if the binary is gone. MOCKTILLYOUDROP!


willkg:
I need someone from DE to review the schema change. I'm pretty sure I did it right, but it'd be best to get a second set of eyes.

willkg:
I forgot we still have the `BreakpadStaclkwalker2015Rule`, so I can't remove the bits from the docker image. I'll do another bug for removing all the old stackwalker stuff.

PR #6003, opened 2022-02-08
bug 1750759: add page view tracking

willkg:
This adds page view tracking which captures time-to-generate, whether
it's an API endpoint, whether it's an AJAX request, and some other
things. No ip addresses or user information is captured. This isn't
remotely equivalent to Google Analytics.

Regardless, this will help us see site usage at a coarse level.

While doing that, I redid how paths are specified and fixed a bunch of
tests that were using fake data.

While doing that, I ran "django-upgrade --target-version=3.2" on all the
files which adjusted some code that looks at request HTTP headers.

### DIFF
--- a/webapp-django/crashstats/crashstats/tests/test_views.py
+++ b/webapp-django/crashstats/crashstats/tests/test_views.py
@@ -8,6 +8,7 @@ import re
 from unittest import mock
 
 import pyquery
+from markus.testing import MetricsMock
 
 from django.conf import settings
 from django.contrib.auth import REDIRECT_FIELD_NAME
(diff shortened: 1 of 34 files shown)


webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -453,6 +456,16 @@ def mocked_processed_crash_get(**params):
         # shows up.
         assert "The crashing address." in content
 
+        metrics_mock.assert_timing(
+            "webapp.view.pageview",
+            tags=[
+                "status:200",
+                "ajax:false",
+                "api:false",
+                "path:report/index/_crashid_crash_id_",
```
willkg, line 465: Oh, hrm... This might not work because it's not the full route but rather the part of the route defined in the matching resolver thing.

I need to add tests to the other views and see if it does what I want it to do. If it doesn't, I'll have to think of something else.

PR #6008, opened 2022-02-16
bug 1753521: refactor indexing and start migrating some raw crash fields

willkg:
This refactors indexing to make it more flexible around moving data around. Also, it starts the move to my ultimate vision where analytics and data flow rely solely on processed (normalized, validated, redacted, etc) data rather than the original values from the crash annotations in the crash report.

Towards that aim, this also starts migrating indexing from pulling from the raw crash and indexing into the `raw_crash` namespace to pull from the processed crash and indexing into the `processed_crash` namespace. We're going to do a slow migration because that's all we have support for at the moment. In 5 months, we'll remove the raw crash sides of these things.

There's a handful of followup bugs from this work to cover issues I bumped into and decided to carve out as separate bugs.

After this lands, we still have to migrate the boolean-ish fields that are being pulled from the raw crash. We have to figure out how to process those and unify how we index them--right now we do a variety of different things and it's a right mess.

To test this, I did a few things:

1. Generate a mapping using the `main` branch (`socorro-cmd es list` and then `socorro-cmd es print_mapping`) and compare with the mapping generated from this branch. The `raw_crash` namespace should be identical between the two mappings. The `processed_crash` namespace has a bunch of new fields in it.
2. Process crash reports with the main branch and capture their Elasticsearch document sources (`curl "http://elasticsearch:9200/${INDEX}/crash_reports/${CRASHID}" > ${CRASHID}.document.json`) and then switch to this branch, process the same crash reports, capture their Elasticsearch document sources, and compare the two.

   `raw_crash.UptimeTS` switched from a string value to a float. It wasn't being converted correctly in the old indexing code which didn't handle "double".

   `raw_crash.ContentSandboxCapabilities` switched from a string value to a short. It also wasn't being converted correctly in the old indexing code which didn't handle "short".

   There should be a bunch of additional data in the `processed_crash` namespace--that's the data we're migrating over.
3. Supersearch should continue to work--both querying and faceting.
4. Signature generation should be based mostly on the processed crash data (I think I have one more field to fix--that's covered in another bug). While signature generation looks at some different fields for some data, signatures shouldn't change with these changes.

The one thing I didn't check is `TelemetryBotoS3CrashStorage` data. I'm pretty sure none of these changes affect how that's generated. I wrote up another bug to rework that so it's not using `super_search_fields.py` to build its crash data.


socorro/external/es/crashstorage.py
```diff
@@ -31,28 +34,6 @@
...
-
-    """
-    # FIXME(willkg): These should be specified in super_search_fields.py
-    # and not hard-coded
-    datetime_fields = [
-        "submitted_timestamp",
-        "date_processed",
-        "client_crash_date",
-        "started_datetime",
-        "completed_datetime",
```
willkg, line 47: I'm so glad to get rid of this.

socorro/external/es/crashstorage.py
```diff
@@ -87,185 +68,219 @@ def is_valid_key(key):
...
+        return string_to_datetime(value)
+    if not isinstance(value, (datetime.datetime, datetime.time)):
+        return None
+    return value
 
-    "long" can be -9,223,372,036,854,775,808
 
-    :arg dict fields: the super search fields schema
-    :arg dict data: the data to look through
+def build_document(src, crash_document, fields, all_keys):
```
willkg, line 227: This way of building the document is much better than the previous way in that it's much easier to reason about, iterates over fields once, and makes the related code easier to test and modify over time.

socorro/processor/rules/mozilla.py
```diff
@@ -54,16 +54,58 @@ class CopyFromRawCrashRule(Rule):
...
+        ("string", "RemoteType", "remote_type"),
+        ("string", "ShutdownProgress", "shutdown_progress"),
+        ("int", "StartupTime", "startup_time"),
+        ("int", "SystemMemoryUsePercentage", "system_memory_use_percentage"),
+        ("int", "TotalPageFile", "total_page_file"),
+        ("int", "TotalPhysicalMemory", "total_physical_memory"),
+        ("int", "TotalVirtualMemory", "total_virtual_memory"),
+        ("float", "UptimeTS", "uptime_ts"),
+        ("string", "useragent_locale", "useragent_locale"),
+        ("string", "Vendor", "vendor"),
```
willkg, line 108: This is not a great way of doing things, but it's a small step in the right direction.

At some point, we'll write a schema that specifies which fields are sourced from the raw crash and what their types are. Then this rule will use that to do its thing.

socorro/processor/rules/mozilla.py
```diff
@@ -309,12 +359,8 @@ def action(self, raw_crash, dumps, processed_crash, processor_meta):
         if crash_time == submitted_timestamp_epoch:
             processor_notes.append("client_crash_date is unknown")
 
-        # StartupTime: must have started up some time before crash
-        try:
-            startup_time = int(raw_crash.get("StartupTime", crash_time))
-        except ValueError:
-            startup_time = 0
-            processor_notes.append('non-integer value of "StartupTime"')
```
willkg, line 317: We're now doing validation and normalization in the `CopyFromRawCrashRule` so we don't need to validate the data here.

socorro/external/es/super_search_fields.py
```diff
@@ -2923,6 +3253,11 @@ def number_field(
         "name": "useragent_locale",
         "namespace": "raw_crash",
         "permissions_needed": [],
+        "source_key": "processed_crash.useragent_locale",
+        "destination_keys": [
+            "raw_crash.useragent_locale",
```
willkg, line 3258: The crash annotation is snake-case, so this isn't a mistake.

socorro/external/es/super_search_fields.py
```diff
@@ -1624,6 +1869,11 @@ def number_field(
         "name": "ipc_channel_error",
         "namespace": "raw_crash",
         "permissions_needed": [],
+        "source_key": "processed_crash.ipc_channel_error",
+        "destination_keys": [
+            "raw_crash.ipc_channel_error",
```
willkg, line 1874: The crash annotation is snake-case, so this isn't a mistake.

willkg:
I double-checked the spelling of everything in `super_search_fields.py` and in the `CopyFromRawCrash` rule. I think I've been pretty meticulous with this, so it's time to land it. If there are still issues, we'll fix them as we find them.

PR #6013, opened 2022-02-23
bug 1755523: end raw crash use in signature generation

willkg:
This fixes signature generation so that it only uses processed crash
data.

### DIFF
--- a/socorro/signature/cmd_signature.py
+++ b/socorro/signature/cmd_signature.py
@@ -217,27 +217,6 @@ def main(argv=None):
 
             crash_id = parsed_crash_id
 
-            resp = fetch("/RawCrash/", crash_id, api_token)
-            if resp.status_code == 404:
-                out.warning(f"{crash_id}: does not exist.")
-                continue
-            if resp.status_code == 429:
-                out.warning(f"API rate limit reached. {resp.content}")
-                # FIXME(willkg): Maybe there's something better we could do here. Like maybe wait a
-                # few minutes.
-                return 1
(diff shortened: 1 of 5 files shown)


socorro/signature/cmd_signature.py
```diff
@@ -217,27 +217,6 @@ def main(argv=None):
...
-                out.warning(f"HTTP 500: {resp.content}")
-                continue
-
-            raw_crash = resp.json()
-
-            # If there's an error in the raw crash, then something is wrong--probably with the API
-            # token. So print that out and exit.
-            if "error" in raw_crash:
-                out.warning(f"Error fetching raw crash: {raw_crash['error']}")
-                return 1
```
willkg, line 239: This is nice because it means we only have to pull down one thing from Socorro now.

socorro/signature/utils.py
```diff
@@ -109,8 +107,10 @@ def convert_to_crash_data(raw_crash, processed_crash):
         "ipc_message_name": glom(processed_crash, "ipc_message_name", default=None),
         # text
         "moz_crash_reason": glom(processed_crash, "moz_crash_reason", default=None),
-        # text; comma-delimited e.g. "browser,flash1,flash2"
-        "additional_minidumps": glom(raw_crash, "additional_minidumps", default=""),
+        # list of str; e.g. ["browser", "flash1", "flash2"]
```
willkg, line 110: Oops--that comment is wrong. Need to fix that.

PR #6028, opened 2022-03-09
bug 1756223: update docker image and remote docker version

willkg:
* switch to `cimg/base:2022.02`
* update remote docker version
* stop using `/socorro` as working directory (we don't have access to it)
* parallelize jobs so that lint, testsocorro, and testwebapp all run at the same time
* move dockerhub push code to an external script that can be reused with other projects
* fix yamllint issues
* reduce some yaml redundancy


.circleci/config.yml
```diff
@@ -37,111 +53,176 @@ jobs:
...
+          command: |
+            docker load -i workspace/socorro_app.tar.gz
+
       - run:
           name: Run tests
-          working_directory: /socorro
           command: |
-            make test-ci
+            make my.env
+            ls -l my.env
```
willkg, line 139: Oops--I can take this out.

PR #6029, opened 2022-03-10
bug 1758701: add sentry-test mode to sentry-wrap

willkg:
This improves error handling in sentry-wrap, switches it to use click instead of argparse, and adds "sentry-test" and "wrap-process" subcommands.

This updates the `bin/run_migrations.sh` script accordingly.

This also switches all parts of Socorro to use `SENTRY_DSN` environment variable to specify the Sentry host to use. Previously, crontabber and the webapp used `SENTRY_DSN` and the processor and upload_telemetry_schema apps used `secrets.sentry.dsn`.

### DIFF
--- a/bin/sentry-wrap.py
+++ b/bin/sentry-wrap.py
@@ -26,8 +26,31 @@ import sentry_sdk
 from sentry_sdk import capture_exception, capture_message
 
 
-@click.command()
-@click.option("--timeout", default=300, help="Timeout in seconds to wait for process before giving up.")
+@click.group()
+def cli_main():
+    pass
+
+
+@cli_main.command()
+@click.pass_context
+def sentry_test(ctx):
+    sentry_dsn = os.environ.get("SENTRY_DSN")
+
+    if not sentry_dsn:
+        print("SENTRY_DSN is not defined. Exiting.")
+        sys.exit(1)
+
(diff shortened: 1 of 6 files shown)


bin/sentry-wrap.py
```diff
@@ -1,51 +1,78 @@
...
+
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 # Wraps a command such that if it fails, an error report is sent to the Sentry service
 # specified by SENTRY_DSN in the environment.
 #
-# Usage: python bin/sentry-wrap.py -- [CMD]
+# Usage: python bin/sentry-wrap.py wrap -- [CMD]
```
willkg, line 10: Oops--these are wrong.

bin/sentry-wrap.py
```diff
@@ -55,16 +82,28 @@ def main():
                     "stderr": ret.stderr.decode("utf-8"),
                 },
             )
-            capture_message(f"Command {args.cmd} failed.")
+            capture_message(f"Command {cmd!r} failed.")
+            print(ret.stdout.decode("utf-8"))
```
willkg, line 86: Oops--these should use `click.echo`.

PR #6039, opened 2022-03-25
bug 1755528: fix flag/boolean fields

willkg:
* redo how we handle flag/boolean fields in the processor and Elasticsearch crashstorage
* fix throttleable
* fix submitted_from_infobar
* fix safe_mode
* fix is_garbage_collecting
* fix has_device_touch_screen
* fix em_check_compatibility
* fix content_sandbox_enabled
* fix content_sandbox_capable
* fix addons_checked to be bool
* fix accessibility

This also fixes startup_crash, but leaves it as "null_value is False" which differs from the rest of them. This is used in the signature reports and I didn't want to spend the time to figure out what effects changing that would have now.

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -253,16 +253,6 @@ def build_document(src, crash_document, fields, all_keys):
             else:
                 value = fix_string(value, max_size=MAX_STRING_FIELD_VALUE_SIZE)
 
-        elif storage_type == "boolean":
-            # FIXME(willkg): Previous iteration of the crash storage logic had a
-            # convert_booleans which was implemented wrong, so it didn't do anything.
-            # "booleans" is tricky since we have flags and booleans and they're
-            # different, so it requires more analysis to figure out what to do here.
-            # Meanwhile, I commented this out for now which returns us to the previous
-            # behavior.
-            # value = fix_boolean(value)
-            pass
-
         elif storage_type == "integer":
             value = fix_integer(value)
             if value is None:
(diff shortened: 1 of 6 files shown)


socorro/external/es/crashstorage.py
```diff
@@ -253,16 +253,6 @@ def build_document(src, crash_document, fields, all_keys):
...
 
-        elif storage_type == "boolean":
-            # FIXME(willkg): Previous iteration of the crash storage logic had a
-            # convert_booleans which was implemented wrong, so it didn't do anything.
-            # "booleans" is tricky since we have flags and booleans and they're
-            # different, so it requires more analysis to figure out what to do here.
-            # Meanwhile, I commented this out for now which returns us to the previous
-            # behavior.
-            # value = fix_boolean(value)
-            pass
```
willkg, line 264: We're no longer doing any validation or adjustments for boolean values in the crashstorage code--that's now being done in the processor.

socorro/external/es/super_search_fields.py
```diff
@@ -1689,6 +1699,11 @@ def number_field(
         "name": "em_check_compatibility",
         "namespace": "raw_crash",
         "permissions_needed": [],
+        "source_key": "processed_crash.em_check_compatibility",
+        "destination_keys": [
+            "raw_crash.EMCheckCompatibility",
+            "processed_crash.em_check_compatibility",
```
willkg, line 1705: I should add validation of this to the tests. That'll mean I don't have to eyeball this very carefully.

PR #6040, opened 2022-03-25
bug 1755528: fix flag/boolean fields that are shorts

willkg:
There are 5 flag/boolean fields that are being indexed as shorts:

* contains_memory_report
* dom_fission_enabled
* dom_ipc_enabled
* gmp_plugin
* graphics_startup_test

I tested out whether search would work across indexes where these fields were shorts in one index and booleans in another and that doesn't work. It doesn't kick up errors, but it doesn't match values correctly.

We're going to need to migrate from one type to another. This starts accumulating the data in temporary fields. In August 2022, we can delete the old fields and switch to the new fields permanently.

This also fixes `socorro-cmd es delete` which didn't work right.

### DIFF
--- a/socorro/external/es/super_search_fields.py
+++ b/socorro/external/es/super_search_fields.py
@@ -194,32 +194,32 @@ def build_mapping(doctype, fields=None):
     return mapping
 
 
-def is_doc_values_friendly(field_value):
-    """Predicate denoting whether this field should have doc_values added
+def is_doc_values_friendly(storage_value):
+    """Predicate denoting whether this storage should have doc_values added
 
-    ``doc_values=True`` is a thing we can add to certain fields to reduce the
+    ``doc_values=True`` is a thing we can add to certain storages to reduce the
     memory they use in Elasticsearch.
 
     This predicate determines whether we should add it or not for a given
-    field.
+    storage.
 
-    :arg field_value: a field value from super search fields
+    :arg storage_value: a storage value from super search storages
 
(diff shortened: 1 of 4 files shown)


socorro/external/es/super_search_fields.py
```diff
@@ -194,32 +194,32 @@ def build_mapping(doctype, fields=None):
...
+    storage_type = storage_value.get("type")
 
     # No clue what type this is--probably false
-    if not field_type:
+    if not storage_type:
         return False
 
-    # object, and multi_fields fields don't work with doc_values=True
-    if field_type in ("object", "multi_field"):
+    # object, and multi_storages storages don't work with doc_values=True
```
willkg, line 217: oops--this needs to be undone.

socorro/external/es/super_search_fields.py
```diff
@@ -1448,6 +1449,31 @@ def number_field(
         "query_type": "flag",
         "storage_mapping": {"type": "short"},
     },
+    # NOTE(willkg): This populates the new location in processed_crash with data with
+    # the right type. In 8/2022, we can:
+    #
+    # 1. remove contains_memory_report (deprecated)
+    # 2. rename this one to contains_memory_report
+    # 3. change is_exposed and is_returned to True
+    # 4. change the namespace and in_database_name and remove source/destination keys
```
willkg, line 1458: Hopefully future me remembers this and undoes it. I'll probably write up a bug for it.

PR #6042, opened 2022-03-26
Import ABC from collections.abc for Python 3.10 compatibility.

tirkarthi:
Importing ABC from `collections` was deprecated since Python 3.4 and removed in Python 3.10 . Use `collections.abc` instead.


willkg (approved):

willkg:
Thank you!

PR #6045, opened 2022-03-29
Make sure fakesentry is up before running tests

willkg:
This adds a waitfor for fakesentry to make sure it's up and running before the tests start. Hopefully this fixes the issues CI is having today.

This also changes waitfor to default to 15 seconds which matches urlwait.

This also removes the hard-coded database value from the test.sh script.


willkg:
Ugh. So frustrating.

PR #6052, opened 2022-04-01
bug 1755095: fix TelemetryBotoS3CrashStorage to not use super search fields

willkg:
This redoes TelemetryBotoS3CrashStorage to build the crash document it
sends to telemetry ingestion solely from the processed crash.

This no longer uses super search fields and reduces some of the tangling
between crash storage systems which is creating a lot of problems.

The processed crash contains data that has been normalized and
validated, so it's a better source to use than the raw crash. This
handles a few historical naming issues, but otherwise it's pretty
straight-forward and driven entirely by the telemetry_socorro_crash.json
schema.

### DIFF
--- a/socorro/external/boto/crashstorage.py
+++ b/socorro/external/boto/crashstorage.py
@@ -15,7 +15,6 @@ from socorro.external.crashstorage_base import (
     CrashIDNotFound,
     MemoryDumpsMapping,
 )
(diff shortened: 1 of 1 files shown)


willkg:
Let's land this and see how it fares on stage. I tested it with 1,000 crash reports and it seems fine, but testing it on stage over the weekend will be more comprehensive.

PR #6059, opened 2022-04-02
bug 1762207: remove hang_type

willkg:
hang_type is based on Hang and HangID crash annotations which were
removed 10 years ago. This removes them from Socorro.

### DIFF
--- a/socorro/signature/utils.py
+++ b/socorro/signature/utils.py
@@ -22,22 +22,15 @@ def get_crashing_thread(crash_data):
     """
     Takes crash data and returns the crashing thread as an int.
 
-    If it's a chrome hang (hang_type=1), then use thread 0. Otherwise, use the crashing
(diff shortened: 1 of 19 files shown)


socorro/signature/utils.py
```diff
@@ -22,22 +22,15 @@ def get_crashing_thread(crash_data):
...
-    If it's a chrome hang (hang_type=1), then use thread 0. Otherwise, use the crashing
-    thread specified in the crash data.
-
     :arg crash_data: crash data structure that conforms to the schema
 
     :returns: crashing thread as an int
 
     """
-    crashing_thread = 0
-    if crash_data.get("hang_type", None) != 1:
```
willkg, line 34: hang_type always ends up as 0. Thus this is always true and we can reduce it accordingly.

PR #6068, opened 2022-04-11
bug 1761240: fix major/minor version figuring

willkg:
The OSPrettyVersionRule extracts a major and minor version from the os
version. This improves that extraction code so it's more resilient to
interesting possibilities.

### DIFF
--- a/socorro/processor/rules/mozilla.py
+++ b/socorro/processor/rules/mozilla.py
@@ -988,6 +988,8 @@ class OSPrettyVersionRule(Rule):
 
     """
 
+    MAJOR_MINOR_RE = re.compile(r"^(\d+)\.(\d+)")
+
     WINDOWS_VERSIONS = {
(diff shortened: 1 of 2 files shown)


socorro/processor/rules/mozilla.py
```diff
@@ -1018,18 +1020,15 @@ def action(self, raw_crash, dumps, processed_crash, processor_meta):
...
+        if match is None:
+            # The version number is missing or invalid, there's nothing more to do
             return True
 
-        version_split = os_version.split(".")
-        if len(version_split) < 2:
-            # The version number is invalid, there's nothing more to do.
-            return
+        major_version = int(match.group(1))
+        minor_version = int(match.group(2))
```
willkg, line 1029: At this point, the text matches `(\d+)` so it's guaranteed to be a valid int.

PR #6076, opened 2022-05-02
Updates 20220502

willkg:
Update dependencies. This covers:

* Update secondary requirements affected by requiremnts.in
* Bump moment from 2.29.1 to 2.29.2 in /webapp-django (from PR #6066)
* Bump bcrypt from 3.2.0 to 3.2.1 (from PR #6071)
* Bump cryptography from 36.0.1 to 37.0.1 (from PR #6074)
* Bump pygments from 2.11.2 to 2.12.0 (from PR #6075)


willkg:
I removed the django-waffle update because we're not using it and updating it sometimes means we have to do migrations and I don't feel like doing that.

I removed websocket-client because we don't want to update the requirements for docker-compose which we use in CI.

PR #6105, opened 2022-06-28
bug 1776868: implement MissingProcessedCrash API

willkg:


webapp-django/crashstats/api/views.py
```diff
@@ -575,6 +577,89 @@ def finalize_response(self, request, response, *args, **kwargs):
...
+    def get(self, request):
+        qs = models.MissingProcessedCrash.objects.order_by("id").all()
+        paginator = Paginator(qs, self.PAGE_SIZE)
+
+        try:
+            page_number = int(request.GET.get("page", "1"))
+        except ValueError:
+            return http.JsonResponse(
+                {"error": "invalid page number"}, status=status.HTTP_400_BAD_REQUEST
+            )
```
willkg, line 624: We have a handful of ways to return responses and it's kind of weird. I think I looked at this a while back and decided to not do anything because we also have multiple "api frameworks" involved and they're pretty different.

Still, this doesn't thrill me.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -210,7 +210,10 @@ def collected_date(self):
         return "20" + self.crash_id[-6:]
 
     def report_url(self):
-        return reverse("crashstats:report_index", args=(self.crash_id,))
+        try:
+            return reverse("crashstats:report_index", args=(self.crash_id,))
+        except NoReverseMatch:
+            return f"no reverse match: {self.crash_id}"
```
willkg, line 216: Adding this lets me be stupid and create MissingProcessCrash records for crash ids that are invalid.

PR #6106, opened 2022-06-29
bug 1776923: add Thunderbird version rewrite support

willkg:
This fixes the BetaVersionRule which was written for a time when we
didn't have normalized, validated data. Thank goodness those days are
over-ish.

This adds Thunderbird to archivescraper scraping so we have data in the
table. This also fixes the BetaVersionRule to fix versions for
Thunderbird betas, too.

### DIFF
--- a/socorro/processor/rules/mozilla.py
+++ b/socorro/processor/rules/mozilla.py
@@ -873,18 +873,19 @@ class ModulesInStackRule(Rule):
 
 
 class BetaVersionRule(Rule):
-    #: Hold at most this many items in cache; items are a key and a value
-    #: both of which are short strings, so this doesn't take much memory
+    # Hold at most this many items in cache; items are a key and a value
+    # both of which are short strings, so this doesn't take much memory
     CACHE_MAX_SIZE = 5000
 
-    #: Items in cache expire after 30 minutes by default
+    # Items in cache expire after 30 minutes by default
     SHORT_CACHE_TTL = 60 * 30
 
-    #: If we know it's good, cache it for 24 hours because it won't change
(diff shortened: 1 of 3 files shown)


socorro/processor/rules/mozilla.py
```diff
@@ -953,38 +955,32 @@ def predicate(self, raw_crash, dumps, processed_crash, proc_meta):
...
         build_id = processed_crash.get("build", "").strip()
-        release_channel = processed_crash.get("release_channel").strip()
+        release_channel = processed_crash["release_channel"].lower()
 
         # Only run if we've got all the things we need
-        if (
-            product
-            and build_id
-            and release_channel
-            and product in self.SUPPORTED_PRODUCTS
```
willkg, line 976: We don't need this check because the predicate wouldn't pass if this weren't true and then this would never run.

webapp-django/crashstats/crashstats/management/commands/archivescraper.py
```diff
@@ -254,7 +258,11 @@ def scrape_candidate_version(self, link, product_name, final_releases):
             # directories
             for json_link in json_links:
                 json_file = self.download(json_link)
-                data = json.loads(json_file)
+                try:
+                    data = json.loads(json_file)
+                except json.decoder.JSONDecodeError:
+                    self.worker_write(f"not valid json: {json_link}")
+                    continue
```
willkg, line 265: I occasionally get errors where the returned content isn't JSON. We should probably fix something to retry this situation a few times in case it's an ephemeral server error. I'm going to leave it like this for now.

PR #6121, opened 2022-07-21
bug 1764570: implement sentry pii scrubbing

willkg:
This adds Sentry set up, PII scrubbing, and integration tests using the Fillmore library. It also cleans up a lot of Socorro code to make this easier to share between the processor and webapp/crontabber.

* centralize revision bits into libdockerflow.py and add tests
* clean up processor rules so we're passing by keyword arguments only
* rework all the sentry related code in the processor and webapp/crontabber to use Fillmore and libsentry
* add integration tests so we know what's getting scrubbed and whether we need to look at event contents because something has changed
* clean up a bunch of tests that were mocking out sentry and logging so they don't do that anymore (the mocking was wrong in some cases anyhow)

This adds a couple of new metrics:

* `webapp.crashstats.apps.sentry_scrub_error`
* `processor.sentry_scrub_error`

We need to add them to the dashboards when this lands.


socorro/app/socorro_app.py
```diff
@@ -145,8 +146,37 @@ def _app_instance_name(self):
...
+                Boto3Integration(),
+                ExcepthookIntegration(),
+                DedupeIntegration(),
+                StdlibIntegration(),
+                ModulesIntegration(),
+                ThreadingIntegration(),
+            ],
+            # Scrub sensitive data
+            before_send=scrubber,
+        )
```
willkg, line 174: App gets subclassed into the actual apps. To reduce duplicate code and the potential to fix things in one place but miss others, we should nix this and make it required to implement on an app-by-app basis.

socorro/unittest/processor/test_processor_app.py
```diff
@@ -139,180 +143,233 @@ def test_transform_unexpected_exception(self):
...
-    @mock.patch("socorro.lib.sentry_client.get_hub")
-    def test_transform_save_error_with_sentry_configured_successful(
-        self, mock_get_hub, caplogpp
-    ):
-        caplogpp.set_level("DEBUG")
 
-        mock_hub = mock.MagicMock()
-        mock_hub.capture_exception.side_effect = RuntimeError("should not be called")
-        mock_get_hub.return_value = mock_hub
+SAVE_ERROR = {}
```
willkg, line 345: We can remove this.

webapp-django/crashstats/crashstats/apps.py
```diff
@@ -2,10 +2,82 @@
...
+]
+
+
+def count_sentry_scrub_error(msg):
+    metrics.incr("sentry_scrub_error", 1)
+
+
+def configure_sentry():
+    release = get_release_name(settings.SOCORRO_ROOT)
+    scrubber = Scrubber(rules=SCRUB_RULES_DEFAULT + SCRUB_RULES_WEBAPP)
```
willkg, line 56: This needs the error handler.

PR #6126, opened 2022-07-26
bug 1763264: finish super search migration

willkg:
This finishes up the super search field migration so that the
Elasticsearch crash storage is only looking at the processed crash for
data it's indexing.


socorro/external/es/super_search_fields.py
```diff
@@ -3480,13 +3011,8 @@ def number_field(
         "is_exposed": True,
```
willkg, line 3011: Oops--missed lowercasing the `in_database_name` here.

PR #6129, opened 2022-07-29
bug 1782258: expose thread id and name in search

aosmond:
The thread id and name are useful search parameters to help distinguish
between similar signatures, and finding patterns across crash reports in
aggregations.


socorro/processor/rules/mozilla.py
```diff
@@ -1152,6 +1152,12 @@ def action(self, raw_crash, dumps, processed_crash, processor_meta_data):
         processor_meta_data["processor_notes"].extend(result.notes)
         if "proto_signature" in result.extra:
             processed_crash["proto_signature"] = result.extra["proto_signature"]
+        if "crashing_thread_id" in result.extra:
+            processed_crash["crashing_thread_id"] = result.extra["crashing_thread_id"]
+        if "crashing_thread_name" in result.extra:
+            processed_crash["crashing_thread_name"] = result.extra[
+                "crashing_thread_name"
+            ]
```
willkg, line 1160: These aren't related to signature generation, so they shouldn't be here.

Extracting the `crashing_thread_id` and `crashing_thread_name` fields from the stackwalker output and adding them to the processed crash should happen in the `CrashingThreadInfoRule`:

https://github.com/mozilla-services/socorro/blob/6e7aadc03a2ba5315a978fac433285dde8758af3/socorro/processor/rules/breakpad.py#L24-L53

socorro/signature/rules.py
```diff
@@ -525,9 +525,12 @@ def action(self, crash_data, result):
...
             # Otherwise we don't have frames to use.
             if crashing_thread is not None:
+                crashing_thread_mapping = glom(
+                    crash_data, "threads.%d" % crashing_thread, default={}
+                )
+                crashing_thread_name = crashing_thread_mapping.get("thread_name", None)
                 signature_list = self._create_frame_list(
-                    glom(crash_data, "threads.%d" % crashing_thread, default={}),
-                    crash_data.get("os") == "Windows NT",
+                    crashing_thread_mapping, crash_data.get("os") == "Windows NT"
```
willkg, line 533: I don't understand this. You're not adding the `crashing_thread_name` or `crashing_thread_id` to the signature, so why make this change?

socorro/signature/rules.py
```diff
@@ -554,6 +557,10 @@ def action(self, crash_data, result):
             result.debug(self.name, note)
         if signature:
             result.set_signature(self.name, signature)
+        if crashing_thread is not None:
+            result.extra["crashing_thread_id"] = crashing_thread
+        if crashing_thread_name:
+            result.extra["crashing_thread_name"] = crashing_thread_name
```
willkg, line 563: Same here--you're not adding the `crashing_thread_name` or `crashing_thread_id` to the signature, so we don't need to make this change.

socorro/signature/tests/test_rules.py
```diff
@@ -639,6 +639,7 @@ def test_replace_address_trailing_whitespace(self):
         },
     ],
     "threads_index": 0,
+    "thread_name": "MainThread",
```
willkg, line 642: Same here--since we're not adjusting signature generation, we don't need to make any changes to this file.

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -864,6 +865,8 @@ def test_basic(self, client):
             "results": [
                 {
                     "extra": {
+                        "crashing_thread_id": 0,
+                        "crashing_thread_name": "SomeThread",
```
willkg, line 869: Same here. We're not making changes to signature generation, so we don't need to change this file.

socorro/external/es/super_search_fields.py
```diff
@@ -1448,6 +1448,34 @@ def number_field(
...
+        "in_database_name": "crashing_thread_id",
+        "is_exposed": True,
+        "is_returned": True,
+        "name": "crashing_thread_id",
+        "namespace": "processed_crash",
+        "permissions_needed": [],
+        "query_type": "number",
+        "storage_mapping": {"type": "integer"},
+    },
+    "crashing_thread_name": {
```
willkg, line 1465: I think this should probably be implemented as a keyword--not a string. Strings get analyzed when indexed, but keywords don't. I think you should use `keyword_field`. Here's an example:

https://github.com/mozilla-services/socorro/blob/6e7aadc03a2ba5315a978fac433285dde8758af3/socorro/external/es/super_search_fields.py#L1468-L1475

Having said that, I don't exactly have a list of thread name example values in front of me, so maybe keyword isn't a good fit. We'd have to test it out.

socorro/external/es/super_search_fields.py
```diff
@@ -1448,6 +1448,34 @@ def number_field(
         "query_type": "enum",
         "storage_mapping": {"type": "string"},
     },
+    "crashing_thread_id": {
```
willkg, line 1451: Use `number_field` because it sets up the defaults. Here's an example:

https://github.com/mozilla-services/socorro/blob/6e7aadc03a2ba5315a978fac433285dde8758af3/socorro/external/es/super_search_fields.py#L1895-L1903

socorro/external/es/super_search_fields.py
```diff
@@ -1448,6 +1448,34 @@ def number_field(
...
+        "data_validation_type": "int",
+        "description": "",
```
willkg, line 1453: Fields should have descriptions.

willkg (changes requested):
When you implement the value extraction in `CrashingThreadInfoRule`, you'll need to adjust the tests to make sure it's correct:

https://github.com/mozilla-services/socorro/blob/6e7aadc03a2ba5315a978fac433285dde8758af3/socorro/unittest/processor/rules/test_breakpad.py#L165-L190

Let me know if any of this is confusing. I'm happy to take over if that's more helpful / easier.

willkg:
I meant to mention that I really appreciate you tackling this--thank you!

willkg:
I redid how fields work in the processed crash, so I figured I'd reimplement this. I did that in PR #6170 .

PR #6139, opened 2022-08-04
bug 1756264: correct graph legend

edugfilho:
Because Javascript doesn't guarantee the reading order will be the same as the insertion order for objects, the major version order for legend (array) and graph data (object) would not match sometimes. This was fixed by building the legend array along with lineDataArray, which contains graph data and is an array, which guarantees the reading order will be the same as insertion order.


edugfilho:
This shows the green line now matches the legend for the same major version.
Screen Shot 2022-08-04 at 2 35 11 PM

willkg (approved):
Nice job!

PR #6142, opened 2022-08-12
bug 1764395: processed crash schema

willkg:
This fixes a bunch of things around the code I needed to write. Then it implements a processed crash schema and reworks the `CopyFromRawCrashRule` to use the processed crash schema to guide what to copy over.

Things I did to verify this:

1. Wrote a tool for comparing two JSON documents.
2. Processed 1,000 Firefox, 1,000 Fenix, and 100 Thunderbird crash reports and compared the processed crash output from prod to the processed crash output from the local dev environment.
3. Verified that the new behavior for app_notes, url, and user_comments works in the report and signature views of the webapp.

I'm pretty confident this is correct.

The next step is to write permission-based reducers and then redo the processed crash APIs (apparently there are multiple) and the webapp to use the schema and remove the other allow lists floating around.


socorro/processor/rules/general.py
```diff
@@ -131,23 +131,6 @@ def action(self, raw_crash, dumps, processed_crash, processor_meta_data):
...
-
-    """
-
-    KEYS = [
-        "collector_notes",
-    ]
-
-    def action(self, raw_crash, dumps, processed_crash, processor_meta_data):
-        for key in self.KEYS:
-            processed_crash[key] = raw_crash.get(key, None)
```
willkg, line 148: This was never added to the pipeline, so I removed the code. However, we do need to add it.

Since the `collector_notes` value is an array of strings and I don't currently have support for that in `CopyFromRawCrashRule`, I couldn't just add this in this pass. I'll write up a separate bug for it.

willkg: https://bugzilla.mozilla.org/show_bug.cgi?id=1784558

socorro/processor/rules/mozilla.py
```diff
@@ -690,6 +653,8 @@ def action(self, raw_crash, dumps, processed_crash, processor_meta_data):
 class ExploitablityRule(Rule):
     """Copies exploitability bits in json_dump to processed_crash"""
 
+    # FIXME(willkg): We no longer have this field, so we can nix this
```
willkg, line 656: This is covered in https://bugzilla.mozilla.org/show_bug.cgi?id=1784485

socorro/schemas/__init__.py
```diff
@@ -4,20 +4,22 @@
...
-TELEMETRY_SOCORRO_CRASH_SCHEMA_AS_STRING = _get_file_content(
-    "telemetry_socorro_crash.json", parsed=False
-)
 
-JAVA_EXCEPTION_SCHEMA = _get_file_content("java_exception.json")
+PROCESSED_CRASH_SCHEMA = get_file_content("processed_crash.1.schema.yaml")
+
+TELEMETRY_SOCORRO_CRASH_SCHEMA = get_file_content("telemetry_socorro_crash.json")
+
+JAVA_EXCEPTION_SCHEMA = get_file_content("java_exception.json")
```
willkg, line 25: This is now a subschema in the processed crash schema. At some point, I'll fix this so we only have the java exception stuff defined once.

socorro/schemas/validate_processed_crash.py
```diff
@@ -0,0 +1,127 @@
...
+    # keys_not_in_crashes = set(schema_types.keys()) - all_keys
+    # if keys_not_in_crashes:
+    #     click.echo("")
+    #     click.echo(
+    #         f"{len(keys_not_in_crashes)} keys in processed_crash.json Schema, "
+    #         + "but not in crash reports:"
+    #     )
+    #     click.echo("  %s%s" % ("KEY".ljust(60), "TYPE(S)"))
+    #     for key in sorted(keys_not_in_crashes):
+    #         click.echo("  %s%s" % (key.ljust(60), all_schema_types[key]))
```
willkg, line 113: I wrote a schema visitor, so this might be easier to do now. I'll look into that.

socorro/schemas/validate_telemetry_socorro_crash.py
```diff
@@ -119,6 +119,7 @@ def log_all_keys(crash, prefix=""):
 
     all_schema_types = {}
 
+    # FIXME(willkg): This doesn't traverse refs
```
willkg, line 122: I can redo this with the schema visitor.

PR #6143, opened 2022-08-15
bug 1784927: remove elasticsearch redactor code

willkg:
We no longer need the elasticsearch redactor code, so it's time to
remove it and simplify things.

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -10,13 +10,13 @@ import re
 import time
 
 from configman import Namespace
-from configman.converters import class_converter, list_converter
+from configman.converters import class_converter
 import elasticsearch
 from elasticsearch.exceptions import NotFoundError
 import glom
 import markus
 
-from socorro.external.crashstorage_base import CrashStorageBase, Redactor
+from socorro.external.crashstorage_base import CrashStorageBase
 from socorro.external.es.super_search_fields import (
     FIELDS,
     get_destination_keys,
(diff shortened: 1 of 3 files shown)


socorro/external/es/crashstorage.py
```diff
@@ -34,24 +34,6 @@
...
-
-    Not that this overwrites the list of forbidden_keys that would be defined
-    through configuration. That list is hard-coded in the __init__ function.
-    """
-
-    def __init__(self, config):
-        super().__init__(config)
-
-        # Overwrite the list of fields to redact away.
-        self.forbidden_keys = ["StackTraces"]
```
willkg, line 52: We haven't had this key since before I started working on Socorro. So this class (as far as I can tell) has never had an effect on things.

socorro/external/es/crashstorage.py
```diff
@@ -522,92 +504,3 @@ def _submit_crash_to_elasticsearch(
...
-    )
-    required_config.es_redactor.add_option(
-        name="forbidden_keys",
-        doc="a list of keys not allowed in a redacted processed crash",
-        default=(
-            "json_dump, "
-            "upload_file_minidump_flash1.json_dump, "
-            "upload_file_minidump_flash2.json_dump, "
-            "upload_file_minidump_browser.json_dump"
-        ),
```
willkg, line 544: This did have an effect on things, however, recent overhauls of how indexing works have pared down indexing to _only_ index fields listed in super search fields and these four are not in that list. Given that, the indexing already pares down the document to only what's needed for supersearch and now this redactor isn't having an effect on things which is why we can remove it.

socorro/external/es/crashstorage.py
```diff
@@ -522,92 +504,3 @@ def _submit_crash_to_elasticsearch(
...
-
-    required_config = Namespace()
-    required_config.add_option(
-        name="json_dump_allowlist_keys",
-        doc="keys of the json_dump field to keep in the processed crash",
-        default=[
-            "stackwalk_version",
-            "system_info",
-        ],
-        from_string_converter=list_converter,
```
willkg, line 583: This bit was to re-add things that were removed in the superclass. Given we don't need the superclass anymore, we also don't need this class to undo some of the things the superclass did.

PR #6144, opened 2022-08-15
bug 1697001: remove UnredactedCrash API

willkg:
This merges the UnredactedCrash API with the ProcessedCrash API. They
almost did the same thing in some key cases, but in different ways.
Removing UnredactedCrash API simplifies the APIs by 1 and continues to
pave the way for a new data permissions model.


socorro/external/crashstorage_base.py
```diff
@@ -97,50 +97,6 @@ def as_memory_dumps_mapping(self):
         return in_memory_dumps
 
 
-class Redactor(RequiredConfig):
```
willkg, line 100: We're no longer redacting, so we can drop this class which isn't really doing the right thing anymore anyhow.

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -205,180 +204,145 @@ def test_hit_or_not_hit_ratelimit(self):
...
+            "signature": "FakeSignature1",
+            "success": True,
+            "uptime": 14693,
+            "uuid": "11cb72f5-eb28-41e1-a8e4-849982120611",
+            "version": "5.0a1",
+        }
+
+        protected_data = {
+            "url": "https://example.com",
+            "user_comments": None,
```
willkg, line 232: This structure is _so_ much clearer about what's going on and what's correct and not correct.

webapp-django/crashstats/api/tests/test_views.py
```diff
@@ -205,180 +204,145 @@ def test_hit_or_not_hit_ratelimit(self):
...
-        # because we don't have the sufficient permissions yet to use it
-        assert response.status_code == 403
+        for key in protected_data.keys():
+            assert key not in data
 
+        # If you have permissions, you see all the data
+        # If you do have permissions, then you get it all
         user = User.objects.create(username="test")
         self._add_permission(user, "view_pii")
-        self._add_permission(user, "view_exploitability")
```
willkg, line 267: There's another bug for getting rid of all the exploitability stuff. I removed these bits in this PR because the ProcessedCrash API uses an allow list which is triggered solely by view_pii and we have no exploitability data anymore. I'll finish cleaning things up in the "remove exploitability" bug (https://bugzilla.mozilla.org/show_bug.cgi?id=1784485)

webapp-django/crashstats/crashstats/models.py
```diff
@@ -802,15 +767,12 @@ class RawCrash(SocorroMiddleware):
...
         "BIOS_Manufacturer",
-        "bug836263-size",
-        "buildid",
         "BuildID",
         "CrashTime",
         "CoUnmarshalInterfaceResult",
         "DOMIPCEnabled",
         "EMCheckCompatibility",
         "FlashVersion",
-        "id",
```
willkg, line 813: I think this used to be ProductID a long time ago or something like that. It's not sent anymore.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -802,15 +767,12 @@ class RawCrash(SocorroMiddleware):
         "AvailablePhysicalMemory",
         "AvailableVirtualMemory",
         "BIOS_Manufacturer",
-        "bug836263-size",
-        "buildid",
```
willkg, line 806: These are no longer sent anymore.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -777,14 +741,15 @@ class RawCrash(SocorroMiddleware):
...
+        # Crash annotations:
+        # https://searchfox.org/mozilla-central/source/toolkit/crashreporter/CrashAnnotations.yaml
         "Accessibility",
         "AdapterDeviceID",
         "AdapterDriverVersion",
         "AdapterRendererIDs",
         "AdapterSubsysID",
         "AdapterVendorID",
         "ApplicationBuildID",
-        "additional_minidumps",
```
willkg, line 787: Oops--this is still being sent. I'll re-add it.

willkg:
Before this gets deployed to production, we should send out an email about how we're removing the UnredactedCrash API and people should switch to using ProcessedCrash API.

PR #6146, opened 2022-08-16
bug 1764395: redo ProcessedCrash permissions handling

willkg:
This redoes the ProcessedCrash API to reduce the processed crash data using the processed crash schema and the permissions of the user viewing the data.

This reduces the number of places we have annotations for field permissions by one.

This also fixes some tests that had bad test data that didn't match the schema.

To test:

1. process some crashes
2. view the crashes both logged in with view_pii permissions as well as without

Particular points of interest:

1. whether the `java_exception` or `java_exception_raw` show (we can probably remove this at some point)
2. whether the `moz_crash_reason` or `moz_crash_reason_raw` show
3. whether the `java_stacktrace` or `java_stacktrace_raw` show
4. whether user comments show up or not
5. whether the report view kicks up an error because the underlying data has changed


socorro/external/boto/crashstorage.py
```diff
@@ -290,7 +290,12 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         # Create a reducer that traverses documents and reduces them down to the
         # structure of the specified schema
-        self.reducer = Reducer(schema=TELEMETRY_SOCORRO_CRASH_SCHEMA)
+        only_public = permissions_predicate(["public"])
```
willkg, line 293: This would be clearer if we added the argument name.

socorro/schemas/socorro_metadata.1.schema.yaml
```diff
@@ -12,3 +12,34 @@ properties:
...
+      The permissions required to view this field.
+
+      This field if it does not exist or has no value defaults to 'all
+      permissions are required'.
+    items:
+      enum: ["public", "protected", "exploitability"]
+      type: string
+    type: array
+
+  bugs:
```
willkg, line 27: We're not using bugs, dataReview, or notes, yet.

Also, the Glean metrics.yaml files use `data_review` (snake case) instead of `dataReview` (camel-case). I figure it's early days, so we can switch if we need to, but keeping with jsonschema conventions seemed less weird.

socorro/unittest/schemas/test_schemas.py
```diff
@@ -25,21 +26,467 @@ def test_validate_schemas(reporoot):
...
+    "utility_process_sandboxing_kind",
+    "uuid",
+    "vendor",
+    "version",
+    "windows_error_reporting",
+    "xpcom_spin_event_loop_stack",
+}
+
+
 def test_processed_crash_schema():
```
willkg, line 445: I should split the processed crash schema tests into a separate module. It's weird to have one module testing all the schema things.

socorro/unittest/schemas/test_schemas.py
```diff
@@ -25,21 +26,467 @@ def test_validate_schemas(reporoot):
...
+                public_fields.add(".".join(split_path(general_path)))
+
+                parts = list(split_path(general_path))
+                for i in range(len(parts)):
+                    parent = ".".join(parts[:i])
+                    if not parent:
+                        continue
+
+                    if parent not in public_fields:
+                        raise NonPublicParentError(
```
willkg, line 481: Hrm... This might not work because `java_exception_raw` is protected, but it's a ref and the subschema has public fields. This should kick up an error.

willkg: Oh, I see what's going on... yeah, this is definitely wrong. I'll fix this.

socorro/unittest/schemas/test_schemas.py
```diff
@@ -25,21 +26,467 @@ def test_validate_schemas(reporoot):
...
+    "java_exception.exception.values.[].stacktrace.frames",
+    "java_exception.exception.values.[].stacktrace.frames.[]",
+    "java_exception.exception.values.[].stacktrace.frames.[].filename",
+    "java_exception.exception.values.[].stacktrace.frames.[].function",
+    "java_exception.exception.values.[].stacktrace.frames.[].in_app",
+    "java_exception.exception.values.[].stacktrace.frames.[].lineno",
+    "java_exception.exception.values.[].stacktrace.frames.[].module",
+    "java_exception.exception.values.[].stacktrace.module",
+    "java_exception.exception.values.[].stacktrace.type",
+    "java_exception_raw",
```
willkg, line 179: This is marked protected, so it (and it's children) shouldn't be here. It's a ref, so I think the bug is around traversing refs.

willkg:
I hit some issues and had to redo a lot of stuff. Going to close this out and do a new PR.

PR #6148, opened 2022-08-22
bug 1764395: rewrite schemas and reducers and traversers

willkg:
This is a semi-big rewrite of a lot of stuff.

First, I changed the plan. Instead of defining a processed crash schema (jsonschema), we now define a socorro data schema (jsonschema) and then define the processed crash schema (socorrodata). The socorro data schema codifies the bits we were calling "socorro metadata" before as part of the schema. It enforces some of the things we were enforcing in validation tests. It is structurally similar to jsonschema (types are the same, object and array traversal is a subset) without some of the complexity we don't need (items as records vs. items as lists). I copied some bits from the glean-parser metrics schema so it's as close
to that as we can make it. Overall, I think this is a big improvement. 

However, it does mean we need two different sets of traversing things: one for jsonschema schemas (telemetry crash schema) and one for socorro-data schemas (processed crash).

We can share some things. I made `"$ref"` in socorro-data the same-ish as jsonschema. The basic types are the same. socorro-data uses `properties`, `pattern_properties`, and `items` in the same way as jsonschema (roughly).

I wrote a `resolve_references` function that takes a schema, resolves all the references, and drops the `"definitions"` section.

For jsonschema things, I wrote a `JsonSchemaReducer` which we use in `TelemetryBotoS3CrashStorage`. It reduces a document according to the shape of the specified schema. It no longer does other transformations or predicates or whatever, so it's simpler. It also doesn't support the full jsonschema spec--only the bits I needed.

For socorro-data things, I wrote a `SocorroDataReducer` that reduces a document according to the shape of a specified schema. It doesn't do any other transformations. I wrote a `transform_socorro_data_schema` function that takes a schema and a transform function and returns a new schema. We'll use that for handling permissions.

I updated `TelemetryBotoS3CrashStorage` to create a public-only derived schema from the processed crash schema. It reduces the processed crash document with that so we're gating what's public and what's not from the processed crash schema. It then reduces the public-only processed crash data using the telemetry schema giving us a document in the right shape for telemetry ingestion.

I rewrote the ProcessedCrash API to create a series of permission-variations derived schemas from the processed crash schema. It uses the appropriate one for reducing the processed crash to the data the user is permitted to see. No more `API_ALLOWLIST` nonsense.

I updated the processed crash schema to be valid socorro-data. I added permission annotations to it with permissions derived from the `API_ALLOWLIST` from the ProcessedCrash API code coupled with the permissions from `super_search_fields.py` because both were involved in non-overlapping ways (which is one of the reasons we're moving permissions to a single central schema

I also added a few fields to the processed crash schema that I think got lost in the shuffle.

This forces `dom_fission_enabled` and `dom_ipc_enabled` to get converted from a boolean to a string because that's what Telemetry ingestion is expecting.

I fixed the processor to add isoformat strings to the processed crash for `client_crash_date`, `started_datetime`, `completed_datetime`, and `submitted_timestamp`. They get saved to crash storage as isoformat strings and when the processed crash is loaded, they stay as isoformat strings. Given that, it's much easier to _always_ have them be isoformat strings.

The memory report can contain urls of pages, so it should never be public. This adds a comment to the processed crash schema and adjusts the description in the telemetry socorro crash schema.

`cpu_count` and `stackwalk_version` were indexed into Elasticsearch from `processed_crash.json_dump`. The `cpu_count` is in the processed crash already, so we can/should pull the value from there like the other `cpu_` fields. The `stackwalk_version` field should be up a level as well and match the `mdsw_` fields.

This moves `stackwalk_version` up a level and adds a note in the `report_index.html` code to drop the backwards-compatible code in 12/2022.

This moves `cpu_count` up a level and adds a migration and a note to `super_search_fields.py` to drop the backwards-compatible code in 12/2022.

We have several risks with this code:

1. Bugs that cause things to fail. I've tested the affected areas pretty well and I think we're ok. In many cases, the new behavior is more predictable and correct than the old behavior. Once this lands, I'll continue to test on stage. After that point, whatever bugs are left are things we can fix going forward--they shouldn't require rethinking everything.
2. Data leaks. I've taken care to test without api tokens and make sure the output of things matches prod. In several cases, the new code is more restrictive than prod and some data is no longer emitted by the processed crash api. While this PR is a huge change and it's possible I missed something, I'm feeling pretty good about the due diligence done.

What I did to validate this:

1. Process a bunch of Firefox, Fenix, and Thunderbird crash reports and compare the telemetry crash report from before and after making these changes.
2. Process a bunch of Firefox, Fenix, and Thunderbird crash reports and compare the output of the ProcessedCrash API with (public and protected data) and without (public data only) an auth token between prod and local dev environment.
3. Write a bunch of unit tests to make sure graph traversal code works correctly in normal and edge case scenarios.

Next steps:

1. I'll review this code and make any required changes.
2. We'll land this and test on stage.


socorro/external/fs/crashstorage.py
```diff
@@ -17,7 +17,7 @@
     FileDumpsMapping,
     MemoryDumpsMapping,
 )
-from socorro.lib.libdatetime import utc_now, JsonDTISOEncoder
```
willkg, line 20: We had two json encoders that had different things they did with datetimes. That's silly. We only need one and we should have everything unify on that.

This removes the second one.

socorro/lib/libjson.py
```diff
@@ -62,25 +63,230 @@ def lookup_definition(schema, ref):
...
+    #
+    # Deepcopy because this schema can end up in multiple parts of the tree.
+    new_schema = copy.deepcopy(lookup_definition(root_schema, ref))
+
+    # Copy anything from the original schema (other than "$ref") into the
+    # new schema
+    for key, val in schema.items():
+        if key == "$ref":
+            continue
+        new_schema[key] = copy.deepcopy(schema[key])
```
willkg, line 78: I should redo this as a list comprehension.

socorro/lib/libjson.py
```diff
@@ -62,25 +63,230 @@ def lookup_definition(schema, ref):
     return schema
 
 
-def expand_references(schema, schema_item):
-    """Expands $ref items
+def resolve_reference(root_schema, schema, ref):
```
willkg, line 66: This should get a docstring.

socorro/lib/libjson.py
```diff
@@ -62,25 +63,230 @@ def lookup_definition(schema, ref):
...
+            continue
+        new_schema[key] = copy.deepcopy(schema[key])
+
+    return new_schema
+
+
+def resolve_references(schema):
+    """Returns a schema with all ``$ref`` resolved
+
+    This works for both jsonschema schemas and socorro-data schemas.
```
willkg, line 86: This should state why this works for both.

socorro/lib/libjson.py
```diff
@@ -62,25 +63,230 @@ def lookup_definition(schema, ref):
...
+                    for key, val in schema["properties"].items()
+                }
+            if schema.get("pattern_properties"):
+                schema["pattern_properties"] = {
+                    key: _resolve(root_schema, val)
+                    for key, val in schema["pattern_properties"].items()
+                }
+
+        elif "array" in type_:
+            schema["items"] = _resolve(root_schema, schema["items"])
```
willkg, line 159: jsonschema supports record items which we don't use, so this doesn't handle that.

socorro/lib/libjson.py
```diff
@@ -62,25 +63,230 @@ def lookup_definition(schema, ref):
...
+            )
+
+        type_ = schema["type"]
+        if "object" in type_:
+            if schema.get("properties"):
+                schema["properties"] = {
+                    key: _resolve(root_schema, val)
+                    for key, val in schema["properties"].items()
+                }
+            if schema.get("pattern_properties"):
```
willkg, line 152: The docstring says this works for jsonschema and socorrodata schema, but it doesn't because jsonschema has `patternProperties` and socorrodata schema has `pattern_properties`.

socorro/lib/libjsonschema.py
```diff
@@ -100,39 +188,29 @@ def convert_to(document_part, target_format):
...
+       are invalid.
 
     """
-    if pattern not in PATTERN_CACHE:
-        PATTERN_CACHE[pattern] = re.compile(pattern)
-    return PATTERN_CACHE[pattern]
-
 
-class Reducer:
-    def __init__(self, schema, include_predicate=everything_predicate):
```
willkg, line 123: This kind of reducer was interesting, but in practice it was difficult to use because it was doing too many things at the same time. In the "new plan", I broke this up into three things:

1. resolve references producing a resolved schema
2. transform a schema into another schema using some transform function
3. reduce a document using a schema

That's much easier to work with and reason about and made the reducer code more straight-forward.

socorro/lib/libsocorrodataschema.py
```diff
@@ -0,0 +1,455 @@
...
+            )
+
+        elif isinstance(document_part, list):
+            if "array" not in schema_part_types:
+                raise InvalidDocumentError(
+                    f"invalid: {path}: type array not in {schema_part_types}"
+                )
+
+            schema_item = schema_part["items"]
+            schema_item_path = f"{path}.[]"
```
willkg, line 406: This is superfluous.

socorro/lib/libsocorrodataschema.py
```diff
@@ -0,0 +1,455 @@
...
+
+                path_name = f"{path}.{name}"
+
+                new_doc[name] = self._traverse(
+                    schema_part=schema_property,
+                    document_part=document_part[name],
+                    path=path_name,
+                )
+
+            # Verify all required properties exist in the document
```
willkg, line 442: We don't have a "required" thing in socorrodata schema.

socorro/schemas/socorro-data-1-0-0.schema.yaml
```diff
@@ -0,0 +1,154 @@
...
+      - "boolean"
+      - "integer"
+      - "null"
+      - "number"
+      - "object"
+      - "string"
+
+  medium_id:
+    allOf:
+      - maxLength: 40
```
willkg, line 25: glean-parser's metrics schema enforces snake_case as well, but we can't do that because we have some fields we need to support that aren't snake_case. Similarly, we have longer keys to support, so we use 40 instead of 30.

socorro/unittest/processor/rules/test_breakpad.py
```diff
@@ -459,6 +459,7 @@ def test_stackwalker_bad_output(self, tmp_path):
         assert processed_crash["mdsw_stderr"] == "boo hiss"
         assert processed_crash["stackwalk_version"] == rule.stackwalk_version
         assert not processed_crash["success"]
+        assert processed_crash["stackwalk_version"] == rule.stackwalk_version
```
willkg, line 462: This is a duplicate line.

socorro/unittest/lib/test_libjsonschema.py
```diff
@@ -0,0 +1,438 @@
...
+                        },
+                        "permissions": ["ignored"],
+                        "type": "object",
+                    },
+                },
+                "properties": {
+                    "color": {
+                        "description": "the color used here",
+                        "$ref": "#/definitions/color",
+                        "permissions": ["protected"],
```
willkg, line 84: jsonschema doesn't have "permissions" keys and we don't use it that way, so we should nix the permissions bits here.

socorro/unittest/lib/test_libsocorrodataschema.py
```diff
@@ -0,0 +1,740 @@
...
+
+    def test_object_pattern_properties(self):
+        schema = {
+            "$schema": "moz://mozilla.org/schemas/socorro/socorro-data/1-0-0",
+            "type": "object",
+            "properties": {
+                "registers": {
+                    "type": "object",
+                    "pattern_properties": {
+                        "^r.*$": {"type": "string"},
```
willkg, line 311: Should add a comment here because the `r` is easy to miss. Plus it should be a raw string.

socorro/unittest/schemas/test_socorro_data_schemas.py
```diff
@@ -0,0 +1,436 @@
...
+    transform_schema,
+)
+from socorro.schemas import get_file_content
+from socorro.schemas.validate_processed_crash import validate_and_test
+
+
+def test_validate_socorro_data_schema():
+    """Assert that socorro-data-1-0-0.schema.yaml is valid jsonschema"""
+    schema = get_file_content("socorro-data-1-0-0.schema.yaml")
+    jsonschema.Draft7Validator.check_schema(schema)
```
willkg, line 21: `socorro-data-1-0-0.schema.yaml` is a Draft 7 schema because I needed format or something. I forget what it was.

socorro/unittest/schemas/test_socorro_data_schemas.py
```diff
@@ -0,0 +1,436 @@
...
+    """Test whether the module loads and spits out help."""
+    runner = CliRunner()
+    result = runner.invoke(validate_and_test, ["--help"])
+    assert result.exit_code == 0
+
+
+def split_path(path):
+    """Split a general path into parts
+
+    This handles the case where patternProperties parts are enclosed in parens and can
```
willkg, line 34: This should be `pattern_properties`.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -609,93 +630,72 @@ class TelemetryCrash(SocorroMiddleware):
     aliases = {"crash_id": "uuid"}
 
 
+class PermissionsReducerError(Exception):
+    pass
+
+
+@functools.lru_cache
+def get_processed_crash_permissions_reducer(permissions_tuple):
```
willkg, line 638: The variations of permissions_tuple values is pretty minimal, so this will do some work up front and then keep that cached for the duration of the process.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -609,93 +630,72 @@ class TelemetryCrash(SocorroMiddleware):
...
+            permissions_transform_function(permissions_have=permissions_tuple)
+        ),
+    )
+
+    return SocorroDataReducer(schema=permissioned_schema)
+
+
 class ProcessedCrash(SocorroMiddleware):
+    # FIXME(willkg): change this to BotoS3CrashStorage
     implementation = SimplifiedCrashData
```
willkg, line 661: At some point, I want to get rid of `SimplifiedCrashData` which seems like it just adds another abstraction layer without a lot of value.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -609,93 +630,72 @@ class TelemetryCrash(SocorroMiddleware):
...
-        "success",
-        "submitted_from",
-        "topmost_filenames",
-        "upload_file_minidump_*",
-        "uptime",
-        "uuid",
-        "version",
-        "windows_error_reporting",
-    )
+    API_ALLOWLIST = None
```
willkg, line 681: Now field permissions are driven by the processed crash schema.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -609,93 +630,72 @@ class TelemetryCrash(SocorroMiddleware):
...
+
+    def get(self, crash_id, dont_cache=False, refresh_cache=False):
+        data = self.fetch(
```
willkg, line 684: Using this takes advantage of caching and all that.

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -34,37 +34,38 @@
...
+    "reason": "EXC_BAD_ACCESS / KERN_INVALID_ADDRESS",
+    "release_channel": "nightly",
+    "ReleaseChannel": "nightly",
+    "signature": "FakeSignature1",
+    "success": True,
+    "uptime": 14693,
+    "uuid": "11cb72f5-eb28-41e1-a8e4-849982120611",
+    "version": "5.0a1",
+    # protected data
+    "user_comments": "this is a comment",
```
willkg, line 68: I went through and fixed the test data so it's correct, matches the schema, doesn't have weird things in it, and it's easier to eyeball in regards to public and protected.

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -1401,40 +1399,6 @@ def mocked_processed_crash_get(**params):
             if pyquery.PyQuery(row).find("th").text() == "Install Time":
                 assert pyquery.PyQuery(row).find("td").text() == ""
 
-    def test_report_index_empty_os_name(self):
```
willkg, line 1404: This scenario (os_name is empty) isn't possible anymore, so I nixed the test.

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -1578,38 +1540,6 @@ def mocked_publish(**params):
         assert "Please wait..." in content
         assert "Processing this crash report only takes a few seconds" in content
 
-    def test_report_index_with_invalid_date_processed(self):
```
willkg, line 1581: I didn't understand what this was testing, so I nixed it. `date_processed` is derived from `submitted_timestamp` which is set by the collector so... why wouldn't it be valid?

PR #6150, opened 2022-08-23
bug 1764395: remove java_exception_raw

willkg:
Previously, we couldn't handle permissions being different in fields in
nested structures so we did this thing where we'd have a "raw" and
"redacted" copies of data structures in the processed crash.

We now have the ability to set the permissions for fields in nested
structures and prevent them from showing up for people without those
permissions.

Given that, this removes java_exception_raw and all the code to support
that.


webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -650,17 +650,14 @@ <h2>{{ report.product }} {{ report.version }} Crash Report [@ {{ report.signatur
...
-                    </p>
-                  {% endif %}
+                  <h3>
+                    Message: {% if stacktrace.value %}{{ stacktrace.value }}{% else %}--{% endif %}
+                  </h3>
+                  <p>
+                    <img src="{{ static('img/3rdparty/silk/exclamation.png') }}" alt="protected" title="Protected Data" />
+                    Note: Java exception messages are protected data.
+                    <a href="{{ url('documentation:protected_data_access') }}">Protected data policy.</a>
+                  </p>
```
willkg, line 660: This changes the report view so that it's clearer the message is not shown if you don't have access.

PR #6157, opened 2022-08-31
bug 1779630: update rust-minidump to 0.14.0

willkg:
This updates rust-minidump to 0.14.0. It adds the "--no-color" argument
to the command line. It fixes the parsing of the stderr ERROR line to
match the new format.

### DIFF
--- a/socorro/processor/rules/breakpad.py
+++ b/socorro/processor/rules/breakpad.py
@@ -144,9 +144,10 @@ class MinidumpStackwalkRule(Rule):
         command_line=(
             "timeout --signal KILL {kill_timeout} "
             "{command_path} "
-            "--raw-json={raw_crash_path} "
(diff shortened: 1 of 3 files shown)


socorro/processor/rules/breakpad.py
```diff
@@ -144,9 +144,10 @@ def __init__(
         command_line=(
             "timeout --signal KILL {kill_timeout} "
             "{command_path} "
-            "--raw-json={raw_crash_path} "
+            "--evil-json={raw_crash_path} "
             "--symbols-cache={symbol_cache_path} "
             "--symbols-tmp={symbol_tmp_path} "
+            "--no-color "
```
willkg, line 150: It's irksome that we have these command line arguments in two places and kind of have to update both. Took me a while to figure out why the change I made didn't do anything.

PR #6169, opened 2022-09-02
bug 1788971: fix jsonschema deprecation warning

willkg:
bug 1788971: fix jsonschema deprecation warning
    
jsonschema was kicking up this deprecation warning:
    
```
DeprecationWarning: The metaschema specified by $schema was not found.
Using the latest draft to validate, but this will raise an error in the
future.
```
    
This fixes that by passing in the correct validator. Further, since I
think I want to be validating test data because a lot of it is wrong, I
created a function for it that we can use everywhere.

### DIFF
--- a/socorro/lib/libsocorrodataschema.py
+++ b/socorro/lib/libsocorrodataschema.py
@@ -9,6 +9,8 @@ Utilities for working with socorro-data schemas.
 import copy
 import re
 
+import jsonschema
+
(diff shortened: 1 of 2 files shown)


willkg:
Easiest way to test this is to run python with warnings:

```
python -W socorro/schema/validate_processed_crash.py [dir]
```

PR #6171, opened 2022-09-02
bug 1763154: truncate frames from huge stacks

willkg:
When a stack has more than 100 frames, this truncates the middle in the report view so it shows the first 50, a "truncated" line, and then the last 50. This should significantly reduce the HTML involved in stack overflow bugs and make the page load _much_ faster.

Example crash report: 25e8e40f-42a1-4248-b24d-43d2e0220403

I picked 100 because it seemed like a good number and ran it by some others who agreed. However, it's kind of arbitrary. If we decide there are better numbers to truncate on, we can change it in a new bug.


willkg:
Oops--I meant to add a test.

PR #6175, opened 2022-09-04
bug 1788103: show inline functions in details tab of report view

willkg:
This adds display of inline functions to the details tab of the report
view. This shows them for the crashing thread and all other threads.

### DIFF
--- a/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
+++ b/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
@@ -701,17 +701,17 @@
                     </thead>
                     <tbody>
                       {% for frame in parsed_dump.threads[crashing_thread].frames %}
-                        <tr class="{% if frame.missing_symbols %}missingsymbols{% endif %}">
-                          {% if frame.truncated %}
-                            <td colspan="5">truncated {{ frame.truncated|digitgroupseparator }} frames...</td>
-                          {% else %}
-                            <td>
+                        {% if frame.truncated %}
+                          <tr><td colspan="5">truncated {{ frame.truncated|digitgroupseparator }} frames...</td></tr>
+                        {% else %}
+                          <tr class="{% if frame.missing_symbols %}missingsymbols{% endif %}">
+                            <td rowspan="{{ rowspan_for_inlines(frame) }}">
                               {% if frame.missing_symbols %}
                                 <span class="row-notice" title="missing symbol">&Oslash;</span>
                               {% endif %}
                               {{ frame.frame }}
                             </td>
-                            <td>{{ frame.module }}</td>
+                            <td rowspan="{{ rowspan_for_inlines(frame) }}">{{ frame.module }}</td>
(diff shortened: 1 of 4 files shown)


webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -701,17 +701,17 @@ <h2>Crashing Thread ({{ crashing_thread }}){% if parsed_dump.threads[crashing_th
                     </thead>
                     <tbody>
                       {% for frame in parsed_dump.threads[crashing_thread].frames %}
-                        <tr class="{% if frame.missing_symbols %}missingsymbols{% endif %}">
-                          {% if frame.truncated %}
-                            <td colspan="5">truncated {{ frame.truncated|digitgroupseparator }} frames...</td>
-                          {% else %}
-                            <td>
+                        {% if frame.truncated %}
+                          <tr><td colspan="5">truncated {{ frame.truncated|digitgroupseparator }} frames...</td></tr>
```
willkg, line 705: We'll never have a case where the truncated indicator has missing symbols, so we can do a minimal row.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -701,17 +701,17 @@ <h2>Crashing Thread ({{ crashing_thread }}){% if parsed_dump.threads[crashing_th
...
+                        {% else %}
+                          <tr class="{% if frame.missing_symbols %}missingsymbols{% endif %}">
+                            <td rowspan="{{ rowspan_for_inlines(frame) }}">
```
willkg, line 708: We want to show the function plus the inlines. So the rowspan is 1 + len(inlines) which is easier to do as a function than to do in the template.

webapp-django/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -509,3 +509,8 @@ def filter_featured_versions(product_versions):
 @library.global_function
 def filter_not_featured_versions(product_versions):
     return [pv for pv in product_versions if not pv["is_featured"]]
+
+
+@library.global_function
+def rowspan_for_inlines(frame):
+    return 1 + len(frame.get("inlines") or [])
```
willkg, line 516: inlines can be null, so this protects against that.

webapp-django/crashstats/crashstats/utils.py
```diff
@@ -348,6 +349,8 @@ def enhance_json_dump(dump, vcs_mappings):
 
         for frame in frames:
             enhance_frame(frame, vcs_mappings)
+            for inline in frame.get("inlines") or []:
```
willkg, line 352: The inline structure has keys function, file, and line and uses them in the same way as the frame structure, so we can call `enhance_frame` and get the same stuff out. Namely, we want `source_link`.

willkg:
This should have a test even though most of our tests for the report view are pretty janky.

PR #6176, opened 2022-09-05
bug 1604845: remove entropy from raw_crash path

willkg:
This removes the "entropy" section from the raw crash path in AWS S3. This makes it easier to work with raw crash data and also sets us up for moving to GCP where we're not planning to have an "entropy" section at all.

This also fixes the instructions in `bin/remove_field.py` to make it easier to test next time without reading through the code and the PR that created that file.

I went through and tested the following:

* loading raw crashes works with old and new key formats
* verifyprocessed works
* permadelete_crash_data works
* remove_field works
* fetch_crash_data puts crash data in new key location

In relevant places, I added notes about when the deprecated code can be removed.

### DIFF
--- a/bin/permadelete_crash_data.py
+++ b/bin/permadelete_crash_data.py
@@ -33,6 +33,7 @@ Usage::
 import json
 import logging
 import os
+from traceback import print_exception
 
 import click
 from configman import ConfigurationManager
@@ -40,6 +41,7 @@ from configman.environment import environment
 from elasticsearch_dsl import Search
 
 from socorro.external.boto.connection_context import S3Connection
+from socorro.external.boto.crashstorage import build_keys
 from socorro.external.es.connection_context import ConnectionContext
 
 
@@ -108,18 +110,18 @@ def s3_delete(crashid):
     bucket = s3_context.config.bucket_name
     s3_client = s3_context.client
 
-    # Delete the raw crash
-    key = "v2/raw_crash/%(entropy)s/%(date)s/%(crashid)s" % {
-        "entropy": crashid[0:3],
-        "date": "20" + crashid[-6:],
-        "crashid": crashid,
-    }
-    obj = s3_fetch_object(s3_client, bucket, key)
-    if obj:
-        s3_delete_object(s3_client, bucket, key)
+    keys = build_keys("raw_crash", crashid)
+    for key in keys:
+        try:
+            obj = s3_fetch_object(s3_client, bucket, key)
+            if obj:
+                s3_delete_object(s3_client, bucket, key)
+        except Exception as exc:
+            print(f"exception fetching/deleting raw crash: {key}")
+            print_exception(exc)
 
     # Fetch dump_names which tells us which dumps exist
-    key = "v1/dump_names/%(crashid)s" % {"crashid": crashid}
+    key = build_keys("dump_names", crashid)[0]
     dump_names = s3_fetch_object(s3_client, bucket, key)
     if dump_names:
         try:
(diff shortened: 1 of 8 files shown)


bin/permadelete_crash_data.py
```diff
@@ -108,18 +110,18 @@ def s3_delete(crashid):
...
-    }
-    obj = s3_fetch_object(s3_client, bucket, key)
-    if obj:
-        s3_delete_object(s3_client, bucket, key)
+    keys = build_keys("raw_crash", crashid)
+    for key in keys:
+        try:
+            obj = s3_fetch_object(s3_client, bucket, key)
+            if obj:
+                s3_delete_object(s3_client, bucket, key)
```
willkg, line 118: In this case, while it's very unlikely we could have the data at more than one key, it's possible, so we should iterate through all the keys and not do a `break`.

bin/permadelete_crash_data.py
```diff
@@ -108,18 +110,18 @@ def s3_delete(crashid):
...
+        except Exception as exc:
+            print(f"exception fetching/deleting raw crash: {key}")
+            print_exception(exc)
 
     # Fetch dump_names which tells us which dumps exist
-    key = "v1/dump_names/%(crashid)s" % {"crashid": crashid}
+    key = build_keys("dump_names", crashid)[0]
```
willkg, line 124: We should probably do the same thing here where we try every key even though we've never moved `dump_names` so it's only ever had one key.

bin/permadelete_crash_data.py
```diff
@@ -134,10 +136,7 @@ def s3_delete(crashid):
             # Handle dump_name -> key goofiness
             if dump_name in (None, "", "upload_file_minidump"):
                 dump_name = "dump"
-            key = "v1/%(dump_name)s/%(crashid)s" % {
-                "dump_name": dump_name,
-                "crashid": crashid,
-            }
+            key = build_keys(dump_name, crashid)[0]
```
willkg, line 139: Same here re: trying every key.

bin/permadelete_crash_data.py
```diff
@@ -148,11 +147,11 @@ def s3_delete(crashid):
 
         # If there were no errors, then we try to delete dump_names
         if dump_errors == 0:
-            key = "v1/dump_names/%(crashid)s" % {"crashid": crashid}
+            key = build_keys("dump_names", crashid)[0]
             s3_delete_object(s3_client, bucket, key)
 
     # Delete processed crash
-    key = "v1/processed_crash/%(crashid)s" % {"crashid": crashid}
+    key = build_keys("processed_crash", crashid)[0]
```
willkg, line 154: Same here re: trying every key.

socorro/scripts/fetch_crash_data.py
```diff
@@ -82,9 +83,8 @@ def fetch_crash(
 
         # Save raw crash to file system
         raw_crash = resp.json()
-        fn = os.path.join(
-            outputdir, "v2", "raw_crash", crash_id[0:3], "20" + crash_id[-6:], crash_id
-        )
+        key = build_keys("raw_crash", crash_id)[0]
```
willkg, line 86: This script uses the RawCrash API to download the data and uses `build_keys` to figure out where on the file system to put the file. Because of that, we only need to look at the first item in the keys list.

webapp-django/crashstats/crashstats/management/commands/verifyprocessed.py
```diff
@@ -119,6 +132,46 @@ def check_crashids(entropy_chunk, date):
...
+    """Check crash ids for a given firstchars and date"""
+    s3_context = get_s3_context()
+    bucket = s3_context.config.bucket_name
+    s3_client = s3_context.client
+
+    supersearch = SuperSearchUnredacted()
+
+    missing = []
+
+    for firstchars in firstchars_chunk:
```
willkg, line 145: At some point, we might want to change how much chunkification happens. Maybe 2-characters or 1-character is better?

PR #6177, opened 2022-09-06
bug 1604845: switch raw crash key version to v1

willkg:
We're effectively returning to the original v1 from ages ago. Version changing shenanigans are irritating, but in this specific case, I don't think anything else relies on the pathing except antenna and socorro. Given that I control both and I'm changing it anyhow, it seems easier in general to return to v1 which matches the other paths than to keep v2 or go to v3.

### DIFF
--- a/bin/run_mdsw.sh
+++ b/bin/run_mdsw.sh
@@ -44,11 +44,11 @@ do
     fi
 
     # Find the raw crash file
-    RAWCRASHFILE=$(find ${DATADIR}/v2/raw_crash/ -name $CRASHID -type f)
+    RAWCRASHFILE=$(find ${DATADIR}/v1/raw_crash/ -name $CRASHID -type f)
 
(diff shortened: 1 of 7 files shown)


bin/run_mdsw.sh
```diff
@@ -44,11 +44,11 @@ do
     fi
 
     # Find the raw crash file
-    RAWCRASHFILE=$(find ${DATADIR}/v2/raw_crash/ -name $CRASHID -type f)
+    RAWCRASHFILE=$(find ${DATADIR}/v1/raw_crash/ -name $CRASHID -type f)
 
     timeout -s KILL 600 "${STACKWALKER}" \
         --raw-json $RAWCRASHFILE \
-        --symbols-url "https://s3-us-west-2.amazonaws.com/org.mozilla.crash-stats.symbols-public/v1" \
```
willkg, line 51: This should use the symbols server and not access the bucket directly.

PR #6180, opened 2022-09-08
bug 1789705: fix table column widths across the site

willkg:
I threw together some width classes in an earlier commit. This tweaks the widths of the stacks table to fit the data better. This also improves a bunch of other tables to make them more stable layout-wise and remove pixel widths.

I also adjusted borders for table headers. There was a bunch of css code in random places for that. I reduced that some.

### DIFF
--- a/webapp-django/crashstats/crashstats/static/crashstats/css/components/table_sorter.less
+++ b/webapp-django/crashstats/crashstats/static/crashstats/css/components/table_sorter.less
@@ -2,7 +2,8 @@
 
 table.tablesorter {
     th {
-        background-color: @light-grey;
+        background-color: @dark-tan;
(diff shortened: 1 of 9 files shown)


webapp-django/crashstats/crashstats/static/crashstats/css/components/table_sorter.less
```diff
@@ -2,7 +2,8 @@
 
 table.tablesorter {
     th {
-        background-color: @light-grey;
+        background-color: @dark-tan;
+        border: solid 1px @grey;
```
willkg, line 6: The tablesorter thing uses border=black to denote which column is being used to sort. Given that, we should make it grey to start with.

PR #6181, opened 2022-09-08
bug 1788103: redo display of inline functions in report view

willkg:
This fixes the display of inline functions in the report view in two fundamental ways:

1. it switches to displaying functions as depicted in the other mockup
2. it fixes the ordering of frame function vs. the inline functions so it makes sense now

### DIFF
--- a/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
+++ b/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
@@ -704,29 +704,16 @@
                         {% if frame.truncated %}
                           <tr><td colspan="5">truncated {{ frame.truncated|digitgroupseparator }} frames...</td></tr>
                         {% else %}
-                          <tr class="{% if frame.missing_symbols %}missingsymbols{% endif %}">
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -704,14 +704,36 @@ <h2>Crashing Thread ({{ crashing_thread }}){% if parsed_dump.threads[crashing_th
...
+                                  {% if inline.source_link %}
+                                    <a href="{{ inline.source_link }}">{{ inline.file }}:{{ inline.line }}</a>
+                                  {% else %}
+                                    {% if inline.file %}{{ inline.file }}:{{ inline.line }}{% endif %}
+                                  {% endif %}
+                                </td>
+                                <td>inlined</td>
+                              </tr>
+                            {% endfor %}
+                          {% endif %}
```
willkg, line 728: Per these docs, the inline functions should get printed _before_ the frame:

https://github.com/rust-minidump/rust-minidump/blame/188ec7e4bab55a66d76c7171de149a37381d6ca6/minidump-processor/json-schema.md#L309-L317

This change fixes that.

PR #6185, opened 2022-09-08
bug 1749770: fix API endpoints to be more resilient

willkg:
This fixes the API endpoints that derive from SocorroMiddleware to be more resilient to fuzzing.

To test, do things like this:

```
curl 'http://locahost:8000/api/SuperSearchFields/' --data ""
```

That does a POST to SuperSearchFields endpoint. In prod, that kicks up an HTTP 500 because it's using the default post function. This fixes it so now it'll return a `Method not allowed`.

### DIFF
--- a/webapp-django/crashstats/crashstats/models.py
+++ b/webapp-django/crashstats/crashstats/models.py
@@ -629,6 +629,10 @@ class TelemetryCrash(SocorroMiddleware):
     required_params = ("crash_id",)
     aliases = {"crash_id": "uuid"}
 
+    post = None
+    put = None
+    delete = None
+
 
 class PermissionsReducerError(Exception):
     pass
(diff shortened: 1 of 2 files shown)


webapp-django/crashstats/crashstats/models.py
```diff
@@ -629,6 +629,10 @@ class TelemetryCrash(SocorroMiddleware):
     required_params = ("crash_id",)
     aliases = {"crash_id": "uuid"}
 
+    post = None
+    put = None
+    delete = None
```
willkg, line 634: We never use PUT or DELETE, so I don't understand why there's any support for them. I almost removed them, but then I decided not to. Future me will probably regret that. I may rethink this choice tomorrow.

PR #6192, opened 2022-09-09
bug 1790136: don't display line "None"

willkg:
If line data is None, then don't display it at all.

Test with https://crash-stats.mozilla.org/report/index/c8e640f4-6d6d-46aa-a015-9db2c0220909 . After this patch, the first inlined frame should have no line number.

### DIFF
--- a/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
+++ b/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
@@ -3,6 +3,14 @@
   Protected data: <a href="{{ url('documentation:protected_data_access') }}">Protected data policy.</a>
 {%- endmacro %}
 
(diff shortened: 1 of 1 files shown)


webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -3,6 +3,14 @@
   Protected data: <a href="{{ url('documentation:protected_data_access') }}">Protected data policy.</a>
 {%- endmacro %}
 
+{% macro source_link(link, file, line) -%}
+{% if link %}
+  <a href="{{ link }}">{{ file }}{% if line is not none %}:{{ line }}{% endif %}</a>
```
willkg, line 8: "is not none" is the Jinja built-in test for whether the value is a Python `None`.

PR #6193, opened 2022-09-10
bug 1788269: add inline functions to signature

willkg:
This iterates through the inlines if there are any when looking at frames to build the signature.

```
$ socorro-cmd signature 829c31ce-e2d8-4102-9188-d35fa0220908
Crash id: 829c31ce-e2d8-4102-9188-d35fa0220908
Original: js::TenuringTracer::collectToObjectFixedPoint
New:      js::NativeObject::getDenseInitializedLength
Same?:    False
```

### DIFF
--- a/socorro/signature/rules.py
+++ b/socorro/signature/rules.py
@@ -198,7 +198,6 @@ class CSignatureTool:
         line=None,
         module_offset=None,
         offset=None,
-        **kwargs,
     ):
         """Normalizes a single frame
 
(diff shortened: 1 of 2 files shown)


socorro/signature/rules.py
```diff
@@ -198,7 +198,6 @@ def normalize_frame(
         line=None,
         module_offset=None,
         offset=None,
-        **kwargs,
```
willkg, line 201: This makes bugs difficult to find, so I nixed it.

socorro/signature/rules.py
```diff
@@ -233,6 +232,63 @@ def normalize_frame(
...
+            yield {
+                "module": frame.get("module"),
+                "function": frame.get("function"),
+                "file": frame.get("file"),
+                "line": frame.get("line"),
+                "module_offset": frame.get("module_offset"),
+                "offset": frame.get("offset"),
+            }
+
+    def create_frame_list(self, thread_data, make_modules_lower_case=False):
```
willkg, line 267: `create_frame_list` only applies to c signatures, so I moved it to the `CSignatureTool`.

PR #6197, opened 2022-09-13
bug 1763710: prep work for looking at corrupt_symbols stuff

willkg:
This fixes `run_mdsw.sh` so it works again and adds a really basic local symbols server.


bin/entrypoint.sh
```diff
@@ -37,6 +37,9 @@ crontabber)  ## Run crontabber service
 webapp)  ## Run webapp service
     /app/bin/run_webapp.sh "$@"
     ;;
+localsymbols)  ## Runs a local symbols server
+    /app/bin/run_local_symbols.sh
```
willkg, line 41: I should change this to symbolsserver.

PR #6199, opened 2022-09-14
Add enum$<T>::unwrap to prefix list

jrmuizel:
This signature shows up when we do signature generation based on inline stacks. e.g. https://crash-stats.allizom.org/report/index/1fbb67cf-dec1-4729-96d5-061f10220913

### DIFF
--- a/socorro/signature/siglists/prefix_signature_re.txt
+++ b/socorro/signature/siglists/prefix_signature_re.txt
@@ -46,6 +46,7 @@ dvm
 EtwEventEnabled
 extent_
 enum$<T>::unwrap
+enum$<T>::expect
 fastcopy_I
 fastzero_I
 FindElementCommon
--- a/socorro/signature/siglists/prefix_signature_re.txt
+++ b/socorro/signature/siglists/prefix_signature_re.txt
@@ -45,6 +45,7 @@ dlmalloc
 dvm
 EtwEventEnabled
 extent_
+enum$<T>::unwrap
 fastcopy_I
 fastzero_I
 FindElementCommon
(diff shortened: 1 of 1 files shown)


socorro/signature/siglists/prefix_signature_re.txt
```diff
@@ -45,6 +45,8 @@ dlmalloc
 dvm
 EtwEventEnabled
 extent_
+enum$<T>::unwrap
```
willkg, line 48: These are used to build a regular expression so the $ here means "end of line". You need to escape them with a \.

```suggestion
enum\$<T>::unwrap
```

socorro/signature/siglists/prefix_signature_re.txt
```diff
@@ -45,6 +45,8 @@ dlmalloc
...
+enum$<T>::expect
```
willkg, line 49:
```suggestion
enum\$<T>::expect
```

willkg (changes requested):
After escaping the `$`, we get this:

```
app@socorro:/app$ socorro-cmd signature 1fbb67cf-dec1-4729-96d5-061f10220913
Crash id: 1fbb67cf-dec1-4729-96d5-061f10220913
Original: core::result::unwrap_failed | webrender::renderer::Renderer::update_texture_cache
New:      core::result::unwrap_failed | enum$<T>::unwrap | webrender::renderer::upload::copy_into_staging_buffer
Same?:    False
```

The original signature is the one coming from production which doesn't have the inline function changes for signature generation. That's why it looks wrong.

willkg:
Thank you!

PR #6200, opened 2022-09-14
bug 1687987: support raw crash version 2

willkg:
This adds some code to migrate raw crash data from version 1 to version 2. It updates all the places in Socorro to use the new raw crash version 2 locations for things.

It also changes how `minidump_sha256_hash` gets populated in the processed crash. We're going to drop the `MinidumpSha256Hash` fake annotation in the raw crash that the collector adds. The information is in `dump_checksums`, so there's no need to have it in two places and the fake annotation suggests things that are confusing unless you know where it comes from.

### DIFF
--- a/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
+++ b/webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
@@ -1125,11 +1125,11 @@
               <table class="record data-table hardwrapped">
                 <tbody>
                   <tr>
-                    <th class="w-2/12" scope="row">UUID</th>
+                    <th class="w-2/12" scope="row">uuid</th>
                     <td>{{ raw.uuid }}</td>
                   </tr>
                   <tr>
-                    <th scope="row">Collected timestamp</th>
+                    <th scope="row">submitted_timestamp</th>
                     <td>{{ raw.submitted_timestamp | human_readable_iso_date }} UTC</td>
                   </tr>
                   <tr>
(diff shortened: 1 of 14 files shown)


willkg:
I want to wait until after a prod deploy before landing this.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -1144,38 +1144,32 @@ <h4>Collector</h4>
...
-                          {% endfor %}
-                        </table>
-                      {% endif %}
-                    </td>
-                  </tr>
-                  <tr>
-                    <th scope="row">Collector notes</th>
-                    <td><pre>{{ raw.collector_notes|join("\n") }}</pre></td>
-                  </tr>
+                  {% for key, val in raw.metadata | dictsort %}
```
willkg, line 1147: Redoing it so that it shows everything in `metadata` makes it easier to add new fields since we only have to add them in Antenna and don't have to do anything additionally here.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -761,7 +761,6 @@ class RawCrash(SocorroMiddleware):
     API_ALLOWLIST = (
         # Crash annotations:
         # https://searchfox.org/mozilla-central/source/toolkit/crashreporter/CrashAnnotations.yaml
-        "additional_minidumps",
```
willkg, line 764: This field was in the list twice. Further, it's not even something that's in the raw crash, so I'm not sure why it was in there at all.

webapp-django/crashstats/crashstats/models.py
```diff
@@ -799,7 +798,6 @@ class RawCrash(SocorroMiddleware):
         "MacMemoryPressureNormalTime",
         "MacMemoryPressureSysctl",
         "MacMemoryPressureWarningTime",
-        "MinidumpSha256Hash",
```
willkg, line 802: This was a "fake annotation". The processor now pulls this from the `dump_checksums`.

PR #6208, opened 2022-09-22
bug 1787931: fix super search fields to pull permissions from processed crash schema

willkg:
This is a series of commits that move things around to eventually fix super search fields to pull permissions from the processed crash schema.

First, this removes `test_mapping` because it's dead code that's not used anymore. This fixes the `BreadcrumbsRule` and other things that used `get_schema()` and fixes `resolve_references` to return a new version of the data and not mutate the input. This fixes schema validation in `validate_processed_crash` so that it can be used for validating that the schemas themselves are valid in addition to testing them against processed crash data. This also moves some webapp-centric code from `socorro/external/es/super_search_fields.py` to `webapp-django/crashstats/supersearch/libsupersearch.py`. This also fixes the `upload_telemetry_schema` test because it prevented me from running tests using flags. :(

With all that cleanup done, then we add `default_permissions` to the schema so we're not defining defaults across the app.

Finally, we remove permissions from fields in super search fields and add some code to get the permissions from `processed_crash.schema.yaml`. We also add some code to the webapp to convert from schema permissions ("public", "protected", etc) to webapp permissions ("crashstats.view_pii"). This last bit of code is awkward feeling, but I didn't have a good way to generalize it and the webapp permissions are everywhere already.

To test this, I did this before the changes:

```
from socorro.external.es.super_search_fields import FIELDS


for key, val in FIELDS.items():
    print(f"{key}: {val['permissions_needed']}")
```

and this after:

```
from crashstats.supersearch.libsupersearch import SUPERSEARCH_FIELDS as FIELDS           

for key, val in FIELDS.items():                                                          
    print(f"{key}: {val['permissions_needed']}")    
```

And here's the differences:

```
> diff field_permissions_orig.txt field_permissions_new.txt 
39c39
< collector_notes: []
---
> collector_notes: ['crashstats.view_pii']
115c115
< remote_type: []
---
> remote_type: ['crashstats.view_pii']
119d118
< skunk_classification: []
124d122
< support_classification: []
```

The `*_classification` fields were removed because they don't refer to anything in the processed crash. The `remote_type` field had a typo in the `in_database_name` and fixing that fixes its permissions. `collector_notes` were marked protected in the `processed_crash.schema.yaml`, but not in `super_search_fields.py` which is a mistake.

I used `gmp_library_path` to verify that fields that are protected don't show up for people who don't have access to them.

I verified that *all fields* show up in the Super Search API reference.

### DIFF
--- a/socorro/external/es/super_search_fields.py
+++ b/socorro/external/es/super_search_fields.py
@@ -2,13 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-from dataclasses import dataclass
-import datetime
 import logging
-
-from configman import class_converter, Namespace, RequiredConfig
-import elasticsearch
-from elasticsearch_dsl import Search
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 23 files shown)


socorro/external/es/super_search_fields.py
```diff
@@ -252,190 +243,6 @@ def add_doc_values(value):
...
-
-
-@dataclass
-class IndexDataItem:
-    name: str
-    start_date: datetime.datetime
-    count: int
-
-
-class SuperSearchFields(SuperSearchFieldsData):
```
willkg, line 278: This did three things:

1. set up the webapp for getting `FIELDS`: I changed the webapp so it has its own `SUPERSEARCH_FIELDS` and accesses that directly.
2. contain the code for `test_mapping`: This is vestigial code that we don't use anymore, so I removed it entirely.
3. contain the code for supersearch status: This is only used by the admin in the webapp, so I moved all that code there.

Given that, there's no reason to have this code in this module.

socorro/external/es/super_search_fields.py
```diff
@@ -2605,28 +2300,10 @@ def number_field(
             "type": "multi_field",
         },
     },
-    # FIXME(willkg): where does this come from?
```
willkg, line 2608: This hasn't been used in a long time, so I removed it.

socorro/external/es/super_search_fields.py
```diff
@@ -2668,25 +2343,7 @@ def number_field(
             "this crash was submitted by the user: Auto, Infobar, AboutCrashes, "
             "CrashedTab, and Client."
         ),
-        is_protected=False,
     ),
-    # FIXME(willkg): where does this come from?
```
willkg, line 2673: This hasn't been used in a long time, so I removed it.

socorro/processor/rules/mozilla.py
```diff
@@ -434,7 +434,7 @@ def __init__(self, schema):
 
         # NOTE(willkg): if the "breadcrumbs" section ever gets moved in the processed
         # crash schema, we'll need to update this
-        self.BREADCRUMBS_SCHEMA = schema["definitions"]["breadcrumbs"]
+        self.breadcrumbs_schema = schema["properties"]["breadcrumbs"]
```
willkg, line 437: Pulling from "definitions" means we're using a schema that hasn't had references resolved. `breadcrumbs` doesn't currently have references, but it might at some point and it seems prudent to change this to pull from "properties" and use a schema where references have been resolved.

socorro/schemas/socorro-data-1-0-0.schema.yaml
```diff
@@ -10,6 +10,12 @@ description: >
   Schema for defining Socorro-related data schemas.
 
 definitions:
+  simple_permissions:
+    enum:
+      - "public"
+      - "protected"
+    type: string
```
willkg, line 17: This makes it easier to add permission fields in the schema. Now it's defined in one place with all the options.

socorro/schemas/validate_processed_crash.py
```diff
@@ -113,8 +117,12 @@ def match_key(schema_key, document_key):
 @click.argument("crashdir")
 @click.pass_context
 def validate_and_test(ctx, crashdir):
-    jsonschema.Draft4Validator.check_schema(PROCESSED_CRASH_SCHEMA)
-    click.echo("processed crash schema is a valid JSON schema.")
+    socorro_data_schema = get_file_content("socorro-data-1-0-0.schema.yaml")
```
willkg, line 120: This fixes `validate_processed_crash.py` so it can (sort of) be used to validate that the schemas themselves are valid. I say "sort of" because you still have to provide a bunch of processed crash examples and sometimes I pass it an empty directory which allows it to "work", but it's wonky.

socorro/unittest/external/boto/test_upload_telemetry_schema.py
```diff
@@ -7,7 +7,11 @@
 
 
 class TestUploadTelemetrySchema:
-    def test_upload_worked(self, boto_helper, caplogpp):
+    def test_upload_worked(self, boto_helper, caplogpp, monkeypatch):
+        # NOTE(willkg): configman apps look at sys.argv for command line arguments, so
+        # we monkey patch sys.argv so it doesn't pick up pytest arguments and errors
+        # out
+        monkeypatch.setattr("sys.argv", ["upload_telemetry_schema.py"])
```
willkg, line 14: This bug has been incredibly annoying for the last week or two. I finally got around to fixing it today.

webapp-django/crashstats/supersearch/libsupersearch.py
```diff
@@ -0,0 +1,129 @@
...
+            )
+
+        mapping = index_client.get_mapping(index=latest_index)
+        mapping_properties = mapping[latest_index]["mappings"][doctype]["properties"]
+
+        return {
+            "indices": index_data,
+            "latest_index": latest_index,
+            "mapping": mapping_properties,
+        }
```
willkg, line 129: I could probably go one more step and move this code to the admin view where it's used, but I don't think it's worth doing in this PR.

willkg:
This looks good! This is a big win!

PR #6222, opened 2022-10-17
bug 1746630: support unloaded modules in signature generation and report view

willkg:
This updates signature generation to take unloaded modules into account. This uses gankra's notation where it denotes which parts are from unloaded modules from the first commit in PR #6021 . Examples:

```
app@socorro:/app$ socorro-cmd signature edcb4dbc-5516-4b15-9464-26adc0221001 1e8a0180-dd82-4051-b1ac-1006c0221001 9cc1cb90-c92a-4a31-b69c-cebfc0221001
Crash id: edcb4dbc-5516-4b15-9464-26adc0221001
Original: QueryDnsForFamily
New:      (unloaded PrxDrvPE64.dll@0x5387) | QueryDnsForFamily
Same?:    False

Crash id: 1e8a0180-dd82-4051-b1ac-1006c0221001
Original: UserCallWinProcCheckWow
New:      (unloaded RTSUltraMonHook.dll@0xbff8) | UserCallWinProcCheckWow
Same?:    False

Crash id: 9cc1cb90-c92a-4a31-b69c-cebfc0221001
Original: CFGControl::SetFirstVW
New:      (unloaded obs-virtualcam-module64.dll@0xe4df) | CFGControl::SetFirstVW
Same?:    False
```

This also adds support for unloaded modules in the report view in the webapp.

While implementing the relevant tests, I switched them to be pytest parametrized tests because that's a lot cleaner and more maintainable.

Also, while this doesn't implement the functionality in the third commit of PR #6021, this does replace that PR.

### DIFF
--- a/socorro/signature/tests/test_rules.py
+++ b/socorro/signature/tests/test_rules.py
@@ -243,10 +243,10 @@ class TestCSignatureTool:
         # If they're the same, then there was no mutation
         assert frames == frames_copy
 
-    def test_normalize_frame(self):
-        """test_normalize: bunch of variations"""
-        s = self.setup_config_c_sig_tool()
-        a = [
+    @pytest.mark.parametrize(
+        "args, expected",
+        [
+            # module, function, file, line, module_offset, offset, unloaded_modules
             (("module", "", "source/", "23", "0xfff"), "source#23"),
             (("module", "", "source\\", "23", "0xfff"), "source#23"),
             (("module", "", "/a/b/c/source", "23", "0xfff"), "source#23"),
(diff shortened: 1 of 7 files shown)


willkg:
I should fix the create-a-bug links in this PR, too.

gabrielesvelto:
I've got mixed feelings about putting the address in the signature... on the one hand it shows at a glance if we've got an issue with only one version of the unloaded DLL or several, on the other hand it'll require changing signatures on *a lot* of existing bugs.

I'm tempted to say that I prefer the former but I don't have the bandwidth to go through our backlog of crashes and adjust the signatures so if possible I'd prefer if we could stick with the old behavior that would give something like:
```
Crash id: edcb4dbc-5516-4b15-9464-26adc0221001
Original: QueryDnsForFamily
New:      PrxDrvPE64.dll | QueryDnsForFamily
```

willkg:
I think if we don't make the notation change now, we're never going to do it because we'll always be in a situation where  signatures are going to change. If you're fine with that, then I can adjust it so it does what it used to do.

gabrielesvelto:
Alright let's bite the bullet, the crash spike detector should help us out with this one

willkg:
The "omg--signatures are changing" thing sucks up a lot of time. At one point, I was thinking about versioning signature algorithms, but I couldn't figure out how to do it nicely.

I wrote up https://bugzilla.mozilla.org/show_bug.cgi?id=1795762 to think about this more.

socorro/signature/tests/test_rules.py
```diff
@@ -268,10 +268,38 @@ def test_normalize_frame(self):
...
-            assert e == r
+            # Handle normalization with unloaded modules
+            (
+                (
+                    "",
+                    "",
+                    "",
+                    "",
+                    "",
+                    "",
```
willkg, line 279: These tests are the worst. "Is that 4 or 5 empty strings?" Ugh.

webapp-django/crashstats/crashstats/utils.py
```diff
@@ -255,6 +255,25 @@ def _json_clean(value):
...
+
+    :param text: the text to strip leading zeros from
+
+    :returns: hex string with leading zeros stripped
+
+    """
+    try:
+        return hex(int(text, base=16))
+    except (ValueError, TypeError):
+        return text
```
willkg, line 274: This is copied from signature generation code. Though, we can probably import it. I'll fix that now.

PR #6237, opened 2022-11-03
bug 1791208: Quote regex in test.env

jwhitlock:
This was required to run `make build` with docker-compose v2.10.2.


willkg:
Thank you for this!

PR #6248, opened 2022-11-08
bug 1799539: support specifying signature lists

willkg:
This adjusts signature generation to allow someone to build a SignatureGenerator and specify the directory containing signature lists. This functionality is extended to the `socorro-cmd signature` subcommand via a `--signature-list-dir` argument.

### DIFF
--- a/socorro/signature/generator.py
+++ b/socorro/signature/generator.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 import dataclasses
+import inspect
 import sys
 from typing import Any, Dict, List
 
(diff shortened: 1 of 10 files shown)


socorro/signature/generator.py
```diff
@@ -72,15 +73,54 @@ def to_dict(self):
...
 class SignatureGenerator:
-    def __init__(self, pipeline=None, error_handler=None):
+    def __init__(self, ruleset=None, error_handler=None, **cfg):
         """
-        :arg pipeline: list of rules to use for signature generation
-        :arg error_handler: error handling function with signature
+        :param ruleset: list of rule classes to use for signature generation
+        :param error_handler: error handling function with signature
             ``fun(signature_data, exc_info, extra)``
+        :param cfg: configuration keyword arguments to pass to rules being
```
willkg, line 81: I'm not thrilled about the `cfg` kwargs thing, but it was either that or carrying configuration through the layers explicitly or setting up a configuration system ... I figure we can start with this and if/when we see it causing problems, we can adjust as necessary.

socorro/signature/siglists_utils.py
```diff
@@ -27,25 +28,50 @@ class BadRegularExpressionLineError(Exception):
     """Raised when a file contains an invalid regular expression."""
 
 
-def build_filepath(source):
+class _PackageSource:
```
willkg, line 31: I'm going to rename this to "INCLUDED" because that's what we're using in the comments and other places.

PR #6259, opened 2022-11-22
bug 1626698: raw crash schema

willkg:
This implements a raw crash schema just like the processed crash schema we did in bug 1764395. In addition to implementing the schema, this adds tests, a validator script, reworks the RawCrash API to use the schema, reworks the report view to use the schema, and adds some additional tests to the processed crash schema to assert invariants about source annotations.

Some other changes that are riding along with this:

1. removing support for `BIOS_Manufacturer` and `MemoryErrorCorrection` annotations. These were removed from Firefox in bug 1710152, so we no longer need to support them in crash ingestion.
2. removing support for `NumberOfProcessors` annotation. Looks like this hasn't been used in a long time and it's not in `CrashAnnotations.yaml`. The data is in the minidump now.
3. removing support for `AdapterRendererIDs` annotation. This hasn't been emitted by products in a long time and it's not in `CrashAnnotations.yaml`.

To verify this:

1. I went through the existing unit tests, fixed them up to cover the changes (and in some cases, fix errors in the tests that were unearthed by these changes), implemented new ones to check new invariants.
4. Downloaded data using the RawCrash API and no api token (thus, this is public data) from production and localhost and compared the output to make sure we weren't leaking data.
5. Downloaded data using the RawCrash API and an api token (thus, this is all the raw crash data--public and protected) from production and localhost and compared the output to make sure the raw crash api didn't change.
6. Went through the report view to make sure it continued to work and didn't break because of a change in permissions including the Crash Annotations and Raw Data and Minidumps tabs.
7. Used Firefox private window to go through crash reports on localhost to make sure they worked correctly when logged out.

I'm fairly confident everything is good at this point.


socorro/schemas/socorro-data-1-0-0.schema.yaml
```diff
@@ -18,6 +18,7 @@ definitions:
 
   simple_types:
     enum:
+      - "any"
```
willkg, line 21: I added `any` so that I could include `metadata` in the schema. This is a collection of collector-generated fields that capture what the payload looked like, the throttling rule that accepted the crash report, etc. These fields are diagnostic and relevant only to socorro engineers debugging collector issues. It's helpful to be able to make changes to the collector without having to do a lock-step set of changes with the rest of socorro across multiple repositories. That's what went into making `metadata` type `any` and using `pattern_properties` to cover all the sub fields.

webapp-django/crashstats/crashstats/tests/test_models.py
```diff
@@ -400,14 +400,11 @@ def mocked_get(**params):
...
                 "Version": "5.0a1",
                 "Vendor": "Mozilla",
-                # This is not allowlisted
-                "URL": "https://example.com/",
             }
 
         model.implementation().get.side_effect = mocked_get
         r = api.get(crash_id="some-crash-id")
         assert r["Vendor"] == "Mozilla"
-        assert "URL" in r  # no filtering at this level
```
willkg, line 410: This is different now. If you don't explicitly set an `api_user` that has protected-data permissions, then you now get a filtered public-only set of data. public-by-default.

webapp-django/crashstats/crashstats/tests/test_views.py
```diff
@@ -1440,13 +1440,15 @@ def mocked_raw_crash_get(**params):
...
-                }
+                crash["TelemetryEnvironment"] = json.dumps(
+                    {
+                        "key": ["values"],
+                        "plainstring": "I am a string",
+                        "plainint": 12345,
+                        "empty": [],
+                        "foo": {"keyA": "AAA", "keyB": "BBB"},
+                    }
+                )
```
willkg, line 1451: This test was bad. The value for `TelemetryEnvironment` (and all annotations) is always a string.

webapp-django/crashstats/crashstats/views.py
```diff
@@ -184,13 +184,12 @@ def report_index(request, crash_id, default_context=None):
...
-    context["public_raw_keys"] = [
-        x for x in context["raw"] if x in models.RawCrash.API_ALLOWLIST()
-    ]
+    all_public_raw_keys = models.RawCrash.public_keys()
+    context["public_raw_keys"] = [x for x in context["raw"] if x in all_public_raw_keys]
     if request.user.has_perm("crashstats.view_pii"):
         # If the user can see PII or this is their crash report, include everything
         context["protected_raw_keys"] = [
-            x for x in context["raw"] if x not in models.RawCrash.API_ALLOWLIST()
+            x for x in context["raw"] if x not in all_public_raw_keys
```
willkg, line 192: It might be better to provide a "is_public" function that takes a field name and returns True or False. I'm planning to do a pass on reducing the complexity of the code that generates the report view and I can think about this more then.

socorro/schemas/validate_raw_crash.py
```diff
@@ -0,0 +1,202 @@
...
+    # key is in the schema, but is missing a type
+    keys_not_in_schema = document_keys.keys - reduced_keys.keys
+    if keys_not_in_schema:
+        click.echo("")
+        click.echo(
+            f"{len(keys_not_in_schema)} keys in crash reports but not in schema:"
+        )
+        click.echo(f"  {'KEY':90}  TYPE(S)")
+        for key, val in sorted(keys_not_in_schema):
+            click.echo(f"  {key:90}  {val}")
```
willkg, line 198: It might be interesting to have this also fetch `CrashAnnotations.yaml` and then list out the annotations that are in the raw crash schema and not in `CrashAnnotations.yaml` and vice-versa.

willkg:
There will probably be additional changes to make after this. Plus there are a bunch of bugs to work on for follow-up work.

PR #6264, opened 2022-11-29
bug 1792255: add telemetry_environment to processed crash schema

willkg:
This adds the telemetry_environment bits to the processed crash schema. I cribbed a lot of the types and descriptions from:
    
https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/d7c892dbfe46c37c44d897b8edb8b0d9f5273fec/schemas/telemetry/event/event.4.schema.json
    
A lot of fields don't have descriptions. That's unfortunate. We can fill those in as we go along, I guess.
    
I didn't include things that weren't in the crash reports I spot checked. Whatever isn't in the schema will be treated as protected data, so we can add it as we need it without risk of a data leak.
    
I also fixed the report view and template to use the version in the processed crash. It's normalized and validated now, so it's less likely for weird data to cause an error in the report view.

To test this, I processed and validated 100 Firefox and 100 Fenix crash reports against the changes. I looked at example data.


socorro/lib/libsocorrodataschema.py
```diff
@@ -270,6 +270,8 @@ def transform_schema(schema, transform_function):
     """
 
     def _transform_schema(schema, path, transform_function):
+        if "type" not in schema:
+            raise KeyError(f"'type' not in schema at {path}")
```
willkg, line 274: Prior to this change, all you'd get was something like `KeyError("type")` and it wasn't clear where in the (very long) schema that happened. Now it shows the path which makes fixing these issues (which I keep hitting) much much easier.

socorro/schemas/validate_processed_crash.py
```diff
@@ -95,22 +100,42 @@ def match_key(schema_key, document_key):
...
+    # which needs to match something like this:
+    #
+    # .telemetry_environment.settings.userPrefs.browser.startup.homepage
+    #
+    # because the keys have dots, too.
+    #
+    # Another way to do this matching thing is to store the path as a list of string
+    # parts rather than a .-delimited string, but that involves rewriting a bunch of
+    # stuff about paths and traversals and such.
+    schema_parts = []
```
willkg, line 123: This is kind of ugly, but as I say in the comment, the right fix is to redo paths everywhere and I didn't want to tackle that in this PR. I'll write up a follow-up bug to do that.

PR #6273, opened 2022-12-01
bug 1803558: data dictionary for crash reporting data

willkg:
This adds a data dictionary section to the Crash Stats website. It's similar to the Glean Dictionary where it documents all the annotation and processed crash fields that are currently known and supported by Socorro.

This is a first pass. It's kind of ugly. It's missing some things I would like it to have.

However, it does have some nice features already:

1. it has descriptions, type information, and permission information for all the support fields
2. it supports markdown in descriptions
3. it links to data reviews
4. for fields that are searchable, it includes examples of the data from the last week


webapp-django/crashstats/documentation/views.py
```diff
@@ -58,16 +62,124 @@ def read_whatsnew():
...
 
 @track_view
 @pass_default_context
 def protected_data_access(request, default_context=None):
     context = default_context or {}
-    return render(request, "documentation/protected_data_access.html", context)
+    return render(request, "docs/protected_data_access.html", context)
+
+
+def get_annotation_schema_data():
```
willkg, line 95: Oops--these should be cached.

PR #6285, opened 2022-12-06
bug 1796094: fix JSLargeAllocationFailure OOM detection

willkg:
The OOM is when the value is Reporting and not Reported. This fixes that.


willkg:
```
app@socorro:/app$ socorro-cmd signature ae980f94-a122-43fa-bc31-db3360221206
Crash id: ae980f94-a122-43fa-bc31-db3360221206
Original: OOM | small
New:      OOM | large | mozalloc_abort | moz_xmalloc | new
Same?:    False
```

PR #6287, opened 2022-12-06
Switch to using contextlib.suppress

willkg:
This is mostly cosmetic, but it reads better.

### DIFF
--- a/socorro/lib/__init__.py
+++ b/socorro/lib/__init__.py
@@ -4,6 +4,8 @@
 
 """common library code for socorro modules"""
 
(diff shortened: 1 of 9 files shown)


socorro/lib/__init__.py
```diff
@@ -18,11 +20,9 @@ def __init__(self, arg):
         self.arg = arg
 
     def __str__(self):
-        try:
+        with suppress(Exception):
```
willkg, line 23: This is awful, but I'm not going to fix it today.

PR #6291, opened 2022-12-30
bug 1791210: fix socorro to work with docker compose > 2

willkg:
Previously, Socorro only worked with docker-compose < 2. This fixes everything (including the docs) to work with docker compose > 2.


willkg:
Oops--made this with the wrong bug number.

PR #6292, opened 2022-12-30
bug 1791208: fix socorro to work with docker compose > 2

willkg:
Previously, Socorro only worked with docker-compose < 2. This fixes everything (including the docs) to work with docker compose > 2.

I did this work because I updated my laptop and don't have the old docker-compose anymore, so I needed to update in order to look into bug 1807956. I figured I'd throw it into a PR before I forgot.


willkg:
It's dying with this:

```
Copying my.env.dist to my.env...
/usr/bin/docker compose build  --build-arg userid=10001 --build-arg groupid=10001 app
unexpected character "." in variable name near "resource.logging.level=DEBUG\n\n# postgres\n# --------\n
...
make: *** [Makefile:49: build] Error 15
```

The issue is that the configman-style environment variables don't work with the env things that docker does (or bash)--variable names can't have periods in them. I'll have to think about this. Maybe we change it so that env files get loaded inside the environment with the run script rather than via Docker with the `env_file` things in the `docker-compose.yml` file.

willkg:
Turns out I could update the cimg/python image version and the setup_remote_docker version and that got around the configman style variables issue. Yay!

willkg:
I'll wait to land this until after the code freeze. Also, after landing this, we can remove some of the dependabot PRs that target `requirements-compose.txt`.

PR #6296, opened 2023-01-04
bug 1808406: fix uses of functools.cache decorating methods

willkg:
This fixes uses of `functools.cache` decorating methods. Generally, this is problematic because it uses `self` in the hash key which then would cause instances to never get garbage collected. In these specific cases, the instances are singletons except one where the decorated thing was a classmethod. Given that, I don't think these will cause memory leaks, so I don't think this will change anything.


socorro/external/es/connection_context.py
```diff
@@ -90,6 +89,9 @@ def __init__(self, config):
         super().__init__()
         self.config = config
 
+        # Cached answers for things that don't change
+        self._mapping_cache = {}
```
willkg, line 93: If we find ourselves doing this much more, we should write up a `methodcache`. I didn't do that, yet, because I didn't want to deal with locking and all the other complexities involved in caching on the instance.

socorro/processor/rules/mozilla.py
```diff
@@ -106,17 +105,22 @@ def build_fields(self, schema):
...
             value of
 
         :returns: schema
 
         """
-        return SocorroDataReducer(self.schema["properties"][schema_key])
+        cache_key = f"_socorro_data_reducer_{schema_key}"
+        reducer = getattr(self, cache_key, None)
+        if reducer is None:
+            reducer = SocorroDataReducer(self.schema["properties"][schema_key])
```
willkg, line 122: Oops--I missed adding it.

PR #6298, opened 2023-01-04
bug 1808593: rewrite c/s/models.py get_api_allowlist

willkg:
This rewrites the `crashstats/supersearch/models.py` `get_api_allowlist()` so it makes more sense with the current architecture. Many years ago, we moved super search fields from being defined in the database (where they were mutable) to being defined in a Python module (where they aren't).

`get_api_allowlist()` was written when the fields were mutable. Now that they aren't, we can take a lot of the dynamicism out of it.

### DIFF
--- a/webapp-django/crashstats/supersearch/models.py
+++ b/webapp-django/crashstats/supersearch/models.py
@@ -3,9 +3,6 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 import copy
-import functools
-
-from django.core.cache import cache
 
 from crashstats import productlib
 from crashstats.crashstats import models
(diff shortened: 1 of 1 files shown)


webapp-django/crashstats/supersearch/models.py
```diff
@@ -43,26 +40,19 @@
...
+    fields = []
+    for meta in all_fields.values():
+        if (
+            meta["name"] not in fields
+            and meta["is_returned"]
+            and (include_all_fields or not meta["permissions_needed"])
+        ):
+            fields.append(meta["name"])
+
+    return {"hits": fields}
```
willkg, line 55: We can make this return a real value rather than a callable since all this stuff is static between deploys now.

PR #6311, opened 2023-01-11
bug 1807956: truncate stacks in processor

willkg:
This adds a rule to the processor to truncate stacks in the stackwalker output. Stacks can sometimes be egregiously large and stacks that large often contain repeated sequences of frames and the repetition isn't helpful.

This picks a `MAX_FRAMES` of 500 figuring that's enough frames to get a sense of things. If you need more frames, you should download the minidump, run the stackwalker, and do your analysis on the raw stackwalker output.


socorro/processor/rules/breakpad.py
```diff
@@ -70,6 +70,70 @@ def action(self, raw_crash, dumps, processed_crash, status):
...
+class TruncateStacksRule(Rule):
+    """Truncate stacks that are too large
+
+    This rule truncates stacks that are greater than MAX_FRAMES. It truncates the stacks
+    in the middle with the thinking that the beginning and end of the stack are the most
+    interesting.
+
+    """
+
+    MAX_FRAMES = 500
```
willkg, line 82: We picked 500 because it's enough to see what's going on, but not so much that the webapp is overwhelmed. We can change this number if it's too much/little.

socorro/processor/rules/breakpad.py
```diff
@@ -70,6 +70,70 @@ def action(self, raw_crash, dumps, processed_crash, status):
...
+    HALF_MAX_FRAMES = int(MAX_FRAMES / 2)
+
+    def __init__(self):
+        super().__init__()
+        self.metrics = markus.get_metrics("processor.truncatestacksrule")
+
+    def truncation_frame(self, truncated_frames):
+        return {"truncated": {"msg": f"{len(truncated_frames):,} frames truncated"}}
```
willkg, line 90: I made this a structure in case we decide to "summarize" the truncated frames at some point in the future. By making it a structure, we can add additional information in a backwards-compatible way by adding new bits to the structure without changing existing bits.

socorro/processor/rules/breakpad.py
```diff
@@ -70,6 +70,70 @@ def action(self, raw_crash, dumps, processed_crash, status):
...
+                thread["truncated"] = True
+
+    def action(self, raw_crash, dumps, processed_crash, status):
+        # Traverse processed_crash for "json_dump" sections. Each one of these has
+        # multiple stacks in it.
+        for key, value in processed_crash.items():
+            if key == "json_dump":
+                self.truncate_stacks(value)
+
+            elif isinstance(value, dict):
```
willkg, line 132: I did it this way in case we add more "upload_file_whatever" things to the schema. Currently we only have "upload_file_minidump_browser", but we used to have a bunch of them. By being general here, we can add new sections to the schema and don't have to remember to update this code, too.

socorro/schemas/processed_crash.schema.yaml
```diff
@@ -158,7 +158,6 @@ definitions:
       The output of rust-minidump stackwalker.
     properties:
       crash_info:
-        # FIXME(willkg): add unloaded_modules
```
willkg, line 161: We added unloaded_modules, so this FIXME was out of date.

socorro/unittest/processor/rules/test_breakpad.py
```diff
@@ -222,6 +223,202 @@ def test_empty_json_dump(self):
...
+                    {
+                        "frame_count": 100,
+                        "frames": copy.deepcopy(frames)[0:100],
+                    },
+                ],
+                "modules": [],
+            }
+        }
+
+        validate_instance(instance=processed_crash, schema=PROCESSED_CRASH_SCHEMA)
```
willkg, line 295: This validates that the test data we're running this test with is valid processed crash data.

socorro/unittest/processor/rules/test_breakpad.py
```diff
@@ -222,6 +223,202 @@ def test_empty_json_dump(self):
...
+        assert json_dump["threads"][0]["truncated"] is True
+        assert json_dump["threads"][0]["frames"][250] == {
+            "truncated": {"msg": "502 frames truncated"}
+        }
+
+        # Check thread 1 is not truncated
+        assert len(json_dump["threads"][1]["frames"]) == 100
+        assert "truncated" not in json_dump["threads"][1]
+
+        validate_instance(instance=processed_crash, schema=PROCESSED_CRASH_SCHEMA)
```
willkg, line 321: This validates that the output of the rule is valid processed crash data. If we ever change the `truncated` structure, this will make sure we remember to update the schema.

webapp-django/crashstats/crashstats/jinja2/crashstats/report_index.html
```diff
@@ -717,8 +717,8 @@ <h2>Crashing Thread ({{ crashing_thread }}){% if parsed_dump.threads[crashing_th
                     </thead>
                     <tbody>
                       {% for frame in parsed_dump.threads[crashing_thread].frames %}
-                        {% if frame.truncated %}
-                          <tr><td colspan="5">truncated {{ frame.truncated|digitgroupseparator }} frames...</td></tr>
```
willkg, line 721: This changes the template to show the msg the processor created. That's good enough for now.

webapp-django/crashstats/crashstats/utils.py
```diff
@@ -355,17 +349,7 @@ def enhance_json_dump(dump, vcs_mappings):
         if "thread" not in thread:
             thread["thread"] = thread_index
 
-        # We're only going to show 100 frames. If there are more than that in the stack,
-        # we're going to truncate in the middle with a "special" frame
         frames = thread["frames"]
-        if len(frames) > MAX_FRAMES:
```
willkg, line 361: We've moved truncation to the processor, so we don't need it here anymore.

willkg:
Originally, I was thinking I'd do "remove truncation from webapp" as a follow-up bug at some point in the future, but as I was testing this code, I decided I didn't want to reason about crash reports that were truncated in the processor and also truncated by the webapp because the `MAX_FRAMES` numbers were different.

Instead, I removed the truncation from the webapp. There will be crash reports with large stacks that were processed before this code is deployed. We can reprocess them as they come up which will truncate the stacks. There aren't a lot of them and it'll be increasingly less of an issue as data expires out of the system.

socorro/processor/rules/breakpad.py
```diff
@@ -70,6 +70,70 @@ def action(self, raw_crash, dumps, processed_crash, status):
...
+        """Truncates a single stack if it's too big
+
+        :arg frames: the list of frame structures to truncate
+
+        :returns: truncated list of frame structures with a truncation
+            frame in the middle
+
+        """
+        self.metrics.gauge("stack_size", len(frames))
+        self.metrics.incr("truncated")
```
willkg, line 102: This emits a `stack_size` metric so we can see the stack sizes over time. This only emits a metric for stacks > `MAX_FRAMES`. I think that's fine for now, but maybe at some point, we change it to emit a metric for all stacks.

This emits a `truncated` metric so we can see how often we truncate stacks.

PR #6338, opened 2023-02-02
bug 1767282: fix temp files in processor

willkg:
The processor uses a temporary directory to save crash report data in order to process it. Previously there were multiple parts of the processor saving data in different places--some of which weren't safe from thread contention.
    
This fixes that by pulling ownership of the temporary directory (creation, deletion) into the processor app and then passing that
directory around to the pipeline and rules. The processor app now creates a new temporary directory in process_crash that's unique to that processing run. This should fix orphaned files and thread contention issues (and future possible process contention issues).

### DIFF
--- a/docker/config/local_dev.env
+++ b/docker/config/local_dev.env
@@ -52,7 +52,6 @@ resource.boto.s3_endpoint_url=http://localstack:4566/
 resource.boto.access_key=foo
 secrets.boto.secret_access_key=foo
 resource.boto.bucket_name=dev-bucket
-resource.boto.temporary_file_system_storage_path=/tmp
 resource.boto.region=us-west-2
 
 resource.boto.sqs_endpoint_url=http://localstack:4566/
--- a/docker/config/local_dev.env
+++ b/docker/config/local_dev.env
@@ -63,9 +63,6 @@ resource.boto.reprocessing_queue=local-dev-reprocessing
 # processor
 # ---------
 
-# Use the rust-minidump minidump-stackwalk binary
-processor.minidump_stackwalker=rust
-
 # In the docker local dev environment, we store symbol cache and other things
 # in /tmp because there's only one processor node. For server environments, we
 # probably want to store that in a volume. These vars are all affected.
@@ -73,9 +70,9 @@ companion_process.symbol_cache_path=/tmp/symbols/cache
 processor.minidumpstackwalk.symbol_cache_path=/tmp/symbols/cache
 processor.minidumpstackwalk.symbol_tmp_path=/tmp/symbols/tmp
 
-# Drop kill_timeout to 30 because this is a dev environment and 5 minutes is
+# Drop kill_timeout to 60 because this is a dev environment and 5 minutes is
 # a long time
-processor.minidumpstackwalk.kill_timeout=30
+processor.minidumpstackwalk.kill_timeout=120
 
 # Set symbols_urls to two symbol servers so we make sure we can specify multiple
 # --symbols-url values in minidump-stackwalk
--- a/socorro/external/boto/crashstorage.py
+++ b/socorro/external/boto/crashstorage.py
@@ -140,12 +140,6 @@ class BotoS3CrashStorage(CrashStorageBase):
         reference_value_from="resource.boto",
     )
     required_config.add_option(
-        "temporary_file_system_storage_path",
-        doc="a local filesystem path where dumps temporarily during processing",
-        default="/home/socorro/temp",
-        reference_value_from="resource.boto",
-    )
-    required_config.add_option(
         "dump_file_suffix",
         doc="the suffix used to identify a dump file (for use in temp files)",
         default=".dump",
@@ -250,7 +244,7 @@ class BotoS3CrashStorage(CrashStorageBase):
         except self.conn.KeyNotFound as x:
             raise CrashIDNotFound("%s not found: %s" % (crash_id, x))
 
-    def get_dumps_as_files(self, crash_id):
+    def get_dumps_as_files(self, crash_id, tmpdir):
         """Get the dump files for given crash id and save them to tmp.
 
         :returns: dict of dumpname -> file path
@@ -262,7 +256,7 @@ class BotoS3CrashStorage(CrashStorageBase):
         # convert our native memory dump mapping into a file dump mapping.
         return in_memory_dumps.as_file_dumps_mapping(
             crash_id,
-            self.config.temporary_file_system_storage_path,
+            tmpdir,
             self.config.dump_file_suffix,
         )
 
--- a/socorro/external/fs/crashstorage.py
+++ b/socorro/external/fs/crashstorage.py
@@ -214,7 +214,9 @@ class FSPermanentStorage(CrashStorageBase):
         ) as f:
             return f.read()
 
-    def get_dumps_as_files(self, crash_id):
+    def get_dumps_as_files(self, crash_id, tmpdir):
+        # NOTE(willkg): We don't need to use tmpdir here because the files are already
+        # on the file system.
         parent_dir = self._get_radixed_parent_directory(crash_id)
         if not os.path.exists(parent_dir):
             raise CrashIDNotFound
@@ -232,7 +234,7 @@ class FSPermanentStorage(CrashStorageBase):
         )
 
     def get_dumps(self, crash_id):
-        file_dump_mapping = self.get_dumps_as_files(crash_id)
+        file_dump_mapping = self.get_dumps_as_files(crash_id, None)
         # ensure that we return a name/blob mapping
         return file_dump_mapping.as_memory_dumps_mapping()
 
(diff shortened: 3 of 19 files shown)


docker/config/local_dev.env
```diff
@@ -63,19 +62,16 @@ resource.boto.reprocessing_queue=local-dev-reprocessing
 # processor
 # ---------
 
-# Use the rust-minidump minidump-stackwalk binary
-processor.minidump_stackwalker=rust
```
willkg, line 67: This is cruft from 2021 when we switched stackwalkers.

docker/config/local_dev.env
```diff
@@ -63,19 +62,16 @@ resource.boto.reprocessing_queue=local-dev-reprocessing
...
-
 # In the docker local dev environment, we store symbol cache and other things
 # in /tmp because there's only one processor node. For server environments, we
 # probably want to store that in a volume. These vars are all affected.
 companion_process.symbol_cache_path=/tmp/symbols/cache
 processor.minidumpstackwalk.symbol_cache_path=/tmp/symbols/cache
 processor.minidumpstackwalk.symbol_tmp_path=/tmp/symbols/tmp
 
-# Drop kill_timeout to 30 because this is a dev environment and 5 minutes is
+# Drop kill_timeout to 60 because this is a dev environment and 5 minutes is
```
willkg, line 72: Oops--this should be 120.

socorro/external/boto/crashstorage.py
```diff
@@ -262,7 +256,7 @@ def get_dumps_as_files(self, crash_id):
         # convert our native memory dump mapping into a file dump mapping.
         return in_memory_dumps.as_file_dumps_mapping(
             crash_id,
-            self.config.temporary_file_system_storage_path,
```
willkg, line 265: One of the effects of this configuration variable and its use is that all thread stored the dump files of crash reports they were working on in the same directory with no uniqueness to the filename. If two threads happen to be working on the same crash report (this is possible--there's nothing to prevent it), then there's a race condition around deleting files and processing. That's the cause of bugs 1742100 and 1767282.

socorro/external/fs/crashstorage.py
```diff
@@ -214,7 +214,9 @@ def get_raw_dump(self, crash_id, name=None):
         ) as f:
             return f.read()
 
-    def get_dumps_as_files(self, crash_id):
+    def get_dumps_as_files(self, crash_id, tmpdir):
+        # NOTE(willkg): We don't need to use tmpdir here because the files are already
```
willkg, line 218: The file system already has the files on the file system where the stackwalker (and other things) can access them. We don't need to copy them to another place on the file system. We also don't want to delete them when processing of this crash is done. Given that, we can ignore the temporary directory.

socorro/processor/processor_app.py
```diff
@@ -188,63 +197,67 @@ def process_crash(self, crash_id, ruleset_name):
...
-                # Clean up any dump files saved to the file system
-                for a_dump_pathname in dumps.values():
-                    if "TEMPORARY" in a_dump_pathname:
-                        try:
-                            os.unlink(a_dump_pathname)
-                        except OSError as x:
-                            self.logger.info("deletion of dump failed: %s", x)
+            with tempfile.TemporaryDirectory(dir=self.temporary_path) as tmpdir:
+                self.logger.info("using tmpdir %s", tmpdir)
+                # Fetch the raw crash data
```
willkg, line 207: All this code was indented because of the temporary directory context, but is otherwise the same.

socorro/processor/processor_app.py
```diff
@@ -188,63 +197,67 @@ def process_crash(self, crash_id, ruleset_name):
...
-
-            finally:
-                # Clean up any dump files saved to the file system
-                for a_dump_pathname in dumps.values():
-                    if "TEMPORARY" in a_dump_pathname:
-                        try:
-                            os.unlink(a_dump_pathname)
-                        except OSError as x:
-                            self.logger.info("deletion of dump failed: %s", x)
+            with tempfile.TemporaryDirectory(dir=self.temporary_path) as tmpdir:
```
willkg, line 205: When the context exits or the object is garbage collected, it will delete the contents of the temporary directory and the directory itself. That's a lot nicer than what we had before where we assumed the files created during processing were all dump files and in the `dumps` values.

socorro/processor/processor_pipeline.py
```diff
@@ -136,11 +136,6 @@ class ProcessorPipeline(RequiredConfig):
         doc="absolute path to symbol cache",
         default=os.path.join(tempfile.gettempdir(), "symbols"),
     )
-    required_config.minidumpstackwalk.add_option(
```
willkg, line 139: Instead of the minidump stackwalker having its own temporary directory (separate from the one the dumps are using), we use the temporary directory the processor app created for processing this crash report.

socorro/processor/rules/breakpad.py
```diff
@@ -146,30 +144,6 @@ def action(self, raw_crash, dumps, processed_crash, status):
...
-    :param tmp_path: str path to temp storage
-    :param raw_crash: dict of raw crash data
-    :param crash_id: crash id for this crash report
-
-    :yields: absolute path to temp file
-
-    """
-
-    path = os.path.join(
-        tmp_path, f"{crash_id}.{threading.currentThread().getName()}.TEMPORARY.json"
```
willkg, line 162: This uses the thread name to prevent contention between threads processing the same crash report. However, when we switch from multi-threaded to multi-process, this wouldn't work anymore. This was another reason to redo how the processor manages temporary files/directories during processing.

socorro/processor/rules/breakpad.py
```diff
@@ -359,6 +330,14 @@ def run_stackwalker(self, crash_id, command_path, command_line, status):
                 self.logger.warning(f"{msg} ({crash_id})")
                 output = {}
 
+        else:
+            # Log last 500 characters replacing \n with "\\n"
+            self.logger.error(
+                "MinidumpStackwalkRule: minidump-stackwalk failed: %s %s",
```
willkg, line 336: I should wrap the last part in quotes so it's more obvious when it's an empty string.

socorro/processor/rules/breakpad.py
```diff
@@ -397,68 +376,72 @@ def run_stackwalker(self, crash_id, command_path, command_line, status):
...
-
-                if dump_name == self.dump_field:
-                    processed_crash.update(stackwalker_data)
-
-                else:
-                    if dump_name not in processed_crash["additional_minidumps"]:
-                        processed_crash["additional_minidumps"].append(dump_name)
-                    processed_crash.setdefault(dump_name, {})
-                    processed_crash[dump_name].update(stackwalker_data)
+        # Save crash annotations to disk for stackwalker to look at
```
willkg, line 384: This is mostly the same except it's dedented because we no longer have the `tmp_raw_crash_file` context.

socorro/processor/rules/breakpad.py
```diff
@@ -397,68 +376,72 @@ def run_stackwalker(self, crash_id, command_path, command_line, status):
...
+        raw_crash_path = f"{tmpdir}/{crash_id}.json"
```
willkg, line 385: This should be `os.path.join`.

socorro/tests/external/boto/test_crashstorage.py
```diff
@@ -301,8 +299,8 @@ def test_get_dumps_not_found(self, boto_helper):
         with pytest.raises(CrashIDNotFound):
             boto_s3_store.get_dumps("0bba929f-dead-dead-dead-a43c20071027")
 
-    def test_get_dumps_as_files(self, boto_helper, tmpdir):
-        boto_s3_store = self.get_s3_store(tmpdir=tmpdir)
+    def test_get_dumps_as_files(self, boto_helper, tmp_path):
```
willkg, line 302: This switches from the old `tmpdir` pytest fixture which returned a `py.something.Path` to the new `tmp_path` pytest fixture which returns a `pathlib.Path`.

socorro/tests/processor/test_processor_app.py
```diff
@@ -99,15 +101,20 @@ def test_transform_success(self):
         pa.processor.process_crash = mocked_process_crash
         pa.destination.save_processed_crash = mock.Mock()
         finished_func = mock.Mock()
-        patch_path = "socorro.processor.processor_app.os.unlink"
```
willkg, line 102: The processor app is no longer in charge of deleting files manually from the dumps values. Instead, it does it automatically when the temporary directory context exits.

Given that, we don't need to add this testing.

socorro/tests/processor/test_processor_app.py
```diff
@@ -99,15 +101,20 @@ def test_transform_success(self):
...
         pa.source.get_raw_crash.assert_called_with("17")
         pa.processor.process_crash.assert_called_with(
-            "default", fake_raw_crash, fake_dumps, fake_processed_crash
+            ruleset_name="default",
+            raw_crash=fake_raw_crash,
+            dumps=fake_dumps,
+            processed_crash=fake_processed_crash,
+            # FIXME(willkg): testing this is tricky--we'd have to mock
+            # TemporaryDirectory
+            tmpdir=ANY,
```
willkg, line 117: The value here will be something like `/tmp/tmpxidifS43` and changes every time so it's tricky to test unless we mock out parts of the tempfile module.

PR #6339, opened 2023-02-04
bug 1529342: more configman-related cleanup

willkg:
This merges the FetchTransformSaveApp and the ProcessorApp. The ProcessorApp was the only subclass, so it made sense to merge the two allowing us to remove a lot of code getting around the awkward API barrier between the two. This adjusts some of the processor code to reduce the complexity for the `process_crash` method.

While doing that, I discovered a bunch of other uses of configman's `DotDict` that weren't configuration related. This fixes those, too.

After this, I could look at merging the task manager code with the processor app. After that, I'm not sure what else I can do to reduce the configman -> everett lift.

To test this:

1. run tests
2. run `socorro-cmd` and some of the subcommands
3. process some crashes and look at the logging output


socorro/processor/processor_app.py
```diff
@@ -131,6 +180,10 @@ class ProcessorApp(FetchTransformSaveApp):
         from_string_converter=class_converter,
     )
 
+    def __init__(self, config):
+        super().__init__(config)
+        self.waiting_func = None
```
willkg, line 185: I think we can get rid of the `waiting_func`. I'm not sure what it does and it appears to always be set to `None`.

socorro/processor/processor_app.py
```diff
@@ -173,107 +231,138 @@ def _basic_iterator(self):
...
+        # been processed, yet
+        try:
+            processed_crash = self.source.get_processed(crash_id)
+            new_crash = False
+        except CrashIDNotFound:
+            new_crash = True
+            processed_crash = {}
+
+        # Process the crash and remove any temporary artifacts from disk
+        # Process the crash to generate a processed crash
```
willkg, line 313: This is redundant.

socorro/processor/processor_app.py
```diff
@@ -173,107 +231,138 @@ def _basic_iterator(self):
...
+        try:
+            self.logger.debug("saving %s", crash_id)
+            self.destination.save_processed_crash(raw_crash, processed_crash)
+            self.logger.info("saved %s", crash_id)
+        except PolyStorageError as poly_storage_error:
+            # Capture and log the exceptions raised by storage backends
+            for storage_error in poly_storage_error:
+                sentry_sdk.capture_exception(storage_error)
+                self.logger.error("error: crash id %s: %r", crash_id, storage_error)
+            self.logger.warning("error in processing or saving crash %s", crash_id)
```
willkg, line 333: This is wrong now.

socorro/tests/processor/__init__.py
```diff
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-from configman.dotdict import DotDict as DotDict
```
willkg, line 5: I think this is a remnant from the days when we had multiple DotDict things.

PR #6342, opened 2023-02-05
bug 1529342: fix temporary path value

willkg:
The existing infrastructure configuration uses the old resource.boto.temporary_file_system_storage_path variable that we dropped (code-wise) in 1eb8a195. This adds some code to make the code backwards-compatible with the old configuration.

### DIFF
--- a/socorro/processor/processor_app.py
+++ b/socorro/processor/processor_app.py
@@ -165,7 +165,15 @@ class ProcessorApp(App):
             "a local filesystem path that can be used as a workspace for processing "
             + "rules"
         ),
-        default=os.path.join(tempfile.gettempdir(), "workspace"),
+        default=os.path.join(
+            (
+                # FIXME(willkg): this is for backwards compatability with existing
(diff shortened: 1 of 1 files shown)


socorro/processor/processor_app.py
```diff
@@ -165,7 +165,15 @@ class ProcessorApp(App):
             "a local filesystem path that can be used as a workspace for processing "
             + "rules"
         ),
-        default=os.path.join(tempfile.gettempdir(), "workspace"),
+        default=os.path.join(
+            (
+                # FIXME(willkg): this is for backwards compatability with existing
+                # configuration
+                os.environ.get("resource.boto.temporary_file_system_storage_path", "")
```
willkg, line 172: If this environment variable exists, it takes this value as the default value. If not, it takes `tempfile.gettempdir()` (which is a more sensible default).

This will bridge the current state of main with the current configuration in stage/prod until we update stage/prod configuration.

PR #6347, opened 2023-02-06
bug 1529342: convert from configman to everett

willkg:


willkg:
This is an absurdly huge PR. I get that. Here's why this is so huge:

1. configman is integrated in the entirety of Socorro, so removing it required a rearchitecture of how everything goes together
2. the new configuration code is simpler and we're no longer doing configuration-driven component architectures so I removed a lot of component code to simplify the processor as much as possible
3. all of that required me to rewrite a lot of test code
4. many of the webapp tests were heavily mocked out with incorrect test data and bugs, so I ended up rewriting most of those; the mocked out tests should probably have been rewritten when I switched Socorro over to a Docker-based local dev environment years ago
5. removing the mocking from tests required improved test helper fixtures; these new fixtures will make it possible to write tests for `get_versions_for_products` in the future (bug 1548671)
6. removing mocking from tests also involved removing many uses of `unittest.TestCase` (bug 1504067)
7. I removed `MetricsCounter` and `BenchmarkingWrapper` (bug 1453086)
8. I removed the `PolyStorage` and `PolyStorageError` handling--it's much simpler now
9. I removed the companion process code and the symbol cache manager. We need a symbol cache manager, so I need to write a new one.

I did a pass a couple of days ago that chipped some of the bits in this PR off and put those in PRs that have landed already. I think I could do some more of that work, but it's increasingly hard to do and I think it'd require me to implement and test intermediary migration steps which is more work, but not great value at this point.

Now we're at a point where everything passes in CI. Next steps:

1. I need to replace the symbol cache manager.
2. I need to change how the processor starts up to use Honcho and runs the symbol cache manager as a separate process.
3. I need to review everything and make sure it makes sense. I'll keep a list of things that aren't consistent and push off as many of those issues to later PRs or later bugs.
4. I need to go through and make sure everything works in my local dev environment.
5. I need to write up the configuration changes we need to make in the infrastructure.
6. I need to land the configuration changes.
7. I need to write up metrics keys changes we need to make in the dashboards.

Maybe another week of work to go before we can get this on stage and test it thoroughly there.

bin/es_cli.py
```diff
@@ -110,52 +104,57 @@ def cmd_list_crashids(ctx, index):
...
-    context = get_conn()
-    doctype = context.get_doctype()
-    conn = context.connection()
-    indices_client = IndicesClient(conn)
-    resp = indices_client.get_mapping(index=index)
-    mapping = resp[index]["mappings"][doctype]["properties"]
-    click.echo(json.dumps(mapping, indent=2, sort_keys=True))
+    crashstorage = get_crashstorage()
+    doctype = crashstorage.get_doctype()
+    with crashstorage.client() as conn:
```
willkg, line 109: I'm pretty sure I'm being inconsistent with naming things. Crashstorage instances should have a connection which is a ConnectionContext instance and those should have a client which is the client from the library it's using to access the storage system (e.g. boto, elasticsearch-py, etc).

I'll clean up consistency issues in another PR.

bin/permadelete_crash_data.py
```diff
@@ -82,7 +80,7 @@ def s3_delete_object(client, bucket, key):
         click.echo(f"s3: deleted {key!r}")
         return True
     except Exception:
-        logger.exception(f"ERROR: s3: when deleting {key!r}")
+        logger.exception("ERROR: s3: when deleting %r", key)
```
willkg, line 83: While rewriting code, I went through and made sure to fix logging so that messages get interpolated by the logging client. That improves logging because it tracks the arguments.

bin/remove_field.py
```diff
@@ -29,15 +29,13 @@
...
-from configman.environment import environment
 from elasticsearch.exceptions import ConnectionError
 from elasticsearch_dsl import Search
 from more_itertools import chunked
 
-from socorro.external.boto.connection_context import S3Connection
+from socorro import settings
 from socorro.external.boto.crashstorage import build_keys, dict_to_str
-from socorro.external.es.connection_context import ConnectionContext
+from socorro.libclass import build_instance_from_settings
```
willkg, line 38: I thought about putting `libclass.py` into `socorro/lib/`, but then put it one directory up because in bug 1810238, I'm working on reducing redundancy in structure names and paths. In that bug, I'll either move this to `socorro/lib/` or move everything out of `socorro/lib/` up a directory.

bin/remove_field.py
```diff
@@ -89,19 +79,22 @@ def wait_times_access():
     wait_time_generator=wait_times_access,
     module_logger=logger,
 )
-def fix_data_in_s3(fields, bucket, s3_client, crash_data):
```
willkg, line 92: This `s3_client` is a boto client. We swap it for a `s3_crashstorage` which is a `socorro.external.boto.crashstorage.S3CrashStorage` instance which has a `connection` that's the `ConnectionContext`. This means that `remove_field` is using more of our API and less of the boto API directly.

bin/run_processor.sh
```diff
@@ -4,18 +4,19 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-# Usage: bin/run_processor.sh
+# Usage: bin/run_processor_honcho.sh
```
willkg, line 7: Oops--this should be `run_processor.sh`. I'll fix this.

bin/run_processor.sh
```diff
@@ -4,18 +4,19 @@
...
 #
-# Runs the processor.
+# Runs processor process manager.
 #
 # Note: This should be called from inside a container.
 
 set -euo pipefail
 
-CMDPREFIX="${CMDPREFIX:-}"
+PROCESSOR_WORKERS=${PROCESSOR_WORKERS:-"1"}
```
willkg, line 15: We're setting this to 1 for now because the processor is still a multi-threaded application. In a future PR for bug 907277, we'll change it from a multi-threaded app to a multi-process app.

bin/sqs_cli.py
```diff
@@ -10,41 +10,21 @@
 #
```
willkg, line 10: I moved this script into `bin` where all the scripts should live. Plus it should be able to run outside of `socorro-cmd`.

bin/sqs_cli.py
```diff
@@ -10,41 +10,21 @@
...
-        aws_secret_access_key=os.environ.get("secrets.boto.secret_access_key"),
-    )
-    client = session.client(
-        service_name="sqs",
-        region_name=os.environ.get("resource.boto.region"),
-        endpoint_url=os.environ.get("resource.boto.sqs_endpoint_url"),
-    )
-    return client
+    cls = import_class(settings.QUEUE["class"])
+    return cls.build_client(**settings.QUEUE["options"])
```
willkg, line 27: This is weird, but what's going on is that we can't instantiate a `SQSCrashQueue` because it tries to get the queue urls for the standard, priority, and reprocessing queues and if those queues haven't been created, yet, then it fails. Given that this is the script that the local dev environment uses to create, delete, and get information from the state of SQS in the local dev environment, we can't use `SQSCrashQueue` and have to use the boto client directly.

bin/sqs_cli.py
```diff
@@ -160,7 +140,9 @@ def create(ctx, queue):
         return
 
     conn = get_client()
-    validate_queue_name(queue)
+
+    cls = import_class(settings.QUEUE["class"])
+    cls.validate_queue_name(queue)
```
willkg, line 145: `validate_queue_name` is a classmethod on `SQSCrashQueue`.

bin/upload_telemetry_schema.py
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env python
```
willkg, line 1: This used to be a SocorroApp, but I rewrote it as a cli using click and moved it to `bin/`.

bin/upload_telemetry_schema.py
```diff
@@ -0,0 +1,99 @@
...
+
+    """
+
+    basedir = Path(__file__).resolve().parent.parent
+
+    configure_sentry(
+        basedir,
+        host_id=settings.HOST_ID,
+        sentry_dsn=settings.SENTRY_DSN,
+    )
```
willkg, line 82: We don't need to do this anymore. We should make sure upload_telemetry_schema is run with the sentry wrapper.

docker/config/test.env
```diff
@@ -1,10 +1,8 @@
...
 # Logging level at DEBUG so we catch everything in tests
-resource.logging.level=DEBUG
-
-# put filesystem crashstorage stuff in /tmp/test_crashes
-resource.fs.fs_root=/tmp/test_crashes
+LOGGING_LEVEL=DEBUG
+CACHE_MANAGER_LOGGING_LEVEL=DEBUG
 
 # Statsd
 STATSD_CLIENT=django_statsd.clients.null
```
willkg, line 8: I think this is vestigial and can be removed. We're using Markus now.

socorro/__init__.py
```diff
@@ -1,3 +1,116 @@
...
+import glom
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+UPPERCASE_KEY_RE = re.compile(r"^[A-Z][A-Z_]+$")
+
+
+class Settings:
```
willkg, line 21: This needs a docstring.

socorro/__init__.py
```diff
@@ -1,3 +1,116 @@
...
+                continue
+            value = sanitize_value(key, getattr(self, key))
+
+            logger.info("%s: %r", key, value)
+
+    def __repr__(self):
+        return f"<Settings {self._settings_module_path!r}>"
+
+    @contextlib.contextmanager
+    def override(self, **kwargs):
```
willkg, line 58: This needs usage examples.

willkg: This could probably use some tests, too. Does it build up correctly? Does it override correctly? Does it take things down correctly?

socorro/external/boto/connection_context.py
```diff
@@ -124,24 +105,39 @@ def __init__(self, config):
...
     )
-    def build_client(self):
-        """Returns a Boto3 S3 Client."""
+    def build_client(
+        cls,
+        region,
+        access_key=None,
+        secret_access_key=None,
+        endpoint_url=None,
+        **kwargs,
```
willkg, line 114: I don't remember why this kwargs is here.

socorro/external/boto/connection_context.py
```diff
@@ -154,15 +150,15 @@ def build_client(self):
         wait_time_generator=wait_times_access,
         module_logger=logger,
     )
-    def save_file(self, path, data):
+    def save_file(self, bucket, path, data):
```
willkg, line 153: I changed the role of this connection context so it just wraps the boto API and handles the connection specifically. It no longer builds keys, or knows about buckets, or any of that.

socorro/external/boto/connection_context.py
```diff
@@ -202,10 +197,48 @@ def load_file(self, path):
...
+            resp = self.client.get_object(Bucket=bucket, Key=path)
             return resp["Body"].read()
         except self.client.exceptions.NoSuchKey:
             raise KeyNotFound(
-                "%s (bucket=%r key=%r) not found, no value returned"
-                % (id, self.config.bucket_name, path)
+                f"(bucket={bucket!r} key={path}) not found, no value returned"
+            )
+
+    def list_objects_paginator(self, bucket, prefix):
```
willkg, line 207: I moved some code from other places to here so that it's not spread out across the code base. This is used by the webapp command that looks for crash reports we collected, but didn't process.

socorro/external/boto/crash_data.py
```diff
@@ -49,7 +49,7 @@ def get(self, **kwargs):
             # Raw Crash
             "meta": "get_raw_crash",
             # Redacted processed crash
-            "processed": "get_processed",
+            "processed": "get_processed_crash",
```
willkg, line 52: We had "get_raw_crash" and "get_processed'. I made that consistent everywhere so now we have "get_raw_crash" and "get_processed_crash".

socorro/external/boto/crash_data.py
```diff
@@ -60,10 +60,7 @@ def get(self, **kwargs):
             else:
                 return get(params["uuid"])
         except CrashIDNotFound as cidnf:
-            self.logger.warning(
-                "%(datatype)s not found: %(exception)s",
-                {"datatype": params["datatype"], "exception": cidnf},
-            )
+            self.logger.warning("%s not found: %s", params["datatype"], cidnf)
```
willkg, line 63: I went through and fixed logging messages. However, now that I'm revisiting the logging docs, this was valid before I changed it. I think I'm going to leave it like this.

socorro/external/boto/crashstorage.py
```diff
@@ -79,34 +77,17 @@ def build_keys(name_of_thing, crashid):
         entropy = crashid[:3]
         date = get_datestamp(crashid).strftime("%Y%m%d")
         return [
-            "v1/%(nameofthing)s/%(date)s/%(crashid)s"
-            % {
-                "nameofthing": name_of_thing,
-                "date": date,
-                "crashid": crashid,
-            },
+            f"v1/{name_of_thing}/{date}/{crashid}",
```
willkg, line 80: f-strings make this code so much shorter and more legible.

socorro/external/boto/crashstorage.py
```diff
@@ -163,11 +181,11 @@ def save_raw_crash(self, raw_crash, dumps, crash_id):
 
         path = build_keys("raw_crash", crash_id)[0]
         raw_crash_data = dict_to_str(raw_crash).encode("utf-8")
-        self.conn.save_file(path, raw_crash_data)
+        self.save_file(path, raw_crash_data)
```
willkg, line 184: We moved the "ownership" of the bucket to the crash storage, so now it has to pass the bucket along with all the calls. I created `load_file` and `save_file` methods to make it easier.

socorro/external/crashqueue_base.py
```diff
@@ -29,7 +21,7 @@ def __iter__(self):
         ``finished_func`` when it's done processing the crash.
 
         """
-        pass
+        raise NotImplementedError("__iter__ not implemented")
```
willkg, line 24: This matches what `CrashStorageBase` does.

socorro/external/crashstorage_base.py
```diff
@@ -143,31 +136,18 @@ def migrate_raw_crash(data):
...
-        self.config = config
-        self.namespace = namespace
-
+    def __init__(self):
         # Collection of non-fatal exceptions that can be raised by a given storage
         # implementation. This may be fetched by a client of the crashstorge so that it
         # can determine if it can try a failed storage operation again.
-        self.exceptions_eligible_for_retry = ()
         self.logger = logging.getLogger(__name__ + "." + self.__class__.__name__)
+        self.exceptions_eligible_for_retry = ()
```
willkg, line 147: I'm not sure why this line got moved, but we should move it back because the comment applies to it.

socorro/external/crashstorage_base.py
```diff
@@ -253,368 +242,65 @@ def remove(self, crash_id):
         raise NotImplementedError("remove is not implemented")
 
 
-class PolyStorageError(Exception, MutableSequence):
```
willkg, line 256: I removed all the poly-storage stuff and adjusted the processor so it always supports multiple crash destinations. That's a lot easier than having this abstraction layer with the complex poly-storage-error handling.

socorro/external/crashstorage_base.py
```diff
@@ -253,368 +242,65 @@ def remove(self, crash_id):
...
-        return self.exceptions.__contains__(value)
-
-    def __getitem__(self, index):
-        """Get a specific exception by index"""
-        return self.exceptions.__getitem__(index)
-
-    def __setitem__(self, index, value):
-        """Set the value for an index in the sequence"""
-        self.exceptions.__setitem__(index, value)
+class InMemoryCrashStorage(CrashStorageBase):
```
willkg, line 245: I added this to make it easier to write tests, but in retrospect, I probably could have written the tests using the FSCrashStorage. I'm leaving it here for now, but maybe we should remove it later.

socorro/external/crashstorage_base.py
```diff
@@ -253,368 +242,65 @@ def remove(self, crash_id):
...
-                crash_id = processed_crash.get("uuid", "NONE")
-                self.logger.error(
-                    "%r failed (crash id: %s)", store_class, crash_id, exc_info=True
-                )
-                storage_exception.gather_current_exception()
-        if storage_exception.has_exceptions():
-            raise storage_exception
-
-
-class BenchmarkingCrashStorage(CrashStorageBase):
```
willkg, line 470: I removed this. The only thing we used it for was gathering timing metrics for `save_processed_crash` in the crashstorage destinations. It's easier to assume we want that and add a couple of lines to the processor code that's saving the crash data in all the destinations.

socorro/external/crashstorage_base.py
```diff
@@ -253,368 +242,65 @@ def remove(self, crash_id):
...
-
-    def __init__(self, config, namespace=""):
-        self.config = config
-        self.metrics = markus.get_metrics(self.config.metrics_prefix)
-
-    def _make_key(self, *args):
-        return ".".join(x for x in args if x)
-
-
-class MetricsCounter(MetricsEnabledBase):
```
willkg, line 576: I removed this. It's a faux crash storage that is used to keep a count of how many times the processor saves a crash report which is a proxy for how many crash reports the processor processed. It's better to add an incr to the relevant code in the processor than have all this faux crash storage code.

socorro/external/crashstorage_base.py
```diff
@@ -253,368 +242,65 @@ def remove(self, crash_id):
...
+        try:
+            return self._processed_crash_data[crash_id]
+        except KeyError:
+            raise CrashIDNotFound(f"{crash_id} not found")
 
-    def __getattr__(self, attr):
-        wrapped_attr = getattr(self.wrapped_object, attr)
-        if attr in self.config.active_list:
+    def remove(self, crash_id):
+        with suppress(KeyError):
```
willkg, line 299: `suppress` is my new favorite contextlib thing.

socorro/external/es/connection_context.py
```diff
@@ -3,94 +3,29 @@
...
 
-from socorro.external.es.super_search_fields import build_mapping
-from socorro.lib.libdatetime import utc_now
 
-
-# Elasticsearch indices configuration.
-ES_CUSTOM_ANALYZERS = {
-    "analyzer": {"semicolon_keywords": {"type": "pattern", "pattern": ";"}}
-}
-ES_QUERY_SETTINGS = {"default_field": "signature"}
```
willkg, line 21: I reduced the role of the connection context to _just_ the connection things and moved all the crashstorage/search related things to the crashstorage class.

socorro/external/es/connection_context.py
```diff
@@ -153,64 +58,18 @@ def __call__(self, name=None, timeout=None):
...
-                resp = self.indices_client().get_mapping(index=index_name)
-                mapping = resp[index_name]["mappings"][es_doctype]["properties"]
-                self._mapping_cache[cache_key] = mapping
-            except elasticsearch.exceptions.NotFoundError:
-                if reraise:
-                    raise
-        return mapping
-
-    def create_index(self, index_name, mappings=None):
+    def create_index(self, index_name, index_settings):
```
willkg, line 61: I can probably change the arguments here to `name` and `settings`. I'm on the fence about whether to clean this up now or later. This PR is so big.

socorro/external/es/connection_context.py
```diff
@@ -153,64 +58,18 @@ def __call__(self, name=None, timeout=None):
...
         """Create an index that will receive crash reports.
 
         :arg index_name: the name of the index to create
-        :arg mappings: dict of doctype->ES mapping
+        :arg index_settings: index settings including the mapping
 
-        :returns: True if the index was created, False if it already
-            existed
+        :returns: True if the index was created, False if it already existed
```
willkg, line 67: This is sort of a lie. It's True if it was created and False if it already existed or there was some other error. I'm not going to fix this now, but I will add a FIXME.

willkg: My bad--the relevant code is obscured by the lack of diff context. Scratch this.

socorro/external/es/connection_context.py
willkg: I started to change it, but then it was a terrible idea because we'd be using different terms for the same thing in some of the code. Also, "settings" refers to something else. Therefore, I'm leaving it.

socorro/external/es/connection_context.py
```diff
@@ -221,41 +80,20 @@ def create_index(self, index_name, mappings=None):
...
-        ]
-        indices.sort()
+        indices = status["indices"].keys()
         return indices
 
     def delete_index(self, index_name):
         """Delete an index."""
         self.indices_client().delete(index_name)
 
-    def delete_expired_indices(self):
```
willkg, line 239: This is moved to the crashstorage.

socorro/external/es/crashstorage.py
```diff
@@ -264,32 +272,164 @@ def build_document(src, crash_document, fields, all_keys):
...
+                    "analysis": ES_CUSTOM_ANALYZERS,
+                }
+            },
+            "mappings": mappings,
+        }
+
+    def get_mapping(self, index_name, es_doctype, reraise=False):
+        """Retrieves the mapping for a given index and doctype
+
+        NOTE(willkg): Mappings are cached on the ConnectionContext instance. If you
```
willkg, line 355: This is wrong. Need to update it.

socorro/external/es/crashstorage.py
```diff
@@ -264,32 +272,164 @@ def build_document(src, crash_document, fields, all_keys):
...
+            otherwise
+
+        :returns: mapping as a dict or None
+
+        """
+        cache_key = f"{index_name}::{es_doctype}"
+        mapping = self._mapping_cache.get(cache_key)
+        if mapping is None:
+            try:
+                resp = self.client.indices_client().get_mapping(index=index_name)
```
willkg, line 371: The crashstorage code shouldn't access the indices_client directly. We should have an intermediary method.

socorro/external/es/query.py
```diff
@@ -7,33 +7,26 @@
...
-        default="socorro.external.es.connection_context.ConnectionContext",
-        from_string_converter=class_converter,
-    )
-
     filters = [("query", None, "json"), ("indices", None, ["list", "str"])]
 
-    def __init__(self, config):
-        self.config = config
-        self.context = self.config.elasticsearch_class(self.config)
+    def __init__(self, crashstorage, timeout=120):
```
willkg, line 21: I like this change a lot. Instead of query accessing the connection context directly, it now goes through a crashstorage. That makes tons of sense since the crashstorage defines all the structures the query is querying.

This is true for supersearch, too.

socorro/external/fs/crashstorage.py
```diff
@@ -129,10 +110,10 @@ def _cleanup_empty_dirs(base, leaf):
                 break
 
     def _get_dump_file_name(self, crash_id, dump_name):
-        if dump_name == self.config.dump_field or not dump_name:
-            return crash_id + self.config.dump_file_suffix
+        if dump_name == self.dump_field or not dump_name:
+            return crash_id + self.dump_file_suffix
         else:
-            return "%s.%s%s" % (crash_id, dump_name, self.config.dump_file_suffix)
+            return "%s.%s%s" % (crash_id, dump_name, self.dump_file_suffix)
```
willkg, line 116: I should have turned this into an f-string. I'll do that now.

socorro/external/fs/crashstorage.py
```diff
@@ -167,11 +148,12 @@ def _save_files(self, crash_id, files):
                     f.write(contents)
 
     def save_raw_crash(self, raw_crash, dumps, crash_id):
-        if dumps is None:
-            dumps = MemoryDumpsMapping()
+        dumps = dumps or {}
```
willkg, line 151: Previously, this took a different type for `dumps` than other crashstorage. Fun.

socorro/external/sqs/crashqueue.py
```diff
@@ -173,28 +177,51 @@ def get_queue_url(self, queue_name):
...
 
+    def get_queue_url(self, queue_name):
+        """Returns the SQS queue url for the given queue name
+
+        :arg queue_name: the queue name
+
+        :returns: the url for the queue
+
+        """
+        return self.client.get_queue_url(QueueName=queue_name)["QueueUrl"]
```
willkg, line 216: Unlike the crashstorage classes that use a connectioncontext which then creates a client, this creates the boto client without the intermediary. I almost changed this, but theoretically we're moving to GCP so the SQS code is not long for this world and spending more time on it doesn't make a lot of sense.

webapp/crashstats/settings/base.py
```diff
@@ -30,16 +30,25 @@ def path(*dirs):
...
+TOOL_ENV = config("TOOL_ENV", False, cast=bool)
+if TOOL_ENV:
+    fake_values = [
+        ("ELASTICSEARCH_URL", "http://elasticsearch:9200"),
+        ("SECRET_KEY", "ou812"),
+    ]
+    for key, val in fake_values:
+        os.environ[key] = val
 
-# Set this to True to make debugging AJAX requests easier; development-only!
```
willkg, line 37: We don't use this, so I removed it.

webapp/crashstats/settings/base.py
```diff
@@ -343,22 +357,6 @@ def filter(self, record):
 # Base URL for Buildhub
 BUILDHUB_BASE_URL = "https://buildhub.moz.tools/"
 
-ELASTICSEARCH_URLS = config(
```
willkg, line 346: We don't use these keys anymore.

webapp/crashstats/settings/base.py
```diff
@@ -382,11 +380,6 @@ def filter(self, record):
 # Path to the view that gets executed if you hit upon a ratelimit block
 RATELIMIT_VIEW = "crashstats.crashstats.views.ratelimit_blocked"
 
-# We don't want to test the migrations when we run tests.
```
willkg, line 385: South. That was a long time ago.

webapp/crashstats/settings/base.py
```diff
@@ -457,14 +438,8 @@ def filter(self, record):
 
 TIME_ZONE = config("TIME_ZONE", "UTC")
 
-
-# We'll possible reuse this later in this file
-database_url = config("DATABASE_URL", "sqlite://sqlite.crashstats.db")
-
-DATABASES = {"default": dj_database_url.parse(database_url)}
-
-# Uncomment this and set to all slave DBs in use on the site.
```
willkg, line 466: We don't use this.

webapp/crashstats/settings/base.py
```diff
@@ -528,58 +503,10 @@ def filter(self, record):
 # Set to True enable analysis of all model fetches
 ANALYZE_MODEL_FETCHES = config("ANALYZE_MODEL_FETCHES", True, cast=bool)
 
-# This `IMPLEMENTATIONS_DATABASE_URL` is optional. By default, the
```
willkg, line 531: We don't use this.

webapp/crashstats/settings/base.py
```diff
@@ -528,58 +503,10 @@ def filter(self, record):
...
-implementations_database_url = config("IMPLEMENTATIONS_DATABASE_URL", "")
-if not implementations_database_url:
-    implementations_database_url = database_url
-implementations_config = dj_database_url.parse(implementations_database_url)
-
 # The list of valid rulesets for the Reprocessing API
+# FIXME(willkg): we can pluck this from settings or structure
 VALID_RULESETS = ["default", "regenerate_signature"]
 
-# The CrashQueueBase class to use for submitting priority and reprocessing
```
willkg, line 543: All the Socorro configuration bits are now in `socorro.mozilla_settings` which is accessed via `from socorro import settings as socorro_settings`. It's nice to have the settings in one place now.

webapp/crashstats/crashstats/views.py
```diff
@@ -131,9 +131,12 @@ def report_index(request, crash_id, default_context=None):
...
 
-    context["product_details"] = productlib.get_product_by_name(
-        context["report"]["product"]
-    )
+    try:
+        context["product_details"] = productlib.get_product_by_name(
+            context["report"]["product"]
+        )
+    except productlib.ProductDoesNotExist:
+        context["product_details"] = {}
```
willkg, line 139: This should have handled products that don't exist. We could have crash data and then decide to stop supporting a product--you should be able to see the report views.

webapp/crashstats/crashstats/tests/testbase.py
```diff
@@ -2,54 +2,11 @@
...
-        if issubclass(klass, crashstats.crashstats.models.SocorroMiddleware):
-            # Remember, the default thing for SocorroMiddleware is
-            # that it always has a class attribute called `implementation`
-            # it's by default set to None.
-            if klass.implementation:
-                classes_with_implementation.append(klass)
-
 
 class DjangoTestCase(django.test.TestCase):
-    def setUp(self):
```
willkg, line 27: I removed all the implementation class mocking. It was a tremendous undertaking, but it was the only way to verify that the configman-less code worked right.

socorro/tests/external/sqs/conftest.py
```diff
@@ -1,14 +0,0 @@
-# This Source Code Form is subject to the terms of the Mozilla Public
-# License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at https://mozilla.org/MPL/2.0/.
-
-import pytest
-
-from socorro.tests.external.sqs import get_sqs_config, SQSHelper
-
-
-@pytest.fixture
```
willkg, line 10: I moved the sqs_helper to `socorro/tests/conftest.py`.

socorro/tests/external/sqs/__init__.py
```diff
@@ -1,100 +1,3 @@
...
-        [sqs_config],
-        app_name="test-sqs",
-        app_description="",
-        values_source_list=[environment],
-        argv_source=[],
-    )
-    return config_manager.get_config()
-
-
-class SQSHelper:
```
willkg, line 27: I moved `SQSHelper` to `socorro/tests/conftest.py`.

socorro/tests/external/es/base.py
```diff
@@ -1,186 +0,0 @@
...
-            app_version="1.0",
-            app_description="Elasticsearch integration tests",
-            values_source_list=[environment, values_source],
-            argv_source=[],
-        )
-
-        return config_manager.get_config()
-
-
-class ElasticsearchTestCase(TestCaseWithConfig):
```
willkg, line 87: I moved most of this to an `es_helper` fixture that's more convenient to use. That's in `socorro/tests/conftest.py`.

socorro/tests/external/es/base.py
```diff
@@ -1,186 +0,0 @@
...
-    "resource.elasticsearch.elasticsearch_index": "testsocorro_%W",
-    "resource.elasticsearch.elasticsearch_index_regex": "^testsocorro.*$",
-    "resource.elasticsearch.elasticsearch_timeout": 10,
-}
-
-
-CRON_JOB_EXTA_VALUES = {"resource.elasticsearch.backoff_delays": [1]}
-
-
-def minimum_es_version(minimum_version):
```
willkg, line 29: This got moved to `socorro/tests/external/es/__init__.py`.

socorro/lib/task_manager.py
```diff
@@ -102,13 +98,10 @@ def _get_iterator(self):
...
-            return self.job_param_source_iter(self.config)
-        except TypeError:
-            try:
-                return self.job_param_source_iter()
-            except TypeError:
-                return self.job_param_source_iter
+        # FIXME(willkg): test this out; we dropped the class instance one
+        if callable(self.job_source_iterator):
+            return iter(self.job_source_iterator())
+        return iter(self.job_source_iterator)
```
willkg, line 104: I rewrote this, but I'm not sure it does the same thing it used to do because the old code is bereft of "this is what this is intended to do" comments. I redid the related tests as well and it works with the code we have in Socorro, so it's probably fine.

socorro/lib/threaded_task_manager.py
```diff
@@ -8,68 +8,58 @@
...
+            job_source_iterator=job_source_iterator,
+            task_func=task_func,
+        )
         self.thread_list = []  # the thread object storage
-        self.number_of_threads = config.number_of_threads
-        self.task_queue = queue.Queue(config.maximum_queue_size)
+        self.number_of_threads = number_of_threads
+        self.task_queue = queue.Queue(maximum_queue_size)
+
+        self.queueing_thread = None
```
willkg, line 62: "queuing" turns out to be a valid spelling, but I hate it, so I changed it. Further, the `start` code creates this instance variable, but it should be initialized, so I added this line here.

socorro/lib/threaded_task_manager.py
```diff
@@ -119,34 +112,24 @@ def blocking_start(self):
...
-            # it only ends if someone hits  ^C or sends SIGHUP or SIGTERM -
-            # any of which will get translated into a KeyboardInterrupt
         except KeyboardInterrupt:
             while True:
                 try:
                     self.stop()
                     break
                 except KeyboardInterrupt:
-                    self.logger.warning(
-                        "We heard you the first time.  There "
```
willkg, line 131: I took out comments like this. I'm not into snarky applications.

socorro/lib/threaded_task_manager.py
```diff
@@ -156,67 +139,73 @@ def wait_for_empty_queue(self, wait_log_interval=0, wait_reason=""):
...
 
-        This is a blocking call.  The thread using this function will wait for
-        all the worker threads to die."""
+        This is a blocking call. The thread using this function will wait for
+        all the worker threads to end.
+
+        """
         for x in range(self.number_of_threads):
-            self.task_queue.put((None, None))
+            self.task_queue.put(STOP_TOKEN)
```
willkg, line 154: `STOP_TOKEN` is so much clearer.

socorro/lib/threaded_task_manager.py
```diff
@@ -156,67 +139,73 @@ def wait_for_empty_queue(self, wait_log_interval=0, wait_reason=""):
             seconds += 1
             time.sleep(1.0)
 
-    def _kill_worker_threads(self):
-        """This function coerces the consumer/worker threads to kill
```
willkg, line 160: I took out all the suicide, abortion, and homocide language and replaced it with clearer language.

socorro/mozilla_rulesets.py
```diff
@@ -0,0 +1,121 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+from socorro import settings
```
willkg, line 5: This was all in `socorro/processor/processor_pipeline.py` and I moved it to here.

socorro/mozilla_settings.py
```diff
@@ -0,0 +1,309 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+"""
```
willkg, line 5: This module is similar to how Django settings module works. You define `SOCORRO_SETTINGS` in the environment as a Python dotted path to a settings file. Then Socorro loads that into a `Settings` instance and you do stuff with it.

This is the settings for Mozilla's crash ingestion pipeline.

socorro/mozilla_settings.py
```diff
@@ -0,0 +1,309 @@
...
+            doc="S3 region for crash report data.",
+        ),
+        "endpoint_url": AWS_ENDPOINT_URL,
+    },
+}
+
+# Each key in this list corresponds to a key in this dict containing a crash report data
+# destination configuration
+CRASH_DESTINATIONS_ORDER = ["s3", "es", "telemetry"]
+CRASH_DESTINATIONS = {
```
willkg, line 188: This is how we specify crashstorage destinations--no more poly-storage stuff.

socorro/processor/cache_manager.py
```diff
@@ -0,0 +1,378 @@
...
+import sys
+import traceback
+
+from boltons.dictutils import OneToOne
+from fillmore.libsentry import set_up_sentry
+from fillmore.scrubber import Scrubber, SCRUB_RULES_DEFAULT
+from inotify_simple import INotify, flags
+import markus
+
+from socorro import settings
```
willkg, line 34: I thought about having this take command line arguments instead of using settings. I'm not sure which is the better approach, but none of the other services in Socorro take command line arguments (now) so I decided to have it work like them.

socorro/processor/cache_manager.py
```diff
@@ -0,0 +1,378 @@
...
+            | flags.DELETE_SELF
+            | flags.MODIFY
+            | flags.MOVED_FROM
+            | flags.MOVED_TO
+        )
+
+    def set_up(self):
+        set_up_logging(
+            local_dev_env=settings.LOCAL_DEV_ENV,
+            logging_level=settings.CACHE_MANAGER_LOGGING_LEVEL,
```
willkg, line 104: I have this use a different setting for the logging level because this can be pretty spammy when things are going and I'm pretty sure the typical scenario is one needs one of the services logging at DEBUG and not all of them.

socorro/processor/cache_manager.py
```diff
@@ -0,0 +1,378 @@
...
+            host_id=settings.HOST_ID,
+            before_send=scrubber,
+        )
+
+        self.log_config()
+
+        # Create the cachepath if we need to
+        self.cachepath.mkdir(parents=True, exist_ok=True)
+
+        print(f"test {self.cachepath = }")
```
willkg, line 125: Oops--this should get removed.

socorro/processor/cache_manager.py
```diff
@@ -0,0 +1,378 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
willkg, line 1: This was copied over from Eliot where it's been in production for a year. Then I updated it to work in Socorro.

socorro/processor/processor_app.py
```diff
@@ -37,167 +37,51 @@
...
+from socorro.external.crashstorage_base import CrashIDNotFound
+from socorro.libclass import build_instance, build_instance_from_settings
 from socorro.lib.libdatetime import isoformat_to_time
-from socorro.lib.libdockerflow import get_release_name
+from socorro.lib.libdockerflow import get_release_name, get_version_info
+from socorro.lib.liblogging import set_up_logging
 from socorro.lib.task_manager import respond_to_SIGTERM
 
 
-CONFIG_DEFAULTS = {
```
willkg, line 47: A lot of this goes away and the rest is in `socorro/mozilla_settings.py.

socorro/processor/pipeline.py
```diff
@@ -10,60 +10,13 @@
 """
```
willkg, line 10: I renamed this file from `socorro/processor/processor_pipeline.py` to `socorro/processor/pipeline.py` to take out duplicate information.

socorro/processor/processor_app.py
```diff
@@ -293,48 +177,56 @@ def process_crash(self, crash_id, ruleset_name, tmpdir):
...
-            # Capture and log the exceptions raised by storage backends
-            for storage_error in poly_storage_error:
-                sentry_sdk.capture_exception(storage_error)
-                self.logger.error("error: crash id %s: %r", crash_id, storage_error)
-
-            # Re-raise the original exception with the correct traceback
-            raise
+        # Save data to crash storage destinations
+        self.logger.debug("saving %s", crash_id)
+        for dest in self.destinations:
```
willkg, line 211: Previously, we had poly-storage and metrics counters and benchmarking wrappers. All that code goes away with this for loop here where it supports multiple destinations and handles the metrics emitting on its own.

socorro/processor/processor_app.py
```diff
@@ -349,49 +241,42 @@ def process_crash(self, crash_id, ruleset_name, tmpdir):
...
+            setattr(dest_obj, "crash_destination_name", key)
+            destinations.append(dest_obj)
+        self.destinations = destinations
+
+        self.pipeline = build_instance_from_settings(settings.PROCESSOR["pipeline"])
+
+        self.temporary_path = settings.PROCESSOR["temporary_path"]
         os.makedirs(self.temporary_path, exist_ok=True)
 
-        if self.config.companion_process.companion_class:
```
willkg, line 374: I took out all the companion process stuff. There are a few problems with the companion process. The first is that it's kicked off as another thread in the processor process. The second is that it's unmanaged--if the thread dies, the processor just blithely continues on. The third is that it emits no metrics or logging, so we have no idea what it's doing. Also, since the stackwalker populates the LRU cache and it's not emitting metrics or logging either, we have no idea how effective the LRU cache is.

Anyhow, so I removed all the companion process stuff and wrote a new disk cache manager that runs as a separate process, emits metrics and logging, and when it dies, it causes the entire processor to die.

socorro/processor/processor_app.py
```diff
@@ -349,49 +241,42 @@ def process_crash(self, crash_id, ruleset_name, tmpdir):
...
         respond_to_SIGTERM_with_logging = partial(respond_to_SIGTERM, target=self)
         signal.signal(signal.SIGTERM, respond_to_SIGTERM_with_logging)
 
-        self.task_manager = self.config.producer_consumer.producer_consumer_class(
-            self.config.producer_consumer,
-            job_source_iterator=self.source_iterator,
-            task_func=self.transform,
+        # Create task manager
+        manager_class = settings.PROCESSOR["task_manager"]["class"]
+        manager_settings = settings.PROCESSOR["task_manager"]["options"]
```
willkg, line 271: I thought about adding the ability to pass additional arguments to `build_instance_from_settings`, but I'm on the fence about it. We could use it here and another place.

socorro/tests/external/boto/test_connection_context.py
```diff
@@ -2,56 +2,66 @@
...
 class TestS3ConnectionContext:
-    def test_save_file(self, boto_helper):
+    def test_save_file(self, s3_helper):
         """Test saving a file and make sure it's there."""
-        config = get_config(cls=S3Connection)
-        conn = S3Connection(config)
+        conn = build_instance_from_settings(S3_SETTINGS)
 
-        bucket = conn.config.bucket_name
+        bucket = os.environ["CRASHSTORAGE_S3_BUCKET"]
```
willkg, line 29: I think `s3_helper` has a `get_crashstorage_bucket` method now. I'll update this.

socorro/tests/external/boto/test_crash_data.py
```diff
@@ -3,123 +3,143 @@
...
-        return SimplifiedCrashData(config=get_config(SimplifiedCrashData))
+    def test_get_processed(self, s3_helper):
+        crashdata = build_instance_from_settings(CRASHDATA_SETTINGS)
 
-    def test_get_processed(self, boto_helper):
-        boto_s3_store = self.get_s3_store()
-        bucket = boto_s3_store.conn.bucket
-        boto_helper.create_bucket(bucket)
+        bucket = CRASHDATA_SETTINGS["options"]["bucket"]
+        crash_id = "0bba929f-8721-460c-dead-a43c20071027"
```
willkg, line 43: These crash ids should be fresh.

socorro/tests/external/boto/test_crashstorage.py
```diff
@@ -57,26 +77,22 @@ def test_build_keys(kind, crashid, expected):
...
         # Run save_raw_crash
-        boto_s3_store.save_raw_crash(
-            {"submitted_timestamp": "2013-01-09T22:21:18.646733+00:00"},
+        crashstorage.save_raw_crash(
+            raw_crash={"submitted_timestamp": "2013-01-09T22:21:18.646733+00:00"},
             # This is an empty set of dumps--no dumps!
-            MemoryDumpsMapping(),
-            "0bba929f-8721-460c-dead-a43c20071027",
+            dumps=MemoryDumpsMapping(),
+            crash_id="0bba929f-8721-460c-dead-a43c20071027",
```
willkg, line 90: This crashid and submitted_timestamp should be fresh.

socorro/tests/external/es/test_analyzers.py
```diff
@@ -2,42 +2,56 @@
...
 
         That analyzer creates tokens (terms) by splitting the input on
         semicolons (;) only.
 
         """
+        crashstorage = self.build_crashstorage()
+        api = SuperSearch(crashstorage=crashstorage)
+        now = utc_now()
+        crash_id_1 = "936ce666-ff3b-4c7a-9674-367fe2230212"
+        crash_id_2 = "b39a6e4a-1680-4968-99bd-272286230212"
```
willkg, line 32: I think we need to use fresh crash ids here.

socorro/tests/external/es/test_connection_context.py
```diff
@@ -2,62 +2,62 @@
...
-    def test_create_index(self, es_conn):
-        # Delete any existing indices first
-        for index in es_conn.get_indices():
-            es_conn.delete_index(index)
-        es_conn.refresh()
-        es_conn.health_check()
+    def build_conn(self):
+        return build_instance(
+            class_path="socorro.external.es.connection_context.ConnectionContext",
+            kwargs=settings.CRASH_DESTINATIONS["es"]["options"],
```
willkg, line 14: One of the things I spent a lot of time thinking about was how to test everything. Should we be testing with the mozilla_settings? Should we create a second set of settings for testing? The latter is less fragile to how we change mozilla settings around over time. But the former makes it more likely that "if it passes in CI, it'll work in prod".

If we had more time / resources, we could have test_settings AND also test out the mozilla_settings. The latter probably done with integration tests.

However, I don't have time / resources to maintain all that, so for now, we're testing with the mozilla_settings with some minor adjustments (bucket names, index names) so that test settings don't stomp on local dev environment crash storage.

socorro/tests/external/es/test_connection_context.py
```diff
@@ -2,62 +2,62 @@
...
         index_name = now.strftime(template)
 
-        es_conn.create_index(index_name)
-        es_conn.health_check()
-        assert index_name in list(es_conn.get_indices())
-
-        es_conn.delete_index(index_name)
-        assert index_name not in list(es_conn.get_indices())
-
-    def test_delete_old_indices(self, es_conn):
```
willkg, line 39: "Old indices" relates to the retention policy which is no longer part of the connection context--it was moved to crashstorage code. This test was also moved.

socorro/tests/external/es/test_crashstorage.py
```diff
@@ -85,34 +90,35 @@ def test_invalid_key(self, key):
...
         processed_crash["date_processed"] = date_to_string(utc_now())
 
-        es_storage.save_processed_crash(
+        crashstorage = self.build_crashstorage()
+        crashstorage.save_processed_crash(
             raw_crash=raw_crash,
             processed_crash=processed_crash,
         )
 
         # Ensure that the document was indexed by attempting to retreive it.
```
willkg, line 114: Typo.

socorro/tests/external/es/test_crashstorage.py
```diff
@@ -149,34 +155,37 @@ def test_index_crash_indexable_keys(self):
...
-            index_name=self.es_context.get_index_for_date(four_weeks_ago),
+        crashstorage.create_index(
+            index_name=crashstorage.get_index_for_date(four_weeks_ago),
             mappings=mappings,
         )
 
-        es_storage = ESCrashStorage(config=self.config)
-
         # Create a crash for this week and save it
         now_uuid = "00000000-0000-0000-0000-000000120408"
```
willkg, line 190: This looks wrong.

socorro/tests/external/es/test_crashstorage.py
```diff
@@ -204,144 +213,91 @@ def test_index_crash_mapping_keys(self):
...
+            doc = conn.get(
+                index=crashstorage.get_index_for_date(four_weeks_ago),
+                id=old_uuid,
+            )
+            assert field not in doc["_source"]["processed_crash"]
 
 
-class TestESCrashStorage(ElasticsearchTestCase):
+class TestESCrashStorage:
     """These tests are self-contained and use Mock where necessary"""
```
willkg, line 243: I have no idea what this docstring means--we should remove it.

socorro/tests/external/es/test_crashstorage.py
```diff
@@ -204,144 +213,91 @@ def test_index_crash_mapping_keys(self):
...
+            kwargs=settings.CRASH_DESTINATIONS["es"]["options"],
         )
 
-        # Ensure that the indexing function is only called once.
-        assert es_storage._submit_crash_to_elasticsearch.call_count == 1
-
-    @mock.patch("socorro.external.es.connection_context.elasticsearch")
-    def test_success(self, espy_mock):
+    def test_indexing_success(self, es_helper):
         """Test a successful index of a crash report"""
```
willkg, line 252: I think this has already been tested. We should merge this with the integration test above.

socorro/tests/external/es/test_crashstorage.py
```diff
@@ -204,144 +213,91 @@ def test_index_crash_mapping_keys(self):
...
 
-        additional = {
-            "doc_type": "crash_reports",
-            "id": crash_id,
-            "index": self.es_context.get_index_for_date(utc_now()),
-        }
-
-        sub_mock.index.assert_called_with(body=document, **additional)
-
-    @mock.patch("socorro.external.es.connection_context.elasticsearch")
```
willkg, line 308: This didn't test anything real because everything was mocked out so I removed it.

socorro/tests/external/es/test_crashstorage.py
```diff
@@ -378,16 +333,111 @@ def test_index_data_capture(self):
...
+                "processed_crash.memory_explicit",
+                9_223_372_036_854_775_999,
+                REMOVED_VALUE,
+            ),
+            # Booleans are converted
+            # FIXME(willkg): fix_boolean is never called--that's wrong
+            # ("processed_crash.accessibility", "true", True),
+        ],
+    )
+    def test_indexing_bad_data(self, key, value, expected_value, es_helper):
```
willkg, line 401: This is a much better version of tests we had previously that only tested fixing data.

socorro/tests/external/es/test_crashstorage.py
```diff
@@ -378,16 +333,111 @@ def test_index_data_capture(self):
...
+                REMOVED_VALUE,
+            ),
+            # Out-of-bounds longs are removed
+            (
+                "processed_crash.memory_explicit",
+                9_223_372_036_854_775_999,
+                REMOVED_VALUE,
+            ),
+            # Booleans are converted
+            # FIXME(willkg): fix_boolean is never called--that's wrong
```
willkg, line 397: When writing this, I discovered `fix_boolean` is never called. Fixing that is a new bug because it may have consequences on indexing and searching across indexes. I'll look into it later.

socorro/tests/external/es/test_crashstorage.py
```diff
@@ -378,16 +333,111 @@ def test_index_data_capture(self):
...
     "value, expected",
     [
         # Non-string values are converted to "BAD DATA"
         (1, "BAD DATA"),
-        # String values are truncated if > 10,000 characters
-        ("a" * 9_999, "a" * 9_999),
-        ("a" * 10_000, "a" * 10_000),
-        ("a" * 10_001, "a" * 10_000),
+        # Keyword string values are truncated if > 10,000 characters
+        pytest.param("a" * 9_999, "a" * 9_999, id="keyword_9_999"),
```
willkg, line 438: I learned about using `pytest.param` to specify the test id and it fixes these situations where an id with 9,999 characters is unusable.

socorro/tests/external/es/test_supersearch.py
```diff
@@ -23,21 +27,20 @@ def get(self, **kwargs):
...
 
-        # Wait until the cluster is yellow before proceeding.
-        self.health_check()
+    def test_get_indices(self, es_helper):
+        crashstorage = self.build_crashstorage()
+        api = SuperSearchWithFields(crashstorage=crashstorage)
 
-    def test_get_indices(self):
-        now = datetime.datetime(2001, 1, 2, 0, 0)
+        now = datetime.datetime(2021, 1, 2, 0, 0)
```
willkg, line 43: This is poorly named, but I'm going to leave it for now.

socorro/tests/external/fs/test_fspermanentstorage.py
```diff
@@ -3,117 +3,117 @@
...
 from socorro.external.crashstorage_base import CrashIDNotFound, MemoryDumpsMapping
 from socorro.external.fs.crashstorage import FSPermanentStorage
 
 
-FS_ROOT = os.environ["resource.fs.fs_root"]
+class TestFSPermanentStorage:
+    def test_save_raw_crash(self, tmp_path):
+        fs = FSPermanentStorage(fs_root=str(tmp_path))
 
+        crash_id = "0bba929f-8721-460c-dead-a43c20071025"
```
willkg, line 17: This should be a fresh crash id.

webapp/crashstats/supersearch/models.py
```diff
@@ -86,8 +86,6 @@ def validate_products(product_value):
 
 
 class SuperSearch(ESSocorroMiddleware):
-    implementation = supersearch.SuperSearch
```
willkg, line 89: Previously, you could define the `implementation` class attribute and the SocorroMiddleware would instantiate the specified class passing in the configman configuration and all that. We no longer have configman configuration, so now each middleware class needs to supply its own `get_implementation` method which configures the socorro class it's using (e.g. ESCrashStorage, SuperSearch, etc).

I changed the code that creates the implementation so it only looked at the `get_implementation` method. Then I went through all the middleware classes and removed `implementation` class attributes replacing them with the appropriate `get_implementation` code.

webapp/crashstats/supersearch/models.py
```diff
@@ -118,6 +116,12 @@ def __init__(self):
             + tuple(self.extended_fields)
         )
 
+    def get_implementation(self):
+        s3_crash_dest = build_instance_from_settings(
+            socorro_settings.CRASH_DESTINATIONS["es"]
```
willkg, line 121: We can change this to `socorro_settings.ES_STORAGE` now.

webapp/crashstats/supersearch/views.py
```diff
@@ -284,20 +286,20 @@ def search_custom(request, default_context=None):
...
             error = e
 
-    schema = settings.ELASTICSEARCH_INDEX_SCHEMA
-    now = timezone.now()
+    es_crashstorage = build_instance_from_settings(
+        socorro_settings.CRASH_DESTINATIONS["es"]
+    )
 
     possible_indices = []
-    for i in range(26):
```
willkg, line 291: I reworked this code so it calls `get_indices` in the crash storage instance. That way things like the retention policy, how index names are generated, etc are all in the crash storage class and not spread across the code base.

webapp/crashstats/supersearch/views.py
```diff
@@ -284,20 +286,20 @@ def search_custom(request, default_context=None):
...
+    for index in es_crashstorage.get_indices():
         possible_indices.append({"id": index, "text": index})
 
     context = default_context
     context["elasticsearch_indices"] = possible_indices
 
     if query:
         context["query"] = json.dumps(query["query"])
-        context["indices"] = ",".join(query["indices"])
+        context["indices"] = ",".join(sorted(query["indices"]))
```
willkg, line 302: I added a `sorted` thing here to stabilize the output so tests passed consistently.

webapp/crashstats/crashstats/models.py
```diff
@@ -550,16 +534,21 @@ class takes.
...
     put = None
     delete = None
 
+    def get_implementation(self):
+        # FIXME(willkg): change this to BotoS3CrashStorage
+        s3_settings = socorro_settings.CRASH_DESTINATIONS["telemetry"]["options"]
+        return TelemetryCrashData(**s3_settings)
+
+        telemetry_settings = socorro_settings.CRASH_DESTINATIONS["telemetry"]
+        return build_instance_from_settings(telemetry_settings)
```
willkg, line 550: Oops--this is vestigial code.

webapp/crashstats/api/tests/test_views.py
```diff
@@ -196,7 +192,10 @@ def test_hit_or_not_hit_ratelimit(self):
             # eventually you'll run into the limit there too.
             assert response.status_code == 429
 
-    def test_ProcessedCrash(self):
+
+class TestProcessedCrashAPI(BaseTestViews):
+    @mock.patch("crashstats.crashstats.models.ProcessedCrash.get_implementation")
```
willkg, line 197: I took out the code that mocks all the implementations out before the tests run. I removed a vast majority of the mocking, but left this one in. Because of that, we have to add the mocking as a decorator.

socorro/tests/processor/test_cache_manager.py
```diff
@@ -0,0 +1,397 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
willkg, line 1: The cache manager code and its tests were pulled in from Eliot which has been running it in production for over a year now. Having said that, the Socorro minidump stackwalker saves files in a deep directory tree, but Eliot saves all its files in the same directory--no subdirectories. The code for reacting to new directories (which Eliot didn't use) was buggy.

Thus this version and its tests is just like the Eliot version, but with the support for directory trees fixed.

socorro/tests/processor/test_cache_manager.py
```diff
@@ -0,0 +1,397 @@
...
+        lru["key1"] = 1
+        lru["key2"] = 2
+
+        oldest = lru.pop_oldest()
+        assert oldest == ("key1", 1)
+        assert list(lru.items()) == [("key2", 2)]
+
+
+@pytest.fixture
+def cm(tmp_path):
```
willkg, line 52: "cm" stands for "cache manager". We use "cache_manager" in some places and I needed a shorter name for things, but I'm not thrilled about where I ended up with naming things.

socorro/tests/processor/test_cache_manager.py
```diff
@@ -0,0 +1,397 @@
...
+    """Test cache manager setup with tmp_path."""
+    with settings.override(
+        SYMBOLS_CACHE_PATH=str(tmp_path),
+        SYMBOLS_CACHE_MAX_SIZE=10,
+    ):
+        cache_manager = DiskCacheManager()
```
willkg, line 58: I thought about this a bunch. Should the `DiskCacheManager` class take these as arguments or should they be pulled in from settings? I decided to leave it as settings since we're testing the process which pulls from settings.

socorro/tests/processor/test_processor_app.py
```diff
@@ -25,89 +22,74 @@ def foo(*fargs, **fkwargs):
...
-    config.queue.crashqueue_class = FakeCrashQueue
 
-    config.companion_process = DotDict()
-    mocked_companion_process = mock.Mock()
-    config.companion_process.companion_class = mock.Mock(
-        return_value=mocked_companion_process
-    )
-
-    return config
+class TestProcessorApp:
```
willkg, line 55: We can probably remove the mocking in these tests since we're using the `InMemoryCrashStorage` and can control what it's got in it.

webapp/crashstats/topcrashers/tests/test_views.py
```diff
@@ -12,259 +12,284 @@
...
-            bug_id=22222, signature="FakeSignature1 \u7684 Japanese"
-        )
-        BugAssociation.objects.create(
-            bug_id=33333, signature="FakeSignature1 \u7684 Japanese"
-        )
+    def test_topcrashers_first_appearance(self, client, db, es_helper):
+        signature = "FakeSignature1"
 
         Signature.objects.create(
-            signature="FakeSignature1 \u7684 Japanese",
```
willkg, line 32: We don't allow non-ascii characters in signatures, so this is weird to test.

webapp/crashstats/topcrashers/tests/test_views.py
```diff
@@ -12,259 +12,284 @@
...
+            first_date=datetime.datetime(2021, 1, 1, 12, 23, 34, tzinfo=utc),
             first_build="20000101122334",
         )
-        Signature.objects.create(
-            signature="mozCool()",
-            first_date=datetime.datetime(2016, 5, 2, 0, 0, 0, tzinfo=utc),
-            first_build="20160502000000",
-        )
 
-        def mocked_supersearch_get(**params):
```
willkg, line 42: This mocked output was wrong in a bunch of weird ways, so it's not clear what these tests were telling us about correctness in the system.

webapp/crashstats/topcrashers/tests/test_views.py
```diff
@@ -12,259 +12,284 @@
...
+            url,
+            {"product": "Firefox", "version": "1.0", "_facets_size": "100"},
         )
         assert response.status_code == 200
         doc = pyquery.PyQuery(response.content)
         selected_count = doc('.tc-result-count a[class="selected"]')
         assert selected_count.text() == "100"
 
-        print(smart_str(response.content))
+    def test_topcrashers_startup_crash(self, client, db, es_helper):
```
willkg, line 123: I rewrote this test to verify all possibilities of the startup crash icons.

webapp/crashstats/supersearch/tests/test_views.py
```diff
@@ -15,31 +16,17 @@
...
-
-class TestViews(BaseTestViews):
-    def setUp(self):
-        super().setUp()
-        # Mock get_versions_for_product() so it doesn't hit supersearch breaking the
-        # supersearch mocking
-        self.mock_gvfp = mock.patch(
-            "crashstats.crashstats.utils.get_versions_for_product"
-        )
-        self.mock_gvfp.return_value = ["20.0", "19.1", "19.0", "18.0"]
```
willkg, line 31: I took out the mock for featured versions. Because of that, we now have to index a sufficient amount of crash report data to coerce the featured version code to return the values we're looking for.

willkg: Also, we have to use real supported products--we can't use Waterwolf and other fake products.

webapp/crashstats/supersearch/tests/test_views.py
```diff
@@ -105,209 +87,141 @@ def test_search_fields_metrics(self):
...
-    def test_search_results(self):
+    def test_search_results(self, client, db, es_helper):
         BugAssociation.objects.create(
             bug_id=123456, signature="nsASDOMWindowEnumerator::GetNext()"
         )
 
-        def mocked_supersearch_get(**params):
-            assert "_columns" in params
-
-            if "WaterWolf" in params.get("product", []):
```
willkg, line 116: These results were all mocked out and were wrong in a lots of ways. Also, when we go to update Elasticsearch, these tests would continue to pass even if the API changed which is deeply unhelpful.

webapp/crashstats/supersearch/tests/test_views.py
```diff
@@ -105,209 +87,141 @@ def test_search_fields_metrics(self):
...
-                return results
-
-            else:
-                return {"hits": [], "facets": [], "total": 0}
-
-        SuperSearchUnredacted.implementation().get.side_effect = mocked_supersearch_get
+        def build_crash_data(**params):
+            crash_id = create_new_ooid()
+            data = {
+                "date_processed": utc_now(),
```
willkg, line 98: Elasticsearch crash storage is date-sensitive. When redoing all these tests and taking the mocks out, I switched to using fresh crash ids. The `date_processed` is related to the crash id in that they're usually the same day, so I'm basing it on the same timestamp.

webapp/crashstats/supersearch/tests/test_views.py
```diff
@@ -105,209 +87,141 @@ def test_search_fields_metrics(self):
...
-
-        SuperSearchUnredacted.implementation().get.side_effect = mocked_supersearch_get
-
+    def test_search_results_metrics(self, client, db, es_helper):
         url = reverse("supersearch:search_results")
         with MetricsMock() as metrics_mock:
-            response = self.client.get(url, {"product": "WaterWolf"})
+            response = client.get(url, {"product": "Firefox"})
         assert response.status_code == 200
         metrics_mock.assert_timing(
```
willkg, line 226: This should be indented because it should be run in the same context.

webapp/crashstats/supersearch/tests/test_views.py
```diff
@@ -630,48 +469,44 @@ def mocked_supersearch_get(**params):
...
+            now.strftime(template),
+        ]
+        assert ",".join(indices) in smart_str(response.content)
+
+    def test_search_query(self, client, db, es_helper, user_helper):
+        user = user_helper.create_protected_plus_user()
+        client.force_login(user)
 
         url = reverse("supersearch:search_query")
-        response = self.client.post(url, {"query": '{"query": {}}'})
```
willkg, line 667: This is not a valid Elasticsearch query but since it was all mocked out, the test never failed.

webapp/crashstats/signature/tests/test_views.py
```diff
@@ -2,506 +2,524 @@
...
-    def setUp(self):
-        super().setUp()
-        # Mock get_versions_for_product() so it doesn't hit supersearch breaking the
-        # supersearch mocking
-        self.mock_gvfp = mock.patch(
-            "crashstats.crashstats.utils.get_versions_for_product"
-        )
-        self.mock_gvfp.return_value = ["20.0", "19.1", "19.0", "18.0"]
-        self.mock_gvfp.start()
+def get_date_range(crash_id=None, date=None):
```
willkg, line 23: The supersearch only returns items with a certain date range, so we need to make sure the requests made by the tests use a date range that works for the crash reports it indexed. This function makes that easier.

webapp/crashstats/signature/tests/test_views.py
```diff
@@ -2,506 +2,524 @@
...
         # Test with no results.
         url = reverse("signature:signature_reports")
-        response = self.client.get(
-            url, {"signature": DUMB_SIGNATURE, "date": "2012-01-01"}
+        response = client.get(
+            url,
+            {
+                "signature": "wrong signature",
+                "product": "Firefox",
+                "date": [f">={start_date}", f"<{end_date}"],
```
willkg, line 87: Now that we're not mocking everything out, we discover that passing dates is important, so this adds dates to all the supersearch filters.

webapp/crashstats/signature/tests/test_views.py
```diff
@@ -2,506 +2,524 @@
...
+            url = reverse("signature:signature_reports")
+            response = client.get(
+                url,
+                {
+                    "signature": TEST_SIGNATURE,
+                    "product": ["Firefox", "Thunderbird"],
+                    "address": ["0x0", "0xa"],
+                    "reason": ["^hello", "$thanks"],
+                    "date": [f">={date}"],
+                    "java_stack_trace": "Exception",
```
willkg, line 168: This is a contrived impossible scenario since we'll never have a Firefox or Thunderbird crash report with `java_stack_trace`.

webapp/crashstats/signature/tests/test_views.py
```diff
@@ -2,506 +2,524 @@
...
-        assert len(struct["aggregates"]) == 2
+        assert len(struct["aggregates"]) == 3
+
+        # We have three kinds of products: Firefox, Fenix, Thunderbird
         assert "term_counts" in struct
-        assert len(struct["term_counts"]) == 1
+        assert len(struct["term_counts"]) == 3
 
-    def test_signature_comments_no_permission(self):
+    def test_signature_comments_no_permission(self, client, db):
```
willkg, line 374: This test is effectively done twice. We could take out this one.

webapp/crashstats/crashstats/tests/test_views.py
```diff
@@ -72,6 +77,36 @@
...
+    processed_crash["date_processed"] = date_to_string(date_from_ooid(crash_id))
+    return crash_id, raw_crash, processed_crash
+
+
+def upload_crash_data(s3_helper, raw_crash, processed_crash):
+    """Validate crash data and upload it to crash bucket"""
+    crash_id = processed_crash["uuid"]
+    bucket = s3_helper.get_crashstorage_bucket()
+
+    validate_instance(raw_crash, RAW_CRASH_SCHEMA)
```
willkg, line 95: This guarantees the data we're building is valid raw crash data. This is pretty important when testing the webapp.

webapp/crashstats/crashstats/tests/test_views.py
```diff
@@ -72,6 +77,36 @@
...
+    raw_key = build_keys("raw_crash", crash_id)[0]
+    s3_helper.upload_fileobj(
+        bucket_name=bucket, key=raw_key, data=dict_to_str(raw_crash).encode("utf-8")
+    )
+
+    validate_instance(processed_crash, PROCESSED_CRASH_SCHEMA)
```
willkg, line 101: This guarantees the data we're building is valid processed crash data. This is pretty important when testing the webapp.

webapp/crashstats/crashstats/tests/test_views.py
```diff
@@ -438,15 +458,13 @@ def mocked_processed_crash_get(**params):
         assert "OS X 10.11" in content
 
         # the protected data will appear if we log in
-        user = self._login()
-        group = self._create_group_with_permission("view_pii")
-        user.groups.add(group)
-        assert user.has_perm("crashstats.view_pii")
+        user = user_helper.create_protected_user()
```
willkg, line 461: The `user_helper` makes creating users with appropriate permissions much easier.

willkg:
The configuration changes landed in the infra repo. I landed a PR with some minor dependency updates to trigger an autodeploy to stage to make sure the configuration changes work with the existing codebase. Stage autodeployed, I did some brief testing, and everything looks ok.

Next step is to land this and have it autodeploy to stage and then begin the process of validating everything on stage. As we find issues, we'll fix them. I'll update the dashboards as part of validation.

PR #6348, opened 2023-02-07
bug 1815535: remove raw crash path with entropy from verifyprocessed

willkg:
We phased out raw crash paths with entropy in September, so verifyprocessed doesn't need to check those anymore.

The remaining bits of entropy stuff are in the boto crash storage code. We need to keep that there until March 7th, 2023 when all the old data has expired out.

### DIFF
--- a/webapp/crashstats/crashstats/tests/test_verifyprocessed.py
+++ b/webapp/crashstats/crashstats/tests/test_verifyprocessed.py
@@ -64,13 +64,13 @@ class TestVerifyProcessed:
 
     def test_get_threechars(self):
         cmd = Command()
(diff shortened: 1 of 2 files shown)


webapp/crashstats/crashstats/tests/test_verifyprocessed.py
```diff
@@ -64,13 +64,13 @@ def create_processed_crash_in_es(self, es_conn, crash_id):
 
     def test_get_threechars(self):
         cmd = Command()
-        entropy = list(sorted(cmd.get_threechars()))
+        threefhars = list(sorted(cmd.get_threechars()))
```
willkg, line 67: Oops!

PR #6350, opened 2023-02-13
bug 1529342: remove tests and fix docstrings for things that will be changed

willkg:
I've got a PR that's too big. I'm taking out some parts from that PR and landing them on their own. In the future PR, I'm removing the PolyStorage crashstorage, MetricsCounter, the benchmarking wrapper, and cleaning up the task manager code.

This removes the tests for those things and fixes the doc strings in the task manager code.

This doesn't affect the run-time behavior of webapp, processor, or crontabber. There are no functionality/behavior changes in this PR.


socorro/tests/external/es/test_crashstorage.py
```diff
@@ -336,245 +328,6 @@ def test_fatal_failure(self, espy_mock):
                 crash_id=crash_id,
             )
 
-    @mock.patch("elasticsearch.client")
```
willkg, line 339: These tests mock everything out and hard-code messages that Elasticsearch 1.4 returns. What happens when we update to a later version of Elasticsearch? Does this help us? _shrug_

Further, when we hit problems with indexing, we really want to write code to fix things so the problems _don't_ occur. 

Having these tests doesn't help us.

socorro/tests/external/test_crashstorage_base.py
```diff
@@ -150,204 +145,6 @@ def test_basic_crashstorage(self):
 
             crashstorage.close()
 
-    def test_polyerror(self):
```
willkg, line 153: Instead of having a poly-storage thing, we're going to move to a "processing supports multiple destinations" model. It's much easier to implement, it doesn't require additional poly-storage code, and it removes the weirdness around polystorageerror which is hard to debug in the logs and in sentry.

socorro/tests/external/test_crashstorage_base.py
```diff
@@ -150,204 +145,6 @@ def test_basic_crashstorage(self):
...
-
-            poly_store.stores["B"].close.side_effect = Exception
-            with pytest.raises(PolyStorageError):
-                poly_store.close()
-
-            for v in poly_store.stores.values():
-                v.close.assert_called_with()
-
-
-class TestBench:
```
willkg, line 281: We only use these benchmarking wrappers to capture the `save_processed_crash` timings. Instead of having benchmarking wrappers, we're going to add timings capture code to the processor so we're capturing the things we want to capture there without the need of wrapper classes.

socorro/tests/external/test_crashstorage_base.py
```diff
@@ -358,51 +155,3 @@ def test_simple(self):
         fdm = mdm.as_file_dumps_mapping("a", "/tmp", "dump")
         assert fdm.as_file_dumps_mapping() is fdm
         assert fdm.as_memory_dumps_mapping() == mdm
-
-
-class TestMetricsCounter:
```
willkg, line 363: This metrics counter acts as a crashstorage destination. Instead of the complexity around that, it's easier to add a `METRICS.incr` call in the processor.

PR #6372, opened 2023-03-15
bug 1822648: improve filenotfound resiliency and switch to &

willkg:
This switches event checking to use & (it's a bitmask) instead of "in" which iterates over the list. & should be a little faster even if the lists are short.

This improves other cases where the file might not exist when we check the size and handles those cases appropriately.

### DIFF
--- a/socorro/processor/cache_manager.py
+++ b/socorro/processor/cache_manager.py
@@ -222,37 +222,39 @@ class DiskCacheManager:
                 try:
                     for event in self.inotify.read(timeout=timeout):
                         processed_events = True
-                        event_flags = flags.from_mask(event.mask)
+                        event_mask = event.mask
 
-                        flags_list = ", ".join([str(flag) for flag in event_flags])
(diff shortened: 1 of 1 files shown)


socorro/processor/cache_manager.py
```diff
@@ -222,37 +222,39 @@ def _event_generator(self, nonblocking=False):
...
                         processed_events = True
-                        event_flags = flags.from_mask(event.mask)
+                        event_mask = event.mask
 
-                        flags_list = ", ".join([str(flag) for flag in event_flags])
-                        self.logger.debug("EVENT: %s: %s", event, flags_list)
+                        # flags_list = ", ".join(
+                        #     [str(flag) for flag in flags.from_mask(event_mask)]
+                        # )
+                        # self.logger.debug("EVENT: %s: %s", event, flags_list)
```
willkg, line 230: This is some extra work and the loop shouldn't be doing this work _every_ event if it's not going to be logged. This is only helpful for development. Ergo I commented it out.

PR #6374, opened 2023-03-17
Bug 1822931 - Fix constant reference qualifiers in signatures

gabrielesvelto:
Fix processing functions with constant reference qualifiers that have a space between the `const` and `&` symbols. Note that I didn't drop the `const&` version because I'm not 100% sure that it's wrong or unused.


willkg:
Thank you for doing this!

PR #6375, opened 2023-03-17
bug 1529342: switch from ACCESS to OPEN; minor improvements

willkg:
This moves debugging logging to an `is_verbose` conditional block which reduces some work in the hot loop.

More importantly, this switches from using ACCESS to using OPEN to update the access time in the LRU. There's one OPEN but multiple ACCESSes when using a file, so this will radically reduce the number of inotify events being handled.

This also wraps another os operation in a FileNotFoundError check.

### DIFF
--- a/socorro/processor/cache_manager.py
+++ b/socorro/processor/cache_manager.py
@@ -3,13 +3,15 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 """
-This defines the DiskCacheManager application. It's designed to run as
-a standalone application separate from the rest of the processor.
+This defines the DiskCacheManager application. It's designed to run as a standalone
+application separate from the rest of the processor.
 
-It keeps track of files in a directory and evicts files least recently
-used in order to keep the total size under a max number.
+It keeps track of files in a directory and evicts files least recently used in order to
+keep the total size under a max number.
 
 It uses inotify to cheaply watch the files.
+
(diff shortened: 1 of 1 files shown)


socorro/processor/cache_manager.py
```diff
@@ -91,15 +93,19 @@ def __init__(self):
         self._generator = None
         self.inotify = None
         self.watch_flags = (
-            flags.ACCESS
```
willkg, line 94: I should have used OPEN instead of ACCESS originally. When a file is used, there's (usually) one OPEN and then a bazillion ACCESS events, so it floods the inotify queue which is why we got a few IN_Q_OVERFLOW events earlier. This should fix most/all the issues we're having now.

socorro/processor/cache_manager.py
```diff
@@ -91,15 +93,19 @@ def __init__(self):
...
             | flags.DELETE_SELF
             | flags.MODIFY
             | flags.MOVED_FROM
             | flags.MOVED_TO
+            | flags.OPEN
         )
 
+        # If the logging level is DEBUG, then we'll treat that as meaning we should be
+        # in verbose mode and emit additional run information
+        self.is_verbose = settings.CACHE_MANAGER_LOGGING_LEVEL == "DEBUG"
```
willkg, line 107: This makes it easier to keep the debugging output, but minimize the work when it's not being emitted.

PR #6376, opened 2023-03-20
bug 1529342: make adding watches more resilient to errors

willkg:
This adds some additional error handling around inotify calls to handle the case where it goes to add a watch, but the directory is no longer there.

### DIFF
--- a/socorro/processor/cache_manager.py
+++ b/socorro/processor/cache_manager.py
@@ -146,6 +146,15 @@ class DiskCacheManager:
         settings.log_settings(logger=self.logger)
 
     def add_watch(self, path):
+        """Add a watch
+
+        :arg path: the absolute path to the directory to add a watch to
+
+        :raises OSError: if inotify was unable to add the watch, possibly because
+            the file does not exist
+
+        """
+        path = str(path)
         if path not in self.watches:
             wd = self.inotify.add_watch(path, self.watch_flags)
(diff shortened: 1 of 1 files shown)


socorro/processor/cache_manager.py
```diff
@@ -240,7 +267,17 @@ def _event_generator(self, nonblocking=False):
                             flags_list = ", ".join(
                                 [str(flag) for flag in flags.from_mask(event_mask)]
                             )
-                            logger.debug("EVENT: %s: %s", event, flags_list)
+                            if event.wd > 0:
+                                dir_path = self.watches.inv[event.wd]
+                            else:
+                                dir_path = ""
```
willkg, line 273: This makes it easier to read the EVENT data in the logs. Previously, you'd get an event.name for a directory like `""` and you had to guess at what that was. Now it shows the full path.

socorro/processor/cache_manager.py
```diff
@@ -333,10 +376,10 @@ def _event_generator(self, nonblocking=False):
 
                             else:
                                 if is_verbose:
-                                    logger.debug("ignored %s %s", path, event)
+                                    logger.debug("unhandled event: %s %s", path, event)
```
willkg, line 379: "ignored" has a specific event meaning, so I changed the msg here to be "unhandled event".

PR #6381, opened 2023-03-24
bug 1529342: fix possible race condition when adding directories

willkg:
We keep running out of space in a way that suggests the cache manager somehow misses adding a watch for a top-level directory.

The theory is that if multiple stackwalkers are adding subdirectories, the cache manager would handle a CREATE | ISDIR event, but because inventory_existing does a directory traversal which takes a while to run, it's possible that an additional directory gets added, but missed by the traversal. That missed directory would never get watched and it will accrue data that isn't accounted for in the LRU cache until it eats up all the disk.

This changes the handling for CREATE | ISDIR to stop using inventory_existing and instead create a series of fake inotify Events just for the contents of the directory created. The fake events will get handled by the loop and cover adding watches for new directories and handling new files. The smaller focus should reduce the likelihood of a race condition causing a missed watch.

### DIFF
--- a/socorro/processor/cache_manager.py
+++ b/socorro/processor/cache_manager.py
@@ -30,7 +30,7 @@ import traceback
 from boltons.dictutils import OneToOne
 from fillmore.libsentry import set_up_sentry
 from fillmore.scrubber import Scrubber, SCRUB_RULES_DEFAULT
-from inotify_simple import INotify, flags
+from inotify_simple import INotify, flags, Event
 import markus
 
 from socorro import settings
@@ -150,6 +150,8 @@ class DiskCacheManager:
 
         :arg path: the absolute path to the directory to add a watch to
 
+        :returns: the watch descriptor for the path
+
         :raises OSError: if inotify was unable to add the watch, possibly because
             the file does not exist
 
@@ -158,6 +160,7 @@ class DiskCacheManager:
         if path not in self.watches:
             wd = self.inotify.add_watch(path, self.watch_flags)
             self.watches[path] = wd
+        return self.watches[path]
 
     def remove_watch(self, path):
         """Remove a watch
@@ -259,7 +262,10 @@ class DiskCacheManager:
         try:
             while self.running:
                 try:
-                    for event in self.inotify.read(timeout=timeout):
+                    events = self.inotify.read(timeout=timeout)
+                    while events:
+                        event = events.pop(0)
+
                         processed_events = True
                         event_mask = event.mask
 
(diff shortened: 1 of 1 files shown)


socorro/processor/cache_manager.py
```diff
@@ -217,11 +220,21 @@ def make_room(self, size):
             total_size -= rm_size
             removed += rm_size
             try:
+                # Delete the evicted file
                 os.remove(rm_path)
             except FileNotFoundError:
                 # If there was an OSError, then this file is gone already. We need to
                 # update our bookkeping, so continue.
                 pass
+            try:
```
willkg, line 229: This implements pruning empty directories which will in turn kick up DELETE | ISDIR events which will remove the corresponding watch. That'll prevent the cache manager from running out of inotify watches.

socorro/processor/cache_manager.py
```diff
@@ -217,11 +220,21 @@ def make_room(self, size):
...
+                # Attempt to prune empty directories. This will trigger DELETE | ISDIR
+                # events and get cleaned up by the event loop.
+                os.removedirs(os.path.dirname(rm_path))
+            except OSError as exc:
+                if self.is_verbose:
+                    self.logger.debug("unable to remove directory: %s %s", rm_path, exc)
```
willkg, line 235: I can take this out. It's almost assuredly because there's stuff in the directory.

socorro/processor/cache_manager.py
```diff
@@ -259,7 +270,10 @@ def _event_generator(self, nonblocking=False):
         try:
             while self.running:
                 try:
-                    for event in self.inotify.read(timeout=timeout):
+                    events = self.inotify.read(timeout=timeout)
```
willkg, line 273: The `inotify.read` docs say it returns a generator, but that's a lie--it's returning a list. If they change it to return a generator, calling `.pop` and `.insert` on it will fail and we'll know immediately in CI.

socorro/processor/cache_manager.py
```diff
@@ -293,15 +307,45 @@ def _event_generator(self, nonblocking=False):
                             # Handle directory events which update our watch lists
                             if flags.CREATE & event_mask:
                                 try:
-                                    self.add_watch(path)
+                                    created_wd = self.add_watch(path)
                                 except OSError:
+                                    logger.exception("add watch error for %s", path)
```
willkg, line 312: Adding a watch should never silently fail--that was a mistake. :(

socorro/processor/cache_manager.py
```diff
@@ -293,15 +307,45 @@ def _event_generator(self, nonblocking=False):
...
                                     continue
 
-                                # This is a new directory to watch, so we should add the
-                                # contents to our LRU and watches
-                                self.inventory_existing(path)
+                                # This is a new directory to watch, so we add faked
+                                # events for subdirectories and files
+                                for name in os.listdir(path):
```
willkg, line 317: We do a shallow directory listing because we're going to add events that, when handled, will cover sub directories.

PR #6398, opened 2023-04-19
bug 1816751: support product details version files for featured versions

willkg:
This adds support for specifying a product details version file (https://product-details.mozilla.org/1.0/mobile_versions.json) and which fields in that file to include in featured versions.

This switches Fenix to use that rather than auto-calculate it from crash data.


willkg:
This is hard to test in my dev environment, so I'm landing this now and will test it on stage.

PR #6408, opened 2023-05-11
bug 1827416, 1827495: update stackwalker to v0.16.0; add support for adjusted_address

willkg:
This updates the stackwalker from v0.15.0 to v0.16.0.

This fixes the handling for `cpu_microcode_version` since it changed from a u32 to a hexcode and the stackwalker now handles looking at the `CPUMicrocodeVersion` crash annotation if the minidump doesn't have the data.

This adds support for `crash_info.adjusted_address` and using that information to determine the final `address` in the processed crash.

### DIFF
--- a/socorro/processor/rules/breakpad.py
+++ b/socorro/processor/rules/breakpad.py
@@ -34,22 +34,48 @@ class CrashingThreadInfoRule(Rule):
         return processed_crash.get("json_dump", None) is not None
 
     def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
-        processed_crash["crashing_thread"] = glom.glom(
-            processed_crash, "json_dump.crash_info.crashing_thread", default=None
-        )
-        if processed_crash["crashing_thread"] is None:
+        crash_info = glom.glom(processed_crash, "json_dump.crash_info", default={})
+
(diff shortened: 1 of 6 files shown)


socorro/processor/rules/breakpad.py
```diff
@@ -34,22 +34,48 @@ def predicate(self, raw_crash, dumps, processed_crash, tmpdir, status):
...
+        # crash_info.adjusted_address.kind field is non-canonical then use
+        # crash_info.adjusted_address.address instead of crash_info.address for the
+        # crashing address.
+        address = crash_info.get("address")
+        adjusted_address = crash_info.get("adjusted_address")
+
+        if adjusted_address is not None:
+            kind = adjusted_address.get("kind")
+            if kind == "null-pointer":
+                # Infer the width from the previous address value
```
willkg, line 69: @gabrielesvelto Does this look ok to you for inferring the width? In this case, it's the difference between `0x00000000` and `0x0000000000000000` which feels cosmetic to me, so if this is good enough, we don't need to do more.

gabrielesvelto: Yes, this is fine!

PR #6415, opened 2023-06-13
bug 1837409: fix create-a-bug links for Fenix Java crashes

willkg:
The JavaStackTrace annotation is a long vaguely structured string. The create-a-bug code uses that to build the bug description. If that value is too long, it exceeds the max url length and then gets clipped causing the create-a-bug links to not work.

This truncates the JavaStackTrace value.

At some point, we should look at switching to JavaException which is structured when it's available and fall back to JavaStackTrace if necessary.

### DIFF
--- a/webapp/crashstats/crashstats/jinja2/crashstats/bug_comment.txt
+++ b/webapp/crashstats/crashstats/jinja2/crashstats/bug_comment.txt
@@ -6,7 +6,7 @@ Crash report: {{ full_url(request, "crashstats:report_index", crash_id=uuid ) }}
 {% if java_stack_trace %}
 Java stack trace:
 ```
-{{ java_stack_trace|safe }}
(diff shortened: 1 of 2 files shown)


webapp/crashstats/crashstats/jinja2/crashstats/bug_comment.txt
```diff
@@ -6,7 +6,7 @@ Crash report: {{ full_url(request, "crashstats:report_index", crash_id=uuid ) }}
 {% if java_stack_trace %}
 Java stack trace:
 ```
-{{ java_stack_trace|safe }}
+{{ java_stack_trace|truncate(5000, True)|safe }}
```
willkg, line 9: We have to truncate it here so we don't mess up the markdown formatting. Also, if we ever add additional bits after the stack trace, we don't want them truncated by the view.

PR #6421, opened 2023-06-14
bug 1837423: support __null__ for number fields in supersearch

willkg:
This fixes `__null__` and `!__null__` for number fields. Support was partially implemented a couple of years ago in the search library code, but the Django form fields which parse and validate the request query string were never adjusted to support `__null__` and `!__null__`, so it never worked.
    
This adds support for `__null__` and `!__null__` and also adds appropriate validation for combinations of these values and other filters for a number field. This adds appropriate tests as well.

### DIFF
--- a/socorro/lib/search_common.py
+++ b/socorro/lib/search_common.py
@@ -182,16 +182,17 @@ class SearchBase:
                                 break
 
                     # ensure the right data type
-                    try:
-                        value = convert_to_type(value, param.data_type)
-                    except ValueError as exc:
-                        raise BadArgumentError(
-                            param.name,
-                            msg=(
-                                "Bad value for parameter %s: not a valid %s"
-                                % (param.name, param.data_type)
-                            ),
-                        ) from exc
+                    if operator != "__null__":
+                        try:
+                            value = convert_to_type(value, param.data_type)
+                        except ValueError as exc:
+                            raise BadArgumentError(
+                                param.name,
+                                msg=(
+                                    "Bad value for parameter %s: not a valid %s"
+                                    % (param.name, param.data_type)
+                                ),
+                            ) from exc
 
                     if param.name not in parameters:
                         parameters[param.name] = []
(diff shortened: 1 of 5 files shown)


socorro/lib/search_common.py
```diff
@@ -182,16 +182,17 @@ def get_parameters(self, **kwargs):
...
-                        value = convert_to_type(value, param.data_type)
-                    except ValueError as exc:
-                        raise BadArgumentError(
-                            param.name,
-                            msg=(
-                                "Bad value for parameter %s: not a valid %s"
-                                % (param.name, param.data_type)
-                            ),
-                        ) from exc
+                    if operator != "__null__":
```
willkg, line 185: In the case of NumberField and using `__null__`, the value is `__null__` and that's not a valid int. That's the only exception and it'll probably be an exception for other things, too, so I added a test here.

webapp/crashstats/supersearch/form_fields.py
```diff
@@ -72,8 +72,8 @@ def to_python(self, value):
 
         return super().to_python(value)
 
-    def clean(self, *args, **kwargs):
-        cleaned_value = super().clean(*args, **kwargs)
+    def clean(self, value):
```
willkg, line 75: I fixed `clean` things to conform to the Django Field API. This reduces likelihood it breaks in a future Django upgrade.

webapp/crashstats/supersearch/form_fields.py
```diff
@@ -145,16 +147,39 @@ def validate_ordered_values(self, values, operators):
...
                     )
 
 
-class IntegerField(MultiplePrefixedValueField, forms.IntegerField):
-    pass
+class NumberField(MultiplePrefixedValueField, forms.IntegerField):
+    existence_strings = ["__null__", "!__null__"]
+
+    def to_python(self, value):
+        if isinstance(value, str) and value in self.existence_strings:
```
willkg, line 173: I can probably skip the isinstance checks here.

PR #6428, opened 2023-06-20
Bug 1826703 - Add specific signatures for hung processes

gabrielesvelto:


willkg:
`hang` isn't in the processed crash, yet. We need to add support for it in the processing rules. I don't think we should land this until we've got crash reports coming in with `Hang` annotation that we can test signature generation on.

gabrielesvelto:
> `hang` isn't in the processed crash, yet. We need to add support for it in the processing rules.

That's the part I was missing! Can I add the processor rules within this PR or should I open a separate bug?

> I don't think we should land this until we've got crash reports coming in with `Hang` annotation that we can test signature generation on.

I'll tell Alex he can go forward and land his changes then so we start getting these reports.

willkg:
While I've tried to make signature generation changes self-service-friendly, the processor rules kind of aren't at this time. Once there's crash report data with `Hang` annotations, I can look into this PR and the needed processor rule and see what's involved. My prediction is that it'll be pretty straight-forward and take me like 30 minutes.

willkg:
I forgot about this PR--sorry about that! I'll review it this week and push it over the finish line.

willkg (approved):
I tested this out with 5e876d3d-38d9-4ab9-a46e-80d320230820 and that now gets this signature:

```
hang: ui | bad hardware | LdrpDrainWorkQueue
```

PR #6434, opened 2023-07-05
bug 1667997: report_type field, processor rule, and adjustment to Top Crasher report

willkg:
This adds a `report_type` field with values "crash" (the default) and "hang". It implements a processor rule to populate this field using heuristics we've been using for signature generation for a while. It also adjusts the Top Crasher report to allow users to filter on "report type" with values "any", "crash" (the default), and "hang".

This eases triage work by splitting the hang type crash reports from the crash type crash reports which are more actionable.

### DIFF
--- a/socorro/processor/rules/mozilla.py
+++ b/socorro/processor/rules/mozilla.py
@@ -1110,16 +1110,34 @@ class ReportTypeRule(Rule):
 
     """
 
-    def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
-        # "crash" is the default type
-        report_type = "crash"
-
+    def identify_report_type(self, processed_crash, status):
         # NOTE(willkg): This comes from the ipc_channel_error crash annotation
         if "ipc_channel_error" in processed_crash:
-            report_type = "hang"
+            return "hang"
 
         # NOTE(willkg): This comes from the AsyncShutdownTimeout crash annotation
         if "async_shutdown_timeout" in processed_crash:
-            report_type = "hang"
+            return "hang"
+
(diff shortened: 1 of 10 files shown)


socorro/processor/rules/mozilla.py
```diff
@@ -1097,3 +1097,47 @@ def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
...
+            try:
+                crash_info = crash_data.get("crash_info") or {}
+                crashing_thread = int(crash_info.get("crashing_thread", 0))
+            except (TypeError, ValueError):
+                crashing_thread = 0
+
+            stack = glom(crash_data, "threads.%d.frames" % crashing_thread, default=[])
+            for frame in stack:
+                if "RunWatchdog" in (frame.get("function") or ""):
+                    return "hang"
```
willkg, line 1135: This matches the equivalent heuristics for the shutdown rule in signature generation. Because this rule runs before the signature generation rule in the processor pipeline, we can't use the results of signature generation to tell us it's a hang.

At some point, we might want to adjust signature generation to reflect the work done in processing rather than doing it's own thing.

socorro/schemas/processed_crash.schema.yaml
```diff
@@ -2864,6 +2864,12 @@ properties:
     type: string
     permissions: ["protected"]
     source_annotation: RemoteType
+  report_type:
+    description: >
+      Type of the crash report. Can be set to `crash` (the default for reports),
+      `hang`, etc.
```
willkg, line 2870: This should be fixed--there are only two possible values.

PR #6444, opened 2023-07-25
Bug 1845406 - Add nsTArrayElementTraits to the irrelevant signature list

amccreight:


amccreight:
Yeah, that looks good, thanks. I think some further work might be needed to deal with the nsTArray_Impl gunk but I don't think we want to get rid of all of it so it'll be a bit subtler.

willkg (approved):

PR #6445, opened 2023-07-27
bug 1845341: fix float support in super search fields

willkg:
Previously, "number" was used to support short, integer, long, and float. However, uptime_ts was the only float field so it was effectively treated as integer and the code didn't _actually_ support floats at all.

We need support for float values in super search filters and histograms, so this reworks the super search related code to separate "number" into "integer" and "float" types, add "float" as a valid histogram type, correctly support "integer" and "float" as histogram interval values, and support "float" validation through all the query type and data validation type using code.

This also updates all the super search documentation.


docs/crashstorage.rst
```diff
@@ -175,7 +175,8 @@ Here is an explanation of each properties of a field:
     Specifies how values are validated when passed to filters of this field
     in Super Search.
 
-    Possible values: ``str``, ``enum``, ``bool``, ``int``, ``datetime``,
+    Possible values: ``str``, ``enum``, ``bool``, ``int``, ``float``,
+    ``datetime``,
```
willkg, line 179: NB: I looked at whether we could merge `query_type` and `data_validation_type` and we might be able to at some point in the future, but there are some fields that have different values.

socorro/external/es/super_search_fields.py
```diff
@@ -358,26 +358,26 @@ def keyword_field(
...
 ):
-    """Generates a numeric field.
+    """Generates a whole number field.
 
     :param name: the name used to query the field in super search
     :param namespace: either "raw_crash" or "processed_crash"; note that we're moving
         to a model where we pull everything from the processed_crash, so prefer that
     :param in_database_name: the field in the processed crash to pull this data from
-    :param number_type: "short", "integer", "long", "double"
+    :param number_type: "short", "integer", "long"
```
willkg, line 373: I should fix this--it should talk about what `number_type` affects.

willkg: We might even want to rename it to `storage_mapping_type`. That's infinitely clearer.

socorro/tests/lib/test_search_common.py
```diff
@@ -22,7 +22,6 @@
         "data_validation_type": "str",
         "query_type": "string",
         "namespace": "processed_crash",
-        "permissions_needed": [],
```
willkg, line 25: We don't specify these in super search fields anymore. They're specified in the processed crash schema.

willkg:
After this deploys to stage, I'll go through and test `uptime` and `uptime_ts` filters, histograms, signature report, top crashers, etc. I'll also double-check the super search documentation.

PR #6447, opened 2023-07-28
bug 1826345: first part of possible_bit_flips support

willkg:
This is the first part of adding possible_bit_flips support. It:

1. adds `possible_bit_flips` to the stackwalker output in the processed crash schema
2. adds a processor rule to compute `possible_bit_flips_max_confidence` (0 to 100) and adds that field to the processed crash schema
3. adds `possible_bit_flips_max_confidence` to the search index so we can filter on it and do aggregations / histograms
4. adds `possible_bit_flips_max_confidence` to the Details tab in the report view

It's hard to find crash reports with possible bit flips data. After this lands and `possible_bit_flips_max_confidence` data accumulates in the index, it'll be easier to find crash reports allowing us to work on the next part where we add `possible_bit_flips_max_confidence` to the signature and top crashers reports.


Mrvirt222 (commented):
New. Sorry

PR #6450, opened 2023-08-10
Bug 1848162 - add mozilla::Span<T> as substring to ignore for signature generation

Archaeopteryx:


Archaeopteryx:
@amccreight Does the third example will provided fit crash triager's needs?

amccreight:
Yes, that looks fine. Thanks. The new signatures are still a bit generic but if that's an issue there can be followup work to add JS scope stuff to the prefix list. (Sorry for failing to reply yesterday, I don't always pay enough attention to my GitHub queue.)

willkg (approved):

willkg:
Thank you!

PR #6454, opened 2023-08-28
Bug 1850325 - Ignore `<std::collections::hash::map::HashMap<K, V, S> as core::ops::index::Index<&Q>>::index` when generating crash signatures

gabrielesvelto:


willkg (approved):
```
app@socorro:/app$ ./mybin/test_signature_change.sh '<std::collections::hash::map::HashMap<K, V, S> as core::ops::index::Index<&Q>>::index'
Working on '<std::collections::hash::map::HashMap<K, V, S> as core::ops::index::Index<&Q>>::index' ...
Crash id: 1a977ca2-f082-4339-90f8-992690230824
Original: core::option::expect_failed | <std::collections::hash::map::HashMap<K, V, S> as core::ops::index::Index<&Q>>::index
New:      core::option::expect_failed | webrender::scene_building::NodeIdToIndexMapper::get_spatial_node_index
Same?:    False

Crash id: 570e4563-1733-4f63-b9c3-a67980230824
Original: core::option::expect_failed | <std::collections::hash::map::HashMap<K, V, S> as core::ops::index::Index<&Q>>::index
New:      core::option::expect_failed | webrender::clip::ClipStore::get_template
Same?:    False
```

PR #6456, opened 2023-08-29
bug 1850428: rework stackwalker execution and timeout

willkg:
This reworks how we run the stackwalker. Previously, it was run using the timeout gnu coreutils command which would kill the stackwalker process after some timeout if it was still running.

processor -> timeout -> stackwalker

Now it runs the stackwalker using the timeout argument to subprocess.run() and subprocess.run() keeps track of time using threads. This removes an intermediary process and a possibility for the stackwalker process to get orphaned.

processor -> stackwalker

### DIFF
--- a/socorro/processor/rules/breakpad.py
+++ b/socorro/processor/rules/breakpad.py
@@ -200,23 +200,37 @@ class TruncateStacksRule(Rule):
                     self.truncate_stacks(value["json_dump"], f"{key}.json_dump", status)
 
 
-def execute_process(command_line):
+def execute_process(command_line, timeout=120):
     """Executes process and returns completed process data.
 
     :param command_line: the complete command line to run
+    :param timeout: the timeout in seconds for the command to complete before it's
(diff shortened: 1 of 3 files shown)


socorro/processor/rules/breakpad.py
```diff
@@ -200,23 +200,37 @@ def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
...
+            # NOTE(willkg) stdout and stderr here are bytes
+            "stdout": completed.stdout,
+            "stderr": completed.stderr,
+            "returncode": completed.returncode,
+        }
+
+    except subprocess.TimeoutExpired as timeout_exc:
+        ret = {
+            # NOTE(willkg) stdout and stderr here are bytes
+            "stdout": timeout_exc.stdout or b"",
```
willkg, line 228: `stdout` is an alias for `output` and it can be `None` in the case where there's no output. This normalizes things so we're always getting at least an empty byte string.

PR #6459, opened 2023-08-30
bug 1850750: add disk cache file size metrics

willkg:
This adds:

* `processor.cache_manager.file_sizes.median`
* `processor.cache_manager.file_sizes.ninety_five`

metrics.

It also redoes the metrics emitting so it's in a clearer "heartbeat" section that uses less cpu to debounce.

### DIFF
--- a/socorro/processor/cache_manager.py
+++ b/socorro/processor/cache_manager.py
@@ -20,11 +20,11 @@ To run::
 """
 
 from collections import OrderedDict
-import datetime
 import logging
 import os
 import pathlib
 import sys
+import time
 import traceback
 
 from boltons.dictutils import OneToOne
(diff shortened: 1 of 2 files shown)


socorro/processor/cache_manager.py
```diff
@@ -449,14 +479,24 @@ def _event_generator(self, nonblocking=False):
                         len(self.lru),
                         f"{self.total_size:,d}",
                     )
-                    # Emit a usage metric, but debounce it so it only gets emitted at
-                    # most once per second
-                    now = datetime.datetime.now()
-                    if now > (last_usage_update + datetime.timedelta(seconds=1)):
```
willkg, line 455: It's weird that I was creating datetimes and timedeltas to calculate this rather than using floats. What was I thinking?

socorro/processor/cache_manager.py
```diff
@@ -266,7 +296,7 @@ def _event_generator(self, nonblocking=False):
         self.running = True
         processed_events = False
         num_unhandled_errors = 0
-        last_usage_update = datetime.datetime.now()
+        next_heartbeat = time.time() + 1
```
willkg, line 299: `time.time()` returns seconds since the epoch as a float with fractional seconds. We add 1 here to determine the next heartbeat and debounce until we hit that.

socorro/processor/cache_manager.py
```diff
@@ -449,14 +479,24 @@ def _event_generator(self, nonblocking=False):
...
+                # Emit usage metric, but debounce it so it only gets emitted at most
+                # once per second
+                now = time.time()
+                if now > next_heartbeat:
+                    if is_verbose:
+                        METRICS.gauge("usage", value=self.total_size)
+
+                    sorted_sizes = list(sorted(self.lru.values()))
+                    if sorted_sizes:
+                        METRICS.gauge("file_sizes.median", get_index(sorted_sizes, 50))
```
willkg, line 493: These two metrics felt like the ones I want to see over time to figure out if the size of files in the disk cache (and thus the read/write work to use them) was increasing over time as the processor aged.

PR #6460, opened 2023-08-31
bug 1850750: add additional metrics for disk cache

willkg:
This adds:

* `processor.cache_manager.files.count`
* `processor.cache_manager.files.gt_500`
* `processor.cache_manager.file_sizes.avg`
* `processor.cache_manager.file_sizes.max`

### DIFF
--- a/socorro/processor/cache_manager.py
+++ b/socorro/processor/cache_manager.py
@@ -497,6 +497,7 @@ class DiskCacheManager:
                         METRICS.gauge(
                             "file_sizes.ninety_five", get_index(sorted_sizes, 95)
                         )
+                        METRICS.gauge("file_sizes.max", sorted_sizes[-1])
 
                         # Some metrics about what's in the cache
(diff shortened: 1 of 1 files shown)


socorro/processor/cache_manager.py
```diff
@@ -488,12 +488,21 @@ def _event_generator(self, nonblocking=False):
...
                             "file_sizes.ninety_five", get_index(sorted_sizes, 95)
                         )
+                        METRICS.gauge("file_sizes.max", sorted_sizes[-1])
+
+                        # Some metrics about what's in the cache
+                        METRICS.gauge("files.count", len(sorted_sizes))
+                        gt_500 = len([fs for fs in sorted_sizes if fs > 500_000_000])
+                        METRICS.gauge("files.gt_500", gt_500)
 
                     next_heartbeat = now + 1
```
willkg, line 507: This is really spammy in a local dev environment and I don't think we need down-to-the-second granularity, so I'm going to change the interval to 60s.

PR #6464, opened 2023-09-01
bug 1847429: implement signature generation for JavaException

willkg:
When there's no `JavaStackTrace` annotation, but there is a `JavaException` annotation, this will kick in and generate a signature that matches what it would look like if it came from the `JavaStackTrace` annotation.

The code is a little over-engineered, but gets us a step closer to changing the format for Java signatures to be more like the C signatures and to allow us to iterate over more than the first frame and include/ignore them like we do with C signatures.

This also moves the schema for the signature generation crash data structure to the socorro repository.

### DIFF
--- a/socorro/signature/rules.py
+++ b/socorro/signature/rules.py
@@ -415,7 +415,7 @@ class CSignatureTool:
 
 
 class JavaSignatureTool:
-    """This is the signature generation class for Java signatures."""
+    """Signature generation class for signatures based on JavaStackTrace."""
 
     # The max length of a java exception description--if it's longer than this,
     # drop it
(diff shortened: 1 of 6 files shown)


socorro/signature/rules.py
```diff
@@ -496,6 +496,67 @@ def generate(self, source, delimiter=": "):
...
+    def generate(self, source):
+        if not isinstance(source, dict):
+            return (
+                "EMPTY: Java exception not in expected format",
+                ["JavaExceptionSignatureTool: value not in expected format"],
+                [],
+            )
+
+        # This looks at the first frame (the one that raised the exception) of the first
+        # exception value (the exception that started a series of cascading exceptions).
```
willkg, line 531: The `JavaException` structure is modeled after the one that Sentry uses in their event payloads but we have the opposite order for stacktraces and for frames. My fault--should have caught it earlier. So when/if we implement support for Sentry event payloads, we'll have to do some normalizing of the values in the processor.

willkg:
```
app@socorro:/app$ cat crashids.txt | socorro-cmd signature
Crash id: 1e40591c-39a5-4753-9319-901f80230901
Original: EMPTY: no frame data available
New:      java.lang.RuntimeException: at android.graphics.BaseCanvas.throwIfCannotDraw(BaseCanvas.java)
Same?:    False

Crash id: 0713f9c5-1823-48d8-b215-ea8af0230901
Original: EMPTY: no frame data available
New:      java.lang.OutOfMemoryError: at java.util.Arrays.copyOf(Arrays.java)
Same?:    False

Crash id: b36f3e39-ef44-41ad-819f-cff5f0230901
Original: EMPTY: no frame data available
New:      java.lang.NoSuchFieldException: at java.lang.Class.getDeclaredField(Class.java)
Same?:    False

Crash id: b9c6e055-0d93-46dd-bce2-5eec20230901
Original: EMPTY: no frame data available
New:      java.lang.NoSuchFieldException: at java.lang.Class.getDeclaredField(Class.java)
Same?:    False

Crash id: 214076ff-620b-469f-9802-2575f0230901
Original: EMPTY: no frame data available
New:      java.lang.IllegalStateException: at mozilla.components.feature.sitepermissions.SitePermissionsFeature.onContentPermissionRequested$feature_sitepermissions_release(SitePermissionsFeature.kt)
Same?:    False

Crash id: 8d52a1c1-5f58-48e0-95f8-924f50230901
Original: EMPTY: no frame data available
New:      java.lang.IllegalStateException: at mozilla.components.feature.sitepermissions.SitePermissionsFeature.onContentPermissionRequested$feature_sitepermissions_release(SitePermissionsFeature.kt)
Same?:    False

Crash id: 0defac0e-4504-4f72-a97c-140100230901
Original: EMPTY: no frame data available
New:      java.lang.IllegalStateException: at android.app.ContextImpl.startServiceCommon(ContextImpl.java)
Same?:    False
```

socorro/signature/rules.py
```diff
@@ -496,6 +496,67 @@ def generate(self, source, delimiter=": "):
...
+        exc_type = stacktrace.get("type", "Unknown")
+        if exc_module and exc_type:
+            exc_class = f"{exc_module}.{exc_type}"
+        else:
+            exc_class = exc_type
+
+        frames = stacktrace.get("frames", [])
+        normalized_frames = [self.normalize_frame(frame) for frame in frames]
+        if frames:
+            signature = f"{exc_class}: at {normalized_frames[0]}"
```
willkg, line 553: When we redo signature generation for Java, we'll iterate through the frames the same way we do with C signatures. But for now, we only look at the first one.

socorro/signature/rules.py
```diff
@@ -496,6 +496,67 @@ def generate(self, source, delimiter=": "):
...
+        debug_notes = []
+
+        exc_module = stacktrace.get("module", "")
+        exc_type = stacktrace.get("type", "Unknown")
+        if exc_module and exc_type:
+            exc_class = f"{exc_module}.{exc_type}"
+        else:
+            exc_class = exc_type
+
+        frames = stacktrace.get("frames", [])
```
willkg, line 550: Oops--this needs to protect against frames being something unhelpful like `None`.

PR #6483, opened 2023-10-12
bug 1858717: switch stackwalker to use output file

willkg:
This switches the way Socorro runs the stackwalker to use an output file for output (previously used stdout) and a log file for error output (previously used stderr). This simplifies the code plus it makes it easier to differentiate between the output artifact and stackwalker logging.

This fixes the problem with bin/run_mdsw.sh where it didn't create the symbols cache directories before running the stackwalker. Now it does.

This also changes `bin/run_mdsw.sh` to use `--verbose=debug` since we're likely using it to debug stackwalker issues.

### DIFF
--- a/bin/run_mdsw.sh
+++ b/bin/run_mdsw.sh
@@ -15,6 +15,7 @@ set -euo pipefail
 
 DATADIR=./crashdata_mdsw_tmp
 STACKWALKER="/stackwalk-rust/minidump-stackwalk"
+SYMBOLSCACHE="/tmp/symbols"
 
 # This will pull symbols from the symbols server
 SYMBOLS="--symbols-url=https://symbols.mozilla.org"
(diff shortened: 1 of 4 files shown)


bin/run_mdsw.sh
```diff
@@ -35,25 +36,29 @@ if [[ $# -eq 0 ]]; then
 fi
 
 mkdir "${DATADIR}" || true
+mkdir -p "${SYMBOLSCACHE}/cache" || true
+mkdir -p "${SYMBOLSCACHE}/tmp" || true
```
willkg, line 40: I discovered this was a problem after adding `--log-file` and making sure the log file was working.

docs/overview.rst
```diff
@@ -117,11 +118,10 @@ The collector saves the crash report data to AWS S3 as a :term:`raw crash` and
 
 .. code-block:: text
 
-   v2/
```
willkg, line 120: We moved everything to the `v1/` directory a bunch of months ago.

PR #6486, opened 2023-10-23
bug 1860580: Make Docker Hub upload script more robust.

smarnach:
This is the same change as https://github.com/mozilla-services/antenna/pull/947


willkg (approved):
Thank you! :100:

PR #6488, opened 2023-10-30
webapp: remove unused pkg.json package

biancadanforth:
This package is several major versions out of date, and after reviewing the code, it does not appear to be used anywhere. Removing it will prevent us from receiving N/A warnings from GitHub/npm.


willkg (approved):
Looks good to me as far as I can tell. Thank you for tackling this!

PR #6490, opened 2023-11-01
Update Django to 3.2.23

willkg:
This updates Django to 3.2.23 to pick up a security fix.

https://www.djangoproject.com/weblog/2023/nov/01/security-releases/


smarnach (approved):
If I'm reading this correctly, the vulnerability would only affect us if we'd run the server on Windows, but it shouldn't hurt picking this up.

willkg:
Yes. I should have noted in the description I don't think this affects us. Generally, I try to pick up security fixes with some urgency regardless of whether they help or not because it makes for easier bookkeeping.

Thank you for reviewing!

PR #6494, opened 2023-11-02
Update psutil from 5.9.5 to 5.9.6

willkg:


willkg:
@biancadanforth I need to trigger a stage deploy and the easiest thing to do is land a change. This is a pretty minor update to psutil, so there isn't much to review here. Pretty much if CI passes, then it's probably fine.

willkg:
Boo hiss. I'll look into this tomorrow if no one figures it out before then.

smarnach (approved):

PR #6498, opened 2023-11-16
Update docs to use hyphens when referencing bugs

biancadanforth:
Because:
* We have enabled GitHub autolinking[1] in the crash ingestion repos (socorro, tecken, eliot and antenna)
* GitHub autolinking doesn't support spaces when pattern matching, so we've added support to rob-bugson (#willkg/rob-bugson#51) to recognize hyphens, and we want to use hyphens when referencing bugs going forward.

This commit:
* Updates our docs re: conventions when referencing bug ID(s) in git commit subjects and GitHub PR titles

[1]: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls#custom-autolinks-to-external-resources

Screenshot 2023-11-16 at 5 12 23 PM
Screenshot 2023-11-16 at 5 11 53 PM

### DIFF
--- a/docs/dev.rst
+++ b/docs/dev.rst
@@ -171,10 +171,14 @@ additional details by posting comments in the bug.
 Pull requests
 -------------
 
-Pull request summary should indicate the bug the pull request addresses. For
+Pull request summary should indicate the bug the pull request addresses. Use a hyphen between "bug" and the bug ID(s). For
 example::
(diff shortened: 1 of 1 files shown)


willkg (approved):
Awesome--thank you for doing this!

PR #6508, opened 2023-12-20
bug-1799125: switch from urllib.parse's urlparse to urlsplit

relud:
and from `urllib.parse.urlunparse` to `urllib.parse.SplitResult.geturl` instead of `urllib.parse.urlunsplit` so that we consistently use `geturl` instead of a mix of both

### DIFF
--- a/bin/service-status.py
+++ b/bin/service-status.py
@@ -24,7 +24,7 @@ import argparse
 import json
 import os
 import sys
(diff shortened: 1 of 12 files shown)


smarnach (approved):
Looks good to me. I double checked we are not using `.params` or (less likely) attribute access by index rather than name, but seems all good.

PR #6510, opened 2024-01-04
bug-1761322: add support for `Android_PackageName`

willkg:
This adds support for the `Android_PackageName` annotation by adding it to the raw and processed crash schemas, adding it to super search fields, and adding it to the crash stats report view details tab.

This also updates the description and data review url for all the `Android_*` crash annotation fields in the raw crash schema.

To test:

1. run `make run`
2. process one or more Fenix crash reports
3. view a crash report
   1. verify that "Android Package Name" shows up in the report view
4. go to super search
   1. search for product=Fenix, android packagename=org.mozilla.fenix and verify that reports come up
   2. facet on android packagename and verify that the facet works


socorro/external/es/super_search_fields.py
```diff
@@ -802,6 +802,7 @@ def apply_schema_properties(fields, schema):
             "type": "multi_field",
         },
     },
+    "android_packagename": keyword_field(name="android_packagename"),
```
willkg, line 805: `keyword_field` is a function that generates all the data for a keyword field. There are a few other functions that work like this for other data types. I always intended to go through `super_search_fields.py` and rework `FIELDS` to use the field functions since it's much easier to work with.

socorro/schemas/processed_crash.schema.yaml
```diff
@@ -2181,57 +2181,82 @@ properties:
     type: ["string", "null"]
     permissions: ["public"]
   android_board:
-    description: The board used by the Android device.
+    description: >
+      The name of the underlying board used by the Android device.
```
willkg, line 2185: The descriptions are copied over from [CrashAnnotations.yaml](https://searchfox.org/mozilla-central/source/toolkit/crashreporter/CrashAnnotations.yaml).

relud (approved):

willkg:
Thank you!

PR #6511, opened 2024-01-05
Fix opt-out language

willkg:
Turns out I've had the "opt-out by default" language wrong all these years. This fixes cases of that to clearer "default on/off" language.


willkg:
self-r+ -- this is a docs fix based on a discussion I had with Nneka.

PR #6512, opened 2024-01-10
bug-1862204: add missing_symbols field

willkg:
This adds a missing_symbols field to the processed crash and the search index.

This field is a `;` delimited list of "filename/version/debugid" tuples.

This is generated from the json_dump.modules list of modules that have a "filename" and "missing_symbols=True". "missing_symbols=True" indicates that the stackwalker was unable to find a symbols file at the time it was processing the minidump.

Because I'm working with debug ids, I decided to add a debug_id_helper pytest fixture for creating valid debug ids.

To test:

1. download some crash reports and process them
2. go to supersearch and do a facet on "missing symbols"

Unless you're really lucky, there should be one or two missing symbols.

image

### DIFF
--- a/socorro/external/es/super_search_fields.py
+++ b/socorro/external/es/super_search_fields.py
@@ -1617,6 +1617,18 @@ FIELDS = {
         "query_type": "string",
         "storage_mapping": {"type": "string"},
     },
+    "missing_symbols": {
+        "data_validation_type": "str",
+        "form_field_choices": [],
+        "has_full_version": False,
+        "in_database_name": "missing_symbols",
+        "is_exposed": True,
+        "is_returned": True,
+        "name": "missing_symbols",
+        "namespace": "processed_crash",
+        "query_type": "string",
+        "storage_mapping": {"analyzer": "semicolon_keywords", "type": "string"},
+    },
     "modules_in_stack": {
         "data_validation_type": "str",
         "form_field_choices": [],
--- a/socorro/processor/rules/mozilla.py
+++ b/socorro/processor/rules/mozilla.py
@@ -626,10 +626,57 @@ class TopMostFilesRule(Rule):
                 return
 
 
+class MissingSymbolsRule(Rule):
+    """
+    Adds ``missing_symbols`` field where the value is a semi-colon separated set of
+    ``module/version/debugid`` strings for modules where the stackwalker couldn't find a
+    symbols file.
+    """
+
+    # Filenames should contain A-Za-z0-9_.- and that's it.
+    BAD_FILENAME_CHARACTERS = re.compile(r"[^a-zA-Z0-9_\.-]", re.IGNORECASE)
+
+    # Debug ids are hex strings
+    BAD_DEBUGID_CHARACTERS = re.compile(r"[^a-f0-9]", re.IGNORECASE)
+
+    NULL_DEBUG_ID = "0" * 33
+
+    def format_module(self, item):
+        filename = item["filename"]
+        filename = self.BAD_FILENAME_CHARACTERS.sub("", filename)
+
+        version = item.get("version", "") or "None"
+        version = version.replace("/", "\\/")
+
+        debugid = item.get("debug_id", self.NULL_DEBUG_ID)
+        debugid = self.BAD_DEBUGID_CHARACTERS.sub("", debugid)
+
+        return f"{filename}/{version}/{debugid}"
+
+    def predicate(self, raw_crash, dumps, processed_crash, tmpdir, status):
+        return "json_dump" in processed_crash
+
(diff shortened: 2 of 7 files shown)


socorro/tests/conftest.py
```diff
@@ -79,6 +80,19 @@ def caplogpp(caplog):
         logger.propagate = False
 
 
+class DebugIdHelper:
+    """Breakpad debug id helper class."""
+
+    def generate(self):
+        """Returns 33-character uppercase hex string"""
+        return uuid.uuid4().hex.upper() + "A"
```
willkg, line 88: `uuid.uuid4().hex.upper()` generates a string that's 32 hex characters long:

```python
Python 3.10.12 (main, Jul  3 2023, 09:21:40) [GCC 12.2.0] on linux
>>> import uuid
>>> len(uuid.uuid4().hex.upper())
32
```
So we add an `"A"` to that to get 33 characters.

https://tecken.readthedocs.io/en/latest/glossary.html#term-debug_id

socorro/tests/processor/rules/test_mozilla.py
```diff
@@ -1439,6 +1440,107 @@ def test_missing_crashing_thread(self, tmp_path):
...
+                "libxul.so/None/" + MissingSymbolsRule.NULL_DEBUG_ID,
+            ),
+            (
+                {
+                    "filename": "libxul_2.dll",
+                    "version": "1.0",
+                    "debug_id": "51C36FAFFD214DB4A0D91D93B38336CEA",
+                    "missing_symbols": True,
+                },
+                "libxul_2.dll/1.0/51C36FAFFD214DB4A0D91D93B38336CEA",
```
willkg, line 1526: It's tricky to use the debug id helper in the parametrized options, so I hard-coded the debug id here.

socorro/external/es/super_search_fields.py
```diff
@@ -1617,6 +1617,18 @@ def apply_schema_properties(fields, schema):
...
+        "form_field_choices": [],
+        "has_full_version": False,
+        "in_database_name": "missing_symbols",
+        "is_exposed": True,
+        "is_returned": True,
+        "name": "missing_symbols",
+        "namespace": "processed_crash",
+        "query_type": "string",
+        "storage_mapping": {"analyzer": "semicolon_keywords", "type": "string"},
+    },
```
willkg, line 1631: This is identical to `modules_in_stack` value after substituting `"modules_in_stack"` with `"missing_symbols"`

socorro/processor/rules/mozilla.py
```diff
@@ -626,10 +626,57 @@ def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
...
+
+    def format_module(self, item):
+        filename = item["filename"]
+        filename = self.BAD_FILENAME_CHARACTERS.sub("", filename)
+
+        version = item.get("version", "") or "None"
+        version = version.replace("/", "\\/")
+
+        debugid = item.get("debug_id", self.NULL_DEBUG_ID)
+        debugid = self.BAD_DEBUGID_CHARACTERS.sub("", debugid)
```
willkg, line 652: I've seen wild filenames, so normalizing those to valid values is important. I haven't seen wild versions or wild debug ids, yet, so the validation here is half-hearted.

socorro/processor/rules/mozilla.py
```diff
@@ -626,10 +626,57 @@ def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
...
+
+        return f"{filename}/{version}/{debugid}"
+
+    def predicate(self, raw_crash, dumps, processed_crash, tmpdir, status):
+        return "json_dump" in processed_crash
```
willkg, line 657: I should fix this to cover `processed_crash["json_dump"]["modules"]` exists and is not empty. That would reduce some of the work in `action`.

socorro/tests/processor/rules/test_mozilla.py
```diff
@@ -1439,6 +1440,107 @@ def test_missing_crashing_thread(self, tmp_path):
         assert "topmost_filenames" not in processed_crash
 
 
+def generate_debug_id():
+    """Generates a valid debug id."""
+
+
```
relud, line 1446: this isn't used, doesn't appear to do anything, and looks like it was replaced by `debug_id_helper.generate()`
```suggestion
```

relud (approved):
r+wc

socorro/tests/processor/rules/test_mozilla.py
willkg: You're right! I wrote it and then decided to do a pytest fixture because I could use it elsewhere. I'll remove it.

willkg:
Thank you!

PR #6513, opened 2024-01-10
bug-1873658: fix params parsing date=>=2024-01-10 bug

willkg:
This fixes a bug that occurs when breaking up the querystring into key=val parameters when the value has an = in it. Instead of dropping everything after the second = from the value, this takes more care to split on the first = and ensures everything after that is part of the value.

Before this fix, something like this kicks up a JS error in the console and prevents the page from fetching search results:

http://localhost:8000/search/?date=%3E=2024-01-09&date=%3C2024-01-10#crash-reports

After this fix, the params are parsed correctly, the from date is filled in, and search results show correctly.

### DIFF
--- a/webapp/crashstats/crashstats/static/crashstats/js/socorro/utils.js
+++ b/webapp/crashstats/crashstats/static/crashstats/js/socorro/utils.js
@@ -184,6 +184,7 @@
         var params = {};
         var queries;
         var temp;
+        var firstEquals;
         var i;
         var len;
 
@@ -197,7 +198,8 @@
 
         // Convert the array of strings into an object
         for (i = 0; i < len; i++) {
-          temp = queries[i].split('=');
+          firstEquals = queries[i].indexOf('=');
+          temp = [queries[i].slice(0, firstEquals), queries[i].slice(firstEquals + 1)];
           var key = temp[0];
           var value = decodeURIComponent(temp[1]);
 


webapp/crashstats/crashstats/static/crashstats/js/socorro/utils.js
```diff
@@ -197,7 +198,8 @@
 
         // Convert the array of strings into an object
         for (i = 0; i < len; i++) {
-          temp = queries[i].split('=');
```
willkg, line 200: When `queries[i]` is something like `date=>%3D2024-01-10`, then we get:

```javascript
temp = ['date', '>%3D2024-01-10']
```

and then everything works super.

The problem scenario here is when `queries[i]` is something like `date=>=2024-01-10`. In that case, we get:

```javascript
temp = ['date', '>', '2024-01-10']
```

Then the `key` becomes `date` and the `value` becomes `>` and the `2024-01-10` part is completely dumped. Then shenanigans ensue and everything goes sideways.

The fix is to only split on the first `=`, but doing `.split('=', 1)` doesn't do what it does in Python--it does this other weird thing which is wildly unhelpful.

So instead, I went with manually splitting the string using `.indexOf('=')` which gives us the index and then creating `temp` by using `slice` to get the key and value as the 0 and 1 indexes.

biancadanforth (approved):
Thank you for the detailed explanation. It took me a while to follow what the code's doing, since I've not worked on Crash Stats yet. I can confirm that I can reproduce the error with the link you provided and no longer see the issue after your patch.

In general, we should replace `search.parseQueryString` with `URL.searchParams`, which I think is a relatively new (compared to the age of this code) built-in JavaScript interface that parses query strings for you. I tested it with this scenario, and it correctly parsed the value with the extra `=`. I can file a bug for that if one doesn't already exist.

willkg:
Thank you!

re: updating the code, I added bug-1874265 to the user interface tracker bug-1699182. There's a lot of javascript that should get updated, refactored, etc.

PR #6516, opened 2024-01-17
bug-1835881: add support for mac_boot_args

willkg:
The stackwalker parses the `MozMacosBootargsStream` if it exists in the minidump and puts the value in `json_dump.mac_boot_args`. If that field exists and the value is non-empty, this will add two fields to the processed crash:

* `has_mac_boot_args`: a flag (public) which is True if the field exists and is non-empty or omitted otherwise
* `mac_boot_args`: a string (protected) which has the contents of the `mac_boot_args` field

Both fields are added to the index and are searchable.

Crash reports that have `mac_boot_args` without sensitive information:

crash id | mac boot args value
--- | ---
eaa702fa-940a-40a8-98e6-ac6b20231203 | `-no_compat_check`
cc9a3ac6-7904-4fb5-8b3b-6ead20231202 | `-v`
8a8fc36d-0c7c-48cd-aeef-d7d460231202 | `-arm64e_preview_abi`
ade170c6-b5b0-4cf5-9ff1-87e690231202 | `-no_compat_check`

When not logged in or without protected data access:

1. filter on `has_mac_boot_args` using `exists` and `is True` should work and bring up crash reports
2. `has_mac_boot_args` shows up in the data dictionary in the "Processed crash fields" section
3. `mac_boot_args` shows up in the data dictionary in the "Processed crash fields" section

When logged in with protected data access:

1. filter on `has_mac_boot_args` using `exists` and `is True` should work and bring up crash reports
4. filter on `mac_boot_args` using `contains` and `no_compat_check` should bring up crash reports with `no_compat_check` in the `mac_boot_args`
5. facet on `mac_boot_args` should work
6. adding `mac_boot_args` to show columns should work
7. `mac_boot_args` and `has_mac_boot_args` should show up in the processed crash data


socorro/processor/rules/mozilla.py
```diff
@@ -449,6 +449,33 @@ def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
...
+    If there's a mac_boot_args in the json_dump and it's not an empty value, this
+    copies it to the processed_crash.
+
+    """
+
+    def predicate(self, raw_crash, dumps, processed_crash, tmpdir, status):
+        return bool(glom(processed_crash, "json_dump.mac_boot_args", default=None))
+
+    def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
+        value = processed_crash["json_dump"]["mac_boot_args"]
```
relud, line 464: am I understanding correctly that `processed_crash["json_dump"]["mac_boot_args"]` won't raise KeyError, because predicate ensures that if it didn't exist before it's now None?

relud: wait no, it won't throw KeyError because predicate ensures that action isn't called unless there is a truthy value at that path?

socorro/tests/processor/rules/test_mozilla.py
```diff
@@ -791,6 +792,58 @@ def test_absent_seconds_since_last_crash(self, tmp_path):
...
 
 
+class TestMacBootArgsRule:
+    @pytest.mark.parametrize(
+        "processed, expected",
+        [
+            ({}, False),
+            ({"json_dump": {}}, False),
+            ({"json_dump": {"mac_boot_args": None}}, False),
+            ({"json_dump": {"mac_boot_args": ""}}, False),
```
relud, line 802: seems like it is covered by downstream tests, but I feel like this test should still cover at least one truthy case:
```suggestion
            ({"json_dump": {"mac_boot_args": ""}}, False),
            ({"json_dump": {"mac_boot_args": "-v"}}, True),
```

relud (approved):
r+wc (including above comment)

socorro/tests/processor/rules/test_mozilla.py
willkg: I can add that. :+1:

willkg:
Thank you!

PR #6517, opened 2024-01-18
bug-1812771: clean up processor rules for java/android things

willkg:
This moves some java/android related processor rule code into separate modules. It makes it a little easier to find rules.

### DIFF
--- a/socorro/mozilla_rulesets.py
+++ b/socorro/mozilla_rulesets.py
@@ -20,6 +20,7 @@ from socorro.processor.rules.general import (
     IdentifierRule,
     OSInfoRule,
 )
+from socorro.processor.rules.java import JavaStackTraceRule
 from socorro.processor.rules.memory_report_extraction import MemoryReportExtraction
 from socorro.processor.rules.mozilla import (
     AccessibilityRule,
@@ -32,7 +33,6 @@ from socorro.processor.rules.mozilla import (
     DistributionIdRule,
     ESRVersionRewrite,
     FenixVersionRewriteRule,
-    JavaProcessRule,
     MajorVersionRule,
     MissingSymbolsRule,
     ModulesInStackRule,
@@ -91,7 +91,7 @@ DEFAULT_RULESET = [
     OutOfMemoryBinaryRule(),
     PHCRule(),
     BreadcrumbsRule(schema=get_schema("processed_crash.schema.yaml")),
-    JavaProcessRule(),
+    JavaStackTraceRule(),
     MacBootArgsRule(),
     MacCrashInfoRule(),
     MozCrashReasonRule(),
--- a/socorro/mozilla_rulesets.py
+++ b/socorro/mozilla_rulesets.py
@@ -4,6 +4,7 @@
 
 from socorro import settings
 from socorro.lib.libsocorrodataschema import get_schema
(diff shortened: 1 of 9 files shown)


socorro/mozilla_rulesets.py
```diff
@@ -91,14 +92,15 @@
     OutOfMemoryBinaryRule(),
     PHCRule(),
     BreadcrumbsRule(schema=get_schema("processed_crash.schema.yaml")),
-    JavaProcessRule(),
```
willkg, line 94: "JavaProcessRule" isn't wildly helpful for what it does. It's a rule focusing on the `JavaStackTrace` annotation and processing that specifically. So I renamed it.

socorro/processor/rules/android.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def predicate(self, raw_crash, dumps, processed_crash, tmpdir, status):
+        cpu_arch = processed_crash.get("cpu_arch", "unknown")
+        return cpu_arch == "unknown" and "Android_CPU_ABI" in raw_crash
+
+    def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
+        android_cpu_abi = raw_crash["Android_CPU_ABI"]
+        cpu_arch = self.ANDROID_CPU_ABI_MAP.get(android_cpu_abi, android_cpu_abi)
+
+        processed_crash["cpu_arch"] = cpu_arch
```
willkg, line 35: `CPUInfoRule` sets `cpu_info` to "unknown" and `cpu_count` to "0" both of which aren't helpful, so we don't set them here.

socorro/processor/rules/general.py
```diff
@@ -96,13 +96,6 @@ class CPUInfoRule(Rule):
 
     """
 
-    # Map of Android_CPU_ABI values to cpu_arch values
-    ANDROID_CPU_ABI_MAP = {
-        "armeabi-v7a": "arm",
-        "arm64-v8a": "arm64",
-        "x86_64": "amd64",
-    }
-
```
willkg, line 105: Oops--this needs a predicate so it only kicks in if `json_dump.system_info.cpu_info` exists.

willkg: Actually, scratch that--the way I have it now, it does exactly what the old code did.

socorro/processor/rules/android.py
willkg: Actually, scratch that--the way I have it now, it does exactly what the old code did.

socorro/schemas/processed_crash.schema.yaml
```diff
@@ -2848,12 +2848,16 @@ properties:
...
+      For Android, this is `Android` plus the sdk version.
     type: string
     permissions: ["public"]
   os_version:
-    description: >
+    description: |
       Version of the operating system. This comes from `system_info.os_ver`
       from the stackwalker output of the minidump.
+
+      For Android, this is the sdk version.
```
willkg, line 2860: Oops--this shouldn't change in this PR.

willkg:
There's some other reorganization work I want to do at some point, but I limited this cleanup work to the minimum to make the next change easier.

relud (approved):

willkg:
Thank you!

PR #6518, opened 2024-01-18
bug-1812771: fix os_name, os_version, and os_pretty_version for Android

willkg:
We want to pull the version for Android from the `Android_Version` annotation which holds the sdk version and the release name. Then we use that to build an os_pretty_version that works for C++/Rust crash reports as well as Java crash reports and is much more useful than what we've got now.

I did this implementation so that it only changes the `os_name`, `os_version`, and `os_pretty_version` for crashes on Android platforms.

What we've got now:

image

*Figure 1: Bunch of "Unknown" and the version is set to the Linux kernel. Unhelpful.*

What the changes in this PR do:

image

*Figure 2: Android crashes are marked "Android" and the version is always the sdk version.*

### DIFF
--- a/socorro/processor/rules/mozilla.py
+++ b/socorro/processor/rules/mozilla.py
@@ -848,16 +848,20 @@ class BetaVersionRule(Rule):
 
 
 class OSPrettyVersionRule(Rule):
-    """Populate os_pretty_version with most readable operating system version string.
+    """Sets os_pretty_version with most readable operating system version string.
 
     This rule attempts to extract the most useful, singular, human understandable field
-    for operating system version. This should always be attempted.
+    for operating system version.
+
+    * os_pretty_version
 
     For Windows, this is a lookup against a map.
 
     For Mac OSX, this pulls from os_name and os_version.
 
     For Linux, this uses json_dump.lsb_release.description if it's available.
+
+    Must be run after OSInfoRule.
 
     """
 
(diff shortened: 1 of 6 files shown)


socorro/processor/rules/mozilla.py
```diff
@@ -879,60 +883,97 @@ class OSPrettyVersionRule(Rule):
...
+        )
+        pretty_version = pretty_version or "Linux"
+        return pretty_version
+
+    def compute_android_pretty_version(self, os_name, os_version):
+        if os_version:
+            return f"{os_name} {os_version}"
+        return os_name
+
+    def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
```
willkg, line 949: This method was kind of a big mess. I split out the platform-specific code into separate methods to make it much easier to reason about and follow the flow.

socorro/tests/processor/rules/test_mozilla.py
```diff
@@ -1876,16 +1879,6 @@ def test_junk_data(self, tmp_path, os_name, os_version, expected):
         rule.act(raw_crash, dumps, processed_crash, str(tmp_path), status)
         assert processed_crash["os_pretty_version"] == expected
 
-    def test_dotdict(self, tmp_path):
```
willkg, line 1879: This is a vestige of the past when we had `DotDict` and the raw and processed crash structures were `DotDict` structures sometimes except on days where sunrise is on the 00 of the hour when they're dicts. Those days are so over so we can remove this.

It probably should have been removed in d727e09e4c5d4f0f7c818cc7c164a818c38e7e0c.

relud (approved):

willkg:
Thank you!

PR #6520, opened 2024-01-25
bug-1527460: switch webapp to everett

relud:


smarnach (approved):
This is looking good to me.

webapp/crashstats/settings/base.py
```diff
@@ -41,12 +50,20 @@ def path(*dirs):
...
 
-# Debugging displays nice error messages, but leaks memory. Set this to False
-# on all server instances and True only for development.
-DEBUG = config("DEBUG", False, cast=bool)
+DEBUG = config(
+    "DEBUG",
+    default="false",
+    parser=parse_bool,
+    doc="Debugging displays nice error messages, but leaks memory. Set this to false"
+    " on all server instances and true only for development.",
```
willkg, line 65: I would put this in parentheses so it's easier to see that this is two strings concatenated by the Python parser into a single string.

```suggestion
    doc=(
        "Debugging displays nice error messages, but leaks memory. Set this to false"
        " on all server instances and true only for development."
    ),
```

I also put the whitespace at the end of the line, so this looks different than how I've been doing it. However, I'm not sure that matters, so I wouldn't change it.

Vaguely related to this, but not really--one thing to note here is that Everett will pick up a single string where the concatenation is implied and handled by the Python parser like this, but it won't pick up a string where the concatenation is done at runtime with the `+` operator. So what you've got works and we shouldn't change it to something like `"string 1" + "string 2"`. That's a bug in Everett that I keep meaning to write up an issue for. I did that just now: https://github.com/willkg/everett/issues/240

webapp/crashstats/settings/base.py
```diff
@@ -11,13 +11,14 @@
 import re
 import socket
 
-from decouple import config, Csv
+from everett.manager import ConfigManager, parse_bool
```
willkg, line 14: Now that we're using Everett for configuration, we can use `automoduleconfig` to auto-document configuration.

We need to add something this to `docs/configuration.rst` in the webapp section:

```rest
.. automoduleconfig:: crashstats.settings.base._config
   :show-table:
   :hide-name:
   :case: upper
```

Note the configuration variable name will depend on what you name the configuration variable in line 21.

webapp/crashstats/settings/base.py
```diff
@@ -394,12 +423,14 @@ def filter(self, record):
 # Below are settings that can be overridden using
 # environment variables.
 
-CACHE_IMPLEMENTATION_FETCHES = config("CACHE_IMPLEMENTATION_FETCHES", True, cast=bool)
+CACHE_IMPLEMENTATION_FETCHES = config(
+    "CACHE_IMPLEMENTATION_FETCHES", default="true", parser=parse_bool
+)
```
willkg, line 428: I can improve these settings docs in a future PR.

webapp/crashstats/settings/base.py
```diff
@@ -433,9 +464,9 @@ def filter(self, record):
     }
 }
 
-TIME_ZONE = config("TIME_ZONE", "UTC")
+TIME_ZONE = config("TIME_ZONE", default="UTC")
```
willkg, line 467: This is weird. We never set the `TIME_ZONE` so I have no idea why this was configurable.

willkg (changes requested):
I had some things that should get fixed, but otherwise this is great! I'm psyched the changes seem to be relatively straightforward.

webapp/crashstats/settings/base.py
```diff
@@ -29,9 +30,17 @@ def path(*dirs):
     return os.path.join(ROOT, *dirs)
 
 
-# Whether or not we're running in a tool environment where we want to ignore required
-# configuration
-TOOL_ENV = config("TOOL_ENV", False, cast=bool)
+def parse_csv(val: str) -> list[str]:
+    return [v.strip() for v in val.split(",")]
```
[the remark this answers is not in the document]
willkg: https://github.com/willkg/everett/issues/241

webapp/crashstats/settings/base.py
relud: I would imagine this ensures that we use UTC in dev configurations, which is good for consistency

docs/configuration.rst
```diff
@@ -84,4 +84,7 @@ Gunicorn configuration:
 
 Webapp configuration:
 
-To be documented.
+.. automoduleconfig:: crashstats.settings.base._config
```
relud, line 87: should this be under webapp?
```suggestion
.. automoduleconfig:: webapp.crashstats.settings.base._config
```

webapp/crashstats/settings/base.py
willkg: We don't set this in server environments, so it should always be `UTC`. It's curious.

docs/configuration.rst
willkg: Mmm... I want to say no it shouldn't because `webapp/` shouldn't be a Python module. But it turns out that it has a `__init__.py` file, so it is a Python module so maybe that works? The more correct way would be to do `crashstats.settings.base._config` and have `webapp/` be part of the Python path.

It should be easy to test. If you run `make docs` and it complains, then that's the wrong thing.

relud: make docs only succeeds when i use `webapp.`, but even then it's outputting an empty configuration table when i try it:
Screenshot 2024-01-30 at 1 51 51 PM

willkg (changes requested):
I checked out the main branch and did:

```
$ cd webapp
$ ./manage.py diffsettings > settings_main.txt
```

Then did the same thing with this branch. Then I compared them:

```
diff -W 250 -y webapp/settings_main.py webapp/settings_everett.py
```

and the two are identical. Ergo, we're assured that no Django settings or values changed.

We still need to fix the Python path in `docs/conf.py` and the `automoduleconfig` directive and then the docs will work. But then this is good to go!

webapp/crashstats/settings/base.py
```diff
@@ -473,44 +506,71 @@ def filter(self, record):
...
-# `django.views.decorators.clickjacking.xframe_options_sameorigin`
-# decorator on specific views that can be in a frame.
-X_FRAME_OPTIONS = config("X_FRAME_OPTIONS", "DENY")
+X_FRAME_OPTIONS = _config(
+    "X_FRAME_OPTIONS",
+    default="DENY",
+    doc=(
+        "By default, we don't want to be inside a frame. "
+        "If you need to override this you can use the "
+        "`django.views.decorators.clickjacking.xframe_options_sameorigin` "
```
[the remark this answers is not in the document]
relud: single backtick does italic, double backtick does what i want, switched to that

docs/configuration.rst
relud: this fixed it, tyty

willkg (approved):
Looks super! Thank you!

PR #6535, opened 2024-02-26
bug-1850981: update to python 3.11

willkg:
This updates the socorro services to use Python 3.11.

Because we're using 3.11, we can drop some explicit requirements we had to fix a nuissance where Dependabot compiled requirements with Python 3.11 and dropped some backport packages we need.

Django 3.2 still imports cgi, so we add an ignore for that DeprecationWarning.

Note: Socorro doesn't have the VSCode devcontainer stuff set up, yet, so that didn't need to be updated to Python 3.11.


socorro/signature/siglists_utils.py
```diff
@@ -5,7 +5,10 @@
 from pathlib import Path
 import re
 
-import importlib_resources
+try:
+    from importlib import resources as importlib_resources
+except ImportError:
+    import importlib_resources
```
willkg, line 11: This allows the signature generation code to work with Python 3.8+.

relud (approved):

willkg:
Thank you!

PR #6538, opened 2024-02-29
bug-1809726: update to django 4.2

willkg:
This updates Socorro from Django 3.2 LTS to 4.2 LTS. This is a more involved update than Tecken was because Socorro uses more of Django's stuff for the web ui whereas Tecken is predominantly a react app.

This starts with updating Django which then required me to update django-pipeline. That unearthed some problems one of which was fixed by adjusting the bundle specification. The other required us to switch from one jsonview js lib which isn't maintained to another one which sort of is.

On top of that, Django deprecated a bunch of stuff that were were doing with the plan of removing those things in Django 5.0. So I ran `django-upgrade` and I also fixed some things by hand.

To test:

1. run `make build` which builds everything and also builds the static assets
2. run the tests
3. run the webapp, process some crash reports, click around
   1. check report view "Raw data and minidumps" and "Telemetry Environment" tabs which use the jsonview library
   2. check the admin


docker/Dockerfile
```diff
@@ -31,7 +31,7 @@ COPY --chown=app:app ./webapp/package*.json /webapp-frontend-deps/
 RUN cd /webapp-frontend-deps/ && npm install
 
 COPY --chown=app:app requirements.txt /app/
-RUN pip install -U 'pip==23.3.1' && \
+RUN pip install -U 'pip==24.0' && \
```
willkg, line 34: I got tired of the pip warning, so I updated pip, too.

webapp/crashstats/api/tests/test_views.py
```diff
@@ -110,7 +110,7 @@ def test_base_classes_raise_not_found(self):
     def test_option_CORS(self):
         """OPTIONS request for model_wrapper returns CORS headers"""
         url = reverse("api:model_wrapper", args=("NoOp",))
-        response = self.client.options(url, HTTP_ORIGIN="http://example.com")
+        response = self.client.options(url, headers={"origin": "http://example.com"})
```
willkg, line 113: The Django test client takes a `headers` argument instead of headers as kwargs. I don't know which Django version changed this--it's possible it's been this way for a while and we never updated.

webapp/crashstats/authentication/admin.py
```diff
@@ -38,12 +38,12 @@ class UserAdminBetter(UserAdmin):
         "get_policyexception",
     ]
 
+    @admin.display(description="Hacker (PII)")
```
willkg, line 41: Instead of setting properties on the methods, there's a decorator that you can use to set things. We switched to that.

webapp/crashstats/crashstats/admin.py
```diff
@@ -30,14 +30,14 @@ class LogEntryAdmin(admin.ModelAdmin):
...
-        "admin",
+        "admin_email",
         "object_link",
         "action",
         "get_change_message",
     ]
     list_display_links = ["action_time", "get_change_message"]
 
-    def admin(self, obj):
+    def admin_email(self, obj):
```
willkg, line 40: We needed to change this method name because it overrides `admin` in the class namespace preventing us from using `admin.display()` decorator.

webapp/crashstats/crashstats/static/crashstats/css/components/message.less
```diff
@@ -24,9 +24,9 @@ div#message {
 }
 .message.warning {
     background-color: @pink;
-    border: 1px solid @jsonview-bool;
+    border: 1px solid @orange;
```
willkg, line 27: I got rid of `jsonview-bool` when I updated the CSS for the jsonview. It's weird that the messages system was using that color. It was kind of a lighter tangerine color, so I replaced it with orange.

webapp/crashstats/crashstats/tests/test_views.py
```diff
@@ -201,7 +202,7 @@ def test_html(self):
         handler500 = getattr(views, end)
 
         # to make a mock call to the django view functions you need a request
-        fake_request = RequestFactory().request(**{"wsgi.input": None})
+        fake_request = RequestFactory().request(**{"wsgi.input": StringIO("")})
```
willkg, line 205: Turns out when you pass `None` in as a value, something in Django internals calls `.read()` on it and then everything gets mad. `wsgi.input` should be a stream, so I replaced it with a `StringIO` and that seemed to work.

webapp/crashstats/settings/base.py
```diff
@@ -320,11 +320,6 @@ def filter(self, record):
 }
 
 
-# No need to load it because we don't do i18n in this project
-USE_I18N = False
-
-USE_L10N = False
```
willkg, line 326: These are deprecated. I think that means Django always has localization enabled. We don't use that in Crash Stats, so I think this is probably fine and if not, someone will tell us and we can figure out what to do then.

webapp/crashstats/settings/base.py
```diff
@@ -470,6 +465,12 @@ def filter(self, record):
 DATABASES = {"default": dj_database_url.parse(DATABASE_URL)}
 
 
+STORAGES = {
```
willkg, line 468: This used to be `STATICFILES_STORAGE` and is now just `STORAGES` with a different structure.

https://docs.djangoproject.com/en/5.0/releases/4.2/#custom-file-storages

webapp/crashstats/settings/bundles.py
```diff
@@ -34,12 +34,15 @@
     "@fortawesome/fontawesome-free": ["css/all.min.css", "webfonts/*"],
     "tablesorter": ["dist/css/theme.default.min.css", "dist/js/jquery.tablesorter.js"],
     "d3": ["dist/*"],
-    "jssha": ["dist/*.js"],
+    "jssha": ["dist/*.js", "dist/*.map"],
```
willkg, line 37: The `.js` file has an indicator to the `.map` file and something uses that now (I have no idea what), so we need to copy over the `.map` file now.

webapp/crashstats/settings/bundles.py
```diff
@@ -34,12 +34,15 @@
...
     "qs": ["dist/*"],
     "moment": ["moment.js"],
     "filesize": ["lib/*"],
     "underscore": ["*.js"],
-    "jquery-jsonview": ["dist/*"],
+    "jquery.json-viewer": [
+        "json-viewer/jquery.json-viewer.css",
+        "json-viewer/jquery.json-viewer.js",
+    ],
```
willkg, line 45: jquery-jsonview has a file that indicates there's a `.map` file just like jssha did, but there is no `.map` file in the package. The library hasn't been touched in 8 years. So I switched us to jquery.json-viewer which has some minor differences, but it's been updated more recently and doesn't have the `.map` problem.

webapp/crashstats/supersearch/form_fields.py
```diff
@@ -206,7 +206,7 @@ class IsoDateTimeField(forms.DateTimeField):
     def to_python(self, value):
         if value:
             try:
-                return parse_isodate(value).replace(tzinfo=utc)
+                return parse_isodate(value).replace(tzinfo=datetime.timezone.utc)
```
willkg, line 209: `django.utils.timezone.utc` is deprecated in Django 4.1 in favor of `datetime.timezone.utc` which it was an alias for since Django 4.0.

http://docs.djangoproject.com/en/5.0/releases/4.1/

webapp/crashstats/tokens/jinja2/tokens/home.html
```diff
@@ -115,7 +115,7 @@ <h2>Generate a New Token</h2>
         <form method="post">
           {% csrf_token %}
           <table class="data-table">
-            {{ form }}
+            {{ form.as_table() }}
```
willkg, line 118: Django project created a new div form renderer that's accessibility friendly which the table form renderer is not. Calling `{{ form }}` will change what it's doing from the table form renderer it does now to the div form renderer. So we call `form.as_table()` here because I didn't want to spend time on redoing the API token form using the div form renderer and hunt for other forms that need to be similarly updated.

I'll write up a bug to cover that work.

http://docs.djangoproject.com/en/5.0/releases/4.1/

relud (approved):

webapp/crashstats/tokens/jinja2/tokens/home.html
willkg: https://bugzilla.mozilla.org/show_bug.cgi?id=1883044

willkg:
Thank you!

PR #6540, opened 2024-03-01
bug 1883055: stop using django-session-csrf

willkg:
We used django-session-csrf which is a Mozilla maintained library that we wrote back in the heady Mozilla webdev days which had properties our webdev and websec folks wanted. That was a long time ago, Django's built-in csrf module has improved over time, the world has changed (no more Flash), and the library hasn't been actively maintained in some time.

This drops the django-session-csrf and some of the interesting code around it in favor of Django's built-in csrf.

To test:

1. set up a local dev environment
2. process some crash reports
3. try a POST with no csrf token: `curl -X POST http://localhost:8000` -- should return an HTTP 403
4. open the webapp in your browser
5. while not logged in
   1. test quick search (top right-hand corner) with a crash id that you processed
   2. test supersearch, add a facet, add a column
   3. go to topcrashers site and change the filters
   4. go to signature report, add an aggregation, click around
   5. test an API on the API example page: http://localhost:8000/api/
6. log into webapp
   1. test quick search (top right-hand corner) with a crash id that you processed
   2. test supersearch, add a facet, add a column
   3. go to topcrashers site and change the filters
   4. go to signature report, add an aggregation, click around
   5. test an API on the API example page: http://localhost:8000/api/
   6. go to API token page, create a token, delete a token

I think that's most/all of the places where CSRF comes up.

### DIFF
--- a/webapp/crashstats/api/views.py
+++ b/webapp/crashstats/api/views.py
@@ -16,7 +16,6 @@ from rest_framework.response import Response
 from rest_framework import status
 from rest_framework.utils.urls import replace_query_param
 from rest_framework.views import APIView
-from session_csrf import anonymous_csrf_exempt
 
 from django import http
 from django import forms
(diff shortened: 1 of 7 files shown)


webapp/crashstats/api/views.py
```diff
@@ -134,33 +133,6 @@ def _clear_empty_session(request, *args, **kwargs):
     return _clear_empty_session
 
 
-def no_csrf_i_mean_it(fun):
```
willkg, line 137: I remember writing this. Anybody want a peanut?

willkg:
Thank you!

PR #6551, opened 2024-03-06
bug-1884002: Add devcontainer configuration

relud:
mostly copied from https://github.com/mozilla-services/antenna/pull/964/files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ devcontainerbuild: my.env  ## | Build VS Code development container.
 	touch .devcontainer-build
 
 .PHONY: devcontainer
-devcontainer: .env .devcontainer-build  ## | Run VS Code development container.
+devcontainer: my.env .devcontainer-build  ## | Run VS Code development container.
 	${DC} up --detach devcontainer
 
 .PHONY: setup
--- a/Makefile
+++ b/Makefile
@@ -44,12 +44,24 @@ my.env:
 .docker-build:
 	make build
 
+.devcontainer-build:
+	make devcontainerbuild
+
 .PHONY: build
 build: my.env  ## | Build docker images.
 	${DC} build ${DOCKER_BUILD_OPTS} --build-arg userid=${SOCORRO_UID} --build-arg groupid=${SOCORRO_GID} --progress plain app
 	${DC} build --progress plain oidcprovider fakesentry
 	${DC} build --progress plain statsd postgresql memcached localstack elasticsearch symbolsserver
 	touch .docker-build
+
+.PHONY: devcontainerbuild
+devcontainerbuild: my.env  ## | Build VS Code development container.
+	${DC} build devcontainer
+	touch .devcontainer-build
+
+.PHONY: devcontainer
+devcontainer: .env .devcontainer-build  ## | Run VS Code development container.
+	${DC} up --detach devcontainer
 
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -44,13 +44,25 @@ my.env:
...
+
 .PHONY: build
 build: my.env  ## | Build docker images.
 	${DC} build ${DOCKER_BUILD_OPTS} --build-arg userid=${SOCORRO_UID} --build-arg groupid=${SOCORRO_GID} --progress plain app
 	${DC} build --progress plain oidcprovider fakesentry
 	${DC} build --progress plain statsd postgresql memcached localstack elasticsearch symbolsserver
 	touch .docker-build
 
+.PHONY: devcontainerbuild
+devcontainerbuild: my.env  ## | Build VS Code development container.
```
biancadanforth, line 58: Aside: I noticed the other Crash Ingestion repos just use `.env`, but Socorro uses `my.env`. Doesn't matter really as long as your filename matches the filename it's looking for. Just a minor inconsistency across the repos.

Makefile
```diff
@@ -44,13 +44,25 @@ my.env:
...
+	${DC} build devcontainer
+	touch .devcontainer-build
+
+.PHONY: devcontainer
+devcontainer: .env .devcontainer-build  ## | Run VS Code development container.
```
biancadanforth, line 63: FWICT trying this locally, `.devcontainer-build` is empty (it's created in the above `devcontainerbuild` command) and doesn't appear to do anything -- is it doing something?

biancadanforth (approved):
WFM! Just have one question about the purpose of `.devcontainer-build`...

Makefile
relud: to my knowledge absolutely nothing, I just copied the pattern from `.docker-build` assuming it was for make reasons.

@willkg can you elaborate on why `.docker-build` exists?

Makefile
relud: i think it's worth standardizing on `.env` because that also automatically pulled in by docker compose, but definitely a concern for a separate bug. I'll file one and link it here shortly.

Makefile
willkg: `.docker-build` is a real file on the file system and has a rule for building it. As such, you can use it as a dependency in Make rules and if it doesn't exist, Make will run the `.docker-build` rule which will build the container.

Makefile
willkg: Tecken uses `.env` because that's what peterbe did. Eliot uses `.env` because I extracted it out of Tecken.

Socorro and Antenna use `my.env` instead because I don't like using `.env` for service things because so many other things (docker-compose, honcho, etc) use `.env` that it sometimes creates problems. Plus `.env` is a dotfile so it doesn't show up in listings.

Having said that, we should probably standardize on `.env` and suffer whatever issues come up.

Makefile
willkg: For example, you can start with a completely clean repo and do `make test` and Make will see there's no `.docker-build` file, run that rule building everything, and then run the tests.

Makefile
relud: filed https://bugzilla.mozilla.org/show_bug.cgi?id=1884446 about changing to `.env` here and in antenna

PR #6556, opened 2024-03-12
bug-1867814: remove irrelevant comment

willkg:
Previously `CrashAnnotations.yaml` listed `ipc_system_error` as a string, but Socorro treated it as an integer. We had a note because of the discrepancy.

That's no longer true--`CrashAnnotations.yaml` has been fixed. We can remove the note.


relud (approved):

willkg:
Thank you!

PR #6560, opened 2024-03-20
bug-1885978: remove raw crash from indexing code

willkg:
In bug 1763264, we finished all the work to stop using the raw crash when indexing crash report data. The only data we're indexing now comes from the processed crash after being normalized and validated by processor rules. This radically reduces problems with indexing from weird values.

This cleans up the remnants of those old days by removing `raw_crash` from building documents, indexing, and related documentation. This also removes a deepcopy call which will speed up indexing a little.

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -216,7 +216,7 @@ def fix_datetime(value):
 def build_document(src, crash_document, fields, all_keys):
     """Given a source document and fields and valid keys, builds a document to index.
 
-    :param dict src: the source document with raw_crash and processed_crash keys
+    :param dict src: the source document with "processed_crash" key
     :param dict crash_document: the document to fill
     :param list fields: the list of fields in super search fields
     :param set all_keys: the list of valid keys
@@ -270,7 +270,7 @@ def build_document(src, crash_document, fields, all_keys):
 
 
 class ESCrashStorage(CrashStorageBase):
-    """This sends raw and processed crash reports to Elasticsearch."""
+    """Indexes documents based on the processed crash to Elasticsearch."""
 
     # These regex will catch field names from Elasticsearch exceptions. They
     # have been tested with Elasticsearch 1.4.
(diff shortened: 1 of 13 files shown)


socorro/external/es/crashstorage.py
```diff
@@ -498,14 +498,10 @@ def save_processed_crash(self, raw_crash, processed_crash):
         es_doctype = self.get_doctype()
         all_valid_keys = self.get_keys(index_name, es_doctype)
 
-        src = {
-            "raw_crash": copy.deepcopy(raw_crash),
```
willkg, line 502: One less deepcopy should make it a little faster, though I'm not sure we'd notice.

socorro/tests/external/es/test_crashstorage.py
```diff
@@ -143,11 +137,8 @@ def test_index_crash_indexable_keys(self, es_helper):
                 id=processed_crash["uuid"],
             )
 
-        # Verify keys that aren't in super_search_fields aren't in the raw or processed
-        # crash parts
-        raw_crash = doc["_source"]["raw_crash"]
-        assert list(sorted(raw_crash.keys())) == []
```
willkg, line 149: One less "sort an empty list" in the codebase. :tada:

socorro/tests/external/es/test_crashstorage.py
```diff
@@ -250,13 +234,12 @@ def test_crash_size_capture(self):
...
             crashstorage.save_processed_crash(
-                raw_crash=raw_crash,
+                raw_crash={},
                 processed_crash=processed_crash,
             )
 
-            mm.assert_histogram("processor.es.raw_crash_size", value=2)
             mm.assert_histogram("processor.es.processed_crash_size", value=96)
-            mm.assert_histogram("processor.es.crash_document_size", value=186)
+            mm.assert_histogram("processor.es.crash_document_size", value=169)
```
willkg, line 242: This value changes because the document no longer contains a `"raw_crash":{},`.

willkg: Now that I'm thinking about this, we could stop emitting the processed crash size. The "Socorro prod app metrics" dashboard only has a panel for the crash document size. 

I'll fix this now--one less metric.

socorro/tests/external/es/test_super_search_fields.py
```diff
@@ -218,8 +217,7 @@ def test_validate_super_search_fields(name, properties):
         if properties.get("destination_keys"):
             for key in properties["destination_keys"]:
                 possible_keys = [
-                    # Old keys we're probably migrating from
-                    f"raw_crash.{properties['in_database_name']}",
```
willkg, line 222: None of the keys start with `raw_crash` and we want to make sure that's true going forward. We only want to index values that have been normalized and validated and those are in the processed crash.

relud (approved):

PR #6563, opened 2024-03-28
bug 1849266: change generate_create_bug_url to truncate symbols and refactor

willkg:
This changes how frame symbols in a stack are rendered in the bug comment from dropping function arguments (which was buggy and didn't work right in various cases) to truncating at 80 characters.

This also refactors how `generate_create_bug_url` works. It takes out the jinja template from generating the comment and some of the hacks we had to alleviate issues with that.

It reimplements that in a libbugzilla library. It also radically cleans up the arguments being passed around so now `generate_create_bug_url` only requires a `crash_report_url`, a template, and a processed crash. That's a lot less nebulous than what we had before and makes it much easier for us to extend and for others to extract for external uses.

https://crash-stats.mozilla.org/report/index/098c2d8a-5f54-436f-8e8b-ea2d20240327

Before:

```
0  xul.dll  MOZ_Crash  mfbt/Assertions.h:301
0  xul.dll  Abort  xpcom/base/nsDebugImpl.cpp:513
0  xul.dll  NS_DebugBreak  xpcom/base/nsDebugImpl.cpp:496
1  xul.dll  nsDebugImpl::Abort  xpcom/base/nsDebugImpl.cpp:129
2  xul.dll  XPTC__InvokebyIndex  
3  xul.dll  NS_InvokeByIndex  xpcom/reflect/xptcall/md/win32/xptcinvoke_x86_64.cpp:57
3  xul.dll  CallMethodHelper::Invoke  js/xpconnect/src/XPCWrappedNative.cpp:1621
3  xul.dll  CallMethodHelper::Call  js/xpconnect/src/XPCWrappedNative.cpp:1174
3  xul.dll  XPCWrappedNative::CallMethod  js/xpconnect/src/XPCWrappedNative.cpp:1120
4  xul.dll  XPC_WN_CallMethod  js/xpconnect/src/XPCWrappedNativeJSOps.cpp:966
```

After:

```
0  xul.dll  MOZ_Crash(char const*, int, char const*)  mfbt/Assertions.h:301
0  xul.dll  Abort(char const*)  xpcom/base/nsDebugImpl.cpp:513
0  xul.dll  NS_DebugBreak(unsigned int, char const*, char const*, char const*, int)  xpcom/base/nsDebugImpl.cpp:496
1  xul.dll  nsDebugImpl::Abort(char const*, int)  xpcom/base/nsDebugImpl.cpp:129
2  xul.dll  XPTC__InvokebyIndex
3  xul.dll  NS_InvokeByIndex(nsISupports*, unsigned int, unsigned int, nsXPTCVariant*)  xpcom/reflect/xptcall/md/win32/xptcinvoke_x86_64.cpp:57
3  xul.dll  CallMethodHelper::Invoke()  js/xpconnect/src/XPCWrappedNative.cpp:1621
3  xul.dll  CallMethodHelper::Call()  js/xpconnect/src/XPCWrappedNative.cpp:1174
3  xul.dll  XPCWrappedNative::CallMethod(XPCCallContext&, XPCWrappedNative::CallMode)  js/xpconnect/src/XPCWrappedNative.cpp:1120
4  xul.dll  XPC_WN_CallMethod(JSContext*, unsigned int, JS::Value*)  js/xpconnect/src/XPCWrappedNativeJSOps.cpp:966
```

Here's one with truncation:

https://crash-stats.mozilla.org/report/index/4ba8e300-c512-42f0-8716-7965b0240327

```
0  xul.dll  mozilla::dom::quota::QuotaManager::Shutdown::<lambda_0>::operator()(nsITimer*...  dom/quota/ActorsParent.cpp:2273
0  xul.dll  mozilla::dom::quota::QuotaManager::Shutdown::<lambda_0>::__invoke(nsITimer*, ...  dom/quota/ActorsParent.cpp:2238
1  xul.dll  nsTimerImpl::Fire::<lambda_14>::operator()(nsTimerImpl::FuncCallback const&) ...  xpcom/threads/nsTimerImpl.cpp:681
1  xul.dll  mozilla::detail::VariantImplementation<unsigned char, 3, nsTimerImpl::FuncCal...  mfbt/Variant.h:309
1  xul.dll  mozilla::detail::VariantImplementation<unsigned char, 2, nsCOMPtr<nsIObserver...  mfbt/Variant.h:318
1  xul.dll  mozilla::detail::VariantImplementation<unsigned char, 1, nsCOMPtr<nsITimerCal...  mfbt/Variant.h:318
1  xul.dll  mozilla::detail::VariantImplementation<unsigned char, 0, nsTimerImpl::Unknown...  mfbt/Variant.h:318
1  xul.dll  mozilla::Variant<nsTimerImpl::UnknownCallback, nsCOMPtr<nsITimerCallback>, ns...  mfbt/Variant.h:902
1  xul.dll  mozilla::Variant<nsTimerImpl::UnknownCallback, nsCOMPtr<nsITimerCallback>, ns...  mfbt/Variant.h:857
1  xul.dll  nsTimerImpl::Fire(int)  xpcom/threads/nsTimerImpl.cpp:675
```

**Why refactor this?** Several reasons:

1. The way the code originally worked is that the `report_index` view renders the `report_index.html` template which calls the `generate_create_bug_url` jinja function which rendered the `bug_comment.txt` template and then generated a bugizlla url to create a bug with pre-filled fields. It's complicated and it's hard to understand which bits come from where. As crash reports get more complex, it's harder to support new situations.
2. The code is _really_ hard to maintain to the point where I hate touching it. To create the bug comment, it uses Jinja which is a templating engine meant for HTML where whitespace (mostly) doesn't matter for producing a string where whitespaces _does_ matter. There are multiple scenarios of crash reports we need to support, so there are a bunch of if/then blocks. Because of that, the `bug_comment.txt` form is bonkers.
3. The current test code is mediocre because it tests existence or non-existence of substrings and we have no idea if changes to the `bug_comment.txt` form affect the structure of the comment.
4. Suhaib wants to use the same code for bugbot. It behooves us to create a module that he can copy that takes arguments available in the context bugbot is using and produces a good comment.

**Why refactor it in this PR?** It was hard to test the changes in the previous situation and I've got another bug I want to work on for better Fenix support, so I decided to do it here.


webapp/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -157,21 +147,17 @@ def show_bug_link(bug_id):
...
 MAX_TITLE_LENGTH = 255
 
 
 @library.global_function
-def generate_create_bug_url(
-    request, template, raw_crash, report, parsed_dump, crashing_thread
-):
-    parsed_dump = parsed_dump or {}
-
+def generate_create_bug_url(crash_report_url, template, processed_crash):
```
willkg, line 154: I made big changes to the signature here:

I don't want to pass the request into a utility function because that requires us to create a lot of HTTP state which we shouldn't be requiring for a utility function. By dropping the request, it's possible for someone else (like Suhaib) to copy our link generation code verbatim without having to make changes to it.

However, we need the full url to the crash report which depends on service details that are in the request--the scheme and host. Thus, we now pass in the fully-formed crash report url and the caller can generate that and the utility function doesn't have to know anything about requests.

This function and downstream functions don't use the raw crash data, so I removed that from the arguments.

The parsed dump is really the `processed_crash['json_dump']` with a bunch of processing that gets used on the Details page of the crash report, but no where else--including in `generate_create_bug_url`, so we can use `processed_crash['json_dump']` and drop `parsed_dump` from the arguments.

webapp/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -183,39 +169,14 @@ def generate_create_bug_url(
     elif op_sys in ("Windows Unknown", "Windows 2000"):
         op_sys = "Windows"
 
-    frames = None
-    threads = parsed_dump.get("threads")
-    if threads:
-        thread_index = crashing_thread or 0
-        frames = bugzilla_thread_frames(parsed_dump["threads"][thread_index])
-
-    comment = render_to_string(
```
willkg, line 192: Rendering to a string using Jinja templates was a complete and utter nightmare. I ditched the template and rewrote the code to build the string using Python.

webapp/crashstats/crashstats/templatetags/jinja_helpers.py
```diff
@@ -183,39 +169,14 @@ def generate_create_bug_url(
...
-            "moz_crash_reason": report.get("moz_crash_reason", None),
-            "reason": report.get("reason", None),
-            "frames": frames,
-            "crashing_thread": crashing_thread,
-        },
-    )
-
-    # Whitespace is a nightmare when using Jinja2 templates to render text that's not
-    # destined for HTML. So we do a pass on removing extra line endings.
-    comment = EXTRA_NEWLINES_RE.sub("\n\n", comment)
```
willkg, line 211: Because we're not using the Hellraiser-inspired pain firehose that is building whitespace-sensitive strings using Jinja templates, we don't need to do this anymore.

webapp/crashstats/crashstats/tests/test_jinja_helpers.py
```diff
@@ -113,7 +113,6 @@ def test_show_bug_link_with_cache(self):
 
 
 class Test_generate_create_bug_url:
-    EMPTY_PARSED_DUMP = {}
```
willkg, line 116: Most of the changes in this file are because we changed the signature for `generate_create_bug_url` removing the raw crash and request and using the `json_dump` value from the processed crash.

webapp/crashstats/crashstats/tests/test_jinja_helpers.py
```diff
@@ -126,26 +125,25 @@ class Test_generate_create_bug_url:
...
         + "format=__default__"
     )
+    CRASH_ID = "70dda764-a402-4ca3-b806-c38dd0240328"
 
-    @staticmethod
-    def _create_report(**overrides):
+    def _create_report(self, **overrides):
         default = {
             "signature": "$&#;deadbeef",
-            "uuid": "00000000-0000-0000-0000-000000000000",
```
willkg, line 134: This is not a valid crash id. I generated a valid crash id and switched everything to use that. This will save future us from having to deal with consequences of invalid test data.

webapp/crashstats/crashstats/tests/test_jinja_helpers.py
```diff
@@ -126,26 +125,25 @@ class Test_generate_create_bug_url:
...
+            "uuid": self.CRASH_ID,
             "cpu_arch": "x86",
             "os_name": None,
         }
         return dict(default, **overrides)
 
-    @staticmethod
-    def _extract_query_string(url):
```
willkg, line 141: There's no need for these to be static methods and when they are static methods, they can't access instance properties/methods. So I changed them.

webapp/crashstats/crashstats/tests/test_jinja_helpers.py
```diff
@@ -165,139 +163,131 @@ def _create_frame(
...
         assert qs["format"] == ["__default__"]
         assert qs["product"] == ["Firefox"]
         assert qs["rep_platform"] == ["x86"]
         assert qs["short_desc"] == ["Crash in [@ $&#;deadbeef]"]
         assert qs["op_sys"] == ["Windows"]
         assert qs["bug_type"] == ["defect"]
+        comment_lines = [
+            f"Crash report: http://localhost:8000/report/index/{self.CRASH_ID}"
+        ]
+        comment = "\n".join(comment_lines)
```
willkg, line 192: I should make this less silly given there's only one line.

webapp/crashstats/crashstats/tests/test_jinja_helpers.py
```diff
@@ -329,343 +313,448 @@ def test_with_unicode_signature(self):
...
+        qs = self._extract_query_string(url)
+        comment = (
+            "Crash report: http://localhost:8000/report/index/70dda764-a402-4ca3-b806-c38dd0240328\n"
+            "\n"
+            "Top 3 frames of crashing thread:\n"
+            "```\n"
+            "0  fake_module  foo::bar(char* x, int y)  fake.cpp:1\n"
+            "1  fake_module  foo::bar(char* x, int y)  fake.cpp:1\n"
+            "2  fake_module  foo::bar(char* x, int y)  fake.cpp:1\n"
+            "```"
```
willkg, line 346: I changed all the comment tests so that instead of asserting a substring in a url which is quoted which is impossible to infer anything useful about and excrutiating to verify, it's now explicit and sensible.

webapp/crashstats/crashstats/tests/test_jinja_helpers.py
```diff
@@ -329,343 +313,448 @@ def test_with_unicode_signature(self):
...
 
     def test_comment_missing_everything(self):
         """If a frame is missing everything, do not throw an error."""
-        req = RequestFactory().get("/report/index")
-        raw_crash = {}
-        report = self._create_report()
-        parsed_dump = self._create_dump(threads=[self._create_thread(frames=[{}])])
-        generate_create_bug_url(req, self.TEMPLATE, raw_crash, report, parsed_dump, 0)
-
-    def test_comment_no_html_escaping(self):
```
willkg, line 544: We needed this test to make sure that the rendered string from Jinja didn't escape all the HTML which is what Jinja does because Jinja expects to be rendering a string _for HTML_ and not _to be put into a url to prefill a bug comment form_. Given we're not using Jinja anymore, we don't need this.

webapp/crashstats/crashstats/tests/test_jinja_helpers.py
```diff
@@ -329,343 +313,448 @@ def test_with_unicode_signature(self):
...
+            "5  fake_module  foo::bar(char* x, int y)  fake.cpp:1\n"
+            "6  fake_module  foo::bar(char* x, int y)  fake.cpp:1\n"
+            "7  fake_module  foo::bar(char* x, int y)  fake.cpp:1\n"
+            "8  fake_module  foo::bar(char* x, int y)  fake.cpp:1\n"
+            "9  fake_module  foo::bar(char* x, int y)  fake.cpp:1\n"
+            "```"
+        )
+        assert qs["comment"][0] == comment
+
+    def test_comment_frame_long_signature(self):
```
willkg, line 402: This is a new test that tests truncation of long frame signatures.

biancadanforth (approved):
As you noted, the code here goes a few layers deep, so it took me a while to understand what you were changing. I really appreciate your detailed comments and explanations.

I know you've already made a ton of improvements here going way beyond what was asked for in the initial ticket, but noticed a couple of things about the `frames` output that could also potentially be improved (not blocking of course, and maybe you've already considered doing this in the past):
* Left align the values in each column, otherwise it's hard to see e.g. where Source starts following the Signature for non-truncated frames.
* Add column headers (i.e. Frame, Module, Signature, Source, Trust IIUC).

webapp/crashstats/crashstats/templatetags/jinja_helpers.py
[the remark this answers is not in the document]
willkg: It's not wildly clear but a "jinja function" is a Python function that's wrapped with the decorator that registers the function with the Jinja engine so it can be used in templates.

webapp/crashstats/crashstats/templatetags/jinja_helpers.py
willkg: Yeah. Jinja has some operators that let you strip whitespace in various directions, but the `bug_comment.txt` is complex enough I never got them to work exactly right. We'd end up with extra newlines in random places and then someone would complain.

willkg:
Thank you! You have done a great deed this day. Together we have banished the wicked and dishonorable `bug_comment.txt` from our universe. The world is a little bit more pleasant now.

PR #6566, opened 2024-04-01
bug-1884041: support java_exception in crash_report_to_description

willkg:
This adds support for java_exception field when generating the
description for creating new bugs.

This also creates `webapp/crashstats/tests/test_libbugzilla.py` and
moves a bunch of tests from `test_jinja_helpers.py` so that
`test_libbugzilla.py` tests all the requirements for the functions.

I also switched the comment-building tests to use dedent instead of
string delimiters, newlines, and such. It's a little easier to
manipulate and a lot easier to read and verify.

This also re-does `test_corrected_os_version_name` so it's parametrized.

### DIFF
--- a/webapp/crashstats/crashstats/tests/test_jinja_helpers.py
+++ b/webapp/crashstats/crashstats/tests/test_jinja_helpers.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 import datetime
+from textwrap import dedent
 import time
 from urllib.parse import parse_qs, urlsplit
 
@@ -139,36 +140,6 @@ class Test_generate_create_bug_url:
     def _extract_query_string(self, url):
         return parse_qs(urlsplit(url).query)
 
-    def _create_frame(
-        self,
-        frame=1,
-        module="fake_module",
-        signature="foo::bar(char* x, int y)",
-        file="fake.cpp",
-        line=1,
-        inlines=None,
-        unloaded_modules=None,
-    ):
-        data = {
-            "frame": frame,
-            "module": module,
-            "signature": signature,
-            "file": file,
-            "line": line,
-        }
-        if inlines:
-            data["inlines"] = inlines
-        if unloaded_modules:
-            data["unloaded_modules"] = unloaded_modules
-
-        return data
-
-    def _create_thread(self, frames=None):
-        return {"frames": frames or []}
-
-    def _create_dump(self, threads=None):
-        return {"threads": threads or []}
-
     def test_basic_url(self):
         report = self._create_report(
             os_name="Windows", crashing_thread=self.CRASHING_THREAD
@@ -186,8 +157,13 @@ class Test_generate_create_bug_url:
         assert qs["short_desc"] == ["Crash in [@ $&#;deadbeef]"]
         assert qs["op_sys"] == ["Windows"]
         assert qs["bug_type"] == ["defect"]
-        comment = f"Crash report: http://localhost:8000/report/index/{self.CRASH_ID}"
-        assert qs["comment"] == [comment]
+        comment = dedent(
+            f"""\
+            Crash report: http://localhost:8000/report/index/{self.CRASH_ID}
+
+            No stack."""
+        )
+        assert qs["comment"][0] == comment
 
     def test_truncate_short_desc(self):
         report = self._create_report(
@@ -204,10 +180,24 @@ class Test_generate_create_bug_url:
         assert len(qs["short_desc"][0]) == 255
         assert qs["short_desc"][0].endswith("...")
 
-    def test_corrected_os_version_name(self):
+    @pytest.mark.parametrize(
+        "os_name, os_pretty_version, op_sys",
+        [
+            ("Windoooosws", "Windows 10", "Windows 10"),
+            # os_name if the os_pretty_version is there, but empty
+            ("Windoooosws", "", "Windoooosws"),
+            # "OS X <Number>" becomes "macOS"
+            ("OS X", "OS X 11.1", "macOS"),
+            # "Windows 8.1" becomes "Windows 8"
+            ("Windows NT", "Windows 8.1", "Windows 8"),
+            # "Windows Unknown" becomes plain "Windows"
+            ("Windows NT", "Windows Unknown", "Windows"),
+        ],
+    )
+    def test_corrected_os_version_name(self, os_name, os_pretty_version, op_sys):
         report = self._create_report(
-            os_name="Windoooosws",
-            os_pretty_version="Windows 10",
+            os_name=os_name,
+            os_pretty_version=os_pretty_version,
             crashing_thread=self.CRASHING_THREAD,
         )
         url = generate_create_bug_url(
@@ -216,69 +206,12 @@ class Test_generate_create_bug_url:
             report,
         )
         qs = self._extract_query_string(url)
-        assert qs["op_sys"] == ["Windows 10"]
-
-        # os_name if the os_pretty_version is there, but empty
-        report = self._create_report(
-            os_name="Windoooosws",
-            os_pretty_version="",
-            crashing_thread=self.CRASHING_THREAD,
-        )
-        url = generate_create_bug_url(
-            f"http://localhost:8000/report/index/{self.CRASH_ID}",
-            self.TEMPLATE,
-            report,
-        )
-        qs = self._extract_query_string(url)
-        assert qs["op_sys"] == ["Windoooosws"]
-
-        # "OS X <Number>" becomes "macOS"
-        report = self._create_report(
-            os_name="OS X",
-            os_pretty_version="OS X 11.1",
-            crashing_thread=self.CRASHING_THREAD,
-        )
-        url = generate_create_bug_url(
-            f"http://localhost:8000/report/index/{self.CRASH_ID}",
-            self.TEMPLATE,
-            report,
-        )
-        qs = self._extract_query_string(url)
-        assert qs["op_sys"] == ["macOS"]
-
-        # "Windows 8.1" becomes "Windows 8"
-        report = self._create_report(
-            os_name="Windows NT",
-            os_pretty_version="Windows 8.1",
-            crashing_thread=self.CRASHING_THREAD,
-        )
-        url = generate_create_bug_url(
-            f"http://localhost:8000/report/index/{self.CRASH_ID}",
-            self.TEMPLATE,
-            report,
-        )
-        qs = self._extract_query_string(url)
-        assert qs["op_sys"] == ["Windows 8"]
-
-        # "Windows Unknown" becomes plain "Windows"
-        report = self._create_report(
-            os_name="Windows NT",
-            os_pretty_version="Windows Unknown",
-            crashing_thread=self.CRASHING_THREAD,
-        )
-        url = generate_create_bug_url(
-            f"http://localhost:8000/report/index/{self.CRASH_ID}",
-            self.TEMPLATE,
-            report,
-        )
-        qs = self._extract_query_string(url)
-        assert qs["op_sys"] == ["Windows"]
+        assert qs["op_sys"] == [op_sys]
 
     def test_with_os_name_is_null(self):
         """Some processed crashes have a os_name but it's null."""
         report = self._create_report(
             os_name=None,
-            signature="java.lang.IllegalStateException",
             crashing_thread=self.CRASHING_THREAD,
         )
         url = generate_create_bug_url(
@@ -314,19 +247,36 @@ class Test_generate_create_bug_url:
 
     def test_comment(self):
         report = self._create_report(
-            crashing_thread=1,
-            json_dump=self._create_dump(
-                threads=[
-                    self._create_thread(),  # Empty thread 0
-                    self._create_thread(
-                        frames=[
-                            self._create_frame(frame=0),
-                            self._create_frame(frame=1),
-                            self._create_frame(frame=2),
+            crashing_thread=0,
+            json_dump={
+                "threads": [
+                    {
+                        "frames": [
+                            {
+                                "frame": 0,
+                                "module": "fake_module",
+                                "signature": "foo::bar(char* x, int y)",
+                                "file": "fake.cpp",
+                                "line": 10,
+                            },
+                            {
+                                "frame": 1,
+                                "module": "fake_module",
+                                "signature": "foo::bar(char* x, int y)",
+                                "file": "fake.cpp",
+                                "line": 20,
+                            },
+                            {
+                                "frame": 2,
+                                "module": "fake_module",
+                                "signature": "foo::bar(char* x, int y)",
+                                "file": "fake.cpp",
+                                "line": 30,
+                            },
                         ]
-                    ),
+                    },
                 ]
-            ),
+            },
         )
         url = generate_create_bug_url(
             f"http://localhost:8000/report/index/{self.CRASH_ID}",
@@ -335,21 +285,21 @@ class Test_generate_create_bug_url:
         )
 
         qs = self._extract_query_string(url)
-        comment = (
-            "Crash report: http://localhost:8000/report/index/70dda764-a402-4ca3-b806-c38dd0240328\n"
-            "\n"
-            "Top 3 frames of crashing thread:\n"
-            "```\n"
-            "0  fake_module  foo::bar(char* x, int y)  fake.cpp:1\n"
-            "1  fake_module  foo::bar(char* x, int y)  fake.cpp:1\n"
-            "2  fake_module  foo::bar(char* x, int y)  fake.cpp:1\n"
-            "```"
+        comment = dedent(
+            f"""\
+            Crash report: http://localhost:8000/report/index/{self.CRASH_ID}
+
+            Top 3 frames:
+            ```
+            0  fake_module  foo::bar(char* x, int y)  fake.cpp:10
+            1  fake_module  foo::bar(char* x, int y)  fake.cpp:20
+            2  fake_module  foo::bar(char* x, int y)  fake.cpp:30
+            ```"""
         )
         assert qs["comment"][0] == comment
 
-    def test_comment_no_threads(self):
-        """If json_dump has no threads available, do not output any frames."""
-        report = self._create_report(crashing_thread=0)
+    def test_comment_from_missing_data(self):
+        report = self._create_report()
         url = generate_create_bug_url(
             f"http://localhost:8000/report/index/{self.CRASH_ID}",
             self.TEMPLATE,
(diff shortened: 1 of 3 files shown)


webapp/crashstats/crashstats/tests/test_jinja_helpers.py
```diff
@@ -139,36 +140,6 @@ def _create_report(self, **overrides):
     def _extract_query_string(self, url):
         return parse_qs(urlsplit(url).query)
 
-    def _create_frame(
```
willkg, line 142: These kinds of methods make it difficult to read and verify tests because it's not clear how they're setting up the test scenario. Since I moved most of the tests that these were used in to `test_libbugzilla.py`, I fixed the ones that remained in here to not uses these methods and then removed the methods.

webapp/crashstats/crashstats/tests/test_jinja_helpers.py
```diff
@@ -186,8 +157,13 @@ def test_basic_url(self):
         assert qs["short_desc"] == ["Crash in [@ $&#;deadbeef]"]
         assert qs["op_sys"] == ["Windows"]
         assert qs["bug_type"] == ["defect"]
-        comment = f"Crash report: http://localhost:8000/report/index/{self.CRASH_ID}"
-        assert qs["comment"] == [comment]
+        comment = dedent(
```
willkg, line 160: I switched these test assertions to build the expected value using dedent and triple-quoted Python strings. The Python compiler will parse the triple-quoted and keep all the characters from the source file including the whitespace (spaces, newlines, tabs and everything) in the value. The `\` at the end of a line tells the Python compiler to ignore the newline, so that's _not_ part of the value.

When the Python interpreter runs this code, the dedent function will take that string and de-dent it by the number of leading spaces in the first line.

Thus with this source file:

```python
        comment = dedent(
            f"""\
            Crash report: http://localhost:8000/report/index/{self.CRASH_ID}

            No stack."""
        )
```

the `comment` variable points to this string value:

```
Crash report: http://localhost:8000/report/index/70dda764-a402-4ca3-b806-c38dd0240328

No stack.
```

webapp/crashstats/crashstats/tests/test_jinja_helpers.py
```diff
@@ -204,66 +180,24 @@ def test_truncate_short_desc(self):
...
+            ("Windoooosws", "", "Windoooosws"),
+            # "OS X <Number>" becomes "macOS"
+            ("OS X", "OS X 11.1", "macOS"),
+            # "Windows 8.1" becomes "Windows 8"
+            ("Windows NT", "Windows 8.1", "Windows 8"),
+            # "Windows Unknown" becomes plain "Windows"
+            ("Windows NT", "Windows Unknown", "Windows"),
+        ],
+    )
+    def test_corrected_os_version_name(self, os_name, os_pretty_version, op_sys):
```
willkg, line 197: I cleaned up these tests to use `pytest.mark.parametrize` because it's some boilerplate repeated a whole bunch of times.

webapp/crashstats/crashstats/tests/test_jinja_helpers.py
```diff
@@ -313,112 +246,37 @@ def test_with_unicode_signature(self):
...
-                    self._create_thread(
-                        frames=[
-                            self._create_frame(
-                                frame=0,
-                                module="test_module",
-                                signature=long_signature,
-                                file="foo.cpp",
-                                line=7,
-                            )
+            json_dump={
```
willkg, line 251: The diffing here stinks, but this is the rest of the `test_comment` test case.

webapp/crashstats/libbugzilla.py
```diff
@@ -80,40 +106,56 @@ def crash_report_to_description(crash_report_url, processed_crash):
...
+                lines.append("No crashing thread identified; using thread 0.")
+
+            thread_index = processed_crash.get("crashing_thread") or 0
+            frames = minidump_thread_to_frames(threads[thread_index])
+
+    elif "java_exception" in processed_crash:
+        stacktraces = (
+            processed_crash["java_exception"].get("exception", {}).get("values", [])
+        )
+        print(stacktraces)
```
willkg, line 125: Oops--need to remove these. Also, this shouldn't be in an elif--I thought I fixed that.

webapp/crashstats/libbugzilla.py
```diff
@@ -50,12 +51,10 @@ def frame_generator(thread):
     for frame in islice(frame_generator(thread), MAX_FRAMES):
         # Source is an empty string if data isn't available
         source = frame.get("file") or ""
-        if frame.get("line"):
-            source += ":{}".format(frame["line"])
+        if source and frame.get("line") is not None:
+            source = f"{source}:{frame['line']}"
```
willkg, line 55: I redid how these get constructed to ditch the `.format()` call and also fix the situation where there isn't a file, but there is a line. Previously, that would result with a source of something like `:20` which is weird. Twentieth line of what exactly?

Does that happen in the wild? I have no idea and no good way to hunt for an instance. But it is what the previous implementation did.

webapp/crashstats/libbugzilla.py
```diff
@@ -69,6 +68,33 @@ def frame_generator(thread):
     return frames
 
 
+def java_exception_to_frames(stack):
```
willkg, line 71: This is new. I modeled it after the `minidump_thread_to_frames`, but I think we're guaranteed to have actual values for the different fields. Instead of verifying that by looking at the crash reporter code for Android, I figured we could be defensive and handle the cases where data may not exist.

webapp/crashstats/libbugzilla.py
```diff
@@ -80,40 +106,55 @@ def crash_report_to_description(crash_report_url, processed_crash):
         lines.append("")
         lines.append(f"Reason: ```{processed_crash['reason']}```")
 
-    # If there's a java_stack_trace, add that; this is a big string blob, so we truncate
-    # it at 5,000 characters but don't otherwise format it
-    if processed_crash.get("java_stack_trace"):
```
willkg, line 85: Previously, this would prefer `java_stack_trace` over `java_exception` or processed minidump values. We want it the other way around--processed minidump takes precedence, followed by `java_exception`, followed by `java_stack_trace` if the other two didn't produce anything.

We're phasing out `java_stack_trace`, so at some point, we can remove the `java_stack_trace` nonsense altogether.

webapp/crashstats/libbugzilla.py
```diff
@@ -80,40 +106,55 @@ def crash_report_to_description(crash_report_url, processed_crash):
...
+        lines.append("```")
+        lines.append(truncate(java_stack_trace, 5000))
+        lines.append("```")
 
-        else:
-            lines.append("")
-            lines.append("No stack.")
+    else:
+        lines.append("")
+        lines.append("No stack.")
```
willkg, line 157: I changed things around so that if there isn't anything, then it now adds `No stack.`. Previously, it would have nothing which is weird and it's unclear what happened.

webapp/crashstats/tests/test_libbugzilla.py
```diff
@@ -0,0 +1,677 @@
...
+    ],
+)
+def test_truncate(text, length, expected):
+    assert truncate(text, length) == expected
+
+
+@pytest.mark.parametrize(
+    "thread, expected",
+    [
+        pytest.param(
```
willkg, line 31: One problem with this test code is we don't know if it's valid data and we won't know if things change in the future such that it becomes invalid data at some point.

[Bug 1795699](https://bugzilla.mozilla.org/show_bug.cgi?id=1795699) covers implementing some bits to make it possible to validate test data as valid processed crash data. We can then use that in the tests to make sure that any test data we have doesn't become invalid in the future due to changes in the schema. Some day....

webapp/crashstats/tests/test_libbugzilla.py
```diff
@@ -0,0 +1,677 @@
...
+        """If there's a java stack trace, use that instead
+
+        Also verify tabs get converted to 4-spaces.
+
+        """
+        processed_crash = {
+            "java_stack_trace": dedent(
+                """\
+                java.lang.NoClassDefFoundError: kotlinx.coroutines.channels.BufferedChannel
+                \tat kotlinx.coroutines.channels.ChannelKt.Channel$default(Channel.kt:44)
```
willkg, line 566: The `\t` inserts a tab. The `java_stack_trace` values have tabs at the beginning of stack frame lines. Fun times.

webapp/crashstats/libbugzilla.py
relud: is the change from `frame.get("line")` to `frame.get("line") is not None` because line can be 0?

webapp/crashstats/libbugzilla.py
```diff
@@ -80,40 +106,55 @@ def crash_report_to_description(crash_report_url, processed_crash):
...
         lines.append(f"Reason: ```{processed_crash['reason']}```")
 
-    # If there's a java_stack_trace, add that; this is a big string blob, so we truncate
-    # it at 5,000 characters but don't otherwise format it
-    if processed_crash.get("java_stack_trace"):
+    frames = None
+    if "json_dump" in processed_crash:
+        # Generate frames from the stackwalker output from parsing a minidump
+        threads = processed_crash.get("json_dump", {}).get("threads")
+        if threads:
```
relud, line 113: this seems like exactly what `:=` is for:
```suggestion
        if threads := processed_crash.get("json_dump", {}).get("threads"):
```

webapp/crashstats/libbugzilla.py
```diff
@@ -80,40 +106,55 @@ def crash_report_to_description(crash_report_url, processed_crash):
...
+                lines.append("")
+                lines.append("No crashing thread identified; using thread 0.")
+
+            thread_index = processed_crash.get("crashing_thread") or 0
+            frames = minidump_thread_to_frames(threads[thread_index])
+
+    if not frames and "java_exception" in processed_crash:
+        stacktraces = (
+            processed_crash["java_exception"].get("exception", {}).get("values", [])
+        )
```
relud, line 124: also a good place for `:=`
```suggestion
    if not frames and java_exception := processed_crash.get("java_exception"):
        stacktraces = java_exception.get("exception", {}).get("values", [])
```

webapp/crashstats/libbugzilla.py
```diff
@@ -80,40 +106,55 @@ def crash_report_to_description(crash_report_url, processed_crash):
...
+        if stacktraces:
+            frames = java_exception_to_frames(
+                stacktraces[0]["stacktrace"].get("frames", [])
```
relud, line 127: seems weird for `stacktrace` to be the only key that can't be missing:
```suggestion
                stacktraces[0].get("stacktrace", {}).get("frames", [])
```

relud (commented):
stopping review because i just noticed @biancadanforth is doing it

webapp/crashstats/tests/test_libbugzilla.py
biancadanforth: I recall you and Sven talking about this a while back in another context, but another option that's more generalizable is to use static typing (I'm coming from TypeScript which is what FxA uses).

I was looking for the TypeScript analog, and Python does have some libraries and packages where you can define types, declare variables as those types and set up static type checking.

I know that'd be a lot more work to set up, but a test scenario like this seems like a good, low-risk way of introducing something like that to the code base.

webapp/crashstats/crashstats/tests/test_jinja_helpers.py
biancadanforth: Much cleaner

biancadanforth (approved):
LGTM. Very thorough test cases FWICT. Had a couple of non-blocking questions and suggestions in line.

I have to step out of the office briefly, but one thing I'd like to know how to do generally is how to get an example crash report from prod into local to see this in action. I'm sure it's probably covered somewhere, and I'll poke around.

webapp/crashstats/libbugzilla.py
willkg: Yes!

webapp/crashstats/libbugzilla.py
willkg: `java_stack_trace` is a big string blob that we can't do much with in Socorro. `java_exception` is a structure that's JSON-encoded in the crash annotations. Because it's structured, we don't have to deal with parsing it.

This is the tracker bug covering phasing out `JavaStackTrace` and its usage: https://bugzilla.mozilla.org/show_bug.cgi?id=1851903

This is the one for Fenix (I think): https://bugzilla.mozilla.org/show_bug.cgi?id=1792902

There's no decision brief. We've had `JavaStackTrace` for a long time and it has a lot of issues and prevents us from improving Fenix support in Socorro. Roger and I spec'd out `JavaException` which is based on the Sentry error event exception structure. He implemented that a while ago (I forget when it was) and the plan was to phase out `JavaStackTrace` in favor of `JavaException`.

webapp/crashstats/libbugzilla.py
willkg: I almost never use `:=` for whatever reason, but yes--this would work here. I can change it.

willkg: In other parts of the codebase, we use `glom` to deeply extract values from the structures, but I didn't want to use that here and introduce another dependency for this module. But ... it sure would read better if I rolled my own `glom` for this module. I'll do that.

webapp/crashstats/libbugzilla.py
willkg: This is a mistake. Oops!

webapp/crashstats/libbugzilla.py
```diff
@@ -14,21 +14,58 @@
...
+
+    :returns: value
+
+    :raises KeyError: if no default and the path part is a key in a map
+    :raises IndexError: if no default and the path part is an index in a list/tuple
+
+    """
+    node = structure
+    for part in path.split("."):
+        if part.startswith("["):
```
willkg, line 37: Whoops--this doesn't work like glom. I need to fix it.

```
app@socorro:/app$ python
Python 3.11.8 (main, Feb 13 2024, 10:25:57) [GCC 10.2.1 20210110] on linux
>>> from glom import glom
>>> struct = {"key": {"sub_key": [{"sub_sub_key": "val1"}]}}
>>> glom(struct, "key.sub_key.[0].sub_sub_key")
  File "/usr/local/lib/python3.11/site-packages/glom/core.py", line 2296, in glom
    raise err
glom.core.PathAccessError: error raised while processing, details below.
 Target-spec trace (most recent last):
 - Target: {'key': {'sub_key': [{'sub_sub_key': 'val1'}]}}
 - Spec: 'key.sub_key.[0].sub_sub_key'
glom.core.PathAccessError: could not access '[0]', part 2 of Path('key', 'sub_key', '[0]', 'sub_sub_key'), got error: ValueError("invalid literal for int() with base 10: '[0]'")
>>> glom(struct, "key.sub_key.0.sub_sub_key")
'val1'
>>> 
```

webapp/crashstats/tests/test_libbugzilla.py
biancadanforth: Ah okay. Yeah what I was thinking would replace however it's being represented now. It would mean we wouldn't have to write/maintain any custom validation code for example for this or any other schema.

Relatedly, I recall [using validation libraries](https://github.com/mozilla/fxa/blob/3354f25b44b06326bfa368950d8909fdfd4262d4/packages/fxa-auth-server/lib/routes/validators.js#L431) like [`joi`](https://joi.dev/api/?v=17.12.3) for complex/composable schemas in Node in FxA. Not sure if there are similar packages available in Python. Sounds like [`schematic`](https://schematics.readthedocs.io/en/latest/) [might be similar/well-liked](https://www.reddit.com/r/Python/comments/4xv0dk/python_validation_framework/).

Anyway, I'm not proposing any dramatic changes here; I'm just thinking out loud about different ways of accomplishing this kind of task.

willkg: When I switched the processor to a schema-based system, I went with JSON schema defined schemas because that's what Data Org had standardized on and crash ingestion was part of Data Org and there was an impetus to unify how we were doing things with how they did things. After settling on a JSON schema-based system, I hunted for libraries and didn't find any Python libraries that did what I needed.

This is partially figured out in this bug:

https://bugzilla.mozilla.org/show_bug.cgi?id=1626698

After building out the schemas, I built some reducer and validation code that allows us to do permission-based reduction and some other things.

https://github.com/mozilla-services/socorro/tree/main/socorro/schemas

https://github.com/mozilla-services/socorro/blob/main/socorro/lib/libsocorrodataschema.py

Some of the processor and large portions of the webapp are built on top of that schema foundation, so switching at this point would be a big project.

I wrote a retrospective here:

https://bluesock.org/~willkg/blog/mozilla/socorro_schema_based_overhaul.html

webapp/crashstats/libbugzilla.py
```diff
@@ -80,7 +116,7 @@ def java_exception_to_frames(stack):
     frames = []
     for i, frame in enumerate(islice(stack, MAX_FRAMES)):
         source = frame.get("filename") or "<nofile>"
-        if frame.get("lineno"):
+        if source and frame.get("lineno") is not None:
```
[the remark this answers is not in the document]
willkg: I missed the `<nofile>` bit. Also, now that I'm looking at this, there are no tests for `java_exception_to_frames`. I'll fix that, too.

webapp/crashstats/libbugzilla.py
```diff
@@ -69,6 +100,35 @@ def frame_generator(thread):
...
+        "source"
+
+    """
+    new_frames = []
+    for i, frame in enumerate(islice(frames, MAX_FRAMES)):
+        source = frame.get("filename") or ""
+        if source and frame.get("lineno") is not None:
+            source = f"{source}:{frame['lineno']}"
+
+        module = truncate(frame.get("module") or "?", 80)
```
willkg, line 118: I looked at some `java_exception` data and some of the modules are hella long, so I added module truncation. In the `java_exception` data I looked at, I didn't find a case where the function signature was too long, but I figured we should keep it just in case.

webapp/crashstats/libbugzilla.py
```diff
@@ -153,13 +151,12 @@ def crash_report_to_description(crash_report_url, processed_crash):
         frames = minidump_thread_to_frames(threads[thread_index])
 
     if not frames:
-        frames = mini_glom(
+        # Generate frames from java_exception structure
+        if frames := mini_glom(
             processed_crash,
```
biancadanforth, line 156: I don't think what you have would cause an error, but wouldn't `if frames :=...` replace the above `if not frames:` statement?

webapp/crashstats/libbugzilla.py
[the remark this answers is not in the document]
biancadanforth: Okay now I am seeing some JavaException frames that previously showed e.g.:

`kotlinx.coroutines.BuildersKt  launch  <nofile>:24`

To now show:
`kotlinx.coroutines.BuildersKt  launch`

This is from the [reference JavaException crash](https://crash-stats.mozilla.org/report/index/125840f3-a9b1-4307-b605-a04da0240327) you linked in the bug.

Is this what we want? I guess a line number with no file is not very useful, but is it useful for people to know that, for some reason, there was no file included for this frame and line number?

biancadanforth (approved):
LGTM. Just asked a couple of clarifying questions on some of the logic and chosen output in certain cases.

webapp/crashstats/libbugzilla.py
willkg: I think we need the `if not frames` thing separate because we don't want to stomp on `frames` with a new value if it already has a value from `minidump_thread_to_frames`.

webapp/crashstats/libbugzilla.py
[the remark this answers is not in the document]
willkg: I intended to match the behavior we use for stackwalker frames. There's no specification for how this should work--it's evolved over time.

Also, this is really low stakes in terms of risk of it not being what people might want. There's no spec plus the code for generating the stackwalker output and the java exception structure change, so we can implement something that seems reasonable now and if there are issues, someone will write up a bug and we can fix it then.

willkg:
Thank you!

webapp/crashstats/libbugzilla.py
biancadanforth: Ah okay yeah I just tried this locally and it doesn't work the way I thought it would, so good to leave it as you had it.

PR #6571, opened 2024-04-03
bug-1886018: Fix flaky tests due to pubsub duplicate delivery

relud:
by comparing delivered crash ids as sets instead of as lists

### DIFF
--- a/socorro/tests/external/pubsub/test_crashqueue.py
+++ b/socorro/tests/external/pubsub/test_crashqueue.py
@@ -5,7 +5,6 @@
 # This is tested using test settings (docker/config/test.env) and Pub/Sub emulator.
 
 import time
-from unittest.mock import ANY
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


socorro/tests/external/pubsub/test_crashqueue.py
```diff
@@ -109,6 +95,6 @@ def test_publish_many(self, pubsub_helper, queue):
         crashqueue.publish(queue, [crash_id_3])
 
         published_crash_ids = pubsub_helper.get_published_crashids(queue)
-        assert sorted(published_crash_ids) == sorted(
-            [crash_id_1, crash_id_2, crash_id_3]
-        )
+        assert set(published_crash_ids) == {
+            crash_id_1, crash_id_2, crash_id_3
+        }
```
willkg, line 100: The original version of this test is curious. I would think we'd want to assert the order of the crash ids published is correct, but I guess this is fine.

willkg (approved):
Looks good. Thank you!

PR #6573, opened 2024-04-04
Clarify how to process and view processed crashes locally

biancadanforth:
This is a doc improvement based on my experience following these instructions to test PR #6566.

### DIFF
--- a/bin/process_crashes.sh
+++ b/bin/process_crashes.sh
@@ -62,5 +62,8 @@ fi
 # Print urls to make it easier to look at them
 for crashid in "$@"
 do
-    echo "Check webapp: http://localhost:8000/report/index/${crashid}"
+    echo "The crash is queued for processing."
+    echo "To process it, start up the processor."
+    echo "To view the processed crash, start up the webapp,"
+    echo "and then check the webapp: http://localhost:8000/report/index/${crashid}"
 done
--- a/bin/process_crashes.sh
+++ b/bin/process_crashes.sh
@@ -62,8 +62,8 @@ fi
 # Print urls to make it easier to look at them
 for crashid in "$@"
 do
-    echo "The crash is queued for processing."
-    echo "To process it, start up the processor."
-    echo "To view the processed crash, start up the webapp,"
-    echo "and then check the webapp: http://localhost:8000/report/index/${crashid}"
+    echo "Check webapp: http://localhost:8000/report/index/${crashid}"
 done
+
+echo "The crash(es) has/have been queued."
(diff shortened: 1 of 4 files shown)


relud (approved):

bin/process_crashes.sh
```diff
@@ -62,5 +62,8 @@ fi
 # Print urls to make it easier to look at them
 for crashid in "$@"
 do
-    echo "Check webapp: http://localhost:8000/report/index/${crashid}"
+    echo "The crash is queued for processing."
+    echo "To process it, start up the processor."
+    echo "To view the processed crash, start up the webapp,"
+    echo "and then check the webapp: http://localhost:8000/report/index/${crashid}"
```
willkg, line 68: This will print four lines for every crash id being processed. If you're processing 20 crash ids, this will print a lot of lines. Maybe move the explanation out of the for loop?

docs/dev.rst
```diff
@@ -720,8 +720,9 @@ You can also use it with ``fetch_crashids``:
 
    app@socorro:/app$ socorro-cmd fetch_crashids --num=1 | bin/process_crashes.sh
 
-After running ``bin/process_crashes.sh``, you will need to run the
-processor which will do the actual processing.
+After running ``bin/process_crashes.sh``, you will need to :ref:`run the processor <run-the-processor-locally>` which will do the actual processing.
+
+If you want to view the processed crash, you will need to :ref:`run the webapp <run-the-webapp-locally>`.
```
willkg, line 725: You need to run both the processor and webapp since the processor uses the `VersionString` API. Instead of this, maybe we should say something like:

> Run the processor and webapp with ``make run`` to process the crash reports.

willkg (changes requested):

willkg:
These are good changes! They make things less implicit. Thank you for doing this!

docs/dev.rst
biancadanforth: Ah okay this is a crash-stats API that the processor uses. Sure I can change it thanks.

willkg (approved):
Thank you for doing this!

biancadanforth:
IIUC the [unrelated test failure](https://app.circleci.com/pipelines/github/mozilla-services/socorro/3071/workflows/ea38fc3c-8dc8-49b4-a683-302b64cb53ac/jobs/9161) is part of bug-1886018. Willkg included a link to the failure in [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1886018#c7).

biancadanforth:
Good point! Thanks.

PR #6575, opened 2024-04-08
bug-1886018: add delay in pubsub tests between publish and pull

relud:
this is intended to fix flaky tests in CI where sometimes expected messages are missing from the pubsub pull response


willkg (approved):
Looks good!

Another way we could have done some of this is similar to how we index crash reports in Elasticsearch helper. We could have `publish` take an additional argument indicating it should sleep.

https://github.com/mozilla-services/socorro/blob/4cc18deb50c193e790467e897464cffeb0224155/socorro/tests/conftest.py#L246-L254

Then the sleeping-after-publish is part of the test harness and not something we have to do in each individual test for most of the tests.

PR #6577, opened 2024-04-09
bug-1886019: generalize s3 references to make room for gcs

relud:
the GCS implementation PR was quite large, this PR is my attempt at breaking some of that work out to be reviewed separately.

<details><summary>test plan</summary>

### run CI tests locally
```bash
make build
make lint
make test
```

### check aws mode works
```bash
echo CLOUD_PROVIDER=AWS > my.env
make runservices
make setup
make run
```
separate shell
```bash
make shell
./bin/socorro_aws_s3.sh ls --recursive telemetry-bucket
./socorro-cmd fetch_crashids --num=3 | tee crashids | ./bin/process_crashes.sh
sleep 5
./bin/socorro_aws_s3.sh ls --recursive dev-bucket
./bin/socorro_aws_s3.sh ls --recursive telemetry-bucket
cat crashids | xargs -i bash -c 'curl -s webapp:8000/api/CrashVerify/?crash_id={}; echo'
```

### expectation:

before sleep: 3 raw crashes and dumps in dev-bucket, nothing in telemetry-bucket
after sleep: 3 processed crashes added to dev-bucket and telemetry-bucket, CrashVerify api returns true for all locations

</details>

### DIFF
--- a/docs/tests/system_checklist.rst
+++ b/docs/tests/system_checklist.rst
@@ -82,7 +82,7 @@ Checklist
 
     * Log into Sentry and check for errors.
 
-    * Submit a crash to the collector. Verify raw crash made it to S3.
+    * Submit a crash to the collector. Verify raw crash made it to cloud storage.
 
 
     Processor
@@ -103,11 +103,11 @@ Checklist
       * stage: check Grafana stage dashboard
       * prod: check Grafana prod dashboard
 
-    Is the processor saving crash data to Elasticsearch? To S3?
+    Is the processor saving crash data to Elasticsearch? To cloud storage?
 
     * Check Grafana "processor.es.save_processed_crash"
 
-    * Check Grafana "processor.s3.save_processed_crash"
+    * Check Grafana "processor.storage.save_processed_crash"
 
     * Check Grafana "processor.telemetry.save_processed_crash"
 
--- a/socorro/external/boto/connection_context.py
+++ b/socorro/external/boto/connection_context.py
@@ -204,7 +204,7 @@ class S3Connection:
                 f"(bucket={bucket!r} key={path}) not found, no value returned"
             ) from exc
 
-    def list_objects_paginator(self, bucket, prefix):
+    def list_objects_paginator(self, bucket, prefix, page_size=None):
         """Returns S3 client paginator of objects with key prefix in bucket
 
         :arg bucket: the name of the bucket
@@ -214,7 +214,11 @@ class S3Connection:
 
         """
         paginator = self.client.get_paginator("list_objects_v2")
-        page_iterator = paginator.paginate(Bucket=bucket, Prefix=prefix)
+        page_iterator = paginator.paginate(
+            Bucket=bucket,
+            Prefix=prefix,
+            PaginationConfig={} if page_size is None else {"PageSize": page_size},
+        )
         return page_iterator
 
     def head_object(self, bucket, key):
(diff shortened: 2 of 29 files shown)


webapp/crashstats/api/views.py
```diff
@@ -476,22 +476,22 @@ def crash_verify(request):
...
-    # Check S3 crash bucket for raw and processed crash data
+    # Check crash bucket for raw and processed crash data
     raw_api = models.RawCrash()
     try:
         raw_api.get(crash_id=crash_id, dont_cache=True, refresh_cache=True)
         has_raw_crash = True
     except CrashIDNotFound:
         has_raw_crash = False
-    data["s3_raw_crash"] = has_raw_crash
+    data["raw_crash"] = has_raw_crash
```
relud, line 486: i don't know how important changing these keys is, if it's important i'm happy to revert this change and worry about the s3 prefix being less than accurate later.

docs/tests/system_checklist.rst
```diff
@@ -103,11 +103,11 @@ Checklist
       * stage: check Grafana stage dashboard
       * prod: check Grafana prod dashboard
 
-    Is the processor saving crash data to Elasticsearch? To S3?
+    Is the processor saving crash data to Elasticsearch? To cloud storage?
 
     * Check Grafana "processor.es.save_processed_crash"
 
-    * Check Grafana "processor.s3.save_processed_crash"
+    * Check Grafana "processor.storage.save_processed_crash"
```
willkg, line 110: We can't change this unless we're planning to conflate s3 and gcs metrics which I don't think we should do.

We could do something like this:

```
* In AWS:

   * Check Grafana "processor.s3.save_processed.crash"

* In GCP:

   * Check Grafana "processor.gcs.save_processed.crash"
```

socorro/external/boto/connection_context.py
```diff
@@ -214,7 +214,11 @@ def list_objects_paginator(self, bucket, prefix):
 
         """
         paginator = self.client.get_paginator("list_objects_v2")
-        page_iterator = paginator.paginate(Bucket=bucket, Prefix=prefix)
+        page_iterator = paginator.paginate(
+            Bucket=bucket,
+            Prefix=prefix,
+            PaginationConfig={} if page_size is None else {"PageSize": page_size},
+        )
```
willkg, line 221: If we pass an empty dict here is that the same as the default `Paginationconfig` for `paginate`?

socorro/external/boto/connection_context.py
```diff
@@ -242,3 +246,29 @@ def head_object(self, bucket, key):
...
+
+        """
+        try:
+            return self.client.delete_object(Bucket=bucket, Key=key)
+        except self.client.exceptions.NoSuchKey as exc:
+            raise KeyNotFound(
+                f"(bucket={bucket!r} key={key}) not found, no value returned"
+            ) from exc
+        except self.client.exceptions.ClientError as exc:
+            if "404" in str(exc):
```
willkg, line 270: Is there a more specific substring we could look for? Maybe "http: 404" or something like that? If the msg ever included a crash id, those could have a 404 in them.

Having said that, this code isn't likely to be used and the S3 stuff is leaving soon anyhow, so we don't _need_ to look into this. But if we knew offhand, it'd be good to improve it.

socorro/external/boto/crashstorage.py
```diff
@@ -100,9 +100,12 @@ def str_to_list(a_string):
     return json.loads(a_string)
 
 
-class BotoS3CrashStorage(CrashStorageBase):
+class BotoS3CrashStorage(SimplifiedCrashDataMixin, CrashStorageBase):
```
willkg, line 103: This seems like a bigger architectural change with crashstorage classes, but there's no explanation of what the big picture is.

What are the main aims for this PR? What're we doing here other than generalizing "s3" to "storage"?

willkg (changes requested):
Seems like there are a lot of changes in this PR beyond just generalizing "s3" to "storage".  I only got about 1/3 of the way through it before I realized I don't know what we're trying to do here. It'd help to have an explanation of what the goals are, what problems are being solved, etc.

relud:
closing in favor of a new plan to break up the gcs changes

PR #6578, opened 2024-04-09
bug-1886018: update dev.rst to mention pubsub

relud:
while working on #6577 i noticed some places where pubsub should be mentioned, and for compatibility with `bin/sqs_cli.py` I added support for reading crash ids from stdin for `socorro-cmd pubsub publish`

### DIFF
--- a/bin/pubsub_cli.py
+++ b/bin/pubsub_cli.py
@@ -10,6 +10,8 @@
 #
 # Usage: ./bin/pubsub_cli.py [SUBCOMMAND]
 
+import sys
+
 import click
 from google.cloud import pubsub_v1
 from google.api_core.exceptions import AlreadyExists, NotFound
@@ -119,6 +121,15 @@ def publish(ctx, project_id, topic_name, crashids):
         batch_settings=pubsub_v1.types.BatchSettings(max_messages=len(crashids))
     )
     topic_path = publisher.topic_path(project_id, topic_name)
+
+    # Pull crash ids from stdin if there are any
+    if not crashids and not sys.stdin.isatty():
+        crashids = list(click.get_text_stream("stdin").readlines())
+
+    if not crashids:
+        raise click.BadParameter(
+            "No crashids provided.", ctx=ctx, param="crashids", param_hint="crashids"
+        )
 
     # publish all crashes before checking futures to allow for batching
     futures = [
--- a/docs/dev.rst
+++ b/docs/dev.rst
@@ -71,7 +71,7 @@ Setup quickstart
 
    That will build the app Docker image required for development.
 
-5. Initialize Postgres, Elasticsearch, S3, and SQS.
+5. Initialize Postgres, Elasticsearch, Pub/Sub, S3, and SQS.
 
    Then you need to set up services. To do that, run:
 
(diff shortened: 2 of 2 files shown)


docs/dev.rst
```diff
@@ -71,7 +71,7 @@ Setup quickstart
 
    That will build the app Docker image required for development.
 
-5. Initialize Postgres, Elasticsearch, S3, and SQS.
+5. Initialize Postgres, Elasticsearch, Pub/Sub, S3, and SQS.
```
willkg, line 74: During the reviews so far, I was figuring that while we're starting a period of transition to GCP things, these services still require AWS things, so we could leave the docs as is for now. As part of cleaning up after the transition, we'd go through and overhaul the docs replacing all the AWS things with GCP things.

But updating the docs as we go is cool, too. :+1:

bin/pubsub_cli.py
```diff
@@ -120,6 +122,15 @@ def publish(ctx, project_id, topic_name, crashids):
     )
     topic_path = publisher.topic_path(project_id, topic_name)
 
+    # Pull crash ids from stdin if there are any
+    if not crashids and not sys.stdin.isatty():
+        crashids = list(click.get_text_stream("stdin").readlines())
```
willkg, line 127: It took me a while to figure out how to do this and get the right behavior such that it doesn't _block_ on stdin when it shouldn't. Then I switched everything to use click which makes it a lot easier.

docs/dev.rst
```diff
@@ -884,9 +886,14 @@ Let's process crashes for Firefox from yesterday. We'd do this:
...
   app@socorro:/app$ bin/socorro_aws_s3.sh sync ./crashdata s3://dev-bucket/
 
-  # Add all the crash ids to the queue
+  # if using CLOUD_PROVIDER=AWS (default)
+  # Add all the crash ids to the sqs queue
   app@socorro:/app$ cat crashids.txt | socorro-cmd sqs publish local-dev-standard
 
+  # or if using CLOUD_PROVIDER=GCP
+  # Add all the crash ids to the pubsub topic
+  app@socorro:/app$ cat crashids.txt | socorro-cmd pubsub publish local-standard-topic
```
willkg, line 895:
```suggestion
  app@socorro:/app$ cat crashids.txt | socorro-cmd pubsub publish test local-standard-topic
```

As a side note, "test" is the project id set in both `docker/config/local_dev.env` and `docker/config/test.env`. Maybe we should change the local dev one to "local"? If we did that, we'd use "local" here, too.

willkg (changes requested):
One minor issue which uncovered another minor issue, but otherwise this is good.

docs/dev.rst
willkg: Typically we try to keep local dev and test environments separate. Would it affect local dev state when we run the tests which include setup/teardown if both have the same project id?

willkg (approved):
Looks good!

docs/dev.rst
relud: > Would it affect local dev state when we run the tests which include setup/teardown if both have the same project id?

no. the only thing we do with pubsub that has a broader impact than specific topics is restarting the container which drops everything from both dev and test

PR #6580, opened 2024-04-12
Reverse preference for importlib_resources

willkg:
Python 3.8 has a `importlib.resources`, but it doesn't include the `files()` function. So for Python 3.8, siggen will require `importlib_resources`.

That's great, except we needed to invert the `importlib_resources` preference so it now prefers the backport library if it's there and uses the built-in library if it's not.

This fixes https://github.com/willkg/socorro-siggen/issues/124 .

### DIFF
--- a/socorro/signature/siglists_utils.py
+++ b/socorro/signature/siglists_utils.py
@@ -5,10 +5,11 @@
 from pathlib import Path
 import re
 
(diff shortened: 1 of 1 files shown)


relud (approved):

willkg:
Thank you!

PR #6589, opened 2024-04-30
bug-1857938: change tag pattern to vYYYY.MM.DD[-N]

willkg:
Note: This also requires a change in Jenkins which I'll do after this lands.


willkg:
Thank you!

PR #6599, opened 2024-05-02
bug-1887629: add CLOUD_PROVIDER to `/__version__` output

willkg:
This adds the `CLOUD_PROVIDER` environment variable value to `/__version__` output so we can easily determine which cloud the service is running in.

To test:

1. run the webapp
2. go to `http://localhost:8000/__version__` and verify there's a "cloud" key with the value of `CLOUD_PROVIDER` in it

Note that this is a dirty way of doing what we want and the Dockerflow "standard" doesn't necessarily allow for other things to be in the `/__version__` output, but we'll plan to clean this up post-GCP-migration so it's temporary.

### DIFF
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -32,7 +32,7 @@ urlwait "${ELASTICSEARCH_URL}"
 urlwait "http://${PUBSUB_EMULATOR_HOST}" 10
 urlwait "${STORAGE_EMULATOR_HOST}/storage/v1/b" 10
 python ./bin/waitfor.py --verbose --codes=200,404 "${SENTRY_DSN}"
-python ./bin/waitfor.py --verbose "${LOCAL_DEV_AWS_ENDPOINT_URL}"
+python ./bin/waitfor.py --timeout 20 --verbose "${LOCAL_DEV_AWS_ENDPOINT_URL}"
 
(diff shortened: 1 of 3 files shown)


webapp/crashstats/settings/base.py
```diff
@@ -65,6 +65,7 @@ def path(*dirs):
     ),
 )
 
+
```
relud, line 68: was this on purpose? if so that's fine, just looked out of place and want to make sure

relud (approved):

webapp/crashstats/settings/base.py
willkg: Whoops! I added something to this file and then changed how I did everything and removed it, but missed an empty line. I'll fix that.

willkg:
That waitfor failed again. I'm going to increase the value in this PR and see if that makes it less annoying.

willkg:
@relud Does that waitfor timeout change look ok to you as a "fix"? Any idea what else might cause it to timeout occasionally when waiting for localstack?

relud (approved):
lgtm

willkg:
I deleted the branch, but forgot I hadn't merged the PR, yet. Yay for git reflog.

willkg:
Thank you!

PR #6608, opened 2024-05-07
bug 1894925: change blocking timeout to 5

willkg:
This changes the blocking timeout to 5s for waitfor--previously it used the socket default timeout of 1s.

The theory is that in CI, it's taking longer than 1s for localstack to respond sometimes which causes CI to fail.


relud (approved):

willkg:
Thank you!

This passed two CI attempts, so I think it probably helps. Maybe.

PR #6609, opened 2024-05-08
bug 1889156: script to load processed crashes into Elasticsearch

biancadanforth:
Because:
* We need to migrate Elasticsearch from AWS to GCP as part of the Socorro GCP migration.
* Due to our version of Elasticsearch being very old, the best (only?) possible migration path is to [load data into the GCP Elasticsearch cluster from the original source](https://www.elastic.co/guide/en/cloud/current/ec-migrating-data.html), which is S3 in our case.

This PR:
* Adds a script that loads processed crashes from either S3 or GCS (depending on `settings.CRASH_SOURCE`) into Elasticsearch by either date (YYYY-MM-DD) or crash ID.

<details>
  <summary>How to test this locally</summary>

I have painstakingly added test cases for all the major scenarios, so running the tests in `socorro/tests/test_load_processed_crashes_into_es.py` (or replicating the setup before running the script with similar args) should be sufficient.

~~The test cases conditionally skip based on `settings.CLOUD_PROVIDER`, so to test the script with GCS as the crash storage source locally, you can edit your `my.env` so that `CLOUD_PROVIDER=GCP` and rebuild the test container (which in my case is the `devcontainer`). Just don't forget to revert that change when you're done.~~ See https://github.com/mozilla-services/socorro/pull/6609#discussion_r1607340132 -- this PR now runs both test cases against both cloud providers thanks to some clever pytest config.

I can add some manual testing instructions if things aren't clear from the test. I might just add those here tomorrow.
</details>

### DIFF
--- a/bin/load_processed_crashes_into_es.py
+++ b/bin/load_processed_crashes_into_es.py
@@ -1,0 +1,203 @@
+#!/usr/bin/env python
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+# Loads a processed crash by either crash ID or date from either
+# GCS or S3 into Elasticsearch, depending on `settings.CRASH_SOURCE`,
+# skipping crashes already in Elasticsearch.
+
+# Uses a variation of `check_crash_ids_for_date`
+# from the `verifyprocessed` command in Crash Stats to get crash IDs from S3/GCS:
+# https://github.com/mozilla-services/socorro/blob/3f39c6aaa7f294884f3261fd268e8084d5eec93a/webapp/crashstats/crashstats/management/commands/verifyprocessed.py#L77-L115
+
+# Usage: ./bin/load_processed_crash_into_es.py [OPTIONS] [CRASH_ID | DATE]
+
+import concurrent.futures
+import datetime
+from functools import partial
+from isodate import parse_date
+from more_itertools import chunked
+
+import click
+
+from socorro import settings
+from socorro.external.es.super_search_fields import FIELDS
+from socorro.external.es.supersearch import SuperSearch
+from socorro.lib.libooid import date_from_ooid
+from socorro.libclass import build_instance_from_settings
+
+NUM_CRASHIDS_TO_FETCH = "all"
+# Number of prefix variations to pass to a check_crashids subprocess
+CHUNK_SIZE = 4
+# Number of seconds until we decide a worker has stalled
+WORKER_TIMEOUT = 15 * 60
+
+
+def is_in_storage(crash_storage, crash_id):
+    """Is the processed crash in storage."""
+    return crash_storage.exists_object(f"v1/processed_crash/{crash_id}")
+
+
+def get_threechars():
+    """Generate all combinations of 3 hex digits."""
+    chars = "0123456789abcdef"
+    for x in chars:
+        for y in chars:
+            for z in chars:
+                yield x + y + z
+
+
+def check_elasticsearch(supersearch, crash_ids):
+    """Checks Elasticsearch and returns list of missing crash ids.
+
+    Crash ids should all be on the same day.
+
+    """
+    crash_ids = [crash_ids] if isinstance(crash_ids, str) else crash_ids
+    crash_date = date_from_ooid(crash_ids[0])
+
+    # The datestamp in the crashid doesn't match the processed date sometimes especially
+    # when the crash came in at the end of the day.
+    start_date = (crash_date - datetime.timedelta(days=5)).strftime("%Y-%m-%d")
+    end_date = (crash_date + datetime.timedelta(days=5)).strftime("%Y-%m-%d")
+
+    params = {
+        "uuid": crash_ids,
+        "date": [">=%s" % start_date, "<=%s" % end_date],
+        "_columns": ["uuid"],
+        "_facets": [],
+        "_facets_size": 0,
+        "_fields": FIELDS,
+    }
+    search_results = supersearch.get(**params)
+
+    crash_ids_in_es = [hit["uuid"] for hit in search_results["hits"]]
+    return set(crash_ids) - set(crash_ids_in_es)
+
+
+def check_crashids_for_date(firstchars_chunk, date):
+    """Check crash ids for a given subset of all crash id prefixes and date"""
+    crash_source = build_instance_from_settings(settings.CRASH_SOURCE)
+    crash_dest = build_instance_from_settings(settings.CRASH_DESTINATIONS["es"])
+
+    supersearch = SuperSearch(crash_dest)
+
+    in_crash_source = []
+    missing_in_es = []
+
+    for firstchars in firstchars_chunk:
+        # Grab all the crash ids at the given date directory
+        page_iterator = crash_source.list_objects_paginator(
+            prefix=f"v1/raw_crash/{date}/{firstchars}",
+        )
+
+        for page in page_iterator:
+            # NOTE(bdanforth): Keys here look like /v1/raw_crash/DATE/CRASHID
+            crash_ids = [item.split("/")[-1] for item in page]
+
+            if not crash_ids:
+                continue
+
+            # Check crashstorage source first
+            for crash_id in crash_ids:
+                if is_in_storage(crash_source, crash_id):
+                    in_crash_source.append(crash_id)
+                else:
+                    click.echo(
+                        f"Could not find processed crash for raw crash {crash_id}."
+                    )
+
+            # Check Elasticsearch in batches
+            for crash_ids_batch in chunked(in_crash_source, 100):
+                missing_in_es_batch = check_elasticsearch(supersearch, crash_ids_batch)
+                missing_in_es.extend(missing_in_es_batch)
+
+    return list(set(missing_in_es))
+
+
+@click.command()
+@click.option(
+    "--date",
+    default=None,
+    type=str,
+    help=("Date to load processed crashes from as YYYY-MM-DD. Defaults to None."),
+)
+@click.option(
+    "--crash-id",
+    default=None,
+    type=str,
+    help="A single crash ID to load into ES from the source. E.g. 64f9a777-771d-4db3-82fa-b9c190240430. Defaults to None.",
+)
+@click.option(
+    "--num-workers",
+    default=4,
+    type=int,
+    help="The number of workers to use to check for crash IDs in the crashstorage source. Defaults to 4.",
+)
+@click.pass_context
+def load_crashes(ctx, date, crash_id, num_workers):
+    """
+    Loads processed crashes into Elasticsearch by crash source (S3 or GCS)
+    and either crash ID or date.
+
+    Must specify either CRASH_ID or DATE.
+
+    """
+    crash_ids = []
+
+    if crash_id:
+        crash_ids.append(crash_id)
+    elif date:
+        check_date = parse_date(date)
+        if not check_date:
+            raise click.ClickException(f"Unrecognized run_time format: {date}")
+
+        check_date_formatted = check_date.strftime("%Y%m%d")
+        click.echo(
+            f"Checking for missing processed crashes for: {check_date_formatted}"
+        )
+
+        check_crashids = partial(check_crashids_for_date, date=check_date_formatted)
+
+        firstchars_chunked = chunked(get_threechars(), CHUNK_SIZE)
+
+        if num_workers == 1:
+            for result in map(check_crashids, firstchars_chunked):
+                crash_ids.extend(result)
+        else:
+            with concurrent.futures.ProcessPoolExecutor(
+                max_workers=num_workers
+            ) as executor:
+                for result in executor.map(
+                    check_crashids, firstchars_chunked, timeout=WORKER_TIMEOUT
+                ):
+                    crash_ids.extend(result)
+    else:
+        raise click.BadParameter(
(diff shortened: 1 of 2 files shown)


socorro/external/es/crashstorage.py
```diff
@@ -650,3 +650,9 @@ def _submit_crash_to_elasticsearch(
                     exc_info=True,
                 )
                 raise
+
+    def document_exists(self, date, crash_id):
+        index = self.get_index_for_date(date)
+        # TODO: Are there different `doc_type`s that we use? Is `_all` fine?
+        # https://elasticsearch-py.readthedocs.io/en/1.9.0/api.html#elasticsearch.Elasticsearch.exists
+        return self.client.connection().exists(index=index, doc_type="_all", id=crash_id)
```
willkg, line 658: This kind of breaks the API structure we've got here where crashstorage should use the connection context to do the things in Elasticsearch.

Instead of this, we can do what we do in the VerifyCrash API and create an `ESCrashStorage`, feed that into a `SuperSearch` and then use existing things to figure out if the item is in ES or not. We can do that entirely in the `load_processed_crash_into_es.py` script such that when we no longer need that script, we don't have anything additional lying around.

```python
import sys

from socorro import settings
from socorro.external.es.supersearch import SuperSearch
from socorro.external.es.super_search_fields import FIELDS
from socorro.libclass import build_instance_from_settings

crash_id = sys.argv[1]

crashstorage = build_instance_from_settings(settings.ES_STORAGE)
supersearch = SuperSearch(crashstorage)

params = {
    "_columns": ["uuid"],
    "_results_number": 1,
    "uuid": crash_id,
    "dont_cache": True,
    "refresh_cache": True,
    "_fields": FIELDS,
}
results = supersearch.get(**params)
exists = results["total"] == 1 and results["hits"][0]["uuid"] == crash_id

print(f"exists?: {exists}")
```

biancadanforth: Thanks for the suggestion! I had some trouble importing `SuperSearch` when I tried that last week, but I couldn't find anything in my notes about it, so I'll give it another shot.

socorro/tests/test_load_processed_crashes_into_es.py
```diff
@@ -0,0 +1,132 @@
...
+    runner = CliRunner()
+    result = runner.invoke(load_crashes, ["--help"])
+    assert result.exit_code == 0
+
+
+@pytest.mark.skipif(
+    settings.CLOUD_PROVIDER != "AWS",
+    reason="Skipping test because CLOUD_PROVIDER is not set to 'AWS'",
+)
+def test_it_loads_processed_crashes_by_date_s3(s3_helper, es_helper):
```
biancadanforth, line 69: This test and the corresponding GCS one testing the date argument are quite slow. Since this is testing a one-off GCP migration script, we should consider some options for speeding things up.

One option is to skip the date tests in CI.

Another option is to add test-only config to the `load_crashes` command to use only the date in the raw_crash prefix when searching for raw crashes by date, since the majority of the slowness comes from including all possible combinations of the first three characters of the crash ID in the prefix.

Perhaps there are other options (other than do nothing!)?

bin/load_processed_crashes_into_es.py
```diff
@@ -0,0 +1,203 @@
...
+            crash_dest.save_processed_crash(None, processed_crash)
+
+            click.echo(
+                f"Crash with ID {crash_id!r} loaded from {type(crash_source).__name__!r}."
+            )
+        except Exception as exc:
+            click.echo(
+                f"Unable to load crash with ID {crash_id!r} from {type(crash_source).__name__!r}; error: {exc}."
+            )
+            continue
```
biancadanforth, line 199: I couldn't think of any special types of exceptions that we'd want to handle differently -- are there any?

Also, please let me know if we'd like to handle exceptions in the general case differently -- right now we just log to stdout that we couldn't load it and move on.

bin/load_processed_crashes_into_es.py
```diff
@@ -0,0 +1,203 @@
...
+                    click.echo(
+                        f"Could not find processed crash for raw crash {crash_id}."
+                    )
+
+            # Check Elasticsearch in batches
+            for crash_ids_batch in chunked(in_crash_source, 100):
+                missing_in_es_batch = check_elasticsearch(supersearch, crash_ids_batch)
+                missing_in_es.extend(missing_in_es_batch)
+
+    return list(set(missing_in_es))
```
relud, line 117: this chunking occurs inside both the `for firstchars in firstchars_chunk` loop and the `for page in page_iterator`, and given that `in_crash_source` is defined outside of both of those, i believe it also needs to occur outside both those loops.

Additionally, checking elasticsearch should be optional: after discussing with willkg a week ago, it's probably faster and/or more desirable not to check in most cases.

```suggestion
    if only_missing_in_es:
        # Check Elasticsearch in batches
        for crash_ids_batch in chunked(in_crash_source, 100):
            missing_in_es_batch = check_elasticsearch(supersearch, crash_ids_batch)
            missing_in_es.extend(missing_in_es_batch)
        return list(set(missing_in_es))
    else:
        return in_crash_source
```

and then `only_missing_in_es` would need to be added as a boolean click option that gets passed through to this method

socorro/tests/test_load_processed_crashes_into_es.py
relud, line 69: rather than using `s3_helper` or `gcs_helper` and skipping tests based on `settings.CLOUD_PROVIDER`, these tests should use `storage_helper` and no `skipif` condition and no duplicate tests per cloud provider.

Then to run the gcp tests requires these changes:
* https://github.com/mozilla-services/socorro/pull/6617/files#r1602263443
* https://github.com/mozilla-services/socorro/pull/6617/files#r1607103528

this ensures the tests fail CI when incorrectly selected, rather than skipping and silently passing CI, and keeps things in line with the pattern set in https://github.com/mozilla-services/antenna/pull/985/files (because my tests were in fact being incorrectly skipped and silently passing CI :sob:)

bin/load_processed_crashes_into_es.py
relud, line 199: this seems like a good place to apply another `concurrent.futures.ProcessPoolExecutor`

socorro/tests/test_load_processed_crashes_into_es.py
relud: any reason we can't drop the crash-prefix splitting entirely?

relud: or make it an option that gets passed into the script with variable length, and then in tests we would submit an empty prefix to do effectively disable the prefix splitting?

relud: note the gcp tests no longer require additional changes, they were merged as part of #6617

biancadanforth: > any reason we can't drop the crash-prefix splitting entirely?

That's a really good question.

I assume the reason we do this, is because in prod, we expect a "large" volume of crashes on any given day, so I tried to dig up some data:
* Using the `supersearch` CLI, I looked up a few dates, and we get between 80-100k crashes for a 24 hour period.
* All other things being equal, that means there are about 25 processed crashes (i.e. unique crash IDs) for each unique three-character crash ID prefix[1].

But, as you've noted elsewhere, not only do we chunk the search prefix, we also get paginated results back from the crashstorage source AND we split all of these checks across multiple workers (4 by default). So my thinking is that no, we don't need the crash-prefix splitting on the surface of things, and it would be simplest/fastest to remove it entirely from the script.

However, this [`firstchars` prefix splitting](https://github.com/mozilla-services/socorro/blob/00d064a90eaf533fda9a4ea0c3b1dddf10781c78/webapp/crashstats/crashstats/management/commands/verifyprocessed.py#L85) is based off of the `check_crashids_for_date` function that's part of an existing `verifyprocessed` Django command which verifies that a raw crash has been processed and saved to all three crashstorage locations (S3, S3 telemetry and ES in this case). @willkg's: To your recollection, why was this done, and what do you think is best here?

[1] Since the first 3 chars can be any of 16 hexadecimal characters, that's 16^3 or 4096 possible combinations. If crash IDs are randomly assigned, and we assume a uniform distribution, that gives 100k/4096 or 25 unique crash IDs per combination.

bin/load_processed_crashes_into_es.py
biancadanforth: > this chunking occurs inside both the for firstchars in firstchars_chunk loop and the for page in page_iterator, and given that in_crash_source is defined outside of both of those, i believe it also needs to occur outside both those loops.

Ah I see what you're saying -- the `for` loop to check Elasticsearch is doubly nested under `for firstchars in firstchars_chunk`, but actually, that block is to build up my list of `in_crash_source`, and the Elasticsearch check should be done outside of that block once I have the completed list, `in_crash_source`. Right now I'm performing a bunch of checks multiple times for each prefix and each page in each prefix result. I'll fix that.

FWIW, I copied this code from the `verifyprocessed.py` command I [mentioned earlier](https://github.com/mozilla-services/socorro/pull/6609#discussion_r1608861136), though I didn't catch that it [does this](https://github.com/mozilla-services/socorro/blob/00d064a90eaf533fda9a4ea0c3b1dddf10781c78/webapp/crashstats/crashstats/management/commands/verifyprocessed.py#L100-L103). I'll change it there too.

> Additionally, checking elasticsearch should be optional: after discussing with willkg a week ago, it's probably faster and/or more desirable not to check in most cases.

Can you elaborate on why this is likely faster? More desirable? I can add that flag to the CLI as an option.

relud: in the specific case of using this during gcp migration, it's faster and more desirable because we know there is nothing in elasticsearch yet, and that also applies more generally in cases where we know most or all of the crashes being loaded are not in es yet.

i also got the impression from @willkg that checking for crash ids in elasticsearch is very slow, potentially so slow that it's faster to just re-upload a crash than to check if it's present, but i'm not clear on that part.

socorro/tests/test_load_processed_crashes_into_es.py
biancadanforth: That is really cool. I don't totally grok yet how willkg did that; I'll have to spend a bit more time with the patch. But it works beautifully.

biancadanforth:
Okay I've updated the PR for all your feedback except for https://github.com/mozilla-services/socorro/pull/6609#discussion_r1607344147 . I'll submit a separate commit for that tomorrow, so it's easier to revert if willkg weighs in and says it's a bad idea.

relud (approved):
:shipit:

bin/load_processed_crashes_into_es.py
```diff
@@ -77,47 +75,58 @@ def check_elasticsearch(supersearch, crash_ids):
...
+
+    get_crashids = partial(
+        get_crashids_in_storage, only_missing_in_es=only_missing_in_es
+    )
+
+    with concurrent.futures.ProcessPoolExecutor(max_workers=num_workers) as executor:
+        for result in executor.map(get_crashids, page_iterator, timeout=WORKER_TIMEOUT):
+            crash_ids.extend(result)
+
+    return crash_ids
```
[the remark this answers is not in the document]
biancadanforth: I plan to open a separate PR to fix the ES check in `verifyprocessed` per https://github.com/mozilla-services/socorro/pull/6609#discussion_r1607325591. If this change looks good to everyone for this script, I will also include it as part of that PR (or as a separate PR if preferrable).

relud (approved):

socorro/tests/test_load_processed_crashes_into_es.py
willkg: This is how we split the work across multiple workers in `verifyprocessed`. The combinations are added to a queue and the workers pull a combination from the queue, get a listing, then run verification for that crash id.

https://github.com/mozilla-services/socorro/blob/00d064a90eaf533fda9a4ea0c3b1dddf10781c78/webapp/crashstats/crashstats/management/commands/verifyprocessed.py#L132-L150

Whether we want to do that same thing in this script depends on how it's structured in regards to breaking up the work across multiple workers.

willkg: The AWS/GCP test scaffolding was all @relud . It's nice. I'm not 100% sure I understand pytest markers, but it works very nicely.

bin/load_processed_crashes_into_es.py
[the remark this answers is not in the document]
willkg: The way we test verifyprocessed is by mocking the firstchars generator so it only goes through a couple of variations:

https://github.com/mozilla-services/socorro/blob/00d064a90eaf533fda9a4ea0c3b1dddf10781c78/webapp/crashstats/crashstats/tests/test_verifyprocessed.py#L17-L23

I'm not sure I understand what we need to fix about verifyprocessed. Can we push that off to a new PR?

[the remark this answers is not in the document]
biancadanforth: Thanks! Yes actually I was just going to open this separate PR, but I realized it's sufficiently different from what I'm doing here (though it was originally the basis of this logic), that it doesn't actually suffer from the same issue. In short, I don't plan to open a separate PR for that after all.

socorro/tests/test_load_processed_crashes_into_es.py
biancadanforth: Okay thanks. I'm going to defer making any changes to `verifyprocessed` until we try this script out for the migration. I've gone in a different direction in terms of what work is being done in parallel, but I think either way works, and which is "better" depends on what our crashstorage data looks like in prod. This might be in the realm of premature optimization, so I'll table the discussion for now.

PR #6610, opened 2024-05-08
bug-1895527: remove addresses from signatures

willkg:
This a couple of fundamental changes to signature generation:

1. stop including module offsets in frame signatures

   * `libxul.so@0x8ecff` becomes `libxul.so`
   * `libgtk-3.so.0@0x3ad8c6` becomes `libgtk-3.so.0`
   * `(unloaded SnagAA64.dll@0x2e92)` becomes `(unloaded SnagAA64.dll)`

2. collapse repeated `XUL`, `.dll`, `.so`, `<name omitted>`, and `<unknown in >` frames

This has a couple of consequences for signatures where we're missing module symbols:

1. signatures based on modules with missing symbols will fall into the same bucket now so we don't have to have bugs with twenty different variations of `libGLESv2_adreno.so@0xNNNNN` which require ongoing upkeep

2. some kinds of consecutive frames where we don't have a function symbol will now collapse down to a single frame reducing the likelihood the signature is truncated before an interesting bit

Examples of signatures that changed:

```
Crash id: 35af4662-3adc-4591-8322-c3a1e0240507
Original: libxul.so@0x54e5bd2 | libxul.so@0x6c6b461 | libxul.so@0x1e282d4 | libxul.so@0x1e38c88 | libxul.so@0x1c98d7e | libxul.so@0xdb8fff | libxul.so@0xd8ecff | libxul.so@0x9760a8f | libxul.so@0x1c9ee25 | libxul.so@0x9760a8f | libxul.so@0xd8ec2b | libxul.so@0xd...
New:      libxul.so | __clock_gettime_2
Same?:    False

Crash id: 787d2f23-1065-4853-bb2c-8e03c0240507
Original: libxul.so@0x4e5acd8 | libxul.so@0x7f79ba3 | _pt_root
New:      libxul.so | _pt_root
Same?:    False

Crash id: fd518273-515a-4ef5-8423-3f7950240507
Original: libxul.so@0x652b380 | libxul.so@0x652c08c | libxul.so@0x652c5c9 | libxul.so@0x67512fd | libxul.so@0xb65427 | <name omitted> | <name omitted> | <name omitted> | malloc | libxul.so@0x17a2a4a | libxul.so@0x17a2a4a | <name omitted> | <name omitted> | arena...
New:      libxul.so | <name omitted> | malloc | libxul.so | <name omitted> | arena_dalloc | <name omitted> | arena_t::RallocSmallOrLarge | <name omitted> | BaseAllocator::realloc | <name omitted> | BaseAllocator::realloc | <name omitted> | realloc | libxul.so | ...
Same?:    False
Notes:    (1)
          SigTruncate: SigTrunc: signature truncated due to length

Crash id: 4f975e81-68e6-44f3-af65-6fc740240507
Original: libxul.so@0x503dbd4 | libxul.so@0x9466eff | libxul.so@0x50a137a | libxul.so@0x947107f | libxul.so@0x56c8508 | libxul.so@0x505be2f | libxul.so@0x27b1caf | libxul.so@0x950329f | libxul.so@0x95032af | libxul.so@0x950329f | libxul.so@0x50a693e | libxul.so@...
New:      libxul.so | update_get_addr
Same?:    False

Crash id: a4ed6532-136c-44f8-936d-482a70240507
Original: libxul.so@0x6d17d80 | libxul.so@0x8958ce9 | libxul.so@0xa8d1fff | libxul.so@0x5872e33 | RedBlackTree<T>::Remove
New:      libxul.so | RedBlackTree<T>::Remove
Same?:    False

Crash id: c6db49a9-75df-4fc1-8553-2587f0240507
Original: (unloaded SnagAA64.dll@0x2e92) | LdrpCallInitRoutine
New:      (unloaded SnagAA64.dll) | LdrpCallInitRoutine
Same?:    False

Crash id: 117fdf74-8980-4ca9-904e-4ff7b0240507
Original: XUL@0x6b4e60 | XUL@0x6c54bc | XUL@0x49a8550 | XUL@0x497cf74 | XUL@0x4952820 | XUL@0x1246988 | XUL@0x1243eb4 | XUL@0x124ee20 | XUL@0x3e27b0 | XUL@0x3eab1c | XUL@0x3f9324 | XUL@0x6bbbcc | XUL@0x67d5dc | XUL@0x119b1ec | XUL@0x11aee34 | XUL@0x16b537c | XUL...
New:      XUL | content_process_main | IPC_Message_Name=PDocAccessible::Msg_Cache
Same?:    False

Crash id: cd7a425f-38cc-4649-bc8b-a36ef0240507
Original: libcryptoki.4.1.4.dylib@0x14580
New:      libcryptoki.4.1.4.dylib
Same?:    False
```


socorro/signature/rules.py
```diff
@@ -216,19 +216,23 @@ def normalize_frame(
     ):
         """Normalizes a single frame
 
+        Takes a frame and returns the most meaningful single-string representation of
+        the frame. Takes into account the following in this order of precedence
+
         Returns a structured conglomeration of the input parameters to serve as a
         signature. The parameter names of this function reflect the exact names of the
         fields from the JSON MDSW frame output. This allows this function to be invoked
         by passing a frame as ``**a_frame``.
```
willkg, line 225: Oops--I meant to remove this paragraph.

socorro/signature/rules.py
```diff
@@ -216,19 +216,23 @@ def normalize_frame(
...
 
         """
+        # If there's a function symbol, use that--it's the best
```
willkg, line 228: I restructured this method to make it structurally like the precedence rules we want and less all-over-the-place if/then/else-wise which was hard to follow.

socorro/signature/rules.py
```diff
@@ -240,26 +244,23 @@ def normalize_frame(
...
+        if unloaded_modules:
+            unloaded_module = unloaded_modules[0]
+            unloaded_module_module = unloaded_module.get("module")
+            return f"(unloaded {unloaded_module_module})"
 
-            if offset:
-                return f"@{strip_leading_zeros(offset)}"
+        # If there's only an offset, then begrudgingly use that
+        if offset:
+            return f"@{strip_leading_zeros(offset)}"
```
willkg, line 259: We keep the offset here because otherwise there isn't anything in this resulting frame signature. This either doesn't come up much or doesn't come up at all. I wasn't able to find a crash report that hits this scenario. Maybe it still happens but rarely? Maybe I was searching poorly?

I figured we can think about this more another day.

socorro/signature/rules.py
```diff
@@ -366,26 +367,37 @@ def generate(self, source_list, crashed_thread=None, delimiter=" | "):
         # have already been normalized at this point.
         new_signature_list = []
         for a_signature in source_list:
-            # If the signature matches the irrelevant signatures regex, skip to the next frame.
+            # If the frame signature is an empty string, skip it
+            if not a_signature:
+                debug_notes.append(f'empty string; ignoring: "{a_signature}"')
+                continue
```
willkg, line 373: Previously, the frame signature could have at least a module offset, but we nixed that so now frame signatures for frames that are missing all the interesting information end up as `""`. In that case, we don't want it in the signature, so we continue to the next frame.

socorro/signature/rules.py
```diff
@@ -366,26 +367,37 @@ def generate(self, source_list, crashed_thread=None, delimiter=" | "):
...
+            if not a_signature.startswith("(unloaded") and (
+                # A library on Windows or Linux
+                a_signature.lower().endswith((".dll", ".so"))
+                # XUL library on macOS
+                or a_signature == "XUL"
+                # Case where symbols file has "name omitted" as function symbol
+                or a_signature == "<name omitted>"
+                # Case where symbols file only has public function names. bug
+                # #1685178
+                or a_signature.startswith("<unknown in")
```
willkg, line 391: This adds a few more cases to collapse plus it adds some explanation of each case.

socorro/signature/rules.py
```diff
@@ -366,26 +367,37 @@ def generate(self, source_list, crashed_thread=None, delimiter=" | "):
...
             ):
-                a_signature = a_signature.split("@")[0]
-
-                # If this trimmed DLL signature is the same as the previous frame's, skip it.
```
willkg, line 381: Given that we're no longer including the module offset, we don't need to do this anymore.

socorro/signature/siglists/irrelevant_signature_re.txt
```diff
@@ -16,8 +16,8 @@ AnnotateMozCrashReason
 ashmem
 __ascii_stricmp
 __assert
-app_process@0x
-AppleIntelHD3000GraphicsGLDriver@
```
willkg, line 20: We're not including the module offset, so we can drop all variations of `@0x` from the end of these frame signatures.

Note that we keep the two `@0x...` cases at the top of this file because we still have one case where it could be just an offset.

socorro/signature/siglists/irrelevant_signature_re.txt
```diff
@@ -87,18 +87,18 @@ JS::loader::ScriptLoadRequest::GetWorkerLoadContext
...
-libandroid_runtime\.so@0x
-libbinder\.so@0x
+libandroid\.so
+libandroid_runtime\.so
+libbinder\.so
 __libc_fatal
 __libc_message
-libEGL\.so@
-libgui\.so@0x
-libicudata.so\..*@
```
willkg, line 97: We don't need the `\..*` here anymore.

socorro/signature/rules.py
smarnach: Would it make sense to make the irrelevant signatures regex match the empty string instead?

socorro/signature/rules.py
smarnach: Doing all of this with string matching looks kind of brittle to me. It feels like the processing should rather be done on typed and structured data, and only at the end we convert that to a string. (I no a change like this is totally out of scope for this PR, so I'm not asking for any change here.)

socorro/signature/rules.py
```diff
@@ -216,19 +216,18 @@ def normalize_frame(
     ):
         """Normalizes a single frame
 
-        Returns a structured conglomeration of the input parameters to serve as a
-        signature. The parameter names of this function reflect the exact names of the
-        fields from the JSON MDSW frame output. This allows this function to be invoked
-        by passing a frame as ``**a_frame``.
+        Takes a frame and returns the most meaningful single-string representation of
+        the frame. Takes into account the following in this order of precedence
```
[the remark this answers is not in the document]
willkg: No, it isn't--oops! I'll fix that.

socorro/signature/rules.py
willkg: The irrelevant signatures regex is built from the `socorro/signature/siglists/irrelevant_signatures.txt` file. I don't have a good way to encode an empty string in that file.

So in this case, it seemed a lot easier to do this.

socorro/signature/rules.py
willkg: Currently, all things symbols is done in strings. The strings come from multiple places:

1. compiler generated symbols from builds Mozilla does and builds done by Microsoft, Apple, Nvidia, various Linux distributions, etc
2. compiler symbols that dump_syms fixes in its demangling before putting them in symbols files
3. changes that the stackwalker makes to symbols before emitting the stackwalker output when processing a minidump
4. changes we make to normalize symbols in signature generation

It could be easier if the data were more structured, but getting from here to there feels like an insurmountable task across a bunch of projects, organizations, and decades of software development history.

willkg:
The examples in the PR description are generated this way:

```
socorro-cmd fetch_crashids --num=1000 | socorro-cmd signature --different-only
```

The `signature` command can tell you a lot about what happened in signature generation. The `--different-only` flag causes it to only emit output if the new signature it would generate is different than the one for that crash report in prod. It also has a `--verbose` flag which emits all the notes from generating the signature.

willkg:
Thank you!

socorro/signature/rules.py
smarnach: I wasn't really thinking of the symbols themselves, but rather of the frame signatures generated by `normalize_frame()`. Instead of being just strings, they could be `Function(name, line)`, `Module(name)`, `UnloadedModule(name)`, `Offset(offset)` or similar. Then instead of `a_signature.startswith("(unloaded")` we could do something more like `isinstance(a_signature, UnloadedModule)`.

I didn't really put much thought into this comment, and I'm only looking at a tiny piece of this whole system without having much experience with the rest, but sometimes a fresh perspective is useful. Often, it isn't.

PR #6613, opened 2024-05-08
OBS-158: Add github actions CI

relud:
copied from antenna with `Build Docker images` through `Run tests` commands changed to match circleci and `local_image` in the last step set to `local/socorro_app:latest`.

This will push socorro images to GAR for deployment in GCP.


willkg (approved):
I compared it with antenna and everything looked the same barring some differences which should be fine.

PR #6615, opened 2024-05-15
Remove unused files

willkg:
This removes some unused files from webapp/ . These stem from when we were using pep8 linter and when we packaged the webapp up as an RPM.


smarnach (approved):

willkg:
Thank you!

PR #6620, opened 2024-05-16
Fix token admin headers and add createtoken command

willkg:
This fixes the token admin headers like we did with Tecken. Further, it adds a createtoken command like Tecken has allowing us to create tokens from the command line in the local dev environment.

To test:

```
cd webapp/
./manage.py createtoken jimmy@example.com c7d1f8cad79545b6a06dc4122f0eb3cd
```


smarnach (approved):
Looks good to me, and I tested it's working as expected.

willkg:
Thank you!

PR #6623, opened 2024-05-23
bug-1898341, bug-1898345: add "host" tag, fix metrics key prefix

willkg:
This adds a "host" tag to emitted metrics when running in GCP. This is derived from "HOSTNAME" if it exists, otherwise it defaults to `socket.gethostname()` like our other services.

This changes Sentry and logging to use `HOSTNAME` configuration variable rather than `HOST_ID`. This brings us in line with other services as we migrate to GCP.

This also adds `"socorro."` prefix to all emitted keys, but only for the GCP environments so we don't make a major change to socorro running in AWS at this time. This brings keys in line with our other services.

In order to do this, I had to create a singleton `METRICS` in `socorro/libmarkus.py` and then rework everything to use that.

To test:

1. run the tests--they have assertions about keys and tags
2. run socorro, process crashes, make sure metrics are emitted and look correct
   * processor, webapp, and crontabber metrics keys **are NOT** prefixed with `socorro`
   * keys **do NOT** include a `host` tag
3. set `CLOUD_PROVIDER=GCP`, run socorro, process crashes, make sure metrics are emitted and look correct
   * processor, webapp, and crontabber metrics keys **are** prefixed with `socorro`
   * keys **do** include a `host` tag

### DIFF
--- a/docker/config/gunicorn_config.py
+++ b/docker/config/gunicorn_config.py
@@ -12,19 +12,19 @@ from decouple import config as CONFIG
 
 LOGGING_LEVEL = CONFIG("LOGGING_LEVEL", "INFO")
 LOCAL_DEV_ENV = CONFIG("LOCAL_DEV_ENV", False, cast=bool)
-HOST_ID = socket.gethostname()
+HOSTNAME = CONFIG("HOSTNAME", default=socket.gethostname())
 
 
-class AddHostID(logging.Filter):
+class AddHostname(logging.Filter):
     def filter(self, record):
-        record.host_id = HOST_ID
+        record.hostname = HOSTNAME
         return True
 
 
 logconfig_dict = {
     "version": 1,
     "disable_existing_loggers": False,
-    "filters": {"add_hostid": {"()": AddHostID}},
+    "filters": {"add_hostname": {"()": AddHostname}},
     "handlers": {
         "console": {
             "level": LOGGING_LEVEL,
@@ -35,7 +35,7 @@ logconfig_dict = {
             "level": LOGGING_LEVEL,
             "class": "logging.StreamHandler",
             "formatter": "mozlog",
-            "filters": ["add_hostid"],
+            "filters": ["add_hostname"],
         },
     },
     "formatters": {
--- a/socorro/external/boto/crashstorage.py
+++ b/socorro/external/boto/crashstorage.py
@@ -5,8 +5,6 @@
 import json
 import logging
 
-import markus
-
 from socorro.external.crashstorage_base import (
     CrashStorageBase,
     CrashIDNotFound,
@@ -97,8 +95,6 @@ class BotoS3CrashStorage(CrashStorageBase):
         )
         self.bucket = bucket
         self.dump_file_suffix = dump_file_suffix
-
-        self.metrics = markus.get_metrics(metrics_prefix)
 
     @classmethod
     def build_connection(cls, region, access_key, secret_access_key, endpoint_url):
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -295,7 +295,10 @@ class ESCrashStorage(CrashStorageBase):
 
         # Create a MetricsInterface that includes the base prefix plus the prefix passed
         # into __init__
-        self.metrics = markus.get_metrics(build_prefix(METRICS.prefix, metrics_prefix))
+        self.metrics = markus.get_metrics(
+            build_prefix(METRICS.prefix, metrics_prefix),
+            filters=list(METRICS.filters),
+        )
 
         self.index = index
         self.index_regex = index_regex
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -23,7 +23,7 @@ from socorro.external.es.super_search_fields import (
     is_indexable,
     parse_mapping,
 )
-
+from socorro.libmarkus import METRICS, build_prefix
 from socorro.lib.libdatetime import JsonDTEncoder, string_to_datetime, utc_now
 
 
@@ -293,7 +293,9 @@ class ESCrashStorage(CrashStorageBase):
 
         self.client = self.build_client(url=url, timeout=timeout)
 
-        self.metrics = markus.get_metrics(metrics_prefix)
+        # Create a MetricsInterface that includes the base prefix plus the prefix passed
+        # into __init__
+        self.metrics = markus.get_metrics(build_prefix(METRICS.prefix, metrics_prefix))
 
         self.index = index
         self.index_regex = index_regex
--- a/socorro/libmarkus.py
+++ b/socorro/libmarkus.py
@@ -1,0 +1,78 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Holds Markus utility functions and global state."""
+
+import logging
+import os
+
+import markus
+from markus.filters import AddTagFilter
+
+
+_IS_MARKUS_SETUP = False
+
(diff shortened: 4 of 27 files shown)


docker/config/gunicorn_config.py
```diff
@@ -12,19 +12,19 @@
...
-HOST_ID = socket.gethostname()
+HOSTNAME = CONFIG("HOSTNAME", default=socket.gethostname())
 
 
-class AddHostID(logging.Filter):
+class AddHostname(logging.Filter):
     def filter(self, record):
-        record.host_id = HOST_ID
+        record.hostname = HOSTNAME
         return True
```
willkg, line 21: Once we've switched to GCP, we can remove `AddHostname` because we won't need it anymore. I'll write up a follow-up bug for that.

socorro/external/boto/crashstorage.py
```diff
@@ -98,8 +96,6 @@ def __init__(
         self.bucket = bucket
         self.dump_file_suffix = dump_file_suffix
 
-        self.metrics = markus.get_metrics(metrics_prefix)
```
willkg, line 101: This never got used in the crashstorage classes except for `ESCrashStorage`. I removed it from everywhere and for `ESCrashStorage` I did something different.

socorro/external/es/crashstorage.py
```diff
@@ -293,7 +293,9 @@ def __init__(
 
         self.client = self.build_client(url=url, timeout=timeout)
 
-        self.metrics = markus.get_metrics(metrics_prefix)
+        # Create a MetricsInterface that includes the base prefix plus the prefix passed
+        # into __init__
+        self.metrics = markus.get_metrics(build_prefix(METRICS.prefix, metrics_prefix))
```
willkg, line 298: The `METRICS.prefix` will be either `""` or `"socorro"` depending on the cloud environment. `build_prefix` will append the crash storage metrics prefix (i.e. `"processor.es"` here) to that and that results in the prefix it uses.

willkg: I wrote up a bug in Markus to make this a little easier. It'd be nice to hone a `MetricsInterface` iteratively. Then we wouldn't need this wonky looking thing.

https://github.com/willkg/markus/issues/142

socorro/libmarkus.py
```diff
@@ -0,0 +1,78 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+"""Holds Markus utility functions and global state."""
```
willkg, line 5: This is effectively a copy from Tecken with some minor adjustments.

socorro/processor/pipeline.py
```diff
@@ -33,14 +33,14 @@ def add_notes(self, notes):
...
+    def __init__(self, rulesets, hostname=None):
         """
         :arg rulesets: either a dict of name -> list of rules or a Python dotted
             string path to a dict of name -> list of rules
-        :arg host_id: the id of the host this is running on; used for logging
+        :arg hostname: the id of the host this is running on; used for logging
         """
         self.logger = logging.getLogger(__name__ + "." + self.__class__.__name__)
-        self.host_id = host_id or "unknown"
+        self.hostname = hostname or "unknown"
```
willkg, line 43: I should change this to force the caller to set the hostname.

webapp/crashstats/crashstats/apps.py
```diff
@@ -48,7 +45,7 @@
 
 
 def count_sentry_scrub_error(msg):
-    metrics.incr("sentry_scrub_error", 1)
+    METRICS.incr("webapp.sentry_scrub_error", 1)
```
willkg, line 48: This key changes completely. Before we had `webapp.crashstats.apps.sentry_scrub_error` and now we've got `webapp.sentry_scrub_error` which matches the other services.

After this lands, we'll need to update dashboards.

webapp/crashstats/crashstats/tests/test_views.py
```diff
@@ -475,15 +476,30 @@ def test_report_index(self, client, db, storage_helper, user_helper):
...
+                "webapp.view.pageview",
+                tags=[
+                    "ajax:false",
+                    "api:false",
+                    "path:/report/index/_crashid_crash_id_",
+                    "status:200",
+                ],
+            )
+        else:
+            records = metrics_mock.filter_records(
```
willkg, line 490: So the problem here that makes this wonky looking is that a `host` tag gets added and it's hard to know what the value would be. Markus doesn't have a good way of dealing with this. I wrote this up:

https://github.com/willkg/markus/issues/141

I might have been able to implement something here, but Markus tries hard to allow for unstable orders and does this:

https://github.com/willkg/markus/blob/8dcc2fad9178dc03853e666c1a84e90a30411150/src/markus/testing.py#L109-L110

In the meantime, I did this ridiculous thing. We can change it later when Markus grows an `AnyValue` thing.

socorro/mozilla_settings.py
```diff
@@ -73,27 +73,9 @@ def _or_none(val):
...
-                parser=int,
-                doc="statsd port.",
-            ),
-        },
-    },
-]
-if LOCAL_DEV_ENV:
-    MARKUS_BACKENDS.append({"class": "markus.backends.logging.LoggingMetrics"})
+
+STATSD_HOST = (_config("STATSD_HOST", default="localhost", doc="statsd host."),)
```
willkg, line 77: Whoops.

docker/config/gunicorn_config.py
relud: why won't need it anymore?

relud: because logs in gcp are already tagged with pod and container name?

socorro/external/es/crashstorage.py
relud: this causes the host tag to be lost for socorro.processor.es.* metrics in GCP. For example i got these two log lines when testing, the first with host tag working for gcs storage, the second with host tag missing for es:
```
processor-1      | 2024-05-24 15:20:13,012 INFO - processor - markus - Thread-2 - METRICS|2024-05-24 15:20:13|timing|socorro.processor.storage.save_processed_crash|12.225023994687945|#host:160ce9ffa6cc
processor-1      | 2024-05-24 15:20:13,039 INFO - processor - markus - Thread-1 - METRICS|2024-05-24 15:20:13|histogram|socorro.processor.es.crash_document_size|4454|
```

relud (changes requested):

webapp/crashstats/__init__.py
```diff
@@ -1,3 +1,5 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+default_app_config = "crashstats.crashstats.apps.CrashstatsAppConfig"
```
relud, line 5: I don't understand what this is doing or why it's needed for this PR

socorro/external/es/crashstorage.py
relud: note, this only breaks the host tag for metrics emitted in this file, metrics emitted in the base class still get those host tag, for example the first metric here has no host tag, but the second does have it:
```
processor-1      | 2024-05-24 15:20:13,148 INFO - processor - markus - Thread-2 - METRICS|2024-05-24 15:20:13|histogram|socorro.processor.es.index|95.75057029724121|#outcome:successful
processor-1      | 2024-05-24 15:20:13,148 INFO - processor - markus - Thread-2 - METRICS|2024-05-24 15:20:13|timing|socorro.processor.es.save_processed_crash|136.01990399183705|#host:160ce9ffa6cc
```

willkg: Oh, whoops. It's because I'm not bringing over the filters, too. Good catch!

willkg:
Thank you!

After this autodeploys to stage, I'll update the dashboards.

PR #6625, opened 2024-05-28
bug-1898800: fix csrf token storage

willkg:
This changes Crash Stats to store the csrf token in the session rather than in the cookie.

This also adjusts the order of the middleware per Django's documented conventions.

This also sets the `CSRF_COOKIE_NAME` to "crashstatscsrftoken" to make it distinguishable in the local dev environment from other things creating a "csrftoken". It shouldn't get created anymore and now it's a lot easier to verify that.

To test:

1. check out this pr, then switch to the first commit
2. set up your local dev environment
3. run `make run`
4. go to `localhost:8000` in a fresh private window and log in
5. check firefox dev tools Storage tab Cookies

You'll see the csrf cookie from crash stats code (crashstatscsrftoken) and the one from the oidc provider (csrftoken) which we use in the local dev environment.

image

1. do the above steps with the tip of this pr branch -- make sure to use a fresh private window so you're not picking up cookies from the last round

You'll see only the csrf cookie from the oidc provider (csrftoken).

image

### DIFF
--- a/webapp/crashstats/settings/base.py
+++ b/webapp/crashstats/settings/base.py
@@ -126,13 +126,13 @@ MIDDLEWARE = [
     # CORS needs to go before other response-generating middlewares
     "corsheaders.middleware.CorsMiddleware",
     "whitenoise.middleware.WhiteNoiseMiddleware",
-    "django.middleware.common.CommonMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
+    "django.middleware.common.CommonMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "crashstats.tokens.middleware.APIAuthenticationMiddleware",
-    "django.middleware.csrf.CsrfViewMiddleware",
     "mozilla_django_oidc.middleware.SessionRefresh",
     "django.contrib.messages.middleware.MessageMiddleware",
+    "django.middleware.csrf.CsrfViewMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     # If you run crashstats behind a load balancer, your `REMOTE_ADDR` header
     # will be that of the load balancer instead of the actual user. The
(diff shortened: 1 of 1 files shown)


webapp/crashstats/settings/base.py
```diff
@@ -126,13 +126,13 @@ def path(*dirs):
...
     "whitenoise.middleware.WhiteNoiseMiddleware",
-    "django.middleware.common.CommonMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
+    "django.middleware.common.CommonMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "crashstats.tokens.middleware.APIAuthenticationMiddleware",
-    "django.middleware.csrf.CsrfViewMiddleware",
     "mozilla_django_oidc.middleware.SessionRefresh",
     "django.contrib.messages.middleware.MessageMiddleware",
+    "django.middleware.csrf.CsrfViewMiddleware",
```
willkg, line 135: I shifted some of the middlewares around per this Django advice:

https://docs.djangoproject.com/en/4.2/ref/middleware/#middleware-ordering

webapp/crashstats/settings/base.py
```diff
@@ -490,6 +490,9 @@ def filter(self, record):
     doc="Make this unique, and don't share it with anybody. It cannot be blank.",
 )
 
+CSRF_COOKIE_NAME = "crashstatscsrfcookie"
+CSRF_USE_SESSIONS = True
```
willkg, line 494: The first item helps us validate things. The second item fixes the bug.

relud (approved):

willkg:
Thank you!

webapp/crashstats/settings/base.py
biancadanforth: I know this PR already merged, but IIUC, the only purpose of `CSRF_COOKIE_NAME` is to validate this change as part of the PR review. Why not set it as a temporary env var in one's local config instead of permanently adding a new env var to settings?

willkg: The `CSRF_COOKIE_NAME` is a Django setting used by the build-in CSRF middleware and it defaults to `"csrftoken"`. All I did here was change the value to `"crashstatscsrfcookie"`. This doesn't introduce a new environment variable or otherwise affect things.

We can take the `CSRF_COOKIE_NAME` setting out now that this has landed, been pushed to production, and verified.

PR #6626, opened 2024-05-30
bug-1877291: process crash reports we discover weren't processed

willkg:
verifyprocessed is a Django management command that runs every day, looks at the previous day of raw crashes, checks to see if there's a processed crash and data in Elasticsearch, and adds a MissingProcessedCrash record if there wasn't.

This fixes the command so it also tosses the crash id in the reprocessing queue.

This fix allows Socorro to self-heal. Currently it notes that some crash reports weren't processed, but doesn't do anything to fix that.

In 2024, we've had one crash report that wasn't processed:

https://crash-stats.mozilla.org/siteadmin/crashstats/missingprocessedcrash/

It doesn't happen often except when the collector has problems connecting to the queue. In that case, we get a burst of crash reports that weren't processed and it's time-consuming to fix manually.

### DIFF
--- a/webapp/crashstats/crashstats/management/commands/verifyprocessed.py
+++ b/webapp/crashstats/crashstats/management/commands/verifyprocessed.py
@@ -20,7 +20,7 @@ from django.db.utils import IntegrityError
 from django.utils import timezone
 from django.utils.dateparse import parse_date, parse_datetime
 
-from crashstats.crashstats.models import MissingProcessedCrash
+from crashstats.crashstats.models import MissingProcessedCrash, Reprocessing
 from crashstats.supersearch.models import SuperSearchUnredacted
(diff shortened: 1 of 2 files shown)


webapp/crashstats/crashstats/management/commands/verifyprocessed.py
```diff
@@ -163,6 +164,8 @@ def handle_missing(self, date, missing):
                         pass
                     else:
                         raise
+                reprocessing_api.post(crash_ids=crash_id)
```
willkg, line 167: Given that these were never processed, it's probably the case we should put them in the standard queue. However, it doesn't really matter and we have an API for reprocessing, so I opted to reuse that.

relud (approved):

willkg:
Thank you!

PR #6635, opened 2024-06-04
bug-1900646: fix submitter to only consume standard queue

willkg:
This fixes the submitter to only consume from the standard subscription rather than from standard, priority, and reprocessing.

It does it by stomping on the subscription_name settings for priority and reprocessing in the PubSubCrashQueue class. That means this will never run in AWS, but that's what our plan was anyhow.

### DIFF
--- a/socorro/external/pubsub/crashqueue.py
+++ b/socorro/external/pubsub/crashqueue.py
@@ -128,15 +128,19 @@ class PubSubCrashQueue(CrashQueueBase):
             publisher_options=PublisherOptions(retry=None, timeout=publish_timeout),
         )
 
-        self.standard_topic_path = self.publisher.topic_path(
-            project_id, standard_topic_name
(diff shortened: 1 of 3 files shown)


socorro/external/pubsub/crashqueue.py
```diff
@@ -244,7 +261,11 @@ def publish(self, queue, crash_ids):
                 try:
                     future.result()
                 except Exception:
-                    logger.exception(f"Crashid failed to publish: {batch[i]}")
+                    logger.exception(
```
willkg, line 264: This is an unrelated fix to a bad exception logging line I did earlier. Boo me.

relud (approved):
lgtm, but after reading i realize we could forgo this complexity and instead specify the same subscription for every queue in helm. i'm happy either way.

PR #6636, opened 2024-06-05
bug-1900806: add code_id to missing_symbols field

willkg:
This adds the code_id to the misisng_symbols field. Now the field will have one of two shapes:

* module/version/debugid
* module/version/debugid/codeid

This doesn't affect how we index this field in Elasticsearch. Further, we can reprocess existing crash reports to pick up the codeid information once this is deployed to production if we want to do that.

To test:

1. build socorro local dev environment
2. index some crash reports with missing symbols:
   * 2cca63d3-2d4a-4981-adac-562030240605
   * 6a61e999-b227-4779-8fff-475960240605
   * e27ddaf3-3760-4d8b-9380-a5b570240605
   ```
   ./bin/process_crashes.sh 2cca63d3-2d4a-4981-adac-562030240605 6a61e999-b227-4779-8fff-475960240605 e27ddaf3-3760-4d8b-9380-a5b570240605
   ```
3. use supersearchfacet to get a facet of `missing_symbol` (make sure to include today in the date range)
   ```
   $ supersearchfacet --host=http://localhost:8000/ --_facets=missing_symbols --end-date=2024-06-06
   missing_symbols
    missing_symbols                                                                  | count 
   ----------------------------------------------------------------------------------|-------
    ashshell.dll/24.4.9067.0/279a5e694253439293510083cfe5d1461/6628e8933c2000        | 1     
    ctxwebauthnhook.dll/7.33.3000.10/9009a93593484a878b1bc5bbe06813f31/644fd20a7c000 | 1     
    umppc18511.dll/7.15.18511.0/bc9ce3500d8c4469a2b36acfc3de1b581/663f762514000      | 1     
    total                                                                            | 2     
   ```

### DIFF
--- a/socorro/processor/rules/mozilla.py
+++ b/socorro/processor/rules/mozilla.py
@@ -630,15 +630,20 @@ class TopMostFilesRule(Rule):
 class MissingSymbolsRule(Rule):
     """
     Adds ``missing_symbols`` field where the value is a semi-colon separated set of
-    ``module/version/debugid`` strings for modules where the stackwalker couldn't find a
-    symbols file.
+    module information strings for modules where the stackwalker couldn't find a symbols
+    file. Module information strings have one of two forms depending on whether there's
+    a codeid value in the module data or not:
+
+    * ``module/version/debugid``
+    * ``module/version/debugid/codeid``
+
     """
 
     # Filenames should contain A-Za-z0-9_.- and that's it.
     BAD_FILENAME_CHARACTERS = re.compile(r"[^a-zA-Z0-9_\.-]", re.IGNORECASE)
 
-    # Debug ids are hex strings
-    BAD_DEBUGID_CHARACTERS = re.compile(r"[^a-f0-9]", re.IGNORECASE)
+    # Debug ids and code ids are hex strings
+    BAD_HEXID_CHARACTERS = re.compile(r"[^a-f0-9]", re.IGNORECASE)
 
     NULL_DEBUG_ID = "0" * 33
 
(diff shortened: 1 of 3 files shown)


socorro/processor/rules/mozilla.py
```diff
@@ -650,9 +655,16 @@ def format_module(self, item):
         version = version.replace("/", "\\/")
 
         debugid = item.get("debug_id", self.NULL_DEBUG_ID)
-        debugid = self.BAD_DEBUGID_CHARACTERS.sub("", debugid)
+        debugid = self.BAD_HEXID_CHARACTERS.sub("", debugid)
+
+        codeid = item.get("code_id", None)
+        if codeid:
+            codeid = self.BAD_HEXID_CHARACTERS.sub("", codeid)
```
relud, line 662: kinda feel like we should be redacting whole invalid values, rather than just invalid characters, e.g.:
```suggestion
        if self.BAD_HEXID_CHARACTERS.match(debugid):
            debugid = ""

        codeid = item.get("code_id", None)
        if self.BAD_HEXID_CHARACTERS.match(codeid):
            codeid = ""
```
or at least just for codeid, since it can be omitted:
```suggestion
        debugid = self.BAD_HEXID_CHARACTERS.sub("", debugid)

        codeid = item.get("code_id", None)
        if self.BAD_HEXID_CHARACTERS.match(codeid):
            codeid = ""
```

relud (approved):
r+ even without my comment, because it's consistent with existing behavior as-is.

socorro/processor/rules/mozilla.py
willkg: Crash ingestion is fun because we often get bit flips and bad encoding changes and compiler bugs and all manner of curiosities that result in slightly bad values. For things like this, I hedge towards "having some value that's indicative of what _should_ have been there" is better than "remove the value altogether because it's malformed". It's been helpful to have some signal to debug weird situations.

If it wasn't the wily world of crash ingestion, I'd go with your suggestion.

willkg:
Thank you!

socorro/processor/rules/mozilla.py
willkg: This is definitely a difference between data platform ingestion where the end result is something we need to be clean and proper for data analysis and crash ingestion where the end result is something we need to give us signal we can use for debugging crashes.

PR #6637, opened 2024-06-06
bug-1900995: remove SignatureParentIDNotEqualsChildID rule

willkg:
This removes a signature generation rule that used moz_crash_reason data to stomp on the signature in a certain situation. The assertion that triggered crash reports for this situation is no longer in the code base so this rule no longer has any effect. This removes the rule and the use of moz_crash_reason.

There isn't a good way to test this beyond running the tests. The related bug covers investigation of the situation.


willkg:
Thank you!

PR #6644, opened 2024-06-25
Update sentry-sdk, fillmore, and kent

willkg:
This updates sentry-sdk to 2.5.1, fillmore to 2.0.0, and kent to 2.0.0. This gets us over the sentry-sdk API change.

This also fixes some flakey cache manager tests that sometimes fails and I think it has to do with there not being enough time between creating the files and reading the first file such that the first file doesn't end up with a measurably recent mtime.


relud (approved):

willkg:
Thank you!

PR #6647, opened 2024-06-26
bug-1100790: fix super search platform queries

willkg:
In Crash Stats, if you go to Super Search and then search for `platform="Linux"`, you get back crash reports on the Linux platform. If you search for `platform="Mac OS X"`, you get back crash reports on the Mac OS X platform. However, if you want to search for `platform=["Linux", "Mac OS X"]`, you _only_ get back crash reports for Linux.

This happens for a couple of reasons:

1. the `platform` super search field is an analyzed field despite there being only 5 possible values and being treated as an enum in the search form -- it shouldn't be analyzed, but that's a problem for a different day
2. when you search on a field with a single value, the supersearch code does a "simple_query_search" query, but if you search on a field with 2 or more values, it does a "terms" query -- "simple_query_search" and "terms" queries are _COMPLETELY DIFFERENT_ and "terms" queries will do the wrong thing for analyzed fields and that's why searching for more than one platform with one of them being "Mac OS X" will always fail

Further, it's weird that we're using "simple_query_string" in the first place since that supports a query dsl and all kinds of other stuff.

Fun fact, but the form uses "Windows", but the actual value is "Windows NT". If we were using the actual value, it would also be failing in this way. I bumped into the Windows/Windows NT thing when looking at [bug 1768738](https://bugzilla.mozilla.org/show_bug.cgi?id=1768738).

This has been going on since we built Super Search. This bug is almost 10 years old and really irritating.

Relevant docs:

* https://www.elastic.co/guide/en/elasticsearch/reference/1.4/query-dsl-simple-query-string-query.html
* https://www.elastic.co/guide/en/elasticsearch/reference/1.4/query-dsl-terms-query.html
* https://www.elastic.co/guide/en/elasticsearch/reference/1.4/query-dsl-match-query.html

This PR does a few things:

1. it changes the default query type to a "match" query regardless of whether it's a single value or more than one value
2. it updates the documentation and the query type form field drop down options for "has terms" and "does not have terms" to "matches" and "does not match" because:
   a. that is a lot clearer
   b. it's more correct now that we've changed the code in the way we changed it
4. it fixes a test that was just plain wrong that started failing after I made the code changes

To test this:

1. run the tests
2. set up a local dev environment and process crash reports for Linux, Mac OS X, and Windows NT
3. run the webapp and experiment with different kinds of queries and look for ones where the results are clearly wrong
5. particularly, searching for `platform="Linux"`, `platform="Mac OS X"`, and `platform=["Linux", "Mac OS X"]` should all do the right thing

Here's a list of crash ids I used:

```
# linux
e8b5640c-27a9-4369-81b4-6767c0240625
1f81f372-e460-4b1c-a138-600210240625
42ce5fd9-e630-49b3-b27f-810880240625
c7a504aa-409c-4d09-ad18-a74350240625
c5911799-f37f-4fc0-ad33-d6ee30240625
# mac os x
78f4f538-0763-4c67-9aea-0716b0240625
7aefe685-c192-4d69-b393-d09000240625
82212b3a-b206-488d-8711-7279e0240625
0e8d75ee-8814-4522-af59-2e2570240625
21a8e11d-0033-4814-92b6-087f10240625
# windows nt
afeac5b5-23d2-4a55-ae19-568340240626
30fa927d-17f0-4620-a1a6-b298d0240626
2b57fc66-a481-4744-90dd-405130240626
6df6650e-dc97-4c77-822b-815720240626
8818ad7d-98d2-46df-bb91-0eff40240626
```

Risks of this PR:

The tests are terrible and it's possible that in fundamentally changing the query used for the default operator, we're changing undocumented behavior of Super Search. I spent some quality time today testing Super Search and how queries get built and executed. I believe that any changes here are in good directions, but it's hard to know for sure. If there are problems with these changes, I think we should hone them further in future bugs.

Niceties of this PR:

This forced me to reckon with query building and make it less sus. Improving the search code, docs, and tests is important for the upcoming Elasticsearch migration.

How I'm feeling right now after fixing an almost-10-year-old-bug that has bothered me for a long time:

:100:

### DIFF
--- a/socorro/tests/external/es/test_supersearch.py
+++ b/socorro/tests/external/es/test_supersearch.py
@@ -1594,6 +1594,79 @@ class TestIntegrationSuperSearch:
         for hit in res["hits"]:
             assert "4.0b" in hit["version"]
 
+    def test_get_with_platform(self, es_helper):
+        crashstorage = self.build_crashstorage()
+        api = SuperSearchWithFields(crashstorage=crashstorage)
+        now = utc_now()
+        es_helper.index_crash(
+            processed_crash={
+                "uuid": create_new_ooid(timestamp=now),
+                "signature": "OOM | small | linux",
+                "os_name": "Linux",
+                "product": "WaterWolf",
+                "version": "4.0b2",
+                "date_processed": now,
+            },
+        )
+        es_helper.index_crash(
+            processed_crash={
+                "uuid": create_new_ooid(timestamp=now),
+                "signature": "mozilla::dom::ClientHandle::Control | linux",
+                "os_name": "Linux",
+                "product": "WaterWolf",
+                "version": "4.0b3",
+                "date_processed": now,
+            },
+        )
+        es_helper.index_crash(
+            processed_crash={
+                "uuid": create_new_ooid(timestamp=now),
+                "signature": "js::gc::detail::CellHasStoreBuffer | mac os x",
+                "os_name": "Mac OS X",
+                "product": "WaterWolf",
+                "version": "5.0a1",
+                "date_processed": now,
+            },
+        )
+        es_helper.index_crash(
+            processed_crash={
+                "uuid": create_new_ooid(timestamp=now),
+                "signature": "hmpalert.dll | windows",
+                "os_name": "Windows",
+                "product": "WaterWolf",
+                "version": "5.0a1",
+                "date_processed": now,
+            },
+        )
+        es_helper.refresh()
+
+        # Test querying a single platform
+        resp = api.get(platform=["Linux"])
+        assert resp["total"] == 2
+        signatures = [x["signature"] for x in resp["hits"]]
+        assert list(sorted(signatures)) == [
+            "OOM | small | linux",
+            "mozilla::dom::ClientHandle::Control | linux",
+        ]
+
+        # Test querying a single platform--this one has spaces in the name
+        resp = api.get(platform=["Mac OS X"])
+        assert resp["total"] == 1
+        signatures = [x["signature"] for x in resp["hits"]]
+        assert signatures == [
+            "js::gc::detail::CellHasStoreBuffer | mac os x",
+        ]
+
+        # Test querying multiple platforms--one with spaces in the name
+        resp = api.get(platform=["Linux", "Mac OS X"])
+        assert resp["total"] == 3
+        signatures = [x["signature"] for x in resp["hits"]]
+        assert list(sorted(signatures)) == [
+            "OOM | small | linux",
+            "js::gc::detail::CellHasStoreBuffer | mac os x",
+            "mozilla::dom::ClientHandle::Control | linux",
+        ]
+
     def test_get_against_nonexistent_index(self, es_helper):
         crashstorage = self.build_crashstorage()
         api = SuperSearchWithFields(crashstorage=crashstorage)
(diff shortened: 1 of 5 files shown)


socorro/tests/external/es/test_supersearch.py
```diff
@@ -463,17 +463,11 @@ def test_get_with_combined_operators(self, es_helper):
         now = utc_now()
         crashstorage = self.build_crashstorage()
         api = SuperSearchWithFields(crashstorage=crashstorage)
-        sigs = (
```
willkg, line 466: This test started failing after I fixed the code. I don't like tests that define things far away from where you need to look to understand what the test is doing, so I moved these to be inline even though it becomes repeated strings.

socorro/tests/external/es/test_supersearch.py
```diff
@@ -500,35 +494,46 @@ def test_get_with_combined_operators(self, es_helper):
...
         assert res["total"] == 0
         assert len(res["hits"]) == 0
 
-        res = api.get(signature=["js", "~::"], product=["WaterWolf", "EarthRacoon"])
-        assert res["total"] == 3
-        assert sorted([x["signature"] for x in res["hits"]]) == sorted(
-            [sigs[0], sigs[1], sigs[2]]
-        )
-
-        res = api.get(signature=["js", "~::"], app_notes=["foo bar"])
```
willkg, line 527: This is the case that fails with the new code. What used to happen is it did a "terms" query and only returned the item that matched exactly "foo bar". But that's not what the default operator should be doing--that's what the "is exactly" operator does.

It started failing because the search text gets analyzed per the field `storage_mapping` so this gets tokenized into "foo" and "bar" and those strings appear in several crash report data app_notes values.

socorro/tests/external/es/test_supersearch.py
```diff
@@ -1594,6 +1599,79 @@ def test_get_with_beta_version(self, es_helper):
         for hit in res["hits"]:
             assert "4.0b" in hit["version"]
 
+    def test_get_with_platform(self, es_helper):
```
willkg, line 1602: This test captures the problem and fails without the code changes in this PR.

socorro/tests/external/es/test_supersearch.py
```diff
@@ -500,35 +494,46 @@ def test_get_with_combined_operators(self, es_helper):
...
                 "product": "EarthRacoon",
                 "date_processed": now,
             },
         )
         es_helper.refresh()
 
+        # Search for signatures matching "js" or containing "::"
         res = api.get(signature=["js", "~::"])
         assert res["total"] == 3
-        assert sorted([x["signature"] for x in res["hits"]]), sorted(
```
willkg, line 513: This always passes because it mistakenly uses a `,` instead of a `==` between the two expressions. So the assertion is asserting that `sorted([x["signature"] for x in res["hits"]])` is truthy which it is unless it's an empty list. That's not what we want, though. We want to compare the first expression to the second with a `==`. I fixed that.

relud (approved):

willkg:
On Friday, GitHub Actions weren't working, so I did other things.

Then today, I spent an hour writing a test harness that does this:

1. fetches all the known supersearch fields
2. for each field
   1. does a facet to get the top 3 values
   2. for each value
      1. does a supersearch using the default operator ('has terms" before this PR lands and "matches" after this PR lands) for 3 results and prints out the uuid and value
   3. does a superssearch using the default operator ("has terms" before this PR lands and "matches" after this PR lands) for all three values and prints out the uuid and value

This gives us a baseline of results before and after the change. It's hard coded to look at `date >= 2024-06-24 < 2024-06-25` and at stage.

PR #6648, opened 2024-06-27
Bug 1905094 - Improve signatures for JS context checks.

amccreight:

PR #6652, opened 2024-07-01
Bug 1905714 - Only match JS::Call exactly in signatures.

amccreight:


biancadanforth (approved):
The regexp looks correct.

Verifying this is a little trickier, since it's building off of your [earlier patch](https://github.com/mozilla-services/socorro/pull/6648) late last week, which isn't yet in prod.

I chose one of the crash IDs from the [previous PR](https://github.com/mozilla-services/socorro/pull/6648#pullrequestreview-2149601123), `0ba5c8d8-e0eb-4777-9669-5e94f0240622`, and I looked at its crash signature on `main` relative to its signature on this feature branch (via `socorro-cmd signature 0ba5c8d8-e0eb-4777-9669-5e94f0240622`) -- does this look like what you wanted @amccreight ?

* `main`: `"OOM | unknown | js::AutoEnterOOMUnsafeRegion::crash_impl | JS::CallbackTracer::onEdge | js::GenericTracerImpl<T>::onObjectEdge"`
* This branch: `"OOM | unknown | js::AutoEnterOOMUnsafeRegion::crash_impl | JS::CallbackTracer::onEdge"`

amccreight:
Yes, that's right. Thanks. I didn't mean to change the signature for this crash with my change.

Here's an example of a crash report that should end up with JS::Call in the crash signature with both patches: https://crash-stats.mozilla.org/report/index/54664941-776c-4107-8a03-164a80240617

biancadanforth:
@amccreight , willkg tried to do a prod deploy for Socorro just now (bug-1906174), but the deployment didn't kick off as expected. We just migrated to GCP, so we may still need to iron out some kinks. Hopefully relud (the lead engineer on the GCP migration effort) can take a look when they return next week. That's tracked in obs-264.

biancadanforth:
Update: 2024-07-08 - the deploy issue was fixed, and Socorro was successfully deployed to prod: https://bugzilla.mozilla.org/show_bug.cgi?id=1906174#c3 .

PR #6665, opened 2024-07-11
bug-1907363: Update service-status hosts after gcp migration

relud:
technically we could add the mozgcp.net hosts here, but i figure we're fine without them because this covers all the endpoints we serve.


willkg (approved):
Thank you!

PR #6670, opened 2024-07-17
bug-1908543: remove s3_ prefix from keys in crash verify response

relud:
because it's not in s3 anymore. this changes the api, and i'm not sure how impactful that is.


willkg (approved):
I wrote the `CrashVerify` API for [bug 1425245](https://bugzilla.mozilla.org/show_bug.cgi?id=1425245), but I don't remember using it for much and I don't have `CrashVerify` in any of the scripts I have floating around.

Ergo, I think this is fine to change the way you're changing it. It's also possible we may want to remove it altogether. Though if we did that, we'd have no way of auditing whether a crash report has data in all the right places.

PR #6671, opened 2024-07-17
bug-1908543: fix remove_field.py for gcs

relud:
by modifying it to only use the existing crash storage api

blocks #6669

### DIFF
--- a/bin/remove_field.py
+++ b/bin/remove_field.py
@@ -6,7 +6,7 @@
 
 """
 Given a set of crash report ids via a file and a list of fields to remove, removes the
-fields from the raw crash file in S3 and the document in Elasticsearch.
+fields from the raw crash file and the document in Elasticsearch.
 
 Usage::
 
@@ -22,19 +22,16 @@ Lines prefixed with "#" are ignored.
 
 import concurrent.futures
 from functools import partial
-import io
 import json
 import logging
 import os
 
-from botocore.client import ClientError
 import click
 from elasticsearch.exceptions import ConnectionError
 from elasticsearch_dsl import Search
 from more_itertools import chunked
 
 from socorro import settings
-from socorro.external.boto.crashstorage import build_keys, dict_to_str
 from socorro.libclass import build_instance_from_settings
 from socorro.lib.util import retry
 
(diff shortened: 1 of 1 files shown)


bin/remove_field.py
```diff
@@ -74,51 +71,26 @@ def wait_times_access():
...
-        s3_client.save_file(
-            bucket=bucket,
-            path=path,
-            data=io.BytesIO(dict_to_str(data).encode("utf-8")),
-        )
-        click.echo("# s3: fixed raw crash")
+        # NOTE(relud): this is not guaranteed to save the modified raw crash in the same
+        # location as the original, but it will still be read instead of the original if
+        # that happens. In practice as of 2024-07-17 it always overwrites the original.
+        raw_crashstorage.save_raw_crash(raw_crash=data, dumps=None, crash_id=crashid)
```
relud, line 90: I considered exposing a modified `get_raw_crash_and_path` method, but I think this is sufficient and more maintainable.

willkg (approved):
I threw this script together and I think I used it once. I added it to the repo so we had something to start with in the future if we ever needed to adjust data again, but if we ever have to do something like that, we'll need to update this script first.

For example, we never save raw crash data in Elasticsearch anymore--we only use processed crash data. There will be nothing in the `raw_crash` part of the document (I'm not even sure there is a `raw_crash` part of the document anymore).

Another example is that if we have to remove a raw crash field that gets processed, it'll have vestiges in field-specific places in the processed crash and in Elasticsearch. This doesn't cover that.

Part of me thinks we should remove this and put it in a place where we can get it if we need to write something similar, but it's not something we maintain over time.

In the meantime, I'll approve it because the code changes look fine.

relud:
> Part of me thinks we should remove this and put it in a place where we can get it if we need to write something similar, but it's not something we maintain over time.

I'm a fan of the way the data team handled things like this with a `GRAVEYARD.md` in the git root that lists deleted things that could be worth reviving and a link to the last commit that still had it or the PR that removed them. for example https://github.com/mozilla/telemetry-batch-view/blob/main/GRAVEYARD.md and https://github.com/mozilla/gcp-ingestion/blob/main/GRAVEYARD.md

willkg:
I :heart: that idea! Let's do that.

PR #6672, opened 2024-07-17
 bug-1908543: fix permadelete for gcs

relud:
this adds `delete_crash` methods to the gcs and es crash storage classes, so that backend-specific code lives within the implementation.

blocks #6669

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -10,6 +10,7 @@ import time
 
 import elasticsearch
 from elasticsearch.exceptions import NotFoundError
+from elasticsearch_dsl import Search
 import glom
 import markus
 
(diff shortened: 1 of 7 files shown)


socorro/external/es/crashstorage.py
```diff
@@ -655,3 +656,18 @@ def _submit_crash_to_elasticsearch(
                     exc_info=True,
                 )
                 raise
+
+    def delete_crash(self, crash_id):
```
willkg, line 660: `socorro/ext/crashstorage.py:CrashStorageBase` has this as `remove`. I think I like `delete_crash` better since it lines up with the other methods (`save_raw_crash`, `get_processed_crash`, etc) and is clearer about what it's doing.

Can you rename the `remove` stub in `CrashStorageBase` to `delete_crash`? You'll need to update all the `CrashStorageBase` subclasses that implemented `remove`, too.

```
> rg "def remove"
crashstorage_base.py
198:    def remove(self, crash_id):
260:    def remove(self, crash_id):

fs/crashstorage.py
234:    def remove(self, crash_id):
```

willkg (changes requested):
This looks good with that one issue about `remove` vs. `delete_file`.

I can't remember offhand why I implemented `permadelete_crash_data.py` as effectively a standalone script. I feel like past me would have had a reason. Maybe it was expediency or that I didn't want a `remove` implemented in crash stores in case it got used accidentally. It was probably something like that. We don't need to adhere to that anymore.

willkg (approved):
Looks super!

PR #6674, opened 2024-07-23
bug-1908543: Add GRAVEYARD.md and remove outdated scripts

relud:
as requested in https://github.com/mozilla-services/socorro/pull/6669#discussion_r1686815849 and https://github.com/mozilla-services/socorro/pull/6671#issuecomment-2243388346

This PR creates a GRAVEYARD.md to list deleted things that could be worth reviving and a link to the last commit that still had it or the PR that removed them, following the data team's example https://github.com/mozilla/telemetry-batch-view/blob/main/GRAVEYARD.md


GRAVEYARD.md
```diff
@@ -0,0 +1,20 @@
...
+
+This document records interesting code that we've deleted for the sake of discoverability for the future.
+
+## Various scripts
+
+* [Removal PR](https://github.com/mozilla-services/socorro/pull/6674)
+
+`bin/remove_field.py` which removes fields from raw crashes in crash storage and elasticsearch.
+
+`bin/permadelete_crash_data.py` which deletes crashes from crash storage and elasticsearch.
```
willkg, line 11: I think we want to keep this one because we want to make sure it works over time in case we need it (though that's very infrequent).

Can you un-graveyard this one script, please?

willkg (approved):
This is a great idea. Thank you for doing this!

PR #6677, opened 2024-07-29
bug-1910376: increase stage submitter request timeout to 10 seconds

relud:
because stage submitter threw some errors last night due to the stage collector latency spiking above the current timeout of 3 seconds for a short while.


willkg (approved):
Looks good!

PR #6684, opened 2024-08-06
Enable token expiration date selector

krzepka:
This introduces a new field to the `GenerateTokenForm` with a widget to select the token expiration date. By default this widget is set to the date `TOKENS_DEFAULT_EXPIRATION_DAYS` days from the current date, but users can set it to any date.

### DIFF
--- a/webapp/crashstats/tokens/forms.py
+++ b/webapp/crashstats/tokens/forms.py
@@ -8,10 +8,15 @@ from crashstats.crashstats.forms import BaseModelForm
 from crashstats.tokens import models
 
 
+class DateInput(forms.DateInput):
+    input_type = "date"
+
+
 class GenerateTokenForm(BaseModelForm):
     class Meta:
         model = models.Token
-        fields = ("permissions", "notes")
+        fields = ("permissions", "notes", "expires")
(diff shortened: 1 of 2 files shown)


willkg:
Can you talk more about why you need this change and for what purpose?

krzepka:
I have a socorro instance deployed and we utilize its API in some external applications. It would be quite time-consuming and annoying to regenerate the API token every 3 months for each app that uses it. I could simply change the default expiration days in my settings, but in my opinion it is better to allow the let users to decide for how long to set it.

It's not that I need it in the original repository, but I thought I'll share this code here because you may find it useful.

willkg:
Allowing the user to change the token expiration has security ramifications this doesn't address. Not only should we have a setting for the default expiration, but also a setting for allowing the user to set it at all. Further, this would need tests. This isn't something we need at Mozilla and I don't want to maintain it in perpetuity, so I'm going to close this out.

At Mozilla, for tokens that are used by scripts, we create the token and then request an admin extend the token expiration. That's worked for us for a long time. I don't know if that helps you or not.

Thank you for the pull request! I appreciate your effort.

PR #6686, opened 2024-08-08
bug-1912356: Add elasticsearch 8 instance in dev

relud:
Must leave existing elasticsearch instance intact, and not require any changes to stage/prod configuration


socorro/mozilla_settings.py
```diff
@@ -181,7 +181,11 @@ def _or_none(val):
             default="^socorro[0-9]{6}$",
             doc="Regex for matching Elasticsearch index names.",
         ),
-        "url": _config("ELASTICSEARCH_URL", doc="Elasticsearch url."),
+        "url": _config(
+            "LEGACY_ELASTICSEARCH_URL",
+            alternate_keys=["ELASTICSEARCH_URL"],
```
relud, line 186: if i understand correctly, this will use LEGACY_ELASTICSEARCH_URL if present and fall back to ELASTICSEARCH_URL if not, which should retain backwards compatibility for configuration (for now)

willkg (changes requested):
The tests fail and socorro doesn't run because the web, processor, crontabber, and test services don't require legacy-elasticsearch.

Can you fix that? Then I can look at this again.

docker-compose.yml
```diff
@@ -214,12 +214,16 @@ services:
...
   # https://hub.docker.com/_/elasticsearch/
-  # Note: This image is deprecated, but the new one requires fiddling.
-  elasticsearch:
+  legacy-elasticsearch:
     image: mozilla/socorro_elasticsearch:1.4.5
     ports:
       - "9200:9200"
-      - "9300:9300"
+
+  # https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html
```
biancadanforth, line 222: Eventually `current` won't point to 8.14.3 anymore, right? If we're going to fix the version we're upgrading to, can this link point to that specific version as well?

docker-compose.yml
```diff
@@ -214,12 +214,16 @@ services:
       - "8126:8126"
 
   # https://hub.docker.com/_/elasticsearch/
-  # Note: This image is deprecated, but the new one requires fiddling.
-  elasticsearch:
+  legacy-elasticsearch:
     image: mozilla/socorro_elasticsearch:1.4.5
     ports:
       - "9200:9200"
-      - "9300:9300"
```
biancadanforth, line 222: IIUC this is used primarily for internode communication within an ES cluster. I was reading about some reasons for why one may or may not want to expose it.

The `git blame` [commit](https://github.com/mozilla-services/socorro/commit/fb9fcae74f13c02fb826a36ce9c34df99878dc41) doesn't mention anything specific about it when it was initially added. Any idea why it may have been, and why it's safe to remove now?

biancadanforth (commented):
Super delayed, but I had wanted to peek in on this patch and the other ones for the ES upgrade. I asked a couple questions inline.

docker-compose.yml
relud: i learned this while setting up our elasticsearch cluster in GCP! elasticsearch uses 9300 only for inter-node communication, such as data transfer and master selection (but *not* node discovery), and we only have one node in each dev "cluster", so we're guaranteed not to use it.

biancadanforth: Thanks! From the ES class I'm taking, IIRC 1 node is the default now for the ES cluster when it starts up, so I guess there wouldn't be an explicit config for this for dev, right? How is this config set for stage and prod though, which I imagine have multiple nodes?

docker-compose.yml
```diff
@@ -214,12 +221,20 @@ services:
...
+
+  # https://www.elastic.co/guide/en/elasticsearch/reference/8.14/docker.html
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:8.14.3
+    command:
+    - bin/elasticsearch
+    - -Expack.security.enabled=false
+    - -Ediscovery.type=single-node
```
willkg, line 235: Nit: These need to be indented two more spaces.

```suggestion
      - bin/elasticsearch
      - -Expack.security.enabled=false
      - -Ediscovery.type=single-node
```

willkg: Also, in the current configuration, I think elasticsearch 8 checks the environment to see how much memory it can take and then _takes all the memory_. On my laptop, running the elasticsearch container and anything else causes my laptop to run out of memory and then a bunch of processes get killed (usually Slack and some others).

If I do `docker-compose up legacy-elasticsearch`, it doesn't take up much memory--I don't notice it in the performance monitor. It also doesn't log much. There's nothing about heap sizes in the logs.

If I do `docker-compose up elasticsearch`, it immediately takes 15gb+ of memory and my system gets sad. If I look at the logs, it's _very chatty_. In the extraordinary amount of output, I see this line:

```
socorro-elasticsearch-1         | {"@timestamp":"2024-08-28T15:24:05.701Z", "log.level": "INFO", "message":"JVM arguments [-Des.networkaddress.cache.ttl=60, -Des.networkaddress.cache.negative.ttl=10, -Djava.security.manager=allow, -XX:+AlwaysPreTouch, -Xss1m, -Djava.awt.headless=true, -Dfile.encoding=UTF-8, -Djna.nosys=true, -XX:-OmitStackTraceInFastThrow, -Dio.netty.noUnsafe=true, -Dio.netty.noKeySetOptimization=true, -Dio.netty.recycler.maxCapacityPerThread=0, -Dlog4j.shutdownHookEnabled=false, -Dlog4j2.disable.jmx=true, -Dlog4j2.formatMsgNoLookups=true, -Djava.locale.providers=SPI,COMPAT, --add-opens=java.base/java.io=org.elasticsearch.preallocate, --add-opens=org.apache.lucene.core/org.apache.lucene.store=org.elasticsearch.vec, --enable-native-access=org.elasticsearch.nativeaccess, -Des.cgroups.hierarchy.override=/, -XX:ReplayDataFile=logs/replay_pid%p.log, -Djava.library.path=/usr/share/elasticsearch/lib/platform/linux-x64:/usr/java/packages/lib:/usr/lib64:/lib64:/lib:/usr/lib, -Djna.library.path=/usr/share/elasticsearch/lib/platform/linux-x64:/usr/java/packages/lib:/usr/lib64:/lib64:/lib:/usr/lib, -Des.distribution.type=docker, -XX:+UnlockDiagnosticVMOptions, -XX:G1NumCollectionsKeepPinned=10000000, -XX:+UseG1GC, -Djava.io.tmpdir=/tmp/elasticsearch-6759128300942634220, --add-modules=jdk.incubator.vector, -XX:+HeapDumpOnOutOfMemoryError, -XX:+ExitOnOutOfMemoryError, -XX:HeapDumpPath=data, -XX:ErrorFile=logs/hs_err_pid%p.log, -Xlog:gc*,gc+age=trace,safepoint:file=logs/gc.log:utctime,level,pid,tags:filecount=32,filesize=64m, -Xms15882m, -Xmx15882m, -XX:MaxDirectMemorySize=8329887744, -XX:InitiatingHeapOccupancyPercent=30, -XX:G1ReservePercent=25, --module-path=/usr/share/elasticsearch/lib, --add-modules=jdk.net, --add-modules=ALL-MODULE-PATH, -Djdk.module.main=org.elasticsearch.server]", "ecs.version": "1.2.0","service.name":"ES_ECS","event.dataset":"elasticsearch.server","process.thread.name":"main","log.logger":"org.elasticsearch.node.Node","elasticsearch.node.name":"fac4b151d2b3","elasticsearch.cluster.name":"docker-cluster"}
```

The interesting bit I think is this:

```
-Xms15882m, -Xmx15882m
```

I checked what we were doing when we were trying to migrate to elasticsearch 7:

5a7480001736a2d5c0b484520651e2255cfa1311

We don't set any memory params there. Maybe elasticsearch 7 did different things with heap sizes? Maybe we weren't actually running the container so I never got to find out?

Anyhow, for the local dev environment, I think we need to cap the heap size. If I add these params, then everything works fine for me:

```
-Xms2g
-Xmx2g
```

I picked those values mostly to test an example in the advanced configuration docs. I don't know if those are appropriate values.

https://www.elastic.co/guide/en/elasticsearch/reference/current/advanced-configuration.html#set-jvm-heap-size

Also, we might want to specify some logging related configuration to reduce the amount of chattiness and make it more human readable. It's in a structured format currently.

https://www.elastic.co/guide/en/elasticsearch/reference/current/logging.html

willkg (changes requested):
Can we change the memory settings, please?

docker-compose.yml
relud: we use docker-compose.yml for dev, and we're currently using a helm chart for es 1.4 for stage and prod, which creates an elasticsearch.yml file with configs. we will probably be using terraform to deploy es 8 via elastic cloud in stage and prod.

becaase docker-compose.yml is only used for dev, and dev is configured to use single nodes, we don't need 9300 here.

docker-compose.yml
relud: per https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html
> 5. Start an Elasticsearch container.
> ```
> docker run --name es01 --net elastic -p 9200:9200 -it -m 1GB docker.elastic.co/elasticsearch/elasticsearch:8.15.0
> ```
> > Use the -m flag to set a memory limit for the container. This removes the need to [manually set the JVM size](https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html#docker-set-heap-size).

equivalent option in docker compose is [`mem_limit`](https://docs.docker.com/reference/compose-file/services/#mem_limit), which I have set to `1g` to match the above example

willkg (approved):
I had a suggested change to indent the list in the yaml file another level, but otherwise this looks good.

relud:
> I had a suggested change to indent the list in the yaml file another level, but otherwise this looks good.

oops, got mixed up and thought i'd already merged that, tyty

PR #6688, opened 2024-08-12
bug-1906112: upgrade markus to 5.0.0 and fix metrics-related issues

willkg:
This upgrades markus to 5.0.0. It adds a specification for all metrics emitted. It adds autodocumentation of those metrics to the docs.

While doing this, I found a handful of issues:

1. markus wasn't set up at test session start, so tests may or may not be running with the markus filters set up; now we set up markus at pytest session start
2. `sentry_scrub_error` had a key per service; now we have a single key with the service specified as a tag which makes it easier to have a single dashboard panel covering all sentry scrub errors
3. cronrun emitted separate metrics for success and failure with the timing as a gauge; now it's a timing and one key with the result as a tag

### DIFF
--- a/socorro/processor/cache_manager.py
+++ b/socorro/processor/cache_manager.py
@@ -48,9 +48,7 @@ HEARTBEAT_INTERVAL = 60
 
 
 def count_sentry_scrub_error(msg):
-    # NOTE(willkg): we re-use the processor prefix here and differentiate with the
-    # service tag.
-    METRICS.incr("processor.sentry_scrub_error", value=1, tags=["service:cachemanager"])
+    METRICS.incr("sentry_scrub_error", value=1, tags=["service:cachemanager"])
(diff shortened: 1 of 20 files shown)


relud (approved):

socorro/processor/cache_manager.py
```diff
@@ -50,7 +50,7 @@
 def count_sentry_scrub_error(msg):
     # NOTE(willkg): we re-use the processor prefix here and differentiate with the
     # service tag.
```
relud, line 52: seems like this comment is no longer relevant

willkg: Good catch!

willkg:
I already updated the panel for sentry_scrub_errors. I still need to update the panels for cron job metrics.

Thank you!

PR #6689, opened 2024-08-13
bug-1912979: fix rerererereprocessing problem in local dev env

willkg:
We need to set the ack_deadline_seconds otherwise it defaults to 10 seconds which isn't long enough and causes crash ids to get read again. And again. And again. And again again.

This _only_ affects the local dev environment, so I'm putting it up for review for reals.

### DIFF
--- a/bin/pubsub_cli.py
+++ b/bin/pubsub_cli.py
@@ -79,7 +79,11 @@ def create_subscription(ctx, project_id, topic_name, subscription_name):
     subscriber = pubsub_v1.SubscriberClient()
     subscription_path = subscriber.subscription_path(project_id, subscription_name)
     try:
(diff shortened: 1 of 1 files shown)


bin/pubsub_cli.py
```diff
@@ -79,7 +79,11 @@ def create_subscription(ctx, project_id, topic_name, subscription_name):
     subscriber = pubsub_v1.SubscriberClient()
     subscription_path = subscriber.subscription_path(project_id, subscription_name)
     try:
-        subscriber.create_subscription(name=subscription_path, topic=topic_path)
+        subscriber.create_subscription(
+            name=subscription_path,
+            topic=topic_path,
+            ack_deadline_seconds=600,
```
willkg, line 85: 600 is the number we use in stage and prod.

relud (approved):

PR #6706, opened 2024-09-05
Fix deprecations during docker compose build

relud:
because I noticed some deprecations while working on es8 support

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,5 +1,5 @@
 # This should be kept in sync with the python version used in docker/Dockerfile
-FROM --platform=linux/amd64 mcr.microsoft.com/devcontainers/python:3.11-bullseye@sha256:105bf6a63ab025206f019a371a735fec6553db0be520030c7a2fd0e002947232
+FROM --platform=linux/amd64 mcr.microsoft.com/devcontainers/python:3.11-bullseye@sha256:105bf6a63ab025206f019a371a735fec6553db0be520030c7a2fd0e002947232 AS devcontainer_amd64
 
 ARG userid=10001
 ARG groupid=10001
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,4 @@
 ---
-version: "2"
 services:
   # Socorro app image
   app:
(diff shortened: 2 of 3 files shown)


docker-compose.yml
```diff
@@ -1,5 +1,4 @@
 ---
-version: "2"
```
relud, line 2: fixes:
```
WARN[0000] /Users/relud/socorro/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion
```

.devcontainer/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # This should be kept in sync with the python version used in docker/Dockerfile
-FROM --platform=linux/amd64 mcr.microsoft.com/devcontainers/python:3.11-bullseye@sha256:105bf6a63ab025206f019a371a735fec6553db0be520030c7a2fd0e002947232
+FROM --platform=linux/amd64 mcr.microsoft.com/devcontainers/python:3.11-bullseye@sha256:105bf6a63ab025206f019a371a735fec6553db0be520030c7a2fd0e002947232 AS devcontainer_amd64
```
relud, line 2: fixes:
```
WARN: FromPlatformFlagConstDisallowed: FROM --platform flag should not use constant value "linux/amd64"
```
by using stage names as described in https://docs.docker.com/reference/build-checks/from-platform-flag-const-disallowed/

docker/Dockerfile
```diff
@@ -5,7 +5,7 @@
 # https://hub.docker.com/_/python
 # NOTE(willkg): stick with slim-bullseye which matches how we build the
 # stackwalker
-FROM --platform=linux/amd64 python:3.11.9-slim-bullseye@sha256:8850f5e6e8da9081a6d156252a11161aa22f04d6ed1723c57ca2d5a5d48132bc
+FROM --platform=linux/amd64 python:3.11.9-slim-bullseye@sha256:8850f5e6e8da9081a6d156252a11161aa22f04d6ed1723c57ca2d5a5d48132bc AS app_amd64
```
relud, line 8: fixes:
```
WARN: FromPlatformFlagConstDisallowed: FROM --platform flag should not use constant value "linux/amd64"
```
by using stage names as described in https://docs.docker.com/reference/build-checks/from-platform-flag-const-disallowed/

willkg (approved):
Looks good--thank you for the links!

PR #6707, opened 2024-09-08
bug-1911612: add logger decorator to ESCrashStorage and SuperSearchUnredacted methods

biancadanforth:
This is a WIP to share my progress with others and to provide context for questions I expect I may ask in the coming days as I pursue some additional lines of inquiry.

### Overall approach

To comprehensively identify where and how Elasticsearch is used by the webapp, willkg suggested adding a logging decorator to output the traceback among other important information for key ES interfaces. This could then be used to audit the tests locally, but also to report this information in the webapp container in production on a temporary basis (perhaps a month?), so that any gaps in test coverage can also be identified.

### DIFF
--- a/socorro/external/crashstorage_base.py
+++ b/socorro/external/crashstorage_base.py
@@ -106,7 +106,15 @@ class CrashStorageBase:
         # implementation. This may be fetched by a client of the crashstorge so that it
         # can determine if it can try a failed storage operation again.
         self.exceptions_eligible_for_retry = ()
+
+        # Configure logger
         self.logger = logging.getLogger(__name__ + "." + self.__class__.__name__)
+        handler = logging.StreamHandler()
+        formatter = logging.Formatter(
+            "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+        )
+        handler.setFormatter(formatter)
+        self.logger.addHandler(handler)
 
     def close(self):
         """Close resources used by this crashstorage."""
--- a/socorro/lib/util.py
+++ b/socorro/lib/util.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 from functools import wraps
+import json
 import traceback
 import time
 
@@ -156,9 +157,16 @@ def es_usage_logger(func):
         stack_trace = "".join(traceback.format_stack())
         class_name = self.__class__.__name__
         method_name = func.__name__
-        self.logger.info(
-            f"Bug 1911612,\n{class_name}.{method_name} method called,\nArguments were: {args}, {kwargs},\nTraceback:\n{stack_trace}",
-        )
+
+        log_entry_part_one = {
+            "bug_id": "1911612",
+            "method": f"{class_name}.{method_name}",
+            "arguments": {"args": args, "kwargs": kwargs},
+        }
+
+        log_entry_part_two = f"\ntraceback:\n{stack_trace}"
+
+        self.logger.info(json.dumps(log_entry_part_one, indent=4) + log_entry_part_two)
 
         # Call the original function with its arguments and ensure we
         # return whatever it returns.
--- a/socorro/lib/util.py
+++ b/socorro/lib/util.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 from functools import wraps
+import traceback
 import time
 
 from more_itertools import peekable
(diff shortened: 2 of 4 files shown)


socorro/external/crashstorage_base.py
```diff
@@ -106,7 +106,15 @@ def __init__(self):
...
         self.exceptions_eligible_for_retry = ()
+
+        # Configure logger
         self.logger = logging.getLogger(__name__ + "." + self.__class__.__name__)
+        handler = logging.StreamHandler()
+        formatter = logging.Formatter(
+            "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+        )
+        handler.setFormatter(formatter)
+        self.logger.addHandler(handler)
```
biancadanforth, line 117: This was previously using a logger that inherited from `RootLogger` that had no handlers. I had to add logger config to get it to output at all.

socorro/lib/util.py
```diff
@@ -144,3 +146,30 @@ def _retry_fun(*args, **kwargs):
...
+        class_name = self.__class__.__name__
+        method_name = func.__name__
+
+        log_entry_part_one = {
+            "bug_id": "1911612",
+            "method": f"{class_name}.{method_name}",
+            "arguments": {"args": args, "kwargs": kwargs},
+        }
+
+        log_entry_part_two = f"\ntraceback:\n{stack_trace}"
```
biancadanforth, line 167: I had to break this up into two parts, because `json.dumps` escapes newline characters, which I wanted to preserve in the traceback.

Open to a better way to make these logs more searchable/parseable downstream if others have better ideas!

socorro/lib/util.py
```diff
@@ -144,3 +146,30 @@ def _retry_fun(*args, **kwargs):
...
+
+        self.logger.info(json.dumps(log_entry_part_one, indent=4) + log_entry_part_two)
```
biancadanforth, line 169: This is very verbose right now, thanks largely to the traceback. I'm a little concerned about increases in log volume, even if temporary.

For example, a single call to `SuperSearchUnredacted.get_implementation` had 154 lines where all but 10 were the traceback.

Here's a snippet of the first chunk as a reference:
```
webapp-1         | 2024-09-12 21:16:22,736 INFO - webapp - crashstats.supersearch.models.SuperSearchUnredacted - {
webapp-1         |     "bug_id": "1911612",
webapp-1         |     "method": "SuperSearchUnredacted.get_implementation",
webapp-1         |     "arguments": {
webapp-1         |         "args": [],
webapp-1         |         "kwargs": {}
webapp-1         |     }
webapp-1         | }
webapp-1         | traceback:
webapp-1         |   File "/usr/local/lib/python3.11/threading.py", line 1002, in _bootstrap
webapp-1         |     self._bootstrap_inner()
webapp-1         |   File "/usr/local/lib/python3.11/threading.py", line 1045, in _bootstrap_inner
...
```

I'd like some way to reduce the log volume without sacrificing signal. Sampling is one option, but that may mean missing some little used code paths. A lot of the traceback is framework/middleware code, so another option could be to e.g. remove lines that start with `'File "/usr/local/'` and the next line after and lines that contain the substring "middleware".

Is this additional log volume a concern? Is there a good way to estimate the temporary increase? Where might I find log volume costs/limits in GCP? Maybe it's okay as-is particularly if the increase is timeboxed to say a month.

biancadanforth (commented):
Added some questions inline to get some feedback on what I've got thus far.

socorro/lib/util.py
willkg: I don't think I understand what you're doing here. Are you planning to land these changes and have them run on stage and prod? Or are these changes limited to exploration in your local dev environment?

biancadanforth: I added an "Overall approach" paragraph to the [PR description](https://github.com/mozilla-services/socorro/pull/6707#issue-2512509905) that I think answers your question. Basically, the plan right now is to temporarily deploy the changes to prod to gather some usage data and then revert the change.

willkg: I misunderstood the "This could ..." sentence as being theoretical and not the plan.

I think the underlying assumption for putting this in production is that the way production is used will organically uncover most/all possible uses of Elasticsearch by the code, that we'd be able to extract this usage information from the logs, and that this information will be more comprehensive than what we'd get out of running unit tests and test scenarios. I'm skeptical that this is compellingly true and I'd be hesitant to resort to this. I've made significant changes to the Elasticsearch code over the years by tracing the code in local dev and test environments and not doing something like this in production.

Have you tried capturing HTTP requests to Crash Stats and replaying them in a local dev environment? That continues this initiative, but without having to land and deploy anything.

biancadanforth: Thanks for the suggestion! That sounds like it could be lower risk/faster, though I'm not sure I've done something like that before. Do you have any tooling recommendations for doing this? I'll do a search as well after lunch.

biancadanforth: Oh, I primarily meant tooling for "capturing HTTP requests to Crash Stats and replaying them in a local dev environment", but I'm interested in any tooling recommendations in general for this approach.

biancadanforth:
I typed up a [summary](https://bugzilla.mozilla.org/show_bug.cgi?id=1911612#c1) of the things I've tried and what my new plan is that addresses some of willkg's feedback in this [comment](https://github.com/mozilla-services/socorro/pull/6707#discussion_r1759114219).

I'm going to close this PR, since at most, I will use this decorator only locally going forward.

PR #6727, opened 2024-09-19
Update ElasticSearch

krzepka:
This adapts socorro to the newer ElasticSearch version. I modified it by testing it with ES 8.15. I didn't run it on this fork yet, because it needs dockerfile, requirements and import updates.

I did run this exact code on my another fork - most tests passed and it seemed to work well when I tested it manually.
There are lots of changes which in my opinion look kind of incorrect, such as adding "keyword" everywhere, ["properties"] or accessing the first element of the array. I'm sure there are better ways to code it, as I'm not experienced with Elastic, but I managed to get that working.
In the next days I'll try to modify more code here and try to test it.

I hope it'll help you.

### DIFF
--- a/bin/es_cli.py
+++ b/bin/es_cli.py
@@ -13,8 +13,8 @@ import datetime
 import json
 
 import click
-from elasticsearch_dsl_0_0_11 import Search
-from elasticsearch_1_9_0.client import IndicesClient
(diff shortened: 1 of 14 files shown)


relud:
fixed by #6741

PR #6728, opened 2024-09-20
bug-1918223: fix moz_crash_reason and reason formatting in create bug Description field.

biancadanforth:
Because:
* The `moz_crash_reason` value (and similarly, the `reason` value) was not displaying properly in the Bugzilla create bug page due to some extra whitespace and missing newline characters in the value of the `comment` query parameter passed in the create bug URL.

This commit:
* Trims any whitespace on these values.
* Adds a newline character before and after the triple backticks.

Note: Shortly after bug-1919789 was fixed (which temporarily blocked this bug), I can no longer reproduce the issue in this bug locally or in prod, however, it makes sense to make these changes anyway, so that the formatting will reflect standard Markdown formatting regardless of the current Bugzilla create bug URL implementation.

Here's a screenshot of what the Preview looks like in the Bugzilla create bug page with the crash from the ticket (`2827f0eb-4a80-421e-a287-4c77f0240911`) in local development:

Screenshot 2024-09-20 at 12 01 19 PM

### DIFF
--- a/webapp/crashstats/libbugzilla.py
+++ b/webapp/crashstats/libbugzilla.py
@@ -134,10 +134,12 @@ def crash_report_to_description(crash_report_url, processed_crash):
     ]
     if processed_crash.get("moz_crash_reason"):
         lines.append("")
-        lines.append(f"MOZ_CRASH Reason: ```{processed_crash['moz_crash_reason']}```")
+        lines.append(
+            f"MOZ_CRASH Reason:\n```\n{processed_crash['moz_crash_reason'].strip()}\n```\n"
+        )
     elif processed_crash.get("reason"):
         lines.append("")
-        lines.append(f"Reason: ```{processed_crash['reason']}```")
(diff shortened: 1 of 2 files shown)


relud (approved):

webapp/crashstats/libbugzilla.py
```diff
@@ -134,10 +134,12 @@ def crash_report_to_description(crash_report_url, processed_crash):
     ]
     if processed_crash.get("moz_crash_reason"):
         lines.append("")
-        lines.append(f"MOZ_CRASH Reason: ```{processed_crash['moz_crash_reason']}```")
+        lines.append(
+            f"MOZ_CRASH Reason:\n```\n{processed_crash['moz_crash_reason'].strip()}\n```\n"
```
willkg, line 138: The `lines` array is joined using `\n` as a separator, so I probably would have done this like we do the stack below:

```
lines.append("```")
lines.append(processed_crash("moz_crash_reason").strip())
lines.append("```")
```

PR #6733, opened 2024-09-24
bug-1916848: relocate search_common.py into es and legacy_es modules

relud:
as discussed in zoom.

note `socorro/external/legacy_es/search_common.py` is a copy of `socorro/external/es/search_common.py` with no changes and `socorro/tests/external/legacy_es/test_search_common.py` is a copy of `socorro/tests/external/es/test_search_common.py` with one change `from socorro.external.es` -> `from socorro.external.legacy_es`


willkg (approved):
:100:

PR #6749, opened 2024-10-10
prevent bash from auto-exiting on keypress for `make shell` and `make testshell` commands

biancadanforth:
Because:
* Several developers intermittently had issues where the shells from these commands were exiting.
* These commands originally relied on the default entrypoint of the base Docker image, i.e. `ENTRYPOINT ["/usr/bin/tini", "--", "/app/bin/entrypoint.sh"]` in `docker/Dockerfile`, and we believe Tini is interfering with this shell and causing it to exit prematurely[1].

This commit:
* Passes a different entrypoint into these commands to bypass using Tini in the default entrypoint command.

[1]: As @relud observed, and I've confirmed: `docker compose run --rm -ti --entrypoint '/usr/bin/tini' app bash` does the bad thing and `docker compose run --rm -ti --entrypoint '/bin/bash' app` works

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,7 +104,7 @@ stop: my.env  ## | Stop all service containers.
 
 .PHONY: shell
 shell: my.env .docker-build  ## | Open a shell in the app container.
-	${DC} run --rm app shell
+	${DC} run --rm --entrypoint /bin/bash app
 
 .PHONY: clean
 clean:  ## | Remove all build, test, coverage, and Python artifacts.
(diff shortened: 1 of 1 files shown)


relud (approved):

willkg:
Instead of doing this, we can remove tini altogether and add `init: true` in `docker-compose.yaml` to the relevant services.

https://docs.docker.com/reference/compose-file/services/#init

biancadanforth:
@willkg Since we're not sure that's going to work the same way, and since that's a larger change than what I'm proposing here that involves removing Tini, using a newer Docker Compose feature (`init`), and making changes across repos, I'd recommend that be filed as an OBS ticket for tracking and handled separately. I'm happy to work on it though. Would that be okay?

biancadanforth:
> @willkg Since we're not sure that's going to work the same way, and since that's a larger change than what I'm proposing here that involves removing Tini, using a newer Docker Compose feature (`init`), and making changes across repos, I'd recommend that be filed as an OBS ticket for tracking and handled separately. I'm happy to work on it though. Would that be okay?

Okay I filed https://mozilla-hub.atlassian.net/browse/OBS-361 as a more comprehensive, longer term solution to this that will hopefully fix this issue across the Crash Ingestion repos (or at least the ones that use Tini).

PR #6758, opened 2024-10-14
bug-1924072: Update minidump-stackwalk to 0.22.2

relud:


willkg (approved):
I verified that the stackwalker still works and that `crash_inconsistencies` shows up in stackwalker output for the example crash reports from the other bug.

PR #6759, opened 2024-10-21
bug-1918240: Add crash_inconsistencies field from rust-minidump

relud:
from :btsoi in bug-1918240:
> Rust-minidump has added a feature for detecting inconsistent crashes (see the crash_inconsistencies field in the [JSON Schema](https://github.com/rust-minidump/rust-minidump/blob/main/minidump-processor/json-schema.md)). It will be helpful to show this on crash-stat to indicate that the crash doesn't make sense and is likely caused by bad hardware.
> I think it probably makes sense to treat the field similar to bitflips and have the following support:
> 
>  - make it publicly viewable
>  - add it to the Details tab in the report view
>  - make it searchable in Super Search
>  - make it aggregatable in Super Search and Signature report


socorro/schemas/processed_crash.schema.yaml
```diff
@@ -2457,6 +2464,14 @@ properties:
...
     type: string
     permissions: ["public"]
+  crash_inconsistencies:
+    description: >
+      A list of inconsistencies detected by comparing crash reason/crash
+      address with crashing instruction and memory information.
+    type: ["array", "null"]
+    items:
+      type: string
+    permissions: ["public"]
```
relud, line 2474: I thought this line was supposed to make the field show up in the details tab when I'm not logged in, but in dev it's only showing up when I'm logged in and I'm not sure what to change

relud: ah, didn't specify it enough times

socorro/external/legacy_es/super_search_fields.py
```diff
@@ -1060,6 +1060,9 @@ def apply_schema_properties(fields, schema):
         "query_type": "enum",
         "storage_mapping": {"type": "string"},
     },
+    "crash_inconsistencies": keyword_field(
+        "crash_inconsistencies",
+    ),
```
relud, line 1065: I can't seem to search this field. In dev I process the crashes from the bug, then go to http://localhost:8000/search/?crash_inconsistencies=!__null__ I get no results, but I can see the field is populated via the details pane on individual crashes.

socorro/external/es/super_search_fields.py
```diff
@@ -1053,6 +1053,9 @@ def apply_schema_properties(fields, schema):
         "query_type": "enum",
         "storage_mapping": {"type": "string"},
     },
+    "crash_inconsistencies": keyword_field(
+        "crash_inconsistencies",
```
smarnach, line 1057: Virtually all other invocations of `keyword_field` pass the argument as a keyword parameter (except for one), so I suggest we do the same here for consistency:
```suggestion
        name="crash_inconsistencies",
```

socorro/external/legacy_es/super_search_fields.py
```diff
@@ -1060,6 +1060,9 @@ def apply_schema_properties(fields, schema):
         "query_type": "enum",
         "storage_mapping": {"type": "string"},
     },
+    "crash_inconsistencies": keyword_field(
+        "crash_inconsistencies",
```
smarnach, line 1064:
```suggestion
        name="crash_inconsistencies",
```

smarnach (approved):
This looks good to me, but I'm not really familiar with this codebase. Here are a couple of questions I can't answer:

* Do we really need `CrashInconsistenciesRule`, or is there some other mechanism that can handle trivial cases like this one? Can this possibly be done by `CopyFromRawCrashRule`?
* If we need the rule, what file in the socorro/processor/rules directory should it go in?

I presume you know the answers and did everything correctly, so I'll approve. :)

relud:
>     * Do we really need `CrashInconsistenciesRule`, or is there some other mechanism that can handle trivial cases like this one? Can this possibly be done by `CopyFromRawCrashRule`?

creating a new rule was suggested by @willkg via zoom when I learning how to do this, so I'm going to take that as a new rule is the right way to go here.

>     * If we need the rule, what file in the socorro/processor/rules directory should it go in?

I'm confident this should go in `breakpad.py` because the data source is rust-minidump and all other data generated specfically in that package is handled by rules in that file.

PR #6764, opened 2024-10-23
Rework how supersearch fields permissions are handled

willkg:
In PR #6741, we're moving the code where we convert between the processed crash schema permissions in super search fields to webapp permissions to the general Socorro code. We try really hard to not have webapp things leak into the general Socorro code, so I wanted to experiment with other ways of solving the problem of having two elasticsearch crash storage implementations in respects to super search fields.

This tests that out by adding a `get_supersearch_fields()` function to `webapp/crashstats/supersearch/libsupersearch.py` which imports the specified Elasticsearch crashstorage, gets `SUPERSEARCH_FIELDS` from it, converts the permissions and returns it.

While doing that, I fixed some clarity issues in the code. `convert_permissions` will never have a case where the field has permissions `["public", "protected"]`--that's nonsensical so we shouldn't test that. Also, `convert_permissions` shouldn't stomp on the `permissions_needed` value. Instead, it should put the webapp permissions in a new key and then webapp code should look there.

### DIFF
--- a/webapp/crashstats/supersearch/tests/test_libsupersearch.py
+++ b/webapp/crashstats/supersearch/tests/test_libsupersearch.py
@@ -15,7 +15,7 @@ def test_convert_permissions():
             "permissions_needed": ["public"],
         },
         "version": {
-            "permissions_needed": ["public", "protected"],
+            "permissions_needed": ["protected"],
         },
     }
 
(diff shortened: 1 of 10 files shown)


webapp/crashstats/supersearch/tests/test_libsupersearch.py
```diff
@@ -15,22 +15,25 @@ def test_convert_permissions():
             "permissions_needed": ["public"],
         },
         "version": {
-            "permissions_needed": ["public", "protected"],
```
willkg, line 18: This is nonsensical. We shouldn't test this case.

relud (approved):
I like this better than my solution 💯

PR #6765, opened 2024-10-24
bug-1926948: upgrade dev to postgresql 16.4

relud:
socorro dev is on PostgreSQL 12 right now, which will be EoL on November 14, 2024. for additional details see https://mozilla-hub.atlassian.net/browse/OBS-367


willkg (approved):
:100:

PR #6772, opened 2024-10-29
Dependabot ignore python beyond 3.11

smarnach:
https://mozilla-hub.atlassian.net/browse/OBS-375

### DIFF
--- a/docs/dev.rst
+++ b/docs/dev.rst
@@ -646,6 +646,23 @@ pick up changes:
    $ make devcontainer
 
 
+Upgrading to a new Python version
+=================================
+
+To upgrade Python to a new minor or major version, you need to change the version in
+these files:
+
+* ``.devcontainer/Dockerfile``
+* ``.github/dependabot.yml``
+* ``.readthedocs.yaml``
+* ``docker/Dockerfile``
+* ``docker/Dockerfile.fakesentry``
+* ``docker/set_up_legacy_es_requirements.sh``
+* ``pyproject.toml``
+* ``socorro/tests/processor/test_processor_app.py``
+* ``webapp/crashstats/crashstats/tests/test_sentry.py``
+
+
 Processing crashes
 ==================
 
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,9 @@ updates:
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 10
+    ignore:
+      - dependency-name: python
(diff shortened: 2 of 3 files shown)


docs/dev.rst
```diff
@@ -646,6 +646,23 @@ pick up changes:
...
+
+* ``.devcontainer/Dockerfile``
+* ``.github/dependabot.yml``
+* ``.readthedocs.yaml``
+* ``docker/Dockerfile``
+* ``docker/Dockerfile.fakesentry``
+* ``docker/set_up_legacy_es_requirements.sh``
+* ``pyproject.toml``
+* ``socorro/tests/processor/test_processor_app.py``
+* ``webapp/crashstats/crashstats/tests/test_sentry.py``
```
willkg, line 663: That's unfortunate about the last two. If we forgot to update those two tests, they would fail so we'll definitely remember as opposed to the rest of these files where there's nothing that fails causing us to remember.

We could write a `ANY_ENDSWITH` that always matches if the other thing is a string and ends with some specified substring. That would let us partial match and gloss over the python 3.11 specific file path. Maybe something like this:

```python
class ANY_ENDSWITH:
    def __init__(self, endstr):
        self.endstr = endstr

    def __eq__(self, other):
        return (
            isinstance(other, str) and other.endswith(self.endstr)
            or isinstance(other, ANY_ENDSWITH) and self.endstr == other.endstr
        )

    def __ne__(self, other):
        return not self.__eq__(other)


print("/some/python3.11/path/base.py" == "/some/python3.12/path/base.py")
# False
print("/some/python3.11/path/base.py" == ANY_ENDSWITH("path/base.py"))
# True
print(ANY_ENDSWITH("path/base.py") == "/some/python3.11/path/base.py")
# True
print(["a", "b", "/some/python3.11/path/base.py"] == ["a", "b", ANY_ENDSWITH("path/base.py")])
# True
print(ANY_ENDSWITH("path/base.py") == ANY_ENDSWITH("path/base.py"))
# True
```

smarnach: Sounds reasonable. I'd personally just use `ANY`, since I can't really think of a situation where a change in the `abs_path` of a stack frame points to a bug in our code.

smarnach: Should I file a bug for this? We have a similar thing in Tecken. We could also decide it's unfortunate, but ultimately irrelevant.

relud: seems like the kind of case where we could use `sys.version_info` to define the expected value

PR #6776, opened 2024-10-31
bug-1911367: Remove processor heartbeat and process metrics

willkg:
This removes processor heartbeat (and the scaffolding for that heartbeat) which generated process metrics for the processor container. We don't need those anymore--the problem those were added to help us understand is long gone now.

### DIFF
--- a/socorro/lib/task_manager.py
+++ b/socorro/lib/task_manager.py
@@ -9,8 +9,6 @@ import time
 
 LOGGER = logging.getLogger(__name__)
 
-HEARTBEAT_INTERVAL = 60
-
 
 def default_task_func(a_param):
     """Default task function.
@@ -19,16 +17,6 @@ def default_task_func(a_param):
     to demonstrate the api and not really for any other purpose.
 
     """
-
-
-def default_heartbeat():
-    """Runs once a second from the main thread.
-
-    Note: If this raises an exception, it could kill the process or put it in a
-    weird state.
-
-    """
-    LOGGER.info("THUMP")
 
 
 def default_iterator():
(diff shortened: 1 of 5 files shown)


socorro/lib/task_manager.py
```diff
@@ -76,7 +64,6 @@ def __init__(
         idle_delay=7,
         quit_on_empty_queue=False,
         job_source_iterator=default_iterator,
-        heartbeat_func=default_heartbeat,
```
willkg, line 79: Socorro still has this TaskManager and ThreadedTaskManager which is the last remnants for how socorro service processes were structured a long long time ago.

In this PR, I removed the heartbeat function from the processor app as well as the scaffolding for heartbeats for any app that uses the task managers figuring we don't need that anymore. (We can always re-add it if we do need it.)

socorro/lib/task_manager.py
```diff
@@ -109,7 +94,7 @@ def _get_iterator(self):
         job_source_iterator can be one of a few things:
 
         * a class that can be instantiated and iterated over
-        * a function that returns an interator
+        * a function that returns an iterator
```
willkg, line 97: I set up pre-commit and it's now fixing typos.

socorro/processor/processor_app.py
```diff
@@ -270,73 +269,13 @@ def _set_up_task_manager(self):
...
-                METRICS.gauge(
-                    "processor.processes_by_type", val, tags=[f"proctype:{proc_type}"]
-                )
-            for status, val in processes_by_status.items():
-                METRICS.gauge(
-                    "processor.processes_by_status", val, tags=[f"procstatus:{status}"]
-                )
-
-        except Exception as exc:
-            sentry_sdk.capture_exception(exc)
```
willkg, line 338: Buh-bye!

relud (approved):

willkg:
Thank you!

This change will be very nice because it'll result in a lot less console output from the processor.

PR #6777, opened 2024-10-31
bug-1901997: handle invalid BuildID for redacted crash reports

willkg:
If the user viewing the crash report does not have protected data access and the crash report has a crash annotation value that doesn't match our schema, then the redaction process can fail. In this scenario, Crash Stats can't recover and redact the document properly.
    
In bug 1901997, we were seeing this scenario when `BuildID=null`. Because crash annotation data comes from the crash reporter which might be running on a machine that has bad hardware, this scenario can happen with other fields.
    
These changes handle that scenario and shows a page indicating the crash report is malformed and providing a link the user can use to report the malformed crash report.

While doing this, I adjusted `bin/process_crashes.sh` so it allows you to pause it before it uploads data to the bucket. This makes it easier to adjust crash reports to test specific scenarios.


relud (approved):

willkg:
Thank you!

PR #6778, opened 2024-10-31
bug-1920739: update sentry_sdk to 2.17.0

willkg:
This updates sentry_sdk to 2.17.0. In order to do that, we had to do two fundamental changes:

1. Switch `sentry_sdk.push_scope()` calls to `sentry_sdk.new_scope()` which does roughly the same thing.
2. Switch from using `sentry_sdk.add_extra()` to `sentry_sdk.set_context()`. `add_extra()` was nice in that it allowed us to add additional things to the sentry event payload that gave context to the error. We used this to capture the processor rule, ruleset, crash_id, signature generation rule, etc in an additive way.

   `add_extra()` is deprecated and Sentry suggests to switch to `set_context()` which is not an additive kind of thing. So now we have to set a new context section everywhere we set the context. That's a little annoying, but it'll be fine.

I updated the code and tests accordingly.

I verified that the final context for a sentry event looks correct. Here's a screenshot of the `context` section in Kent for an error kicked up by a signature generation rule:

2024-10-31_sentry_context

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -48,7 +48,7 @@ requests==2.32.3
 requests-mock==1.12.1
 ruff==0.7.1
 semver==3.0.2
-sentry-sdk==2.8.0
+sentry-sdk==2.17.0
 Sphinx==8.1.3
 sphinx_rtd_theme==3.0.1
 statsd==4.0.1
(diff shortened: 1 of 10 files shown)


relud (approved):

willkg:
Thank you!

PR #6779, opened 2024-10-31
bug-1901998: fix buginfo view to stop sending sentry errors

willkg:
When Bugzilla is in a mood, the buginfo view queries the BugInfo API and that API throws a requests.exceptions.RetryError which the buginfo view didn't handle and thus we end up with a Sentry error.

When this happens, the report view Bugzilla tab gets back an HTTP 500 from querying the buginfo view and kind of shrugs and doesn't show bug details. The user isn't really impacted--they can continue their merry day.

2024-10-31_bugzilla_tab

Bugzilla seems to be down a lot, so we get these Sentry errors we're not going to do anything with. Thus I did a couple of things:

1. increase the backoff_factor for Bugzilla requests giving us a little more time for Bugzilla to find its happy place
2. change the buginfo view to handle the RetryError by returning an HTTP 500 with an error message

After this, the user sees the same thing (maybe we could improve that a bit, but I'm not going to do that in this bug) and we _don't_ get a Sentry event we're not going to do anything with.


webapp/crashstats/crashstats/models.py
```diff
@@ -57,7 +57,7 @@ def get_bugs_and_related_bugs(self, signatures):
 
 
 class BugAssociation(models.Model):
-    """Specifies assocations between bug ids in Bugzilla and signatures"""
+    """Specifies associations between bug ids in Bugzilla and signatures"""
```
willkg, line 60: I set up pre-commit and it's now fixing typos before I can create commits.

webapp/crashstats/crashstats/views.py
```diff
@@ -332,8 +333,10 @@ def buginfo(request, signatures=None):
     bug_ids = form.cleaned_data["bug_ids"]
 
     bzapi = models.BugzillaBugInfo()
-    result = bzapi.get(bug_ids)
-    return result
+    try:
+        return bzapi.get(bug_ids)
+    except RetryError:
+        return {"error": "Max retries exceeded with Bugzilla."}, 500
```
willkg, line 339: This return is a structure that `utils.json_view` recognizes and will split into "response, status".

relud (approved):

willkg:
Thank you!

webapp/crashstats/crashstats/models.py
biancadanforth: Does that work for all local development, or is that config specific to your local development? If the former, where was that added?

willkg: I used it a long time ago and then decided I didn't like the way it was structured. Rehan built Therapist which I used for a while, but I think that's unmaintained now.

I noticed SRE uses pre-commit for their repositories and decided to try it out again.

https://github.com/mozilla-it/webservices-infra/blob/main/.pre-commit-config.yaml
https://github.com/mozilla/terraform-modules/blob/main/.pre-commit-config.yaml
https://github.com/mozilla-it/global-platform-admin/blob/main/.pre-commit-config.yaml

I was going to give it a little more time and then bring it up at an Observeneers meeting.

In the meantime, it's fixing spelling in PRs I'm doing in files I made a change to and I wanted to point out that's a cosmetic change.

PR #6780, opened 2024-11-01
bug-1863963: fix site title to Mozilla Crash Stats

willkg:
"crash reports" is the name of the collector, so it's always been a dissonance that the header of the Crash Stats site said Crash Reports.

While doing this, I also fixed the hierarchy of error templates and removed an unneeded header_title.html template.

Currently we have:

image

This fixes it to:

image


relud (approved):

willkg:
Thank you!

PR #6781, opened 2024-11-01
bug-1851708: fix ESRVersionRewriteRule and FenixVersionRewriteRule

willkg:
This fixes these two rules to stop editing the raw crash--that's soo bad!

Now they run after we copy everything from the raw crash to the processed crash (validated and normalized), use data from the processed crash to figure out what to do, and edit data in the processed crash if necessary.


relud (approved):

willkg:
Thank you!

PR #6782, opened 2024-11-01
bug-1764633: remove non-printable/non-ascii characters from signature

willkg:
The crash signature contains data from stackwalker output as well as JavaException annotations and other places, too. Because of this, it's possible for the signature to contain non-ascii and non-printable characters. This rule restricts signatures to printable ASCII characters.


relud (approved):

willkg:
Thank you!

PR #6783, opened 2024-11-02
bug-1902001: fix source highlight with bad urls

willkg:
If the url provided isn't a valid url, treat it as if the document doesn't exist and return an HTTP 404.

To test:

```
> curl "http://localhost:8000/sources/highlight/?url=https://crash-stats.allizom.org\[@a.xxx.org/?"
```

Compare that with:

```
> curl "https://crash-stats.allizom.org/sources/highlight/?url=https://crash-stats.allizom.org\[@a.xxx.org/?"
```

(You need the `\[` because otherwise curl thinks the `[` means something and kicks up `curl: (3) bad range specification in URL position 78:`.)

### DIFF
--- a/webapp/crashstats/sources/tests/test_views.py
+++ b/webapp/crashstats/sources/tests/test_views.py
@@ -67,6 +67,14 @@ def test_highlight_url(client, requests_mock):
     assert response["content-security-policy"]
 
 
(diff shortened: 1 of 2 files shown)


relud (approved):

willkg:
Thank you!

PR #6785, opened 2024-11-02
bug-1892787: handle case where Bugzilla returns bad payload

willkg:
Sometimes Bugzilla returns an HTTP 200 with JSON content, but it doesn't have "bugs" in it.

Rather than throw an HTTP 500 because of an unhandled `KeyError`, this changes the code to raise a `BugzillaRestHTTPUnexpectedError` with the payload it did get. This will give us a better idea of what's happening so we can figure out what to do about it if anything.


relud (approved):

willkg:
Thank you!

PR #6786, opened 2024-11-04
bug-1928934: fix super search fields in signature report forms

willkg:
In 9b425e50, I changed the name of the field the webapp permissions were stored in, but missed updating the signature report view. This fixes that and adds a test.

To test:

1. process a crash report
2. go to `localhost:8000`
3. if you're logged in, log out
4. go to the crash report then click on the signature to go to the signature report
5. go to the "Aggregations" tab and click on the "Aggregate on..." drop down -- there should be fields
6. go to the "Reports" tab -- there should be fields listed in the unlabeled columns field
7. go to the "Graphs" tab and click on "Crashes per day, by ..." drop down -- there should be fields

### DIFF
--- a/webapp/crashstats/documentation/jinja2/docs/supersearch/api.html
+++ b/webapp/crashstats/documentation/jinja2/docs/supersearch/api.html
@@ -92,7 +92,7 @@
 
         <p class="description">
           Use this parameter to get a large number of results instead of setting
-          an arbitraty big <code>_results_number</code> value. Run queries in a
+          an arbitrary big <code>_results_number</code> value. Run queries in a
           loop, incrementing this by the value of <code>_results_number</code>,
           and concatenate the content of the <code>hits</code> key.
         </p>
@@ -146,7 +146,7 @@
           run an aggregation on <code>field_1</code>, then for each bucket of
           that aggregation, it will aggregate on <code>field_2</code>, and then
           for each bucket of that sub-aggregation, it will aggregate on
-          <code>field_3</code>. Theoratically, we could have any number of
+          <code>field_3</code>. Theoretically, we could have any number of
           levels of aggregations, but in practice we only allow all "level 1"
           aggregations, and a few "level 2" and "level 3".
         </p>
@@ -300,6 +300,9 @@
               {% endfor %}
             </p>
 
+            {# NOTE(willkg): we use permissions_needed so it matches the
+            protected crash schema rather than the webapp permissions which
+            users won't understand #}
             {% if filter.permissions_needed %}
               <p class="permissions">{{ filter.permissions_needed | join(', ') }}</p>
             {% endif %}
--- a/webapp/crashstats/signature/tests/test_views.py
+++ b/webapp/crashstats/signature/tests/test_views.py
@@ -9,10 +9,12 @@ from urllib.parse import quote
 
 import pyquery
 
+from django.contrib.auth.models import AnonymousUser
 from django.urls import reverse
 from django.utils.encoding import smart_str
 
 from crashstats.crashstats import models
+from crashstats.signature.views import get_fields
 from socorro.lib.libdatetime import date_to_string, utc_now
 from socorro.lib.libooid import create_new_ooid, date_from_ooid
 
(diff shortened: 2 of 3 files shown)


webapp/crashstats/signature/tests/test_views.py
```diff
@@ -661,3 +663,16 @@ def test_signature_bugzilla(self, client, db, es_helper):
...
+        user = AnonymousUser()
+        fields = get_fields(user)
+        len_public_fields = len(fields)
+        assert len(fields) > 0
+
+        # Create user with protected data access, get fields, and make sure there are
+        # more of them than if the user didn't have protected data access
+        user = user_helper.create_protected_user()
+        fields = get_fields(user)
+        assert len(fields) > len_public_fields
```
willkg, line 678: Super search fields change over time, so I didn't want to create a test we had to update periodically. Instead, this asserts there are fields in the case where the user doesn't have protected data access and asserts that if the user does have protected data access, the user sees more fields than if the user didn't. That feels like a good enough set of assertions that aren't fragile.

webapp/crashstats/documentation/jinja2/docs/supersearch/api.html
```diff
@@ -300,6 +300,9 @@ <h2 id="param-{{ filter.name }}">{{ filter.name }}</h2>
               {% endfor %}
             </p>
 
+            {# NOTE(willkg): we use permissions_needed so it matches the
+            protected crash schema rather than the webapp permissions which
+            users won't understand #}
```
willkg, line 305: I added this note here so someone doesn't think this is a mistake. This will cause the data dictionary to show "public" and "protected" (processed crash schema permission terms) rather than "" and "crashstats.view_pii" (webapp-centric terms). The latter set of permissions are an internal implementation detail using terms we don't use in our documentation.

webapp/crashstats/documentation/jinja2/docs/supersearch/api.html
```diff
@@ -92,7 +92,7 @@ <h2 id="param-_results_offset">_results_offset</h2>
 
         <p class="description">
           Use this parameter to get a large number of results instead of setting
-          an arbitraty big <code>_results_number</code> value. Run queries in a
```
willkg, line 95: I'm pretty arbitraty sometimes.

webapp/crashstats/documentation/jinja2/docs/supersearch/api.html
```diff
@@ -146,7 +146,7 @@ <h2 id="param-_aggs.*">_aggs.*</h2>
           run an aggregation on <code>field_1</code>, then for each bucket of
           that aggregation, it will aggregate on <code>field_2</code>, and then
           for each bucket of that sub-aggregation, it will aggregate on
-          <code>field_3</code>. Theoratically, we could have any number of
```
willkg, line 149: _Theoradically_ should be a word. It's not misspelled that way here, but it made me think a bit about it.

webapp/crashstats/signature/views.py
```diff
@@ -84,17 +105,8 @@ def signature_report(request, params, default_context=None):
...
-        and x["name"] != "signature"  # exclude the signature field
-    )
-    context["fields"] = [
-        {"id": field, "text": field.replace("_", " ")} for field in fields
-    ]
+    fields = get_fields(request.user)
+    context["fields"] = fields
 
     columns = request.GET.getlist("_columns")
     columns = [x for x in columns if x in fields]
```
relud, line 112: fields now contains dictionaries instead of strings, but it looks here like it's being checked for containing strings? and then the test still passes because the default columns are being used when nothing matches?

relud (approved):
r+ with comment

webapp/crashstats/signature/views.py
```diff
@@ -71,6 +71,27 @@ def inner(request, *args, **kwargs):
...
 
+def get_fields(user):
+    """Retrieve super search fields this user has access to
+
+    :arg user: a Django User instance
+
+    :returns: a list of dicts with "id" and "text" keys
+
+    """
+    print(repr(user), user)
```
willkg, line 82: Oops--need to remove this.

webapp/crashstats/signature/views.py
willkg: Oh, whoops--this is a mistake. It's not related to the test because the test only tests `get_fields()` by itself.

willkg:
^^^ @relud Does that look right to you?

relud (approved):
lgtm

PR #6797, opened 2024-11-04
Move extra dockerfiles into docker/images/*

relud:
to match antenna/eliot/tecken. Fixes syntax highlighting in editors. Makes it more clear when files (e.g. `fixtures.json`) are for specific images.

convenient side-effect is that this causes these images to build faster on mac because less context needs to be transferred to the docker vm to build the image


willkg (approved):
Very nice!

PR #6798, opened 2024-11-07
Bug 1929829 - Ignore g_malloc when generating signatures

gabrielesvelto:


willkg (approved):
```
app@socorro:/app$ ./mybin/test_signature_change.sh g_malloc
Working on 'g_malloc' ...
Crash id: 9c924bb4-cc8d-4071-afef-000140241031
Original: g_malloc
New:      g_slice_alloc
Same?:    False

Crash id: 72d845fc-6f33-4f44-9203-1dc490241104
Original: g_malloc
New:      g_strndup
Same?:    False
```

Doesn't seem like it helps a whole lot, but you can do a follow-up PR with additional changes.

PR #6802, opened 2024-11-12
OBS-144: Migrate from socorro-release to obs-common

relud:


willkg (approved):
I double-checked that building and running a local dev environment works. Everything looks good. :100:

willkg:
I messed up the validation here. We didn't test processing crash reports in the local dev environment. `bin/process_crashes.sh` needs to be updated. Also, we missed one more reference in `docs/dev.rst` to `socorro-cmd pubsub` which no longer exists.

@relud Can you fix these two things? ^^^

relud:
fix in https://github.com/mozilla-services/socorro/pull/6813

PR #6803, opened 2024-11-13
Migrate from make to just

relud:
also switch from `SOCORRO_UID` and `SOCORRO_GID` to `USE_UID` and `USE_GID`
also switch from `my.env` to `.env`
also switch to using `docker-compose.override.yml` for volume mounts, so they can be disabled in CI.


.github/workflows/build-and-push.yml
```diff
@@ -18,7 +18,10 @@ jobs:
       contents: read
       deployments: write
       id-token: write
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
relud, line 21: `ubuntu-latest` currently points to `ubuntu-22.04`, we need `ubuntu-24.04` so we can install `just` via `apt`.

justfile
```diff
@@ -0,0 +1,88 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+_default:
+  @just --list
+
+_env:
+  @if [ ! -f .env ]; \
```
willkg, line 9: I think I would have done four space indents. It's a little easier to read, it's what the just examples use, etc. But this is probably fine.

Also, instead of making this a one-liner, you can make it an inline bash script:

https://just.systems/man/en/shebang-recipes.html

Then we don't have to deal with the `\` stuff.

willkg (approved):
Looks good!

PR #6810, opened 2024-11-19
Work around dependabot not updating docker-compose

relud:
this leaves out `hopsoft/graphite-statsd` and `mozilla/socorro_collector:latest` because they are using "latest" which will get updates without dependabot, and leaves out `mozilla/socorro_elasticsearch:1.4.5` because it's going away


smarnach (approved):
Looks good to me.

> this leaves out `hopsoft/graphite-statsd` and `mozilla/socorro_collector:latest` because they are using "latest" which will get updates without dependabot.

I've got two questions about this. First, I don't think we are still updating `mozilla/socorro_collector`, since it's on Docker Hub, and we now only push to GAR. Is this intended? And second, _should_ we be using `latest`? At least for statsd, I think it would be better to pin it to a specific version, in particular now that we have a way to make dependabot take care of upgrades.

relud:
i should make the antenna image public and use the gar version.

for statsd, iirc the image hasn't been touched in like 7 years, and there is no equivalent non-latest image tag, so latest is what we get until we have time to set up a replacement.

PR #6823, opened 2024-11-27
bug-1933774: omit collector-added fields from Crash Annotations tab

willkg:
This removes the collector-added fields from the Crash Annotations tab. For anyone who isn't a crash ingestion system maintainer, these fields are confusing because they're not part of the crash report payload.

After these changes, the fields will be omitted from the public and protected annotations lists in the Crash Annotations tab.

This also adds "version" to the Debug tab. This value represents the schema version of the raw crash structure. It's relevant to crash ingestion internals and nothing else, so having it in the Debug tab is helpful (though at this point, everything is version 2 so it's not wildly interesting anymore).

This also removes a few lines from the report view template that said we could remove them in 2022. That day has long passed.

To test:

1. process a crash report
2. view the crash report--make sure "uuid", "version" (lowercase v--"Version" (uppercase v) is an annotation), "metadata", and "submitted_timestamp" don't show up in the "Crash Annotations" tab
3. log in using an account with protected data access
4. view the crash report--make sure "uuid", "version" (lowercase v--"Version" (uppercase v) is an annotation), "metadata", and "submitted_timestamp" don't show up in the "Crash Annotations" tab

### DIFF
--- a/webapp/crashstats/crashstats/jinja2/crashstats/report_index.html
+++ b/webapp/crashstats/crashstats/jinja2/crashstats/report_index.html
@@ -909,17 +909,19 @@
             </div>
             <table class="record data-table hardwrapped">
               <tbody>
-                {% for key in public_raw_keys %}
-                  <tr title="{{ fields_desc.get(make_raw_crash_key(key), empty_desc) }}">
-                    <th class="w-2/12" scope="row">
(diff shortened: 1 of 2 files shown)


willkg:
Thank you!

PR #6831, opened 2024-12-05
Remove system checklist

willkg:
We moved the system checklist to Confluence, so this is the unmaintained version and we should remove it.


relud (approved):

willkg:
Thank you!

PR #6832, opened 2024-12-06
Update obs-common to pick up the ack_deadline_timeout fix

willkg:
This fixes subscription creation in the local dev environment so the ack deadline is 600 seconds which gives the processor enough time to retrieve a crash id, process the crash, and ack the crash id. Prior to this fix, subscriptions would use the default 10 seconds and that's not enough time to process crash reports.

We fixed this originally in bug-1912979, but that fix was dropped when we moved utilities to obs-common.


relud (approved):

willkg:
Thank you!

PR #6849, opened 2024-12-30
bug-1911367: remove psutil dependency

willkg:
We removed the processor process metrics, so we don't need this dependency anymore.


biancadanforth (approved):
LGTM --  I was trying to understand what the process metrics were, why they were added in the first place, and why they are no longer needed.

I clicked down into all the history of this, and I think I found my answers largely via [this comment ](https://bugzilla.mozilla.org/show_bug.cgi?id=1795017#c19)(and the ticket its part of, and the ticket it's blocked by). Very tricksy!

Also, I confirmed there's no other references to `psutil` in the codebase.

willkg:
Thank you!

willkg:
Oh, I think I see what's going on. It's not re-running the checks, so the "Merge when ready" relates to the original run from last year and doesn't reflect the failed runs from the merge queue attempts. I think this is a beginning-of-year issue with one of the tests that uses week number for the index. I remember fixing something like this a few years back.

I'll look into it and fix it in another PR.

willkg:
PR #6853 fixes the test. Once that lands, I'll rebase this and it should be mergeable.

PR #6864, opened 2025-01-13
OBS-385: Introduce ESBuild for CSS and Images

toufali:
Using a modern build tool such as ESBuild allows us to use best-practice CSS features such as `@import` and advanced optimizations such as bundling, transpilation, and minification without the need for additional 3rd party tools and complex configuration.

This PR is specifically limited to CSS and related image assets, with the goal of introducing ESBuild working in tandem with collectstatic and django-pipeline, which both still process JS files.  A PR for using ESBuild on JS files and the complete removal of django-pipeline is forthcoming.  (Collectstatic will still remain, however its only concern will be the internal Django admin site.) 

**Follow-up PRs needed:**
- ~Add "watch" feature to avoid tearing down, rebuilding, and restarting container/server on every CSS/JS change:~ ~https://github.com/mozilla-services/socorro/pull/6913~  Feature is now included in this PR.
- Fix host machine permissions for `/app/webapp/static/` and `/app/webapp/node_modules/`, in particular for Linux users
- Handle JS file builds with ESBuild, removing django-pipeline altogether

**To test this PR:**
Run the app in both development and production modes following this test plan: https://mozilla-hub.atlassian.net/wiki/spaces/CS1/pages/605028396/Socorro+manual+test+plan#Webapp

1. `just build` and look out for an ESBuild summary that ends similar to "⚡ Done in 218ms".  The collectstatic summary should follow, similar to "328 static files copied to '/app/webapp/static', 4 unmodified, 23 post-processed."
2. `just setup`
3. `just run`
4. You should now be running in development. Follow [test plan](https://mozilla-hub.atlassian.net/wiki/spaces/CS1/pages/605028396/Socorro+manual+test+plan#Webapp) to ensure app functions as expected.
5. To test production, ctrl-c to stop the server.
6. `docker compose run --service-ports webapp shell`
7.  `npm run build --prefix webapp` (this step – and the similar collectstatic step [in the docs](https://socorro.readthedocs.io/en/latest/service/webapp.html#production-style-assets) might not be needed if files haven't changed since the last build?)
8. `DEBUG=False bash bin/run_service_webapp.sh`
9. You should now be running in production. Follow [test plan](https://mozilla-hub.atlassian.net/wiki/spaces/CS1/pages/605028396/Socorro+manual+test+plan#Webapp) to ensure app functions as expected.


webapp/crashstats/crashstats/jinja2/crashstats_base.html
```diff
@@ -16,10 +16,7 @@
       {% endblock %}
     {% endblock %}
 
-    {% block site_css %}
-      {% stylesheet 'crashstats_base' %}
-      {% stylesheet 'fontawesome' %}
```
toufali, line 21: I searched for the classes and fonts specified by the "fontawesome" stylesheet but couldn't find any used.  This stylesheet is apparently only used on this page.  This stylesheet may have originally been used for icons, but I notice [other fonts being specified for icons](https://github.com/mozilla-services/socorro/blob/10e257162602260e453d889cf7840bd44da7171e/webapp/crashstats/crashstats/static/crashstats/css/base/typography.css), likely replacing the fontawesome package.

willkg: Jinja templates are hierarchical, so many of the templates in crashstats *extend* this template.

```
$ rg -l crashstats_base.html
crashstats/topcrashers/jinja2/topcrashers/topcrashers.html
crashstats/tokens/jinja2/tokens/home.html
crashstats/documentation/jinja2/docs/docs_base.html
crashstats/supersearch/jinja2/supersearch/search.html
crashstats/profile/jinja2/profile/profile.html
crashstats/supersearch/jinja2/supersearch/search_custom.html
crashstats/profile/tests/test_views.py
crashstats/crashstats/jinja2/crashstats/report_index_malformed_raw_crash.html
crashstats/crashstats/jinja2/crashstats/home.html
crashstats/crashstats/jinja2/crashstats/report_index_not_found.html
crashstats/crashstats/jinja2/crashstats/product_home.html
crashstats/crashstats/jinja2/crashstats/login.html
crashstats/crashstats/jinja2/crashstats/about_throttling.html
crashstats/crashstats/jinja2/crashstats/report_index_pending.html
crashstats/signature/jinja2/signature/signature_report.html
crashstats/crashstats/jinja2/crashstats/ratelimit_blocked.html
crashstats/crashstats/jinja2/crashstats/report_index.html
```

I see this commit where I redid how we use fontawesome back in 2020:

https://github.com/mozilla-services/socorro/commit/33935348a8fbcd5e69b70bc3ed00eddb4a146c36

Then I removed the Fission bits (Fission is a Firefox re-architecture project that's been done for a while) in 2022:

https://github.com/mozilla-services/socorro/commit/721d41c43f8f4eca4b6896e5bb45ce16ed29d3f9

I think you're right in that we're not using fontawesome anymore and we can remove it.

webapp/package.json
```diff
@@ -33,5 +34,10 @@
     "qs": "6.9.4",
     "Select2": "github:select2/select2#3.5.4",
     "tablesorter": "2.31.3"
+  },
+  "scripts": {
+    "build": "rm -rf static/* && npm run copy:images && node esbuild && npm run build:collectstatic",
+    "build:collectstatic": "TOOL_ENV=True ./manage.py collectstatic --noinput --ignore css --ignore img",
```
toufali, line 40: Note: the two `--ignore` flags also ignore css and images for Django admin.  From my research, the glob-style flags are notoriously brittle/unintuitive – I was unable to set up `crashstats/**/*.css` and `crashstats/**/img/`.  Once JS files are included in the new build process, we can more easily do `--ignore crashstats` so that collectstatic's only concern is admin.

We should remove both ignore flags before merge, but I left them here and in `test.sh` to illustrate/confirm that ESBuild is in fact building/responsible for CSS/img files.

bin/test.sh
```diff
@@ -50,6 +50,6 @@ echo ">>> run tests"
 
 # Collect static and then run pytest in the webapp
 pushd webapp
-${PYTHON} manage.py collectstatic --noinput
+${PYTHON} manage.py collectstatic --noinput --ignore css --ignore img
```
toufali, line 53: See note below for webapp/package.json

willkg (changes requested):
I skimmed through this and as far as I can tell, it needs some additional changes to work the way we talked about.

First, I think some of the static assets aren't in the right place. In the local dev environment, Django runs with `DEBUG=True` using the `runserver` management command, In that mode, it serves assets from the Django app static folders. That mode works with these changes.

However, in our stage and prod environments, Django runs with `DEBUG=False` using `gunicorn` and static assets are served from the `static/` folder. If we follow the directions here:

https://socorro.readthedocs.io/en/latest/service/webapp.html#production-style-assets

Then we can simulate the stage and prod environments. When we do that, some static assets aren't getting served. I'm seeing HTTP 404s for JS and SVG files for the pages I looked at. I couldn't get very far since JS files weren't getting served and that's affects a large chunk of the navigation.

Second, this doesn't implement something that watches CSS and JS files for changes and rebuilds static assets. In main tip now, you can run the web server and edit JS and CSS files and they get rebuilt and served automatically. With this PR, every time you make a change to JS and CSS files, you have to rebuild the Docker image, set everything back up, and start the webapp container. That's a big step backwards for developer ergonomics. There's nothing in the PR description or followup comments about whether we planned to fix that in a followup PR.

The other thing this needs is a followup PR to fix how `/app/webapp/static/` and `/app/webapp/node_modules/` get created on the host machine so they have the right permissions. I offered to do that change since it largely affects Linux users, but I wanted to note that it needs to be done and that's not captured anywhere.

In one of our talks, we discussed having esbuild use a different directory tree than the existing `webapp/crashstats/APP/static/` for source files that it managed. That alleviates the problems you were having that caused you to add `--ignore` to collectstatic. Did that solution not work? It's cleaner and clearer than the one implemented in this PR and we know it'll work correctly with Django apps like the Django admin.

This is a complicated enough change that it's worth formalizing a test plan for it and making sure we go through the test steps as we make changes and review. There's no test plan in the PR description or the project proposal. We should formalize one and include it in the description.

toufali:
@willkg – thanks for taking a look!

> However, in our stage and prod environments ... some static assets aren't getting served. I'm seeing HTTP 404s for JS and SVG files for the pages I looked at. 

I think this issue was twofold:
- I jumped the gun in replacing part of django-pipeline, forgetting it's still processing JS with this PR. Should be fixed via https://github.com/mozilla-services/socorro/pull/6864/commits/d39b1de273034b7477b02193afc51d4816cf2c55
- I didn't provide instruction on how to manually build for production, and so your test was missing a build step for images.  I'll add to the description.

> Second, this doesn't implement something that watches CSS and JS files for changes and rebuilds static assets. 

From our earlier discussion, I planned to add "watch" via [separate PR](https://github.com/mozilla-services/socorro/pull/6913) to keep this one's size/scope down.  If you think those changes are better suited here I can add!  Else I will update the description to call out this follow-up.

> The other thing this needs is a followup PR to fix how `/app/webapp/static/` and `/app/webapp/node_modules/` get created on the host machine so they have the right permissions. 

I will capture this in the description, thanks for the reminder.

> In one of our talks, we discussed having esbuild use a different directory tree than the existing `webapp/crashstats/APP/static/` for source files that it managed. That alleviates the problems you were having that caused you to add `--ignore` to collectstatic. Did that solution not work? It's cleaner and clearer than the one implemented in this PR and we know it'll work correctly with Django apps like the Django admin.

I updated the collectstatic `--ignore` flag to be more specific and not omit the admin folder.  It's a very long line unfortunately.  This is temporary until we add ESBuild support for JS – the flags will become shorter, though still unsightly in my opinion: `-i crashstats -i api ...` instead of `-i crashstats/*.css -i api/*.css ...`  It looks like there's a [more elegant way to accomplish this](https://docs.djangoproject.com/en/5.1/ref/contrib/staticfiles/#customizing-the-ignored-pattern-list), but I didn't have time to figure this out with my limited Django/Python knowledge!  I'm not sure how using a different directory for ESBuild helps here?  I'm trying to avoid collectstatic needlessly processing files that ESBuild has already done.  Would love to chat more about this in case I'm missing something.

> This is a complicated enough change that it's worth formalizing a test plan for it and making sure we go through the test steps as we make changes and review. There's no test plan in the PR description or the project proposal. We should formalize one and include it in the description.

💯 I will add this now.

webapp/package.json
```diff
@@ -33,5 +34,10 @@
     "qs": "6.9.4",
     "Select2": "github:select2/select2#3.5.4",
     "tablesorter": "2.31.3"
+  },
+  "scripts": {
+    "build": "rm -rf static/* && npm run copy:images && node esbuild && npm run build:collectstatic",
+    "build:collectstatic": "TOOL_ENV=True ./manage.py collectstatic --noinput -i crashstats/*.css -i api/*.css -i documentation/*.css -i profile/*.css -i signature/*.css -i status/*.css -i supersearch/*.css -i tokens/*.css -i topcrashers/*.css",
```
willkg, line 40: This is duplicated in `bin/test.sh`. Either `bin/test.sh` should call `npm run build:collectstatic` or we should pull the two calls into a shell script and call that. It's sufficiently complex that we want to centralize it into one place.

toufali: fixed via https://github.com/mozilla-services/socorro/pull/6864/commits/2bb3372cff60ac8d9a052f8ae503138ba6453c9e

bin/test.sh
```diff
@@ -50,6 +50,6 @@ echo ">>> run tests"
 
 # Collect static and then run pytest in the webapp
 pushd webapp
-${PYTHON} manage.py collectstatic --noinput
+${PYTHON} manage.py collectstatic --noinput -i crashstats/*.css -i api/*.css -i documentation/*.css -i profile/*.css -i signature/*.css -i status/*.css -i supersearch/*.css -i tokens/*.css -i topcrashers/*.css
```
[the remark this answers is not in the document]
toufali: addressed via https://github.com/mozilla-services/socorro/pull/6864/commits/2bb3372cff60ac8d9a052f8ae503138ba6453c9e

toufali:
>
> PR #6913 covers adding a watch mode. We should include that in this PR. We're on a tight schedule and those two things need to go all-or-nothing and including that PR in this one will (hopefully) reduce the overall iterations of review-and-fix.
> 

Watch mode added via https://github.com/mozilla-services/socorro/pull/6864/commits/857ab2780971575d1fb74475d57662b37f7e45a4

docker/config/local_dev.env
```diff
@@ -70,7 +70,7 @@ DEBUG=True
 # files generated in the image are not available. For local development,
 # static files for the webapp get put in /tmp/crashstats-static/ so we
 # need to set STATIC_ROOT to that.
-STATIC_ROOT=/tmp/crashstats-static/
+# STATIC_ROOT=/tmp/crashstats-static/
```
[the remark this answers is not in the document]
toufali: addressed via https://github.com/mozilla-services/socorro/pull/6864/commits/89d71d45ff69e9c34ec697a2c32ae963b1b4b666

webapp/crashstats/crashstats/tests/test_finders.py
```diff
@@ -1,30 +0,0 @@
...
-        filenames = busted_pipeline["STYLESHEETS"][key]["source_filenames"]
-
-        # add a junk one
-        filenames += ("neverheardof.css",)
-        busted_pipeline["STYLESHEETS"][key]["source_filenames"] = filenames
-
-        settings.PIPELINE = busted_pipeline
-
-        with pytest.raises(ImproperlyConfigured):
-            call_command("collectstatic", "--noinput", interactive=False)
```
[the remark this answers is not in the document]
toufali: The test was specific to CSS files with a single method `test_missing_css_source_file`, which stopped working after we removed the old CSS pipeline.  I think we should leave LeftoverPipelineFinder for now, since it involves JS files, and this PR attempts to leave JS builds untouched.

webapp/crashstats/settings/bundles.py
```diff
@@ -283,45 +163,3 @@
...
-
-# These are quality checks--primarily for developers. It checks that you haven't haven't
-# accidentally make a string a tuple with an excess comma, no underscores in the bundle
-# name and that the bundle file extension is either .js or .css.
-#
-# We also check, but only warn, if a file is re-used in a different bundle.  That's
-# because you might want to consider not including that file in the bundle and instead
-# break it out so it can be re-used on its own.
-_used = {}
-for config in PIPELINE_JS, PIPELINE_CSS:  # NOQA
```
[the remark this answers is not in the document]
toufali: addressed via https://github.com/mozilla-services/socorro/pull/6864/commits/831b90bdc47458d2963b207f9c08be51fde3a025

webapp/package.json
willkg: Given how long and unwieldy this line is, I think I would have moved this to a shell script and called the shell script from both `bin/test.sh` and here. That gives us syntax highlighting, shellcheck, and better ergonomics for maintaining it.

But this is fine for now.

willkg (approved):
This looks good. I went through the test plan and everything looked fine with that. I also verified we can edit static asset source files and they get rebuilt when running the webapp in a local dev environment.

* I tested a CSS file -- added and removed setting background to red; esbuild watch kicked in and updated it
* I tested updating a png file -- copied a different png over it; esbuild watch kicked in and updated it
* I tested a JS file -- added and removed an alert; django-pipeline sent new one in next request

I really like how esbuild watch works--it explicitly states that it saw something change and rebuilds it.

```
webapp-1      | [watch] build started (change: "crashstats/crashstats/static/crashstats/css/base/layout.css")
webapp-1      | [watch] build finished
```

That's very nice!

The only thing left here is docs. I'm going to approve this now and whatever docs changes you make we'll go with--no need for another review. We can do a followup fix later for anything we need to change.

Once this deploys to stage, we will need to go through the webapp test plan to make sure it's working correctly in a server environment. It probably will, but we should verify that.

webapp/package.json
toufali: ok, I'll keep an eye on it when we add JS to ESBuild's responsibility.

PR #6873, opened 2025-01-23
obs-443: switch to docker compose watch

willkg:
Currently for the local dev environment, we volume mount the local directory as `/app` which causes anything built during docker image build to be shadowed by the user's local directory. We get around this by having docker image build any artifacts into directories in the image that are not under `/app`. This is a little wonky and there are hacks strewn about to alleviate the consequences.

Instead of doing that, this PR is an initial pass and switching to docker compose watch. In this new model, we no longer volume mount the local directory as `/app` and instead docker compose watches for changes on the local file system and syncs them with the container and will run any rebuilding scripts.

This should work with the local dev environment, CI, and server environments.

This requires v2.22.0 of docker compose plugin.

This doesn't update documentation and doesn't document edge cases to be aware of.

How to test syncing changes from host file system to container file system:

1. run `just build` and `just setup` to build the image and set up storage
2. run `just run` to run services including the webapp service -- this passes `--watch` so that docker compose will watch the filesystem and perform specified actions
3. go to `http://localhost:8000/` -- this shows the home page
4. make a change to `webapp/crashstats/crashstats/jinja2/crashstats/home.html` by adding text to the `content` block and save the file
5. you'll see a "Syncing service webapp ..." message in the `just run` terminal indicating that docker compose has synced the change from the host file system to the container file system
6. refresh `http://localhost:8000/` and you should see your change because the jinja renderer pulls the template directly from the source location and doesn't require rebuilding anything or moving anything around

Current problems:

1. If you make changes to the source when _not_ running the container, you _have to rebuild the image before running the container_ otherwise those changes won't be in the container.
   1. Maybe `just run` should always build first, then run? Is that too onerous?
   7. What should we do about `just shell` and other just rules?
8. We have to add the develop -> watch thing to every service listed in `docker-compose.yaml` that needs it. It's not sufficient to do it for the `app` image. I think I got them all, but maybe I missed something.
9. CI will need at least docker compose version 2.22.0 otherwise it'll balk at the `develop` key. Does CI have the right version?
   1. It's running v2.27.1, so I think we're fine.

To do:

* [ ] probably should fix `just run` to always build first
* [ ] update developer documentation
* [ ] fix static asset building so it puts it in `/app` in a default place
   * pretty sure this means we don't have to run `./manage.py collectstatic` after `just testshell` and before running tests anymore
* [ ] add a `rebuild` action to rebuild static assets in the webapp when they change (css files, js files, images) -- need to call `manage.py collectstatic`
* [x] fix node modules installation so it puts it in `/app` in a default place
* [x] figure out what happens if the image builds node modules and then the user updates them--do they sync with the container? are there cases where the user ends up in a weird state and has to `just build` again?
   * 2025-01-30: We should be good here. We're going to add `node_modules` to `.dockerignore` so it won't ever get synced. We'll add `package.json` as a rebuild so if it ever gets updated on the host, that'll get synced to the container and `npm install` will get run thus updating the `node_modules`.


willkg:
@toufali This is the PR with the docker compose watch stuff in it. If we're going to go this route, it'll be easier to land this before landing your next set of changes.

Do you want to look at this and see if you can finish it up?

docker-compose.yml
```diff
@@ -27,6 +27,11 @@ services:
       - postgresql
       - legacy-elasticsearch
       - elasticsearch
+    develop:
+      watch:
+        - action: sync
+          path: .
+          target: /app
```
willkg, line 34: This includes the sync action which will copy files from the host to the container.

There's also a rebuild action which I think we want to use to run `manage.py collectstatic` and `esbuild` when js and css files change.

docker-compose.yml
```diff
@@ -114,6 +134,11 @@ services:
     command: ["webapp", "--dev"]
     ports:
       - "8000:8000"
+    develop:
+      watch:
+        - action: sync
+          path: .
+          target: /app
```
willkg, line 141: This includes the sync action which will copy files from the host to the container.

There's also a rebuild action which I think we want to use to run manage.py collectstatic and esbuild when js and css files change.

We only need this rebuild on the webapp service because that's the only one using js and css files.

toufali (commented):
Thanks for getting this started @willkg !  A couple questions:

Should we ignore the node_modules folder and similar?

Are we now able to change [this line ](https://github.com/mozilla-services/socorro/blob/259e80290e6cb77d3318602f2df825311c987976/docker/Dockerfile#L32) so Node can find modules?

toufali:
> @toufali This is the PR with the docker compose watch stuff in it. If we're going to go this route, it'll be easier to land this before landing your next set of changes.
> 
> Do you want to look at this and see if you can finish it up?

Yes!  Thanks for posting, I'll work on this.  See also my questions above that I'm happy to add if you agree.

toufali:
@willkg I've spent a bit of time testing this – using Compose Watch accomplishes the goal of not overwriting files in the app folder, as discussed.  However, I am not seeing the "sync" behavior as expected.  Would love to pair on this when you're able!

willkg:
@toufali talked with @relud and there are issues with Docker on macOS (that's a very hand-wavy summary). Further, this makes non-webapp-tweaking workflows much harder.

I wrote about it more in https://mozilla-hub.atlassian.net/browse/OBS-443?focusedCommentId=1001635 .

Given that, we're going to abandon this approach and stick with volume mounts.

PR #6874, opened 2025-01-23
OBS-427: Support verify_certs with self-issued certs for elastic cloud

relud:
This only impacts https urls, which right now we only use for elastic cloud.

In stage we are currently depending on `verify_certs=False` to connect to elastic cloud, and this will allow us to instead use a `certs.pem` provided by elastic cloud to verify their https certs.

### DIFF
--- a/socorro/external/es/connection_context.py
+++ b/socorro/external/es/connection_context.py
@@ -18,14 +18,17 @@ class ConnectionContext:
         self,
         url="http://localhost:9200",
         timeout=30,
+        ca_certs=None,
         **kwargs,
     ):
         """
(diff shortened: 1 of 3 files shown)


smarnach (approved):
The code changes are looking good – it's basically just a new configuration setting that's passed through several layers of indirection.

I wonder whether there's a reasonable way to enable SSL in the local stack for testing purposes, and whether that's a useful thing to do.

relud:
> I wonder whether there's a reasonable way to enable SSL in the local stack for testing purposes, and whether that's a useful thing to do.

i could do it, but i'm inclined to think it's not that helpful.

PR #6875, opened 2025-01-27
Bug 1943919 - Add TppRaiseInvalidParameter to the prefix list

gabrielesvelto:
We don't want to add this function to the irrelevant list because it conveys an important bit of information (that we're crashing because an invalid paramter has been passed to a Windows system function), but we still want to skip over it in order to be able to tell different crashes apart.


gabrielesvelto:
Yes, that represents the bulk of the crashes under that signature, but there's more and I've already preemptively filed the bugs for the other signatures.

smarnach (approved):

PR #6882, opened 2025-01-31
bug-1945553: Fix json encoding error with empty aggregations object

relud:
for es8+ search results may contain an AggResponse object under the `aggregations` key, instead a missing attribute like in es1.4, so here we convert that empty object to a dict

### DIFF
--- a/socorro/external/es/supersearch.py
+++ b/socorro/external/es/supersearch.py
@@ -8,7 +8,7 @@ import datetime
 import re
 
 from elasticsearch.exceptions import NotFoundError, BadRequestError
-from elasticsearch_dsl import A, Q, Search
+from elasticsearch_dsl import A, Q, Search, AggResponse
 
 from socorro.external.es.base import generate_list_of_indexes
 from socorro.external.es.search_common import SearchBase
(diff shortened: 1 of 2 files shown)


biancadanforth:
@relud : I filed a bug for this issue, bug-1945553. Can you update your PR/ticket title before merging and attach the PR to the bug when you get a chance?

socorro/external/es/supersearch.py
```diff
@@ -428,7 +428,7 @@ def get(self, **kwargs):
                 total = search.count()
 
                 aggregations = getattr(results, "aggregations", {})
-                if aggregations:
+                if isinstance(aggregations, AggResponse):
                     aggregations = self.format_aggregations(aggregations)
```
willkg, line 432: With ES 8, the results either:

1. does not have an `aggregations` attribute at all in which case the `aggregations` variable points to a `{}` value
2. does have an `aggregations` attribute which points to a `AggResponse` object in which case, that gets run through `format_aggregations()` which returns a dict

That seems good to me.

`aggregations` points to a value with two very different types. We're not doing type checking, so it's just mildly not great. If we were to switch to type checking, then this would be more obviously not great.

willkg (approved):

PR #6890, opened 2025-02-03
bug-1945556: omit example data when elasticsearch would error

relud:
and enable full version of all text fields so they still have example data under elasticsearch 8

context: the page `/documentation/datadictionary/dataset/{dataset}/field/{field}` for dataset=processed currently throws a 500 error for text fields with no full version and no fielddata

this only changes the mapping to include multi-fields where there was none before, which is a valid mapping update, so it can theoretically be applied in-place in stage so we don't need a backfill


webapp/crashstats/documentation/views.py
```diff
@@ -236,7 +236,13 @@ def generate_field_doc(dataset, field):
             search_field_query_type = super_search_field["query_type"]
             # FIXME(willkg): we're only doing this for public fields, but we could do
             # this for whatever fields the user can see
-            if field_data["permissions"] == ["public"]:
+            if field_data["permissions"] == ["public"] and (
+                # text fields only get example data if they have a full version or
+                # include fielddata, like for semicolon_keywords fields.
+                super_search_field["has_full_version"]
+                or super_search_field["storage_mapping"]["type"] != "text"
```
relud, line 243: this will also work for es 1.4 because there the type is "string" and does support aggregations

socorro/external/es/super_search_fields.py
```diff
@@ -1539,80 +1228,19 @@ def apply_schema_properties(fields, schema):
...
-        "is_returned": True,
-        "name": "moz_crash_reason_raw",
-        "namespace": "processed_crash",
-        "query_type": "string",
-        "storage_mapping": {
-            "fields": {"full": {"type": "keyword"}},
-            "type": "text",
-        },
-    },
+    "minidump_sha256_hash": text_field("minidump_sha256_hash"),
```
relud, line 1231: should this be a keyword field?

webapp/crashstats/documentation/jinja2/docs/datadictionary/field_doc.html
```diff
@@ -78,16 +78,15 @@ <h1>{{ field_name }}</h1>
                     {% for item in example_data %}
                       <div class="field-example"><code>{{ item|string|truncate(100, killwords=True, end="...") }}</code></div>
                     {% endfor %}
+                  {% elif "protected" in permissions %}
```
willkg, line 81: Weird. I wonder why I didn't do an elif before.

webapp/crashstats/documentation/jinja2/docs/datadictionary/field_doc.html
```diff
@@ -78,16 +78,15 @@ <h1>{{ field_name }}</h1>
...
-                        No examples available. Either this field is new and has
-                        no data or this field is rarely emitted and has no data
-                        this week.
-                      </i>
-                    {% endif %}
+                    <i>
+                      No examples available. Either this field is new and has
+                      no data or this field is rarely emitted and has no data
+                      this week or this field is not configured to allow it to
+                      be used for aggregations in elasticsearch.
```
willkg, line 88:
```suggestion
                      be used for aggregations in Elasticsearch.
```

willkg (approved):
I computed FIELDS in the main branch and in this branch and manually compared the two structures and everything looks good.

I had one cosmetic suggestion about Elasticsearch and that's it.

PR #6893, opened 2025-02-07
OBS-384: upgrade ESLint to v8 

toufali:
In order to support more modern ECMAScript required for the front-end refactor, we need to upgrade ESLint, which currently blocks merging of the new FE code.

Unfortunately, ESLint 9 dropped support for Node < v18.18 – we’re currently on Node v14.  ESLint 8 still supports Node v14. 

For additional context, see ticket: https://mozilla-hub.atlassian.net/browse/OBS-478

- [x] Upgrade ESLint to 8
- [x] Migrate ESLint config file to compatible format
- [x] Upgrade ESLint related Prettier packages


willkg:
@toufali Can you adjust the commit messages to match our conventions?: https://socorro.readthedocs.io/en/latest/dev.html#git-conventions

In this case, the issues are obs issues, so it should be something like this:

```
obs-384: upgrade ESLint to v8
```

and:

```
obs-384: upgrade Prettier
```

webapp/package.json
```diff
@@ -10,11 +10,11 @@
...
     "cssmin": "0.4.3",
-    "eslint": "7.13.0",
-    "eslint-config-prettier": "6.15.0",
-    "eslint-plugin-prettier": "3.1.4",
+    "eslint": "^8.57.1",
+    "eslint-config-prettier": "^10.0.1",
+    "eslint-plugin-prettier": "^5.2.3",
     "graceful-fs": "4.2.4",
-    "prettier": "2.1.2",
+    "prettier": "^3.5.0",
```
willkg, line 17: Why define a minimum rather than pinning to a specific version?

toufali: @willkg when it's not a critical dependency I often use a minimum in case of minor patches.  But I'm happy to pin it if you prefer!

toufali:
> @toufali Can you adjust the commit messages to match our conventions?: https://socorro.readthedocs.io/en/latest/dev.html#git-conventions
> 
> In this case, the issues are obs issues, so it should be something like this:
> 
> ```
> obs-384: upgrade ESLint to v8
> ```
> 
> and:
> 
> ```
> obs-384: upgrade Prettier
> ```

Sorry @willkg – I assumed since they get squashed anyway that it didn't really matter, but I will fix these and try to remember for next time!

willkg:
@toufali : We landed the PR that removes legacy ES from Socorro. Can you rebase this against main tip? Then I can review it.

toufali:
@willkg – rebase done 👍

webapp/package.json
willkg: We update this stuff rarely, so I'm wondering whether it's better to pin it and then re-enable dependabot to update things. That way it doesn't change without us approving it.

willkg (approved):
Looks good! The only change I'd like to make is for us to pin the dependencies in `package.json` to specific versions. Let's make that change and then we can land this.

PR #6894, opened 2025-02-07
bug-1946895: drop NaN and +/-inf when loading es8

relud:


socorro/external/es/crashstorage.py
```diff
@@ -243,6 +265,11 @@ def build_document(src, crash_document, fields, all_keys):
             if value is None:
                 continue
 
+        elif storage_type == "float":
+            value = fix_float(value)
+            if value is None:
+                continue
```
relud, line 271: at present this only impacts `uptime_ts`

willkg: Mmm... I think inf, -inf, and nan are junk data and should be dropped from the processed crash altogether. I think I'd adjust the `CopyFromRawDataRule` rule to drop these cases and add a processor note:

https://github.com/mozilla-services/socorro/blob/af68b249eee56ff76f6caf0e2aa76510066e45ef/socorro/processor/rules/mozilla.py#L153-L158

Does dropping it here just for Elasticsearch indexing work better in some way?

relud: this isn't better, it's just the thing I immediately knew how to do, i'll go fix it there

willkg: Maybe we want to do it in both places? This change to Elasticsearch enforces Elasticsearch's indexing requirements which is important and may change over time while adding a processor rule reduces junk in the processed crash data.

relud: oh, that's a good point. yeah i'd like to do it in both places.

i'll file a separate pr for the processor rule, while this can ensure we don't have an issue with the es8 migration.

willkg (approved):
Looks good!

PR #6904, opened 2025-02-10
bug-1946895: Fix detecting float/double field when building es8 documents

relud:
renames `fix_float` to `fix_double` because that's the actual es storage type name, and added cases to `test_indexing_bad_data` to show it's actually working this time

This also applies the fix to legacy_es, since the bug also appeared there and that was impacting manual testing on the processor with the problem crash.

### DIFF
--- a/socorro/external/es/crashstorage.py
+++ b/socorro/external/es/crashstorage.py
@@ -191,8 +191,8 @@ def fix_long(value):
     return value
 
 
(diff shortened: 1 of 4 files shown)


willkg (approved):
I verified this fixes processing for efcb32d9-86d3-437f-961d-1e6e90241218.

PR #6917, opened 2025-02-27
obs-512: fix webapp/static and webapp/node_modules directory ownership

willkg:
This ensures these two directories are owned by the host user rather than root during image building and container running.

To test:

1. delete `webapp/static/` and `webapp/node_modules/` if they exist
2. run `just clean`
3. run `just build`
4. ensure `webapp/static/` and `webapp/node_modules/` are owned by the host user and not root (this is only a problem on Linux)
5. run `just build` to make sure it builds even if the directories exist

### DIFF
--- a/justfile
+++ b/justfile
@@ -14,6 +14,9 @@ _env:
 
 # Build docker images
 build *args='app fakesentry oidcprovider elasticsearch postgresql pubsub memcached gcs-emulator': _env
+    # Create these two directories using the current user so they don't end up
+    # created by Docker with the wrong ownership; ignore errors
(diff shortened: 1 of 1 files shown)


toufali (approved):
I ran through test steps successfully.  

Before this change, `static` and `node_modules` were created via `just run` (docker).  With this change, the two folders are created by `just build` as expected with correct ownership.  

Running `just build` a second time results in 2 `File exists` errors, but the build ignores the errors and continues successfully. 

Caveat: I'm not on Linux

willkg:
Thank you!

PR #6944, opened 2025-03-21
bug-1945653: adjust processing queue balance

willkg:
This changes the way the processor pulls from queues. Previously, it pulled pull_max_messages from each queue in order and then processed that. If we had a backlog of standard and reprocessing queue work, then they'd be handled equally. That's bad for standard queue processing and caused alerts and stuff.

This changes it to prefer processing the standard queue 5 over the reprocessing queue 5 to 1 and makes it much less likely the standard queue backs up when someone reprocesses a bunch of crash reports.

When this hits stage, I can test it more thoroughly by reprocessing a bunch of crash reports. It may be we want something bigger than 5 to 1. We can tune that going forward.

### DIFF
--- a/socorro/tests/external/pubsub/test_crashqueue.py
+++ b/socorro/tests/external/pubsub/test_crashqueue.py
@@ -52,6 +52,53 @@ class TestPubSubCrashQueue:
             (reprocessing_crash,),
         }
 
+    def test_pull_max(self, pubsub_helper):
+        standard_crashids = []
+        for _ in range(10):
+            crashid = "000" + create_new_ooid()[3:]
+            standard_crashids.append(crashid)
+
+            pubsub_helper.publish("standard", crashid)
+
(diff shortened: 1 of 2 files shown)


socorro/tests/external/pubsub/test_crashqueue.py
```diff
@@ -52,6 +52,53 @@ def test_iter(self, pubsub_helper):
...
+            "000",
+            "000",
+            "000",
+            "111",
+            "111",
+            "111",
+            "111",
+        ]
+
+        assert list(sorted(new_crashes)) == list(sorted(standard_crashids)) + list(
```
willkg, line 98: I'm not thrilled with how ruff reformatted this. It's not a big deal, but it looks stupid.

relud (approved):

willkg:
Thank you!

PR #7037, opened 2025-07-03
Group dependabot pull requests

alexcottner:
Grouping dependabot pull requests to make it a little less noisy.


say-yawn (changes requested):
Learned about `groups` ([doc](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--)). Minor change suggested.

smarnach (approved):
Thanks, this looks good to me! Let's try it for a month, and if we like the result, we'll change the other dependabot configs as well.

PR #7043, opened 2025-07-21
Bug 1978426 - Add std::unique_ptr to the irrelevant signature list.

amccreight:
Also, remove a UniquePtr thing from the prefix list, as UniquePtr is in the irrelevant signature list already.


smarnach (approved):
Looks good to me. Here's an example signature change:
```
Crash id: d6a940bc-e9c7-43a2-88f3-0f07f0250722
Original: std::unique_ptr<T>::operator->
New:      mozilla::dom::NavigationHistoryEntry::Key
Same?:    False
```

PR #7062, opened 2025-09-03
Bug 1986656 - Skip over MakeUnique() in signature generation even when it doesn't have a specific type

gabrielesvelto:


gabrielesvelto:
Yes, this is what I expected.

smarnach (approved):

PR #7065, opened 2025-09-11
CRINGE-79: automatically create a Crash Stats superuser when starting services locally

biancadanforth:
This is a follow-up ticket from #7064 .

Because:
* While there are docs for creating a Crash Stats superuser in local development, it's a common enough task that it'd be easier to create one automatically when starting up Socorro.

This commit:
* Moves the commands into a new shell script, `bin/create_superuser.sh`.
* Includes running this new script as part of `just setup`.
* Unrelated: the PubSub emulator kept failing its `waitfor` check for me locally, so I increased the timeout, as was done in the past for the Elasticsearch emulator.

Here's a screenshot of the docs section that explains the change:
Screenshot 2025-09-11 at 12 05 20 PM

### DIFF
--- a/bin/create_superuser.sh
+++ b/bin/create_superuser.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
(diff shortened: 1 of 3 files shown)


smarnach (approved):
Looks good to me!

When I suggested to add this to setup_services.sh, I didn't realize we need to run this outside of Docker so we can run commands in two different containers. It makes sense to have this in a separate script.

PR #7066, opened 2025-09-15
Bug 1959806 - Use Firefox for Android as the product

amccreight:
This product has been renamed from "Fenix" to "Firefox for Android".

### DIFF
--- a/product_details/Fenix.json
+++ b/product_details/Fenix.json
@@ -14,8 +14,8 @@
   "in_buildhub": false,
   "bug_links": [
     [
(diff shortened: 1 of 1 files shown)


biancadanforth (approved):
This change is limited to the Bugzilla tab for a given Fenix crash report to fix a broken Bugzilla link on this tab. I've verified this locally, so it looks good to me.

Thank you for bearing with us. We've only had one engineer on Crash Ingestion for the past six months until recently (now we have two).

PR #7182, opened 2026-04-08
bug-1685143: automate revoke access

Haseeb702:
Because:
- We want to automatically off board users that have lost employment by removing them from the hackers group as well as removing their tokens. [bug-1685143](https://bugzilla.mozilla.org/show_bug.cgi?id=1685143)

This PR:
- Added a function `is_blocked_in_auth0` to send a query to Auth0 and check if a user is blocked. A Mozilla user is blocked if they have lost employment or have too many login attempts
- The `auditgroups` cronjob checks if a Mozilla user is blocked in auth0. If so, we remove them from the `hackers_group`

### DIFF
--- a/webapp/crashstats/authentication/management/commands/auditgroups.py
+++ b/webapp/crashstats/authentication/management/commands/auditgroups.py
@@ -92,14 +92,11 @@ class Command(BaseCommand):
             "--dry-run", action="store_true", help="Whether or not to do a dry run."
         )
 
-    def is_policy_exception(self, user):
+    def is_employee_or_exception(self, user):
         # If this user has a policy exception, then they're allowed
         if PolicyException.objects.filter(user=user).exists():
             return True
 
-        return False
-
-    def has_valid_mozilla_email(self, user):
         if user.email.endswith(VALID_EMAIL_DOMAINS):
             return True
 
@@ -128,15 +125,13 @@ class Command(BaseCommand):
                 self.stdout.write(f"Auth0 failed for: {user.email}: {e}")
 
             # User may be blocked as a security mitigation. Eg: too many login attempts
-            if is_blocked and self.has_valid_mozilla_email(user):
+            if is_blocked:
                 users_to_remove.append((user, "user has most likely lost employment"))
 
             elif not user.is_active:
                 users_to_remove.append((user, "!is_active"))
 
-            elif not self.is_policy_exception(
-                user
-            ) and not self.has_valid_mozilla_email(user):
+            elif not self.is_employee_or_exception(user):
                 users_to_remove.append((user, "not employee or exception"))
 
             elif user.last_login and user.last_login < cutoff:
--- a/webapp/crashstats/authentication/management/commands/auditgroups.py
+++ b/webapp/crashstats/authentication/management/commands/auditgroups.py
@@ -13,9 +13,11 @@ from django.contrib.admin.models import LogEntry, CHANGE
 from django.contrib.contenttypes.models import ContentType
 from django.core.management.base import BaseCommand
 from django.utils import timezone
+from django.conf import settings
 
 from crashstats.authentication.models import PolicyException
-
+from socorro.lib.librequests import session_with_retries
+from urllib.parse import urlparse
 
 VALID_EMAIL_DOMAINS = ("mozilla.com", "mozilla.org")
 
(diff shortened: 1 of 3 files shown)


webapp/crashstats/authentication/management/commands/auditgroups.py
```diff
@@ -38,6 +40,50 @@ def delta_days(since_datetime):
     return (timezone.now() - since_datetime).days
 
 
+def get_access_token(client_id, client_secret, domain, session):
```
biancadanforth, line 43: Note for posterity: `get_access_token`, `find_users`, and `is_blocked_auth0` are the same methods used in Tecken https://github.com/mozilla-services/tecken/blob/d742f91d260f4b3e3a6dc8b37e1916aef585effd/tecken/useradmin/middleware.py

We don't have a good way right now to share application code between repos (obs-common is for dev-only stuff). Since this is only 40-ish lines of code, it's not worth doing anything further now.

webapp/crashstats/authentication/management/commands/auditgroups.py
```diff
@@ -46,11 +92,14 @@ def add_arguments(self, parser):
             "--dry-run", action="store_true", help="Whether or not to do a dry run."
         )
 
-    def is_employee_or_exception(self, user):
```
biancadanforth, line 49: During our 1:1 last week, we talked about breaking up this method, but I didn't fully understand at the time why these two checks were paired together. Basically, if someone is NOT a MoCo employee, they won't have a `mozilla.com` email address. Only users without a `mozilla.com` email address are listed in `PolicyExceptions`. So if a user neither has a MoCo email address nor is a policy exception, then we don't want them in the Hackers group. Consequently, can you revert the change to this method? Sorry for the confusion.

webapp/crashstats/authentication/management/commands/auditgroups.py
```diff
@@ -72,10 +121,22 @@ def audit_hackers_group(self, dryrun=True):
...
         for user in hackers_group.user_set.all():
-            if not user.is_active:
+            is_blocked = False
+            try:
+                is_blocked = is_blocked_in_auth0(user.email)
+            except RuntimeError as e:
+                self.stdout.write(f"Auth0 failed for: {user.email}: {e}")
+
+            # User may be blocked as a security mitigation. Eg: too many login attempts
+            if is_blocked and self.has_valid_mozilla_email(user):
```
biancadanforth: Since you tested the Auth0 API and were able to see your @gmail.com email address in Auth0, we can simplify the logic and basically block any user (regardless of their email address' domain) that is blocked in Auth0.

The risk is that we remove someone from the Hackers group who has a temporary Auth0 block (e.g. too many login attempts). But this is a small enough group of people in the first place and a rare enough occurrence that we can deal with it manually if/when it comes up and consider handling this differently at that time.

So this can just check for `is_blocked`.

webapp/crashstats/authentication/tests/test_auditgroups.py
```diff
@@ -104,27 +128,102 @@ def test_old_user_with_active_api_tokens_is_not_removed(self, db):
...
+        )
+
+        assert hackers_group.user_set.count() == 1
+
+        buffer = StringIO()
+        call_command("auditgroups", dry_run=False, stdout=buffer)
+        assert hackers_group.user_set.count() == 1
+        assert "Removing:" not in buffer.getvalue()
+
+    def test_blocked_user_with_invalid_email_and_no_exception(self, db, monkeypatch):
```
biancadanforth: Nit: Can you follow a similar test case naming convention to the test cases above? I know the name is getting long now:

`test_user_blocked_in_auth0_without_policy_exception_with_invalid_email_is_removed`?

biancadanforth (changes requested):
This is looking pretty good! I reviewed this together with @smarnach , so consider this a review by both of us.

In addition to the change requests indicated inline in this review, there are a couple of top-level changes/things to check before this PR can be merged:
* We need to increase the frequency with which this `auditgroups` cron job is run. It's currently weekly, and it should be daily. That configuration is in this repo.
* For your tests, you used Tecken's stage secrets to make a request to the Auth0 API. We need to know if Socorro's stage secrets for the Auth0 API have sufficient permissions for this Users API. Otherwise, we need to get new secrets with these permissions and update the secrets in GSM before merging this PR.
  * To test: 1) Access the Users API with Tecken stage secrets (you've already done this). 2) If successful, access the API with Socorro stage secrets. 3) If this fails, we know we need additional permissions.

webapp/crashstats/authentication/tests/test_auditgroups.py
Haseeb702: I decided to remove this test case because the edge case that occurred after splitting the `is_employee_or_exception` function doesn't exist anymore since we added it back.

Haseeb702 (commented):
I made the relevant changes according to the feedback. When I tried to run the Auth0 query with the socorro stage credentials I received a 403 error which means that we don't have the required permissions.

webapp/crashstats/authentication/management/commands/auditgroups.py
```diff
@@ -128,15 +125,13 @@ def audit_hackers_group(self, dryrun=True):
                 self.stdout.write(f"Auth0 failed for: {user.email}: {e}")
 
             # User may be blocked as a security mitigation. Eg: too many login attempts
-            if is_blocked and self.has_valid_mozilla_email(user):
+            if is_blocked:
```
biancadanforth, line 128: Not required but we could do this check last, since it's more expensive than the other checks.

All of these conditions are, by themselves, sufficient reason to remove the user from the Hackers group. Ideally, we would avoid checking conditions further down if they meet an earlier condition. I think that's why the `if/elif` blocks are set up as they are. So you could swap `is_blocked` with a condition further down and leave the logic otherwise unmodified.

biancadanforth (approved):
This looks good. However, we should hold off merging until we update the Auth0 secrets for stage (and ultimately prod).

Haseeb702 (commented):
To access the API endpoint for Auth0 to check for the `blocked` field, we had to request for upgraded Auth0 permissions for Socorro. Details are provided in [IAM-1952](https://mozilla-hub.atlassian.net/browse/IAM-1952?atlOrigin=eyJpIjoiN2IxNzQ3N2MwMDMyNGMzMDhjZDNkZTU0Y2RhOTFmNDAiLCJwIjoiaiJ9).

PR #7186, opened 2026-04-20
bug-2031036: Failure to remove arguments in signature with "(anonymous namespace)"

Haseeb702:
Because:
- [bug-2031036](https://bugzilla.mozilla.org/show_bug.cgi?id=2031036)

This PR:
- replaces the `(anonymous namespace)` with a placeholder value and removes it from the exceptions, thus allowing the argument with the `(anonymous namespace)` to be removed

Example:
-  Initial signature: `mozilla::SpinEventLoopUntil(nsTSubstring<T> const&, (anonymous namespace)::ParentImpl::ShutdownBackgroundThread::<T>&&, nsIThread*)`
- Post fix signature: `mozilla::SpinEventLoopUntil`

### DIFF
--- a/socorro/signature/rules.py
+++ b/socorro/signature/rules.py
@@ -175,20 +175,23 @@ class CSignatureTool:
             exceptions=("name omitted", "IPC::ParamTraits", " in "),
         )
 
+        def cpp_argument_exceptions(exceptions, before_token, after_token, token):
+            exceptions = ["(anonymous namespace)", "operator"]
+            for s in exceptions:
+                if before_token.endswith(s):
+                    return True
+                if s == token:
+                    return True
+            return False
+
         # Collapse arguments
         if self.collapse_arguments:
-            function = function.replace(
-                "(anonymous namespace)", "anonymous_namespace_placeholder"
-            )
             function = collapse(
                 function,
                 open_string="(",
                 close_string=")",
                 replacement="",
-                exceptions=("operator",),
-            )
-            function = function.replace(
-                "anonymous_namespace_placeholder", "(anonymous namespace)"
+                is_exception=cpp_argument_exceptions,
             )
 
         # Remove PGO cold block labels like "[clone .cold.222]". bug #1397926
--- a/socorro/signature/rules.py
+++ b/socorro/signature/rules.py
@@ -177,13 +177,15 @@ class CSignatureTool:
 
         # Collapse arguments
         if self.collapse_arguments:
+            function = function.replace("(anonymous namespace)", "placeholder")
             function = collapse(
                 function,
                 open_string="(",
                 close_string=")",
                 replacement="",
-                exceptions=("anonymous namespace", "operator"),
+                exceptions=("operator",),
             )
+            function = function.replace("placeholder", "(anonymous namespace)")
 
         # Remove PGO cold block labels like "[clone .cold.222]". bug #1397926
         if "clone .cold" in function:
--- a/socorro/signature/rules.py
+++ b/socorro/signature/rules.py
@@ -12,7 +12,7 @@ from glom import glom
 
 from .siglists_utils import get_signature_list_content
 from .utils import (
-    collapse,
+    replace_enclosed_slices,
     drop_bad_characters,
     drop_prefix_and_return_type,
     get_crashing_thread,
@@ -115,18 +115,20 @@ class CSignatureTool:
         function = drop_prefix_and_return_type(function)
 
         # Collapse types
-        function = collapse(
+        function = replace_enclosed_slices(
             function,
-            open_string="<",
-            close_string=">",
-            replacement="<T>",
-            exceptions=(" as ",),
+            opening_token="<",
+            closing_token=">",
+            replace=lambda before, inside, after: inside if " as " in inside else "<T>",
         )
 
         # Collapse arguments
         if self.collapse_arguments:
-            function = collapse(
-                function, open_string="(", close_string=")", replacement=""
+            function = replace_enclosed_slices(
+                function,
+                opening_token="(",
+                closing_token=")",
+                replace=lambda before, inside, after: "",
             )
 
         if self.signatures_with_line_numbers_re.match(function):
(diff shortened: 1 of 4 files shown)


socorro/signature/rules.py
```diff
@@ -177,12 +177,18 @@ def normalize_cpp_function(self, function, line):
 
         # Collapse arguments
         if self.collapse_arguments:
+            function = function.replace(
```
Haseeb702, line 180: Here I replace the "(anonymous namespace)" with a placeholder and remove it from the exceptions so that it's removed if it appears in an argument. In addition, when the signature is leading with "(anonymous namespace)" it is not removed since we replace it with a placeholder value. After the `collapse()` is done, we replace the placeholder with the "(anonymous namespace)" to get the correct signature.

smarnach (commented):
Thanks for this pull request! This is a clever way of avoiding the problem.

I still have some concerns about the approach. The current implementation of `collapse()` is a bit confusing, and the meaning of the `exceptions` parameter to that function is ambiguous. Your fix simply side-steps these problems by adding yet another special case to the code.

In the long run, this makes a codebase almost impossible to reason about. Part of the codebase are already quite difficult to understand. We should aim to improve that situation, and leave every bit of code that we should behind in a slightly better shape than it was in before.

Here is an idea what you could do instead. Instead of accepting a list of strings as `exceptions`, the `collapse()` function could accept a function `is_exception` that receives the string before a section that would be replaced and the section that would be replaced itself as parameters and returns a Boolean value indicating whether this is an exception. This would allow simplifying the body of `collapse()`, and also make the meaning of the parameter much clearer than it is now.

There are certainly better approaches than this suggestion I just made up without putting much thought into it. My main point is – let's try and make this code better while we are here, rather than just fixing the bug.

In some situations, "proper" fixes are not really possible. They may be unreasonably complex, or you simply may not have time for them, and then you have to settle for a hack and move on. However, I think that in this particular case there is a way to simplify the code while fixing the bug that can be implemented with reasonable effort.

Haseeb702 (commented):
Another solution I tried was to edit the `_is_exceptions()` function in `socorro/signature/utils.py` by changing line 203 to `if token == s:` so that it checks the exact string and doesn't ignore the `(anonymous namespace)` that is embedded in the arguments. This worked for cpp normalization but failed at the rust because of the " as " exception. Now i'm going to look into the solution that @smarnach recommended.

biancadanforth (commented):
~~Since I was answering a question in Slack, and I don't see it mentioned in Sven's review: The issue is that `collapse` considers it an exception in both the case that 1) the exception string occurs before the start token or 2) somewhere in between the start and end token. The reason why `"anonymous namespace"` was included as an exception in the first place was because we do want to keep it in the function string if it occurs in case 1. The bug reported here is that we don't want to keep it if it's case 2.~~

~~So, the patch as currently written breaks the intended behavior to leave (anonymous namespace) in tact for case 1. We can check with amccreight (the reporter in the bug) to confirm that's still desired, but it so, then whatever solution we go with should not regress that.~~

Edit: Nevermind, your solution doesn't break that (as the tests indicate).

socorro/signature/rules.py
```diff
@@ -167,28 +169,50 @@ def normalize_cpp_function(self, function, line):
...
-            close_string=">",
-            replacement="<T>",
-            exceptions=("name omitted", "IPC::ParamTraits", " in "),
+        def get_type_replacement(before, inside, after):
+            exceptions = ["name omitted", "IPC::ParamTraits", " as "]
+            for s in exceptions:
+                if before.endswith(s):
+                    s_without_outer_tokens = inside[1:-1]
+
+                    inside_substring = replace_enclosed_slices(
```
smarnach: I don't understand this recursive call to `replace_enclosed_slices()`. The old code didn't need any recursion; why is it needed in the new version?

socorro/signature/rules.py
```diff
@@ -167,28 +169,50 @@ def normalize_cpp_function(self, function, line):
...
         # NOTE(willkg): The " in " is for handling "<unknown in foobar.dll>". bug
         # #1685178
-        function = collapse(
-            function,
-            open_string="<",
-            close_string=">",
-            replacement="<T>",
-            exceptions=("name omitted", "IPC::ParamTraits", " in "),
+        def get_type_replacement(before, inside, after):
+            exceptions = ["name omitted", "IPC::ParamTraits", " as "]
```
smarnach: You don't really need to check for all these strings in all the places.

* "name omitted" only applies if it matches the entirety of `inside`.
* `IPC::ParamTraits` only applies if `before` ends with the string.
* " as " applies when it appears somewhere in `inside`.

So maybe you could write this body in a much more specific way:
```
if (inside == "<name omitted>" or
    before.endswith("IPC::ParamTraits") or
    " as " in inside):
    return inside
return "<T>"
```

socorro/signature/rules.py
```diff
@@ -167,28 +169,50 @@ def normalize_cpp_function(self, function, line):
...
-        if self.collapse_arguments:
-            function = collapse(
-                function,
-                open_string="(",
-                close_string=")",
-                replacement="",
-                exceptions=("anonymous namespace", "operator"),
-            )
+        def get_argument_replacement(before, inside, after):
+            exceptions = ["(anonymous namespace)", "operator"]
```
smarnach: Same here:
```
if inside == "(anonymous namespace)" or before.endswith("operator"):
    return inside
return ""

File: socorro/signature/rules.py [hunk 5]
```diff
@@ -167,28 +169,50 @@ def normalize_cpp_function(self, function, line):
...
+            for s in exceptions:
+                if before.endswith(s):
+                    return inside
+                if s in inside and not (s.startswith("(") and s.endswith(")")):
```
smarnach: What is this exception about?

Haseeb702:
What changes I made after Sven's review:
- Simplified the replacement function by directly referencing the exceptions in the if statements instead of iterating through an exceptions list
- Added test cases for the `find_enclosed_slices` function

Question: Will we need to reprocess the crashes that have the `(anonymous namespace)` signature?

smarnach:
> Question: Will we need to reprocess the crashes that have the `(anonymous namespace)` signature?

After this lands in prod, we should consider reprocessing any crashes whose signatures we change. It's not a strict requirement. If we don't, new crashes won't be grouped together with old crashes that would have the same signature now. The situation would sort itself out in six months by itself, though, so it's not the end of the world.

socorro/signature/tests/test_utils.py
```diff
@@ -146,15 +182,61 @@ def test_parse_source_file(source_file, expected):
         ),
     ],
 )
-def test_collapse(function, expected):
+def test_replace_enclosed_slices_types(function, expected):
+    def get_type_replacement(before, inside, after):
```
smarnach: This copies the local function from the actual code, so it's likely this will get out of sync. I suggest refactoring this by wrapping `replace_enclosed_slices()` for each use case, and add these wrappers to utils.py as well. This particular function could be wrapped like this:

```
def collapse_types(func_signature: str) -> str:
    def get_type_replacement(before, inside, after):
        if inside == "<name omitted>" or " as " in inside:
            return inside
        if before.endswith("IPC::ParamTraits"):
            s_without_outer_tokens = inside[1:-1]
            inside_substring = replace_enclosed_slices(
                s_without_outer_tokens, "<", ">", get_type_replacement
            )
            return f"<{inside_substring}>"
        return "<T>"
    return replace_enclosed_slices(
        func_signature,
        opening_token="<",
        closing_token=">",
        replace=get_type_replacement
    )
```

This way, both the code in rules.py and the test code here only need to call `collapse_types()`, and it's guaranteed the code you are testing actually matches the code that's used.

smarnach (commented):
Thanks for the changes! We are getting close here, but I've got one more request for you. :)

socorro/signature/tests/test_utils.py
Haseeb702: Ah I see. So when we wrap the function, in the future if the exceptions or the function is changed in any way, it would reflect those changes onto the tests. I'll make these changes. Thanks for pointing this out!

socorro/signature/utils.py
```diff
@@ -184,103 +185,110 @@ def parse_source_file(source_file):
...
 
-        <rayon_core::job::HeapJob<BODY> as rayon_core::job::Job>::execute
-        ^                              ^^^^ exception string inside token
-        open token
+    def get_type_replacement(before, inside, after):
+        if inside == "<name omitted>" or " as " in inside:
+            return inside
+        if before.endswith("IPC::ParamTraits"):
+            s_without_outer_tokens = inside[1:-1]
+            inside_substring = replace_enclosed_slices(
```
smarnach: This looks like it's actually a call to `collapse_types(s_without_outer_tokens)`.

smarnach (approved):
Thanks, this looks good to me! I added one more comment to consider, but feel free to merge this whether you decide to address it or not.

socorro/signature/utils.py
Haseeb702: Oh thanks for catching this! I'm going to fix this and then merge.

PR #7191, opened 2026-05-01
bug-2007193: dedupe json_dump.thread_count in processed crash schema

biancadanforth:
Because:
* I accidentally added `thread_count` underneath `json_dump` when it was already present. This is a follow-up fix for #7087 .
* I also omitted the permissions on the duplicate entry. Since this entry was overwritten, the schema tests didn't catch it the first time around.

This PR:
* Removes the duplicate (lesser) entry. By lesser, I mean its description was less helpful.
* Adds a new yaml linting step in bin/lint.sh to check for duplicate fields among other things.

More in-depth explanation of what is happening:
* The YAML loader we're using, `pyyaml`, silently overwrites any duplicate keys during the loading process, so the one I'm removing is the one that was previously used in the schema.
  * Minimal example:
```py
import yaml
test_yaml = '''
foo: 1
bar: 2
foo: 3
'''
result = yaml.load(test_yaml, Loader = yaml.Loader)
print(result) # {'foo': 3, 'bar': 2}
```
* In this case, since everything was identical except the description, there's no real consequence other than less helpful documentation with the less helpful description.
* In the worst case scenario, if there's a later entry at the same nested level in the schema and it's not identical, this could cause more significant issues. E.g. if the later entry lists a field as public, and the earlier entry lists it as private, this would result in the field being made public.

### DIFF
--- a/.yamllint
+++ b/.yamllint
@@ -1,0 +1,4 @@
+---
+extends: default
+rules:
+  key-duplicates: enable
--- a/bin/lint.sh
+++ b/bin/lint.sh
@@ -44,6 +44,9 @@ else
         license-check .
     fi
 
+    echo ">>> yamllint"
+    yamllint socorro/schemas/
+
     echo ">>> eslint (js)"
     cd /app/webapp
     /app/webapp/node_modules/.bin/eslint .
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,6 +64,7 @@ dependencies = [
     "urlwait",
     "werkzeug",
     "whitenoise",
+    "yamllint",
 ]
 
 [dependency-groups]
--- a/socorro/schemas/processed_crash.schema.yaml
+++ b/socorro/schemas/processed_crash.schema.yaml
@@ -583,6 +583,7 @@ definitions:
         description: >
           Number of threads in the crashing process at the time of the crash.
         type: ["integer", "null"]
+        permissions: ["public"]
       unloaded_modules:
         description: |
           Modules no longer mapped into the process' address space when the
@@ -628,10 +629,6 @@ definitions:
           permissions: ["public"]
           type: object
         type: ["array", "null"]
-      thread_count:
-        description: How many threads there are.
-        type: ["integer", "null"]
-        permissions: ["public"]
       threads:
         items:
           description: Information on a thread.
--- a/socorro/schemas/raw_crash.schema.yaml
+++ b/socorro/schemas/raw_crash.schema.yaml
@@ -733,7 +733,7 @@ properties:
       Set to 1 when Marionette (WebDriver Classic) is enabled.
     type: string
     permissions: ["public"]
-  
+
   ModuleSignatureInfo:
     data_reviews:
       - unknown
@@ -923,7 +923,7 @@ properties:
       Set to 1 when the Remote Agent (WebDriver BiDi) is enabled.
     type: string
     permissions: ["public"]
-  
+
   RemoteType:
     data_reviews:
       - unknown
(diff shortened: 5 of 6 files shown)


socorro/tests/schemas/test_socorro_data_schemas.py
```diff
@@ -13,13 +15,39 @@
...
 )
-from socorro.schemas import get_file_content
+from socorro.schemas import get_file_content, __file__ as socorro_schemas_file
 from socorro.schemas.validate_processed_crash import (
     validate_and_test as processed_validate,
 )
 from socorro.schemas.validate_raw_crash import validate_and_test as raw_validate
 
 
+SCHEMAS_DIR = Path(socorro_schemas_file).parent
```
smarnach: So far we are relying on `get_file_content()` to retrieve the schema files, either directly or indirectly via `get_schema()`. I understand that this will return the parsed file, so it won't help for the new tests, but I still think we should stay consistent with the previous approach and add a function `get_raw_file_content()`, or maybe a `raw` flag to the existing `get_file_content()` function.

socorro/tests/schemas/test_socorro_data_schemas.py
```diff
@@ -13,13 +15,39 @@
...
+
+
+def load_yaml_asserting_no_duplicate_keys(schema_filename):
+    """Load a YAML schema file, raising an error if any mapping has duplicate keys."""
+
+    class UniqueKeyLoader(yaml.Loader):
+        def construct_mapping(self, node, deep=False):
```
smarnach: While I totally believe this does what it says it does, properly understanding this requires a rather deep understanding of the internals of the PyYAML parser, which I don't have. I guess we could just treat it as a black box. I can't tell how likely it is to break with future PyYAML releases.

How did you come up with this approach?

socorro/tests/schemas/test_socorro_data_schemas.py
```diff
@@ -13,13 +15,39 @@
...
+            if isinstance(node, yaml.MappingNode):
+                self.flatten_mapping(node)
+            pairs = self.construct_pairs(node, deep=deep)
+            keys = [pair[0] for pair in pairs]
+            duplicates = {key for key in keys if keys.count(key) > 1}
```
smarnach: This uses a quadratic time approach for finding duplicates. I suggest using `collections.Counter` instead to get amortized linear time. (It might not be relevant for the number of keys we have in our files, but the code doesn't get any more complex, so we might as well change it.)

smarnach (commented):
Thanks for the fix, and thanks for adding tests as well!

I don't fully understand the test code, so I'm wondering whether we'd want to adopt a YAML linter instead of implementing the check for duplicate keys ourselves. Alternatively, we could treat the test code as a black box and hope it will keep working across PyYAML releases, and just remove it again if it proves brittle.

I added a few comments, but I think it's best we talk about this first and decide what direction we'd want to go together.

biancadanforth:
> Thanks for the fix, and thanks for adding tests as well!
> 
> I don't fully understand the test code, so I'm wondering whether we'd want to adopt a YAML linter instead of implementing the check for duplicate keys ourselves. Alternatively, we could treat the test code as a black box and hope it will keep working across PyYAML releases, and just remove it again if it proves brittle.
> 
> I added a few comments, but I think it's best we talk about this first and decide what direction we'd want to go together.

Thanks for the review! I totally agree that this approach is not a good one; sorry for the confusion. We talked about this sync, and I looked into how involved it would be to add a yaml linting step that checks for duplicate keys among other things. It didn't seem that hard to add that here, so that's what I did instead.

smarnach (approved):

PR #7195, opened 2026-05-04
CRINGE-268: Upgrade all dependencies.

smarnach:
Dependabot was unable to resolve all package versions when upgrading some dependencies. This PR upgrades all dependencies to the latest possible versions using

    just uv lock --upgrade

https://mozilla-hub.atlassian.net/browse/CRINGE-268

### DIFF
--- a/webapp/crashstats/documentation/views.py
+++ b/webapp/crashstats/documentation/views.py
@@ -57,7 +57,7 @@ def read_whatsnew():
 
     with open(path, "r") as fp:
         data = fp.read()
-        parts = docutils.core.publish_parts(data, writer_name="html")
+        parts = docutils.core.publish_parts(data, writer="html")
 
     return parts["html_body"]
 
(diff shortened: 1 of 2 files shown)


webapp/crashstats/documentation/views.py
```diff
@@ -57,7 +57,7 @@ def read_whatsnew():
 
     with open(path, "r") as fp:
         data = fp.read()
-        parts = docutils.core.publish_parts(data, writer_name="html")
+        parts = docutils.core.publish_parts(data, writer="html")
```
leplatrem, line 60: https://www.docutils.org/docs/api/publisher.html#publish-parts

driveby?

leplatrem (approved):

webapp/crashstats/documentation/views.py
smarnach: No, they actually renamed this argument, so that's intentional. :)

PR #7197, opened 2026-05-06
bug-2032276: add Authorization header to socorro-submitter requests

biancadanforth:
This is the corresponding stage-submitter ticket for #mozilla-services/antenna#1338

### DIFF
--- a/socorro/mozilla_settings.py
+++ b/socorro/mozilla_settings.py
@@ -41,6 +41,7 @@ TOOL_ENV = _config(
 if TOOL_ENV:
     fake_values = [
         ("ELASTICSEARCH_URL", "http://elasticsearch:9200"),
+        ("BREAKPAD_STAGE_SUBMITTER_BEARER_TOKEN", "faketoken"),
     ]
     for key, val in fake_values:
         os.environ[key] = val
@@ -306,9 +307,8 @@ BETAVERSIONRULE_VERSION_STRING_API = _config(
 # Stage submitter configuration
 BREAKPAD_STAGE_SUBMITTER_BEARER_TOKEN = _config(
     "BREAKPAD_STAGE_SUBMITTER_BEARER_TOKEN",
-    default="",
     doc=(
-        "Stage submitter bearer auth token sent on every /submit HTTP POST"
+        "Stage submitter bearer auth token sent on every /submit HTTP POST "
         "request to Antenna."
     ),
 )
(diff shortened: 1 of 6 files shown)


socorro/tests/stage_submitter/test_submitter.py
```diff
@@ -547,6 +547,44 @@ def always_20(*args, **kwargs):
     metricsmock.assert_incr("socorro.submitter.ignore")
 
 
+def test_bearer_token(pubsub_helper, gcs_helper, mock_collector):
```
biancadanforth, line 550: Modeled after `test_user_agent` below.

socorro/stage_submitter/submitter.py
```diff
@@ -256,6 +256,9 @@ def __init__(self):
         self.logger = logging.getLogger(__name__ + "." + self.__class__.__name__)
 
     def set_up(self):
+        if not settings.BREAKPAD_STAGE_SUBMITTER_BEARER_TOKEN:
+            raise ValueError("BREAKPAD_STAGE_SUBMITTER_BEARER_TOKEN is not set.")
```
biancadanforth: I thought I'd raise an exception if we don't have the config specified in the environment, since this only runs either locally or in stage. If it isn't set, then we won't have prod crashes in stage to test with.

socorro/mozilla_settings.py
```diff
@@ -303,8 +303,16 @@ def _or_none(val):
     doc="URL for the version string API endpoint in the Crash Stats webapp.",
 )
 
-
 # Stage submitter configuration
+BREAKPAD_STAGE_SUBMITTER_BEARER_TOKEN = _config(
+    "BREAKPAD_STAGE_SUBMITTER_BEARER_TOKEN",
+    default="",
```
smarnach: If you don't provide a default here, it will error out if the value isn't provided. You can then remove your explicit check in `set_up()`.

smarnach (approved):

socorro/mozilla_settings.py
biancadanforth: Ah nice, thank you I'll do that instead.

socorro/stage_submitter/submitter.py
biancadanforth: Sven mentioned in another comment that if I exclude the `default` in the config for this, `everett` will throw an error, so I'll do that instead.

PR #7202, opened 2026-05-15
bug-1685143: Changed Auth0 Audience URL

Haseeb702:
Because:
- The `auditgroups` cronjob is raising a 403 error in the function `get_access_token`, meaning the Auth0 API is not able to query the users
- This error occurs because the `audience` parameter in the request payload points to the external domain behind a Fastly WAF (This ticket shows some context on the domain change in socorro: [CRINGE-135](https://mozilla-hub.atlassian.net/browse/CRINGE-135?atlOrigin=eyJpIjoiYTgwZWQ5OWE1ZGExNDU1M2E0OTJhMjA4NjAzOTk3OTIiLCJwIjoiaiJ9))
- The Auth0 management API only has a single identifier, thus we need to switch to using that single internal domain URL so that the Auth0 management API can generate a valid access token.

This PR:
- Adds a new config variable for the Auth0 management endpoint: `AUTH0_MANAGEMENT_API_ENDPOINT`.
- Updates the new variable to be used for the audience parameter in the `get_access_token` function.

### DIFF
--- a/docker/config/local_dev.env
+++ b/docker/config/local_dev.env
@@ -82,6 +82,7 @@ OIDC_RP_CLIENT_SECRET=bd01adf93cfb
 # Redirect for the browser which is running on the docker host
 OIDC_OP_AUTHORIZATION_ENDPOINT=http://localhost:8080/openid/authorize
 # Requests between the webapp container and the oidcprovider container
+AUTH0_MANAGEMENT_API_ENDPOINT=http://oidcprovider:8080/api/v2/
 OIDC_OP_TOKEN_ENDPOINT=http://oidcprovider:8080/openid/token
 OIDC_OP_USER_ENDPOINT=http://oidcprovider:8080/openid/userinfo
 
--- a/webapp/crashstats/authentication/management/commands/auditgroups.py
+++ b/webapp/crashstats/authentication/management/commands/auditgroups.py
@@ -42,12 +42,11 @@ def delta_days(since_datetime):
 
 def get_access_token(client_id, client_secret, domain, session):
     url = f"https://{domain}/oauth/token"
-    audience = f"https://{domain}/api/v2/"
     payload = {
         "client_id": client_id,
         "client_secret": client_secret,
         "grant_type": "client_credentials",
-        "audience": audience,
+        "audience": settings.AUTH0_MANAGEMENT_API_ENDPOINT,
     }
     response = session.post(url, json=payload)
     if response.status_code != 200:
--- a/webapp/crashstats/authentication/management/commands/auditgroups.py
+++ b/webapp/crashstats/authentication/management/commands/auditgroups.py
@@ -117,15 +117,13 @@ class Command(BaseCommand):
         # Go through the users and mark the ones for removal
         users_to_remove = []
         for user in hackers_group.user_set.all():
-            is_blocked = False
-            try:
-                is_blocked = is_blocked_in_auth0(user.email)
-            except RuntimeError as e:
-                self.stdout.write(f"Auth0 failed for: {user.email}: {e}")
-
-            # User may be blocked as a security mitigation. Eg: too many login attempts
-            if is_blocked:
-                users_to_remove.append((user, "user has most likely lost employment"))
+            if not settings.LOCAL_DEV_ENV:
+                try:
+                    # User may be blocked as a security mitigation. Eg: too many login attempts
+                    if is_blocked_in_auth0(user.email):
+                        users_to_remove.append((user, "user has most likely lost employment"))
+                except RuntimeError as e:
+                    self.stdout.write(f"Auth0 failed for: {user.email}: {e}")
 
             elif not user.is_active:
                 users_to_remove.append((user, "!is_active"))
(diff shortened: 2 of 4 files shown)


webapp/crashstats/settings/base.py
```diff
@@ -510,6 +510,7 @@ def filter(self, record):
 OIDC_OP_AUTHORIZATION_ENDPOINT = _config("OIDC_OP_AUTHORIZATION_ENDPOINT", default="")
 OIDC_OP_TOKEN_ENDPOINT = _config("OIDC_OP_TOKEN_ENDPOINT", default="")
 OIDC_OP_USER_ENDPOINT = _config("OIDC_OP_USER_ENDPOINT", default="")
+AUTH0_MANAGEMENT_API_ENDPOINT = _config("AUTH0_MANAGEMENT_API_ENDPOINT", default="")
```
biancadanforth: This has an empty string as a default value -- is it set in the environment for local dev? If not, I don't think socorro will start up locally (or at least, we won't be able to run this cron job locally). Can you test this? You'll likely need to add the config to `local_dev.env` as the other `OIDC*` config are there as well.

Also, I think this group makes sense to put this new config, but can you put it in alphabetical order?

Also also, I regret including "Auth0" in the variable name. Auth0 is a provider for the OIDC protocol standard. But we may change providers in the future, in which case this variable name won't make sense. I'm not going to make you change it; just something to keep in mind in the future for naming variables. Naming things is hard!

biancadanforth (dismissed):
Some high level feedback:
* Your commit subject isn't following our commit subject conventions per the Socorro docs. Ideally we'd have some kind of automated enforcement of this, but we don't. Can you update it?
* In terms of the substance of the PR title/commit message and PR description, you're not really explaining what the issue was, or what caused it.
  * This PR is a resource for the reviewer at the time the patch is written, but also an artifact for future reference and should accurately reflect what the issue was and why the change is needed. Just saying it's a different  and correct URL doesn't explain why the previous URL didn't work. I know in this case we don't want to mention the URL values explicitly, but we can still go a bit further in explaining the bug and the fix. If you have questions about the root cause or why this fixes it, please ask!

webapp/crashstats/settings/base.py
smarnach: I think the user management endpoint we are using is Auth0-specific and not a standard OIDC thing, so it seems fine to have Auth0 in the name.

Haseeb702:
This comment goes through the changes I made to the PR so far:
1. Added commit subject according to the Socorro docs
2. Improved the PR description
3. Put the config in `base.py` in alphabetical order
4. Added config to `local_dev.env`

docker/config/local_dev.env
```diff
@@ -82,6 +82,7 @@ OIDC_RP_CLIENT_SECRET=bd01adf93cfb
 # Redirect for the browser which is running on the docker host
 OIDC_OP_AUTHORIZATION_ENDPOINT=http://localhost:8080/openid/authorize
 # Requests between the webapp container and the oidcprovider container
+AUTH0_MANAGEMENT_API_ENDPOINT=http://oidcprovider:8080/api/v2/
```
Haseeb702, line 85: For the `AUTH0_MANAGEMENT_API_ENDPOINT` in `local_dev.env`, It made the most sense to set the URL to the `oidcprovider` container since that is where the other OIDC endpoint URLs that request from Auth0 point to (such as `OIDC_OP_TOKEN/USER_ENDPOINT`).

Haseeb702: To provide some more context, the `auditgroups` cronjob was not able to run locally since the config var wasn't defined locally, which is why I had to add the endpoint to `local_dev.env`.

webapp/crashstats/authentication/management/commands/auditgroups.py
```diff
@@ -118,15 +117,15 @@ def audit_hackers_group(self, dryrun=True):
...
-            is_blocked = False
-            try:
-                is_blocked = is_blocked_in_auth0(user.email)
-            except RuntimeError as e:
-                self.stdout.write(f"Auth0 failed for: {user.email}: {e}")
-
-            # User may be blocked as a security mitigation. Eg: too many login attempts
-            if is_blocked:
-                users_to_remove.append((user, "user has most likely lost employment"))
+            if not settings.LOCAL_DEV_ENV:
```
Haseeb702, line 120: While running the `auditgroups` cronjob locally I've been getting an `SSL ERROR` from the `get_access_token` function. After debugging, I found out this is due to the function sending an `https` request to the local `oidcprovider` which can only accept `http` requests ([link](https://github.com/mozilla/docker-test-mozilla-django-oidc)). Since the `Auth0` API request can not be run locally, it would be best to skip the API call in local dev. This is similar to what we have implemented in Tecken for the `Auth0` call ([link](https://github.com/mozilla-services/tecken/blob/b8fa8c474723dbbc25d6a0068967b11ffcc8a86b/tecken/useradmin/middleware.py#L87)). To implement this in Socorro, I used an existing variable `LOCAL_DEV_ENV` from `base.py` in the cronjob to determine if we are in local dev before attempting the API call.

smarnach (approved):

PR #7213, opened 2026-06-11
bug-2038160: ensure authorized users can use _columns on protected fields in SuperSearch

biancadanforth:
Because:
* My fix for bug-2038160 regressed some SuperSearch API behavior. Namely, users with `view_pii` permissions were previously able to get protected fields back in the API response with the `_columns` list-of-fields parameter, but now they are not.

This PR:
* Adds a test case for this scenario and restores the original API behavior. I've added a [more detailed explanation](https://bugzilla.mozilla.org/show_bug.cgi?id=2038160#c9) of the fix in the ticket.

### DIFF
--- a/webapp/crashstats/api/tests/test_views.py
+++ b/webapp/crashstats/api/tests/test_views.py
@@ -434,7 +434,6 @@ class TestSuperSearch(BaseTestViews):
                 "_aggs.signature",
                 "_histogram.date",
                 "_sort",
-                "_columns",
             )
             for key in restricted_params:
                 if key in params:
@@ -449,7 +448,7 @@ class TestSuperSearch(BaseTestViews):
                         "signature": "abcdef",
                         "product": "WaterWolf",
                         "version": "1.0",
-                        "url": "http://embarassing.website.com",
+                        "url": "http://embarrassing.website.com",
                         "user_comments": "hey I am thebig@lebowski.net",
                     }
                 ],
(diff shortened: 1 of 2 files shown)


webapp/crashstats/api/tests/test_views.py
```diff
@@ -488,6 +487,38 @@ def mocked_get(**params):
...
+    @mock.patch("crashstats.supersearch.models.SuperSearch.get_implementation")
+    def test_api_with_view_pii(self, mock_implementation):
+        def mocked_get(**params):
+            assert "url" in params["_columns"]
+            assert "signature" in params["_columns"]
+            return {
+                "hits": [
+                    {
+                        "signature": "abcdef",
+                        "url": "http://embarassing.website.com",
```
smarnach: There's a typo in "embarrassing", but the same typo exists in other places in this file. I guess it doesn't matter.

smarnach:
I can't reproduce the original bug. I just created a new API token with "View PII" permission, assigned it to the `CRASHSTATS_API_TOKEN` env variable and ran
```
supersearch --_columns=uuid --_columns=url
```
and it works just fine. How can I reproduce the bug, so I can verify it's fixed?

biancadanforth:
> I can't reproduce the original bug. I just created a new API token with "View PII" permission, assigned it to the `CRASHSTATS_API_TOKEN` env variable and ran
> 
> ```
> supersearch --_columns=uuid --_columns=url
> ```
> 
> and it works just fine. How can I reproduce the bug, so I can verify it's fixed?

We discussed this a bit over Slack. For some reason, gsvelto and I can reproduce the error with the command above, but Sven can't unless he runs the command inside a Docker container.

The way I tested my patch was to do TDD, where I wrote the test first, it failed, made the change and then the test passed, but it would be good to manually verify as well.

I just verified this manually with the following steps:
1. Reproduce the error in prod  via `supersearch --_columns=uuid --_columns=url` (with a valid `CRASHSTATS_API_TOKEN` with `view_pii` permissions).
```
$ supersearch --_columns=uuid --_columns=url
Usage: supersearch [OPTIONS]
Try 'supersearch --help' for help.

Error: url: no data
```
2. Check out this feature branch locally and start up Socorro, including the webapp and processor.
3. Fetch a crash report from prod via `just shell` and then `socorro-cmd fetch_crashids --num=1 | bin/process_crashes.sh`.
4. Create an API token locally with the `view_pii` permission.
5. Run `CRASHSTATS_API_TOKEN=<api_token> supersearch --host=http://localhost:8000 --_columns=uuid --_columns=url`.

Results: This was for a random crash report in prod with a crash id of `27a5f887-7fa1-4a52-93f8-f90c80260610`:

```
$ CRASHSTATS_API_TOKEN=<api_token> supersearch --host=http://localhost:8000 --_columns=uuid --_columns=url
27a5f887-7fa1-4a52-93f8-f90c80260610	<url_value>
```

smarnach:
It turns out I was running a locally modified version of crashstats-tools which was using the SuperSearchUnredacted endpoint for some reason. I can reproduce the issue with an unmodified version, and I can confirm that this patch correctly fixes the issue. I tested it against a locally running version of Socorro with two crash reports processed.

smarnach (approved):
Looks good to me.

webapp/crashstats/api/tests/test_views.py
biancadanforth: TIL: Embarrassing has two 'r's!

PR #7214, opened 2026-06-11
bug-2046410: added exception to get_type_replacement()

Haseeb702:
Purpose:
- Fixes the regression caused by bug-2031036
- During the recent refactor of the `collapse()` function, the original exception function which caught all of the exceptions was replaced. The new exception function `get_type_replacement()` did not have the exception for the `" in "` edge case.
- This regression caused signatures such as `<unknown in ntdll.pdb>` to be incorrectly collapsed and returned as `<T>` which is not the intended function, we want the signature to be returned unchanged

This PR:
- Adds the `" in "` string as an exception in the `get_type_replacement()` function that is used by `collapse_types()`
- Adds a test for the `<unknown in ntdll.pdb>` edge case to confirm it remains unchanged due to the exception


smarnach (approved):
Looks good to me.

PR #7216, opened 2026-06-12
bug-2037932: Set crashing thread for shutdownhang to 0

Haseeb702:
Purpose:
- A fix for bug-2037932
- Fixes an issue where the `"Bugzilla" > "Core"` tab bug form gets pre-filled with the watchdog thread, instead of the main thread (thread 0) for `shutdownhangs`.
- We want to ensure the bug form gets pre-filled with thread 0 (the main thread) by setting the `processed_crash.crashing_thread` to be 0. This will also populate the `details` page with the main thread stack frame, allowing us to remove the code that manually sets the crashing thread to 0.

This PR:
- Adds a post signature processing rule called `ShutDownHangCrashingThreadRule()` to `mozilla.py` which checks if a signature starts with `shutdownhang`. If it does, the processed crash crashing thread is set to 0.
- Adds test cases for the new `ShutDownHangCrashingThreadRule()` in `test_mozilla.py`
- Removes the redundant code in `crashstats/views.py` that sets the `details` page crashing thread to 0 and removes the test case that is associated with it.

### DIFF
--- a/socorro/mozilla_rulesets.py
+++ b/socorro/mozilla_rulesets.py
@@ -126,9 +126,10 @@ DEFAULT_RULESET = [
     MemoryReportExtraction(),
     SoftErrorsRule(),
     ThreadCountRule(),
-    ShutDownHangCrashingThreadRule(),
     # generate signature now that we've done all the processing it depends on
     SignatureGeneratorRule(),
+    # change the crashing thread if it's a shutdownhang
+    ShutDownHangCrashingThreadRule(),
 ]
 
 
--- a/socorro/mozilla_rulesets.py
+++ b/socorro/mozilla_rulesets.py
@@ -59,6 +59,7 @@ from socorro.processor.rules.mozilla import (
     ThemePrettyNameRule,
     TopMostFilesRule,
     UtilityActorsNameRule,
+    ShutDownHangCrashingThreadRule,
 )
 
 
@@ -125,6 +126,7 @@ DEFAULT_RULESET = [
     MemoryReportExtraction(),
     SoftErrorsRule(),
     ThreadCountRule(),
+    ShutDownHangCrashingThreadRule(),
     # generate signature now that we've done all the processing it depends on
     SignatureGeneratorRule(),
 ]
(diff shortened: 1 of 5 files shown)


socorro/processor/rules/mozilla.py
```diff
@@ -1280,3 +1280,20 @@ def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
...
+
+    Crashes with a `shutdownhang` signature always return the watchdog thread as
+    the crashing thread. But, we want the crashing thread to be 0 since its stack
+    frame reveals the actual reason for the crash.
+
+    Bug #2037932
+
+    """
+
+    def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
```
Haseeb702, line 1296: This rule sets the `"crashing_thread"` to 0 if the signature in the processed crash starts with `"shutdownhang"`.

socorro/mozilla_rulesets.py
```diff
@@ -127,6 +128,8 @@
     ThreadCountRule(),
     # generate signature now that we've done all the processing it depends on
     SignatureGeneratorRule(),
+    # change the crashing thread if it's a shutdownhang
+    ShutDownHangCrashingThreadRule(),
```
Haseeb702, line 132: This rule relies on the `SignatureGeneratorRule()`, since we require a `shutdownhang` in the signature to ensure the crash is a `shutdownhang`

webapp/crashstats/crashstats/views.py
```diff
@@ -170,11 +170,6 @@ def report_index(request, crash_id, default_context=None):
         parsed_dump = {}
 
     context["crashing_thread"] = context["report"].get("crashing_thread")
-    if context["report"]["signature"].startswith("shutdownhang"):
```
Haseeb702, line 173: This was the old code that set `shutdownhang` to 0 in the `details` page. This is not necessary anymore as the crashing thread is set to 0 in the processed crash now.

smarnach (approved):
This looks good to me. I tested this locally by processing a shutdownhang crash and a non-shutdownhang crash, and everything worked as expected.

PR #7231, opened 2026-07-08
CRINGE-38: Replace metrics-graphics dependency

Haseeb702:
Purpose:
- To replace the metrics-graphics dependency with a more modern and actively maintained Javascript graphing library: [CRINGE-38](https://mozilla-hub.atlassian.net/browse/CRINGE-38)

This PR:
- Removes all the old dependencies: `D3` and `metrics-graphics`, since they are both used only for the signature graph tab and can be safely removed.
- Removed the custom CSS for `metrics-graphics`
- Rewrites the `signature_tab_graphs.js` using `chart.js`, which is a modern Javascript graphing library to replace the `metrics-graphics` dependency and the old graphing code

Manual Testing Plan:
- Go to any crash signature report and click on the graphs tab
- Verify that the line graph renders properly, and loads within 2 seconds
- Select a different option in the dropdown menu and verify that a new line graph is drawn with the new option
- Hover your cursor near a line and confirm that data points can be hovered over and display the y-axis and x-axis details
- Resize your window to verify if the graph is scaled correctly


webapp/crashstats/signature/static/signature/js/signature_tab_graphs.js
```diff
@@ -67,74 +66,85 @@ SignatureReport.GraphsTab.prototype.loadControls = function () {
 SignatureReport.GraphsTab.prototype.formatData = function (data) {
```
Haseeb702, line 66: First thing to refactor was the data format, since `chart.js` doesn't accept the same data format as `metrics-graphics`.

webapp/crashstats/signature/static/signature/js/signature_tab_graphs.js
```diff
@@ -67,74 +66,85 @@ SignatureReport.GraphsTab.prototype.loadControls = function () {
...
   $.each(termCounts.splice(0, 4), function (i, element) {
-    lineDataObject[element.term] = [];
+    lineDataObject[element.term] = {
+      label: escapeHTML(element.term),
+      data: [],
+    };
   });
 
   // Each object in data.aggregates contains data for one date.
   $.each(data.aggregates, function (i, dateData) {
```
Haseeb702, line 88: Here we go through the top 4 terms being graphed and create a list of daily crash counts that align with the date list. The code iterates through the data for each date and maps the daily crash counts to each term. This is how the dataset for the line graph is built.

webapp/crashstats/signature/static/signature/js/signature_tab_graphs.js
```diff
@@ -67,74 +66,85 @@ SignatureReport.GraphsTab.prototype.loadControls = function () {
...
+  function escapeHTML(str) {
+    let tmpDiv = document.createElement('div');
+    tmpDiv.textContent = str;
+    return tmpDiv.innerHTML;
+  }
+
   $.each(termCounts.splice(0, 4), function (i, element) {
-    lineDataObject[element.term] = [];
+    lineDataObject[element.term] = {
+      label: escapeHTML(element.term),
```
biancadanforth: Chart.js draws legend and tooltip labels onto a `<canvas>` element [via `fillText`](https://github.com/chartjs/Chart.js/blob/9c5cf9fac7ec04a71b516e2aff3f7d76876be369/src/plugins/plugin.legend.js#L368), which does not interpret HTML, so we no longer need to escape HTML here. If there are field values that contain special characters, they would be displayed as e.g. `&amp;` for `&`. Can you just pass in `element.term` here directly?

I don't think there are any other uses of this `escapeHTML` function, so it can probably be removed as well.

webapp/crashstats/signature/static/signature/js/signature_tab_graphs.js
```diff
@@ -145,27 +155,35 @@ SignatureReport.GraphsTab.prototype.drawGraph = function (graphData, contentElem
...
-    utc_time: true,
-    interpolate: d3.curveLinear,
-    area: false,
-    legend: graphData.legend,
-    legend_target: '.new-legend',
-    show_secondary_x_label: false,
-    mouseover: function (d) {
-      $('.mg-active-datapoint', contentElement).text(d.term + ': ' + d.count + (d.count === 1 ? ' crash' : ' crashes'));
+  // Draw the graph on the graphElement using chart.js
+  new Chart(graphElement, {
```
biancadanforth: You didn't touch the `signature_panel.js` file, but that's where the delete panel functionality lives (the red X in the top right corner of each chart on the page):
```js
  $deleteButton.on('click', function (e) {
    e.preventDefault();
    that.$panelElement.remove();
    // Check that onDelete is not undefined
    if (onDelete) {
      onDelete();
    }
  });
```

When a user deletes a panel, the `<canvas>` element is removed from the DOM, but the chart instance (`new Chart(...)`) is never destroyed. This leaves the reference in Chart.js, and the `ResizeObserver`/any event listeners still around. That means every deleted panel leaks memory (if there are references to an object around, then the data can't be garbage collected).

Before re-rendering a `new Chart` below, I would call `destroy`. To do this, you need to have a reference to the current chart and make it available to the event listener. E.g. something like:
```js
var chart = new Chart(...);
contentElement.data('chart', chart);
```
And then in the `deleteButton` click handler:
```js
$deleteButton.on('click', function (e) {
  e.preventDefault();
  var chart = that.$contentElement.data('chart');
  if (chart) {
    chart.destroy();
  }
  that.$panelElement.remove();
  if (onDelete) { onDelete(); }
});
```

webapp/package.json
biancadanforth: Per the comment, it looks like this can be removed now? Just need to make sure everything still builds correctly without it.

webapp/crashstats/signature/static/signature/css/signature_report.css
biancadanforth: I think this can also be removed?

biancadanforth (commented):
Thanks for the PR Haseeb! I'm still reviewing this, but I thought I'd add the comments so far until I can do some more testing/review tomorrow.

alexcottner:
Screenshots below for reference.

The new graphing library appears to perform well and adjust for date ranges based on the data it is asked to render. The main difference that I can find is that it treats null values as 0's. I don't think this is a bad thing and definitely improves the visibility of data where there are only 1-2 points of data for a given grouping or when there are large gaps between values.


**Before:**
Total physical memory
image

Install age
image


Cpu info
image


**After:**
Total physical memory
image


Install age
image


Cpu info
image

webapp/crashstats/signature/static/signature/js/signature_tab_graphs.js
```diff
@@ -67,74 +66,85 @@ SignatureReport.GraphsTab.prototype.loadControls = function () {
...
-        };
-
-        // ... And add it to this term's data array.
-        lineDataObject[termData.term].push(nodeData);
+      currentDateCount[termData.term] = termData.count;
+    });
+
+    // Check if each top 4 element contains crashes for the current date
+    $.each(lineDataObject, function (element, dataObject) {
+      var crashCount = 0;
```
alexcottner: Ah so this is where we are setting the series values to 0 if we don't have any data for that period. I think this is a reasonable thing to do, and would have improved the old graphs if we had done it.

webapp/crashstats/signature/static/signature/js/signature_tab_graphs.js
```diff
@@ -67,74 +66,85 @@ SignatureReport.GraphsTab.prototype.loadControls = function () {
...
+
+      if (Object.prototype.hasOwnProperty.call(currentDateCount, element)) {
+        crashCount = currentDateCount[element];
+      } else {
+        crashCount = 0;
```
alexcottner: This else is redundant as you default to 0 on line 114.
You could also do a ternary like...
```js
let crashCount = Object.prototype.hasOwnProperty.call(currentDateCount, element) ? currentDateCount[element] : 0;
```

webapp/crashstats/signature/static/signature/js/signature_tab_graphs.js
```diff
@@ -67,74 +66,85 @@ SignatureReport.GraphsTab.prototype.loadControls = function () {
...
-          term: termData.term,
-        };
-
-        // ... And add it to this term's data array.
-        lineDataObject[termData.term].push(nodeData);
+      currentDateCount[termData.term] = termData.count;
+    });
+
+    // Check if each top 4 element contains crashes for the current date
+    $.each(lineDataObject, function (element, dataObject) {
```
alexcottner, line 105: nit: I do prefer normal loops over jquery's each function here. We don't really need a callback function in this case and a normal for loop should be faster.

alexcottner (approved):
I think we should do a follow-up and refactor `signature_tab_graphs.js` once this is merged. I'm pretty sure we can cut a lot of code out of here, but it might be annoying to review while also changing the underlying graphing lib.

webapp/crashstats/signature/static/signature/js/signature_tab_graphs.js
alexcottner: or even...
```js
let crashCount = currentDateCount[element] || 0;
```

PR #7236, opened 2026-07-24
bug-2050638: allow querying protected fields for authorized users in SuperSearch

biancadanforth:
Because:
* Users with `view_pii` permissions should be able to get protected data back from the `SuperSearch` API.

This PR:
* First commit:
  * Refactors and renames `sanitize_list_of_fields_params` to `sanitize_params`, and changes its scope to handle sanitizing all of the variants of query parameters and their values that could contain protected fields and values. Previously, it only sanitized list-of-fields query parameters like `_sort`, `_columns`, or `_facet`. Updates the tests accordingly.
    * I went back and forth on making `all_fields` an optional param, since the /search view sanitizes a subset of fields already, but I ultimately decided to keep the function simpler and err on always sanitizing everything, since the extra work of calling `get_all_fields` is cheap in this case.
* Second commit:
  * Removes the public fields only filter in `self.possible_params` and `self.extended_fields` from `SuperSearch.__init__`, which doesn't have a user to be permissions-aware.
  * Passes `self.api_user` (obtained from the API request) into `get_allowed_fields` in `SuperSearch.get`, so that it considers the user's permissions in getting the list of allowed fields rather than returning only public fields.
  * Removes the public fields only filter for `_histogram.*` fields added to `self.extended_fields` in `SuperSearch.get`.
  * Collectively, this means that the set of possible parameters going into the newly refactored `sanitize_params` in `SuperSearch.get` include both public and protected fields that are then pared down based on the now permissions-aware set of `allowed_fields` in `sanitize_params`.
  * Adds new test cases for the additive behavior and some tests to ensure we don't regress existing `SuperSearch` API behavior.

### DIFF
--- a/webapp/crashstats/supersearch/libsupersearch.py
+++ b/webapp/crashstats/supersearch/libsupersearch.py
@@ -2,12 +2,11 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-from dataclasses import dataclass
 import datetime
+from dataclasses import dataclass
 
 from socorro import settings as socorro_settings
 from socorro.libclass import build_instance_from_settings
-
 
 # Map of processed crash schema permissions to webapp permissions
 PROCESSED_CRASH_TO_WEBAPP_PERMISSIONS = {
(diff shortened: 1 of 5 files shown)


webapp/crashstats/supersearch/libsupersearch.py
```diff
@@ -138,22 +137,51 @@ def permissions_condition(field):
...
 
+    # Drop disallowed filters and aggregation/histogram params that name a field
+    # in the param key itself.
+    for key in list(params):
+        if key.startswith(("_aggs.", "_histogram.", "_histogram_interval.")):
+            # e.g. `_aggs.url`, `_aggs.product.version`, `_histogram.url`
+            referenced = key.split(".")[1:]
+            if any(name not in allowed_fields_set for name in referenced):
+                del params[key]
+        elif key in all_fields and key not in allowed_fields_set:
```
smarnach: If I understand correctly, this code only removes fields if they are in `all_fields`, and all fields that aren't in that list are left alone. This essentially renders this a blacklist approach rather than a whitelist approach.

It's hard to tell for me whether this makes it possible to work around the filtering, e.g. by URL-quoting field names, using different capitalization, adding leading spaces or similar. Any of these approaches would result in field names that are not in `all_fields`, so they wouldn't be removed.

I don't know whether it's possible to find a modification that would actually be accepted by Elasticsearch, but my point is that it's really hard to tell whether this is secure. I'd prefer a proper whitelist approach, where we only accept field names we know, if this is at all feasible.

smarnach (commented):
Thanks for taking this on! This is a difficult part of the code to reason about, since it's complex and it has many layers that are relatively tightly coupled, so there's a lot to keep in mind at once. I feel I'd need to learn a lot more about how SuperSearch works under the hood to be fully certain about this, but I _think_ this approach is correct. I added a suggestion how we could make the sanitization more robust, which may not be feasible, but if it is it would make me feel better about the approach.

I like the thorough and easy-to-read tests – they are a good documentation of the intended behaviour.

webapp/crashstats/supersearch/libsupersearch.py
biancadanforth: Thank you Sven for taking a look and this excellent observation.

This `elif` check with `all_fields` is specifically covering filters where the param name is the field name (e.g. `url=...`). You're right that this check is a blocklist check and is not comprehensive. The other checks for the other cases function more like an allowlist, so it is a good idea to do the same here, and it is possible to do so.

FWIW, it does look like we still drop anything outside of an exact match for the equivalent of `all_fields` downstream[1], but we shouldn't implicitly rely on that to catch things that might slip through the cracks.

I'll update the tests and code accordingly.

[1]: We [build the ES query filters from all of the SuperSearch fields in `fields`](https://github.com/mozilla-services/socorro/blob/1a5a411f23c1a5b7546cd405b4e892fe19c2b2a0/socorro/external/es/search_common.py#L105-L108) (`_fields = self.all_fields = get_supersearch_fields()`), and we only pass SuperSearch API filters on if they are [an exact match for the field name in that set of all the SuperSearch fields](https://github.com/mozilla-services/socorro/blob/1a5a411f23c1a5b7546cd405b4e892fe19c2b2a0/socorro/external/es/search_common.py#L157-L158).

webapp/crashstats/supersearch/libsupersearch.py
```diff
@@ -138,22 +143,47 @@ def permissions_condition(field):
...
 
-    :returns: the mutated ``params`` dict, with disallowed entries
-    removed
+    :returns: the mutated ``params`` dict, with disallowed entries removed
     """
     allowed_fields_set = set(allowed_fields)
 
+    # Drop disallowed filters and aggregation/histogram params that name a field
+    # in the param key itself.
+    for key in list(params):
```
smarnach, line 178: You could add a comment why `list(params)` is required here instead of the usual `for key in params`.

webapp/crashstats/supersearch/libsupersearch.py
```diff
@@ -138,22 +143,47 @@ def permissions_condition(field):
...
+        if key.startswith(("_aggs.", "_histogram.", "_histogram_interval.")):
```
smarnach, line 179: Here I was worried before that there might be other aggregations not listed here. With the new allowlist approach, they will fall through to the `elif` branch and get removed.

webapp/crashstats/supersearch/models.py
```diff
@@ -148,13 +147,13 @@ def _get_extended_params(self):
 
         return tuple(extended_fields)
 
-    def get(self, **kwargs):
-        # Sanitize all parameters listing fields and make sure no private data
-        # is requested.
+    def get(self, dont_cache=False, refresh_cache=False, **kwargs):
```
smarnach, line 150: `dont_cache=False` hurts my brain, but that isn't the fault of this PR.

smarnach (approved):
Thanks for the update! This is easier to reason about locally now, and I feel confident enough to approve it. I added just one optional request to add a comment to the code.

PR #7238, opened 2026-07-29
CRINGE-38: Refactor jquery code

Haseeb702:
Purpose:
- After changing the underlying graphing library, we wanted to improve readability in the `signature_tab_graph` file by modernizing some of the old Javascript code

This PR:
- Replaces the Jquery `.each()` loops with more modern javascript `.map()` and `for...of` loops, which makes the code easier to read.
- These changes were made on parts of the code that contribute to drawing the graph. Since we have no test cases for our front end, we would want to manually test the graphs to ensure that they are working as intended.

How to test:
- Load some crash data into your local Socorro
- Go to any crash signature report and click on the graphs tab
- Verify that the line graph renders properly, and loads within 2 seconds
- Select a different option in the dropdown menu and verify that a new line graph is drawn with the new option
- Hover your cursor near a line and confirm that data points can be hovered over and display the y-axis and x-axis details
- Resize your window to verify if the graph is scaled correctly

### DIFF
--- a/webapp/crashstats/signature/static/signature/js/signature_tab_graphs.js
+++ b/webapp/crashstats/signature/static/signature/js/signature_tab_graphs.js
@@ -36,14 +36,14 @@ SignatureReport.GraphsTab.prototype.loadControls = function () {
   var fields = $('#mainbody').data('fields');
 
   // Append an option element for each field.
-  for (var field of fields) {
-    that.$selectElement.append(
+  that.$selectElement.append(
+    fields.map((f) =>
       $('<option>', {
-        value: field.id,
-        text: field.text,
+        value: f.id,
+        text: f.text,
       })
-    );
-  }
+    )
+  );
 
   // Append the control elements.
   this.$controlsElement.append(this.$selectElement, $('<hr>'));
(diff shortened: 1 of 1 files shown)


webapp/crashstats/signature/static/signature/js/signature_tab_graphs.js
```diff
@@ -36,14 +36,14 @@ SignatureReport.GraphsTab.prototype.loadControls = function () {
   var fields = $('#mainbody').data('fields');
 
   // Append an option element for each field.
-  $.each(fields, function (i, field) {
+  for (var field of fields) {
```
alexcottner: nit: jquery's `append` can take an array. So instead of running `append` one at a time in a loop and possibly causing multiple re-renders, you could do it all at once. Something like this should work...

```js
that.$selectElement.append(
  fields.map((f) => {
    $('<option>', {
      value: f.id,
      text: f.text,
    })
  })
);
```

Note: I didn't run this so there is a reasonable chance I have a typo or syntax error.

alexcottner (approved):
This is a good cleanup, have one potential suggestion.

PR #7249, opened 2026-08-14
bug-2052944: fall back to crash annotation values for platform info when there's no minidump

biancadanforth:
Because:
* Critical path consumers facet on some minidump-derived fields (see [Bug 2055875](https://bugzilla.mozilla.org/show_bug.cgi?id=2055875#c10)).
* Crash pings have no minidump (i.e. there's no `json_dump` in the processed crash), so we need to get this same information from elsewhere to continue using these SuperSearch queries with crash pings.
* Bug-2049749 adds five new crash annotations to that end, so that this data can come from there instead.
* Socorro needs to add these crash annotations and ensure the existing processed crash fields are populated correctly in all cases.

This PR:
* Adds these five new crash annotations to Socorro's raw crash schema.
* Updates descriptions in the processed crash schema to mention that we prefer deriving the field's value from crash reports to preserve original behavior but fall back to using the value from a source annotation.
* Creates a new `PlatformInfoRule` and updates the existing `OSPrettyVersionRule` to add the fallback behavior.
* Adds tests for the new behavior.

Notes:
- Annotation names, descriptions, bug, and data review links are pulled from [one of the patches](https://phabricator.services.mozilla.com/D315029) attached to bug-2049749. The information here should be finalized once bug-2049749 is done.
	- I have a test case that asserts the processor rule changes here are a no-op in the case that we receive a crash submission missing these annotations entirely. This means we can deploy this Socorro change whenever we want without consideration for when bug-2049749 lands (other than making sure the schema info is correct first. Of course, this needs to be true anyway, since we receive a long tail of submissions from older clients.).
- Each of these new annotations represent an existing public processed crash field in Socorro: `cpu_arch`, `cpu_info`, `os_name`, `os_version` and `os_pretty_version`, so I am designating each of the corresponding new crash annotations in the raw crash schema as public.
- All of these existing processed crash fields are already indexed for crash reports, so we don't need to make any changes to `SUPERSEARCH_FIELDS` for consumers to query them when we change the default dataset to crash pings in the future.


socorro/tests/processor/rules/test_mozilla.py
```diff
@@ -1855,6 +1857,131 @@ def test_aurora_channel(self, tmp_path):
         assert status.notes == []
 
 
+def run_cpu_and_os_rules(raw_crash, dumps, processed_crash, tmpdir, status):
```
biancadanforth, line 1860: The unit tests here for the new PlatformInfoRule aren't pure unit tests for the test cases that rely on the order/output of other rules to verify correctness. I use the output from the preceding rule(s) as input into the new PlatformInfoRule. That way, if those implementations later change upstream, we can make sure that we don't regress this processing rule.

That means `test_mozilla.py` now imports from `rules.general` and `rules.android`, which is a new cross-module dependency. Maybe there's a better place for these tests that depend on the order in which specific rules execute?

smarnach: I think this is fine. In my experience, black box integration testing is generally more useful than "pure" unit tests. Unit tests tend to require mocking and implicit assumptions that may not hold in future versions of the code.

Will spent significant effort to get rid of mocking in the tests as much as possible, which essentially is a move towards integration testing. I'm all in favour of moving further in that direction.

socorro/processor/rules/mozilla.py
```diff
@@ -882,6 +880,41 @@ def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
...
+    FIELDS = [
+        ("cpu_arch", "CPUArchitecture", "unknown"),
+        ("cpu_info", "CPUInfo", "unknown"),
+        ("os_name", "OS", "Unknown"),
+        ("os_version", "OSVersion", ""),
+    ]
+
+    def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
+        for key, annotation, sentinel in self.FIELDS:
+            current = processed_crash.get(key) or ""
```
smarnach, line 907: Style nit: I'd personally prefer the two-argument form of `dict.get()` here.
```suggestion
            current = processed_crash.get(key, "")
```

socorro/processor/rules/mozilla.py
```diff
@@ -882,6 +880,41 @@ def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
...
+            if current and current != sentinel:
+                # An earlier rule got a real value from the minidump or from an
+                # Android annotation, so leave it alone.
+                continue
+
+            value = (raw_crash.get(annotation) or "").strip()
```
smarnach, line 913: Style nit: I'd personally prefer the two-argument form of `dict.get()` here.
```suggestion
            value = raw_crash.get(annotation, "").strip()
```

socorro/processor/rules/mozilla.py
```diff
@@ -882,6 +880,41 @@ def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
...
+    processing unchanged.
+
+    """
+
+    # (processed crash field, annotation, sentinel the minidump-based rule sets
+    # when it has no value)
+    FIELDS = [
+        ("cpu_arch", "CPUArchitecture", "unknown"),
+        ("cpu_info", "CPUInfo", "unknown"),
+        ("os_name", "OS", "Unknown"),
```
smarnach, line 901: These sentinel values are quite specific. Do they come from minidump-stackwalk? If the mdsw code changes, e.g. using lowercae  "unknown" for `os_name` as well, will we notice somehow?

socorro/mozilla_rulesets.py
```diff
@@ -115,6 +115,9 @@
     DistributionIdRule(),
     OSInfoRule(),
     AndroidOSInfoRule(),
+    # This rule must run after (Android)CPUInfoRule and (Android)OSInfoRule,
+    # but before OSPrettyVersionRule.
```
smarnach, line 119: Out of scope comment: I wish we could declare rule dependencies more formally, and then use topological sorting to figure out the correct order automatically.

smarnach (approved):
Looks good to me! I added a couple of stylistic comments and a question about the sentinel values.

socorro/processor/rules/mozilla.py
biancadanforth: They're the `glom(..., default=...)` values in CPUInfoRule and OSInfoRule when minidump-stackwalk when they're missing from its output. It looks like the inconsistency in case ("Unknown" specifically) is Socorro's own inconsistency. I looked into changing that, but it looks like some of our operators for ES queries are case sensitive. I'd rather not make changes to what we're writing in for that value as part of this PR.

Since `TestPlatformInfoRule` derives its inputs by running the real `CPUInfoRule`/`OSInfoRule`/Android rules, we would catch this. Changing a `glom` default breaks `test_no_minidump_uses_annotations`. The sentinels used in `PlatformInfoRule` and these other upstream rules have to match for that test case to pass. This wouldn't have been true if these were pure unit tests.

socorro/mozilla_rulesets.py
biancadanforth: Yes that is a really good idea. I filed a stub ticket for this in the optional follow-up milestone for the project and referenced this PR: https://bugzilla.mozilla.org/show_bug.cgi?id=2064407.

biancadanforth:
While [Bug-2049749](https://bugzilla.mozilla.org/show_bug.cgi?id=2049749) is still in review, I have confirmed with Alex Franchuk that the parts of that work that this bug depends on will not change (namely, which annotations are being added and their descriptions), so I'm going to go ahead and merge this PR.

PR #7251, opened 2026-08-19
bug-2052944: add Glean client ID as a protected field

biancadanforth:
Because:
* We need the Glean client ID from crash pings to process Firefox telemetry data deletion requests as required per [Bug 2052950](https://bugzilla.mozilla.org/show_bug.cgi?id=2052950#c4).

This commit:
* Adds the Glean client ID to the raw and processed crash schema, linking them via the `source_annotation` which ensures the field is copied into the processed crash via the `CopyFromRawCrashRule` in the processor.
* Indexes the new field as a keyword to ensure we can search for exact matches when processing data deletion requests.

Notes:
* As indicated in the raw crash schema changes, I obtained a `datareview+` [in the ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=2052944#c8).
* Since this field is only used for processing data deletion requests, I did not explicitly add it to any of the Crash Stats UIs.
* We have general tests that cover a lot of the mechanisms that I get "for free" in this patch: copying from the source annotation, asserting the field gets indexed, redacting the field in specific cases, ... I decided to add a couple of new tests specifically for this field to cover the critical processor and search behavior, though there is some overlap with existing tests. I couldn't find a good, recent reference patch for this class of new field (protected, based on a source annotation).

### DIFF
--- a/socorro/external/es/super_search_fields.py
+++ b/socorro/external/es/super_search_fields.py
@@ -884,6 +884,7 @@ FIELDS = {
         "query_type": "bool",
         "storage_mapping": {"type": "boolean"},
     },
+    "glean_client_id": keyword_field(name="glean_client_id"),
     "gmp_library_path": {
         "data_validation_type": "str",
         "form_field_choices": [],
--- a/socorro/processor/rules/mozilla.py
+++ b/socorro/processor/rules/mozilla.py
@@ -904,13 +904,13 @@ class PlatformInfoRule(Rule):
 
     def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
         for key, annotation, sentinel in self.FIELDS:
-            current = processed_crash.get(key) or ""
+            current = processed_crash.get(key, "")
             if current and current != sentinel:
                 # An earlier rule got a real value from the minidump or from an
                 # Android annotation, so leave it alone.
                 continue
 
-            value = (raw_crash.get(annotation) or "").strip()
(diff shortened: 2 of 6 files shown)


socorro/external/es/super_search_fields.py
```diff
@@ -884,6 +884,7 @@ def apply_schema_properties(fields, schema):
         "query_type": "bool",
         "storage_mapping": {"type": "boolean"},
     },
+    "glean_client_id": keyword_field(name="glean_client_id"),
```
biancadanforth: I set this as `keyword` instead of `text_field`, because we only care about exact matches. We don't need to look for substrings or regexps.

socorro/processor/rules/mozilla.py
```diff
@@ -904,13 +904,13 @@ class PlatformInfoRule(Rule):
...
         for key, annotation, sentinel in self.FIELDS:
-            current = processed_crash.get(key) or ""
+            current = processed_crash.get(key, "")
             if current and current != sentinel:
                 # An earlier rule got a real value from the minidump or from an
                 # Android annotation, so leave it alone.
                 continue
 
-            value = (raw_crash.get(annotation) or "").strip()
+            value = raw_crash.get(annotation, "").strip()
```
biancadanforth, line 913: These are nit changes from the first PR, #7249 , that I forgot to commit before merging.

smarnach (approved):
Looks good to me.

PR #7263, opened 2026-09-04
bug-2053525: Add support for the CrashEventID annotation.

smarnach:
This change adds support for the new annotation by adding the required metadata to the schemata and to the field data for Supersearch.

I read through the pipeline and rules tests to look for a good place to unit test this, but came to the conclusion that this isn't required. This change isn't adding any code; it relies on the `CopyFromRawCrash` rule that we [test with simplified schemata](https://github.com/mozilla-services/socorro/blob/2c0152ee1f9c45149bd15ceef36093f0cebc5003/socorro/tests/processor/rules/test_mozilla.py#L194). It doesn't seem helpful to add a separate test for each field that's copied.

I manually tested the change by locally processing a few recent crash reports. I verified that the crash event id is copied into the processed crash with the right key and that I could search for crashes by crash event id.


smarnach:
We are still discussing whether to make the crash event id public immediately on Slack.

smarnach:
The consensus on Slack was that this field should be public, and that the designation as `data-classification-low` in https://phabricator.services.mozilla.com/D286321#10050855 makes an additional data review unnecessary, so I changed the visibility of the annotation and the processed crash field to public.

biancadanforth (approved):
LGTM. Keyword is the right call for Elasticsearch, just as for `glean_client_id` in #7251. I saw the [thread](https://mozilla.slack.com/archives/C0BC72FC39V/p1788445688144279) with gsvelto and afranchuk about making it public in Slack.

PR #7264, opened 2026-09-07
bug-2068008: Fix control flow bug in the auditgroups cron job.

smarnach:
As described in the bug, the `if` sequence got broken in PR #7202, since in production only the `not settings.LOCAL_DEV_ENV` is alays true, so none of the `elif` branches will ever be executed.

I fixed this by moving the `LOCAL_DEV_ENV` check into `is_blocked_in_auth0()`, since that makes the control flow more similar between the dev env and production.

There is no good way to unit test this problem, unfortunately. We have tests for all the behaviours except the Auth0 check, and they all work in the tests because the tests run with `LOCAL_DEV_ENV = True`.

### DIFF
--- a/webapp/crashstats/authentication/management/commands/auditgroups.py
+++ b/webapp/crashstats/authentication/management/commands/auditgroups.py
@@ -69,6 +69,8 @@ def find_users(client_id, client_secret, domain, email, session):
 
 
 def is_blocked_in_auth0(email):
+    if settings.LOCAL_DEV_ENV:
+        return False
     session = session_with_retries(total_retries=5)
     users = find_users(
         settings.OIDC_RP_CLIENT_ID,
@@ -77,10 +79,7 @@ def is_blocked_in_auth0(email):
         email,
         session,
     )
-    for user in users:
-        if user.get("blocked"):
-            return True
-    return False
+    return any(user.get("blocked") for user in users)
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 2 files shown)


webapp/crashstats/authentication/management/commands/auditgroups.py
```diff
@@ -117,17 +116,17 @@ def audit_hackers_group(self, dryrun=True):
...
-                    self.stdout.write(f"Auth0 failed for: {user.email}: {e}")
-
-            elif not user.is_active:
+            try:
+                # User may be blocked as a security mitigation. Eg: too many login attempts
+                if is_blocked_in_auth0(user.email):
+                    users_to_remove.append(
+                        (user, "user has most likely lost employment")
+                    )
+                    continue
```
biancadanforth: IIUC: This `continue` is needed so that we don't double count a blocked user that is also e.g. inactive.

This is a bit harder to read for me than keeping all of the checks as mutually exclusive at the same top level as before. I.e. something like:

```python
try:
    # User may be blocked as a security mitigation. Eg: too many login attempts
    is_blocked = is_blocked_in_auth0(user.email)
except RuntimeError as e:
    self.stdout.write(f"Auth0 failed for: {user.email}: {e}")
    is_blocked = False

if is_blocked:
    users_to_remove.append((user, "user has most likely lost employment"))
elif not user.is_active:
    users_to_remove.append((user, "!is_active"))
elif not self.is_employee_or_exception(user):
```

Your version is functionally equivalent, so this is a style preference. Not blocking.

webapp/crashstats/authentication/management/commands/auditgroups.py
```diff
@@ -69,6 +69,8 @@ def find_users(client_id, client_secret, domain, email, session):
 
 
 def is_blocked_in_auth0(email):
+    if settings.LOCAL_DEV_ENV:
```
biancadanforth, line 72: Maybe add a comment explaining that the Auth0 service isn't available locally, and without this, the cron job would crash?

biancadanforth (approved):
LGTM. I spent a bit of time trying to think how we could add a test here, but they all involved setting LOCAL_DEV_ENV to `False` in the test, and then that was moved into the function, so it wouldn't work for the auditgroups test anyway... Also, the test env is `LOCAL_DEV_ENV` of `True`, so it's confusing to say otherwise, even temporarily in a test case. Trying not to get too deep here for this fix.

Nit: Looks like you could remove the monkeypatch to set LOCAL_DEV_ENV to `False` in `test_user_blocked_in_auth0_is_removed`, since that moved into the `is_blocked_in_auth0` function.

webapp/crashstats/authentication/management/commands/auditgroups.py
smarnach: I considered that as well and couldn't really decide, but if you prefer the other version I'll go with that.

I also changed the `except` clause to `except Exception`, since we want to continue the loop no matter what.

smarnach:
> LGTM. I spent a bit of time trying to think how we could add a test here, but they all involved setting LOCAL_DEV_ENV to `False` in the test, and then that was moved into the function, so it wouldn't work for the auditgroups test anyway... Also, the test env is `LOCAL_DEV_ENV` of `True`, so it's confusing to say otherwise, even temporarily in a test case. Trying not to get too deep here for this fix.

I considered basically the same things and came to the same conclusion. Not everything can be prevented by tests, and I think the new structure with the `if` condition moved into the function is a good defence against this accidentally happening again.

> Nit: Looks like you could remove the monkeypatch to set LOCAL_DEV_ENV to `False` in `test_user_blocked_in_auth0_is_removed`, since that moved into the `is_blocked_in_auth0` function.

You are right, did that.

PR #7265, opened 2026-09-09
bug-2053871: Add support for the submission_type field

smarnach:
This adds the submission type as a public field. The code currently assumes that Antenna is not adding the field to the raw crash in all cases, so we need to validate it and I needed to add a custom rule. If we change Antenna to unconditionally add the field, and overwrite anything that was potentially set by the client, we can replace the custom rule with the generic copying rule.

The field will be public. In general, it will be possible to tell from the included annotations whether something is a ping or a report, so it doesn't really add information, but rather just convenience. It's also not directly coming from the client, but is only added in Antenna. I don't think we need a data review for this.


smarnach:
I tested this locally by processing a few crash reports. All of them got a `submission_type` field in the processed crash, and filtering by the field in Supersearch worked as expected, and showed the expected choices.

socorro/processor/rules/mozilla.py
```diff
@@ -237,6 +237,22 @@ def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
...
+    """Determine the submission_type field value - either "report" or "ping".
+
+    This looks at the submission_type field in the raw crash that is added by the
+    collector. If the field is missing in the raw crash or has an invalid value,
+    it defaults to "report" in the processed crash.
+    """
+
+    def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
+        submission_type = raw_crash.get("submission_type", "report")
+        if submission_type not in ["report", "ping"]:
```
biancadanforth, line 250: Nit: Maybe convert to lowercase before performing this check? Unlike most of the other raw crash fields, we control the input value here, so it's less important than usual...

socorro/processor/rules/mozilla.py
```diff
@@ -237,6 +237,22 @@ def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
...
+            status.add_note("invalid submission type in raw crash")
```
biancadanforth, line 251: Perhaps also note that we are defaulting to "report" as a result? I do think that's the right call, since this goes into effect much sooner than we're actually submitting crash pings to Socorro.

socorro/processor/rules/mozilla.py
```diff
@@ -267,18 +283,6 @@ def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
...
-    """Fix SubmittedFromInfobar annotation values to "1" """
-
-    def predicate(self, raw_crash, dumps, processed_crash, tmpdir, status):
-        return "SubmittedFromInfobar" in raw_crash and raw_crash[
-            "SubmittedFromInfobar"
-        ] in ("true", True)
-
-    def action(self, raw_crash, dumps, processed_crash, tmpdir, status):
-        raw_crash["SubmittedFromInfobar"] = "1"
-
```
biancadanforth, line 280: Ah, this rule is not invoked at all in our processor ruleset. I noticed the `SubmittedFromInfobar` annotation is no longer in CrashAnnotations.yaml, and our own `SubmittedFromRule` says the corresponding processed crash field can be removed. I don't expect to follow this all the way down to the source in this ticket of course! If we start looking, I'm sure we will find many such examples of this sort of drift. I filed https://mozilla-hub.atlassian.net/browse/CRINGE-436 to capture that.

socorro/schemas/raw_crash.schema.yaml
```diff
@@ -40,6 +40,18 @@ properties:
     type: object
     permissions: ["protected"]
 
+  submission_type:
+    description: |
+      Takes the values "report" or "ping" depending on whether this crash item
+      represents a submission as a crash report or a crash ping. The same crash
+      event can be submitted as both a report and a ping.
```
biancadanforth: > The same crash event can be submitted as both a report and a ping.

Maybe mention something like "as indicated by the `CrashEventID` annotation"?

socorro/tests/processor/rules/test_mozilla.py
```diff
@@ -476,6 +477,30 @@ def test_object_value_with_dict(self, tmp_path):
...
 
+class TestSubmissionTypeRule:
+    @pytest.mark.parametrize(
+        "submission_type, expected_submission_type",
+        [
+            ("report", "report"),
+            ("ping", "ping"),
+            ("", "report"),
+            (None, "report"),
+            ("glonk", "report"),
```
biancadanforth, line 488: 😆

biancadanforth (approved):
LGTM. I have some minor suggestions. After re-reading the ticket description I wrote ages ago, I agree it does warrant having its own rule for validation, since CopyFromRawCrashRule only sets a default if the field is absent, not if it's garbage or otherwise some value that's neither `"report"` nor `"ping"`.

Edit: The processed crash schema already supports a default on `source_annotation` fields, but `CopyFromRawCrashRule` only applies it when the annotation is absent and in no other case. `enum` is supported by the validator, so we could also update the `CopyFromRawCrashRule` to support it as an alternative method.

socorro/processor/rules/mozilla.py
smarnach: I noticed that this rule wasn't imported in mozilla_rulesets.py. I searched for the class name and the only hit is the class definition. This case felt clear-cut enough to deal with in an otherwise unrelated PR.

Looking into fields we can remove makes sense as well. There are so many things that could be cleaned up.

socorro/processor/rules/mozilla.py
smarnach: This note is stored in the `processing_notes` field on the processed crash, and it's not surfaced anywhere in the web interface unless you explicitly show it in Supersearch results. We are the main consumers of this field, so we can see whether certain kinds of processing issues are happening frequently. The main purpose of the message is making it easy to find the relevant source code quickly. It will also be obvious from looking at the processed crash what it defaulted to, and it's possible to see the offending invalid value by looking at the raw crash.

socorro/processor/rules/mozilla.py
smarnach: Sounds good, though I won't add a test case for any mixed-case patterns.

I think "we" control the input values for all raw crash fields, since they are generally emitted by our crash reporting client. Of course people can submit random junk to our crash reporting endpoint, but we have no obligation to support anything other than the values that are emitted by our crash reporter, so we don't really need to be permissive about what inputs we accept.

PR #7266, opened 2026-09-11
bug-2053863: Add StackTraces annotation to raw crash schema.

smarnach:
As noted on the bug, for the time being we can only add the `StackTraces` annotation to the raw schema. Once we have support for the symbolicate-stacks binary in our pipeline, we can consider if any information from the `StackTraces` annotation is missing, and decide whether to add extracting that information to symbolicate-stacks or to Socorro itself.

I refrained from making this public, since it doesn't seem necessary, and since it's hard to tell whether all information the annotation might contain can be public. I couldn't find any data review for the annotation, but I asked on the bug, so maybe someone finds a link I could add there.


smarnach:
We were unable to find a data review for the annotation, so I'll leave it as "unknown".

socorro/schemas/raw_crash.schema.yaml
```diff
@@ -1069,6 +1069,15 @@ properties:
     type: string
     permissions: ["public"]
 
+  StackTraces:
+    data_reviews:
+      - unknown
+    description:
+      Stack traces extracted from the crash minidump, if available, by the
+      crash reporting client. A serialized JSON object.
```
biancadanforth: Nit: The convention for these schema in the README is to use a folded or literal block scalar indicator (`>` or `|`), and it's also a little less fragile than a plain scalar. I think `>` is the right call here based on that. Can you update it?

The [upstream description](https://searchfox.org/firefox-main/rev/6c74efe2fcddf84b6f320959064a66946b4a1759/toolkit/crashreporter/CrashAnnotations.yaml#1294-1296) in `CrashAnnotations.yml` says:
> Stack traces extracted from the crash minidump, if available. A serialized JSON object (with a structure matching the associated Glean object description).

It may be useful to reference that the structure is documented in the corresponding `StackTraces` entry in the upstream CrashAnnotations.yaml file. We do something like this for CPUArchitecture (though that's from rust-minidump).

biancadanforth (approved):
This turned out to be a very small PR that probably could have been rolled up into another bug (i.e. bug-2070650 -- though that bug is pretty big already).

I thought briefly we may want to add the field to the UI, but this is handled automatically for raw crash fields in the Crash Annotations tab, and while there are some exceptions, the Details tab is typically for processed crash fields.

socorro/schemas/raw_crash.schema.yaml
smarnach: Right, I missed the `>` here. Thanks for catching this!

I had originally linked the upstream description here, but wasn't sure whether it would trip up the linter to have a line longer than 80 characters, and I also wasn't sure how useful it is, since it's the case for all JSON fields that their structure is defined in the crash annotations schema, so it doesn't feel necessary to explicitly say so for each of them.

I added the link back, and it turns out the linter is smart enough to notice that the line can't be broken up any further, so it doesn't complain, and having the link definitely won't hurt.

